PR #7852, opened 2016-09-02
Fixed git commit and branch info in ComponentVersion

lutovich:
Previously git commit and branch exported same property and incorrect git commit property name was used.

**Note:** this PR should be null-merged forward. there will be separate PR for 2.3 because whole `ComponentVersion` thingy is different there.


burqen:
@lutovich Looks good. Gonna let it build before merge.

PR #7853, opened 2016-09-02
Cleanup git commit and branch info in kernel manifest file

lutovich:
Added default `unknown-commit` value if env variable is not set, removed profile that exported git branch because value was never used.


burqen:
`env.GIT_BRANCH` should also be removed from the team city builds
e.g. "Packaging" build "Export Git Information" step

PR #7854, opened 2016-09-03
Don't misleadingly report http port for https port conflict

craigtaverner:
When a server cannot start due to HTTPS port conflict, it reported the error was in HTTP port conflict. The code in question does not have an easy way to determine if the underlying jetty exception was due to http or https, so we just report both ports if https is set, and http only otherwise.

changelog: [server] Fixed misleading error message on https port conflict


benbc:
@craigtaverner This looks good from my point of view. Thanks for fixing it.

PR #7855, opened 2016-09-04
Explicit listen and advertised addresses for Bolt and HTTP.

apcj:
Separates the two concerns of:

A. Choosing a network interface and port on which to listen for
   incoming connections: "listen address", and
B. Advertising the address which a client should use to connect
   to the server: "advertised address".

Background:

A web server has no need to advertise its address; the server
always responds according its identity from the client's perspective.

However, when there are multiple connectors available, or
multiple servers in a cluster, an advertised address is required.

Before this commit, Bolt and HTTP connectors tried to deduce their
advertised address from a single "address" setting that was also
used for binding. This made it difficult to support network
configurations where there was a difference in DNS between client
and server, or where a proxy was used.

This distinction was already in place for core-edge clusters,
but this commit extends the concept to Bolt and HTTP.

Configuration changes:

All connectors now have a pair of properties "listen_address"
and "advertised_address". The "address" is deprecated, and for
backwards compatibility its value is interpreted as "listen_address".

advertised_address need not be specified; by default it derives
its value from listen address. Therefore there is no additional
configuration required unless using the new ability to override
the default advertised address.

Additionally, there are some defaults to further reduce
configuration. Both listen_address and advertised_address
properties can be specified as just a port without a hostname
or address, e.g. ":7474" in which case the hostname part is
inherited from global settings dbms.connectors.default_listen_address
or dbms.connectors.default_advertised_hostname respectively.

This means that when opening up a server from its default
local-only configuration, there is only one property that needs
updating to 0.0.0.0, and all connectors will be open.

Implementation notes:

Two tiny types ListenSocketAddress and AdvertisedSocketAddress
are used to easily keep track of how the respective settings are
used. These classes mostly replace HostnamePort in the affected
modules.


community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -502,11 +512,18 @@ private static String defaultPageCacheMemory()
...
+            "To listen for connections on all interfaces, use \"0.0.0.0\". " +
+            "To bind specific connectors to a specific network interfaces, " +
+            "specify the listen_address properties for the specific connector.")
+    public static final Setting<String> default_listen_address =
+            setting( "dbms.connectors.default_listen_address", STRING, "localhost" );
 
-    @Description("Hostname/IP address and port that we can be connected to be the driver.")
-    public static final Setting<HostnamePort> bolt_advertised_address =
-            setting( "bolt_advertised_address", HOSTNAME_PORT, "localhost:7687" );
+    @Description("Default hostname or IP address the server uses advertise itself its connectors. " +
```
benbc: A couple of typos in this line.

benbc:
Why is it necessary to move the HTTP connector config definition down into the Kernel?

community/server/src/main/java/org/neo4j/server/rest/discovery/DiscoveryService.java
```diff
@@ -28,35 +28,37 @@
...
     @Produces( MediaType.APPLICATION_JSON )
     public Response getDiscoveryDocument() throws URISyntaxException
     {
-        String managementUri = configuration.get( ServerSettings.management_api_path ).getPath() + "/";
-        String dataUri = configuration.get( ServerSettings.rest_api_path ).getPath() + "/";
-        String boltAddress = configuration.get( GraphDatabaseSettings.bolt_advertised_address ).toString();
+        String managementUri = config.get( ServerSettings.management_api_path ).getPath() + "/";
+        String dataUri = config.get( ServerSettings.rest_api_path ).getPath() + "/";
+        String boltAddress = boltConnectors( config ).stream().findFirst()
+                .map( boltConnector -> config.get( boltConnector.advertised_address ).toString() ).orElse( null );
```
benbc, line 61: Any reason not to propagate the `Optional` to avoid the null-check?

packaging/neo4j-desktop/src/main/resources/org/neo4j/desktop/config/neo4j-default.conf
```diff
@@ -10,32 +10,39 @@ dbms.directories.import=import
...
+# To accept non-local connections, uncomment this line:
+#dbms.connectors.default_listen_address=0.0.0.0
+
+# You can also choose a specific network interface, and configure a non-default
+# port for each connector, by setting their individual listen_address.
+
+# Neo4j advertises the addresses its connectors.
+# To specify the hostname or IP address of this server used in those
+# adverised addresses, uncomment this line:
+#dbms.connectors.default_advertised_hostname=localhost
```
benbc: The "uncomment this line" approach is reserved for cases where the line _doesn't_ contain the default. I think it does here, doesn't it?

Also "adverised".

packaging/neo4j-desktop/src/main/resources/org/neo4j/desktop/config/neo4j-default.conf
```diff
@@ -10,32 +10,39 @@ dbms.directories.import=import
...
+# individual advertised_address.
+
 # Bolt connector
-#
 dbms.connector.bolt.type=BOLT
 dbms.connector.bolt.enabled=true
 dbms.connector.bolt.tls_level=OPTIONAL
-# To have Bolt accept non-local connections, uncomment this line:
-# dbms.connector.bolt.address=0.0.0.0:7687
+dbms.connector.bolt.listen_address=:7687
```
benbc: Isn't this the default? Should be commented out.

packaging/neo4j-desktop/src/main/resources/org/neo4j/desktop/config/neo4j-default.conf
```diff
@@ -10,32 +10,39 @@ dbms.directories.import=import
...
 
-#
 # HTTP Connector
-#
 dbms.connector.http.type=HTTP
 dbms.connector.http.enabled=true
-#dbms.connector.http.encryption=NONE
-# To have HTTP accept non-local connections, uncomment this line:
-#dbms.connector.http.address=0.0.0.0:#{default.http.port}
+dbms.connector.https.encryption=NONE
```
benbc: "https"

packaging/neo4j-desktop/src/main/resources/org/neo4j/desktop/config/neo4j-default.conf
```diff
@@ -10,32 +10,39 @@ dbms.directories.import=import
...
+dbms.connector.http.listen_address=:#{default.http.port}
```
benbc: More stuff that should be commented out?

benbc:
My comments to one config file apply to the others as well

benbc:
One benefit of the config migrators is that they log when deprecated settings are used. Can you reproduce that with your _ad hoc_ migration here?

benbc:
There are docs changes to go along with this, of course.

packaging/neo4j-desktop/src/main/resources/org/neo4j/desktop/config/neo4j-default.conf
apcj: The default value is "localhost", which is consistent with the default values of the individual listen addresses before this change: https://github.com/neo4j/neo4j/pull/7855/files#diff-49289564650a8e514c0639aa114a7f60R520

mneedham: @benbc  are you suggesting we take this one out?

benbc: No. I'm just saying that it's incorrect to say "to specify the hostname [...] uncomment this line". Uncommenting the line gives the default (localhost). To be consistent with how we document settings elsewhere in the file you should just describe the setting in that comment.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -502,11 +512,18 @@ private static String defaultPageCacheMemory()
...
+            "To advertise a specific hostname or IP address for a specific connector, " +
+            "specify the +advertised_address+ property for the specific connector.")
+    public static final Setting<String> default_advertised_hostname =
```
jjaderberg: Should it be `default_advertised_address` here, as it is in the `@Description` and in the connector-specific settings?

PR #7840, opened 2016-08-31
Removed pointless comparison

pontusmelke:
Code was comparing two things that never could be equal,
an instance with a type.

PR #7856, opened 2016-09-05
Fix comparison bug due to refactoring

Mats-SX:


pontusmelke:
@Mats-SX duplicate of https://github.com/neo4j/neo4j/pull/7840

PR #7857, opened 2016-09-05
Make `packaging` a default submodule

Mats-SX:
The `packaging` module has only been incorporated as a proper submodule through activation of the `-DfullBuild` maven flag. This commit adds it to the list of default submodules, which also decouples it from the `manual` module, also activated by the mentioned flag.

I threw in removal of a profile that seems obsolete. That commit can be removed from this PR if it is inappropriate.


benbc:
@Mats-SX This is fine by me as long as it still builds successfully.

PR #7860, opened 2016-09-05
Repeat backup test on ComException

davidegrohmann:


davidegrohmann:
Subsumed by https://github.com/neo4j/neo4j/pull/7876

PR #7863, opened 2016-09-05
Uses RecordCursor in ReadNodeRecordsByCacheStep

tinwelint:
for more efficient reading of records.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ReadNodeRecordsByCacheStep.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ReadNodeRecordsByCacheStep.java
@@ -57,7 +57,7 @@ public class ReadNodeRecordsByCacheStep extends AbstractStep<NodeRecord[]>
     public void start( int orderingGuarantees )
     {
         super.start( orderingGuarantees );
-        recordCursor.acquire( 0, RecordLoad.CHECK );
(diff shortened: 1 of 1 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ReadNodeRecordsByCacheStep.java
```diff
@@ -80,7 +95,9 @@ public void run()
         @Override
         public void change( long nodeId, ByteArray array )
         {
-            batch[cursor++] = nodeStore.getRecord( nodeId, nodeStore.newRecord(), RecordLoad.CHECK );
+            boolean inUse = recordCursor.next( nodeId );
+            assert inUse;
```
lutovich: Would it make sense to acquire cursor in `#start()` with `RecordLoad.NORMAL` and skip this assertion?

lutovich:
@tinwelint lgtm, one small question

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ReadNodeRecordsByCacheStep.java
tinwelint: Yeah, that's much better, I'll change it

tinwelint:
@lutovich question resulted in a change for the better

PR #7864, opened 2016-09-05
Correct import page size in test environments

tinwelint:
by using same calculation as to match the small page cache size


community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -532,7 +534,8 @@ public int denseNodeThreshold()
             @Override
             public int pageSize()
             {
-                return pageSize;
+                // Let's call super if not specifically configured, so that proper calculations can be made
+                return pageSize == -1 ? super.pageSize() : pageSize;
```
burqen: Would be nice with a variable for this magic number.

tinwelint:
@burqen constant introduced, reads a bit better now

burqen:
Looks good, ready for merge

PR #7882, opened 2016-09-06
NeoStoreFileListing now use StorageEngine to list store files

burqen:
And also include meta data about the files, such as StoreType and recordSize.

co-author: @chrisvest

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
@@ -23,6 +23,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -77,7 +78,7 @@ public class NeoStoreFileListing
 
     private void gatherNonRecordStores( Collection<StoreFileMetadata> files, boolean includeLogs )
     {
-        for ( File file : storeDir.listFiles() )
+        for ( File file : Objects.requireNonNull( storeDir.listFiles()) )
         {
             if ( file.getName().equals( IndexConfigStore.INDEX_DB_FILE_NAME ) )
             {
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
@@ -23,17 +23,21 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.neo4j.graphdb.Resource;
 import org.neo4j.graphdb.ResourceIterator;
-import org.neo4j.helpers.collection.Iterators;
 import org.neo4j.kernel.api.labelscan.LabelScanStore;
 import org.neo4j.kernel.impl.api.LegacyIndexProviderLookup;
 import org.neo4j.kernel.impl.api.index.IndexingService;
 import org.neo4j.kernel.impl.index.IndexConfigStore;
 import org.neo4j.kernel.impl.store.MetaDataStore;
+import org.neo4j.kernel.impl.store.format.RecordFormat;
 import org.neo4j.kernel.spi.legacyindex.IndexImplementation;
+import org.neo4j.storageengine.api.StorageEngine;
+import org.neo4j.storageengine.api.StoreFileMetadata;
 
 import static java.util.Arrays.asList;
 import static org.neo4j.helpers.collection.Iterators.resourceIterator;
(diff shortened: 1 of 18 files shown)


tinwelint:
It has a test failure

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListingTest.java
```diff
@@ -20,238 +20,120 @@
...
         // Given
-        filesInStoreDirAre( STANDARD_STORE_DIR_FILES, STANDARD_STORE_DIR_DIRECTORIES );
-        ResourceIterator<File> scanSnapshot = scanStoreFilesAre( new String[]{"blah/scan.store", "scan.more"} );
-        ResourceIterator<File> indexSnapshot = indexFilesAre( new String[]{"schema/index/my.index"} );
-        NeoStoreFileListing fileListing = newFileListing();
+        LabelScanStore labelScanStore = mock( LabelScanStore.class );
+        IndexingService indexingService = mock( IndexingService.class );
+        LegacyIndexProviderLookup legacyIndexes = mock( LegacyIndexProviderLookup.class );
+        when( legacyIndexes.all() ).thenReturn( Collections.emptyList() );
+        File storeDir = mock( File.class );
```
chrisvest, line 179: You need this mock to produce some reasonable output on calls to `listFiles()`, as the test will otherwise run into an NPE.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -66,77 +75,56 @@ public NeoStoreFileListing( File storeDir, LabelScanStore labelScanStore,
                 new MultiResource( asList( labelScanStoreSnapshot, schemaIndexSnapshots, legacyIndexSnapshots ) ) );
     }
 
-    private Resource gatherLegacyIndexFiles( Collection<File> files ) throws IOException
+    private void gatherNonRecordStores( Collection<StoreFileMetadata> files, boolean includeLogs )
+    {
+        for ( File file : storeDir.listFiles() )
```
chrisvest: The `File.listFiles()` is actually allowed to return `null` if the given file is not a directory, or if an IO error occurs. We need to check for that before iterating. I think a `null` return value should be treated the same as an empty array in this case.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -77,7 +78,7 @@ public NeoStoreFileListing( File storeDir, LabelScanStore labelScanStore, Indexi
 
     private void gatherNonRecordStores( Collection<StoreFileMetadata> files, boolean includeLogs )
     {
-        for ( File file : storeDir.listFiles() )
+        for ( File file : Objects.requireNonNull( storeDir.listFiles()) )
```
chrisvest: This is functionally equivalent to the previous code, because `Objects.requireNonNull` just throws NPE if the argument is `null`. I don't think it matters very much, though.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -66,77 +76,56 @@ public NeoStoreFileListing( File storeDir, LabelScanStore labelScanStore,
...
-                {
-                    targetFiles.add( dbFile );
-                }
-                else if ( includeTransactionLogs && transactionLogFile( name ) )
-                {
-                    targetFiles.add( dbFile );
-                }
-            }
-        }
-        targetFiles.add( neostoreFile );
```
tinwelint, line 123: The (now deleted) method used to add the `neostore` file last of all the stores. This is iirc to guard for some cases of a semi-copied store, where the existence of the neostore file would signify that a complete copy of the store had been made. I don't see this behaviour above (and if indeed it is so then there's seemingly no test case for it either).

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -66,77 +76,56 @@ public NeoStoreFileListing( File storeDir, LabelScanStore labelScanStore,
...
-        for ( File dbFile : Objects.requireNonNull( storeDir.listFiles() ) )
-        {
-            String name = dbFile.getName();
-            if ( dbFile.isFile() )
-            {
-                if ( name.equals( MetaDataStore.DEFAULT_NAME ) )
-                {   // Keep it, to add last
-                    neostoreFile = dbFile;
-                }
-                else if ( neoStoreFile( name ) )
```
tinwelint, line 113: By the looks of it this (now deleted) method included both counts store files, whereas the new implementation above only includes the current one. Is that correct behaviour and do we know what behavioural differences that would introduce?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
chrisvest: The `StoreFile` enum dictates the order in `StorageEngine.listStorageFiles`, and indeed places the `neostore` last. It'd be good to add a test for this, though.

burqen:
@tinwelint @chrisvest Comments adressed.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
tinwelint: OK, but good anyway (not a regression then)

PR #7883, opened 2016-09-06
Terminate queries

boggle:
Add a new method to KernelTransactions that helps listing all transactions that are registered to execute a given cypher query (ExecutingQuery) and introduces new procedures that use this method to kill the transactions for some given queries.

Also includes some additional cosmetic changes:
- queryIds are now strings on the surface (i.e. what comes back/goes into procedures)
- moves transaction/connection/query management procedures over to kernel
- adds procedure descriptions 

changelog [procedures] Added procedures for listing and killing currently running queries


enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/BuiltInProcedures.java
```diff
@@ -54,56 +70,263 @@
...
+            throw new AuthorizationViolationException( PERMISSION_DENIED );
+        }
+        return enterpriseAuthSubject;
+    }
+
+    private EnterpriseAuthSubject ensureSelfOrAdminEnterpriseAuthSubject( String username )
+            throws InvalidArgumentsException
+    {
+        EnterpriseAuthSubject subject = EnterpriseAuthSubject.castOrFail( authSubject );
+        subject.ensureUserExistsWithName( username );
```
OliviaYtterbrink: Shouldn't this check be after the check for admin or self? Otherwise we could leak information about what users are available to a non admin user.

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/BuiltInProcedures.java
```diff
@@ -54,56 +70,263 @@
...
+    public Stream<ConnectionResult> terminateConnectionsForUser( @Name( "username" ) String username )
+            throws InvalidArgumentsException
+    {
+        EnterpriseAuthSubject subject = EnterpriseAuthSubject.castOrFail( authSubject );
+        if ( !subject.isAdmin() && !subject.hasUsername( username ) )
+        {
+            throw new AuthorizationViolationException( PERMISSION_DENIED );
+        }
+
+        subject.ensureUserExistsWithName( username );
```
OliviaYtterbrink: Could use `ensureSelfOrAdminEnterpriseAuthSubject` here as well

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/BuiltInProcedures.java
```diff
@@ -54,56 +70,263 @@
...
+            .filter( query -> isAdminEnterpriseAuthSubject() || authSubject.hasUsername( query.authSubjectName() ) )
+            .map( this::queryStatusResult );
+    }
+
+    @Description( "Kill all transactions executing the query with the given query id." )
+    @Procedure( name = "dbms.killQuery", mode = DBMS )
+    public Stream<QueryTerminationResult> killQuery( @Name( "id" ) String idText )
+            throws InvalidArgumentsException, IOException
+    {
+        long queryId = parseQueryId( idText ).kernelQueryId();
```
OliviaYtterbrink: Don't we want to check permissions for the user running this procedure?

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryId.java
```diff
@@ -0,0 +1,90 @@
...
+        try
+        {
+            if ( queryIdText.startsWith( QUERY_ID_PREFIX ) )
+            {
+                return new QueryId( Long.parseLong( queryIdText.substring( QUERY_ID_PREFIX.length() ) ) );
+            }
+        }
+        catch ( NumberFormatException e )
+        {
+            throw new IllegalArgumentException( e );
```
OliviaYtterbrink: Should use `InvalidArgumentsException` instead since that has status.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/StandardBoltConnectionTracker.java
```diff
@@ -53,7 +53,11 @@ public void onTerminate( ManagedBoltStateMachine machine )
     @Override
     public Set<ManagedBoltStateMachine> getActiveConnections( String owner )
     {
-        return sessions.entrySet().stream().filter( entry -> entry.getValue().equals( owner ) )
+        return sessions
+                .entrySet()
+                .stream()
+                .filter( entry -> entry.getValue()
+                .equals( owner ) )
```
fickludd: This line should probably join it's buddy above, in lack of anything important to say...

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/AuthProcedures.java
```diff
@@ -362,12 +276,13 @@ private StandardEnterpriseAuthSubject ensureAdminAuthSubject()
...
 
-    private StandardEnterpriseAuthSubject ensureSelfOrAdminAuthSubject( String username ) throws InvalidArgumentsException
+    private StandardEnterpriseAuthSubject ensureSelfOrAdminAuthSubject( String username )
+            throws InvalidArgumentsException
     {
         StandardEnterpriseAuthSubject subject = StandardEnterpriseAuthSubject.castOrFail( authSubject );
         subject.getUserManager().getUser( username );
 
-        if ( subject.isAdmin() || subject.doesUsernameMatch( username ) )
+        if ( subject.isAdmin() || subject.hasUsername( username ) )
```
fickludd, line 446: This should be reordered as Olivia said above. I guess
`if ( subject.hasUsername( username ) || 
     ( subject.isAdmin() && subject.getUserManager().getUser( username ) ) )` would cut it

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/BuiltInProcedures.java
```diff
@@ -54,56 +70,263 @@
...
     @Procedure( name = "dbms.listQueries", mode = DBMS )
     public Stream<QueryStatusResult> listQueries() throws InvalidArgumentsException, IOException
     {
-        return getKernelTransactions().executingQueries().stream().filter(
-                ( query ) -> isAdmin() || authSubject.doesUsernameMatch( query.authSubjectName() ) )
-                .map( this::queryStatusResult );
+        Set<ExecutingQuery> executingQueries = getKernelTransactions().executingQueries();
+        return executingQueries
+            .stream()
+            .filter( query -> isAdminEnterpriseAuthSubject() || authSubject.hasUsername( query.authSubjectName() ) )
```
henriknyman: query.authSubjectName() is using name() which is not the same as username()

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/BuiltInProcedures.java
```diff
@@ -54,56 +70,263 @@
...
+
+    private Stream<ExecutingQuery> executingQueriesWithId( long id, KernelTransactionHandle txHandle )
+    {
+        return txHandle.executingQueries().filter( q -> q.kernelQueryId() == id );
+    }
+
+    private QueryTerminationResult killQueryTransaction( Pair<KernelTransactionHandle, ExecutingQuery> pair )
+    {
+        ExecutingQuery query = pair.other();
+        if ( isAdminEnterpriseAuthSubject() || authSubject.hasUsername( query.authSubjectName() ) )
```
henriknyman: query.authSubjectName() is using name() which is not the same as username()

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryId.java
boggle: No, that exception is tied to the security package (same for status). Don't think that's appropriate to use here.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/StandardBoltConnectionTracker.java
boggle: Indeed, let's not keep them lonely :)

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryId.java
fickludd: I think we can move the `InvalidArgumentsException` and `Status.Security.InvalidArguments` to somewhere more appropriate then. They really have nothing in particular to do with security. Getting an exception with a status gives much nicer handling in driver/browser/logging.

community/kernel/src/main/java/org/neo4j/kernel/api/security/AuthSubject.java
```diff
@@ -54,6 +54,22 @@
...
+     */
+    boolean hasUsername( String username );
+
+    /**
+     * Ensure that the provided username is the name of an existing user known to the system.
+     *
+     * @param username a username
+     * @throws InvalidArgumentsException if the provided user name is not the name of an existing user
+     */
+    default void ensureUserExistsWithName( String username ) throws InvalidArgumentsException {
```
MishaDemianenko, line 68: I find 2 strange things with thin method here: 
- i think ensureUserExistsWithName should not be part of AuthSubject interface at all since it does not look like that AuthSubject should be responsible for creation of users
- since it's our first implementation of AuthSubject interface i believe that we should not declare it `default`. If we need some default behaviour it should be an abstract class instead

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/BuiltInProcedures.java
```diff
@@ -54,56 +70,258 @@
...
     {
         return resolver.resolveDependency( KernelTransactions.class );
     }
 
-    private boolean isAdmin()
+    // ----------------- helpers ---------------------
+
+    private Stream<TransactionTerminationResult> terminateTransactionsForValidUser( String username )
+    {
+        long terminatedCount = 0;
```
MishaDemianenko: hm looks like a place where stream api should be used to count terminatedCount instead of for loop

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
```diff
@@ -30,16 +30,21 @@
...
+        long queryId,
+        String userNameOrNull,
+        String queryText,
+        Map<String,Object> queryParameters,
+        long startTime
+    )
     {
         this.queryId = queryId;
-        this.authSubjectName = authSubjectName;
+        this.username = userNameOrNull;
```
MishaDemianenko: silent loss of can be null information, Optional instead?

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
```diff
@@ -69,14 +74,14 @@ public int hashCode()
         return (int) (queryId ^ (queryId >>> 32));
     }
 
-    public long queryId()
+    public long kernelQueryId()
```
MishaDemianenko: silent conversion from query id to kernelQueryId

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
```diff
@@ -201,9 +203,10 @@ final void forceClose()
         }
     }
 
-    final String authSubjectName()
+    final String username()
     {
-        return transaction.mode().name();
+        AccessMode mode = transaction.mode();
```
MishaDemianenko: i find it strange that kernel statement perform this dance, when we will have mode not an instance of AuthSubject?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
```diff
@@ -202,6 +205,24 @@ protected void dispose( KernelTransactionImplementation tx )
     }
 
     /**
+     * Give an approximate set of all transactions currently running together with associated metadata as
+     * computed by the provided selector function.
+     * This is not guaranteed to be exact, as transactions may stop and start while this set is gathered.
+     *
+     * @return the (approximate) set of open transactions.
+     */
+    public <T> Set<Pair<KernelTransactionHandle, T>> activeTransactions(
```
MishaDemianenko: do not think this method should be introduced in KernelTransaction at all:
 it's copy paste of active transactions method, it should not return too generic pair thingy, used only in build in procedures so mapping, if its necessary can be produced there

community/kernel/src/test/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContextTest.java
```diff
@@ -154,4 +154,78 @@ private void setUpMocks()
         when( dependencyResolver.resolveDependency( ThreadToStatementContextBridge.class ) ).thenReturn( statementContextBridge );
         when( dependencyResolver.resolveDependency( Guard.class ) ).thenReturn( guard );
     }
+
+    @SuppressWarnings( "ConstantConditions" )
+    @Test
+    public void rollsBackNewlyCreatedTransactionIfTerminationDetectedOnCloseDuringPeriodicCommit()
+    {
+        // Given
+        GraphDatabaseQueryService graph = mock( GraphDatabaseQueryService.class );
```
MishaDemianenko, line 163: i find this test extremely unreadable and unsupportable, only one thing you will do when it will fail - make mocks happy

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryId.java
```diff
@@ -0,0 +1,90 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.enterprise.builtinprocs;
+
+public final class QueryId
```
MishaDemianenko: Why do we need to represent query externally as query-_queryId_ ? would it be more simpler just to have an numeric id?

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryId.java
```diff
@@ -0,0 +1,90 @@
...
+{
+    public static final String QUERY_ID_PREFIX = "query-";
+    private final long kernelQueryId;
+
+    public static QueryId queryId( long queryId )
```
MishaDemianenko: i believe static factory methods here should be called _of_

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryId.java
MishaDemianenko: This exception missing error details and expected format. It would be strange to get illegal argument exception without explanation what format is actually expected from you

MishaDemianenko:
@boggle @systay i finish my first round of review

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryId.java
boggle: Moved InvalidArguments out of security and used it. This requires though catching it and rethrowing from all use in stream pipelines :/

community/kernel/src/main/java/org/neo4j/kernel/api/security/AuthSubject.java
boggle: This is unrelated to this PR. We're just moving code around here, I'll try to factor our the default implementation. This whole hierarchy needs some overhaul but out of scope for this PR.

boggle: I think factoring this out just complicates things uneccesarliy and having the default implementation generate the correct errors seems reasonable to me so I'm not going to touch it.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
boggle: Good catch, Optional it is.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
boggle: Always, AuthSubject extends AccessMode. This is in need of refactoring and hopefully that hopefully will happen when security support is added to core api. For now, we need this dance and this situation will arise in community.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
boggle: I'll pull it up.

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryId.java
boggle: we want to keep the door open for cluster wide ids and we're not sure longs will be good enough for that.

boggle:
@MishaDemianenko addressed feedback

community/kernel/src/test/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContextTest.java
boggle: That would be bad practice. If this fails, we should stop and think why.  I understand that this is not perfect yet it is the only good test we felt we could write here. We didn't find a better way to test the behavior we need to rely on in the built in procedures without resorting to heavy use of mocking or probabilistic (yuck) testing. How would you test this?

MishaDemianenko:
@boggle @systay do not have kernel questions anymore feel free to merge on green! thx

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
```diff
@@ -98,7 +104,7 @@ public long startTime()
     public String toString()
     {
         return format(
-                "ExecutingQuery{queryId=%d, authSubjectName='%s', queryText='%s', queryParameters=%s, startTime=%d}",
-                queryId, authSubjectName, queryText, queryParameters, startTime );
+            "ExecutingQuery{queryId=%d, username='%s', queryText='%s', queryParameters=%s, startTime=%d}",
+            queryId, username == null ? "<unavailable>" : username, queryText, queryParameters, startTime );
```
OliviaYtterbrink: Not using username as optional

enterprise/backup/data/databases/to/logs/debug.log
```diff
@@ -0,0 +1,598 @@
+2016-09-12 14:11:52.756+0000 INFO  [o.n.k.i.f.CommunityEditionModule] No locking implementation specified, defaulting to 'forseti'
```
OliviaYtterbrink: I don't believe these files should be here... ;p

PR #7888, opened 2016-09-07
Tweak cost model constants

systay:
Also updated `ActualCostCalculationTest` to add property filtering

Interesting changes:
- Index scans are as fast as label scans (from 3.0 to 1.0)
- Index seeks are tweaked from 3.0 to 1.9

Test output:

```
NodeIndexSeek: COST = 1.9243582552871088 * NROWS
NodeIndexScan: COST = 1.0081223395694658 * NROWS
NodeByID: COST = 6.183850510483064 * NROWS
AllNodeScan: COST = 1.207560834893734 * NROWS
RelByID: COST = 10.812422179582548 * NROWS
LabelScan followed by filter on property: COST = 1.9820059304920197 * NROWS
Expand: COST = 1.5051578949201065 * NROWS
NodeByLabelScan: COST = 1.0 * NROWS
```

### DIFF
--- a/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/planner/logical/CardinalityCostModel.scala
+++ b/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/planner/logical/CardinalityCostModel.scala
@@ -25,43 +25,44 @@ import org.neo4j.cypher.internal.frontend.v3_1.ast.{HasLabels, Property}
 
 object CardinalityCostModel extends CostModel {
 
-  /*
-   * These constants are approximations derived from test runs,
-   * see ActualCostCalculationTest
-   */
-  private val CPU_BOUND: CostPerRow = 0.1
-  private val FAST_STORE: CostPerRow = 1.0
-  private val SLOW_STORE: CostPerRow = 12.0
   private val PROBE_BUILD_COST: CostPerRow = 3.1
   private val PROBE_SEARCH_COST: CostPerRow = 2.4
   private val EAGERNESS_MULTIPLIER: Multiplier = 2.0
 
   private def costPerRow(plan: LogicalPlan): CostPerRow = plan match {
+    /*
+     * These constants are approximations derived from test runs,
+     * see ActualCostCalculationTest
+     */
 
-    case  _: NodeByLabelScan |
-          _: NodeIndexScan |
-          _: ProjectEndpoints
-    => FAST_STORE
+    case _: NodeByLabelScan |
+         _: NodeIndexScan |
+         _: ProjectEndpoints
+    => 1.0
 
     // Filtering on labels and properties
     case Selection(predicates, _) if predicates.exists {
       case _: Property | _: HasLabels => true
       case _ => false
     }
-    => FAST_STORE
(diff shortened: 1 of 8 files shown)


community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/planner/logical/CardinalityCostModel.scala
```diff
@@ -39,28 +39,30 @@ object CardinalityCostModel extends CostModel {
   private def costPerRow(plan: LogicalPlan): CostPerRow = plan match {
 
     case  _: NodeByLabelScan |
-         _: ProjectEndpoints
+          _: NodeIndexScan |
+          _: ProjectEndpoints
+    => FAST_STORE
```
pontusmelke: At some point we should remove FAST_STORE etc, because it is just confusing to mix them up with actual numbers

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/planner/logical/CardinalityCostModel.scala
```diff
@@ -39,28 +39,30 @@ object CardinalityCostModel extends CostModel {
...
+      case _: Property | _: HasLabels => true
+      case _ => false
+    }
     => FAST_STORE
 
     case _: AllNodesScan => 1.2
 
     case _: Expand |
-         _: VarExpand  => 2.0
+         _: VarExpand  => 0.5
```
pontusmelke: The data you posted suggested 1.5?

systay: 1.5 is the cost including the labelscan. Since we have labelscans at 1.0, I thought this was a good number. Is that sound science?

pontusmelke: ...or wait no thumbs up I think. The cost for the scan is already being considered

``` scala
private def expandResult(graph: GraphDatabaseQueryService, scan: Pipe) = {
    val scanCost = medianPerRowCount(runSimulation(graph, scan)).head
    val simulation = runSimulation(graph, expand(scan, RELATIONSHIP)).map(_.subtractTime(scanCost.elapsed))

    simulation
  }
```

That thing first figures out the cost of the scan and then run an expand on top of the expand and subtracts the cost of the scan.

systay: Thanks! Well spotted.

jexp:
Did you also add numeric vs. non-numeric property for the index seek?
B/c the non-numeric one (hopefully) doesn't do an additional property-read?

jexp:
So NodeIndexScan is as fast as NodeByLabelScan?

Interesting that rel-by-id is faster than node-by-id? Shouldn't it be the same or even the other way round?

Nodes are simpler records that rel-records.

PR #7820, opened 2016-08-26
Add log for security-related events

Mats-SX:
Replaces #7817 

changelog: Log all security related events to the new security.log file

PR #7890, opened 2016-09-07
Reload changed auth files when modified on disk

fickludd:
Every 5 seconds the auth and role files are checked for modifications, and reloaded if changed. This is to support clustered native auth, where the files are synchronized using rsync or similar.

changelog: Automatically reload auth files from disk every 5 seconds to enable external changes and externally driven cluster synchronisation

### DIFF
--- a/community/bolt/src/test/java/org/neo4j/bolt/v1/transport/integration/Neo4jWithSocket.java
+++ b/community/bolt/src/test/java/org/neo4j/bolt/v1/transport/integration/Neo4jWithSocket.java
@@ -26,14 +26,15 @@ import org.junit.runners.model.Statement;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 import org.neo4j.bolt.BoltKernelExtension;
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.config.Setting;
+import org.neo4j.io.fs.FileSystemAbstraction;
 import org.neo4j.test.TestGraphDatabaseFactory;
 
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.BoltConnector.EncryptionLevel.OPTIONAL;
@@ -41,23 +42,32 @@ import static org.neo4j.graphdb.factory.GraphDatabaseSettings.boltConnector;
 
 public class Neo4jWithSocket implements TestRule
 {
+    private Supplier<FileSystemAbstraction> fileSystemProvider;
     private final Consumer<Map<Setting<?>,String>> configure;
-    TestGraphDatabaseFactory graphDatabaseFactory;
+    private TestGraphDatabaseFactory graphDatabaseFactory;
     private GraphDatabaseService gdb;
 
     public Neo4jWithSocket()
     {
-        this( new TestGraphDatabaseFactory(), settings -> {} );
+        this( new TestGraphDatabaseFactory(), () -> null, settings -> {} );
     }
 
     public Neo4jWithSocket( Consumer<Map<Setting<?>, String>> configure )
     {
-        this( new TestGraphDatabaseFactory(), configure );
+        this( new TestGraphDatabaseFactory(), () -> null, configure );
     }
 
     public Neo4jWithSocket( TestGraphDatabaseFactory graphDatabaseFactory, Consumer<Map<Setting<?>, String>> configure )
     {
+        this( graphDatabaseFactory, () -> null, configure );
+    }
+
+    public Neo4jWithSocket( TestGraphDatabaseFactory graphDatabaseFactory,
+            Supplier<FileSystemAbstraction> fileSystemProvider,
+            Consumer<Map<Setting<?>, String>> configure )
+    {
         this.graphDatabaseFactory = graphDatabaseFactory;
+        this.fileSystemProvider = fileSystemProvider;
         this.configure = configure;
     }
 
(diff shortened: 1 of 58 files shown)


community/io/src/test/java/org/neo4j/graphdb/mockfs/DelegatingFileSystemAbstraction.java
```diff
@@ -19,6 +19,7 @@
  */
 package org.neo4j.graphdb.mockfs;
 
+import java.awt.image.ConvolveOp;
```
chrisvest: What is this?

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
```diff
@@ -592,6 +613,17 @@ public void truncate( File file, long size ) throws IOException
         data.truncate( size );
     }
 
+    @Override
+    public long lastModifiedTime( File file ) throws IOException
+    {
+        EphemeralFileData data = files.get( canonicalFile( file ) );
+        if ( data == null )
+        {
+            throw new FileNotFoundException( "File " + file + " not found" );
```
chrisvest, line 623: This should be `NoSuchFileException`.

chrisvest:
@fickludd Needs rebase.

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
```diff
@@ -133,8 +135,13 @@ public void deleteRecursively( File directory ) throws IOException
...
     @Override
-    public boolean renameFile( File from, File to ) throws IOException
+    public boolean move( File from, File to, CopyOption... copyOptions ) throws IOException
     {
+        if ( copyOptions.length > 0 )
+        {
+            Files.move( from.toPath(), to.toPath(), copyOptions );
+            return true; // will throw if failure
+        }
         return FileUtils.renameFile( from, to );
```
chrisvest, line 145: This is pretty ugly. The behaviour, potential exceptions and side-effects change depending on whether there are copy options or not, even if they otherwise should make no functional difference.

Have you tried using `Files.move` always, and checked if it works or breaks on all relevant platforms?

fickludd:
@chrisvest on it, but found threading bug as well, so takes a little time to straighten out

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
fickludd: It's ugly indeed, but I selected this implementation after looking at the even uglier windows specific bug-fix in `FileUtils.renameFile`. I think the full move over to `Files.move` should happen when more time is available to handle the potential bugs.

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
fickludd: Could be, but all the other methods in `EphemeralFileSystemAbstraction` return FileNotFoundException when the file data is null. This is only following the pattern.

community/security/src/main/java/org/neo4j/server/security/auth/FileRepositorySerializer.java
```diff
@@ -34,34 +38,59 @@
...
-        }
+        OutputStream o = fs.openAsOutputStream( file, false );
+        o.write( bytes );
+        o.close();
+    }
+
+    public static byte[] readFromFile( FileSystemAbstraction fs, File file ) throws IOException
+    {
+        InputStream s = fs.openAsInputStream( file );
+        byte[] bytes = sun.misc.IOUtils.readFully( s, -1, false );
```
davidegrohmann: Did we test this on the IBM jdk?

community/security/src/main/java/org/neo4j/server/security/auth/FileRepositorySerializer.java
```diff
@@ -34,34 +38,59 @@
...
+    {
+        File tempFile = getTempFile( fileSystem, recordsFile );
 
-        Path tempFile = Files.createTempFile( directory, recordsFile.getFileName().toString() + "-", ".tmp" );
         try
         {
-            Files.write( tempFile, serialize( records ) );
-            Files.move( tempFile, recordsFile, ATOMIC_MOVE, REPLACE_EXISTING );
+            writeToFile( fileSystem, tempFile, serialize( records ) );
+            fileSystem.move( tempFile, recordsFile, ATOMIC_MOVE, REPLACE_EXISTING );
```
davidegrohmann, line 81: What are we gonna do if `ATOMIC_MOVE` is unsupported?

fickludd: We recover from transient errors during move, as tested in the FileUserRepository test. If neo4j is installed on a filesystem were an atomic move within the same directory is not possible then the flat-file based security is not recommended, and LDAP should be used instead.

community/bolt/src/test/java/org/neo4j/bolt/v1/transport/integration/Neo4jWithSocket.java
```diff
@@ -26,38 +26,48 @@
...
 
     public Neo4jWithSocket( Consumer<Map<Setting<?>, String>> configure )
     {
-        this( new TestGraphDatabaseFactory(), configure );
+        this( new TestGraphDatabaseFactory(), () -> null, configure );
     }
 
     public Neo4jWithSocket( TestGraphDatabaseFactory graphDatabaseFactory, Consumer<Map<Setting<?>, String>> configure )
     {
+        this( graphDatabaseFactory, () -> null, configure );
```
davidegrohmann, line 62: can we have a proper default here? What does `null` give you? "Ephemeral" or "real"?

fickludd: Nice point, will fix as part of separate PR https://github.com/neo4j/neo4j/pull/7820, which needs to be rebased on this one.

PR #7896, opened 2016-09-08
Add a file rename method to the PageCache interface

chrisvest:
We need this since only the page cache is aware of any custom IO subsystem configurations we might have, and renaming files is an operation that is used in a fair number of places.

Co-authored with @burqen

### DIFF
--- a/community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
+++ b/community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
@@ -135,14 +135,9 @@ public class DefaultFileSystemAbstraction implements FileSystemAbstraction
     }
 
     @Override
-    public boolean move( File from, File to, CopyOption... copyOptions ) throws IOException
+    public void renameFile( File from, File to, CopyOption... copyOptions ) throws IOException
     {
-        if ( copyOptions.length > 0 )
-        {
-            Files.move( from.toPath(), to.toPath(), copyOptions );
-            return true; // will throw if failure
-        }
-        return FileUtils.renameFile( from, to );
+        Files.move( from.toPath(), to.toPath(), copyOptions );
     }
 
     @Override
--- a/community/io/src/main/java/org/neo4j/io/pagecache/PageCache.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/PageCache.java
@@ -97,12 +97,12 @@ public interface PageCache extends AutoCloseable
     int maxCachedPages();
 
     /**
-     * Rename source file to the given target file.
+     * Rename source file to the given target file, effectively moving the file from source to target.
      *
-     * Both files have to be unmapped when performing the move, otherwise an exception will be thrown.
+     * Both files have to be unmapped when performing the rename, otherwise an exception will be thrown.
      *
-     * @param sourceFile The name of the file to move.
-     * @param targetFile The new name of the file after the move.
+     * @param sourceFile The name of the file to rename.
+     * @param targetFile The new name of the file after the rename.
      * @param copyOptions Options to modify the behaviour of the move in possibly platform specific ways. In particular,
      * {@link java.nio.file.StandardCopyOption#REPLACE_EXISTING} may be used to overwrite any existing file at the
      * target path name, instead of throwing an exception.
--- a/community/io/src/main/java/org/neo4j/io/pagecache/PageCache.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/PageCache.java
@@ -21,6 +21,7 @@ package org.neo4j.io.pagecache;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.CopyOption;
 import java.nio.file.OpenOption;
 import java.nio.file.StandardOpenOption;
 import java.util.Optional;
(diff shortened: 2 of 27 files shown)


community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
```diff
@@ -133,9 +134,9 @@ public void deleteRecursively( File directory ) throws IOException
     }
 
     @Override
-    public boolean renameFile( File from, File to ) throws IOException
+    public void renameFile( File from, File to, CopyOption... copyOptions ) throws IOException
     {
-        return FileUtils.renameFile( from, to );
+        FileUtils.renameFile( from, to, copyOptions );
```
MishaDemianenko: Why we using FileUtils.renameFile here but use Files.move in DelegateFileSystemAbstraction  shouldn't we use the same thing in both?

community/io/src/main/java/org/neo4j/io/pagecache/PageCache.java
```diff
@@ -94,4 +95,17 @@
 
     /** The max number of cached pages. */
     int maxCachedPages();
+
+    /**
+     * Move the file from the given source path, to the given target path.
```
MishaDemianenko: javadoc describe paths here but we operate files

community/io/src/main/java/org/neo4j/io/pagecache/PageCache.java
```diff
@@ -94,4 +95,17 @@
...
+     *
+     * Both files have to be unmapped when performing the move, otherwise an exception will be thrown.
+     *
+     * @param sourceFile The name of the file to move.
+     * @param targetFile The new name of the file after the move.
+     * @param copyOptions Options to modify the behaviour of the move in possibly platform specific ways. In particular,
+     * {@link java.nio.file.StandardCopyOption#REPLACE_EXISTING} may be used to overwrite any existing file at the
+     * target path name, instead of throwing an exception.
+     */
+    void moveFile( File sourceFile, File targetFile, CopyOption... copyOptions ) throws IOException;
```
MishaDemianenko: i find it's really confusing that FS abstraction doing same functionality by method called 'renameFile' and here we define method 'moveFile'. i think those 2 should have same name.

MishaDemianenko:
@chrisvest @burqen in general i do not think that PageCache and PageSwapperFactory should have anything to do with file moves don't they? Shouldn't that be some other abstraction?
It does not look like move of unopened files have anything in common with creating page swappers for example.

MishaDemianenko:
@chrisvest @burqen i finish my first round of review

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
burqen: Because they take different arguments.
`FileUtils.renameFile` take a file and use `Files.move` underneath.
`Files.move` take a string which is necessary for `DelegateFileSystemAbstraction` to be able to delegate to provided `FileSystem`.

MishaDemianenko: this is how it looks now:

```
    @Override
    public boolean renameFile( File from, File to ) throws IOException
    {
        Files.move( path( from ), path( to ) );
        return true;
    }
```

why it's can be:

```
    @Override
    public boolean renameFile( File from, File to ) throws IOException
    {
        return FileUtils.renameFile( from, to );
    }
```

for example?

burqen: Because `path( from )` is what delegates to provided file system in delegate.
I guess we could let `DefaultFileSystemAbstraction` do

```
    @Override
    public void renameFile( File from, File to, CopyOption... copyOptions ) throws IOException
    {
        Files.move( from.toPath(), to.toPath(), copyOptions );
    }
```

burqen:
@MishaDemianenko Rebased and adressed comments.

I agree with your point about PageCache and PageSwapperFactory should not do file operations. We should adress that problem as part of future work.

community/io/src/main/java/org/neo4j/io/pagecache/PageCache.java
```diff
@@ -94,4 +95,17 @@
 
     /** The max number of cached pages. */
     int maxCachedPages();
+
+    /**
+     * Rename source file to the given target file.
+     *
+     * Both files have to be unmapped when performing the move, otherwise an exception will be thrown.
```
MishaDemianenko: Java doc should be fully updated for rename, not partially

community/io/src/main/java/org/neo4j/io/pagecache/PageSwapperFactory.java
```diff
@@ -97,4 +99,21 @@ PageSwapper createPageSwapper(
...
     void syncDevice() throws IOException;
+
+    /**
+     * Rename the given source file given target file name, effectively moving the file from source to target.
+     *
+     * The provided list of {@link CopyOption CopyOptions} can be used to modify and influence platform specific
+     * behaviour. In particular, {@link java.nio.file.StandardCopyOption#REPLACE_EXISTING} may be used to overwrite any
+     * existing file at the target path name, instead of throwing an exception.
+     *
+     * Implementors are free to assume that neither file name will be mapped by the page cache at the time of the move,
```
MishaDemianenko: Still mention move operation

MishaDemianenko (changes requested):
Please create a card about update of PageCache and PageSwapperFactory, also there are 2 small java doc comments. As soon as those items will be addressed - PR will be merged

MishaDemianenko:
ping @burqen

burqen:
@MishaDemianenko 
Java docs comments addressed.
We have a card in technical dept candidates. That's what you meant right?
https://trello.com/c/AXcSEusY/2229-refactor-pagecache-renamefile

MishaDemianenko (approved):

PR #7898, opened 2016-09-08
Fix netty buffer leak

pontusmelke:
Flush can be called from multiple threads at the same time which
leads can lead to multiple buffers being allocated and not all of them
will be released before garbage collected.

This is tricky to test, currently we see this error only when running tests
with `-Dio.netty.leakDetectionLevel=PARANOID`


pontusmelke:
will reopen against 3.0

PR #7900, opened 2016-09-08
List queries post updates and cleanups

MishaDemianenko:


MishaDemianenko:
@systay can you please take a look

PR #7904, opened 2016-09-09
Run IO module tests in parallel

chrisvest:
Inspired by our recent discussions, @benbc

### DIFF
--- a/community/io/pom.xml
+++ b/community/io/pom.xml
@@ -105,6 +105,22 @@ the relevant Commercial Agreement.
   <build>
     <plugins>
       <plugin>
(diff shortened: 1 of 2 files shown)


benbc:
Does it run faster?

chrisvest:
Doesn't look like it makes any useful difference on the build agents. Only helpful for local builds, apparently :(

benbc:
Enough difference that we should consider upgrading the build agents somehow?

chrisvest:
Locally the IO module build time goes from 2m 40s to 1m 30s or so. I don't know if that's enough. It's not very much compared to building the whole project.

PR #7907, opened 2016-09-09
Exclude the store lock file from archives

benbc:
The behaviour of file locks on Windows is a constant source of
amazement (or bugs; and sometimes both). This code throws an exception
when reading:

```
File file = File.createTempFile("tmp", null);
file.deleteOnExit();
try (FileChannel channel = new RandomAccessFile(file, "rw").getChannel();
     FileLock lock = channel.tryLock()) {
    assert lock != null;
    new FileInputStream(file).read();
}
```

This is surprising to the casual reader of the Java API docs (to whit:
https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileLock.html);
but it is consistent with the locking semantics of the Win32 APIs
which specify, to anyone who cares to plumb their perlucid depths (so
far as this:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365202(v=vs.85).aspx),
that "[i]f the locking process opens the file a second time, it cannot
access the [locked] region through this second handle until it unlocks
the region". Caveat lockor.

Having locked the store lock for the duration of the archive process,
we must therefore avoid reading the contents of that file. This is no
great loss since the file is empty and there is nothing to be gained
(nor, it should be said, lost) from including it in the archive.

### DIFF
--- a/community/common/src/main/java/org/neo4j/function/ThrowingAction.java
+++ b/community/common/src/main/java/org/neo4j/function/ThrowingAction.java
@@ -21,13 +21,22 @@ package org.neo4j.function;
 
 /**
  * An action that takes no parameters and returns no values, but may have a side-effect and may throw an exception.
+ *
  * @param <E> The type of exception this action may throw.
  */
 public interface ThrowingAction<E extends Exception>
 {
     /**
      * Apply the action for some or all of its side-effects to take place, possibly throwing an exception.
+     *
      * @throws E the exception that performing this action may throw.
      */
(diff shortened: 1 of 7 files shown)


praveenag:
LGTM, minor feedback the `FileVisitors` doesn't seem to have tests. It is especially an interesting class because of methods such as `onlyMatching`.

Also, me when reading your comments :(

Dictionary cat

benbc:
Agreed with @praveenag that I'll add the tests in a separate commit.

PR #7909, opened 2016-09-09
Single active update puller

MishaDemianenko:
After introduction of detached state components lifecycle was change in a way that now it's possible to start new set of components while you still have old one active.
For case of update puller it mean that for some time we will have 2 update pullers active and they will try to apply transactions concurrently.
That is incorrect and causing errors during transactions apply phase.
This commit introduce new phase in component switch lifecycle, just after old component shutdown. And we will use this phase to start new update puller

@pair-with @lutovich

changelog: Update ha component lifecycle to make sure we have only one active update puller on slaves


MishaDemianenko:
@digitalstain can you please take a look, thx

PR #7910, opened 2016-09-09
Enable and update execution guard tests for driver.

MishaDemianenko:


MishaDemianenko:
@zhenlineo here we go, message router updated and everything looks fine; can you please check driver side of it? ta!

zhenlineo:
@MishaDemianenko You are free to merge. The driver state machine needs more work but as your change will not bring any observable change to the bolt state machine. So just go ahead. We will come up with better approach to address streaming result for discard_all latter.

MishaDemianenko:
@zhenlineo ta!

PR #7912, opened 2016-09-12
Warn that neo4j-wrapper.conf is deprecated

benbc:


spacecowboy:
Looks good.

Upon 3.0 upgrades, do we actually move the settings into the new file? Or do we leave a left-over file?

benbc:
@spacecowboy You mean the Debian package specifically? We don't do anything special.

spacecowboy:
@benbc ok

PR #7914, opened 2016-09-12
Introduces limit to the size of netty direct buffers

digitalstain:
Unless a limit is set on the size of Netty buffers used for serializing
 and sending messages, they can grow to occupy all of the direct memory
 available, leading to unrecoverable errors. This change allows for
 setting a ceiling to how big a buffer can grow.
This currently affects transaction state serialization, effectively
 limiting how big a transaction can get and still be eligible for
 replication in a CE cluster.

### DIFF
--- a/enterprise/core-edge/src/main/java/org/neo4j/coreedge/core/state/machines/tx/ReplicatedTransactionFactory.java
+++ b/enterprise/core-edge/src/main/java/org/neo4j/coreedge/core/state/machines/tx/ReplicatedTransactionFactory.java
@@ -26,7 +26,7 @@ import java.util.List;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
-
+import org.neo4j.coreedge.messaging.MessageTooBigException;
 import org.neo4j.coreedge.messaging.NetworkFlushableChannelNetty4;
 import org.neo4j.coreedge.messaging.NetworkReadableClosableChannelNetty4;
 import org.neo4j.kernel.impl.storageengine.impl.recordstorage.RecordStorageCommandReaderFactory;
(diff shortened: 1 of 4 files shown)


enterprise/core-edge/src/main/java/org/neo4j/coreedge/core/state/machines/tx/ReplicatedTransactionFactory.java
```diff
@@ -41,18 +41,25 @@
 
 public class ReplicatedTransactionFactory
 {
+    private static final long MAX_SERIALIZED_TX_SIZE = 1024 * 1024 * 1024L; // 1 GB
```
lutovich: `ByteUnit` would make this line even prettier

enterprise/core-edge/src/main/java/org/neo4j/coreedge/messaging/NetworkFlushableChannelNetty4.java
```diff
@@ -20,78 +20,114 @@
...
+import static java.lang.String.format;
+
 public class NetworkFlushableChannelNetty4 implements FlushableChannel
 {
+    /**
+     * This implementation puts an upper limit to the size of the state serialized in the buffer. The default
+     * value for that should be sufficient for all replicated state except for transactions, the size of which
+     * is unbounded.
+     */
+    private static final long DEFAULT_SIZE_LIMIT = 2 * 1024 * 1024; // 2 MB
```
lutovich: maybe use `ByteUnit` here as well

maxsumrall (approved):
Changed to use ByteUnit in some places. LGTM.

PR #7915, opened 2016-09-12
Using label index for schema index creation

MishaDemianenko:
Switch to label scan store to determine list of node ids for specified label for index population.
In cases when label is not used to mark significant amount of nodes this gives noticeable performance boost.
For cases when label scan store index does not exist yet - full node scan is used.
Feature is experimental and turn off by default.
To turn it on please see feature flag 'use.label.index' in DynamicIndexStoreView.
To be able to adjust when label either node store should be used please see feature toggle 'all.nodes.visit.percentage.threshold' in DynamicIndexStoreView.

changelog: Add possibility to use label store as a source of affected nodes for schema index creation.


MishaDemianenko:
@tinwelint i have update proposal to include updated _idea ready_ code version that should give mostly clean picture of how things working/should work. PR still contains some updates and TODOs that here only temporary/for ongoing testing, but i think if you interesting you can already take a look.

burqen:
@MishaDemianenko If this change speeds up index creation a lot I think it is suitable for changelog.

MishaDemianenko:
@burqen for now idea is that this feature will be disabled by default in old versions

burqen:
@MishaDemianenko Guarded by some experimental config option? If we want people to try it out (which is useful for feedback) we should not hide it, instead promote it as an experimental feature to try at own risk.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/LabelScanViewNodeStoreScan.java
```diff
@@ -0,0 +1,211 @@
...
+            labels = update.getLabelsBefore();
+            break;
+        default:
+            throw new UnsupportedOperationException( "unsupported update mode" );
+        }
+        int property = update.getPropertyKeyId();
+        long nodeId = update.getNodeId();
+        for ( long label : labels )
+        {
+            PrimitiveLongObjectMap<PrimitiveLongSet> labelsNodes = propertyLabelNodes.get( property );
```
tinwelint: This looks like a pretty heavy datastructure, `map<long,set<long>>` could we instead only make due with a `set<long>` of changed nodes and go and read the node labels from the store on completion? That's what e.g. recovery does for reindexing purposes. It would require way less memory.

tinwelint: Another thing which may be good would be to, instead of queueing up _all_ updates during the scan, to look at number of pending updates and if they are rising then the label scan could do this incrementally: i.e. at some point during the scan stop, apply pending updates and create a new reader, resetting the update queue.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/populator/DirectNonUniqueIndexSampler.java
```diff
@@ -0,0 +1,99 @@
...
+        catch ( IOException e )
+        {
+            throw new RuntimeException( e );
+        }
+    }
+
+    @Override
+    public IndexSample result( int numDocs )
+    {
+        throw new UnsupportedOperationException( "Not implemented." );
```
tinwelint, line 97: And there's no point implementing it? Perhaps we could throw something else then... or at least a different message

enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/DynamicIndexStoreIT.java
```diff
@@ -0,0 +1,167 @@
...
+import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.graphdb.factory.GraphDatabaseFactory;
+import org.neo4j.helpers.progress.ProgressMonitorFactory;
+import org.neo4j.kernel.configuration.Config;
+import org.neo4j.logging.FormattedLogProvider;
+import org.neo4j.test.SuppressOutput;
+import org.neo4j.test.TargetDirectory;
+
+public class DynamicIndexStoreIT
```
tinwelint: Do this test need to be in enterprise-kernel?

tinwelint: also this test takes 40s on my laptop run, probably much more in CI. We should be wary of our test times. Try to test the gist of what you're testing without exaggerating

tinwelint (commented):
Generally looks good.

I don't understand the index sampling changes or problem quite, but I'm OK with that as long as we have tests there.

As a side note one might argue that a more generic approach to deciding whether or not to use label scan store to scan nodes with certain label might in the future be moved to be a statistical thing so that labels which nodes above a certain percentage of all nodes have wouldn't even be indexed in the label scan store and decision to use or not use label scan store could be transparent on a lower level.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/LabelScanViewNodeStoreScan.java
MishaDemianenko: second one sounds like a nice suggestion will try that...

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/populator/DirectNonUniqueIndexSampler.java
MishaDemianenko: What other exception you can think of?

enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/DynamicIndexStoreIT.java
MishaDemianenko: that should be somewhere where we actually use lucene to populate indexes, that' why enterprise kernel was chosen. Update test to pass much faster!

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/populator/DirectNonUniqueIndexSampler.java
tinwelint: Dunno, at least a better message. The current message looks like a TODO

enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/DynamicIndexStoreIT.java
tinwelint: Cool, for that we usually put those in community/neo4j

MishaDemianenko:
@tinwelint should be ready for, all comments should be addressed.

MishaDemianenko:
@burqen change log and description updated

burqen:
@MishaDemianenko Nice! How do I activate feature?

MishaDemianenko:
@burqen description mentions that right?

burqen:
@MishaDemianenko Sure, but I wont see that in the changelog. Although, there will be a link to the PR in the changelog, so maybe that is good enough. Let's agree it is.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/BatchingMultipleIndexPopulator.java
```diff
@@ -131,14 +129,11 @@ protected IndexPopulation createPopulation( IndexPopulator populator,
...
-        if ( queue.size() >= QUEUE_THRESHOLD )
-        {
-            log.debug( "Populating from queue." + EOL + this );
-            flushAll();
-            awaitCompletion();
-            super.populateFromQueue( currentlyIndexedNodeId );
-            log.debug( "Drained queue and all batched updates." + EOL + this );
-        }
+        log.debug( "Populating from queue." + EOL + this );
+        flushAll();
```
tinwelint, line 133: Doing this flushing every time there's an update will produce more and smaller tasks to the executor, won't it? Could it be less efficient to do this? Have you measured?

tinwelint:
The simplification where the label scan reader is marked as outdated looks really good to me, there's no refresh involved in closing/re-opening the reader, right?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/BatchingMultipleIndexPopulator.java
MishaDemianenko: @tinwelint no no, now flushing condition moved up to MultipleIndexPopulator and we will call `populateFromQueue` now only when `isQueueThresholdReached` will allow us to do that; so in terms of behaviour it should be exactly the same

MishaDemianenko:
No, new refreshes where not introduced, we should see updates since they will be applied during commit process and in the end of that index should be refreshed.

tinwelint:
Neo4j is now relying on results to be sorted from any `LabelScanStore` (`LabelScanReader` that is). I'd like to see that reflected in its javadocs and compatibility tests if such exist

MishaDemianenko:
@tinwelint i have added `retrieveNodeIdsInAscendingOrder` that check that we get nodes from label scan store always in ascending order. Or do you have some other compatibility tests in mind?

tinwelint:
@MishaDemianenko hmm actually there doesn't seem to be such a test suite, then it's fine. Creating such could be done as separate piece of work. OK I think we're quite done here actually!

MishaDemianenko:
LGTM!

PR #7916, opened 2016-09-12
Bumps store version of HighLimit format

tinwelint:
there was a change in 3.0.5 which made a backwards compatible change to
HighLimit format. The store version wasn't bumped since there was no need
for actual migration. Not bumping store version leads to problems if
user, after running on 3.0.5 would start db on older 3.0.x patch.
The db would start normally since the store version would be correct,
but the data in the records would look completely different.

This commit fixes this problem and adds a temporary, 3.0-specific
special case as to not do actual migration when bumping store version.
3.1 has better mechanics so that this can be done w/o special-casing as well.

There will still be a problem with 3.0.5 in that going back from that version
to a lower 3.0.x patch release if using high limit record format will
fail on reading those records, or succeed but reading completely invalid
data in those records.

changelog: (Enterprise-only) Bumps store version for change in high-limit format. This requires setting `dbms.allow_format_migration=true` config for stores using that format, although no actual store upgrade will be performed since it's a backwards compatible change. The store will be marked as being of the newer version so that older version can't start on an upgraded store.


burqen:
@tinwelint or @MishaDemianenko Can you please write a changelog message for this PR about the necessity of using the migration flag when opening an old store on 3.0.6 because of the store version bump.

PR #7918, opened 2016-09-12
Fixed a bug in the Debian packages which caused an "upgrading from 2.x" warning to be printed even when upgrading from 3.x

srbaker:


spacecowboy:
lgtm

PR #7920, opened 2016-09-13
Write to STDERR for Powershell management scripts

glennsarti:
Previously the Powershell management scripts did not process the `help` command
correctly.  Also it would put the error message onto STDERR as is expected. This
commit changes the command behaviour to output the helptext and an exit code of
zero for the `help` command.  This commit also writes the error message for an
unknown command to STDERR instead of the default STDOUT.  This commit also
changes the exitcode for an unknown command from 255 to 1 to be more inline with
linux based management scripts.


glennsarti:
I didn't see any value in adding unit tests for the new Write-StdErr function.

benbc:
Merging despite a failing build (flaky test). Checked that the important stuff actually ran.

PR #7895, opened 2016-09-08
Fixes an issue where log rotation would cut transaction in half

tinwelint:
The rotation done in PhysicalLogFile only flushed the channel,
i.e. bytes already written to it. It didn't tell the writer to
empty its buffer into the channel first. This could sometimes
result in rotation happening in the middle of a transaction.
All the bytes would be there, just split across files and be
more difficult to read later on if not using proper log version
bridging and would eventually cause problems when first part of
split transaction would be pruned.

This change will have transactions be fully written in the log
they started being written to before rotation happens, without
introducing any performance penalty of appending, forcing
or rotation.

Included in this PR is also a fix for having recovery use LogicalTransactionStore
so that it's able to read split transactions, this also simplifies and reduces
the recovery code since lots of special code for reading log files is
now deleted in favour of existing ways of doing that.

changelog: Fixes [#7204](https://github.com/neo4j/neo4j/issues/7204) an issue where log rotation sometimes would cut a transaction in half making it problematic for recovery to read such transactions. Transactions are now never split into multiple log files. Any db experiencing this issue can now also start that db with this fix included due to improved log reading abilities.

PR #7924, opened 2016-09-13
Fixes an issue where log rotation would cut transaction in half

tinwelint:
The rotation done in PhysicalLogFile only flushed the channel,
i.e. bytes already written to it. It didn't tell the writer to
empty its buffer into the channel first. This could sometimes
result in rotation happening in the middle of a transaction.
All the bytes would be there, just split across files and be
more difficult to read later on if not using proper log version
bridging and would eventually cause problems when first part of
split transaction would be pruned.

This change will have transactions be fully written in the log
they started being written to before rotation happens, without
introducing any performance penalty of appending, forcing
or rotation.

Included in this PR is also a fix for having recovery use LogicalTransactionStore
so that it's able to read split transactions, this also simplifies and reduces
the recovery code since lots of special code for reading log files is
now deleted in favour of existing ways of doing that.

NOTE: this is a port of https://github.com/neo4j/neo4j/pull/7895 (which was null-merged). Changelog entry will be pulled from that PR as well, so not marking this with that label.


tinwelint:
Yet again, tools compilation error is expected

PR #7928, opened 2016-09-13
Add connection details to listQueries

systay:
This adds connection details to the output of the procedure `listQueries`.

Depending on how the user is connecting to Neo4j, different outputs will be seen in the result table. Bolt and REST report connection address, whereas an embedded session will just be listed as `embedded`.


community/server/src/main/java/org/neo4j/server/rest/web/InternalJettyServletRequest.java
```diff
@@ -361,4 +358,50 @@ public HttpChannelState getHttpChannelState()
...
+        {
+            return new RequestData(
+                    req.getRemoteAddr(),
+                    req.getRemoteHost(),
+                    req.isSecure(),
+                    req.getRemotePort(),
+                    req.getLocalName(),
+                    req.getLocalAddr(),
+                    req.getLocalPort(),
+                    req.getAuthType()
```
MishaDemianenko: please note that AuthType can be null

community/server/src/main/java/org/neo4j/server/rest/web/InternalJettyServletRequest.java
```diff
@@ -361,4 +358,50 @@ public HttpChannelState getHttpChannelState()
...
+            this.localAddress = localAddress;
+            this.localPort = localPort;
+            this.authType = authType;
+        }
+
+        public static RequestData from( HttpServletRequest req )
+        {
+            return new RequestData(
+                    req.getRemoteAddr(),
+                    req.getRemoteHost(),
```
MishaDemianenko: please note that in case if implementation of servlet api will try to resolve fully qualified name 
that can take significant time

community/server/src/test/java/org/neo4j/server/rest/batch/BatchOperationsTest.java
```diff
@@ -69,36 +70,35 @@ public void testSchemeInInternalJettyServletRequestForHttp() throws UnsupportedE
...
 
         InternalJettyServletRequest req = new InternalJettyServletRequest( "POST",
                 "https://localhost:7473/db/data/node", "", new InternalJettyServletResponse(),
                 mock );
 
         // When & then
-        assertEquals( "authorization/auth", req.getAuthType());
         assertEquals( "127.0.0.1", req.getRemoteAddr());
         assertEquals( "localhost", req.getRemoteHost());
+        assertEquals( true, req.isSecure() );
```
MishaDemianenko: assertTrue

community/server/src/main/java/org/neo4j/server/rest/web/ServerQuerySession.java
```diff
@@ -21,28 +21,14 @@
...
-
-    @Override
-    public String toString()
+    public static QuerySource describe( HttpServletRequest request )
     {
         return request == null ?
-               format("server-session\t%s", username) :
-               format("server-session\thttp\t%s\t%s\t%s", request.getRemoteAddr(), request.getRequestURI(), username );
+            new QuerySource( "server-session" ) :
+            new QuerySource( "server-session", "http", request.getRemoteAddr(), request.getRequestURI() );
```
MishaDemianenko: it looks like it should be `request.getScheme()` instead of http always

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -216,6 +219,7 @@ public void init( SPI spi, Config config )
                         .getOrCreateRelationshipIndex( InternalAutoIndexing.RELATIONSHIP_AUTO_INDEX, null ) ),
                 spi.autoIndexing().relationships() );
         this.indexManager = new IndexManagerImpl( spi::currentStatement, idxProvider, nodeAutoIndexer, relAutoIndexer );
+        this.contextFactorySupplier = lazySingleton( () -> new Neo4jTransactionalContextFactory( spi, locker ) );
```
MishaDemianenko: why do we need contextFactory to be supplier at all? why not just create instance of `Neo4jTransactionalContextFactory` like:

```
this.contextFactory = new Neo4jTransactionalContextFactory( spi, locker );
```

and use it?

community/kernel/src/main/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContext.java
```diff
@@ -270,4 +267,9 @@ public AccessMode accessMode()
     {
         return mode;
     }
+
+    public static Function<TransactionalContext, String> describe()
```
MishaDemianenko: i do not think this belongs to this class

community/kernel/src/main/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContext.java
```diff
@@ -212,13 +204,18 @@ public TransactionalContext provideContext()
...
             Statement statement = txBridge.get();
+            statement.queryRegistration().registerExecutingQuery( executingQuery );
             return new Neo4jTransactionalContext(
-                graph,
-                transaction,
-                statement,
-                executingQuery.queryText(),
-                executingQuery.queryParameters(),
-                locker
+                    graph,
```
MishaDemianenko: shouldn't we use newly introduced context factory for this sort of things?

community/kernel/src/test/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContextTest.java
```diff
@@ -66,9 +66,21 @@ public void setUp()
...
+        KernelTransaction.Type transactionType = null;
+        AccessMode transactionMode = null;
+        ExecutingQuery executingQuery = null;
+        PropertyContainerLocker locker = null;
+        DbmsOperations.Factory dbmsOperationsFactory = null;
+        ThreadToStatementContextBridge txBridge = null;
         Neo4jTransactionalContext transactionalContext =
-            new Neo4jTransactionalContext(
-                databaseQueryService, transaction, statement, "", Collections.emptyMap(), propertyContainerLocker );
+//<<<<<<< HEAD
```
MishaDemianenko: i do not think we should ever resolve conflicts in this way

enterprise/query-logging/src/main/java/org/neo4j/kernel/impl/query/QueryLoggerKernelExtension.java
```diff
@@ -131,10 +131,6 @@ private Lifecycle createEmptyAdapter()
 
     static class QueryLogger implements QueryExecutionMonitor
     {
-        private static final MetadataKey<Long> START_TIME = new MetadataKey<>( Long.class, "start time" );
-        private static final MetadataKey<String> QUERY_STRING = new MetadataKey<>( String.class, "query string" );
-        private static final MetadataKey<Map<String,Object>> PARAMS = new MetadataKey<>( paramsClass(), "parameters" );
```
MishaDemianenko, line 136: static method paramsClass is not used anymore and should be deleted

community/kernel/src/main/java/org/neo4j/kernel/impl/query/QuerySource.java
```diff
@@ -0,0 +1,67 @@
...
+        String[] newParts = new String[parts.length + 1];
+        System.arraycopy( parts, 0, newParts, 0, parts.length );
+        newParts[parts.length] = newPart;
+        return new QuerySource( newParts );
+    }
+
+    @Override
+    public String toString()
+    {
+        return toString( '\t' );
```
MishaDemianenko: please use `return String.join( "\t", parts );` instead

MishaDemianenko (changes requested):
@systay @boggle first round of review completed, couple of small comments

enterprise/query-logging/src/main/java/org/neo4j/kernel/impl/query/QueryLoggerKernelExtension.java
fickludd: Done.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
fickludd: This lazy instantiation is needed as the spi is not fully configured during the `GraphDatabaseFacade.init`. Only creating the factory at first use solves this problem with little fuss.

community/kernel/src/main/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContext.java
fickludd: We are using several private field which are more easily accessed here. Renamed the method instead to improved clarity.

community/server/src/main/java/org/neo4j/server/rest/web/InternalJettyServletRequest.java
fickludd: Yes, but this seems not to be even used. Possibly there's a need for some large clean-ups here, but I don't think we should perform that as a part of this PR. Essentially any null-Strings gotten from the Request will simply be passed on transparently to InternalJettyServletRequest users.

fickludd:
@MishaDemianenko Thanks for catching some sloppy parts of this code. Addressed most and commented on others.

community/server/src/main/java/org/neo4j/server/rest/web/InternalJettyServletRequest.java
MishaDemianenko: representing it as an optional will help a lot for example, printing AuthType: default is much better then print AuthType: null

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -197,7 +195,7 @@ public GraphDatabaseFacade()
     /**
      * Create a new Core API facade, backed by the given SPI.
      */
-    public void init( SPI spi, Config config )
+    public void initSPI( SPI spi, Config config )
```
MishaDemianenko: initSPI is a bit confusing imho, since it looks like we initialise SPI here, but actually we just provide facade with dependencies, so i would keep old name instead

MishaDemianenko: Will it work if in init method we just provide dependencies - store them as fields and then do al initialisation of fields all together in next method together with context factory?

fickludd: I tried this, but it failed a lot of tests that do not need the TransactionalContextFactory. It seems like here is work to do, but I would rather not spend more time on this is the scope of this PR.

PR #7930, opened 2016-09-13
Completes a previous change on InputEntityDecorators

tinwelint:
where everything should be Decorator, not Function there

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityDecorators.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityDecorators.java
@@ -19,7 +19,7 @@
  */
 package org.neo4j.unsafe.impl.batchimport.input;
 
-import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityDecorators.java
```diff
@@ -75,15 +75,26 @@
...
+                    from = decorator.apply( from );
+                }
+                return from;
+            }
+
+            @Override
+            public boolean isMutable()
             {
-                from = decorator.apply( from );
+                return Stream.of( decorators ).anyMatch( decorator -> decorator.isMutable() );
```
alexaverbuch: can be replaced with method reference
`return Stream.of( decorators ).anyMatch( Decorator::isMutable );`

tinwelint: Oh OK, even prettier :)

PR #7931, opened 2016-09-13
Expose ProcedureGuard to check tx termination and guards from within …

boggle:
…long running procedures

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/guard/TimeoutGuard.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/guard/TimeoutGuard.java
@@ -48,7 +48,7 @@ public class TimeoutGuard implements Guard
         check( statement.getTransaction() );
     }
 
-    private void check( KernelTransactionImplementation transaction )
+    public void check( KernelTransactionImplementation transaction )
     {
         if ( transaction.timeout() > UNSPECIFIED_TIMEOUT )
         {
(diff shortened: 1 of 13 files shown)


community/kernel/src/main/java/org/neo4j/kernel/guard/TimeoutGuard.java
```diff
@@ -48,7 +48,7 @@ public void check( KernelStatement statement )
         check( statement.getTransaction() );
     }
 
-    private void check( KernelTransactionImplementation transaction )
+    public void check( KernelTransactionImplementation transaction )
```
MishaDemianenko: Please add `@Override` here

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/DataSourceModule.java
```diff
@@ -421,6 +428,37 @@ private Procedures setupProcedures( PlatformModule platform, EditionModule editi
         return procedures;
     }
 
+    private ComponentRegistry.Provider<TerminationGuard> procedureGuard( final Dependencies dependencies )
+    {
+        Guard guard = dependencies.resolveDependency( Guard.class );
+        return (ctx) ->
+        {
+            KernelTransaction ktx = ctx.get( KERNEL_TRANSACTION );
+            return (TerminationGuard) () ->
```
MishaDemianenko: please extract this class into separate class and put it into some procedure related package, most probably where `TerminationGuard` live

MishaDemianenko (changes requested):
@boggle  Two small comments and also please note there is test failure atm. Review round complete

boggle:
This was green; just rebased but since it's weekend waiting for another build.

PR #7935, opened 2016-09-14
Add maven option to skip HA: -DskipHA

maxsumrall:


enterprise/pom.xml
```diff
@@ -67,6 +66,17 @@ terms of the relevant Commercial Agreement.
 
   <profiles>
     <profile>
+      <id>include-ha</id>
+      <activation>
+        <property>
+          <name>!skipHA</name>
```
maxsumrall, line 72: Great idea!

enterprise/pom.xml
```diff
@@ -34,7 +34,6 @@
     <module>cluster</module>
     <module>backup</module>
     <module>kernel</module>
-    <module>ha</module>
```
maxsumrall, line 37: Nice catch!

maxsumrall (commented):
Overall looks great. I suggest you merge this.

enterprise/pom.xml
maxsumrall: Thanks!

PR #7937, opened 2016-09-14
Use page cache for all operation on record store file during store copy

burqen:
Builds on #7896, new code start at `e297132` (  StoreCopyServer now read from paged file is mapping exists). Will be rebased after #7896 is merged.

When storing the record storage file on some storage external to the underlying file system, only the page cache know how to write to and move those files.

Summary of changes:
- StoreCopyServer read file through PageCache if mapping exists
- ToFileStoreWriter will write through PageCache if file is considered to be a record store file
- Add record size to store copy protocol to allow new mapping of record store files
- Use page cache to move record store files from temp directory to store directory after store copy


chrisvest:
Rebased. Marking it as ready for review & merge. /cc @burqen @tinwelint

community/io/src/main/java/org/neo4j/io/pagecache/PagedFile.java
```diff
@@ -131,6 +131,11 @@
     int pageSize();
 
     /**
+     * Size of file, in bytes.
+     */
+    long fileSize() throws IOException;
```
chrisvest: There should be tests for this in `PageCacheTest`

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StoreType.java
```diff
@@ -227,15 +228,28 @@ void close( NeoStores me, Object object )
...
+                () -> new IllegalArgumentException( "No enum constant for " + storeFileName + " file." ) );
+    }
+
+    /**
+     * Determine if a file name corresponds to any store type.
+     *
+     * @param fileName - name of the file to check for corresponding store type
+     * @return {@link Optional} wrapping store type, or empty optional if no store type corresponds to file name
+     */
+    public static Optional<StoreType> isStoreType( String fileName )
```
chrisvest: Maybe this should be called `findStoreType` or `determineStoreType`, or `guessStoreType`.

enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyServer.java
```diff
@@ -154,13 +159,32 @@ public RequestContext flushStoresAndStreamStoreFiles( String triggerName, StoreW
...
+                    }
+                    else
                     {
-                        monitor.startStreamingStoreFile( file );
-                        writer.write( relativePath( storeDirectory, file ), fileChannel,
-                                temporaryBuffer, file.length() > 0 );
-                        monitor.finishStreamingStoreFile( file );
+                        try ( ReadableByteChannel fileChannel = fileSystem.open( file, "r" ) )
+                        {
+                            long fileSize = file.length();
```
chrisvest: `fileSystem.getFileSize( file )` ?

enterprise/com/src/main/java/org/neo4j/com/storecopy/ToFileStoreWriter.java
```diff
@@ -23,46 +23,77 @@
...
+import static java.nio.file.StandardOpenOption.WRITE;
+import static org.neo4j.kernel.impl.store.StoreType.isStoreType;
+import static org.neo4j.kernel.impl.store.format.RecordFormat.NO_RECORD_SIZE;
 
 public class ToFileStoreWriter implements StoreWriter
 {
     private final File basePath;
     private final StoreCopyClient.Monitor monitor;
+    private final PageCache pageCache;
+    private final List<FileMoveAction> fileMoveActions; // Todo need thread safe?
```
chrisvest: No, I don't think this needs to be thread safe. The files are streaming over one at a time, controlled on our side by a single thread.

enterprise/com/src/main/java/org/neo4j/com/storecopy/ToFileStoreWriter.java
```diff
@@ -23,46 +23,77 @@
...
+                    {
+                        final long written = writeDataThroughPageCache( pagedFile, data, temporaryBuffer, hasData );
+                        addPageCacheMoveAction( file );
+                        return written;
+                    }
+                }
+                else
+                {
+
+                    return writeDataThroughFileSystem( file, data, temporaryBuffer, hasData );
```
chrisvest: Why don't this add a move action? Then we'd only need one loop in `StoreCopyClient`.

chrisvest (commented):
I had a few comments on a look through.

chrisvest (commented):
Had a few comments

enterprise/com/src/main/java/org/neo4j/com/storecopy/ToFileStoreWriter.java
```diff
@@ -75,6 +106,64 @@ public long write( String path, ReadableByteChannel data, ByteBuffer temporaryBu
...
+    {
+        fileMoveActions.add( ( ( toDir, copyOptions ) ->
+                pageCache.renameFile( file, new File( toDir, file.getName() ), copyOptions ) ) );
+    }
+
+    private int filePageSize( int recordSize )
+    {
+        final int pageCacheSize = pageCache.pageSize();
+        return recordSize == NO_RECORD_SIZE ? pageCacheSize
+                                            : pageCacheSize - pageCacheSize % recordSize;
```
MishaDemianenko: i really miss clarifying parentheses here

enterprise/com/src/main/java/org/neo4j/com/storecopy/ToNetworkStoreWriter.java
```diff
@@ -56,9 +56,10 @@ public long write( String path, ReadableByteChannel data, ByteBuffer temporaryBu
         long totalWritten = 2 + chars.length*2 + 1;
         if ( hasData )
         {
+            targetBuffer.writeInt( recordSize );
+            totalWritten += 4;
```
MishaDemianenko: Integer.BYTES please

enterprise/com/src/main/java/org/neo4j/com/storecopy/ToNetworkStoreWriter.java
```diff
@@ -56,9 +56,10 @@ public long write( String path, ReadableByteChannel data, ByteBuffer temporaryBu
         long totalWritten = 2 + chars.length*2 + 1;
         if ( hasData )
         {
+            targetBuffer.writeInt( recordSize );
```
MishaDemianenko: Shouldn't we update protocol version as result of doing this?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
```diff
@@ -248,13 +252,9 @@ private boolean transfer( EntryVisitor<WritableBuffer> producer, EntryVisitor<Re
...
     {
         List<File> existingFiles = new ArrayList<>();
-        for ( File candidateFile : rotationStrategy.candidateFiles() )
-        {
-            if ( candidateFile.exists() )
-            {
-                existingFiles.add( candidateFile );
-            }
-        }
+        StreamSupport.stream( rotationStrategy.candidateFiles().spliterator(), false )
```
MishaDemianenko: i think all files can be just this:

```
        return StreamSupport.stream( rotationStrategy.candidateFiles().spliterator(), false )
                .filter( File::exists )
```

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StoreType.java
```diff
@@ -227,15 +228,28 @@ void close( NeoStores me, Object object )
      */
     public static StoreType typeOf( String storeFileName )
     {
+        final Optional<StoreType> optional = findStoreType( storeFileName );
```
MishaDemianenko: what if we will just update typeOf(filename) to return Optional instead?

enterprise/com/src/main/java/org/neo4j/com/storecopy/ToFileStoreWriter.java
```diff
@@ -23,46 +23,77 @@
...
         try
         {
             temporaryBuffer.clear();
             File file = new File( basePath, path );
-
             file.getParentFile().mkdirs();
+
+            String filename = file.getName();
+            final Optional<StoreType> storeType = findStoreType( filename );
+            final Optional<PagedFile> existingMapping = pageCache.getExistingMapping( file );
```
MishaDemianenko: I think it would be useful to have a comment explaining here what files we expect to be mapped, not mapped, not a stores to simplify future investigations

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StoreType.java
chrisvest: Of course.

enterprise/com/src/main/java/org/neo4j/com/storecopy/ToFileStoreWriter.java
chrisvest: That's a really good question. In `ToFileStoreWriter` we don't actually need it because we always write to a temporary location. The call in `StoreCopyServer` is seeing more use, so I'm adding a comment there.

chrisvest:
@MishaDemianenko I have addressed your comments.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StoreType.java
```diff
@@ -221,21 +222,21 @@ void close( NeoStores me, Object object )
...
      */
-    public static StoreType typeOf( String storeFileName )
+    public static Optional<StoreType> typeOf( String fileName )
     {
         StoreType[] values = StoreType.values();
         for ( StoreType value : values )
         {
-            if ( value.getStoreName().equals( storeFileName ) ||
-                    storeFileName.equals( MetaDataStore.DEFAULT_NAME + value.getStoreName() ) )
+            if ( value.getStoreName().equals( fileName ) ||
```
tinwelint: To match both (w/ and w/o `neostore.` in front) looks like a source of confusion and a potential edge-case problems finding unrelated files. It'd be better to decide exactly what it looks for, document it and have all callers adhere to it.

MishaDemianenko:
@chrisvest thx, all fixes looks nice so far!

enterprise/com/src/main/java/org/neo4j/com/storecopy/ToNetworkStoreWriter.java
```diff
@@ -45,20 +45,21 @@ public ToNetworkStoreWriter( ChannelBuffer targetBuffer, Monitors monitors )
...
         targetBuffer.writeShort( chars.length );
         Protocol.writeChars( targetBuffer, chars );
         targetBuffer.writeByte( hasData ? 1 : 0 );
         // TODO Make use of temporaryBuffer?
         BlockLogBuffer buffer = new BlockLogBuffer( targetBuffer, bufferMonitor );
-        long totalWritten = 2 + chars.length*2 + 1;
+        long totalWritten = Short.BYTES + chars.length* Character.BYTES + Byte.BYTES;
         if ( hasData )
         {
+            targetBuffer.writeInt( recordSize );
```
tinwelint: It looks like you're changing the network protocol. For compatibility, and detection of incompatibility you have to increment protocol version. I think this is a case where the internal protocol version should be incremented, see https://github.com/neo4j/neo4j/blob/3.1/enterprise/com/src/main/java/org/neo4j/com/ProtocolVersion.java#L24

tinwelint (changes requested):

enterprise/com/src/main/java/org/neo4j/com/storecopy/ToNetworkStoreWriter.java
chrisvest: Yes, I had @digitalstain explain to me in vague terms how this is done.

chrisvest:
@tinwelint I think I've addressed all of your comments.

tinwelint (approved):

enterprise/ha/src/main/java/org/neo4j/kernel/ha/MasterClient214.java
```diff
@@ -19,32 +19,78 @@
...
+public class MasterClient214 extends Client<Master> implements MasterClient
 {
+    /* Version 1 first version
+     * Version 2 since 2012-01-24
+     * Version 3 since 2012-02-16
+     * Version 4 since 2012-07-05
+     * Version 5 since ?
+     * Version 6 since 2014-01-07
+     * Version 7 since 2014-03-18
+     * Version 8 since 2014-08-27
```
tinwelint: A legend like this is always good to have, this could be moved to MasterClient310 (which is now the latest) and have `version 9` included as well

enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/SwitchToSlave.java
```diff
@@ -561,7 +561,7 @@ private MasterClient newMasterClient( URI masterUri, URI me, StoreId storeId, Li
     {
         MasterClient masterClient = masterClientResolver.instantiate( masterUri.getHost(), masterUri.getPort(),
                 me.getHost(), monitors, storeId, life );
-        if ( masterClient.getProtocolVersion().compareTo( MasterClient210.PROTOCOL_VERSION ) < 0 )
+        if ( masterClient.getProtocolVersion().compareTo( MasterClient214.PROTOCOL_VERSION ) < 0 )
         {
             idGeneratorFactory.enableCompatibilityMode();
```
tinwelint: Since we now have removed `MasterClient210`, then `HaIdGeneratorFactory#enableCompatibilityMode()` should also go since it was a work-around for cross-2.0->2.1 cluster migrations. `MasterClient214` has a misleading name, it's when it was created, but was actually used up until now. This condition and the compatibility mode should be removed now.

enterprise/metrics/src/main/java/org/neo4j/metrics/source/cluster/NetworkMetrics.java
```diff
@@ -64,7 +64,7 @@ public void start()
         monitors.addMonitorListener( masterNetworkTransactionWrites, MasterServer.class.getName() );
         monitors.addMonitorListener( masterNetworkStoreWrites, ToNetworkStoreWriter.class.getName(),
                 ToNetworkStoreWriter.STORE_COPIER_MONITOR_TAG );
-        monitors.addMonitorListener( slaveNetworkTransactionWrites, MasterClient210.class.getName() );
+        monitors.addMonitorListener( slaveNetworkTransactionWrites, MasterClient214.class.getName() );
```
tinwelint: Shouldn't this be the current version instead?

tinwelint:
@chrisvest @burqen had a couple of minor more comments, we're getting there...

chrisvest:
@tinwelint Comments addressed!

chrisvest:
@tinwelint Note that the build actually succeeded, but CI instability caused the PR to be marked as failed.

PR #7938, opened 2016-09-14
Fixes an issue with concurrently reading log during rotation

tinwelint:
Problem was observed when reading transaction stream from
LogicalTransactionLog when all of a sudden an IOException saying
that log header couldn't be read was thrown. This happened since
there was a concurrent rotation, log version was incremented before
new log file was fully created. There was a guard for that in
the default log version bridge but the guard only guarded for
the file not being present, not for the header not yet having
been written.

Fix is to detect this state and treat it the same way as a missing
log file for that version. We do this by introducing a specific
IncompleteLogHeaderException (extending IOException) and letting
the ReaderLogVersionBridge catch it.

As a side-effect a lot of states and how recovery and concurrent
readers may parry for different states is now well documented in
PhysicalLogFile.

Another thing was that log readers always opened the files in
read/write mode even when purely reading, which was unnecessary.
This was changed to be specified from the outside.

PhysicalLogFileRotateAndReadRaceIT captures and showcases the issue.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFile.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFile.java
@@ -178,7 +178,8 @@ public class PhysicalLogFile extends LifecycleAdapter implements LogFile
      * {@link FileNotFoundException} and tell the reader that the stream has ended</li>
      * <li>2-3: Same as (1-2)</li>
      * <li>3-4: Here the new log file exists, but the header may not be fully written yet.
-     * TODO currently the reader will fail when trying to read the header since it's reading it strictly</li>
+     * the reader will fail when trying to read the header since it's reading it strictly and bridge
+     * catches that exception, treating it the same as if the file didn't exist.</li>
      * </ol>
      *
      * @param currentLog current {@link LogVersionedStoreChannel channel} to flush and close.
@@ -268,7 +269,6 @@ public class PhysicalLogFile extends LifecycleAdapter implements LogFile
         }
 
         StoreChannel rawChannel = null;
-        boolean successful = false;
         try
         {
             rawChannel = fileSystem.open( fileToOpen, write ? "rw" : "r" );
@@ -278,7 +278,6 @@ public class PhysicalLogFile extends LifecycleAdapter implements LogFile
             assert header != null && header.logVersion == version;
             PhysicalLogVersionedStoreChannel result =
                     new PhysicalLogVersionedStoreChannel( rawChannel, version, header.logFormatVersion );
-            successful = true;
             return result;
         }
         catch ( FileNotFoundException cause )
@@ -286,17 +285,21 @@ public class PhysicalLogFile extends LifecycleAdapter implements LogFile
             throw Exceptions.withCause( new FileNotFoundException( String.format( "File could not be opened [%s]",
                     fileToOpen.getCanonicalPath() ) ), cause );
         }
-        finally
+        catch ( Throwable unexpectedError )
         {
-            if (    // we managed to open the file
-                    rawChannel != null &&
-                    // but we didn't successfully do the rest, f.ex. reading the header
-                    !successful )
+            if ( rawChannel != null )
             {
-                // then close the channel
-                rawChannel.close();
-                // ... and the exception causing us to not be successful will be thrown
+                // If we managed to open the file before failing, then close the channel
+                try
+                {
+                    rawChannel.close();
+                }
+                catch ( IOException e )
+                {
+                    unexpectedError.addSuppressed( e );
+                }
             }
+            throw unexpectedError;
         }
     }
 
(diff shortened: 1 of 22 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFile.java
```diff
@@ -142,6 +142,49 @@ public synchronized void rotate() throws IOException
...
+     * </ol>
+     *
+     * Reading: what happens when rotation is between:
+     * <ol>
+     * <li>1-2: Reader bridge will see that there's a new version (when asking {@link LogVersionRepository}
+     * and try to open it. The log file doesn't exist yet though. The bridge can parry for this by catching
+     * {@link FileNotFoundException} and tell the reader that the stream has ended</li>
+     * <li>2-3: Same as (1-2)</li>
+     * <li>3-4: Here the new log file exists, but the header may not be fully written yet.
+     * TODO currently the reader will fail when trying to read the header since it's reading it strictly</li>
```
chrisvest: Is this a problem that we need to address any time soon?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFile.java
```diff
@@ -207,31 +261,45 @@ public static PhysicalLogVersionedStoreChannel openForVersion( PhysicalLogFiles
...
+        {
+            if (    // we managed to open the file
+                    rawChannel != null &&
+                    // but we didn't successfully do the rest, f.ex. reading the header
+                    !successful )
+            {
+                // then close the channel
+                rawChannel.close();
+                // ... and the exception causing us to not be successful will be thrown
+            }
```
chrisvest: `rawChannel.close()` is allowed to throw an `IOException`, and if it does so from a finally-clause, then the exception from `close` will shadow whatever other exception that caused our `openForVersion` operation to fail. I think it'd be better if you put this code in a catch-clause, so you could add any exception from `close` as a suppressed exception to the original exception. Then you also won't need the `successful` boolean.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/IncompleteLogHeaderException.java
```diff
@@ -0,0 +1,42 @@
...
+ * Used to signal an incomplete log header, i.e. if file is smaller than the header.
+ * This exception is still an {@link IOException}, but a specific subclass of it as to make possible
+ * special handling.
+ */
+public class IncompleteLogHeaderException extends IOException
+{
+    public IncompleteLogHeaderException( File file, int readSize )
+    {
+        super( "Unable to read log version and last committed tx from '" + file.getAbsolutePath() + "', " +
+                "was only able to read " + readSize + " bytes" );
```
chrisvest: … but expected how many bytes?

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFileRotateAndReadRaceIT.java
```diff
@@ -0,0 +1,152 @@
...
+                {
+                    logFile.rotate();
+                    // Let's just close the gap to the reader so that it gets closer to the "hot zone"
+                    // where the rotation happens.
+                    writer.getCurrentPosition( startPosition );
+                    rotations.incrementAndGet();
+                }
+            }
+            return null;
+        } );
```
chrisvest, line 112: Perhaps wait for the `writeFuture` job to have started, by letting it release a latch of some kind as the first thing it does, and then let the main thread wait on that latch.

chrisvest (commented):
Generally looks good, but I have a few minor nits. Also, it needs a rebase.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFile.java
tinwelint: This TODO is actually fixed (_the_ problem this PR fixes), will remove the TODO

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFile.java
tinwelint: True, although I still need the `successful` flag to know if the `rawChannel` should be closed or not in the `finally` block, won't I? Oh, scratch that... you mean of course putting this in a `catch ( Throwable t )` clause instead.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/IncompleteLogHeaderException.java
tinwelint: Made the message include that

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFileRotateAndReadRaceIT.java
tinwelint: Yes, it now does that

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFile.java
tinwelint: Fixed

tinwelint:
@chrisvest comments addressed

PR #7941, opened 2016-09-14
Clarified descriptions of parameters

mariascharin:
Modified dbms.directories.run and dbms.logs.query.enabled.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -306,7 +306,7 @@ public abstract class GraphDatabaseSettings
             setting("dbms.index_sampling.background_enabled", BOOLEAN, TRUE );
 
     @Description("Size of buffer used by index sampling. " +
-                 "This configuration setting is no longer applicable as from Neo4j 3.0.3." +
+                 "This configuration setting is no longer applicable as from Neo4j 3.0.3. " +
                  "Please use dbms.index_sampling.sample_size_limit instead.")
     @Deprecated
     public static final Setting<Long> index_sampling_buffer_size = setting("dbms.index_sampling.buffer_size", BYTES, "64m",
@@ -466,8 +466,11 @@ public abstract class GraphDatabaseSettings
     @Description( "Relationship count threshold for considering a node to be dense" )
     public static final Setting<Integer> dense_node_threshold = setting( "dbms.relationship_grouping_threshold", INTEGER, "50", min(1) );
 
-    @Description( "Log executed queries that takes longer than the configured threshold. "
-            + "_NOTE: This feature is only available in the Neo4j Enterprise Edition_." )
+    @Description( "Log executed queries that take longer than the configured threshold, dbms.logs.query.threshold. " +
+            "Log entries are written to the file _query.log_ located in the Logs directory. " +
+            "For location of the Logs directory, see <<file-locations>>. " +
+            "This feature is available in the Neo4j Enterprise Edition." )
+
     public static final Setting<Boolean> log_queries = setting("dbms.logs.query.enabled", BOOLEAN, FALSE );
 
     @Description("Path of the logs directory")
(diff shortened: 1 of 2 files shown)


mariascharin:
@benbc Please can you review the changes to the descriptions? Thanks!

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -466,9 +466,10 @@ private static String defaultPageCacheMemory()
     @Description( "Relationship count threshold for considering a node to be dense" )
     public static final Setting<Integer> dense_node_threshold = setting( "dbms.relationship_grouping_threshold", INTEGER, "50", min(1) );
 
-    @Description( "Log executed queries that takes longer than the configured threshold. "
-            + "_NOTE: This feature is only available in the Neo4j Enterprise Edition_." )
-    public static final Setting<Boolean> log_queries = setting("dbms.logs.query.enabled", BOOLEAN, FALSE );
+    @Description( "Log executed queries that takes longer than the configured threshold, dbms.logs.query.threshold. " +
+            "Log entries are written to the file query.log located in the log directory. " +
```
benbc: "log directory"

Just checking: is this intended to be a description of the directory (in which case it's fine) or the name of the directory (in which case it should be "logs"). For consistency, would you refer to the directory holding configuration as "the configuration directory" or "the conf directory"?

community/server/src/main/java/org/neo4j/server/configuration/ServerSettings.java
```diff
@@ -215,7 +215,8 @@ private ThirdPartyJaxRsPackage createThirdPartyJaxRsPackage( String packageAndMo
     Setting<Long> gc_logging_rotation_size = setting("dbms.logs.gc.rotation.size", BYTES, "20m", min(0L), max( Long.MAX_VALUE ) );
 
     @SuppressWarnings("unused") // used only in the startup scripts
-    @Description("Path of the run directory")
+    @Description( "Path of the run directory. This directory holds the pid (process id) file for when Neo4j is running in the background. "
+            + "It may be useful to mount this to the equivalent of /tmp on Linux systems, so that it is deleted when rebooting." )
```
benbc: The _contract_ here is that `run` contains runtime state; the pidfile is the only runtime state we currently have. The bit about `/tmp` isn't strictly correct and I think that the "deleted when rebooting" is obvious to anyone who is likely to actually make use of this information.

So I suggest this instead:

> Path of the run directory. This directory holds Neo4j's runtime state, such as a pidfile when it is running in the background. It may be placed on an in-memory filesystem such as tmpfs.

Really we should have content like this on the File Locations page in the operations manual, but that page needs restructuring a bit to allow for descriptions like this.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -466,9 +466,10 @@ private static String defaultPageCacheMemory()
...
+            "_NOTE: This feature is available in the Neo4j Enterprise Edition_." )
```
jjaderberg: Could we get rid of the `_NOTE:_` while we're at it?

jjaderberg: Could we get rid of the `_NOTE:_` while we're at it?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
jjaderberg: italics for paths and file names: _query.log_

mariascharin: Yes I meant the directory, and fixed it up to be called Logs as that's how we refer to it in the File Locations section.

mariascharin:
@benbc Would you care to take a final look?

benbc:
Looks good.

PR #7942, opened 2016-09-14
Change how different editions register procedures

systay:


community/kernel/src/main/java/org/neo4j/kernel/impl/factory/EditionModule.java
```diff
@@ -68,7 +67,15 @@
  */
 public abstract class EditionModule
 {
-    public abstract void setupProcedures( Procedures procedures ) throws KernelException;
+    public void registerCrossEditionProcedures( Procedures procedures ) throws KernelException
```
fickludd: I think this could still be named simply `registerProcedures`, as it is actually registering both cross-edition and edition-specific procs.

fickludd (approved):
Well needed refactoring and tests.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/EditionModule.java
jimwebber: OK. Done.

PR #7943, opened 2016-09-14
Added dbms.setTXMetaData procedure

fickludd:
Meta-data is set as a Map<String, Object>

After procedure call completion, subsequent queries in the current transaction will be associated with this meta-data. This meta-data is included in the query log, as well as returned by the `dbms.listQueries` procedure.

This builds on #7928.

changelog [procedures] Added dbms.setTXMetaData procedure for setting application and query specific data that can clarify the results of `listQueries` results or query log entries


burqen:
@fickludd Please add a changelog tag in the description with a changelog description. Like
`changelog: A suitable description`

craigtaverner:
@burqen You only need a changelog comment if you want something different than the PR subject. In this case the subject looks good as a changelog entry to me.

burqen:
@craigtaverner Right, I don't agree that this is a useful description for an end user. What does the procedure allows me to do for example?

PR #7947, opened 2016-09-15
Clarified descriptions of parameters

mariascharin:


mariascharin:
@digitalstain In discussion with Jim, he indicated that there will never be a reason for a customer to modify these parameters on their own. If there were ever a reason to modify them at all, that would be the result of a deep analysis by a developer (i.e. you). The addition to the description is intended to convey this. Quick sanity check - OK?

digitalstain:
@mariascharin That is correct. However, these values will be set implicitly by setting the default_timeout which is not marked with the same description, for good reason. I would therefore suggest the addition of "explicitly" in the description, since the user can still implicitly change these values and that is allowed.
I would also expect the same caveat to be applied to the "learn_timeout" setting.

mariascharin:
@digitalstain Thank you! I have updated the description to include your point.
Another PR will be issued shortly to add to related ha parameters; your input will be requested there as well.

mariascharin:
Superseded by #8070

PR #7948, opened 2016-09-15
Disable writes on followers

mneedham:
This is a placeholder PR for disabling writes on followers in core-edge. At the moment you only get stopped from writing when you try to take a lock

### DIFF
--- a/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
+++ b/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
@@ -479,10 +479,12 @@ public interface Status
     enum Cluster implements Status
     {
         // transient errors
-        NoLeader( TransientError,
+        NoLeaderAvailable( TransientError,
                 "No leader available at the moment. Retrying your request at a later time may succeed." ),
 
-        ;
+        NotALeader( TransientError,
+                "The request cannot be processed by this server. Write requests can only be processed by the leader." ),
+                ;
 
         private final Code code;
 
--- a/community/graphdb-api/src/main/java/org/neo4j/graphdb/security/WriteOperationsNotAllowedException.java
+++ b/community/graphdb-api/src/main/java/org/neo4j/graphdb/security/WriteOperationsNotAllowedException.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.graphdb.security;
+
+import org.neo4j.kernel.api.exceptions.Status;
+
+public class WriteOperationsNotAllowedException extends RuntimeException implements Status.HasStatus
+{
+    private final Status statusCode;
+
+    public WriteOperationsNotAllowedException( String msg, Status statusCode )
+    {
+        super( msg );
+        this.statusCode = statusCode;
+    }
+
+    /** The Neo4j status code associated with this exception type. */
+    @Override
+    public Status status()
+    {
+        return statusCode;
+    }
+}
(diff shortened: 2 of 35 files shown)


community/graphdb-api/src/main/java/org/neo4j/graphdb/security/WriteOperationsNotAllowedException.java
```diff
@@ -0,0 +1,39 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.graphdb.security;
+
+import org.neo4j.kernel.api.exceptions.Status;
+
+public class WriteOperationsNotAllowedException extends RuntimeException implements Status.HasStatus
+{
+    private Status statusCode = Status.General.DatabaseUnavailable;
```
davidegrohmann: Maybe `ForbiddenOnReadOnlyDatabase` is a better status code?

davidegrohmann:
It looks fine, I am only wondering why we cannot reuse the `AccessMode` mechanism by wrapping the original role in a core-edge specific `OverriddenAccessMode` which does the leader check.

community/graphdb-api/src/main/java/org/neo4j/graphdb/security/WriteOperationsNotAllowedException.java
mneedham: Hmm yeh, I'm not sure how we can make it clear why the write was rejected. In the HA/Community modes the reason would be that you had a 'read_only' database but in Core-Edge mode it'd be because you weren't leader which isn't a read-only database so it might be confusing to see that in the logs.

mneedham:
Yeh I tried to figure out how to hook it into `AccessMode` but it has quite a wide surface where it's used e.g. it's called from Cypher, Core API, and Bolt as well I think.

community/graphdb-api/src/main/java/org/neo4j/graphdb/security/WriteOperationsNotAllowedException.java
davidegrohmann: Maybe the best is having a core-edge specific status code?

davidegrohmann:
I know we override the `AccessMode` to "read_only" when invoking read only procedures, that's why I thought it was relevant, but if it is too complicated I am happy to ship the current solution.

community/graphdb-api/src/main/java/org/neo4j/graphdb/security/WriteOperationsNotAllowedException.java
pontusmelke: +1 on @davidegrohmann suggestion, `ForbiddenOnFollower`? In order for the driver to handle this nicely we need something pretty specific.

mneedham: I'm making a specific code for this now. Will push in a bit.

mneedham: @pontusmelke How about this? https://github.com/neo4j/neo4j/pull/7948/commits/2623e3a26ba357325077f09b510b7beaecfe170f#diff-6bfb36f00f6d3f9f704f1617f43af598R485

community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
```diff
@@ -482,7 +482,9 @@ public Code code()
         NoLeader( TransientError,
                 "No leader available at the moment. Retrying your request at a later time may succeed." ),
 
-        ;
+        NotLeader( TransientError,
```
digitalstain: Given the previous status being named `NoLeader`, it may be confusing with the `NotLeader` name. Can we rename `NoLeader` to `NoLeaderAvailable` and `NotLeader` to `NotALeader`

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/ReadOnly.java
```diff
@@ -0,0 +1,34 @@
...
+import org.neo4j.kernel.api.exceptions.Status;
+
+public class ReadOnly implements AccessCapability
+{
+    @Override
+    public void assertCanWrite()
+    {
+        throw new WriteOperationsNotAllowedException(
+                "No write operations are allowed on this database. It is in read only mode.",
+                Status.General.DatabaseUnavailable );
```
digitalstain: I don't know what the user visible effect is, but having a status of `DatabaseUnavailable` in `WriteOperationsNotAllowedException` seems a bit contradictory.

digitalstain (commented):
No compelling feedback, just some wording choices that stood out.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/ReadOnly.java
mneedham: @digitalstain have fixed this to use `ForbiddenOnReadDatabase` instead

community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
mneedham: done

davidegrohmann:
@mneedham needs reabse

mneedham:
@davidegrohmann rebased

maxsumrall (approved):

mneedham:
Rebased again.

PR #7949, opened 2016-09-15
Ensure security log is closed on LifeCycle.shutdown

craigtaverner:
There is a bug that only happens on windows, and can be seen in the StoreCopyClientTest when deleting log files, the security log cannot be deleted due to a lock.

There are many differences between the implementation of SecurityLog and the other logs. One that appears very relevant to this case is the fact that the debug.log is in LogService which implements LifeCycle, and when the shutdown() method is called, the LogService closes the OutputStream to the debug.log (See StoreLogService.java:195).

So I've implemented LifeCycle on SecurityLog and added calls to init and shutdown from the existing LifeCycle methods on the enterprise auth manager. We could have added SecurityLog to the lifecycle directly, as is done with debug.log in PlatformModule, but the edition differences make that messy, so it was easier to add these calls in the MultiRealmAuthManager where the security log is created and managed directly.

changelog: Fixed problem on windows with security log not being closed before backup is run, by tying it to the lifecycle shutdown

### DIFF
--- a/enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/SecurityLog.java
+++ b/enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/SecurityLog.java
@@ -29,6 +29,7 @@ import org.neo4j.kernel.api.security.AuthSubject;
 import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.impl.enterprise.configuration.EnterpriseEditionSettings;
 import org.neo4j.kernel.impl.util.JobScheduler;
+import org.neo4j.kernel.lifecycle.LifecycleAdapter;
 import org.neo4j.logging.FormattedLog;
 import org.neo4j.logging.Log;
 import org.neo4j.logging.Logger;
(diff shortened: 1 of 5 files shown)


enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
```diff
@@ -187,6 +188,7 @@ public void shutdown() throws Throwable
                 ((RealmLifecycle) realm).shutdown();
             }
         }
+        securityLog.shutdown();
```
henriknyman: I noted that `EnterpriseAuthManagerFactory` has a piece of defensive code that can either accept an externally injected log or create a new one. 
`private SecurityLog getSecurityLog( Log allegedSecurityLog )`

If we get the log injected we should be careful about how we manage its life cycle.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/SecurityLog.java
```diff
@@ -38,35 +39,31 @@
...
 
     public SecurityLog( Config config, FileSystemAbstraction fileSystem, Executor executor ) throws IOException
     {
-        this( createLog( config, fileSystem, executor ) );
-    }
-
-    public SecurityLog( Log log )
-    {
-        inner = log;
-    }
```
henriknyman, line 53: I believe this alternative constructor is still used by both tests and `EnterpriseAuthManagerFactory.getSecurityLog`.

henriknyman (changes requested):

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/SecurityLog.java
```diff
@@ -38,35 +39,31 @@
 import static org.neo4j.helpers.Strings.escape;
 import static org.neo4j.kernel.impl.enterprise.configuration.EnterpriseEditionSettings.security_log_filename;
 
-public class SecurityLog implements Log
+public class SecurityLog implements Log, Lifecycle
```
MishaDemianenko: Why not to extend LifecycleAdapter instead for example and just override `shutdown`?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/SecurityLog.java
craigtaverner: I reviewed the usage and the only non-test place it is used it gets from EditionModel.authManagerLog, which is always either the real security log, or a null log. So this means we're back to two cases, a closable log or a non-closable one, and that is dealt with in the condition inside the shutdown method.

I would prefer a better approach, but the use of two constructors for setting up SecurityLog is a pre-existing design issue I did not want to resolve in this PR, which focuses only on adding the close() call to the real security log (and not to test or null-logs).

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
craigtaverner: Agreed, but as described in my other response below, I see the issue with the two SecurityLog constructors as orthogonal to this fix. Injected logs are not closable or support lifecycle, so we cannot interact with them at all.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/SecurityLog.java
craigtaverner: Good idea.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/SecurityLog.java
henriknyman: Sure, I mostly meant that it wouldn't compile without it. But I see now that you have added it again below.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
henriknyman: The `if ( this.rotatingSupplier != null )` accomplishes that

henriknyman (approved):

PR #7959, opened 2016-09-16
Add tests for FileVisitor combinators

benbc:


praveenag (approved):
LGTM. Just for future reference, you can use.

https://github.com/junit-team/junit4/wiki/exception-testing#expectedexception-rule

PR #7960, opened 2016-09-16
Reduce build time by making some tests faster

tinwelint:
Mostly in kernel and "neo4j" components for now. More to come as we go.

### DIFF
--- a/community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
+++ b/community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
@@ -161,15 +161,14 @@ public class TestDirectory implements TestRule
         return directory( "graph-db" );
     }
 
-    public File makeGraphDbDir()
+    public File makeGraphDbDir() throws IOException
     {
         return cleanDirectory( "graph-db" );
     }
 
     public void cleanup() throws IOException
     {
-        fileSystem.deleteRecursively( testClassBaseFolder );
-        fileSystem.mkdirs( testClassBaseFolder );
+        clean( fileSystem, testClassBaseFolder );
     }
 
     @Override
@@ -179,14 +178,18 @@ public class TestDirectory implements TestRule
         return format( "%s[%s]", getClass().getSimpleName(), testDirectoryName );
     }
 
-    public File cleanDirectory( String name )
+    public File cleanDirectory( String name ) throws IOException
     {
-        File dir = new File( ensureBase(), name );
-        if ( fileSystem.fileExists( dir ) )
+        return clean( fileSystem, new File( ensureBase(), name ) );
+    }
+
+    private static File clean( FileSystemAbstraction fs, File dir ) throws IOException
+    {
+        if ( fs.fileExists( dir ) )
         {
-            deleteSilently( dir );
+            fs.deleteRecursively( dir );
         }
-        fileSystem.mkdir( dir );
+        fs.mkdirs( dir );
         return dir;
     }
 
@@ -211,19 +214,7 @@ public class TestDirectory implements TestRule
         testDirectory = null;
     }
 
-    private void deleteSilently( File dir )
-    {
-        try
-        {
-            fileSystem.deleteRecursively( dir );
-        }
-        catch ( IOException e )
-        {
-            throw new RuntimeException( e );
-        }
-    }
-
-    private File directoryForDescription( Description description )
+    private File directoryForDescription( Description description ) throws IOException
     {
         if ( owningTest == null )
         {
@@ -267,8 +258,7 @@ public class TestDirectory implements TestRule
         File result = new File( testData, owningTest.getName() ).getAbsoluteFile();
         if ( clean )
         {
-            fs.deleteRecursively( result );
-            fs.mkdirs( result );
+            clean( fs, result );
         }
         return result;
     }
--- a/community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
+++ b/community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
@@ -247,11 +247,7 @@ public class TestDirectory implements TestRule
         }
     }
 
-    /**
-     * This method can be used outside the context of a Rule as a utility. It will return something like:
-     * {@code component/target/test-data/org.neo4j.location.of.my.test.Clazz/}
-     */
-    public static File testDataDirectoryOf( FileSystemAbstraction fs, Class<?> owningTest, boolean clean )
+    private static File testDataDirectoryOf( FileSystemAbstraction fs, Class<?> owningTest, boolean clean )
             throws IOException
     {
         File testData = new File( locateTarget( owningTest ), "test-data" );
--- a/community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
+++ b/community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
@@ -246,8 +246,31 @@ public class TestDirectory implements TestRule
         {
             throw new IllegalStateException(" Test owning class is not defined" );
         }
+        try
+        {
+            testClassBaseFolder = testDataDirectoryOf( fileSystem, owningTest, false );
+        }
+        catch ( IOException e )
+        {
+            throw new RuntimeException( e );
+        }
+    }
+
+    /**
+     * This method can be used outside the context of a Rule as a utility. It will return something like:
+     * {@code component/target/test-data/org.neo4j.location.of.my.test.Clazz/}
+     */
+    public static File testDataDirectoryOf( FileSystemAbstraction fs, Class<?> owningTest, boolean clean )
+            throws IOException
+    {
         File testData = new File( locateTarget( owningTest ), "test-data" );
-        testClassBaseFolder = new File( testData, owningTest.getName() ).getAbsoluteFile();
+        File result = new File( testData, owningTest.getName() ).getAbsoluteFile();
+        if ( clean )
+        {
+            fs.deleteRecursively( result );
+            fs.mkdirs( result );
+        }
+        return result;
     }
 
     private void register( String test, String dir )
--- a/community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
@@ -487,6 +487,7 @@ public class NeoStoreDataSource implements Lifecycle, IndexProviders
             dependencies.satisfyDependency( storageEngine.storeReadLayer() );
             dependencies.satisfyDependency( logEntryReader );
             dependencies.satisfyDependency( storageEngine );
+            dependencies.satisfyDependency( transactionLogModule.checkPointing() );
             satisfyDependencies( kernelModule );
         }
         catch ( Throwable e )
--- a/community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
@@ -487,7 +487,6 @@ public class NeoStoreDataSource implements Lifecycle, IndexProviders
             dependencies.satisfyDependency( storageEngine.storeReadLayer() );
             dependencies.satisfyDependency( logEntryReader );
             dependencies.satisfyDependency( storageEngine );
-            dependencies.satisfyDependency( transactionLogModule.checkPointing() );
             satisfyDependencies( kernelModule );
         }
         catch ( Throwable e )
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/util/DebugUtil.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/util/DebugUtil.java
@@ -29,13 +29,17 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
 
 import static java.lang.String.format;
+import static java.lang.System.currentTimeMillis;
+import static java.lang.System.nanoTime;
 import static java.lang.reflect.Modifier.isPublic;
 import static java.lang.reflect.Modifier.isStatic;
 import static org.neo4j.helpers.Exceptions.stringify;
+import static org.neo4j.helpers.Format.duration;
 
 public class DebugUtil
 {
@@ -363,5 +367,79 @@ public class DebugUtil
             }
         }
         return false;
+    }
+
+    /**
+     * Super simple utility for determining where most time is spent when you don't know where to even start.
+     * It could be used to home in on right place in a test or in a sequence of operations or similar.
+     */
+    public abstract static class Timer
+    {
+        private final TimeUnit unit;
+        private long startTime;
+
+        protected Timer( TimeUnit unit )
+        {
+            this.unit = unit;
+            this.startTime = currentTime();
+        }
+
+        protected abstract long currentTime();
+
+        public void reset()
+        {
+            startTime = currentTime();
+        }
+
+        public void at( String point )
+        {
+            long duration = currentTime() - startTime;
+            System.out.println( duration( unit.toMillis( duration ) ) + " @ " + point );
+            startTime = currentTime();
+        }
+
+        public static Timer millis()
+        {
+            return new Millis();
+        }
+
+        private static class Millis extends Timer
+        {
+            Millis()
+            {
+                super( TimeUnit.MILLISECONDS );
+            }
+
+            @Override
+            protected long currentTime()
+            {
+                return currentTimeMillis();
+            }
+        }
+
+        public static Timer nanos()
+        {
+            return new Nanos();
+        }
+
+        private static class Nanos extends Timer
+        {
+            Nanos()
+            {
+                super( TimeUnit.NANOSECONDS );
+            }
+
+            @Override
+            protected long currentTime()
+            {
+                return nanoTime();
+            }
+        }
+    }
+
+    public static long time( long startTime, String message )
+    {
+        System.out.println( duration( (currentTimeMillis() - startTime) ) + ": " + message );
+        return currentTimeMillis();
     }
 }
--- a/community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
@@ -1106,7 +1106,6 @@ public class BatchInserterImpl implements BatchInserter
     // test-access
     NeoStores getNeoStores()
     {
-        forceFlushChanges();
         return neoStores;
     }
 
--- a/community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
@@ -750,8 +750,7 @@ public class BatchInserterImpl implements BatchInserter
     private void setNodeLabels( NodeRecord nodeRecord, Label... labels )
     {
         NodeLabels nodeLabels = parseLabelsField( nodeRecord );
-        nodeStore.updateDynamicLabelRecords( nodeLabels.put( getOrCreateLabelIds( labels ), nodeStore,
-                nodeStore.getDynamicLabelStore() ) );
+        nodeLabels.put( getOrCreateLabelIds( labels ), nodeStore, nodeStore.getDynamicLabelStore() );
         labelsTouched = true;
     }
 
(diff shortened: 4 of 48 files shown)


tinwelint:
@benbc wdyt?

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -487,6 +487,7 @@ public void start() throws IOException
             dependencies.satisfyDependency( storageEngine.storeReadLayer() );
             dependencies.satisfyDependency( logEntryReader );
             dependencies.satisfyDependency( storageEngine );
+            dependencies.satisfyDependency( transactionLogModule.checkPointing() );
```
benbc: This looks like a surprising change for test speed-up work.

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
```diff
@@ -750,8 +750,7 @@ protected PropertyBlock underlyingObjectToObject( Entry<String, Object> property
     private void setNodeLabels( NodeRecord nodeRecord, Label... labels )
     {
         NodeLabels nodeLabels = parseLabelsField( nodeRecord );
-        nodeStore.updateDynamicLabelRecords( nodeLabels.put( getOrCreateLabelIds( labels ), nodeStore,
-                nodeStore.getDynamicLabelStore() ) );
+        nodeLabels.put( getOrCreateLabelIds( labels ), nodeStore, nodeStore.getDynamicLabelStore() );
```
benbc, line 753: Is this a semantic change? You should probably get someone from the Kernel team to review this as well as me.

community/kernel/src/test/java/org/neo4j/graphdb/GraphDatabaseServiceTest.java
```diff
@@ -71,44 +75,45 @@ public void givenDatabaseAndStartedTxWhenShutdownThenWaitForTxToFinish() throws
...
+        // i.e. wait for transaction to start
+        barrier.await();
+
+        // now there's a transaction open, blocked on continueTxSignal
+        Future<Object> shutdownFuture = t3.execute( state ->
+        {
+            db.shutdown();
+            return null;
+        } );
+        t3.get().waitUntilWaiting( location -> location.isAt( DatabaseAvailability.class, "shutdown" ) );
```
benbc: Does this constitute an assertion that the shutdown is actually blocked by the in-flight transaction?

community/kernel/src/test/java/org/neo4j/graphdb/GraphDatabaseServiceTest.java
```diff
@@ -71,44 +75,45 @@ public void givenDatabaseAndStartedTxWhenShutdownThenWaitForTxToFinish() throws
...
+        barrier.release();
+        try
+        {
+            txFuture.get();
+        }
+        catch ( ExecutionException e )
+        {
+            // expected
+        }
+        shutdownFuture.get();
```
benbc, line 109: Shouldn't this (and other `get()`s) have a timeout?

community/kernel/src/test/java/org/neo4j/graphdb/GraphDatabaseServiceTest.java
```diff
@@ -172,81 +173,44 @@ public void terminateNestedTransactionThrowsExceptionOnNextNestedOperation() thr
                 }
             }
         }
-
-        db.shutdown();
     }
 
     @Test
     public void givenDatabaseAndStartedTxWhenShutdownAndStartNewTxThenBeginTxTimesOut() throws Exception
```
benbc, line 179: Shouldn't this be `BeginTxThrowsAnException`?

community/kernel/src/test/java/org/neo4j/graphdb/GraphDatabaseServiceTest.java
```diff
@@ -172,81 +173,44 @@ public void terminateNestedTransactionThrowsExceptionOnNextNestedOperation() thr
...
+            db.beginTx();
+            fail( "Should fail" );
+        }
+        catch ( DatabaseShutdownException e )
+        {
+            //THEN good
         }
-
-        assertThat( result.get(), instanceOf( DatabaseShutdownException.class ) );
     }
```
benbc, line 214: My understanding from the test name is that this is intended to test the case where a transaction starts while the shutdown is in progress. So I would expect the `beginTx()` to come before the `barrier.release()`. Otherwise I don't see any reason to have the blocking tx in this test -- you  might as well just do a simple `db.shutdown()` followed by `db.beginTx()`.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionsTest.java
```diff
@@ -472,41 +475,28 @@ private static TransactionCommitProcess newRememberingCommitProcess( final Trans
...
+//        return Executors.newSingleThreadExecutor().submit( new Callable<KernelTransaction>()
+//        {
+//            @Override
+//            public KernelTransaction call()
+//            {
+//                aboutToStartTx.countDown();
+//                return kernelTransactions.newInstance( KernelTransaction.Type.explicit, AccessMode.Static.WRITE, 0L );
+//            }
+//        } );
+//    }
```
benbc: Can this just be deleted?

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointerIntegrationTest.java
```diff
@@ -106,25 +112,46 @@ public void shouldCheckPointBasedOnTime() throws Throwable
...
-        TimeUnit.SECONDS.sleep( seconds );
-
-        // give a chance to the check pointer to be scheduled and run...
-        TimeUnit.SECONDS.sleep( 1 );
+        // The scheduled job checking whether or not checkpoints are needed runs more frequently
+        // now that we've set the time interval so low, so we can simply wait for it here
+        while ( !checkPointInTxLog( db ) )
+        {
+            Thread.sleep( millis );
+        }
```
benbc: Does this need a timeout in case checkpointing is completely broken somehow?

community/kernel/src/test/java/org/neo4j/test/OtherThreadExecutor.java
```diff
@@ -217,17 +220,45 @@ public String toString()
...
     public WaitDetails waitUntilBlocked() throws TimeoutException
     {
-        return waitUntilThreadState( Thread.State.BLOCKED );
+        return waitUntilBlocked( details -> true );
+    }
+
+    public WaitDetails waitUntilWaiting( Predicate<WaitDetails> correctWait ) throws TimeoutException
+    {
+        return waitUntilThreadState( details -> true, Thread.State.WAITING, Thread.State.TIMED_WAITING );
+    }
```
benbc: ignores the predicate

community/kernel/src/test/java/org/neo4j/test/OtherThreadExecutor.java
```diff
@@ -217,17 +220,45 @@ public String toString()
...
+
+    public WaitDetails waitUntilBlocked( Predicate<WaitDetails> correctWait ) throws TimeoutException
+    {
+        return waitUntilThreadState( details -> true, Thread.State.BLOCKED );
```
benbc: Ignores the predicate

benbc (commented):
I'm afraid my eyes glazed over a bit towards the end there. Looks like some good improvements, though. How much speed-up do you think there is overall?

benbc:
Please get someone from your team to review as well. I don't feel qualified to have an opinion on some of that stuff.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
tinwelint: This is about exposing a way for tests to force a checkpoint, no biggie

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
tinwelint: If you look at this change in the context of the specific commit, there's a description about the problem and that it fixes this bug

community/kernel/src/test/java/org/neo4j/graphdb/GraphDatabaseServiceTest.java
tinwelint: Yes exactly, this technique has existed for years in the neo4j test code base, although many tests don't take advantage of it, instead waiting an arbitrary amount of seconds, catch a TimeoutException and interpret that is the same result. This technique is actually better in every way because we can very specifically determine where the wait happens and it can be detected much faster.

community/kernel/src/test/java/org/neo4j/graphdb/GraphDatabaseServiceTest.java
tinwelint: We could, but there's no legit reason for shutdown to be blocked here, just as there's no legit reason for a normal shutdown of a db which happens in most of our tests to be blocked.

community/kernel/src/test/java/org/neo4j/graphdb/GraphDatabaseServiceTest.java
tinwelint: I thought so too, but the test looks made for a very specific bug perhaps and so didn't want to change something I didn't understand.

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointerIntegrationTest.java
tinwelint: Yup, added

community/kernel/src/test/java/org/neo4j/test/OtherThreadExecutor.java
tinwelint: Yes, fixed

community/kernel/src/test/java/org/neo4j/test/OtherThreadExecutor.java
tinwelint: Yes, fixed

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionsTest.java
tinwelint: Deleted now

community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
```diff
@@ -246,8 +246,31 @@ private void evaluateClassBaseTestFolder( )
...
+        {
+            throw new RuntimeException( e );
+        }
+    }
+
+    /**
+     * This method can be used outside the context of a Rule as a utility. It will return something like:
+     * {@code component/target/test-data/org.neo4j.location.of.my.test.Clazz/}
+     */
+    public static File testDataDirectoryOf( FileSystemAbstraction fs, Class<?> owningTest, boolean clean )
```
MishaDemianenko: all methods that create test directory is called testDirectory(something...) this static method should be called in the same way

community/kernel/src/test/java/org/neo4j/unsafe/batchinsert/internal/BatchInsertTest.java
```diff
@@ -1722,6 +1559,19 @@ private void setAndGet( BatchInserter inserter, Object value )
...
+    {
+        try ( Transaction ignored = db.beginTx() )
+        {
+            return db.getNodeById( nodeId );
+        }
+    }
+
+    private void forceFlush( BatchInserter inserter )
+    {
+        ((BatchInserterImpl)inserter).getNeoStores();
```
MishaDemianenko: this is extremely confusing. We expose method forceFlushChanges to flush changes

community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
```diff
@@ -246,8 +246,31 @@ private void evaluateClassBaseTestFolder( )
...
+            throws IOException
+    {
         File testData = new File( locateTarget( owningTest ), "test-data" );
-        testClassBaseFolder = new File( testData, owningTest.getName() ).getAbsoluteFile();
+        File result = new File( testData, owningTest.getName() ).getAbsoluteFile();
+        if ( clean )
```
MishaDemianenko, line 255: there is already method cleanDirectory that with a bit of adaptation will be suitable for static and general case

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
```diff
@@ -1102,6 +1103,18 @@ private void dumpConfiguration( Map<String,String> config, PrintStream out )
         }
     }
 
+    // test-access
+    NeoStores getNeoStores()
+    {
+        forceFlushChanges();
```
MishaDemianenko: this is confusing i think it's up to test method to do explicit call to flush changes instead.

community/kernel/src/test/java/org/neo4j/unsafe/batchinsert/internal/BatchInsertTest.java
```diff
@@ -188,13 +192,37 @@ public BatchInsertTest( int denseNodeThreshold )
...
 
+    private static BatchInserter globalInserter;
+
+    @BeforeClass
+    public static void startGlobalInserter() throws IOException
+    {
+        // Global inserter can be used in tests which simply want to verify "local" behaviour,
+        // e.g. create a node with some properties and read them back.
+        globalInserter = BatchInserters.inserter(
+                TestDirectory.testDataDirectoryOf( REAL_FS, BatchInsertTest.class, true ),
```
MishaDemianenko: fs should be used here right?

community/kernel/src/test/java/org/neo4j/kernel/impl/locking/StopCompatibility.java
```diff
@@ -556,16 +554,9 @@ private void assertNoLocksHeld()
 
     private void assertThreadIsWaitingForLock( LockAcquisition lockAcquisition ) throws Exception
     {
-        for ( int i = 0; i < 30; i++ )
+        for ( int i = 0; i < 30 && !suite.isAwaitingLockAcquisition( lockAcquisition.executor.waitUntilWaiting() ); i++ )
```
MishaDemianenko, line 557: that's not equivalent behaviour to what we have before since before we where checking that we actually waiting for a lock for some time now it can happen that we just in a method that was expected.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/DebugUtil.java
```diff
@@ -364,4 +368,78 @@ private static boolean hasTestAnnotation( Method method )
         }
         return false;
     }
+
+    /**
+     * Super simple utility for determining where most time is spent when you don't know where to even start.
+     * It could be used to home in on right place in a test or in a sequence of operations or similar.
+     */
+    public abstract static class Timer
```
MishaDemianenko, line 376: Those kind of things are called `Stopwatch` and we have already one from Guava and from apache on our test class path do not see why we need something else...

community/kernel/src/test/java/org/neo4j/kernel/impl/store/MetaDataStoreTest.java
```diff
@@ -399,22 +391,33 @@ public void testRecordTransactionClosed() throws Exception
...
-            AtomicBoolean stopped = new AtomicBoolean();
-            AtomicLong counter = new AtomicLong();
-
-            Runnable writer = untilStopped( stopped, runLatch, () -> {
-                long count = counter.incrementAndGet();
+            AtomicLong writeCount = new AtomicLong();
+            AtomicLong fileReadCount = new AtomicLong();
+            AtomicLong apiReadCount = new AtomicLong();
+            int limit = 10_000;
+            long endTime = currentTimeMillis() + SECONDS.toMillis( 1 );
```
MishaDemianenko: with that small endTime it's really big chance that on CI it will just will not continue based on endTime, i think we need to wait for at least of couple writes to happen

community/kernel/src/test/java/org/neo4j/kernel/impl/store/MetaDataStoreTest.java
```diff
@@ -595,25 +562,16 @@ public void transactionClosedMustBeAtomic() throws Exception
...
-            runLatch.await( 1, TimeUnit.SECONDS );
-            stopped.set( true );
-
-            for ( Future<?> future : readerFutures )
-            {
-                future.get(); // We assert that this does not throw
-            }
+                apiReadCount.incrementAndGet();
+            } ) );
+            race.go();
```
MishaDemianenko, line 574: same as above

community/kernel/src/test/java/org/neo4j/kernel/impl/store/MetaDataStoreTest.java
```diff
@@ -542,46 +508,47 @@ public void transactionCommittedMustBeAtomic() throws Exception
...
-            readerFutures.addAll( forkMultiple( 5, apiReader ) );
-
-            runLatch.await( 1, TimeUnit.SECONDS );
-            stopped.set( true );
-
-            for ( Future<?> future : readerFutures )
-            {
-                future.get(); // We assert that this does not throw
-            }
+            race.go();
```
MishaDemianenko, line 521: same as above here, we need to wait for couple of expected events to happen at least, too aggressive end time can just cancel race

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
MishaDemianenko: i think you need this at all, for example because  https://github.com/neo4j/neo4j/blob/3.1/community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java#L464
hack, and there is for example CheckpointerSupplier that rely on checkpointer to be registered

MishaDemianenko (commented):
@tinwelint  Review round completed please find comments on appropriate lines.
Also general comment: I do not find `Race` thingy readable in most of the cases more confusing....
Also in most of the races place previous assertions just gone and we check only that race did not fail which is strange - we should check at least that our desired state/condition was achieved at least once, also having some assertions during the race should clarify situation.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
tinwelint: Oh cool, then it can probably be removed

community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
tinwelint: It's not a rule factory method and so wanted to call it something else, actually it should be called something very different to not be intermixed with the common rule factory methods, shouldn't it?

community/io/src/test/java/org/neo4j/test/rule/TestDirectory.java
tinwelint: OK, slightly beyond the scope of this PR, but will extract this method anyway

community/kernel/src/test/java/org/neo4j/unsafe/batchinsert/internal/BatchInsertTest.java
tinwelint: Alright, the thinking was that you get NeoStores to look at records things and when you do you want all changes to be present. Sure this method could be used to get the NeoStores at after creation and then just use it. I see your point.

community/kernel/src/test/java/org/neo4j/kernel/impl/locking/StopCompatibility.java
tinwelint: It is equivalent and even asserts more than we did previously. In this piece of code we wait for the thread to go into a _waiting_ state AND see that it waits in the correct place. The previous code only verified that the thread hadn't acquired the lock in N seconds, it could have been caused by anything.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/DebugUtil.java
tinwelint: How is Stopwatch the same thing? This simple utility prints times at different points where it's used. It's the printing which is useful and it's only a utility quite useful when seeing what takes time in a test.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/MetaDataStoreTest.java
tinwelint: Right we could perhaps increase that to 10 seconds or something, just to be safe AND put a lower bound of at least 100 each.

tinwelint:
@MishaDemianenko please tell me more about your concerns with `Race`, imho it makes race testing small, consice and less error prone than any other means I've seen in our testing. Also where have I accidentally removed assertions?

MishaDemianenko:
@tinwelint looks good now, will merge on green

PR #7964, opened 2016-09-17
Tackling buffer NPE in ChunkedOutput

technige:
May be doubling up with @pontusmelke's work. Don't merge until we've confirmed this is still needed.

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/v1/transport/ChunkedOutput.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/v1/transport/ChunkedOutput.java
@@ -24,6 +24,7 @@ import io.netty.channel.Channel;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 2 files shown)


community/bolt/src/main/java/org/neo4j/bolt/v1/transport/ChunkedOutput.java
```diff
@@ -42,10 +43,12 @@
 
     private final int bufferSize;
     private final int maxChunkSize;
+    private final Semaphore bufferToken = new Semaphore( 1 );
```
pontusmelke, line 46: I don't think there is much point in having a semaphore of size 1. I think using synchronized is sufficient and then we don't have to worry about not releasing the locks

pontusmelke:
duplicate of https://github.com/neo4j/neo4j/pull/7963

technige:
Closing as fixed in #7963

PR #7967, opened 2016-09-19
Break out fileSystem method from OutsideWorld interface

burqen:
And have all commands use provided FileSystemAbstraction instead of creating their own.

### DIFF
--- a/community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
+++ b/community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
@@ -24,6 +24,7 @@ import java.util.Optional;
 
 import org.neo4j.helpers.Service;
 import org.neo4j.helpers.collection.Iterables;
+import org.neo4j.io.fs.FileSystemAbstraction;
 
 /**
  * To create a command for {@code neo4j-admin}:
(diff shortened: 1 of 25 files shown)


burqen:
@benbc What do you say about this?

benbc:
@burqen I say: "why?"

burqen:
Decouple file system from output makes the responsibility of the interface clearer.
And having all commands use provided `FileSystemAbstraction` would be the behaviour I would expect.

burqen:
The reason I ended up here is it causes a compilation error in blockdevice where we would have to provide null for file system abstraction.

benbc:
> having all commands use provided FileSystemAbstraction would be the behaviour I would expect

Does refer to the fix in `DumpCommand` to avoid instantiating the FSA? You are certainly right there; well spotted.

I'm not sure that your change makes the responsibility of the interface cleaner. I think it's reasonable to have a single collaborator that encapsulates all I/O. I think generally your change just makes the provision of fakes in tests more of a pain (c.f. the various places that you've created a mock FSA just in order to fake it out.

However if you've got a genuine problem in blockdevice, then we must fix it. Can you expand on that, I haven't understood what you are saying?

burqen:
@benbc I don't have any real problem right now, so I don't have any strong opinion about this. Will gladly close it as you don't find it helpful :)

PR #7969, opened 2016-09-19
Retry Edge Server startup if failed on download from core server.

maxsumrall:
Previously we would retry when attempting to get the store Id. This change will cause us to retry on any error when starting up an edge server. Each retry attempt will randomly pick a core server to try to download from.

### DIFF
--- a/enterprise/core-edge/src/main/java/org/neo4j/coreedge/edge/EdgeStartupProcess.java
+++ b/enterprise/core-edge/src/main/java/org/neo4j/coreedge/edge/EdgeStartupProcess.java
@@ -105,12 +105,12 @@ class EdgeStartupProcess implements Lifecycle
             try
             {
                 Thread.sleep( retryInterval );
-                retryInterval = retryInterval * 2;
+                retryInterval = Math.min( 60_000, retryInterval * 2 );
             }
             catch ( InterruptedException e )
             {
                 Thread.interrupted();
-                throw new RuntimeException( "Interrupted while trying to start edge server. Shutting down.", e );
+                throw new RuntimeException( "Interrupted while trying to start edge server.", e );
             }
         }
         throw new Exception( "Failed to start edge server after " + (attempts - 1) + " attempts" );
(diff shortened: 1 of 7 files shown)


enterprise/core-edge/src/main/java/org/neo4j/coreedge/edge/EdgeStartupProcess.java
```diff
@@ -76,7 +77,47 @@ public void init() throws Throwable
...
+            }
+            catch ( StoreIdDownloadFailedException e )
+            {
+                log.info( "Attempt #%d to start edge server failed while getting store id from %s.", attempts, source );
+            }
+
+            try
+            {
+                Thread.sleep( retryInterval );
+                retryInterval = retryInterval * 2;
```
digitalstain: Maybe it's worth capping this value to something like 30'' or 1' so future changes in the try count do not end up with extreme wait values.

enterprise/core-edge/src/main/java/org/neo4j/coreedge/edge/EdgeStartupProcess.java
```diff
@@ -76,7 +77,47 @@ public void init() throws Throwable
...
+            }
+            catch ( InterruptedException e )
+            {
+                Thread.interrupted();
+                throw new RuntimeException( "Interrupted while trying to start edge server. Shutting down.", e );
```
digitalstain: How is the shut down mentioned in the exception message achieved here?

enterprise/core-edge/src/main/java/org/neo4j/coreedge/edge/EdgeStartupProcess.java
digitalstain: Maybe it's worth capping this value to something like 30'' or 1' so future changes in the try count do not end up with extreme wait values.

enterprise/core-edge/src/test/java/org/neo4j/coreedge/edge/EdgeStartupProcessTest.java
```diff
@@ -109,9 +109,9 @@ public void shouldNotStartWithMismatchedNonEmptyStore() throws Throwable
             edgeStartupProcess.start();
             fail( "should have thrown" );
         }
-        catch ( IllegalStateException ex )
+        catch ( Exception ex )
```
digitalstain, line 113: The tricky part with expecting Exception is that it can be thrown from anywhere in the code. I think that either the code should not throw Exception in start() (introduce a new type, perhaps) or the catch should assert on the exception message.

PR #7971, opened 2016-09-19
Make enterprise built-in procedures work with auth disabled

henriknyman:
Some built-in procedures in enterprise requires an enterprise auth subject to work (the concept of admin only exists in enterprise).


technige:
LGTM :+1:

PR #7973, opened 2016-09-20
Proper namespace resolution

pontusmelke:
It should be possible to use the names of the built-in functions as long
as you put the function in a separate namespace. The lookup was erroneously
using only the name of the function when doing lookup which lead to that
user-defined functions were shadowed by the built-in function with the same name.

Also added capability to test functions in `neo4j-harness`.

### DIFF
--- a/community/neo4j-harness/src/main/java/org/neo4j/harness/internal/AbstractInProcessServerBuilder.java
+++ b/community/neo4j-harness/src/main/java/org/neo4j/harness/internal/AbstractInProcessServerBuilder.java
@@ -255,7 +255,7 @@ public abstract class AbstractInProcessServerBuilder implements TestServerBuilde
     /**
      * A kernel extension used to ensure we load user-registered procedures
      * after other kernel extensions have initialized, since kernel extensions
-     * can addProcedure custom injectables that procedures need.
+     * can add custom injectables that procedures need.
      */
     private static class Neo4jHarnessExtensions extends KernelExtensionFactory<Neo4jHarnessExtensions.Dependencies>
     {
--- a/community/neo4j-harness/src/main/java/org/neo4j/harness/internal/AbstractInProcessServerBuilder.java
+++ b/community/neo4j-harness/src/main/java/org/neo4j/harness/internal/AbstractInProcessServerBuilder.java
@@ -196,7 +196,14 @@ public abstract class AbstractInProcessServerBuilder implements TestServerBuilde
     @Override
     public TestServerBuilder withProcedure( Class<?> procedureClass )
     {
-        procedures.add( procedureClass );
+        procedures.addProcedure( procedureClass );
+        return this;
+    }
+
+    @Override
+    public TestServerBuilder withFunction( Class<?> functionClass )
+    {
+        procedures.addFunction( functionClass );
         return this;
     }
 
@@ -248,7 +255,7 @@ public abstract class AbstractInProcessServerBuilder implements TestServerBuilde
     /**
      * A kernel extension used to ensure we load user-registered procedures
      * after other kernel extensions have initialized, since kernel extensions
-     * can add custom injectables that procedures need.
+     * can addProcedure custom injectables that procedures need.
      */
     private static class Neo4jHarnessExtensions extends KernelExtensionFactory<Neo4jHarnessExtensions.Dependencies>
     {
--- a/community/neo4j-harness/src/test/java/org/neo4j/harness/JavaFunctionsTest.java
+++ b/community/neo4j-harness/src/test/java/org/neo4j/harness/JavaFunctionsTest.java
@@ -65,7 +65,7 @@ public class JavaFunctionsTest
         @Context
         public SomeService service;
 
-        @UserFunction("my.hello")
+        @UserFunction( "my.hello" )
         public String hello()
         {
             return service.hello();
@@ -77,7 +77,7 @@ public class JavaFunctionsTest
         @Context
         public MyCoreAPI myCoreAPI;
 
-        @UserFunction( value = "my.willFail")
+        @UserFunction( value = "my.willFail" )
         public long willFail() throws ProcedureException
         {
             return myCoreAPI.makeNode( "Test" );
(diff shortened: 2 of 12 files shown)


community/neo4j-harness/src/main/java/org/neo4j/harness/internal/AbstractInProcessServerBuilder.java
```diff
@@ -248,7 +255,7 @@ private static String toStringForThirdPartyPackageProperty( List<ThirdPartyJaxRs
     /**
      * A kernel extension used to ensure we load user-registered procedures
      * after other kernel extensions have initialized, since kernel extensions
-     * can add custom injectables that procedures need.
+     * can addProcedure custom injectables that procedures need.
```
Mats-SX: Is this change intentional?

community/neo4j-harness/src/test/java/org/neo4j/harness/JavaFunctionsTest.java
```diff
@@ -0,0 +1,176 @@
...
+        {
+            return myCoreAPI.countNodes();
+        }
+    }
+
+    @Test
+    public void shouldLaunchWithDeclaredFunctions() throws Exception
+    {
+        // When
+        try(ServerControls server = TestServerBuilders.newInProcessBuilder().withFunction( MyFunctions.class ).newServer())
```
Mats-SX: Missing some padding whitespace, also in `try-with-resources` below

integrationtests/src/test/java/org/neo4j/procedure/UserFunctionIT.java
```diff
@@ -685,6 +685,7 @@ public void shouldListAllFunctions() throws Throwable
         //Given/When
         Result res = db.execute( "CALL dbms.functions()" );
 
+
```
Mats-SX: This is the newline that breaks the PR build.

integrationtests/src/test/java/org/neo4j/procedure/UserFunctionIT.java
```diff
@@ -712,14 +713,26 @@ public void shouldListAllFunctions() throws Throwable
...
+        "26 rows" + lineSeparator();
 
         assertThat(res.resultAsString(), equalTo(expected.replaceAll( "'", "\"" )));
     }
 
+    @Test
+    public void shouldCallFunctionWithSameNameAsBuiltIn() throws Throwable
+    {
+        //Given/When
+        Result res = db.execute( "RETURN org.neo4j.procedure.sum([1337, 2.718281828, 3.1415]) AS result");
```
Mats-SX: Missing ending padding whitespace

Mats-SX (commented):
Nice test improvements!

integrationtests/src/test/java/org/neo4j/procedure/UserFunctionIT.java
```diff
@@ -712,14 +713,26 @@ public void shouldListAllFunctions() throws Throwable
         "| 'org.neo4j.procedure.simpleArgument'                | 'org.neo4j.procedure.simpleArgument(someValue :: INTEGER?) :: (INTEGER?)'                                                                                    | ''                      |" + lineSeparator() +
         "| 'org.neo4j.procedure.squareDouble'                  | 'org.neo4j.procedure.squareDouble(someValue :: FLOAT?) :: (FLOAT?)'                                                                                          | ''                      |" + lineSeparator() +
         "| 'org.neo4j.procedure.squareLong'                    | 'org.neo4j.procedure.squareLong(someValue :: INTEGER?) :: (INTEGER?)'                                                                                        | ''                      |" + lineSeparator() +
+        "| 'org.neo4j.procedure.sum'                           | 'org.neo4j.procedure.sum(numbers :: LIST? OF NUMBER?) :: (NUMBER?)'                                                                                          | ''                      |" + lineSeparator() +
```
boggle: This is useful but I have mixed feelings exposing a function with the same name (even in  a namespace) as a built in aggregate function. Especially given that we've said we want to look into how aggregate functions might be called on lists. Could we rename this to make it a bit different from ordinary `sum`?

boggle (changes requested):

integrationtests/src/test/java/org/neo4j/procedure/UserFunctionIT.java
jexp: Actually it's intentional, as users will have namespace functions where the simple function name clashes with built in ones. Which is fine as we have the namespace.

Actually that was the reason for this fix, the clash. And we will not force users to make up awkward names for common operations.

jexp (approved):
Thanks for the fix, looks good to me.

integrationtests/src/test/java/org/neo4j/procedure/UserFunctionIT.java
pontusmelke: @boggle this is not a built-in-function it is a normal user-defined function added in this test class. I think there is just confusion over this very tests it just tests that we can list all user-defined function. Since I added a new one in this update I also had to update the test.

community/neo4j-harness/src/main/java/org/neo4j/harness/internal/AbstractInProcessServerBuilder.java
pontusmelke: hmm...github now longer remove stuff that has been fixed, anyway fixed this.

community/neo4j-harness/src/test/java/org/neo4j/harness/JavaFunctionsTest.java
pontusmelke: fixed

integrationtests/src/test/java/org/neo4j/procedure/UserFunctionIT.java
pontusmelke: thanks! fixed

integrationtests/src/test/java/org/neo4j/procedure/UserFunctionIT.java
pontusmelke: fixed

integrationtests/src/test/java/org/neo4j/procedure/UserFunctionIT.java
pontusmelke: @boggle I clarified it for you :)

boggle: Thank you so much :) Less confusion for the win

boggle (approved):

PR #7974, opened 2016-09-20
Fix: removed log stack trace on authentication failure

fickludd:
Removed server side logging of auth events in Bolt

changelog: Fixed unnecessary stack trace logging of authentication failure

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltProtocolBreachFatality.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltProtocolBreachFatality.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.bolt.v1.runtime;
+
+/**
+ * Indicates that bolt connection has been fatally misused and therefore the server should close the connection.
+ */
+public class BoltProtocolBreachFatality extends BoltConnectionFatality
+{
+    public BoltProtocolBreachFatality( String message )
+    {
+        super( message );
+    }
+}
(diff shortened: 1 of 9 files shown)


community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltProtocolBreachFatality.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.bolt.v1.runtime;
+
+/**
+ * Indicates that bolt connection has been fatally misused and therefore the server should close the connection.
+ */
+public class BoltProtocolBreachFatality extends BoltConnectionFatality
```
technige, line 25: Can we not just stick with a basic `BoltConnectionFatality` for the general case?

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/concurrent/RunnableBoltWorker.java
```diff
@@ -118,7 +129,6 @@ private void execute( Job job ) throws BoltConnectionFatality
     {
         if ( job == SHUTDOWN )
         {
-            machine.close();
```
technige, line 121: Why has this been removed?

fickludd: The close is now called in a finally-clause of the run method

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltProtocolBreachFatality.java
fickludd: In this way, we can differentiate more easily between fatalities where we want to log stack traces, and fatalities where this is not interesting.

technige: I would prefer minimal change and minimal hierarchy. Specifically to use the base exception in all cases unless we know we need to special case it, as you are doing with auth.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/concurrent/RunnableBoltWorker.java
technige: OK, makes sense.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltProtocolBreachFatality.java
fickludd: I definitely agree on minimal change and hierarchy! We do use the new exception to differentiate between the error messages in `RunnableBoltWorker` L103. Would you prefer some other mechanism to achieve this? It's nice to avoid logging stack traces where possible.

technige: I was thinking of something like...

```
catch ( BoltConnectionAuthFatality e )
{
    // handle specific auth-based fatality
}
catch ( BoltConnectionFatality e )
{
    // handle any other kind of connection fatality
}
```

Could that work?

fickludd: Well, that would not longer separate protocol breaches (eg. L554 in `BoltStateMachine`) and other unexpected errors (eg. L352 in `BoltStateMachine`).

PR #7976, opened 2016-09-20
Terminate bolt connections on auth expiration

fickludd:
If the authorization info expires where credentials will be needed to renew it, the server side bolt state machine will send one final failure message and terminate the connection. This functionality uses a new status `Status.Security.AuthorizationExpired`.

Builds on https://github.com/neo4j/neo4j/pull/7974

changelog: Terminate Bolt connection on authorization expiration with specific status code so client apps can differentiate between cache expiry and invalid credentials

### DIFF
--- a/community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/BoltStateMachineTest.java
+++ b/community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/BoltStateMachineTest.java
@@ -22,16 +22,23 @@ package org.neo4j.bolt.v1.runtime;
 
 import org.junit.Test;
 
+import java.util.Collections;
+
 import org.neo4j.bolt.testing.BoltResponseRecorder;
 import org.neo4j.bolt.v1.runtime.spi.BoltResult;
 import org.neo4j.graphdb.TransactionFailureException;
+import org.neo4j.graphdb.security.AuthExpirationException;
 import org.neo4j.kernel.api.exceptions.Status;
 
 import static java.util.Collections.emptyMap;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.fail;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyBoolean;
 import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 import static org.neo4j.bolt.testing.BoltMatchers.canReset;
 import static org.neo4j.bolt.testing.BoltMatchers.failedWithStatus;
 import static org.neo4j.bolt.testing.BoltMatchers.hasNoTransaction;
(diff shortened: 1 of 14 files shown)


community/graphdb-api/src/main/java/org/neo4j/graphdb/security/AuthExpirationException.java
```diff
@@ -0,0 +1,44 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.graphdb.security;
+
+import org.neo4j.kernel.api.exceptions.Status;
+
+/**
+ * Thrown when the database is asked to perform an action that is not authorized based on the AccessMode settings.
+ *
+ * For instance, if attempting to write with READ_ONLY rights.
```
Mats-SX: This comment seems a bit strange. It seems to describe a case where the client is not authorized, not the case where the authorization has _expired_. Are those two concepts the same?

community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/BoltStateMachineTest.java
```diff
@@ -401,4 +411,39 @@ public void actionsDisallowedBeforeInitialized() throws Throwable
         }
     }
 
+    @SuppressWarnings( "unchecked" )
+    @Test
+    public void shouldTerminateOnAuthExpiryDuringREADYRun() throws Throwable
+    {
+        // Given
+        TransactionStateMachine.SPI transactionSPI = mock( TransactionStateMachine.SPI.class );
+        when( transactionSPI.beginTransaction( any() ) ).thenThrow( new AuthExpirationException( "Auth expired!" ) );
```
Mats-SX: I kinda liked the below test call better, with `doThrow(...).when(...)`. Is there a reason it's asymmetrical like this here?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/plugin/TestCacheableAdminAuthPlugin.java
```diff
@@ -0,0 +1,58 @@
...
+
+        if ( principal.equals( "neo4j" ) && credentials.equals( "neo4j" ) )
+        {
+            return CacheableAuthInfo.of( "neo4j", "neo4j".getBytes(),
+                    Collections.singleton( PredefinedRoles.ADMIN ) );
+        }
+        return null;
+    }
+
+    // For testing purposes
```
Mats-SX, line 56: Isn't the whole class a test class?

Mats-SX (commented):
Looks good to me, but I don't feel confident on the Bolt changes.

community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/BoltStateMachineTest.java
henriknyman: It is a minor Mockito quirk. You have to use `doThrow(...).when(...)` syntax when the method is `void` (then it cannot wrap the return value)

community/graphdb-api/src/main/java/org/neo4j/graphdb/security/AuthExpirationException.java
henriknyman: You are right, this looks like a copy-paste comment from the other authorization violation exception.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/plugin/TestCacheableAdminAuthPlugin.java
henriknyman: Well, this is not part of the plugin's supposed "real" behavior :)

community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/BoltStateMachineTest.java
Mats-SX: Aha, thanks for letting me know :)

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/concurrent/RunnableBoltWorker.java
```diff
@@ -94,12 +96,21 @@ public void run()
...
+        {
+            // this is logged in the SecurityLog
+        }
+        catch ( BoltProtocolBreachFatality e )
+        {
+            log.error( "Bolt protocol breach in session '" + machine.key() + "'" );
+        }
+        catch ( Throwable t )
         {
-            log.error( "Worker for session '" + machine.key() + "' crashed: " + e.getMessage(), e );
```
systay: This message seems to have been dropped from the Throwable case. Was that intentional?

fickludd: Yes this was intentional. At the moment the `log` and `userLog` both end up in `debug.log`, and having this message twice is not so helpful.

PR #7990, opened 2016-09-21
Ignore redundant ACK_FAILURE messages

technige:
This addresses a complexity around dealing with multithreading within
drivers by making the server more tolerant of ACK_FAILURE messages that
may not be required. In a failed scenario where the driver sends both a
RESET and an ACK_FAILURE from separate threads (due to a manual
interrupt request) it is hard to guarantee that the ACK_FAILURE will be
processed by the server before the RESET. This change forgives a late
ACK_FAILURE by simply ignoring it.


pontusmelke (approved):
LGTM

zhenlineo:
I close this as it is not needed

Issue #7771, opened 2016-08-22
count() caches off by after some failure


dotdotdotpaul:
I have an instance of 3.0.3 and after monkeying around with various data, and doing "MATCH (v) DETACH DELETE v;" a lot, I've found that I've gotten into an odd situation where the database tells me there are -1 nodes when I do "MATCH (v) RETURN count(v)".  I saved off the entire database path for reference, not sure if it's appropriate to load the whole thing here (it's 2.5MB, even though it's theoretically "empty").

I would get -1 for both MATCH (v) RETURN count(v) and for MATCH (v:Claim) RETURN count(v) but not for MATCH(v:Foobar) RETURN count(v) -- I had been monkeying with Claim nodes, so I must've managed to kill it somehow in the middle of handling those, causing the count caching to decrement prior to actually deleting the node; a subsequent delete then decremented the count again -- although I suppose it's possible that a node was created and the count wasn't incremented on create, so when it was deleted, the count was then off-by-one.

Since it was just a test database, I just deleted the graph.db directory and restarted, and the new database loaded just fine and shows me a zero count.
- Neo4j version: 3.0.3
- Operating system: Mac OS X El Capitan
- API/Driver: Web interface for queries, testing, using neo4j-shell -file to import data.
- Reproduction steps:  Noted above for showing the problem, not sure how I got it into the state in the first place.
- Expected: count() should always accurately reflect the count of nodes.
- Actual: count() appears to be using an independent cached counter which may have been modified in a non-transactional way from the actual node creation/deletion.

Please let me know if you want me to upload the 2.5MB graph.db directory I have, or if I should email to it someone.

spacecowboy:
If you find a clear way of reproducing this (short of killing the database in the middle of a delete process?) we'd like to know

jexp:
Can you zip and upload or send to michael at neo4j.com

Von meinem iPhone gesendet

dotdotdotpaul:
Emailed.

tinwelint:
Would it be possible to include the logs as well, namely `debug.log` and `console.log`?

dotdotdotpaul:
I'll email you what I have, not sure if the date back to when the original event happened.

tinwelint:
Cheers, that would be grand

MishaDemianenko:
The fix was merged long time ago, closing.

PR #7991, opened 2016-09-21
Lock node without relationships during detach delete

MishaDemianenko:
Update node detach delete code to properly acquire node locks in case when node does not have any relationships.
Update test cases to cover this case.
cl: Lock node without relationships during detach delete. Fixes: [#7771](https://github.com/neo4j/neo4j/issues/7771)


tinwelint:
@MishaDemianenko Great find. Actually I think this is an issue in 2.3 as well, consider backporting

jreichhold:
Any ETA on the release of this fix for 3.0.6? Hitting on a regular basis under load

tinwelint:
@jreichhold Look out for 3.0.7, should be out in one or a couple of weeks

PR #7995, opened 2016-09-21
Fixed transient empty file/user repo during reload from disk

fickludd:
During reloading of the auth flat files from disk, a small window existed where the mapping of usernames and role names to users and roles was empty before setting the new reloaded mapping. This is now fixed so that users and roles that are not changed by the reload will always be available.

### DIFF
--- a/community/security/src/main/java/org/neo4j/server/security/auth/FileUserRepository.java
+++ b/community/security/src/main/java/org/neo4j/server/security/auth/FileUserRepository.java
@@ -81,7 +81,7 @@ public class FileUserRepository extends AbstractUserRepository
                 throw new IllegalStateException( "Failed to read authentication file: " + authFile );
             }
 
-            return new ListSnapshot<>( readTime, readUsers );
+            return new ListSnapshot<>( readTime, readUsers, true );
         }
         return null;
     }
@@ -101,7 +101,7 @@ public class FileUserRepository extends AbstractUserRepository
         }
         synchronized ( this )
         {
-            return new ListSnapshot<>( lastLoaded.get(), users.stream().collect( Collectors.toList() ) );
+            return new ListSnapshot<>( lastLoaded.get(), users.stream().collect( Collectors.toList() ), false );
         }
     }
 }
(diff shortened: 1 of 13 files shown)


community/security/src/main/java/org/neo4j/server/security/auth/FileUserRepository.java
```diff
@@ -81,7 +81,7 @@ public void start() throws Throwable
                 throw new IllegalStateException( "Failed to read authentication file: " + authFile );
             }
 
-            return new ListSnapshot<>( readTime, readUsers );
+            return new ListSnapshot<>( readTime, readUsers, true );
```
systay: Booleans are difficult to understand. Could we either add a comment or extract into a helpfully named variable?

community/security/src/main/java/org/neo4j/server/security/auth/FileUserRepository.java
```diff
@@ -101,7 +101,7 @@ protected void persistUsers() throws IOException
         }
         synchronized ( this )
         {
-            return new ListSnapshot<>( lastLoaded.get(), users.stream().collect( Collectors.toList() ) );
+            return new ListSnapshot<>( lastLoaded.get(), users.stream().collect( Collectors.toList() ), false );
```
systay: Same as above

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/InternalFlatFileRealm.java
```diff
@@ -536,7 +563,10 @@ public void suspendUser( String username ) throws IOException, InvalidArgumentsE
             User suspendedUser = user.augment().withFlag( IS_SUSPENDED ).build();
             try
             {
-                userRepository.update( user, suspendedUser );
+                synchronized ( this )
```
systay, line 567: The comment block a couple of lines up says that this method is not synchronized. Maybe that should be updated?

systay (commented):
Looks good from where I am sitting. Just a couple of questions

PR #7998, opened 2016-09-21
Declare that a database is empty when it has no files.

maxsumrall:


enterprise/core-edge/src/main/java/org/neo4j/coreedge/catchup/storecopy/LocalDatabase.java
```diff
@@ -140,9 +138,24 @@ public void delete() throws IOException
...
+    }
+
+    private boolean hasStoreFiles()
+    {
+        for ( StoreType storeType : StoreType.values() )
+        {
+            StoreFile storeFile = storeType.getStoreFile();
+            if(storeFile != null)
+            {
+                boolean exists = new File( storeDir,storeFile.storeFileName() ).exists();
```
apcj: Can you use FileSystemAbstraction.fileExists() here instead of File.exists()? Maybe you actually want to delegate to the StoreFiles type - you have one of those as a field and it already has a FileSystemAbstraction.

PR #7999, opened 2016-09-22
Make `CALL` possible inside `FOREACH`

Mats-SX:
`FOREACH` only allows instances of `UpdateClause` to exist in its
inner scope. `CALL` may have side effects, and may be very useful
in a `FOREACH` construct, so this commit makes it an `UpdateClause`
to support it inside `FOREACH`.

### DIFF
--- a/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/InQueryProcedureCallAcceptanceTest.scala
+++ b/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/InQueryProcedureCallAcceptanceTest.scala
@@ -26,6 +26,18 @@ import org.neo4j.kernel.api.proc.Neo4jTypes
 
 class InQueryProcedureCallAcceptanceTest extends ProcedureCallAcceptanceTest {
 
+  test("should work inside FOREACH") {
+    val query = """WITH [1, 2, 3] AS list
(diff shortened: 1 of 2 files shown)


community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/InQueryProcedureCallAcceptanceTest.scala
```diff
@@ -26,6 +26,18 @@ import org.neo4j.kernel.api.proc.Neo4jTypes
...
+    val query = """WITH [1, 2, 3] AS list
+                  |FOREACH (i IN list |
+                  |  CALL db.labels() YIELD label
+                  |)
+                """.stripMargin
+
+    val result = execute(query)
+
+    result.toList shouldBe empty
+  }
```
davidegrohmann, line 39: Can we actually test a procedure with side effects and check that everything worked as expected?

Mats-SX: Good idea. Turns out it actually wasn't even being planned!

Mats-SX:
This change turned out to be a lot larger than I had anticipated. Not suitable for a maintenance branch.

Mats-SX:
Replaced by #8014

PR #8000, opened 2016-09-22
improve import tool

srbaker:


community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -396,6 +392,22 @@ public static void main( String[] incomingArguments, boolean defaultSettingsSuit
             }
         }
 
+        configuration = importConfiguration( processors, defaultSettingsSuitableForTests, dbConfig, pageSize );
+        input = new CsvInput( nodeData( inputEncoding, nodesFiles ), defaultFormatNodeFileHeader(),
+                relationshipData( inputEncoding, relationshipsFiles ), defaultFormatRelationshipFileHeader(),
+                idType, csvConfiguration( args, defaultSettingsSuitableForTests ), badCollector,
```
spacecowboy: You're using `badCollector` here but you have already closed its output in the finally block above. If any bad lines are detected, your import will crash because it tries to write to a file which had been closed already.

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -396,6 +392,22 @@ public static void main( String[] incomingArguments, boolean defaultSettingsSuit
...
+                configuration.maxNumberOfProcessors() );
+
+        doImport( out, err, storeDir, nodesFiles, relationshipsFiles,
```
spacecowboy: Disregarding the fact that badOutput is closed before it is used (see previous comment), here badOutput is never closed.

Any Streams should be closed when

1) An error occurs, and we are terminating execution
2) Once there is no more need for the stream (we're done)

And it would typically be the responsibility of whoever opened the stream to then close it again.

spacecowboy (changes requested):

community/dbms/src/main/java/org/neo4j/commandline/dbms/CsvImporter.java
```diff
@@ -54,29 +85,52 @@ public static String description()
...
+        pageSize = toIntExact( parseLongWithUnit( args.get( "page-size", String.valueOf( -1 ) ) ) );
+        this.config = config;
     }
 
+
     @Override
     public void doImport() throws IOException
     {
-        ImportTool.main( args );
+        BufferedOutputStream badOutput = new BufferedOutputStream( outsideWorld.errorStream() );
```
spacecowboy: Should flush the bufferedstream you construct here, since it can't be closed (in case errorStream is stdErr)

```
 BufferedOutputStream badOutput = new BufferedOutputStream( outsideWorld.errorStream() );
try {
    // ... all other code
} finally {
    badOutput.flush();
}
```

spacecowboy (changes requested):

spacecowboy (approved):
I approve of this so far

PR #8001, opened 2016-09-22
Lock node without relationships during detach delete

MishaDemianenko:
Update node detach delete code to properly acquire node locks in case when node does not have any relationships.
Update test cases to cover this case.
Fixes: #7771 

<b> N.B. please null merge and use separate #7991 for 3.x versions </B>

changelog [2.3]


tinwelint:
@MishaDemianenko as for the `Compile Tools` failure, please rebase on latest 2.3. As for the `Linux` failure, it failed in reference manual on environment issues. And actually, with this, I'm going to merge this.

PR #8002, opened 2016-09-22
Remove Findbugs plugin

benbc:
Since this wasn't failing the build and no one was inspecting the
output, it wasn't giving any value. It added over ten minutes to a full
build, so we're better off without it.


chrisvest:
Those who miss FindBugs can get the Intellij plug-in, which works quite well.

spacecowboy (approved):
lgtm (as long as build turns green)

PR #8003, opened 2016-09-22
Retry count store lookup in case of concurrent state change.

### LINKED ISSUE #6787: Import via py2neo fails in 2.3.2 with `neostore.counts.db` getting unmapped
I am using py2neo 2.0.8 to import my network into neo4j 2.3.2. The network has ~50 thousand nodes and 3 million relationships. I am running into an intermittent error that `File has been unmapped` regarding `graph.db/neostore.counts.db.a`. The error happens in only ~20% of the imports. [Here is my notebook](https://github.com/dhimmel/integrate/blob/e5ebbec3a7b96cc05b7a01a8d5776d6d94ed192a/neo4j-import.ipynb) where I import 6 networks to 6 separate neo4j servers in parallel, with one import failing. The failure happens midway, so the final database is incomplete. Note that I specified `keep_logical_logs=false` in `neo4j.properties` to reduce space.

I believe the `File has been unmapped` error must come from [this line](https://github.com/neo4j/neo4j/blob/6645e641527d32da1aefa61b80bee8c689a43b36/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPagedFile.java#L371) in `community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPagedFile.java`.
## Error in Python

The error that gets thrown in Python is:

```
File has been unmapped: /home/dhimmels/Documents/serg/rephetio/construct/integrate/neo4j/neo4j-community-2.3.2_hetio-ind_perm-2/data/graph.db/neostore.counts.db.a
```
## `messages.log`

The first 3 lines in `messages.log` below are believed to be okay (making up the vast majority of the log). I believe the process starts to derail on line 4. The shutdown is called by my python script due to an error in the py2neo import. The last line before this shutdown registers in the log is `GC Monitor: Application threads blocked for 222ms.`.

```
2016-03-24 08:54:50.600+0000 INFO  [o.n.c.i.CypherCompiler] Discarded stale query from the query cache: Query(None,SingleQuery(List(Match(false,Pattern(List(EveryPath(NodePattern(Some(Identifier(n)),List(),None,false)))),List(),Some(Where(And(Equals(Property(Identifier(n),PropertyKeyName(identifier)),Parameter(V)),HasLabels(Identifier(n),List(LabelName(Disease))))))), Return(false,ReturnItems(false,List(AliasedReturnItem(Identifier(n),Identifier(n)), AliasedReturnItem(FunctionInvocation(FunctionName(labels),false,Vector(Identifier(n))),Identifier(labels(n))))),None,None,Some(Limit(SignedDecimalIntegerLiteral(1)))))))
2016-03-24 08:54:50.656+0000 INFO  [o.n.c.i.ServerExecutionEngine] Discarded stale query from the query cache: CYPHER 2.3 MATCH (n:MolecularFunction {identifier:{V}}) RETURN n,labels(n) LIMIT 1
2016-03-24 08:54:50.657+0000 INFO  [o.n.c.i.CypherCompiler] Discarded stale query from the query cache: Query(None,SingleQuery(List(Match(false,Pattern(List(EveryPath(NodePattern(Some(Identifier(n)),List(),None,false)))),List(),Some(Where(And(Equals(Property(Identifier(n),PropertyKeyName(identifier)),Parameter(V)),HasLabels(Identifier(n),List(LabelName(MolecularFunction))))))), Return(false,ReturnItems(false,List(AliasedReturnItem(Identifier(n),Identifier(n)), AliasedReturnItem(FunctionInvocation(FunctionName(labels),false,Vector(Identifier(n))),Identifier(labels(n))))),None,None,Some(Limit(SignedDecimalIntegerLiteral(1)))))))
2016-03-24 08:54:51.093+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by scheduler for time threshold [2820]:  Starting check pointing...
2016-03-24 08:54:51.093+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by scheduler for time threshold [2820]:  Starting store flush...
2016-03-24 08:54:51.094+0000 INFO  [o.n.k.i.s.c.CountsTracker] About to rotate counts store at transaction 2820 to [/home/dhimmels/Documents/serg/rephetio/construct/integrate/neo4j/neo4j-community-2.3.2_hetio-ind_perm-2/data/graph.db/neostore.counts.db.b], from [/home/dhimmels/Documents/serg/rephetio/construct/integrate/neo4j/neo4j-community-2.3.2_hetio-ind_perm-2/data/graph.db/neostore.counts.db.a].
2016-03-24 08:54:51.099+0000 INFO  [o.n.k.i.s.c.CountsTracker] Successfully rotated counts store at transaction 2820 to [/home/dhimmels/Documents/serg/rephetio/construct/integrate/neo4j/neo4j-community-2.3.2_hetio-ind_perm-2/data/graph.db/neostore.counts.db.b], from [/home/dhimmels/Documents/serg/rephetio/construct/integrate/neo4j/neo4j-community-2.3.2_hetio-ind_perm-2/data/graph.db/neostore.counts.db.a].
2016-03-24 08:54:51.815+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by scheduler for time threshold [2820]:  Store flush completed
2016-03-24 08:54:51.815+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by scheduler for time threshold [2820]:  Starting appending check point entry into the tx log...
2016-03-24 08:54:52.041+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by scheduler for time threshold [2820]:  Appending check point entry into the tx log completed
2016-03-24 08:54:52.041+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by scheduler for time threshold [2820]:  Check pointing completed
2016-03-24 08:54:52.042+0000 INFO  [o.n.k.i.t.l.p.LogPruningImpl] Log Rotation [1]:  Starting log pruning.
2016-03-24 08:54:52.042+0000 INFO  [o.n.k.i.t.l.p.LogPruningImpl] Log Rotation [1]:  Log pruning complete.
2016-03-24 08:54:52.042+0000 WARN  [o.n.k.i.c.MonitorGc] GC Monitor: Application threads blocked for 222ms.
2016-03-24 08:54:52.058+0000 INFO  [o.n.k.i.f.GraphDatabaseFacade] Shutdown started
2016-03-24 08:54:52.060+0000 INFO  [o.n.k.i.f.CommunityFacadeFactory] Database is now unavailable
2016-03-24 08:54:52.410+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by database shutdown [2820]:  Starting check pointing...
2016-03-24 08:54:52.410+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by database shutdown [2820]:  Starting store flush...
2016-03-24 08:54:53.106+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by database shutdown [2820]:  Store flush completed
2016-03-24 08:54:53.106+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by database shutdown [2820]:  Starting appending check point entry into the tx log...
2016-03-24 08:54:53.553+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by database shutdown [2820]:  Appending check point entry into the tx log completed
2016-03-24 08:54:53.553+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by database shutdown [2820]:  Check pointing completed
2016-03-24 08:54:53.555+0000 INFO  [o.n.k.i.t.l.p.LogPruningImpl] Log Rotation [1]:  Starting log pruning.
2016-03-24 08:54:53.555+0000 INFO  [o.n.k.i.t.l.p.LogPruningImpl] Log Rotation [1]:  Log pruning complete.
2016-03-24 08:54:55.496+0000 INFO  [o.n.k.NeoStoreDataSource] NeoStores closed
2016-03-24 08:54:55.719+0000 INFO  [o.n.k.i.DiagnosticsManager] --- STOPPING diagnostics START ---
2016-03-24 08:54:55.720+0000 INFO  [o.n.k.i.DiagnosticsManager] --- STOPPING diagnostics END ---
```
## Conclusion

`neostore.counts.db.a` is mentioned in `messages.log` and in the python error. I am thinking the log pruning or cycling leads to a failure in a py2neo command/query/transaction. Unfortunately, I'm not exactly sure where to put error handling wrt py2neo to try again in these instances.

I suspect using the default setting for `keep_logical_logs` may resolve this issue. I can always delete logs after shutting down the server, if I need to reduce the database size. I will try this next as well as upgrading to 2.3.3. Not sure if anyone will be able to help, but thought I'd share in case this can lead to a bugfix.

MishaDemianenko:
Count store perform all lookup operation using current state without grabbing any kind of locks.
During any rotation that state will be substituted by new state and closed.
All new readers will use new state automatically.
The problem manifest itself when reader start lookup operation using state that was active when reader starts, but is already rotated and closed by count store rotation.
The most simple solution was chosen: we will retry lookup operation in case of IllegalStateException and state change.

Fixes #6787

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
@@ -89,7 +89,7 @@ public abstract class AbstractKeyValueStore<Key> extends LifecycleAdapter
             ProgressiveState<Key> originalState = this.state;
             try
             {
-                return lookup.value( !this.state.lookup( key, lookup ) );
+                return lookup.value( !originalState.lookup( key, lookup ) );
             }
             catch ( IllegalStateException e )
             {
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
@@ -49,7 +49,7 @@ public abstract class AbstractKeyValueStore<Key> extends LifecycleAdapter
     private final Format format;
     final RotationStrategy rotationStrategy;
     private RotationTimerFactory rotationTimerFactory;
-    private volatile ProgressiveState<Key> state;
+    volatile ProgressiveState<Key> state;
     private DataInitializer<EntryUpdater<Key>> stateInitializer;
     final int keySize;
     final int valueSize;
@@ -84,7 +84,21 @@ public abstract class AbstractKeyValueStore<Key> extends LifecycleAdapter
     protected final <Value> Value lookup( Key key, Reader<Value> reader ) throws IOException
     {
         ValueLookup<Value> lookup = new ValueLookup<>( reader );
-        return lookup.value( !state.lookup( key, lookup ) );
+        while ( true )
+        {
+            ProgressiveState<Key> originalState = this.state;
+            try
+            {
+                return lookup.value( !this.state.lookup( key, lookup ) );
+            }
+            catch ( IllegalStateException e )
+            {
+                if ( originalState == this.state )
+                {
+                    throw e;
+                }
+            }
+        }
     }
 
     /** Introspective feature, not thread safe. */
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStoreTest.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStoreTest.java
@@ -45,6 +45,10 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
 import static org.neo4j.kernel.impl.store.kvstore.DataProvider.EMPTY_DATA_PROVIDER;
 import static org.neo4j.kernel.impl.store.kvstore.Resources.InitialLifecycle.STARTED;
 import static org.neo4j.kernel.impl.store.kvstore.Resources.TestPath.FILE_IN_EXISTING_DIRECTORY;
@@ -78,6 +82,21 @@ public class AbstractKeyValueStoreTest
             return "txId";
         }
     };
+
+    @Test
+    @Resources.Life( STARTED )
+    public void retryLookupOnConcurrentStoreStateChange() throws IOException
+    {
+        Store testStore = resourceManager.managed( createTestStore( TimeUnit.DAYS.toMillis( 2 ) ) );
+        ProgressiveState progressiveState = mock( ProgressiveState.class );
+        CountingErroneousReader countingErroneousReader = new CountingErroneousReader( testStore, progressiveState );
+        testStore.put( "test", "value" );
+
+        testStore.lookup( "test", countingErroneousReader );
+
+        assertEquals( 1, countingErroneousReader.getInvocationCounter() );
+        verify( progressiveState ).lookup( eq( "test" ), any( ValueSink.class ) );
+    }
 
     @Test
     public void shouldStartAndStopStore() throws Exception
@@ -143,7 +162,7 @@ public class AbstractKeyValueStoreTest
         try ( Lifespan life = new Lifespan() )
         {
             Store store = life.add( createTestStore() );
-            assertEquals( 10l, store.headers().get( TX_ID ).longValue() );
+            assertEquals( 10L, store.headers().get( TX_ID ).longValue() );
         }
     }
 
@@ -213,7 +232,7 @@ public class AbstractKeyValueStoreTest
         {
             life.add( store );
 
-            assertEquals( 64l, store.headers().get( TX_ID ).longValue() );
+            assertEquals( 64L, store.headers().get( TX_ID ).longValue() );
         }
     }
 
@@ -316,7 +335,7 @@ public class AbstractKeyValueStoreTest
 
         assertEquals( 2, rotation.rotate() );
 
-        Future<Long> rotationFuture = threading.executeAndAwait( store.rotation, 5l, new Predicate<Thread>()
+        Future<Long> rotationFuture = threading.executeAndAwait( store.rotation, 5L, new Predicate<Thread>()
         {
             @Override
             public boolean test( Thread thread )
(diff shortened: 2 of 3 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
```diff
@@ -84,7 +84,21 @@ public String toString()
     protected final <Value> Value lookup( Key key, Reader<Value> reader ) throws IOException
     {
         ValueLookup<Value> lookup = new ValueLookup<>( reader );
-        return lookup.value( !state.lookup( key, lookup ) );
+        while ( true )
+        {
+            ProgressiveState<Key> originalState = this.state;
+            try
+            {
+                return lookup.value( !this.state.lookup( key, lookup ) );
```
thobe: I'd use `originalState` rather than `this.state` here, to not have to do multiple volatile reads.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStoreTest.java
```diff
@@ -80,6 +84,21 @@ public String toString()
...
+    public void retryLookupOnConcurrentStoreStateChange() throws IOException
+    {
+        Store testStore = resourceManager.managed( createTestStore( TimeUnit.DAYS.toMillis( 2 ) ) );
+        ProgressiveState progressiveState = mock( ProgressiveState.class );
+        CountingErroneousReader countingErroneousReader = new CountingErroneousReader( testStore, progressiveState );
+        testStore.put( "test", "value" );
+
+        testStore.lookup( "test", countingErroneousReader );
+
+        assertEquals( 1, countingErroneousReader.getInvocationCounter() );
```
thobe: Should we not have the next state contain the key we are looking for (at least as a separate test), so that we can observe that the retry actually happens (by observing the invocation counter having a higher value despite the number of invocations to `testStore.lookup` only being 1).

community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
```diff
@@ -84,7 +84,21 @@ public String toString()
...
+            }
+            catch ( IllegalStateException e )
```
thobe, line 94: Are we sure it is `IllegalStateException` that gets thrown in the case of the state being closed?

thobe (commented):
I like the overall structure of the fix, but have a small comment about it.

Personally I would have used a for-loop rather than a while-loop, but that's a matter of taste.

I would have tested this in a very different way - instead of emulating the rotation in the same thread (which is clever), I would have used two threads, and delayed the thread performing the lookup while the other thread performs a rotation. That would be closer to the actual underlying issue we are trying to avoid here, rather than emulating the symptoms of it. One benefit of that is that we would know what exception type we should be catching, and wouldn't have to guess... I'm fine either way though.

MishaDemianenko:
@thobe , @lutovich comments addressed

lutovich:
Changes look good to me. @thobe do you think we can merge this PR?

thobe:
Looks good to me.

PR #7822, opened 2016-08-27
Optimisations and asynchronous apply method for WorkSync

chrisvest:
These changes speed up the `WorkSync` mechanism. Quite noticeable in micro-benchmarks, but probably a negligible difference to the product as a whole. This also fixes a potential GC nepostism issue, where both the prior `tryLock` based locking mechanism, and the batch stack, could end up with prematurely promoted dead objects in the old-gen, that had "live" references to objects in the new-gen.

This also adds a `WorkSync.applyAsync` method, that may be useful for getting #7588 to a working state - specifically, the `TransactionPropagator` might have a use for it.

PR #8005, opened 2016-09-22
Optimisations for WorkSync

chrisvest:
This is the first part of #7822 that is immediately useful without adding too much complexity.

@thobe please take a look.

### DIFF
--- a/community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
+++ b/community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
@@ -45,7 +45,7 @@ import java.util.concurrent.locks.LockSupport;
  *
  * @see Work
  */
-@SuppressWarnings( "unchecked" )
+@SuppressWarnings( {"unchecked", "NumberEquality"} )
 public class WorkSync<Material, W extends Work<Material,W>>
 {
     private final Material material;
@@ -96,7 +96,7 @@ public class WorkSync<Material, W extends Work<Material,W>>
     private WorkUnit<Material,W> enqueueWork( W work )
     {
         WorkUnit<Material,W> unit = new WorkUnit<>( work, Thread.currentThread() );
-        unit.next = stack.getAndSet( unit ); // benign race, see reverse()
+        unit.next = stack.getAndSet( unit ); // benign race, see the batch.next read-loop in combine()
         return unit;
     }
 
@@ -169,7 +169,7 @@ public class WorkSync<Material, W extends Work<Material,W>>
 
     private WorkUnit<Material,W> grabBatch()
     {
-        return reverse( stack.getAndSet( (WorkUnit<Material,W>) stackEnd ) );
+        return stack.getAndSet( (WorkUnit<Material,W>) stackEnd );
     }
 
     private Throwable doSynchronizedWork( WorkUnit<Material,W> batch )
@@ -191,28 +191,6 @@ public class WorkSync<Material, W extends Work<Material,W>>
         return failure;
     }
 
-    private WorkUnit<Material,W> reverse( WorkUnit<Material,W> batch )
-    {
-        WorkUnit<Material,W> result = (WorkUnit<Material,W>) stackEnd;
-        while ( batch != stackEnd )
-        {
-            WorkUnit<Material,W> tmp = batch.next;
-            while ( tmp == null )
-            {
-                // We may see 'null' via race, as work units are put on the
-                // stack before their 'next' pointers are updated. We just spin
-                // until we observe their volatile write to 'next'.
-                // todo Java9: Thread.onSpinWait() ?
-                Thread.yield();
-                tmp = batch.next;
-            }
-            batch.next = result;
-            result = batch;
-            batch = tmp;
-        }
-        return result;
-    }
-
     private W combine( WorkUnit<Material,W> batch )
     {
         W result = null;
@@ -227,7 +205,17 @@ public class WorkSync<Material, W extends Work<Material,W>>
                 result = result.combine( batch.work );
             }
 
-            batch = batch.next;
+            WorkUnit<Material,W> tmp = batch.next;
+            while ( tmp == null )
+            {
+                // We may see 'null' via race, as work units are put on the
+                // stack before their 'next' pointers are updated. We just spin
+                // until we observe their volatile write to 'next'.
+                // todo Java9: Thread.onSpinWait() ?
+                Thread.yield();
+                tmp = batch.next;
+            }
+            batch = tmp;
         }
         return result;
     }
--- a/community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
+++ b/community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
@@ -45,11 +45,12 @@ import java.util.concurrent.locks.LockSupport;
  *
  * @see Work
  */
+@SuppressWarnings( "unchecked" )
 public class WorkSync<Material, W extends Work<Material,W>>
 {
     private final Material material;
     private final AtomicReference<WorkUnit<Material,W>> stack;
-    private final WorkUnit<Material,W> stackEnd;
+    private static final WorkUnit<?,?> stackEnd = new WorkUnit<>( null, null );
     private final AtomicReference<Thread> lock;
 
     /**
@@ -62,8 +63,7 @@ public class WorkSync<Material, W extends Work<Material,W>>
     public WorkSync( Material material )
     {
         this.material = material;
-        this.stackEnd = new WorkUnit<>( null, null, null );
-        this.stack = new AtomicReference<>( stackEnd );
+        this.stack = new AtomicReference<>( (WorkUnit<Material,W>) stackEnd );
         this.lock = new AtomicReference<>();
     }
 
@@ -95,7 +95,7 @@ public class WorkSync<Material, W extends Work<Material,W>>
 
     private WorkUnit<Material,W> enqueueWork( W work )
     {
-        WorkUnit<Material,W> unit = new WorkUnit<>( work, Thread.currentThread(), stackEnd );
+        WorkUnit<Material,W> unit = new WorkUnit<>( work, Thread.currentThread() );
         unit.next = stack.getAndSet( unit ); // benign race, see reverse()
         return unit;
     }
@@ -169,7 +169,7 @@ public class WorkSync<Material, W extends Work<Material,W>>
 
     private WorkUnit<Material,W> grabBatch()
     {
-        return reverse( stack.getAndSet( stackEnd ) );
+        return reverse( stack.getAndSet( (WorkUnit<Material,W>) stackEnd ) );
     }
 
     private Throwable doSynchronizedWork( WorkUnit<Material,W> batch )
@@ -193,7 +193,7 @@ public class WorkSync<Material, W extends Work<Material,W>>
 
     private WorkUnit<Material,W> reverse( WorkUnit<Material,W> batch )
     {
-        WorkUnit<Material,W> result = stackEnd;
+        WorkUnit<Material,W> result = (WorkUnit<Material,W>) stackEnd;
         while ( batch != stackEnd )
         {
             WorkUnit<Material,W> tmp = batch.next;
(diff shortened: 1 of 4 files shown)


chrisvest:
Build was green but has failed to notify Github, which is why there's no green check-mark.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -158,29 +191,127 @@ private W combine( WorkUnit<Material,W> batch )
...
         }
     }
 
-    private static class WorkUnit<Material, W extends Work<Material,W>>
+    private static class WorkUnit<Material, W extends Work<Material,W>> extends AtomicInteger
     {
+        static final int STATE_QUEUED = 0;
+        static final int STATE_PARKED = 1;
+        static final int STATE_DONE = 2;
+        static final int STATE_CANCELLED = 3;
```
thobe: This state is never `set`.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -158,29 +191,127 @@ private W combine( WorkUnit<Material,W> batch )
...
+
+        boolean isDone()
+        {
+            return get() == STATE_DONE;
+        }
+
+        boolean isCancelled()
+        {
+            return get() == STATE_CANCELLED;
+        }
```
thobe: Since `STATE_CANCELLED` is never set, this can never be true.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -158,29 +191,127 @@ private W combine( WorkUnit<Material,W> batch )
             {
                 result = result.combine( batch.work );
             }
-            batch = batch.next;
+
+            do
+            {
+                batch = batch.next;
+            }
+            while ( batch.isCancelled() );
```
thobe: Since `isCancelled()` can never be true, this loop is superfluous.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -103,14 +104,45 @@ public void apply( W work )
...
 
+    private boolean tryLock( int tryCount, WorkUnit<Material,W> unit ) throws InterruptedException
+    {
+        if ( lock.compareAndSet( null, Thread.currentThread() ) )
+        {
+            // Got the lock!
+            return true;
+        }
+
+        // Did not get the lock, spend some time until our work has aither been completed,
```
thobe: spelling: `s/aither/either/`

thobe (commented):
There are some changes I quite liked from the previous PR that didn't make it in here:
- Exception handling for when the jobs fail. It seems like a very sensible thing to do, and throwing ExecutionException for encapsulating that exception also seemed perfectly fine.
- Breaking up `apply()` into `tryDoWork` and `checkFailure` - I thought that was an elegant split.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -58,9 +59,9 @@
     public WorkSync( Material material )
     {
         this.material = material;
-        this.stackEnd = new WorkUnit<>( null );
+        this.stackEnd = new WorkUnit<>( null, null, null );
         this.stack = new AtomicReference<>( stackEnd );
-        this.lock = new ReentrantLock();
+        this.lock = new AtomicReference<Thread>();
```
thobe: explicit type argument not needed.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -71,7 +72,7 @@ public WorkSync( Material material )
     public void apply( W work )
     {
         // Schedule our work on the stack.
-        WorkUnit<Material,W> unit = new WorkUnit<>( work );
+        WorkUnit<Material,W> unit = new WorkUnit<Material,W>( work, Thread.currentThread(), stackEnd );
```
thobe: explicit type arguments not needed.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -158,29 +191,127 @@ private W combine( WorkUnit<Material,W> batch )
...
 
     private void markAsDone( WorkUnit<Material,W> batch )
     {
+        // We will mark the first work unit as the successor - the thread in charge of unparking
+        // the other blocked threads - before we begin marking the units as done.
+        // This way, if the successor wakes up before we finish marking the units as done, it will
+        // know that it should wait and complete its task. In fact, it may trail our done-marking
+        // process and unpark threads in parallel.
+        batch.setAsSuccessor();
+        WorkUnit<Material,W> successor = batch;
```
thobe: Local variable `successor` not used.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -158,29 +191,127 @@ private W combine( WorkUnit<Material,W> batch )
...
+                // If we encounter a unit that isn't marked as done, then we have raced with the
+                // done-marking, and we just spin until we observe the done flag. This is pretty
+                // unlikely, since unparking a thread is pretty expensive on most platforms.
+                // We also spin on null 'next' references, in the unlikely event that our thread
+                // yet observed the complete enqueueing of work units prior to ours.
+                wakeUpAsSuccessor();
+            }
+        }
+
+        private void wakeUpAsSuccessor()
```
thobe: I'd like it if this method had the same type of comments as other places about why it spins waiting for things.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -158,29 +191,127 @@ private W combine( WorkUnit<Material,W> batch )
...
         return result;
     }
 
     private void markAsDone( WorkUnit<Material,W> batch )
     {
+        // We will mark the first work unit as the successor - the thread in charge of unparking
+        // the other blocked threads - before we begin marking the units as done.
+        // This way, if the successor wakes up before we finish marking the units as done, it will
+        // know that it should wait and complete its task. In fact, it may trail our done-marking
+        // process and unpark threads in parallel.
```
thobe: This whole "successor" business was not included in #7822, but strangely this comment was.
Digging through the commits, the "successor" logic was introduced in _Make WorkSync implement its own locking_, and then removed in _Optimisations and asynchronous apply method for WorkSync_.
Perhaps those same optimizations could be applied here, because the "successor" stuff is quite complex (quite possibly overly complex), and at best under documented (more comments please!), at worst unnecessary.

thobe (changes requested):
It seems there was not a clean cut in the commit history of #7822 complexity wise. We traded the unused complexity of cancellation there for the complexity of "successors" here, while not getting the full benefit of not needing cancellation.

chrisvest:
@thobe I believe I've addressed all of your comments.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -19,109 +19,176 @@
...
+ * to overcome the overhead of collecting and batching up the work units.
  * </p>
+ *
  * @see Work
  */
 public class WorkSync<Material, W extends Work<Material,W>>
 {
     private final Material material;
     private final AtomicReference<WorkUnit<Material,W>> stack;
     private final WorkUnit<Material,W> stackEnd;
```
thobe: Remind me again why `stackEnd` cannot be static? Then put that reminder into a source code comment for this field.

If it was static the sharing of it that takes place between `WorkSync` and `WorkUnit` would be simpler.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -19,109 +19,176 @@
...
+            {
+                failure = throwable;
+            }
         }
-
-        markAsDone( batch );
+        return failure;
     }
 
     private WorkUnit<Material,W> reverse( WorkUnit<Material,W> batch )
```
thobe: Given that work combining should be commutative, do we really need to reverse the stack of `WorkUnit`s before combining them?

thobe (commented):
The code is much clearer now. So much clearer that I find myself thinking about higher order things about this code.

thobe (approved):

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
chrisvest: It can be made static at the cost of some unchecked casts to get around the generics.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
chrisvest: Hmm... I guess we don't.

chrisvest:
@thobe I decided to also include your last two suggestions.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -19,130 +19,176 @@
...
- *     where the performance improvements from batching effects is large enough
- *     to overcome the overhead of collecting and batching up the work units.
+ * Obviously this only makes sense for work that a) can be combined, and b)
+ * where the performance improvements from batching effects is large enough
+ * to overcome the overhead of collecting and batching up the work units.
  * </p>
+ *
  * @see Work
  */
+@SuppressWarnings( {"unchecked", "NumberEquality"} )
```
thobe, line 48: Isn't the `NumberEquality` warning being erroneously raised?

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
```diff
@@ -158,7 +204,18 @@ private W combine( WorkUnit<Material,W> batch )
...
+            WorkUnit<Material,W> tmp = batch.next;
+            while ( tmp == null )
+            {
+                // We may see 'null' via race, as work units are put on the
+                // stack before their 'next' pointers are updated. We just spin
+                // until we observe their volatile write to 'next'.
+                // todo Java9: Thread.onSpinWait() ?
+                Thread.yield();
+                tmp = batch.next;
+            }
```
thobe, line 217: Hmm... this is a tricky one. Here is indeed the first place where we are trying to read `WorkUnit.next` now, but it doesn't "feel" like the right place for this spin loop. Accounting for this race would feel more like it "belongs" in the `grabBatch()` method. On the other hand, the longer we can defer the check, the more likely it is we don't have to retry, meaning we should try to defer this look as long as possible - i.e. to here.

Perhaps it would be cleaner in the code to encapsulate this logic in a method, but then it would feel more natural to always use that method for accessing the field, and in `markAsDone()` we don't need to compensate for the race, since `combine()` was invoked before and thus already compensated for it. Although, the overhead of null-checks is really low, so perhaps that's acceptable.

thobe (commented):
Looks good. Just some minor thoughts.

thobe:
The build failure looks unrelated:
```
[09:44:32][Step 5/5] [INFO] --- findbugs-maven-plugin:3.0.3:check (default) @ neo4j-unsafe ---
[09:44:32][Step 5/5] [INFO] BugInstance size is 2
[09:44:32][Step 5/5] [INFO] Error size is 0
[09:44:32][Step 5/5] [INFO] Total bugs: 2
[09:44:32][Step 5/5] [INFO] Exception is caught when Exception is not thrown in org.neo4j.unsafe.impl.internal.dragons.UnsafeUtil.getSharedStringConstructorMethodHandle(MethodHandles$Lookup) [org.neo4j.unsafe.impl.internal.dragons.UnsafeUtil] At UnsafeUtil.java:[line 218] REC_CATCH_EXCEPTION
[09:44:32][Step 5/5] [INFO] org.neo4j.unsafe.impl.internal.dragons.UnsafeUtil$FreeTrace defines compareTo(UnsafeUtil$FreeTrace) and uses Object.equals() [org.neo4j.unsafe.impl.internal.dragons.UnsafeUtil$FreeTrace] At UnsafeUtil.java:[line 370] EQ_COMPARETO_USE_OBJECT_EQUALS
```

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
chrisvest: The message for the warning was relevant, so I think it's just the name of the warning. Looks like any use of reference equality will raise this warning. A bit daft, but I felt that I might as well quiet the last yellow lines while I was there.

community/primitive-collections/src/main/java/org/neo4j/concurrent/WorkSync.java
chrisvest: Having cake vs. eating cake.

PR #8007, opened 2016-09-22
Decoupling Core Topology Lookup from Hazelcast

jimwebber:
We now refresh our local view of topology on topology change,
when asked, we now return that view.


jimwebber:
Owing to some keyboard-git-clownery, this is superseded by #8008

PR #8008, opened 2016-09-22
Decoupling Core Topology Lookup from Hazelcast

jimwebber:
We now refresh our local view of topology on topology change,
when asked, we now return that view.

### DIFF
--- a/enterprise/core-edge/src/main/java/org/neo4j/coreedge/discovery/HazelcastCoreTopologyService.java
+++ b/enterprise/core-edge/src/main/java/org/neo4j/coreedge/discovery/HazelcastCoreTopologyService.java
@@ -94,9 +94,9 @@ class HazelcastCoreTopologyService extends LifecycleAdapter implements CoreTopol
         membershipRegistrationId = hazelcastInstance.getCluster().addMembershipListener( new OurMembershipListener() );
         mapRegistrationId = hazelcastInstance.getMap( EDGE_SERVER_BOLT_ADDRESS_MAP_NAME )
                 .addEntryListener( new OurEntryListener(), true );
-        listenerService.notifyListeners( coreServers() );
         refreshCoreTopology();
         refreshEdgeTopology();
+        listenerService.notifyListeners( coreServers() );
     }
 
     @Override
@@ -189,6 +189,8 @@ class HazelcastCoreTopologyService extends LifecycleAdapter implements CoreTopol
     private void refreshCoreTopology()
     {
         latestCoreTopology = HazelcastClusterTopology.getCoreTopology( hazelcastInstance, log );
+        log.info( "Current core topology is %s", coreServers() );
+        listenerService.notifyListeners( coreServers() );
     }
 
     private void refreshEdgeTopology()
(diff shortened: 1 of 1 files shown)


enterprise/core-edge/src/main/java/org/neo4j/coreedge/discovery/HazelcastCoreTopologyService.java
```diff
@@ -206,15 +212,17 @@ public void memberAdded( MembershipEvent membershipEvent )
         {
             log.info( "Core member added %s", membershipEvent );
             log.info( "Current core topology is %s", coreServers() );
-            listenerService.notifyListeners( coreServers());
+            listenerService.notifyListeners( coreServers() );
```
apcj: Need to call refreshCoreTopology() before logging and notifying listeners, otherwise the log message will have the old topology, which will be very confusing. Maybe move both calls inside refreshCoreTopology() ?

enterprise/core-edge/src/main/java/org/neo4j/coreedge/discovery/HazelcastCoreTopologyService.java
```diff
@@ -206,15 +212,17 @@ public void memberAdded( MembershipEvent membershipEvent )
...
+            refreshCoreTopology();
         }
 
         @Override
         public void memberRemoved( MembershipEvent membershipEvent )
         {
             log.info( "Core member removed %s", membershipEvent );
             log.info( "Current core topology is %s", coreServers() );
-            listenerService.notifyListeners( coreServers());
+            listenerService.notifyListeners( coreServers() );
```
apcj: ditto

enterprise/core-edge/src/main/java/org/neo4j/coreedge/discovery/HazelcastCoreTopologyService.java
```diff
@@ -92,8 +92,10 @@ public void start()
         hazelcastInstance = createHazelcastInstance();
         log.info( "Cluster discovery service started" );
         membershipRegistrationId = hazelcastInstance.getCluster().addMembershipListener( new OurMembershipListener() );
-        mapRegistrationId = hazelcastInstance.getMap( EDGE_SERVER_BOLT_ADDRESS_MAP_NAME ).addEntryListener( new OurEntryListener(), true );
-        listenerService.notifyListeners( coreServers());
+        mapRegistrationId = hazelcastInstance.getMap( EDGE_SERVER_BOLT_ADDRESS_MAP_NAME )
+                .addEntryListener( new OurEntryListener(), true );
+        listenerService.notifyListeners( coreServers() );
```
apcj: First time around coreServers() will return null, won't it?

enterprise/core-edge/src/main/java/org/neo4j/coreedge/discovery/HazelcastCoreTopologyService.java
jimwebber: Okey.

enterprise/core-edge/src/main/java/org/neo4j/coreedge/discovery/HazelcastCoreTopologyService.java
jimwebber: Re-ordered. Ta for that.

enterprise/core-edge/src/main/java/org/neo4j/coreedge/discovery/HazelcastCoreTopologyService.java
jimwebber: Ditto, done.

PR #8012, opened 2016-09-23
Fail to startup as Core or Edge if using custom IO configuration.

maxsumrall:


thobe:
I would like it if the test case for this had a comment specifying _why_ we want to reject custom IO configurations, so that when we get back to this code two years from now, we will remember what the difficulties are that we must overcome for supporting such a feature.

chrisvest:
@maxsumrall I concur with Tobias. It otherwise looks good, though.

enterprise/core-edge/src/test/java/org/neo4j/coreedge/RejectCustomIOTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.neo4j.helpers.collection.MapUtil.stringMap;
+import static org.neo4j.kernel.impl.pagecache.PageSwapperFactoryForTesting.TEST_PAGESWAPPER_NAME;
+import static org.neo4j.kernel.impl.store.format.RecordFormatSelector.defaultFormat;
+
+/**
+ * Core/Edge makes certain assumptions about the file system that it is running on, and certain IO configurations
+ * will not work out of the box. This will reject their usage, giving us the chance to enable it when or if those
+ * other IO configurations are supported. Making this possible likely requires careful work on the kernel layer.
+ * For example, store copy does not work with custom IO configurations.
+ */
```
thobe: How about we instead document _why_ this test case exits.
Something like:

> At the time of writing this test case, certain operations required by Core/Edge does not work with custom IO configurations. This test case aims to ensure that we fail gracefully with a helpful error message if the user tries to combine Core/Edge with custom IO configurations. Specifically, the functionality that Core/Edge needs that custom IO configurations don't support is store copying.

PR #8013, opened 2016-09-23
Fix Ephemeral fs not returning null when listing files for non existing directory

burqen:
It should now behave like DefaultFileSystemAbstraction

### DIFF
--- a/community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
+++ b/community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
@@ -345,7 +345,7 @@ public class EphemeralFileSystemAbstraction implements FileSystemAbstraction
     @Override
     public File[] listFiles( File directory )
     {
-        if ( files.containsKey( directory ) )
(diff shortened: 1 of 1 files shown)


community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
```diff
@@ -345,7 +345,7 @@ public boolean renameFile( File from, File to ) throws IOException
     @Override
     public File[] listFiles( File directory )
     {
-        if ( files.containsKey( directory ) )
+        if ( files.containsKey( directory ) || !directories.contains( directory ) )
```
thobe, line 348: Could we not assume that `files` does not contain directories, and simplify this check to only `!directories.contains( directory )`?

burqen: Turns out it wasn't that easy. Changes causes `TestEphemeralFileChannel` to break. Looks like a more extensive refactoring of how EFSA works is needed to simulate the behaviour of `DefaultFileSystemAbstraction`. I will revisit this in 3.1 and close this PR.

PR #8014, opened 2016-09-23
Support `CALL` inside `FOREACH`

Mats-SX:


community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ProcedureCallSupportAcceptanceTest.scala
```diff
@@ -22,10 +22,81 @@ package org.neo4j.internal.cypher.acceptance
 import java.util
 
 import org.neo4j.cypher._
-import org.neo4j.kernel.api.proc.Neo4jTypes
 
 class ProcedureCallSupportAcceptanceTest extends ProcedureCallAcceptanceTest {
 
+  test("should work inside FOREACH") {
```
systay, line 28: We should add tests for:
- `CALL` inside `FOREACH` inside `FOREACH`
- `CALL ... YIELD` inside `FOREACH`
- A `CALL` that takes input params from the variable introduced in the `FOREACH`
- `NULL` is another thing that usually exposes problems, but I don't know how to get that in here

community/cypher/frontend-3.1/src/main/scala/org/neo4j/cypher/internal/frontend/v3_1/ast/Clause.scala
```diff
@@ -274,11 +274,17 @@ case class Foreach(variable: Variable, expression: Expression, updates: Seq[Clau
...
     expression.semanticCheck(Expression.SemanticContext.Simple) chain
       expression.expectType(CTList(CTAny).covariant) chain
-      updates.filter(!_.isInstanceOf[UpdateClause]).map(c => SemanticError(s"Invalid use of ${c.name} inside FOREACH", c.position)) ifOkChain
+      updates.filterNot(okInsideForeach).map(c => SemanticError(s"Invalid use of ${c.name} inside FOREACH", c.position)) ifOkChain
       withScopedState {
         val possibleInnerTypes: TypeGenerator = expression.types(_).unwrapLists
         variable.declare(possibleInnerTypes) chain updates.semanticCheck
       }
+
+  private def okInsideForeach(clause: Clause) = clause match {
```
systay, line 283: Nice. very readable.

systay (changes requested):

Mats-SX:
CLG decision: not supported.

PR #8016, opened 2016-09-23
Refactoring of intermediate representation

systay:
This PR creates a new module for the intermediate representation classes in Cypher, and moves a lot of code over to this module.


systay:
will re-open against 3.2

Mats-SX:
(Partially) replaced by #8410.

PR #8017, opened 2016-09-23
Disallow deleting final user in neo4j-admin to match server behaviour

craigtaverner:
The server does not allow you to delete yourself. In neo4j-admin the running user is not identified, but the logged in user is assumed to be the admin. The closest role to match server behaviour is to disallow deleting the last user.

### DIFF
--- a/community/security/src/test/java/org/neo4j/commandline/admin/security/DeleteCommandTest.java
+++ b/community/security/src/test/java/org/neo4j/commandline/admin/security/DeleteCommandTest.java
@@ -57,9 +57,29 @@ public class DeleteCommandTest extends UsersCommandTestBase
     }
 
     @Test
+    public void shouldNotDeleteDefaultUser() throws Throwable
+    {
+        // Given - default user
+        createTestUser( "neo4j", "abc" );
+
+        // When
+        try
+        {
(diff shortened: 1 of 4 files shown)


community/security/src/test/java/org/neo4j/commandline/admin/security/DeleteCommandTest.java
```diff
@@ -57,9 +57,29 @@ public void shouldFailToDeleteMissingUser() throws Throwable
...
+        try
+        {
+            UsersCommand usersCommand = new UsersCommand( graphDir.toPath(), confDir.toPath(), out );
+            usersCommand.execute( new String[]{"delete", "neo4j"} );
+        }
+        catch ( CommandFailed e )
+        {
+            // Expect failure message
+            assertThat( e.getMessage(), containsString( "Deleting the only remaining user 'neo4j' is not allowed" ) );
+        }
```
Mats-SX, line 75: Why is this test not using `assertFailedSubCommand`?

craigtaverner: That method uses AdminTool to run the command (and the tests for deleting users with AdminTool are in UsersCommandIT.java), while this class is for unit tests for working directly with the UserCommand class itself, not via the AdminTool.

PR #8018, opened 2016-09-23
Foundational changes for Cypher chunking

petraselmer:
This must be merged simultaneously with https://github.com/neo-technology/neo4j-manual-modeling/pull/86

### DIFF
--- a/manual/cypher/cypher-docs/src/docs/dev/clause.asciidoc
+++ b/manual/cypher/cypher-docs/src/docs/dev/clause.asciidoc
@@ -4,51 +4,58 @@
 [discrete]
 = Read Clauses
 
-[abstract]
---
 These comprise clauses that read data from the database.
---
 
 The flow of data within a Cypher query is an unordered sequence of maps with key-value pairs -- a set of possible bindings between the variables in the query and values derived from the database.
 This set is refined and augmented by subsequent parts of the query.
 
+[options="header"]
+|===
+|Clause                                     |   Description
+|<<query-match,MATCH>>                      |   Specify the patterns to search for in the database.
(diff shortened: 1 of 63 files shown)


manual/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/FunctionsTest.scala
```diff
@@ -122,7 +122,7 @@ class FunctionsTest extends DocumentingTestBase {
 
   @Test def size() {
     testThis(
-      title = "size()",
+      title = "size() - list",
```
jjaderberg: I believe the title controls the file name. With a title of "size() - list" the file name becomes "size---list.asciidoc", which means the include directive for this content fails. If you run a `find manual/cypher/cypher-docs/ -name "*size*"` you should see the generated files. Mutatis mutandis also the other "size" and the two "length" items.

manual/cypher/cypher-docs/src/docs/dev/clause.asciidoc
```diff
@@ -0,0 +1,157 @@
+[[query-clause]]
+= Clauses
+
+[discrete]
```
jjaderberg: I don't think the AsciiDoc processor knows `[discrete]` so it chokes. Can we make do without it?

jjaderberg (changes requested):
Very nice!

Two trifles indicated need fixing:
- no `[discrete]` styles in AsciiDoc
- file names for include directives must agree with what is generated

There is a third after that which at a glance looks like a renegade null character, but I haven't quite figured it out yet.

petraselmer:
@jjaderberg thanks for the comments. I have addressed the two issues you raised.

jjaderberg (approved):

PR #8021, opened 2016-09-25
Plugin realm API cleanup

henriknyman:
- Make AuthToken an explicit type
- More consistent getter syntax


enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginAuthInfo.java
```diff
@@ -19,57 +19,80 @@
...
+    @Override
+    public Collection<String> getRoles()
+    {
+        return roles;
+    }
+
+    @Override
+    public Collection<String> getStringPermissions()
+    {
+        return null;
```
lutovich: It is better to return empty list instead of `null`

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginAuthInfo.java
```diff
@@ -19,57 +19,80 @@
...
+    }
+
+    @Override
+    public Collection<Permission> getObjectPermissions()
+    {
+        return null;
```
lutovich: Also empty list, please

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/api/AuthToken.java
```diff
@@ -19,20 +19,43 @@
...
+    String principal();
+
+    /**
+     * Returns the credentials that verifies the identity.
+     *
+     * <p>Most commonly this is a password.
+     *
+     * @return the credentials that verifies the identity.
+     */
+    char[] credentials();
```
lutovich: Can this method return `String`? `char[]` seems quite inconvenient for callers, though they are mostly tests.

fickludd: char[] is preferred over String for sensitive data as passwords/credentials, since these can be erased after usage. Then one can decrease the number of (unencrypted) passwords floating around in memory in case of a security breach. We know that this type change has not been done across the database, but since this plugin realm will be a public API we need to future proof it.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/spi/AuthenticationPlugin.java
```diff
@@ -63,38 +62,18 @@
      * @see CustomCacheableAuthenticationInfo
      * @see RealmOperations#setAuthenticationCachingEnabled(boolean)
      */
-    AuthenticationInfo authenticate( Map<String,Object> authToken ) throws AuthenticationException;
+    AuthenticationInfo authenticate( AuthToken authToken ) throws AuthenticationException;
 
-    abstract class Adapter implements AuthenticationPlugin
+    abstract class Adapter extends RealmLifecycle.Adapter implements AuthenticationPlugin
```
fickludd, line 67: This really paid off =)

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/api/AuthToken.java
lutovich: Sounds good, I did not know about `char[]` usefulness for sensitive data. Where do we clear those arrays? Or it is a responsibility of API users?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/plugin/api/AuthTokenTest.java
```diff
@@ -30,9 +30,9 @@
...
     {
         // If you change one of these you need to change the other!
-        assertThat( AuthToken.PRINCIPAL, equalTo( org.neo4j.kernel.api.security.AuthToken.PRINCIPAL ) );
-        assertThat( AuthToken.CREDENTIALS, equalTo( org.neo4j.kernel.api.security.AuthToken.CREDENTIALS ) );
-        assertThat( AuthToken.REALM, equalTo( org.neo4j.kernel.api.security.AuthToken.REALM_KEY ) );
-        assertThat( AuthToken.PARAMETERS, equalTo( org.neo4j.kernel.api.security.AuthToken.PARAMETERS ) );
+//        assertThat( AuthToken.PRINCIPAL, equalTo( org.neo4j.kernel.api.security.AuthToken.PRINCIPAL ) );
+//        assertThat( AuthToken.CREDENTIALS, equalTo( org.neo4j.kernel.api.security.AuthToken.CREDENTIALS ) );
+//        assertThat( AuthToken.REALM, equalTo( org.neo4j.kernel.api.security.AuthToken.REALM_KEY ) );
+//        assertThat( AuthToken.PARAMETERS, equalTo( org.neo4j.kernel.api.security.AuthToken.PARAMETERS ) );
```
fickludd: Remove this test?

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/api/AuthToken.java
fickludd: We don't... yet. There is no point unless we change the entire stack all the way from bolt / REST input. But we will eventually, and then we want the public API to support this.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/plugin/api/AuthTokenTest.java
henriknyman: Yes, I missed that

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginAuthInfo.java
henriknyman: Apache Shiro handles null, but sure I can change that.

community/kernel/src/main/java/org/neo4j/kernel/api/security/AuthToken.java
```diff
@@ -48,6 +48,19 @@ static String safeCast( String key, Map<String,Object> authToken ) throws Invali
...
+            throws InvalidAuthTokenException
+    {
+        Object value = authToken.get( key );
+        if ( value != null && !(value instanceof Map) )
+        {
+            throw new InvalidAuthTokenException(
+                    "The value associated with the key `" + key + "` must be a Map but was: " +
+                    value.getClass().getSimpleName() );
+        }
+        return (Map<String,Object>) value;
```
Mats-SX, line 61: There's a null-check in the if-case above, but that seems to be there only to prevent NPE from `.getClass()`. This method may still return `null`. Is that intentional?

Mats-SX: Additionally, I don't see any added test for this method. Can one (or a few) be added?

community/kernel/src/main/java/org/neo4j/kernel/api/security/AuthToken.java
```diff
@@ -48,6 +48,19 @@ static String safeCast( String key, Map<String,Object> authToken ) throws Invali
         return (String) value;
     }
 
+    static Map<String,Object> safeCastMap( String key, Map<String,Object> authToken )
```
Mats-SX, line 51: Is the variable name `authToken` really appropriate here? It seems to me that this method is used to extract a parameter map from a containing map, with no logical connection to auth tokens. I would suggest `map` or `parameterMap`.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/api/AuthToken.java
```diff
@@ -19,20 +19,43 @@
...
-     * Neo4j will act as a pure transport and will not inspect the contents of this map.
+     * Returns an optional custom parameter map if provided by the client.
+     *
+     * <p>This can be used as a vehicle to send arbitrary auth data from a client application
+     * to a server-side auth plugin. Neo4j will act as a pure transport and will not touch the contents of this map.
+     *
+     * @return a custom parameter map if provided by the client, otherwise <tt>null</tt>
      */
-    String PARAMETERS = "parameters";
+    Map<String,Object> parameters();
```
Mats-SX, line 63: Following the comments above on `null`s, why not returning an empty, immutable map if nothing was provided?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/plugin/TestCacheableAuthPlugin.java
```diff
@@ -38,14 +38,14 @@ public String name()
...
     @Override
-    public AuthInfo authenticateAndAuthorize( Map<String,Object> authToken )
+    public AuthInfo authenticateAndAuthorize( AuthToken authToken )
     {
         getAuthInfoCallCount.incrementAndGet();
 
-        String principal = (String) authToken.get( AuthToken.PRINCIPAL );
-        String credentials = (String) authToken.get( AuthToken.CREDENTIALS );
+        String principal = authToken.principal();
+        char[] credentials = authToken.credentials();
```
Mats-SX, line 46: These API changes are lovely!

Mats-SX (commented):
Looks like a nice shape-up. It would be nice to have tests for the `safeCastMap()` method.

community/kernel/src/main/java/org/neo4j/kernel/api/security/AuthToken.java
henriknyman: Added an integration test for custom parameter map

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/api/AuthToken.java
henriknyman: We could but would then also lose the distinction between if the map was provided by the client or if an empty map was provided (not sure if that is useful, but anyways..)

community/kernel/src/main/java/org/neo4j/kernel/api/security/AuthToken.java
henriknyman: It is intentionally returning null since the map may not be present. I guess it could have been called safeCastMapOrNull to clarify this.

PR #8023, opened 2016-09-26
Improved message when store upgrade can't be done

spacecowboy:
Before, this was printed

```
/home/jonas/workspace/neo4j/community/kernel/target/test-data/org.neo4j.kernel.impl.storemigration.UpgradableDatabaseTest$UnsupportedVersions/E1F262ADB34943A5536165FFEF6AEC90/graph-db/neostore' has a store version 'v0.A.4' that we cannot upgrade from
```

After, this is printed instead

```
Unable to upgrade. Cannot open store with version 'v0.A.4'. Current version is `v0.A.7` (Neo4j 3.0.7-SNAPSHOT).
```

This message is thrown when the version is unknown, e.g. it is
jibberish (not sure if that would ever happen), or more likely, it's a
version that is part of a later version of Neo4j which this version has
no knowledge of. Printing the current version (as well as Neo4j version)
makes it as clear as is possible, that he has screwed up his database versions.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
@@ -33,6 +33,7 @@ import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.impl.storemigration.monitoring.MigrationProgressMonitor;
 import org.neo4j.kernel.impl.storemigration.monitoring.MigrationProgressMonitor.Section;
 import org.neo4j.kernel.impl.util.CustomIOConfigValidator;
+import org.neo4j.kernel.internal.Version;
 import org.neo4j.logging.Log;
 import org.neo4j.logging.LogProvider;
 
@@ -318,11 +319,22 @@ public class StoreUpgrader
 
     public static class UnexpectedUpgradingStoreVersionException extends UnableToUpgradeException
     {
-        protected static final String MESSAGE = "'%s' has a store version '%s' that we cannot upgrade from.";
+        protected static final String MESSAGE =
+                "Not possible to upgrade a store with version '%s' to current store version `%s` (Neo4j %s).";
 
-        public UnexpectedUpgradingStoreVersionException( String filename, String actualVersion )
+        public UnexpectedUpgradingStoreVersionException( String fileVersion, String currentVersion )
         {
-            super( String.format( MESSAGE, filename, actualVersion ) );
+            super( String.format( MESSAGE, fileVersion, currentVersion, Version.getNeo4jVersion() ) );
+        }
+    }
+
+    public static class AttemptedDowngradeException extends UnableToUpgradeException
+    {
+        protected static final String MESSAGE = "Downgrading stores are not supported.";
+
+        public AttemptedDowngradeException()
+        {
+            super( MESSAGE );
         }
     }
 
--- a/community/kernel/src/main/java/org/neo4j/kernel/internal/Version.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/internal/Version.java
@@ -36,6 +36,11 @@ public class Version extends Service
         return getKernel().getVersion();
     }
 
+    public static String getNeo4jVersion()
+    {
+        return getKernel().getReleaseVersion();
+    }
+
     private final String artifactId;
     private final String title;
     private final String vendor;
@@ -73,7 +78,7 @@ public class Version extends Service
 
     /**
      * @return a detailed version string, including source control revision information if that is available, suitable
-     *         for internal use, logging and debugging.
+     * for internal use, logging and debugging.
      */
     public final String getVersion()
     {
@@ -86,37 +91,6 @@ public class Version extends Service
     public String getReleaseVersion()
     {
         return releaseVersion;
-    }
-
-    /**
-     * @return the source control revision information, if that is available
-     */
-    public final String getRevision()
-    {
-        StringBuilder result = new StringBuilder( getReleaseVersion() );
-        result.append( ':' ).append( getBranchName() ).append( ':' );
-        String build = getBuildNumber();
-        if ( !(build.startsWith( "${" ) || build.startsWith( "{" )) )
-        {
-            result.append( build ).append( '/' );
-        }
-        result.append( getCommitId() );
-        return result.toString();
-    }
-
-    protected String getBuildNumber()
-    {
-        return "{BuildNumber}";
-    }
-
-    protected String getCommitId()
-    {
-        return "{CommitId}";
-    }
-
-    protected String getBranchName()
-    {
-        return "{BranchName}";
     }
 
     protected Version( String artifactId, String version )
@@ -146,7 +120,7 @@ public class Version extends Service
         );
 
         Matcher matcher = pattern.matcher( fullVersion );
-        if(matcher.matches())
+        if ( matcher.matches() )
         {
             return matcher.group( 1 );
         }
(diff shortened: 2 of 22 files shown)


benbc:
@spacecowboy Can you show the before and after messages?

spacecowboy:
@benbc updated description with before/after

benbc:
All this talk of store versions is pretty unhelpful for the user, I think. Aren't we losing critical (to the user) information because of the way the code is structured: "you are trying to downgrade a database and this isn't supported"?

spacecowboy:
@benbc I agree. Do you want to explicitly mention "downgrade" in the message?

MishaDemianenko:
@spacecowboy @benbc user will see that exception not only when try to downgrade version B to version A, but also when they try to open enterprise store with community version, for example.

spacecowboy:
@MishaDemianenko I thought we dealt with that? This exception seems to think we did: https://github.com/spacecowboy/neo4j/blob/e9e54932141709953431cbe0bc25fcc73c3ccf5c/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java#L334

benbc:
It looks to me like we know what we've got here: https://github.com/spacecowboy/neo4j/blob/e9e54932141709953431cbe0bc25fcc73c3ccf5c/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/UpgradableDatabase.java#L106.

But this throws a spanner in the works: https://github.com/spacecowboy/neo4j/blob/e9e54932141709953431cbe0bc25fcc73c3ccf5c/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/UpgradableDatabase.java#L122

It looks like that IAEx comes from RecordFormatSelector.selectForVersion( result.actualVersion ). Shouldn't we distinguish that case clearly?

benbc:
@spacecowboy I'm not sure if "downgrade" is the most useful vocabulary to use, but I'd definitely like to make it clear that that's what they're (erroneously) trying to do. So, something like "Neo4j 2.3.1 cannot open a database created with a more recent version of Neo4j".

MishaDemianenko:
@spacecowboy @benbc in case if database start using neo4j start up scripts StoreUpgrader should catch that case that's true. But it's not the case when you start tools like consistency checker, etc. on top of the store. We faced already that situation quite often with our testing tools, and as far as those tools will be more and more expose to users over neo4j-admin or others interfaces we should guard them from failure with friendly message :)

spacecowboy:
@MishaDemianenko I think we all agree here. We all want a "user friendly message" in case of failure. And the current message is not particularly user-friendly. Printing the store-versions doesn't help the user understand what's wrong. No one has any idea which version of Neo4j corresponds to `v0.A.7` etc.

Much more friendly messages would be for example:

```
Can't open store created with Neo4j 1.9.5. Only stores of Neo4j 2.3.8 can be upgraded to current version (Neo4j 3.0.5).
```

```
Can't open store created with a newer version of Neo4j. Current version is Neo4j 2.3.8.
```

```
Can't open enterprise store with Neo4j Community Edition.
```

MishaDemianenko:
@spacecowboy absolutely agree i just trying to help and give even more options and use cases :), My point is that we can go even further and give correct message always (in case that you have already covered and in my case of evil tool).
In https://github.com/neo4j/neo4j/blob/3.1/community/kernel/src/main/java/org/neo4j/kernel/impl/store/NeoStores.java#L194 there is a concept of Format family and you can easily see if that format from the same family, higher or lower in same family - so you can treat exceptional case even more correctly then before!

spacecowboy:
@benbc @MishaDemianenko I have updated the PR and the description.

This is, I believe, as good as we can make it for now. The other failures have relatively sensible messages. The Format family etc you linked @MishaDemianenko is only available in 3.1, so I have left that out.

Changing the messages for other cases should be done in a separate PR, if we choose to do so.

community/kernel/src/main/java/org/neo4j/kernel/internal/Version.java
```diff
@@ -88,44 +95,17 @@ public String getReleaseVersion()
...
-
-    protected String getBranchName()
-    {
-        return "{BranchName}";
-    }
-
     protected Version( String artifactId, String version )
     {
         super( artifactId );
+        if ( version == null )
```
MishaDemianenko: `Objects.requireNonNull` can be used for such cases.

community/kernel/src/main/java/org/neo4j/kernel/internal/Version.java
```diff
@@ -170,6 +149,20 @@ public static void main( String[] args )
     }
 
     static final String KERNEL_ARTIFACT_ID = "neo4j-kernel";
-    private static final Version KERNEL_VERSION = new Version( KERNEL_ARTIFACT_ID,
-            Version.class.getPackage().getImplementationVersion() );
+    private static final Version KERNEL_VERSION = new Version( KERNEL_ARTIFACT_ID, getMavenVersion() );
```
MishaDemianenko: What is the reason for this change? why we need to ship with separate properties file instead of using info from manifest?

MishaDemianenko (commented):
One small code comment and question regarding version extraction thingy

community/kernel/src/main/java/org/neo4j/kernel/internal/Version.java
spacecowboy: Because the previous implementation was broken. It was always null. This actually works.

MishaDemianenko: hm...for example:

```
2016-09-19 20:23:54.589+0000 INFO  [o.n.k.i.DiagnosticsManager] Kernel version: 3.0.6-ENG01,5515f38aa3fbea663116cb3349f01d67c7242cca
```

this is actual line from log file where you can see kernel version together with commit printed out - and it's there; also it's sometimes handy to see commit id since some customers from time to time have special engineering releases

spacecowboy: ok, I'll make sure that information is still available. But the existing implementation only works once the whole project is in a jar-form. It does not work when running tests. Having the properties file ALWAYS works.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
```diff
@@ -318,11 +319,12 @@ public UpgradingStoreVersionNotFoundException( String filenameWithoutStoreVersio
 
     public static class UnexpectedUpgradingStoreVersionException extends UnableToUpgradeException
     {
-        protected static final String MESSAGE = "'%s' has a store version '%s' that we cannot upgrade from.";
+        protected static final String MESSAGE =
+                "Unable to upgrade. Cannot open store with version '%s'. Current version is `%s` (Neo4j %s).";
```
MishaDemianenko: I find this message a bit unclear: we say that current version is _something_and then mention neo4j version. What was that something and what i can do with it to be able to fix?

community/kernel/src/main/java/org/neo4j/kernel/internal/Version.java
```diff
@@ -36,6 +36,11 @@ public static String getKernelVersion()
         return getKernel().getVersion();
     }
 
+    public static String getNeo4jVersion()
+    {
+        return getKernel().getReleaseVersion();
```
MishaDemianenko, line 41: there are a lot of call to Version.getKernel().getReleaseVersion(), would propose we replace it with a new one to unify all of them and add some javadoc for this method what it will return (at least reference where it can be found)

MishaDemianenko:
@spacecowboy complete another review round, have couple of comments

spacecowboy:
@MishaDemianenko updated it as we discussed. Hope I got them all.

MishaDemianenko:
@spacecowboy fine with me, feel free to merge on green

PR #8025, opened 2016-09-26
Moved reflective procedure compilation to init phase of lifecycle

fickludd:
Attempting to fix broken IBM JDK 8 builds. Probably related to different class loading combined with the reflective procedure compiler.


systay:
@fickludd Is this PR still valid?

fickludd:
Nope. It was a bug in the IBM JDK 8, which is now fixed.

PR #7180, opened 2016-05-19
Extract docs tooling, kernel docs and lucene-index docs

jjaderberg:

PR #8026, opened 2016-09-26
3.0 shell, UDC and backup docs

benbc:
Creating a new PR because Github won't let me reopen a PR after I've force-pushed to its branch.

Builds on #7180.


manual/shell/src/docs/dev/shell.asciidoc
```diff
@@ -3,363 +3,399 @@
```
jjaderberg, line 3: This can be deleted on fw merge, same for the neo4j-shell manpage below.

manual/shell/src/test/java/org/neo4j/shell/Documenter.java
```diff
@@ -5,17 +5,17 @@
  * This file is part of Neo4j.
```
jjaderberg, line 5: Documenter tool and the ShellDocTest could also be deleted I suppose.

jjaderberg (approved):
Looks good! Almost all of these seem irrelevant in 3.1 but you know better than me how shell and backup is changing.

PR #8031, opened 2016-09-26
Streaming file listings through the page cache

chrisvest:
**What?**
Stream all files in target directory that is reachable by page cache. The file handles in the stream let user delete or rename the underlying files.

**Why?**
The page cache needs to be able to list files and do recursive deletion to be able to imitate behaviour of `BranchDataPolicy` (needed to support HA). Blockdevice does not have any concept of directories so streaming only the files and deleting directories emptied by `rename` and `delete` makes more sense than listing content.


community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
```diff
@@ -206,6 +206,12 @@ public long lastModifiedTime( File file )
         return file.lastModified();
     }
 
+    @Override
+    public void deleteFileOrThrow( File file ) throws IOException
```
lutovich, line 210: There already exists `boolean deleteFile(File)` method. I'd expect it to use `deleteFileOrThrow()` but it does not and has different implementation (uses `FileUtils` instead of `Files`). This looks quite suspicious. Can we make `deleteFile()` use `deleteFileOrThrow()` and return `true` on success and `false` on exception?

community/io/src/main/java/org/neo4j/io/pagecache/FileHandle.java
```diff
@@ -17,15 +17,19 @@
...
 {
-    public CannotRenameMappedFileException( File file )
-    {
-        super( "Cannot rename mapped file: " + file );
-    }
+    String getAbsolutePath();
+
+    File getFile();
+
+    void renameFile( File to, CopyOption... options ) throws IOException;
```
lutovich: Maybe just `rename()`?

community/io/src/main/java/org/neo4j/io/pagecache/impl/FileIsMappedException.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public enum Operation
+    {
+        RENAME( "Cannot rename mapped file" ),
+        DELETE( "Cannot delete mapped file" );
+
+        private final String message;
+
+        Operation( String message )
```
lutovich, line 54: Should this constructor be public? Class is public and `FileIsMappedException` is fully public as well.

community/io/src/main/java/org/neo4j/io/pagecache/impl/SingleFilePageSwapperFactory.java
```diff
@@ -78,6 +84,37 @@ public void renameUnopenedFile( File sourceFile, File targetFile, CopyOption...
     }
 
     @Override
+    public Stream<FileHandle> streamFilesRecursive( File directory ) throws IOException
+    {
+        try
+        {
+            // We grab a snapshot of the file tree to avoid seeing the same file twice or more due to renames.
+            List<File> snapshot = streamFilesRecursiveInner( directory ).collect( toList() );
```
lutovich: Is eagerness required here? Maybe add a comment explaining why

community/io/src/main/java/org/neo4j/io/pagecache/impl/SingleFilePageSwapperFactory.java
```diff
@@ -100,4 +137,54 @@ public long getRequiredBufferAlignment()
     {
         return 1;
     }
+
+    private class WrappingFileHandle implements FileHandle
```
lutovich: Can this class be top level and get its own unit tests?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCache.java
```diff
@@ -405,16 +407,64 @@ public synchronized void renameFile( File sourceFile, File targetFile, CopyOptio
...
+                synchronized ( MuninnPageCache.this )
+                {
+                    throwIfMapped( getFile(), FileIsMappedException.Operation.DELETE );
+                    fileHandle.delete();
+                }
+            }
+        };
+    }
+
+    private void throwIfMapped( File file, FileIsMappedException.Operation operation ) throws IOException
```
lutovich: I'd vote for `assertNotMapped()` but this is not quite important.

community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
```diff
@@ -4862,4 +4875,434 @@ public void fileSizeMustIncreaseInPageIncriments() throws Exception
...
+        File x = existingFile( "sub/x" );
+        File a = file( "a" );
+        File queryForA = new File( new File( sub, ".." ), "a" );
+
+        Stream<FileHandle> stream = pageCache.streamFilesRecursive( queryForA );
+        List<String> filepaths = stream.map( FileHandle::getAbsolutePath ).collect( toList() );
+        assertThat( filepaths, containsInAnyOrder( a.getCanonicalPath() ) ); // note that we don't go into 'sub'
+    }
+
+    @Test( expected = NoSuchFileException.class )
```
lutovich: Can we avoid using `@Test(expected)` and instead use `ExpectedException` rule or `try-catch`? Otherwise it is really hard to understand where exception is expected to be thrown from (especially true for following tests).

lutovich (changes requested):
@chrisvest review round completed. Made some comments.

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
chrisvest: They have slightly different semantics wrt. to deleting files that don't exist, and I didn't want to uproot the use of deleteFile at this point because it has a lot of uses all over the code base. We'd have to examine each to see if they can tolerate the change in semantics.

community/io/src/main/java/org/neo4j/io/pagecache/FileHandle.java
chrisvest: `renameFile` mirrors `FileUtils.renameFile`, but sure we can change it to just `rename`, or to `move` to mirror `java.nio.file.Files.move`.

community/io/src/main/java/org/neo4j/io/pagecache/impl/FileIsMappedException.java
chrisvest: Enum constructors are private. The enum itself is public because it is passed to the `FileIsMappedException` constructor, which is public.

community/io/src/main/java/org/neo4j/io/pagecache/impl/SingleFilePageSwapperFactory.java
chrisvest: Yes, we need to snapshot because file renames could otherwise make files appear more than once in the stream.

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
lutovich: Fair enough, let's not touch `deleteFile()` now.

community/io/src/main/java/org/neo4j/io/pagecache/impl/SingleFilePageSwapperFactory.java
chrisvest: It can be, but it's an implementation detail of the `PageSwapperFactory`. If something needs more unit testing it would be the `PSF`.

community/io/src/main/java/org/neo4j/io/pagecache/FileHandle.java
lutovich: `renameFile()` is just slightly inconsistent with `delete()`. I think both should either have or omit `File` in their names

community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
chrisvest: Yes, I will do that change once the feature work part has settled down. It'll be a lot of modified lines that should go into a dedicated commit.

community/io/src/main/java/org/neo4j/io/pagecache/FileHandle.java
chrisvest: Fair point.

chrisvest:
@lutovich I have not addressed the following:
- [x] the way exceptions are tested for
- [x] the `PSF` testing
- [x] the javadocs on the new `PSF` APIs

This still needs to be done.

chrisvest:
@lutovich All missing things are implemented and all of your comments from your earlier review have been addressed.

Please give it another gander.

chrisvest:
@lutovich The last build has an unrelated test failure. It's just flaky, I think.

lutovich (approved):

PR #8032, opened 2016-09-26
Removing a user should also remove it from all roles

OliviaYtterbrink:
If a user was assigned a role and later removed, it wasn't removed from the roles. This would make it so that if a new user was created with the same name, it would have the roles that the old user had.

This also fixes a bug where the `BasicAuthManager` ignored the requires-password-change flag and always cleared it on `setUserPassword`.

### DIFF
--- a/community/security/src/main/java/org/neo4j/commandline/admin/security/UsersCommand.java
+++ b/community/security/src/main/java/org/neo4j/commandline/admin/security/UsersCommand.java
@@ -19,12 +19,16 @@
  */
 package org.neo4j.commandline.admin.security;
 
-import java.nio.file.FileSystem;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.StreamSupport;
 
 import org.neo4j.commandline.admin.AdminCommand;
 import org.neo4j.commandline.admin.CommandFailed;
@@ -33,17 +37,15 @@ import org.neo4j.commandline.admin.OutsideWorld;
 import org.neo4j.dbms.DatabaseManagementSystemSettings;
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
 import org.neo4j.helpers.Args;
-import org.neo4j.io.fs.DefaultFileSystemAbstraction;
-import org.neo4j.io.fs.DelegateFileSystemAbstraction;
+import org.neo4j.helpers.Service;
+import org.neo4j.kernel.api.security.AuthManager;
 import org.neo4j.kernel.configuration.Config;
+import org.neo4j.kernel.impl.util.JobScheduler;
+import org.neo4j.logging.NullLog;
 import org.neo4j.logging.NullLogProvider;
 import org.neo4j.server.configuration.ConfigLoader;
-import org.neo4j.server.security.auth.BasicAuthManager;
-import org.neo4j.server.security.auth.BasicAuthManagerFactory;
-import org.neo4j.server.security.auth.BasicPasswordPolicy;
-import org.neo4j.server.security.auth.FileUserRepository;
-import org.neo4j.server.security.auth.PasswordPolicy;
-import org.neo4j.time.Clocks;
+import org.neo4j.server.security.auth.UserManager;
+import org.neo4j.server.security.auth.UserManagerSupplier;
 
 public class UsersCommand implements AdminCommand
 {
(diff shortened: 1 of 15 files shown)


benbc:
@OliviaYtterbrink I think I understand that you've added the command overload mechanism in order to allow the Enterprise version of the `users` command to have a different `UserManager`; is that correct?

If so then I think I'd prefer this problem was solved using service location. That allows you to avoid specialization-by-inheritance of the command (in favour of composition) and means that we don't need to complicate the command mechanism.

How do you choose between the Enterprise and Community versions of auth&auth elsewhere in the codebase? Would this not benefit from a service location-based approach as well?

I haven't reviewed the rest of the PR, assuming that you'll get someone from your team to look at it.

OliviaYtterbrink:
@benbc Yes that was indeed the reasoning behind that mechanism. I have no objection against loading the auth manager through service location instead.

benbc:
@OliviaYtterbrink Thank you.

community/security/src/main/java/org/neo4j/commandline/admin/security/UsersCommand.java
```diff
@@ -77,9 +79,9 @@ public AdminCommand create( Path homeDir, Path configDir, OutsideWorld outsideWo
         }
     }
 
-    private final Path homeDir;
-    private final Path configDir;
-    private OutsideWorld outsideWorld;
+    final Path homeDir;
+    final Path configDir;
+    OutsideWorld outsideWorld;
```
Mats-SX, line 84: Why did these become non-private?

community/security/src/main/java/org/neo4j/commandline/admin/security/UsersCommand.java
```diff
@@ -220,22 +221,154 @@ private static Config loadNeo4jConfig( Path homeDir, Path configDir )
...
-        FileUserRepository userRepository =
-                BasicAuthManagerFactory.getUserRepository( config,
-                        NullLogProvider.getInstance(), outsideWorld.fileSystem() );
-        userRepository.start();
-        return userRepository;
+        String configuredKey = config.get( GraphDatabaseSettings.auth_manager );
+        List<AuthManager.Factory> wantedAuthManagerFactories = new ArrayList<>();
+        List<AuthManager.Factory> backupAuthManagerFactories = new ArrayList<>();
+
+        for ( AuthManager.Factory candidate : Service.load( AuthManager.Factory.class ) )
```
Mats-SX, line 231: All of this code is duplicated from the service loading of `AuthManager` that is done in `EditionModule`, right? Since we refactored that in a separate PR, we should recall to also remove this code properly (and load a security module instead).

community/security/src/main/java/org/neo4j/commandline/admin/security/UsersCommand.java
```diff
@@ -220,22 +221,154 @@ private static Config loadNeo4jConfig( Path homeDir, Path configDir )
...
+                outsideWorld.stdOutLine(
+                        String.format( "Attempted to load configured auth manager with keys '%s', but failed",
+                                String.join( ", ", x.getKeys() ) ) );
+            }
+        }
+        return null;
     }
 
-    private BasicAuthManager getAuthManager() throws Throwable
+    public static class NoOpJobScheduler implements JobScheduler
```
Mats-SX, line 289: Could be package-private.

Mats-SX: Could also put a No-Op scheduler as default implementation on the `JobScheduler` interface.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseAuthManagerFactory.java
```diff
@@ -125,7 +124,7 @@ public EnterpriseAuthAndUserManager newInstance( Config config, LogProvider logP
         return orderedActiveRealms;
     }
 
-    private static InternalFlatFileRealm createInternalRealm( Config config, LogProvider logProvider,
+    public static InternalFlatFileRealm createInternalRealm( Config config, LogProvider logProvider,
```
Mats-SX, line 127: Why did this become non-private? IntelliJ suggests `private` is possible.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/InternalFlatFileRealm.java
```diff
@@ -63,7 +63,7 @@
 /**
  * Shiro realm wrapping FileUserRepository and FileRoleRepository
  */
-class InternalFlatFileRealm extends AuthorizingRealm implements RealmLifecycle, EnterpriseUserManager
+public class InternalFlatFileRealm extends AuthorizingRealm implements RealmLifecycle, EnterpriseUserManager
```
Mats-SX, line 66: This access level change also seems to have been unnecessary.

Mats-SX (approved):
I'm going to merge this as-is, and submit a new PR with a commit of suggested changes (to not risk this PR going red).

PR #8033, opened 2016-09-27
Promote the getId() from the Node/Relationship to the PropertyContainer interface

espiegelberg:
The Node and Relationship objects each contain a getter method to get the instance's id. But, when writing Neo4j code, server extensions, and/or GraphAware-based modules, I often find myself writing code that operates on both flavors. I also tend to code against the PropertyContainer interface, which both Node and Relationship implement. This means, sooner or later, I find myself having to do things such as this:

```
long id = -1;
if (propertyContainer instanceof Node) {
    id = ((Node) node).getId();
} else {
    id = ((Relationship) node).getId();
}
```

In fact, I've seen open source code -- as well as code within GraphAware's PropertyContainerUtils -- that performs the same or similar steps as above.

This trivial pull request promotes the getId() off of Node and Relationship and onto the PropertyContainer. By merely relocating the getter to the parent PropertyConatainer interface, the above code is simplified to:

`long id = propertyContainer.getId();`

While small and trivial in nature, to me this is an improvement to the Neo4j codebase that is worth making.

### DIFF
--- a/community/graphdb-api/src/main/java/org/neo4j/graphdb/IdentifiablePropertyContainer.java
+++ b/community/graphdb-api/src/main/java/org/neo4j/graphdb/IdentifiablePropertyContainer.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.graphdb;
+
+public interface IdentifiablePropertyContainer extends PropertyContainer {
+
+    /**
+     * Returns the unique id of this {@link PropertyContainer}. Id's are garbage
+     * collected over time so they are only guaranteed to be unique during a
+     * specific time span: if the {@link PropertyContainer} is deleted, it's
+     * likely that a new {@link PropertyContainer} at some point will get the old
+     * id. <b>Note</b>: this makes {@link PropertyContainer} id's brittle as
+     * public APIs.
+     *
+     * @return The id of this {@link PropertyContainer}
+     */
+     long getId();
+
+}
--- a/community/graphdb-api/src/main/java/org/neo4j/graphdb/IdentifiablePropertyContainer.java
+++ b/community/graphdb-api/src/main/java/org/neo4j/graphdb/IdentifiablePropertyContainer.java
@@ -19,8 +19,12 @@
  */
 package org.neo4j.graphdb;
 
-public interface IdentifiablePropertyContainer extends PropertyContainer {
-
+/**
+ * Provides the ability to obtain the unique id of the implementing
+ * {@link PropertyContainer}.
+ */
+public interface IdentifiablePropertyContainer extends PropertyContainer
+{
     /**
      * Returns the unique id of this {@link PropertyContainer}. Id's are garbage
      * collected over time so they are only guaranteed to be unique during a
--- a/community/graphdb-api/src/main/java/org/neo4j/graphdb/PropertyContainer.java
+++ b/community/graphdb-api/src/main/java/org/neo4j/graphdb/PropertyContainer.java
@@ -145,15 +145,4 @@ public interface PropertyContainer
      * @return all properties on this property container
      */
     Map<String, Object> getAllProperties();
-
-    /**
-     * Returns the unique id of this <code>PropertyContainer</code>. Ids are garbage collected over time
-     * so they are only guaranteed to be unique during a specific time span: if
-     * the  <code>PropertyContainer</code> is deleted, it's likely that a new
-     * <code>PropertyContainer</code> at some point will get the old id. <b>Note</b>: this makes
-     * <code>PropertyContainer</code> ids brittle as public APIs.
-     *
-     * @return The id of this  <code>PropertyContainer</code>.
-     */
-    long getId();
 }
--- a/community/graphdb-api/src/main/java/org/neo4j/graphdb/PropertyContainer.java
+++ b/community/graphdb-api/src/main/java/org/neo4j/graphdb/PropertyContainer.java
@@ -145,4 +145,15 @@ public interface PropertyContainer
      * @return all properties on this property container
      */
     Map<String, Object> getAllProperties();
+
+    /**
+     * Returns the unique id of this <code>PropertyContainer</code>. Ids are garbage collected over time
+     * so they are only guaranteed to be unique during a specific time span: if
+     * the  <code>PropertyContainer</code> is deleted, it's likely that a new
+     * <code>PropertyContainer</code> at some point will get the old id. <b>Note</b>: this makes
+     * <code>PropertyContainer</code> ids brittle as public APIs.
+     *
+     * @return The id of this  <code>PropertyContainer</code>.
+     */
+    long getId();
 }
--- a/community/graphdb-api/src/main/java/org/neo4j/graphdb/Relationship.java
+++ b/community/graphdb-api/src/main/java/org/neo4j/graphdb/Relationship.java
@@ -63,7 +63,7 @@ package org.neo4j.graphdb;
  * {@link #getOtherNode(Node)} and {@link #getNodes()} are guaranteed to always
  * return valid, non-null nodes.
  * <p>
- * A node's id is unique, but note the following: Neo4j reuses its internal ids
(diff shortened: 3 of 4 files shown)


community/graphdb-api/src/main/java/org/neo4j/graphdb/PropertyContainer.java
```diff
@@ -145,4 +145,15 @@
...
+    /**
+     * Returns the unique id of this <code>PropertyContainer</code>. Ids are garbage collected over time
+     * so they are only guaranteed to be unique during a specific time span: if
+     * the  <code>PropertyContainer</code> is deleted, it's likely that a new
+     * <code>PropertyContainer</code> at some point will get the old id. <b>Note</b>: this makes
+     * <code>PropertyContainer</code> ids brittle as public APIs.
+     *
+     * @return The id of this  <code>PropertyContainer</code>.
+     */
+    long getId();
```
chrisvest: Please use `{@code PropertyContainer}` syntax and remove the extra space in line 3.

chrisvest:
@espiegelberg One comment, looks good otherwise. I've also requested rebuilds after adding you to the internal whitelist.

community/graphdb-api/src/main/java/org/neo4j/graphdb/PropertyContainer.java
chrisvest: You can't actually just define a new method on an interface like this, since it breaks compatibility with implementors. In fact in Neo4j itself there are more implementors of `PropertyContainer` than just `Node` and `Relationship` sub-types.

It needs to be a `default` method on the interface, and I think it should throw `UnsupportedOperationException` in its default implementation. This of course also needs to be documented; that it may throw UOE if the entity in question does not have an Id.

chrisvest (changes requested):

community/graphdb-api/src/main/java/org/neo4j/graphdb/PropertyContainer.java
davidegrohmann: Instead of adding a `default` implementation throwing `UnsupportedOperationException`, maybe we can add an extra interface that extends `PropertyContainer` and has the `getId` method. Only `Node`/`Relationship` (and similar) will implement this new interface and we leave untouched the other implementation of `PropertyContainer`.

espiegelberg: While both suggestions are superior to what I proposed, davidegrohmann's accomplishes the same goal while being the least invasive. I will revert my original changes and implement davidegrohmann's suggestion for review.

espiegelberg: The proposed changes have been coded and pushed to my fork. Please let me know if I need to perform any additional steps.

chrisvest: You didn't push your changes to the branch that you issued this pull request from, so they are not showing up in this PR. Also, I think `IdentifiablePropertyContainer` would be a better name.

espiegelberg: I have changed the name to IdentifiablePropertyContainer, committed, and pushed to the PR branch.

community/graphdb-api/src/main/java/org/neo4j/graphdb/IdentifiablePropertyContainer.java
```diff
@@ -0,0 +1,36 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.graphdb;
+
+public interface IdentifiablePropertyContainer extends PropertyContainer {
```
chrisvest: According to our code style, curly braces should be placed on their own lines.

We have code style settings to various IDEs here: https://github.com/neo4j/neo4j.github.com/tree/master/code-style

You can also check for some of our style rules with `mvn checkstyle:check`

chrisvest: There should also be a javadoc comment for the interface itself, even if it will be brief.

community/graphdb-api/src/main/java/org/neo4j/graphdb/Relationship.java
```diff
@@ -67,9 +67,20 @@
...
     /**
+     * Returns the unique id of this relationship. Ids are garbage collected
+     * over time so they are only guaranteed to be unique during a specific time
+     * span: if the relationship is deleted, it's likely that a new relationship
+     * at some point will get the old id. This makes relationship ids brittle as
+     * public APIs.
+     *
+     * @return the id of this node
+     */
+     long getId();
```
chrisvest, line 81: The `@return` clause mentions nodes, but this is a relationship. This comment is also inconsistent with the other two in that it has no `<b>Note</b>:` part.

chrisvest (commented):
The is approaching a good state, but there are still a few nits left to pick. I think the PR can also be squashed into a single commit because the total change is so small, even though we've had a couple of feedback rounds now.

community/graphdb-api/src/main/java/org/neo4j/graphdb/Relationship.java
espiegelberg: Each of the above points of feedback have been coded, committed, and pushed. Please let me know how the PR can be improved in any way.

chrisvest (approved):

espiegelberg:
Thanks @chrisvest and @davidegrohmann for your feedback and work on this pull request.

espiegelberg:
Please let me know if I need to take any additional action. Otherwise I'll patiently look for this PR to be merged when you guys are able to get to it.

thobe:
The general problem with having `getId()` on a more generic interface than `Node` and `Relationship`, and the reason why it is designed the way it is (yes, this is _by design_), is that the possible id-ranges of nodes and relationships overlap with one another, so you have to know if the ID is for a node or for a relationship in order for it to be useful anyhow.

I'm not saying this to stop the change, I am saying this to start a conversation about the benefits versus drawbacks of the proposed change.

community/graphdb-api/src/main/java/org/neo4j/graphdb/IdentifiablePropertyContainer.java
```diff
@@ -0,0 +1,40 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.graphdb;
+
+/**
+ * Provides the ability to obtain the unique id of the implementing
+ * {@link PropertyContainer}.
+ */
+public interface IdentifiablePropertyContainer extends PropertyContainer
```
thobe, line 26: I am not a fan of the name of this interface.

In fact, I don't like the name `PropertyContainer` either for the responsibilities that interface has (it has too many responsibilities for that name) - and I was the one who introduced that interface way-back-when.

I've been feeling that `PropertyContainer` should only deal with the ability to _read_ property values - that would make it more widely applicable. The ability to write properties should be in a sub-interface that I've been thinking could be called `Entity` (because both `Node` and `Relationship` are entities of the graph database). In `Entity` I would find it less controversial to introduce a `getId()` method (even though I still think the problem of you needing to know the entity type before the ID being useable remains).
The main reason I/we haven't done anything towards this before is because the embedded java API has been down-prioritised in favour of Cypher.

espiegelberg: You make some excellent points.

Let’s start with the easiest, being that you are not a fan of the name of this interface. Do you have a suggestion you would prefer?

In terms of the general problem of typically needing to know if the id is for a node or for a relationship in order for the id to be meaningful, that is not only very valid but I imagine particularly much more so valid for the developers working on the actual Neo4j product code base. That said, for developers, such as myself, who are essentially end user consumers of the Neo4j product and are working in unmanaged server extensions or GraphAware modules to create code to create application-level business logic, I find more often than not that I’m able to code to the PropertyContainer interface and do not need to know if the specific implementation is a Node or a Relationship. In fact, by admittedly personal and professional preference, I consider coding to the PropertyContainer interface a good (better?) practice than coding to the implementation, analogous to coding to List rather than coding to ArrayList or Vector (for a trivial example). So while I definitely see the merit, complexities, pros and cons in the design decisions you mention, I think there is valid uses of the getId() being on the PropertyContainer interface for both developers working within Neo4j (such as the core product team) and those working outside it (such as me). I’m undoubtedly heavily influenced by my experience working “downstream” of the core Neo4j developers, where perhaps making use of getId() on the PropertyContainer interface makes more sense for me than it does internally within Neo4j. In sort, perhaps having getId() on the PropertyContainer is more useful to server extension or module developers that it is to developers working on core Neo4j internals.

So, being the contributor of the pull request (PR) to “promote” getId() to the PropertyContainer interface, I unsurprisingly propose that the two topics be split into two separate efforts: that my PR be merged into the core codebase (the short term improvement) and that the conversation of the larger redesign and change in direction take place independently (the long term improvement).

As for the effort of the larger redesign, I am happy to contribute in any way that I can. That said, recognizing that I am a first time contributor of this minor PR and not a member of the core Neo4j product team, I realistically have a fairly narrow scope of possible contribution and would undoubtedly defer to core team members, such as yourself, that have much more familiarity and experience with the code.

PR #8034, opened 2016-09-27
Simplify security realm configuration

henriknyman:
Introduces a new setting `dbms.security.realm` which sets a single
active realm to be used for both authentication and authorization.
Valid alternatives are either one of the built-in realms `native` or `ldap`,
or a plugin realm, e.g. `plugin-<ExamplePlugin>`.

The pre-existing realm settings are now derived from this and considered
an advanced feature and is marked @Internal.
- Added new internal setting `dbms.securtiy.realms`, derived from the single
  realm setting, that can be overridden with a list of active realms. This defines
  the order in which realms are considered in a multiple realm setup.
- The pre-existing `...authentication_enabled` and `...authorization_enabled`
  settings are derived `true` from the single realm setting, but can be overridden
  to toggle authentication and authorization on/off individually to support existing
  multiple realm scenarios.
- Renamed `dbms.security.realms.internal...` to `dbms.security.realms.native...`
  for consistency with the new name of the native realm.

changelog: Simplify security realm configuration to require only the single line dbms.security.realm=name for most configs

### DIFF
--- a/enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/integration/bolt/EnterpriseAuthenticationTestBase.java
+++ b/enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/integration/bolt/EnterpriseAuthenticationTestBase.java
@@ -121,22 +121,7 @@ public abstract class EnterpriseAuthenticationTestBase extends AbstractLdapTestU
 
     protected static Consumer<Map<Setting<?>,String>> ldapOnlyAuthSettings = settings ->
     {
-        settings.put( SecuritySettings.internal_authentication_enabled, "false" );
-        settings.put( SecuritySettings.internal_authorization_enabled, "false" );
-        settings.put( SecuritySettings.ldap_authentication_enabled, "true" );
-        settings.put( SecuritySettings.ldap_authorization_enabled, "true" );
-        settings.put( SecuritySettings.plugin_authentication_enabled, "false" );
-        settings.put( SecuritySettings.plugin_authorization_enabled, "false" );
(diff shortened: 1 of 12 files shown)


enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/integration/bolt/EnterpriseAuthenticationTestBase.java
```diff
@@ -121,8 +121,9 @@ protected void reconnect() throws Exception
 
     protected static Consumer<Map<Setting<?>,String>> ldapOnlyAuthSettings = settings ->
     {
-        settings.put( SecuritySettings.internal_authentication_enabled, "false" );
-        settings.put( SecuritySettings.internal_authorization_enabled, "false" );
+        settings.put( SecuritySettings.active_realm, SecuritySettings.LDAP_REALM_NAME );
+        settings.put( SecuritySettings.native_authentication_enabled, "false" );
```
craigtaverner: Do we need the internal settings when they simply repeat the implied value from the line above?

craigtaverner:
I like the solution. Elegant and only minimal code changes to achieve a much simpler config.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/integration/bolt/EnterpriseAuthenticationTestBase.java
henriknyman: Removed

PR #8040, opened 2016-09-27
Fixed missing logging of password change via init over Bolt

fickludd:


community/server/src/main/java/org/neo4j/server/rest/dbms/UserService.java
```diff
@@ -136,4 +146,16 @@ public Response setPassword( @PathParam("username") String username, @Context Ht
...
+    {
+        if ( principal instanceof DelegatingPrincipal )
+        {
+            AccessMode mode = ((DelegatingPrincipal) principal).getAccessMode();
+            if ( mode instanceof AuthSubject )
+            {
+                return (AuthSubject) mode;
+            }
+        }
+        return null;
```
henriknyman, line 159: There was already a static AuthorizedRequestWrapper.getAccessModeFromUserPrincipal(). It was hidden in there to hide as much of this dirtiness as possible from the outside world.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/StandardEnterpriseAuthSubject.java
```diff
@@ -71,7 +75,17 @@ public AuthenticationResult getAuthenticationResult()
     public void setPassword( String password, boolean requirePasswordChange )
             throws IOException, InvalidArgumentsException
     {
-        getUserManager().setUserPassword( (String) shiroSubject.getPrincipal(), password, requirePasswordChange );
+        try
+        {
+            getUserManager().setUserPassword( (String) shiroSubject.getPrincipal(), password, requirePasswordChange );
+            securityLog.info( this, "changed password%s",
+                    requirePasswordChange ? ", with password change required" : "" );
```
henriknyman, line 82: I still believe in having the logging inside the user manager will simplify our lives..

OliviaYtterbrink: This will be done in a separate PR

community/server/src/main/java/org/neo4j/server/rest/dbms/UserService.java
fickludd: We plan to fix this in separate PR

PR #8041, opened 2016-09-27
Remove or restore bunch of ignored kernel tests

MishaDemianenko:


community/neo4j/src/test/java/upgrade/TestMigrateToDenseNodeSupport.java
```diff
@@ -126,35 +124,6 @@ public void before() throws Exception
     }
 
     @Test
-    @Ignore( "Used for creating the dataset, using the previous store version" )
```
tinwelint, line 129: Perhaps we can delete this test class completely? Doesn't seem relevant anymore.

tinwelint: Meh, some future cleanup will do it :)

PR #8049, opened 2016-09-28
Mark timed out transaction as terminated.

MishaDemianenko:
Mark transaction killed by TimeoutGuard as terminated.
Add new TransactionTimeOut status for that case.


community/kernel/src/main/java/org/neo4j/kernel/guard/GuardException.java
```diff
@@ -19,11 +19,14 @@
...
+import org.neo4j.kernel.api.exceptions.Status;
+
+public abstract class GuardException extends RuntimeException implements Status.HasStatus
 {
 
     protected GuardException( final String message )
     {
         super( message );
     }
+
```
burqen, line 31: Formatting

PR #8050, opened 2016-09-28
Restore compatibility within BOLT protocol v1

cleishm:
This PR restores compatibility with previous releases of the BOLT protocol v1.

Specifically, it allows a FAILURE of INIT to be acknowledged by the client, thus clearing the failure state. Failure of INIT is a common error path, easily reached by a user error, and thus maintaining correct protocol behaviour is important and necessary for compatibility.


systay:
Let's re-open this if/when we decide to accept the changes. No need to have long opened PRs that are not acted on.

PR #8051, opened 2016-09-29
3.0 extract server docs

benbc:


manual/server-docs/src/test/java/org/neo4j/server/rest/PropertyExistenceConstraintsDocIT.java
```diff
@@ -0,0 +1,344 @@
...
+    private static NeoServer server;
+
+    @Override
+    public GraphDatabaseService graphdb()
+    {
+        return server.getDatabase().getGraph();
+    }
+
+    @BeforeClass
+    public static void initServer() throws Exception
```
jjaderberg: I believe `initServer()` is idempotent in the community tests, but here it tries to raise a new server and fails on port bind. The test requires an enterprise server so must raise it. Mixing enterprise and community test means either using different ports to allow multiple servers or controlling test order to first all of one edition then all of the other. It may be easier to just put the enterprise tests in their own module.

jjaderberg (commented):
Looks good!

`PropertyExistenceConstraintsDocIT` fails per other comment.

Could you also move the manpages for `neo4j` and `neo4j-admin`?

benbc:
@jjaderberg I don't think that I can just move the manpages, can I? We build them into the Debian package.

benbc:
I see you've already moved the other manpages. So I'll move those ones as well and we can figure this out when we move the manual stuff to another repo.

jjaderberg:
I actually thought we had moved to using static manpages to break the packaging=>docs dependency. But yes, it should be fine to move them and we can decide later–maybe turn `neo4j-manual` into a tiny `neo4j-manual-pages` for packaging needs.

PR #8053, opened 2016-09-29
Fix bug around IN predicate in Cypher

systay:
Handles collection with single null element in it

### DIFF
--- a/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/WhereAcceptanceTest.scala
+++ b/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/WhereAcceptanceTest.scala
@@ -39,6 +39,14 @@ class WhereAcceptanceTest extends ExecutionEngineFunSuite with NewPlannerTestSup
     a[IncomparableValuesException] should be thrownBy executeWithAllPlannersAndCompatibilityMode(query)
   }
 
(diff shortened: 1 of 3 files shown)


Mats-SX (changes requested):
I'd be happy to see an acceptance test for this, given my assumption that it is possible to see this bug from a specific query.

community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/WhereAcceptanceTest.scala
```diff
@@ -39,6 +39,14 @@ class WhereAcceptanceTest extends ExecutionEngineFunSuite with NewPlannerTestSup
     a[IncomparableValuesException] should be thrownBy executeWithAllPlannersAndCompatibilityMode(query)
   }
 
+  test("check for contents of collection that contains only a single null") {
+    createNode()
+
+    val result = executeWithAllPlanners("MATCH (a) WHERE 42 IN [a.prop] RETURN *", "param" -> null)
```
Mats-SX, line 45: I don't understand how this query uses the `param`. Is it necessary?

Mats-SX (approved):

Mats-SX:
The test failure is an unrelated, flaky HA test. All modules still reported success. Merging this.

PR #8054, opened 2016-09-29
Correct location of bad.log and debug.log in import tool

tinwelint:
by using the same mechanisms to figure that out as neo4j
does otherwise.


chrisvest:
@tinwelint Reminder that there are test failures that look relevant.

PR #8055, opened 2016-09-29
Improved list handling

pontusmelke:
- Improved handling of lists coming from procedures and user-defined functions
- Make sure ranges we create have performant random access
- Use `toIndexedSeq` everywhere instead of `toSeq` and `toVector`

Queries like 

```
WITH ["Jane","Dan","Adam","Sue","Tina"] AS names, 50000 AS len
UNWIND range(1,len) as id
WITH {name:names[toInt(rand()*5)], id:id} as u,len
WITH len, collect(u) AS users
UNWIND range(1,len*10) as _
WITH users[toInt(rand()*len)] AS u1, users[toInt(rand()*len)] AS u2
RETURN count(*)
```

goes from running for more than 15 minutes to finishing in 3s (on my machine).


systay:
Hmm... My initial reaction is that lists returned from procs or udfs are mutable, and we need to copy them before allowing them into the runtime. Am I wrong in thinking that? /cc @thobe @boggle

thobe:
In some cases I think we can trust values returned from procedures/functions, and then the copy is superfluous.

When returning a list/map from cypher to the user, the copy does seem unnecessary, at least if we know that the runtime is not going to need the contents of it anymore.

pontusmelke:
@systay From the runtime point of view the map/list is still immutable and I don't see a way of modifying the objects after being returned from the procedure. Or maybe if the procedure keeps a static object that it returns that might happen. I don't mind changing the code so that it does the copying. I think we will still gain a lot by not doing  `ArrayList -> scala.List -> SeqWrapper`.

boggle:
We had a similar discussion around procedures where in theory the returned stream could perform side effects on stream contents returned earlier. The answer we arrived at there was to expect reasonable behavior from procedures, mostly for performance reasons. So procedure author get to shoot the Cypher runtime in the foot if they really want to. I don't think that's a big deal.

Here it is the same: a returned list could be still mutated whenever a certain operation is called on it and this could be used to mess things up. In reality though, who's going to do that, even by accident? Is it really necessary to make defensive copies? I'm not convinced. It might be worth being able to enable that for less trusted procedure code but that would require some kind of configuration mechanism.

pontusmelke:
@systay @thobe @boggle I updated it to make a defensive copy and removed the fix for maps for now. I think there still may be an opportunity to get rid of copying altogether but lets postpone that work until later.

There is still a lot of gain to be had by just reducing the number of copy operations to only one as is currently done in this PR. For example given

```
@UserFunction
        public List<Long> list()
        {
            ArrayList<Long> longs = new ArrayList<>( 1000_000 );
            for ( long i = 0L; i < 1000_000L; i++ )
            {
                longs.add( i );
            }

            return longs;
        }
```

and the query

```
WITH org.neo4j.procedure.list() AS list RETURN list[3] + list[9999] + list[5000]  + list[99998] AS result
```

goes from `~300ms` to `~25ms` (and something like `18ms` when no defensive copy)

jexp:
@pontusmelke I agree with the others, that defensive copying is not needed, in Cypher itself there are no mutating operations, the endpoints just return the content to to the driver which doesn't mutate and when the collection is passed back to the next procedure it is wrapped (imho but not copied) in a Wrapper anyway which prohibits mutation.

So from that perspective we can remove the defensive copying for procedure results (and if there is any for procedure inputs too). For functions as well, as those are even more meant to be read-only transformations.

pontusmelke:
@jexp The only way I can see is in the case where a procedure or function maintains a static instance.

``` java
...
private final static List<Long> numbers = new ArrayList<>();
...
@UserFunction
List<Long> list(long number)
{
    number.add(number);
    return numbers;
}
...
```

I think that is the case @systay is worried about.

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/commands/expressions/IndexedInclusiveLongRange.scala
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher.internal.compiler.v3_1.commands.expressions
+
+import scala.collection.immutable
+
+/**
+  * An inclusive long range that is also indexable as long as the total length of the range is less that `Int.MaxValue`
```
Mats-SX: `that` -> `than`

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/commands/expressions/IndexedInclusiveLongRange.scala
```diff
@@ -0,0 +1,57 @@
...
+  * @param start beginning of range (inclusive)
+  * @param end end of range (inclusive)
+  * @param step step between element of range
```
Mats-SX: element`s`

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/commands/expressions/MathFunction.scala
```diff
@@ -254,35 +278,26 @@ case class RandFunction() extends Expression {
...
+
   override def apply(ctx: ExecutionContext)(implicit state: QueryState): Any = {
     val stepVal = asLong(step(ctx))
     if (stepVal == 0L)
       throw new InvalidArgumentException("step argument to range() cannot be zero")
 
     val startVal = asLong(start(ctx))
     val inclusiveEndVal = asLong(end(ctx))
-    val check: (Long, Long) => Boolean = if (stepVal.signum > 0) _ <= _ else _ >= _
+    //val check: (Long, Long) => Boolean = if (stepVal.signum > 0) _ <= _ else _ >= _
```
Mats-SX: This should probably be removed.

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/commands/expressions/MathFunction.scala
```diff
@@ -254,35 +278,26 @@ case class RandFunction() extends Expression {
...
-    }
+
+    IndexedInclusiveLongRange(startVal, inclusiveEndVal, stepVal)
   }
 
   override def arguments = Seq(start, end, step)
 
-  override def rewrite(f: (Expression) => Expression) = f(RangeFunction(start.rewrite(f), end.rewrite(f), step.rewrite(f)))
+  override def rewrite(f: (Expression) => Expression) = f(
+    RangeFunction(start.rewrite(f), end.rewrite(f), step.rewrite(f)))
```
Mats-SX: That's a fairly strange line break change.

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/helpers/JavaListWrapper.scala
```diff
@@ -0,0 +1,43 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher.internal.compiler.v3_1.helpers
+
+import scala.collection.immutable.IndexedSeq
+
+/**
+  * Simple wrapper for a java.util.List which preserves the original list
+  * while lazily converts to scala values if needed.
```
Mats-SX: `converts` -> `converting`

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/helpers/RuntimeScalaValueConverter.scala
```diff
@@ -41,10 +42,12 @@ class RuntimeScalaValueConverter(skip: Any => Boolean, converter: Any => Any) {
...
     case anything if skip(anything) => anything
-    case javaMap: JavaMap[_, _] => immutableMapValues(javaMap.asScala, asDeepScalaValue): immutable.Map[_, _]
-    case javaIterable: JavaIterable[_] => javaIterable.asScala.map(asDeepScalaValue).toList: List[_]
+    case javaMap: JavaMap[_, _] => immutableMapValues(javaMap.asScala, asDeepScalaValue)
+    case javaList: java.util.LinkedList[_] => copyJavaList(javaList,() => new util.LinkedList[Any]())
+    case javaList: java.util.List[_] => copyJavaList(javaList,() => new util.ArrayList[Any](javaList.size()))
+    case javaIterable: JavaIterable[_] => javaIterable.asScala.map(asDeepScalaValue).toIndexedSeq: Seq[_]
     case map: collection.Map[_, _] => immutableMapValues(map, asDeepScalaValue): immutable.Map[_, _]
-    case traversable: TraversableOnce[_] => traversable.map(asDeepScalaValue).toList: List[_]
+    case traversable: TraversableOnce[_] => traversable.map(asDeepScalaValue).toList: Seq[_]
```
Mats-SX: Not `toIndexedSeq` and `IndexedSeq[_]` here?

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/helpers/RuntimeScalaValueConverter.scala
```diff
@@ -41,10 +42,12 @@ class RuntimeScalaValueConverter(skip: Any => Boolean, converter: Any => Any) {
 
   def asDeepScalaValue(value: Any): Any = value match {
     case anything if skip(anything) => anything
-    case javaMap: JavaMap[_, _] => immutableMapValues(javaMap.asScala, asDeepScalaValue): immutable.Map[_, _]
-    case javaIterable: JavaIterable[_] => javaIterable.asScala.map(asDeepScalaValue).toList: List[_]
+    case javaMap: JavaMap[_, _] => immutableMapValues(javaMap.asScala, asDeepScalaValue)
+    case javaList: java.util.LinkedList[_] => copyJavaList(javaList,() => new util.LinkedList[Any]())
+    case javaList: java.util.List[_] => copyJavaList(javaList,() => new util.ArrayList[Any](javaList.size()))
+    case javaIterable: JavaIterable[_] => javaIterable.asScala.map(asDeepScalaValue).toIndexedSeq: Seq[_]
```
Mats-SX: Not `IndexedSeq` here?

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/mutation/CreateRelationship.scala
```diff
@@ -68,7 +68,7 @@ extends UpdateAction
   def localEffects(symbols: SymbolTable) = props.values.foldLeft(Effects(CreatesRelationship(typ)))(_ ++ _.effects(symbols))
 
   override def children =
-    props.values.toSeq ++ Seq(from.node, to.node) ++
+    props.values.toIndexedSeq ++ Seq(from.node, to.node) ++
```
Mats-SX: Will this make `children` an `IndexedSeq`? Perhaps the literal `Seq` that gets appended needs to be a `Vector`?

community/cypher/cypher-compiler-3.1/src/test/scala/org/neo4j/cypher/internal/compiler/v3_1/commands/expressions/IndexedInclusiveLongRangeTest.scala
```diff
@@ -0,0 +1,100 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher.internal.compiler.v3_1.commands.expressions
+
+import org.neo4j.cypher.internal.frontend.v3_1.test_helpers.CypherFunSuite
+
+import scala.util.Random
+
+//TODO test negative ranges and indexes
```
Mats-SX: This seems to still be `TODO`. Should this PR have `NOT READY FOR MERGE`?

community/cypher/cypher/src/test/scala/org/neo4j/cypher/ExecutionEngineTest.scala
```diff
@@ -732,8 +732,7 @@ order by a.COL1""")
       val indexDefinitions = graph.schema().getIndexes(Label.label(labelName)).asScala.toSet
       indexDefinitions should have size 1
 
-      val actual = indexDefinitions.head.getPropertyKeys.asScala.toSeq
-      propertyKeys should equal(actual)
```
Mats-SX, line 736: Why is this assertion removed without replacement?

community/cypher/frontend-3.1/src/main/scala/org/neo4j/cypher/internal/frontend/v3_1/ast/CaseExpression.scala
```diff
@@ -37,3 +39,9 @@ case class CaseExpression(expression: Option[Expression], alternatives: Seq[(Exp
     } chain this.specifyType(possibleTypes)
   }
 }
+
+object CaseExpression {
+  def apply(expression: Option[Expression], alternatives: List[(Expression, Expression)], default: Option[Expression])(position: InputPosition):CaseExpression =
```
Mats-SX, line 44: Isn't it preferable to take an `IndexedSeq` as parameter type instead of `List`, and make callers convert? Or did you do it this way to not change an abundance of callers? I'm thinking there's some calling code that obliviously creates a `List` and sends it here, only to have it converted to `IndexedSeq`, where creating the latter immediately would've been better (it may be a micro-optimisation).

Mats-SX (commented):
Nice improvements!

community/cypher/frontend-3.1/src/main/scala/org/neo4j/cypher/internal/frontend/v3_1/ast/CaseExpression.scala
pontusmelke: It is but parboiled creates Lists so I need this extra constructor to turn it into an IndexedSeq

pontusmelke:
@Mats-SX Thanks for the feedback, I think I addressed all your comments

community/cypher/cypher/src/test/scala/org/neo4j/cypher/ExecutionEngineTest.scala
pontusmelke: mistake, added it back

community/cypher/cypher-compiler-3.1/src/test/scala/org/neo4j/cypher/internal/compiler/v3_1/commands/expressions/IndexedInclusiveLongRangeTest.scala
pontusmelke: I added tests but forgot to remove the TODO, fixed now.

jexp:
@pontusmelke @Mats-SX any chance to backport the change relevant to index/slice of lists to 3.0 ?
It would be very beneficial for everyone.

PR #8058, opened 2016-09-30
Removed DbmsOperations.Factory, which is no longer needed

fickludd:


enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/BuiltInProcedures.java
```diff
@@ -81,8 +82,6 @@
     public GraphDatabaseAPI graph;
 
     @Context
-    public KernelTransaction tx;
-    @Context
```
Mats-SX, line 85: Nice!

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/BuiltInProcedures.java
```diff
@@ -99,7 +98,13 @@ public void setTXMetaData( @Name( value = "data" ) Map<String,Object> data )
...
         }
 
-        tx.acquireStatement().queryRegistration().setMetaData( data );
+        getCurrentTx().acquireStatement().queryRegistration().setMetaData( data );
+    }
+
+    private KernelTransaction getCurrentTx()
+    {
+        return graph.getDependencyResolver().resolveDependency( ThreadToStatementContextBridge.class )
+                .getKernelTransactionBoundToThisThread( true );
```
Mats-SX, line 107: Is there a specific reason as to why this is preferred to injecting via `@Context`? I'm curious.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/AuthProcedures.java
```diff
@@ -427,6 +426,12 @@ private BoltConnectionTracker getBoltConnectionTracker()
         return graph.getDependencyResolver().resolveDependency( BoltConnectionTracker.class );
     }
 
+    private KernelTransaction getCurrentTx()
+    {
+        return graph.getDependencyResolver().resolveDependency( ThreadToStatementContextBridge.class )
+                .getKernelTransactionBoundToThisThread( true );
+    }
```
Mats-SX, line 433: Duplicated code from above. Perhaps there's a good class to add this as a `static` helper?

Mats-SX (approved):
Looks good.

PR #8060, opened 2016-09-30
Fix NPE in dependecy resolver when datasource is null

davidegrohmann:
In some cases if you ask for resolving a dependency from the Neo store
data source when the latter is not available a NPE is thrown.  This
change will fix that behaviour by skipping the unavailable data source
from the list of possible dependency providers.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/PlatformModule.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/PlatformModule.java
@@ -24,9 +24,7 @@ import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Supplier;
 
-import org.neo4j.graphdb.DependencyResolver;
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
 import org.neo4j.graphdb.security.URLAccessRule;
 import org.neo4j.helpers.collection.Iterables;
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/factory/PlatformModule.java
```diff
@@ -115,7 +116,12 @@ public PlatformModule( File providedStoreDir, Map<String, String> params, Databa
             @Override
             public DependencyResolver get()
             {
-                return dataSourceManager.getDataSource().getDependencyResolver();
+                NeoStoreDataSource dataSource = dataSourceManager.getDataSource();
+                if ( dataSource == null /* if data source is unavailable */ )
```
MishaDemianenko: It's hard to see from a change set why do you need this change, but from a change itself it looks like this is a workaround and it does not look like a safe and correct thing to do.
How/when datasource can be null? Test that cover/show that situation?

MishaDemianenko (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/PlatformModule.java
davidegrohmann: Whenever you stop the database, which happens during storecopy for example.
I'll add a test.

davidegrohmann:
@MishaDemianenko added tests

PR #8062, opened 2016-09-30
Introduce lock acquisition timeout

MishaDemianenko:
Introduce property that defines timeout for shared and exclusive lock acquisition in lock clients.
In case if user configure timeout any acquisitions that take more then configured time interval will be terminated and LockAcquisitionTimeoutException will be raised.
Add support of new property to community and forseti lock managers.

Unify clocks usage during database construction/creation: now all components will use exactly the same instance of clock that provided by PlatformModule.


MishaDemianenko:
@lutovich any news?

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/LockAcquisitionTimeoutException.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.neo4j.graphdb.TransactionTerminatedException;
+import org.neo4j.kernel.api.exceptions.Status;
+import org.neo4j.storageengine.api.lock.ResourceType;
+
+/**
+ * Used in lock clients for cases when we unable to acquire a lock for a time that exceed configured
+ * timeout, if any.
+ *
+ * @see Locks.Client
```
lutovich, line 31: Maybe also `@see GraphDatabaseSettings#lock_acquisition_timeout`

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/community/RWLock.java
```diff
@@ -412,12 +422,19 @@ synchronized boolean acquireWriteLock( Object tx ) throws DeadlockDetectedExcept
...
-    private boolean waitUninterruptedly()
+    private boolean waitUninterruptedly( long lockAcquisitionTimeBoundary )
     {
         boolean addLockRequest;
         try
         {
-            wait();
+            if ( lockAcquisitionTimeoutMillis > 0 )
+            {
+                wait( Math.abs( lockAcquisitionTimeBoundary - clock.millis() ) );
```
lutovich: If `clock.millis() >>> lockAcquisitionTimeBoundary` due to thread scheduling or smth, thread will still wait for a while and then throw exception. Would it make sense to make previous condition
`if (lockAcquisitionTimeoutMillis > 0 && lockAcquisitionTimeBoundary > currentMillis)`
and then just `wait(lockAcquisitionTimeBoundary-currentMillis)` ?

lutovich (commented):
@MishaDemianenko sorry, I kind of forgot about this PR.

Changes look good, added one question/suggestion.
Also couple more general points:
- please make newly introduced fields in `LockManagerImpl`, `RWLock`, `ForsetiClient` and `ForsetiLockManager` final
- would be really nice to have ITs for lock acquisition timeout involving community/forseti, shared/exclusive and master/slave. I understand, they might be flaky or long running because of need to rely on time but still it is better than nothing.

MishaDemianenko:
@lutovich additional tests introduced, can you please take a look now!

community/neo4j/src/test/java/org/neo4j/locking/CommunityLockAcquisitionTimeoutIT.java
```diff
@@ -0,0 +1,230 @@
...
+                .setConfig( GraphDatabaseSettings.lock_acquisition_timeout, "2s" ).newGraphDatabase();
+
+        createTestNode( marker );
+    }
+
+    @After
+    public void tearDown()
+    {
+        secondTransactionExecutor.close();
+        clockExecutor.close();
```
lutovich, line 101: Should `database.shutdown()` be here as well?

lutovich (approved):
@MishaDemianenko looks good, one small question.

community/neo4j/src/test/java/org/neo4j/locking/CommunityLockAcquisitionTimeoutIT.java
MishaDemianenko: @lutovich absolutely, we need to shut it down; updated, thx!

PR #8064, opened 2016-09-30
Expose HTTP addresses in dbms.cluster.overview

apcj:
This PR publishes HTTP addresses in `dbms.cluster.overview`, alongside bolt addresses.

The procedure lives in the core-edge module, but there was a dependency problem preventing core-edge from reading HTTP configuration, since it lived in the neo4j-server module. Therefore, this PR also moves the HTTP configuration classes from neo4j-server into neo4j-dbms, which was chosen as lower level module that neo4j-core-edge already depended on.

While moving the configuration I found a number of problems with the way connector config was handled, so I've taken the chance to make it more straight-forward.

We now expect users to use the default names for the default set of connectors: any connector named `bolt`, `http` or `https` is assumed to refer to the respective protocol; you don't need to specify a type if you use these names. There are also sensible defaults for all the settings, so that no configuration at all is required. For example, the section of `neo4j.conf` for HTTPS used to look like this:

```
dbms.connector.https.type=HTTP
dbms.connector.https.enabled=true
dbms.connector.https.encryption=TLS     
dbms.connector.https.listen_address=:7473
```

All those parameters were required. For example, if you omitted a value for `dbms.connector.https.type` you could end up changing the port of the bolt connector!

Now the equivalent settings look like this:

```
#dbms.connector.https.enabled=true
#dbms.connector.https.listen_address=:7473
```

Note that both settings are commented out, because those are the default values; no explicit configuration is required.

The old mechanism, with arbitrary names is still supported, but we can choose to deprecate and then remove it in future.

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/server/configuration/ClientConnectorSettings.java
+++ b/community/dbms/src/main/java/org/neo4j/server/configuration/ClientConnectorSettings.java
@@ -1,0 +1,105 @@
+package org.neo4j.server.configuration;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import org.neo4j.graphdb.config.Setting;
+import org.neo4j.graphdb.factory.Description;
+import org.neo4j.graphdb.factory.GraphDatabaseSettings;
+import org.neo4j.helpers.AdvertisedSocketAddress;
+import org.neo4j.helpers.ListenSocketAddress;
+import org.neo4j.kernel.configuration.Config;
+
+import static java.util.Collections.singletonList;
+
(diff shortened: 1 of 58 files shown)


community/dbms/src/main/java/org/neo4j/server/configuration/ClientConnectorSettings.java
```diff
@@ -0,0 +1,105 @@
+package org.neo4j.server.configuration;
```
spacecowboy, line 20: You're missing a License header in this file

spacecowboy: Never mind

PR #8065, opened 2016-09-30
Remove fallback loading of community auth manager

henriknyman:
This fixes a bug where the auth manager service loading would fall back on community auth manager if an exception was thrown during instantiation.
- Moves all security setup into a `SecurityModule` that manages `AuthManager`, `SecurityLog` and auth procedures.
- Removes configurable auth manager loading from community edition
- Replaces it with configurable security module loading in enterprise edition
- Fails hard if the configured security module could not be instantiated


enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/EnterpriseEditionModule.java
```diff
@@ -85,7 +87,20 @@ protected StatementLocksFactory createStatementLocksFactory( Locks locks, Config
...
+    {
+        if ( platformModule.config.get( GraphDatabaseSettings.auth_enabled ) )
+        {
+            setupSecurityModule( platformModule,
+                    platformModule.logging.getUserLog( EnterpriseEditionModule.class ),
+                    procedures, platformModule.config.get( EnterpriseEditionSettings.security_module ) );
+        }
+        else
+        {
+            platformModule.life.add( platformModule.dependencies.satisfyDependency( EnterpriseAuthManager.NO_AUTH ) );
```
Mats-SX, line 103: Nice! This way we do no service loading at all if auth is disabled, right?

community/security/src/main/java/org/neo4j/server/security/auth/AuthProcedures.java
```diff
@@ -48,7 +51,8 @@ public void createUser(
             throws InvalidArgumentsException, IOException
     {
         BasicAuthSubject subject = BasicAuthSubject.castOrFail( authSubject );
-        subject.getAuthManager().newUser( username, password, requirePasswordChange );
+//        subject.getAuthManager().newUser( username, password, requirePasswordChange );
```
henriknyman: This will be removed by 46a3d6b9c7632293b128d271c3475ebae878dec3

craigtaverner:
This PR is incorporated into https://github.com/neo4j/neo4j/pull/8083

henriknyman:
This was merged together with https://github.com/neo4j/neo4j/pull/8083

PR #8066, opened 2016-09-30
Unbind an instance from the cluster.

jimwebber:
Current implementation merely checks that the database exists,
and that it isn't locked by an active Neo4j instance.

If all is good then it deletes the cluster-state directory reverting the
store to a Neo4j standalone/ha format (which can be used to
seed another core-edge cluster).

### DIFF
--- a/enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
+++ b/enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
@@ -64,24 +64,26 @@ public class UnbindFromClusterCommand implements AdminCommand
         @Override
         public String description()
         {
-            return "Removes cluster state data from the specified database making it suitable for use non-Core-Edge " +
-                    "databases";
+            return "Removes cluster state data from the specified database making it suitable for use in single " +
+                    "instance database, or for seeding a new cluster";
         }
 
         @Override
         public AdminCommand create( Path homeDir, Path configDir, OutsideWorld outsideWorld )
         {
-            return new UnbindFromClusterCommand( homeDir, configDir );
+            return new UnbindFromClusterCommand( homeDir, configDir, outsideWorld );
         }
     }
 
     private Path homeDir;
     private Path configDir;
+    private OutsideWorld outsideWorld;
 
-    UnbindFromClusterCommand( Path homeDir, Path configDir )
+    UnbindFromClusterCommand( Path homeDir, Path configDir, OutsideWorld outsideWorld )
     {
         this.homeDir = homeDir;
         this.configDir = configDir;
+        this.outsideWorld = outsideWorld;
     }
 
     @Override
@@ -95,14 +97,21 @@ public class UnbindFromClusterCommand implements AdminCommand
             Path pathToSpecificDatabase = config.get( DatabaseManagementSystemSettings.database_path ).toPath();
 
             Validators.CONTAINS_EXISTING_DATABASE.validate( pathToSpecificDatabase.toFile() );
-            confirmClusterStateDirectoryExists( Paths.get( pathToSpecificDatabase.toString(), "cluster-state" ) );
-            confirmTargetDirectoryIsWritable( pathToSpecificDatabase );
-            deleteClusterStateIn( clusterStateFrom( pathToSpecificDatabase ) );
+
+            if ( exists( Paths.get( pathToSpecificDatabase.toString(), "cluster-state" ) ) )
+            {
+                confirmTargetDirectoryIsWritable( pathToSpecificDatabase );
+                deleteClusterStateIn( clusterStateFrom( pathToSpecificDatabase ) );
+            }
+            else
+            {
+                outsideWorld.stdOutLine(
+                        format( "No cluster state found in %s. No work perfomed.", pathToSpecificDatabase ) );
+            }
         }
         catch ( StoreLockException e )
         {
-            throw new CommandFailed(
-                    "Database is currently locked. Please check that no other Neo4j instance using it.", e );
+            throw new CommandFailed( "Database is currently locked. Please shutdown Neo4j.", e );
         }
         catch ( IllegalArgumentException e )
         {
@@ -123,14 +132,6 @@ public class UnbindFromClusterCommand implements AdminCommand
     private Path clusterStateFrom( Path target )
     {
         return Paths.get( target.toString(), CLUSTER_STATE_DIRECTORY_NAME );
-    }
-
-    private void confirmClusterStateDirectoryExists( Path clusterStateDirectory ) throws UnbindFailureException
-    {
-        if ( !exists( clusterStateDirectory ) )
-        {
-            throw new UnbindFailureException( "Database %s is not bound to any cluster", clusterStateDirectory );
-        }
     }
 
     private void deleteClusterStateIn( Path target ) throws UnbindFailureException
--- a/enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
+++ b/enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
@@ -1,0 +1,184 @@
+/*
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.commandline.dbms;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+import org.neo4j.commandline.admin.AdminCommand;
+import org.neo4j.commandline.admin.CommandFailed;
+import org.neo4j.commandline.admin.IncorrectUsage;
+import org.neo4j.commandline.admin.OutsideWorld;
+import org.neo4j.helpers.Args;
+import org.neo4j.io.fs.DefaultFileSystemAbstraction;
+import org.neo4j.io.fs.FileSystemAbstraction;
+import org.neo4j.io.fs.FileUtils;
+import org.neo4j.kernel.StoreLockException;
+import org.neo4j.kernel.internal.StoreLocker;
+
+import static java.lang.String.format;
+import static java.nio.file.Files.exists;
+import static org.neo4j.coreedge.core.EnterpriseCoreEditionModule.CLUSTER_STATE_DIRECTORY_NAME;
+
+public class UnbindFromClusterCommand implements AdminCommand
+{
+    public static class Provider extends AdminCommand.Provider
+    {
+        public Provider()
+        {
+            super( "unbind" );
+        }
+
+        @Override
+        public Optional<String> arguments()
+        {
+            return Optional.of( "[--database=<name>] " );
+        }
+
+        @Override
+        public String description()
+        {
+            return "Removes cluster state data from the specified database making it suitable for use non-Core-Edge " +
+                    "databases";
+        }
+
+        @Override
+        public AdminCommand create( Path homeDir, Path configDir, OutsideWorld outsideWorld )
+        {
+            return new UnbindFromClusterCommand( homeDir, new DefaultFileSystemAbstraction() );
+        }
+    }
+
+    private Path homeDir;
+    private FileSystemAbstraction fsa;
+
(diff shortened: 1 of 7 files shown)


enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @Override
+    public void execute( String[] args ) throws IncorrectUsage, CommandFailed
+    {
+        Args parsedArgs = Args.parse( args );
+
+        try
+        {
+            Path pathToSpecificDatabase =
+                    homeDir.resolve( "data" ).resolve( "databases" ).resolve( databaseNameFrom( parsedArgs ) );
```
srbaker: This will not work if the user has changed the `data` directory in configuratoin. It's probably better to load config, and use `GraphDatabaseSettings.database_path`. See `CheckConsistencyCommand` for example.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,184 @@
...
+            throw new UnbindFailureException( e );
+        }
+    }
+
+    private String databaseNameFrom( Args parsedArgs )
+    {
+        String databaseName = parsedArgs.get( "database" );
+        if ( databaseName == null )
+        {
+            throw new IllegalArgumentException(
```
srbaker: We use `IncorrectUsage` in `AdminCommand`s. `CheckConsistencyCommand` is another example of this pattern.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,184 @@
...
+                    "No database name specified. Usage: neo4j-admin " + "unbind --database-<name>" );
+        }
```
srbaker: Usage information gets shown on `IncorrectUsage` errors, which is thrown above. So you can shrink this to: `return parsedArgs.interpretOption( "database", Converters.mandatory(), s -> s );`

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,184 @@
...
+                }
+                catch ( StoreLockException e )
+                {
+                    throw new CommandFailed( "Database is currently locked. Is a Neo4j instance still using it?" );
+                }
+            }
+        }
+    }
+
+    private void confirmTargetDatabaseDirectoryExists( Path target )
```
srbaker: You can probably use `Validators.CONTAINS_EXISTING_DATABASE` for this.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,184 @@
...
+    {
+        return Paths.get( target.toString(), CLUSTER_STATE_DIRECTORY_NAME );
+    }
+
+    private void confirmTargetDirectoryIsWritable( Path dbDir ) throws CommandFailed
+    {
+        Path lockFile = dbDir.resolve( StoreLocker.STORE_LOCK_FILENAME );
+        if ( exists( lockFile ) )
+        {
+            if ( Files.isWritable( lockFile ) )
```
benbc: You fail silently if the lockfile is unwritable (i.e. if the current user doesn't have the correct permissions to operate on the database). Presumably that's not what you want for this command. This case will happen, for example, with the Debian package if the operator forgets to `sudo -u neo4j`.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,184 @@
...
+            {
+                StoreLocker storeLocker = new StoreLocker( fsa );
+                try
+                {
+                    storeLocker.checkLock( dbDir.toFile() );
```
benbc: You don't unlock anywhere. I _think_ that's going to be okay with the current implementation because we're sure that after this command is done, the JVM will just exit. But since this class doesn't have a main method and is instead invoked by code from elsewhere, I don't think it's reasonable for the implementation to rely on an immediate exit. You may also run into problems with test ordering.

benbc (changes requested):
Could you not save yourself some effort by reusing the existing lock handling code from `DumpCommand`?

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
jimwebber: The existing `DumpCommand` code is where I started. I've merged my additions into a common abstraction now (`StoreLockChecker`).

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
jimwebber: Okey.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
jimwebber: Okey.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
benbc: Where is that?

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
jimwebber: Done.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
jimwebber: It's there now. I had to get on a plane prior to pushing.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,180 @@
...
+        catch ( UnbindFailureException | CannotWriteException e )
+        {
+            throw new CommandFailed( "Unbind failed: " + e.getMessage(), e );
+        }
+    }
+
+    private void confirmTargetDirectoryIsWritable( Path pathToSpecificDatabase )
+            throws CommandFailed, CannotWriteException
+    {
+        new StoreLockChecker().withLock( pathToSpecificDatabase );
```
benbc, line 129: Do you have a reason for not unlocking the store once you're done?

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,193 @@
...
+        @Override
+        public Optional<String> arguments()
+        {
+            return Optional.of( "[--database=<name>] " );
+        }
+
+        @Override
+        public String description()
+        {
+            return "Removes cluster state data from the specified database making it suitable for use non-Core-Edge " +
```
maxsumrall: "...suitable for use _in_ non-Core-Edge databases."

Forgot a word.

maxsumrall: My suggestion that doesn't use the term `Core-Edge`:
"Removes cluster data from the specified database making it suitable to use as a single instance or to bind to a different cluster."

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,193 @@
...
+            catch ( IOException e )
+            {
+                wrapIOException( e );
+            }
+
+        }
+        catch ( StoreLockException e )
+        {
+            throw new CommandFailed(
+                    "Database is currently locked. Please check that no other Neo4j instance using it.", e );
```
maxsumrall: "Please check that no other Neo4j instance _is_ using it."
Or, "Please check that it is not being used by another Neo4j instance."

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,193 @@
...
+    private Path clusterStateFrom( Path target )
+    {
+        return Paths.get( target.toString(), CLUSTER_STATE_DIRECTORY_NAME );
+    }
+
+    private void confirmClusterStateDirectoryExists( Path clusterStateDirectory ) throws UnbindFailureException
+    {
+        if ( !exists( clusterStateDirectory ) )
+        {
+            throw new UnbindFailureException( "Database %s is not bound to any cluster", clusterStateDirectory );
```
maxsumrall: Should we add some message saying that everything is fine? For example: "Unbinding not necessary."

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,193 @@
...
+import org.neo4j.server.configuration.ConfigLoader;
+
+import static java.lang.String.format;
+import static java.nio.file.Files.exists;
+import static org.neo4j.coreedge.core.EnterpriseCoreEditionModule.CLUSTER_STATE_DIRECTORY_NAME;
+
+public class UnbindFromClusterCommand implements AdminCommand
+{
+
+    private StoreLockChecker storeLockChecker;
```
maxsumrall: Can be final

maxsumrall (changes requested):

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
benbc: Isn't the most likely problem that _this_ database is running?

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
benbc: Does this surface as an error? I would expect this command to be idempotent -- and therefore exit 0 in this case, possible (but not definitely) with a notification that nothing was done.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
jimwebber: Done.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
jimwebber: String no longer appears in code.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
jimwebber: @benbc do you have some concrete words that you'd prefer?

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
jimwebber: It's now idempotent. If there's no `cluster-state` directory then we'll skip the remaining steps.

maxsumrall (approved):

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
benbc: The equivalent text for `neo4j-admin dump`, which I think is reasonable is "the database is in use -- stop Neo4j and try again". That identifies the most likely cause and gives clear instructions on what needs to be done to fix the problem.

enterprise/core-edge/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -0,0 +1,181 @@
...
+
+            if ( exists( Paths.get( pathToSpecificDatabase.toString(), "cluster-state" ) ) )
+            {
+                confirmTargetDirectoryIsWritable( pathToSpecificDatabase );
+                deleteClusterStateIn( clusterStateFrom( pathToSpecificDatabase ) );
+            }
+            else
+            {
+                outsideWorld.stdOutLine(
+                        format( "No cluster state found in %s. No work perfomed.", pathToSpecificDatabase ) );
```
benbc: Informational output like this should go to stderr.

jimwebber: Done.

benbc (approved):

PR #8073, opened 2016-10-03
Fix bug with `allowed` procedure annotation

Mats-SX:
Previously, the `allowed` flag for procedures would restrict the transaction in such a way that subsequent operations in the Cypher statement were also restricted, leading to cases where write operations that were authorized were refused by the system. This commit pushes down the transaction restriction to wrap only the procedure call, and fall back to the normal access mode once the procedure has returned.

changelog: Fixed bug with `allowed` procedure annotation impacting later parts of same cypher query

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
@@ -1531,18 +1531,39 @@ public class OperationsFacade
     }
 
     private RawIterator<Object[],ProcedureException> callProcedure(
-            QualifiedName name, Object[] input, AccessMode mode )
+            QualifiedName name, Object[] input, final AccessMode mode )
             throws ProcedureException
     {
         statement.assertOpen();
 
+        final RawIterator<Object[],ProcedureException> procedureCall;
         try ( KernelTransaction.Revertable ignore = tx.overrideWith( mode ) )
         {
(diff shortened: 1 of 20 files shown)


Mats-SX:
Still TODO: use an optimised iterator for procedures that are executed by subjects that are not activated through the `allowed` procedure flag.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -1510,19 +1495,88 @@ public void unregisterExecutingQuery( ExecutingQuery executingQuery )
...
             ctx.put( Context.THREAD, Thread.currentThread() );
-            return procedures.callProcedure( ctx, name, input );
+            procedureCall = procedures.callProcedure( ctx, name, input );
         }
+        return new RawIterator<Object[],ProcedureException>()
+        {
+            @Override
+            public boolean hasNext() throws ProcedureException
+            {
+                try ( KernelTransaction.Revertable ignore = tx.overrideWith( mode ) )
```
henriknyman, line 1565: This iterator has overhead that we want to avoid in the common case to not risk a bad performance regression. We only need this iterator if the `allowed` annotation was actually used on the procedure (and only if auth is enabled), and if the procedure call mode was  read-only ( `LazyReadOnlyCallMode`). In the other cases the iterator is eagerly exhausted directly after call, so we do not need to override access mode row-by-row.

PR #8076, opened 2016-10-04
Fix PageSwapperTest failure on Windows

chrisvest:


chrisvest:
Both the Linux and the Windows build fails on unrelated test failures.

PR #8077, opened 2016-10-04
Convert parameters to known cypher types

### LINKED ISSUE #7115: [3.0][Procedure] Can't use Long safely as procedure argument type
**Neo4j Version:** 3.0.1
**Operating System:** OSX
**API:** Browser/PHP

In the below example, using Long as procedure argument's type is not safe, because : 
1. It is treated as long when you pass a number as procedure call argument.
2. It is treated as integer when you pass a the same number as query parameter and use the parameter as procedure call parameter

Which means that we can't use a type of `Integer` or `Long` as procedure argument type, because the type coerced differs from Cypher inline vs Query parameter.
### Steps to reproduce

``` java
package proc;

import org.neo4j.procedure.Context;
import org.neo4j.procedure.Name;
import org.neo4j.procedure.Procedure;

import java.util.stream.Stream;

public class LongInt {

    @Procedure
    public Stream<Record> test(@Name("object") Object object) {
        return Stream.of(new Record(object.getClass()));
    }

    public class Record {

        public final Object value;

        public Record(Object value) {
            this.value = value;
        }

    }
}
```

Procedure parameter passed inline in the Cypher query

``` java
@Test
public void testIt1() {
  Driver driver = GraphDatabase.driver( "bolt://localhost" );
  Session session = driver.session();
  StatementResult rs = session.run("CALL proc.test(500)");
  rs.consume();
  session.close();
  driver.close();

  // Returns class java.lang.Long
}
```

Procedure parameter passed from query parameters map // Procedure parse it as Integer

``` java
@Test
public void testIt2() {
  Driver driver = GraphDatabase.driver( "bolt://localhost" );
  Session session = driver.session();
  Map<String, Object> params = new HashMap<>();
  params.put("v", 500);
  StatementResult rs = session.run("CALL proc.test({v})", params);
  rs.consume();
  session.close();
  driver.close();

  // Returns class java.lang.Integer
}
```

pontusmelke:
Cypher should only know about longs, and doubles, hence when parameters are past in to cypher
we should convert them to these known entities. Currently this causes issues when using parameters to procedures in languages where there is no concept of 64 bit integers and floating point numbers.

fixes #7115

### DIFF
--- a/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/InQueryProcedureCallAcceptanceTest.scala
+++ b/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/InQueryProcedureCallAcceptanceTest.scala
@@ -19,9 +19,10 @@
  */
 package org.neo4j.internal.cypher.acceptance
 
-import java.util
+import java.{lang, util}
 
 import org.neo4j.cypher._
+import org.neo4j.graphdb.Result
 import org.neo4j.kernel.api.proc.Neo4jTypes
 
 class InQueryProcedureCallAcceptanceTest extends ProcedureCallAcceptanceTest {
@@ -143,7 +144,7 @@ class InQueryProcedureCallAcceptanceTest extends ProcedureCallAcceptanceTest {
     createLabeledNode("C")
 
     //When
-    a [SyntaxException] shouldBe thrownBy(
+    a[SyntaxException] shouldBe thrownBy(
       execute("WITH 'Hi' AS label CALL db.labels YIELD label RETURN *")
     )
   }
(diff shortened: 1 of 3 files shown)


davidegrohmann:
I don't like this change for several reasons:
- paramaters are not converted on purpose since it is the only way to store types not supported by cypher in the database (e.g., byte arrays)
- breaks retro compatibility in a bug fix release
- it converts only numbers and not arrays containing numbers or maps containing numbers

systay:
@davidegrohmann Looks to me like this would not interfere with other types, so byte arrays should still work. It also looks like it is recursive, handling incoming arrays and maps like expected: https://github.com/neo4j/neo4j/pull/8077/files#diff-a349ea71ef8eed2b58f05f82d8923fe1R71 and https://github.com/neo4j/neo4j/pull/8077/files#diff-a349ea71ef8eed2b58f05f82d8923fe1R73

The last point is that this change breaks retro compatibility. It does change behaviour, but that is expected since it's fixing a bug. Do you think that keeping the original JVM type is an important feature to keep?

pontusmelke:
@davidegrohmann 
- I wasn't expecting the argument on fixing this in a patch release from you, but I am very happy to just do this is 3.1. That is the way I want to do things but I thought I lost that battle a long time ago.
- I misread the code, I see now this will not work for neither `java.util.List` nor `java.util.Map` but that can be easily fixed. Nice catch!
- Arrays are intentionally left out for the reason you gave, and arrays are not part of the type system (cypher can't do anything with them and we can't pass them in to procedures).

davidegrohmann:
@systay I would find the solution quite confusing if we convert all numbers to either long or double but leave array of number untouched.

PR #8078, opened 2016-10-04
Bump high limit store version

MishaDemianenko:
Bump high limit store version to have clear separation between milestone and release version store format.
Format by itself does not contain any changes.

changelog: Bump high limit store version for 3.1 release. 
Stores created with milestone store versions are no longer acceptable.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/StoreId.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/StoreId.java
@@ -26,13 +26,8 @@ import java.io.ObjectOutput;
 import java.security.SecureRandom;
 import java.util.Random;
 
-import org.neo4j.kernel.impl.store.format.standard.StandardV3_0;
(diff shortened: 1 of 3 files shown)


MishaDemianenko:
@tinwelint can you please take a look

burqen:
@MishaDemianenko To start db on store with old store version you need to set `dbms.allow_format_migration` right? If so -> changelog

MishaDemianenko:
@burqen thx, yes! 🐛

tinwelint:
Store versions and migration is mind boggling and a time travel, but I think this is just what we wanted, yes

thobe:
I would really have liked to see a test to verify the behaviours we want. I tried writing one, but the code did not allow testing this on the unit-level, it requires an integration test with zipped up store files to test this properly. I did manage to convince myself that this looks like what it should look like in the process of trying though, so I'm happy to defer that testing until later.

PR #8079, opened 2016-10-04
3.1 remove neo4j admin usermanag

fickludd:
Due to architectural problems, most user management has been removed from neo4j-admin. The only command that remains is the `neo4j-admin users set-password`. However, this command has been modified to only set an initial user on server startup.

no-changelog due to this being reported in the changelog entry in #8088


community/security/src/main/java/org/neo4j/server/security/auth/BasicAuthManager.java
```diff
@@ -47,46 +47,70 @@
...
 
         if ( userRepository.numberOfUsers() == 0 )
         {
-            newUser( "neo4j", "neo4j", true );
+            if ( initialUserRepository.numberOfUsers() == 0 )
+            {
+                newUser( "neo4j", "neo4j", true );
+            }
+        }
+        for ( String username : initialUserRepository.getAllUsernames() )
```
craigtaverner, line 87: Should it not be that this code for looking at initialUserRepository is only considered for the case when default user might be created (ie. inside the if-block above)?

community/security/src/test/java/org/neo4j/commandline/admin/security/SetPasswordCommandTest.java
```diff
@@ -19,97 +19,132 @@
...
-        catch ( IncorrectUsage e )
-        {
-            assertThat( e.getMessage(), containsString( "password" ) );
-        }
+        assertException( () -> usersCommand.execute( arguments ), IncorrectUsage.class, "password" );
     }
 
     @Test
-    public void shouldFailSetPasswordWithNonExistingUser() throws Exception
+    public void shouldAllowSetPasswordWithPasswordChangeRequiredFlag() throws Exception
```
craigtaverner, line 71: Why remove this test? It tests a required feature.

community/security/src/test/java/org/neo4j/commandline/admin/security/SetPasswordCommandTest.java
```diff
@@ -19,97 +19,132 @@
...
-        {
-            assertThat( e.getMessage(), containsString( "does not exist" ) );
-        }
+        String[] arguments = {"set-password", "neo4j", "123", "--requires-password-change"};
+        usersCommand.execute( arguments );
     }
 
     @Test
-    public void shouldRunSetPasswordWithDefaultUser() throws Throwable
+    public void shouldCreateInitialUserWithPasswordChangeRequired() throws Throwable
```
craigtaverner, line 78: The set-password command cannot create new users. It can only set passwords. The fact that the default user 'neo4j' is _created_ when setting up the database is an implementation detail. We could just as easily ship pre-configured auth and roles files with default users and roles. Treat 'neo4j' as already existing from the users perspective.

community/security/src/test/java/org/neo4j/commandline/admin/security/UsersCommandIT.java
```diff
@@ -40,317 +40,8 @@ public void setup()
...
-        // When running 'list' with filter, expect specified user
-        assertSuccessfulUserCommand( "list", args( "other" ), "another" );
-    }
-
-    //
-    // Tests for set-password command
-    //
-
-    @Test
-    public void shouldGetUsageErrorsWithSetPasswordCommandAndNoArgs() throws Throwable
```
craigtaverner, line 86: Why delete all set-password integration tests? I think these tests help clarify the expected behavior of this command.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/InternalFlatFileRealm.java
```diff
@@ -198,19 +204,38 @@ private void readFilesFromDisk( int attemptLeft, java.util.List<String> failures
...
+            if ( userRepository.numberOfUsers() == 0 )
             {
-                newUser( "neo4j", "neo4j", true );
+                if ( initialUserRepository.numberOfUsers() == 0 )
+                {
+                    newUser( "neo4j", "neo4j", true );
+                    return Collections.singleton( "neo4j" );
+                }
+            }
+            for ( String username : initialUserRepository.getAllUsernames() )
```
craigtaverner, line 219: Why is this code not in the if block above?

craigtaverner (commented):
There are behaviour changes that need to be fixed, at least in a followup PR.

PR #8080, opened 2016-10-04
Ignore ACK_FAILURE messages when in INTERRUPTED state.

maxsumrall:


zhenlineo:
👍

maxsumrall:
I've run the same soak test on a build with this PR, and have not observed any protocol breach messages in any of the debug.log files.

PR #8083, opened 2016-10-04
Remove fallback loading of community auth manager and fix start-up crashes on misconfigured server, due to early security logging

fickludd:
This fixes a bug where the auth manager service loading would fall back on community auth manager if an exception was thrown during instantiation.
- Moves all security setup into a SecurityModule that manages AuthManager, SecurityLog and auth procedures.
- Removes configurable auth manager loading from community edition
- Replaces it with configurable security module loading in enterprise edition
- Fails hard if the configured security module could not be instantiated
- Fixed issue where the security log was used before the `JobScheduler` had been initialized, meaning that the logging failed as it is asynchronous.

changelog: Refine behavior of community and enterprise support for dynamic auth managers
- Remove support for dynamic auth manager loading from community, this will be enterprise only
- Disables enterprise from falling back to community auth manager on security initialization errors, so config mistakes will correctly result in server startup failure
- Removes availability of native user and role management procedures when native realm/provider is not enabled [#06a26bc1](https://github.com/neo4j/neo4j/pull/8083/commits/06a26bc15ccb75fdfaea8c44b7a6cfca857f9b67)
- Fixed issue with `changeUserPassword` command terminating the user session [#27eefe2b](https://github.com/neo4j/neo4j/pull/8083/commits/27eefe2b60ab3edd4d1228cda51ba3943d495dec)


henriknyman:
I noticed that `org.neo4j.kernel.impl.logging.StoreLogService` does not seem to have this problem. This fix solves the immediate issue so I think it should be merged, but I think long-term it would simplify things by moving security log into LogService and sharing the existing implementation.

PR #8084, opened 2016-10-04
Fix problem with terminating inner transactions

systay:
PlaceboTransaction was not getting a KernelTransaction, it was getting a
Supplier<KernelTransaction>, which would return the currently thread bound transaction.

This works well in most cases, since transactions are thread bound and should not be used
from other threads. When using terminate() though, the caller is almost always from a different
thread, and in these situations we can't use a supplier.

changelog: Fix problem where terminating transaction from a different thread could end up terminating wrong transaction when using nested transactions. This could occur when one user attempts to terminate another user's transaction, but incorrectly ends up having their own transaction terminated.


burqen:
@systay Is this...
a) fixing a reported GH issue?
b) fixing a costumer issue?
c) fixing an internally discovered issue?

If a -> Please add changelog tag with link to issue.
If b or c -> Please add a more descriptive changelog tag, like in what scenarios this would show up.

systay:
@burqen Done

PR #8074, opened 2016-10-03
Fix: make dbms.security.showCurrentUser fallback to show LDAP user, and fix flaky log in init tests

fickludd:
Made dbms.security.showCurrentUser procedure show authenticated user even if native security realm is not used

Internal changes:
- Refactored usermanagment to work through a delegating usermanager that now had all responsibility to log user management operations to the security log. This `PersonalUserManager` has the active AuthSubject and also checks admin status where needed.
- Fixed flaky tests for bolt init and REST UserService change password logging

changelog: Made dbms.security.showCurrentUser procedure show authenticated user even if native security realm is not used

PR #8085, opened 2016-10-05
Fix: flaky log on init tests

fickludd:
Resolved flakyness in 
`EnterpriseAuthenticationIT.shouldLogInitPasswordChange`
`EnterpriseAuthenticationIT.shouldLogFailedInitPasswordChange`

builds on #8065


fickludd:
This fix is embedded in #8074 instead

PR #8088, opened 2016-10-05
Create a dedicated `set-initial-password` command

Mats-SX:
With the removal of the `users` command from `neo4j-admin`, one use-case was left unsupported: setting the password of the initial user. This PR introduces a dedicated `neo4j-admin` subcommand `set-initial-password` that accomplishes that, and does nothing more.

changelog: Simplify setting initial database password using `neo4j-admin` command-line tool
- Remove `users` and `roles` command from `neo4j-admin` [#8079](https://github.com/neo4j/neo4j/pull/8079)
- Create a new `set-initial-password` command for first-time setting of password on offline database
- User and role management is now only available in online database through procedures


Mats-SX:
This PR handles the comments posted in #8079

fickludd:
This looks like a nice compromise between the hastily extended functionality and the old functionality. I approve for whatever that's worth.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/InternalFlatFileRealm.java
```diff
@@ -212,26 +212,17 @@ private void readFilesFromDisk( int attemptLeft, java.util.List<String> failures
         {
             if ( userRepository.numberOfUsers() == 0 )
             {
-                if ( initialUserRepository.numberOfUsers() == 0 )
+                User neo4j = newUser( INITIAL_USER_NAME, "neo4j", true );
```
craigtaverner, line 215: Shouldn't this line be in the if-block below?

OliviaYtterbrink: No, we always want the initial user. The next check is only to see if the initial password has been set or not.

henriknyman:
I think it would be more useful with a command that could also set the initial username (i.e. `set-initial-user`), but I guess we could always add that later. Forcing you to have an initial user named `neo4j` seems a bit old-school...

Mats-SX:
@henriknyman I agree.

PR #8098, opened 2016-10-06
Added status code for AcquireLockTimeoutException

zhenlineo:
to better help a client, i.e. a driver, to classify the error for end users.

AcquireLockTimeoutException is mainly used in CE to indicate that we cannot grab a lock on a `Leader`. There is a chance that we will get this exception if a leadership switch happened after we've checked that the server is the `Leader`, but before we really start to take a lock.

### DIFF
--- a/enterprise/core-edge/src/main/java/org/neo4j/coreedge/core/state/machines/locks/LeaderOnlyLockManager.java
+++ b/enterprise/core-edge/src/main/java/org/neo4j/coreedge/core/state/machines/locks/LeaderOnlyLockManager.java
@@ -21,18 +21,20 @@ package org.neo4j.coreedge.core.state.machines.locks;
 
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
-import java.util.concurrent.TimeoutException;
 
 import org.neo4j.coreedge.core.consensus.LeaderLocator;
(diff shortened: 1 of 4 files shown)


enterprise/core-edge/src/main/java/org/neo4j/coreedge/core/state/machines/locks/LeaderOnlyLockManager.java
```diff
@@ -121,17 +121,18 @@ private synchronized int acquireTokenOrThrow()
...
         catch ( ExecutionException e )
         {
-            throw new AcquireLockTimeoutException( e, "Failed to acquire lock token." );
+            throw new AcquireLockTimeoutException( e, "Failed to acquire lock token.", NotALeader );
         }
         catch ( InterruptedException e )
         {
             Thread.currentThread().interrupt();
-            throw new AcquireLockTimeoutException( e, "Failed to acquire lock token." );
+            throw new AcquireLockTimeoutException( e, "Failed to acquire lock token.", NotALeader );
```
digitalstain: Surely the status here should be Interrupted

zhenlineo: Done

PR #8100, opened 2016-10-06
Ensures property tokens are not attempted to be created on read operations

digitalstain:
This fixes an issue where, under certain conditions, a read query would
 attempt to create property tokens for nodes and relationships.


community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/commands/expressions/Property.scala
```diff
@@ -35,11 +35,15 @@ case class Property(mapExpr: Expression, propertyKey: KeyToken)
   def apply(ctx: ExecutionContext)(implicit state: QueryState): Any = mapExpr(ctx) match {
     case null => null
     case n: Node =>
-      val propId = propertyKey.getOrCreateId(state.query)
-      state.query.nodeOps.getProperty(n.getId, propId)
+      propertyKey.getOptId(state.query) match {
+        case None => null
```
pontusmelke, line 39: ~~Here we should create the property key, right? In this case it need to do the `propertyKey.getOrCreateId(state.query)`, or we need to make the `getOrCreate` smarter.~~ 

I was wrong I missed that `getOptId` was going down to the database to fetch the token if unresolved. LGTM.

Mats-SX:
Added a test for this behaviour change in #8113.

PR #8103, opened 2016-10-06
Restructure security settings in neo4j.conf

henriknyman:
- Move the security log settings to the Logging section, but keep the
  `dbms.security.log_successful_authentication` in the top section of the
  Security section.
- Remove the auth cache sub-section and move to the top section.
- Change the default value of ldap host setting to `localhost`
- Set the default value of ldap user search base setting to an example value.


benbc:
Looks good to me.

PR #8110, opened 2016-10-07
Use `neo4j-admin backup` in tests

spacecowboy:
This removes a bunch of deprecation warnings from the test logs (and of
course stops using deprecated code).

Had to make sure `neo4j-admin` (and as a side-effect other tools too)
could be started by the tests where System.env won't have the variables
set by the wrapper script.

Adding the disabled `auth` setting for embedded offline configs was also
necessary, so the tests did not try to load classes which were not
available to the class loader.

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/server/configuration/ConfigLoader.java
+++ b/community/dbms/src/main/java/org/neo4j/server/configuration/ConfigLoader.java
@@ -61,10 +61,10 @@ public class ConfigLoader
                 settingsClassesSupplier );
     }
 
-    public Config loadOfflineConfig( Optional<File> homeDir, Optional<File> configFile,
-            Pair<String,String>... configOverrides )
+    public Config loadOfflineConfig( Optional<File> homeDir, Optional<File> configFile )
     {
-        return overrideBoltSettings( loadConfig( homeDir, configFile, configOverrides ) );
+        return overrideBoltSettings( loadConfig( homeDir, configFile,
+                Pair.of( GraphDatabaseSettings.auth_enabled.name(), Settings.FALSE ) ) );
     }
 
     private Map<String, String> calculateSettings( Optional<File> homeDir,
--- a/community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
+++ b/community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
@@ -31,8 +31,8 @@ public class AdminTool
 {
     public static void main( String[] args )
     {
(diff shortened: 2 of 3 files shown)


community/dbms/src/main/java/org/neo4j/server/configuration/ConfigLoader.java
```diff
@@ -61,6 +61,12 @@ public Config loadConfig( Optional<File> homeDir, Optional<File> configFile,
                 settingsClassesSupplier );
     }
 
+    public Config loadOfflineConfig( Optional<File> homeDir, Optional<File> configFile )
+    {
+        return loadOfflineConfig( homeDir, configFile,
+                Pair.of( GraphDatabaseSettings.auth_enabled.name(), Settings.FALSE ) );
+    }
+
```
benbc: Where does this get called from? I'm worried that it's a bit unexpected to get auth disabled just because you don't pass any overrides.

spacecowboy: It is called by all of our tools, and in tests.

My reasoning is, it's an offline config. E.g., all connectors are disabled. Having auth disabled is fine because there is nothing to connect to, and hence nothing to break into.

Furthermore, it is actually an override of an override, see 

https://github.com/spacecowboy/neo4j/blob/e4e651d401a759136f1e9bfb1e32870c59674566/community/dbms/src/main/java/org/neo4j/server/configuration/ConfigLoader.java#L113

which I found to be very unexpected.

benbc: So why don't callers of this get the same behaviour: https://github.com/spacecowboy/neo4j/blob/e4e651d401a759136f1e9bfb1e32870c59674566/community/dbms/src/main/java/org/neo4j/server/configuration/ConfigLoader.java#L70?

spacecowboy: Because if you have your own overrides, you know what you want?

benbc: I think it would be less surprising to always default auth to off in `loadOfflineConfig` unless it's explicitly enabled in the overrides.

spacecowboy: it was only called by the other `loadOfflineConfig`, so I removed it and made it a single method.

benbc:
Oh, perfect. This now LGTM.

PR #8113, opened 2016-10-07
Enforce write privileges for token writes

Mats-SX:
Cypher was recently changed to not perform token writes in read queries.
This commit adds a test that will fail if it would do so. It further enforces that
acquiring `TokenWriteOperations` requires the `AccessMode` to allow writes.

This PR adds a test for the change in #8100.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
@@ -108,6 +108,12 @@ public class KernelStatement implements TxStateHolder, Statement
     public TokenWriteOperations tokenWriteOperations()
     {
         accessCapability.assertCanWrite();
+
+        if ( !transaction.mode().allowsWrites() )
+        {
+            throw transaction.mode().onViolation(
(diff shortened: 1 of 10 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
```diff
@@ -108,6 +108,12 @@ public ProcedureCallOperations procedureCallOperations()
     public TokenWriteOperations tokenWriteOperations()
     {
         accessCapability.assertCanWrite();
+
+        if ( !transaction.mode().allowsWrites() )
```
lutovich: Similar checks already exist in `#dataWriteOperations()` and `#schemaWriteOperations()`. I think it is worth to create a method like `assertWritesAllowed(String message)` at this point.

lutovich (changes requested):
@Mats-SX change looks good to me. It is possible to prettify code a bit and extract an assertion method. More importantly, build has test failures related to the change.

Mats-SX:
@lutovich Thanks for the review! We made this commit to address your comment: https://github.com/neo4j/neo4j/pull/8113/commits/c0c061de74e884838608f757a3db4934e0b67376

lutovich (approved):

PR #8114, opened 2016-10-07
Snapshot access mode on transaction start

henriknyman:
A snapshot of the access mode is taken on begin transaction, which will keep
the authorization state at that point active for the life time of the transaction.

This fixes the issue that authorization can expire during long running
transactions with LDAP authorization
(with `dbms.security.realms.ldap.authorization.use_system_account=false`) or
plugin authorization using the simplified `AuthPlugin` interface.

changelog: Keep a snapshot of the authorization state for the life time of a transaction


community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/OverriddenAccessMode.java
```diff
@@ -80,6 +80,12 @@ public String name()
         }
     }
 
+    @Override
+    public AccessMode getSnapshot()
+    {
+        return null;
```
Mats-SX: Shouldn't this be `overridden.getSnapshot()`?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
```diff
@@ -586,36 +585,37 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
   }
 
   type KernelProcedureCall = (KernelQualifiedName, Array[AnyRef]) => RawIterator[Array[AnyRef], ProcedureException]
+  type KernelFunctionCall = (KernelQualifiedName, Array[AnyRef]) => AnyRef
+
+  private def shouldElevate(allowed: Array[String]): Boolean = {
+    // We have to be careful with elevation, since we cannot elevate permissions in a nested procedure call
+    // above the original allowed procedure mode. With enforce this by checking if mode is already an overridden mode.
```
Mats-SX: Typo: `With` -> `We`

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/OverriddenAccessMode.java
Mats-SX: It just feels a bit unfriendly to return `null`.

henriknyman: This has already been changed in refactoring

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
henriknyman: Fixed

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
```diff
@@ -586,34 +585,39 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
   }
 
   type KernelProcedureCall = (KernelQualifiedName, Array[AnyRef]) => RawIterator[Array[AnyRef], ProcedureException]
+  type KernelFunctionCall = (KernelQualifiedName, Array[AnyRef]) => AnyRef
+
+  private def shouldElevate(allowed: Array[String]): Boolean = {
+    // We have to be careful with elevation, since we cannot elevate permissions in a nested procedure call
+    // above the original allowed procedure mode. We enforce this by checking if mode is already an overridden mode.
+    val mode = transactionalContext.accessMode
+    allowed.nonEmpty && !mode.isOverridden && mode.getOriginalAccessMode.allowsProcedureWith(allowed)
```
Mats-SX, line 594: How about the case where a `WRITE` procedure calls a `READ` procedure? That's now in the domain of `override` rather than `elevate`, is that correct?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
```diff
@@ -586,34 +585,39 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
...
 
   override def callSchemaWriteProcedure(name: QualifiedName, args: Seq[Any], allowed: Array[String]) = {
-    val call: KernelProcedureCall = transactionalContext.accessMode match {
-      case a: AuthSubject if a.allowsProcedureWith(allowed) =>
-        transactionalContext.statement.procedureCallOperations.procedureCallSchema(_, _, AccessMode.Static.OVERRIDE_SCHEMA)
-      case _ =>
+    val call: KernelProcedureCall =
+      if (shouldElevate(allowed))
+        transactionalContext.statement.procedureCallOperations.procedureCallSchemaElevated(_, _)
+      else
```
Mats-SX: Nice cleanup.

community/kernel/src/main/java/org/neo4j/kernel/api/ProcedureCallOperations.java
```diff
@@ -82,14 +79,28 @@
...
+     */
+    Object functionCall( QualifiedName name, Object[] arguments ) throws ProcedureException;
+
+    /** Invoke a read-only function by name, and elevate the transaction's access mode to
+     * the given access mode for the duration of the function execution.
+     * @param name the name of the function.
+     * @param arguments the function arguments.
+     * @throws ProcedureException if there was an exception thrown during function execution.
+     */
+    Object functionCallElevated( QualifiedName name, Object[] arguments ) throws ProcedureException;
```
Mats-SX: Still not convinced on this one :)

community/kernel/src/main/java/org/neo4j/kernel/api/ProcedureCallOperations.java
```diff
@@ -82,14 +79,28 @@
...
-    RawIterator<Object[], ProcedureException> procedureCallSchema( QualifiedName name, Object[] arguments, AccessMode override )
+    RawIterator<Object[], ProcedureException> procedureCallSchemaElevated( QualifiedName name, Object[] arguments )
             throws ProcedureException;
+
+    /** Invoke a read-only function by name
+     * @param name the name of the function.
+     * @param arguments the function arguments.
+     * @throws ProcedureException if there was an exception thrown during function execution.
+     */
+    Object functionCall( QualifiedName name, Object[] arguments ) throws ProcedureException;
```
Mats-SX, line 97: I guess that wasn't introduced by this PR, but should we really throw `ProcedureException` here?

community/kernel/src/main/java/org/neo4j/kernel/api/security/AccessMode.java
```diff
@@ -228,142 +175,55 @@ public boolean allowsSchemaWrites()
...
+
+    /**
+     * Determines whether this subject is allowed to execute a procedure with the parameter string in its
+     * procedure annotation.
+     *
+     * @param roleNames
+     * @return
+     * @throws InvalidArgumentsException
+     */
+    default boolean allowsProcedureWith( String[] roleNames ) throws InvalidArgumentsException
```
Mats-SX: This sneakily introduces the concept of `role` into community. Let's perhaps rename this parameter to `allowed`?

community/kernel/src/main/java/org/neo4j/kernel/api/security/AccessMode.java
```diff
@@ -228,142 +175,55 @@ public boolean allowsSchemaWrites()
...
+    }
+
+    default boolean isOverridden()
+    {
+        return false;
+    }
+
+    /**
+     * Determines whether this subject is allowed to execute a procedure with the parameter string in its
+     * procedure annotation.
```
Mats-SX: 'this subject' is incorrect in this context. `this` is an `AccessMode`, which is different from a subject. I suggest something like this:

Determines whether this mode allows execution of a procedure with the parameter string array in its procedure annotation.

community/kernel/src/main/java/org/neo4j/kernel/api/security/AccessMode.java
```diff
@@ -228,142 +175,55 @@ public boolean allowsSchemaWrites()
...
+     *
+     * @param roleNames
+     * @return
+     * @throws InvalidArgumentsException
```
Mats-SX: I'd like a JavaDoc for these three as well. Specifically `return` would be helpful. I'm thinking something along the lines of 

`true` if (and only if?) the parameter array contains a string that encodes a privilege to allow this mode to execute it

but it feels a bit contrived. Is this really the correct place for this method?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
```diff
@@ -225,8 +223,8 @@ final void forceClose()
 
     final Optional<String> username()
     {
-        AccessMode mode = transaction.mode();
-        return ( mode instanceof AuthSubject ) ? Optional.of( ((AuthSubject) mode).username() ) : Optional.empty();
+        String username = transaction.mode().username();
+        return Optional.of( username );
```
Mats-SX: Could inline.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -800,7 +799,7 @@ public long getCommitTime()
     public Revertable overrideWith( AccessMode mode )
     {
         AccessMode oldMode = this.accessMode;
-        this.accessMode = new OverriddenAccessMode( oldMode, mode );
+        this.accessMode = mode;
```
Mats-SX, line 802: Nice simplification.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/RestrictedAccessMode.java
```diff
@@ -0,0 +1,55 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api.security;
+
+import org.neo4j.kernel.api.security.AccessMode;
+
+public class RestrictedAccessMode extends LayeredAccessMode
+{
+    public RestrictedAccessMode( AccessMode originalMode,
+            AccessMode overriddenAccessMode )
```
Mats-SX: It is a bit confusing reading these as `original` and `overridden`. How about `original` and `restricting`? Or `restrictedTo`?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/ElevatedAccessMode.java
```diff
@@ -0,0 +1,55 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api.security;
+
+import org.neo4j.kernel.api.security.AccessMode;
+
+public class ElevatedAccessMode extends LayeredAccessMode
+{
+    public ElevatedAccessMode( AccessMode originalMode,
+            AccessMode overriddenAccessMode )
```
Mats-SX: Could we call these `original` and `elevating` or `elevatedTo`?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/LayeredAccessMode.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public AccessMode getSnapshot()
+    {
+        return AccessModeSnapshot.create( this );
+    }
+
+    @Override
+    public boolean isOverridden()
```
Mats-SX: Isn't it a bit strange to find that `ElevatedAccessMode.isOverridden()` exists and is true? It's not overridden, really.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/LayeredAccessMode.java
```diff
@@ -0,0 +1,65 @@
...
+
+abstract class LayeredAccessMode implements AccessMode
+{
+    protected final AccessMode originalMode;
+    protected final AccessMode overriddenMode;
+
+    public LayeredAccessMode( AccessMode originalMode, AccessMode overriddenMode )
+    {
+        this.originalMode = originalMode;
+        this.overriddenMode = overriddenMode;
```
Mats-SX: I guess the field naming comments apply the most to this class. Not sure what a good common word is for the two subclasses though. Perhaps one of `layered`, `additional`, `other`?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/AuthScenariosInteractionTestBase.java
```diff
@@ -536,7 +482,7 @@ public void customRoleWithProcedureAccess() throws Exception
         assertEmpty( adminSubject, "CALL dbms.security.createRole('role1')" );
         testFailTestProcs( mats );
         assertEmpty( adminSubject, "CALL dbms.security.addRoleToUser('role1', 'mats')" );
-        testSuccessfulTestProcs( mats );
+        //testSuccessfulTestProcs( mats ); // TODO: FIXME
```
Mats-SX: Leftover TODO

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/EmbeddedInteraction.java
```diff
@@ -170,6 +171,13 @@ public void assertInitFailed( EnterpriseAuthSubject subject )
     }
 
     @Override
+    public void assertSessionKilled( EnterpriseAuthSubject subject )
+    {
+        assertThat( executeQuery( subject, "MATCH (n:Node) RETURN count(n)", null, r -> {} ),
+                containsString( "Read operations are not allowed" ) );
```
Mats-SX: Why is this parametrised with a subject? This assertion will give a false negative (as in: not throwing when it should) if the parameter subject doesn't have read privileges, right?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ProcedureInteractionTestBase.java
```diff
@@ -559,28 +543,67 @@ TransportConnection startBoltSession( String username, String password ) throws
             return Stream.of( new AuthProceduresBase.StringResult( "static" ) );
         }
 
-        @Procedure( name = "test.allowedProcedure1", allowed = {"role1"}, mode = Mode.READ )
+        @Procedure( name = "test.allowedReadProcedure", allowed = {"role1"}, mode = Mode.READ )
```
Mats-SX, line 551: Nice that you renamed these.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/integration/bolt/LdapAuthenticationIT.java
```diff
@@ -323,15 +321,38 @@ public void shouldFailIfAuthorizationExpiredWithUserLdapContext() throws Throwab
...
+    public void shouldSucceedIfAuthorizationExpiredWithinTransactionWithUserLdapContext() throws Throwable
+    {
+        restartNeo4jServerWithOverriddenSettings( settings -> {
+            settings.put( SecuritySettings.ldap_authorization_use_system_account, "false" );
+        } );
+
         // Then
         assertAuth( "neo4j", "abc123" );
 
+        // We need to replace this with a two thread test.
```
Mats-SX: Should we have a card for it, or fix it on this PR?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
henriknyman: Yes, we only elevate access on the outermost procedure call. Once you got your elevated access rights, they can only be restricted by nested procedure calls. In the case of a WRITE procedure calling a READ procedure, it will get restricted to READ in the nested call. To be able to call the outer WRITE procedure you either need WRITE permissions or to be allowed WRITE permissions by annotation, which then implies READ permissions. Since the procedure modes have a strict order of which mode implies another, this should be sufficient.

Mats-SX (changes requested):
Lots of comments, but most are about naming. I like the refactoring into `LayeredAccessMode`, but it has some fields/methods that may need amending for clarification of the new purpose.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/AuthScenariosInteractionTestBase.java
henriknyman: We should fix this

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/integration/bolt/LdapAuthenticationIT.java
henriknyman: It was just a leftover obsolete comment that should be removed

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/LayeredAccessMode.java
henriknyman: No, the meaning of `isOverridden()` is that it is either restricted or elevated, i.e. you have an access mode that overrides the original access mode.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/LayeredAccessMode.java
henriknyman: overridden here just means either restricted or elevated, it is overriding the original access mode in some way.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/EmbeddedInteraction.java
henriknyman: The subject is used in `BoltInteraction` to get the transport connection. The asserts in REST and embedded doesn't make much sense since there is no session that could be killed. Instead they basically check that something happened so that the subject now doesn't have any permissions (e.g. user was deleted or suspended). This is an old remnant of the unifying of the tests across different surfaces.

community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/BoltResponseMessageHandler.java
```diff
@@ -41,4 +41,5 @@
 
     void onFailure( Status status, String message ) throws E;
 
+    default void onFatal( Status status, String message ) throws E { onFailure( status, message ); }
```
MishaDemianenko, line 44: I find this super confusing: it's default implementation that will delegate to `onFailure` handler silently while in the code we try to do distinction:

```
            if ( error.isFatal() )
            {
                out.onFatal( error.status(), error.message() );
            }
            else
            {
                out.onFailure( error.status(), error.message() );
            }
```

that kind of have the idea that we will it's important to distinguish those cases

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/Neo4jError.java
```diff
@@ -148,29 +150,53 @@ public static Status codeFromString( String codeStr )
...
             if ( cause instanceof Status.HasStatus )
             {
-                return new Neo4jError( ((Status.HasStatus) cause).status(), any.getMessage(), any );
+                return new Neo4jError( ((Status.HasStatus) cause).status(), any.getMessage(), any, isFatal );
             }
 
             if (cause instanceof OutOfMemoryError)
             {
-                return new Neo4jError( Status.General.OutOfMemoryError, cause );
+                return new Neo4jError( Status.General.OutOfMemoryError, cause, isFatal );
```
MishaDemianenko, line 164: In case of `OutOfMemoryError` or `StackOverflowError` we will pass `cause` as cause instead of source `any`. Why it's like that? Would it make sense to always see full exception?

community/kernel/src/main/java/org/neo4j/kernel/api/security/AccessMode.java
```diff
@@ -228,142 +175,56 @@ public boolean allowsSchemaWrites()
...
 
     boolean allowsReads();
     boolean allowsWrites();
     boolean allowsSchemaWrites();
-    boolean overrideOriginalMode();
+
     AuthorizationViolationException onViolation( String msg );
     String name();
+
+    default String username()
```
MishaDemianenko, line 200: I think we missing some javadocs here: how access mode is connected with username and what username we expecting here

community/kernel/src/main/java/org/neo4j/kernel/api/security/AccessMode.java
```diff
@@ -228,142 +175,56 @@ public boolean allowsSchemaWrites()
...
+    {
+        return ""; // Should never clash with a valid username
+    }
+
+    default AccessMode getOriginalAccessMode()
```
MishaDemianenko, line 205: i find weird combinations of default and non default methods that implemented in 
accessMode and Static. I would suggest we remove all the defaults since that will force users override required methods instead of spend hours in debug

community/kernel/src/main/java/org/neo4j/kernel/api/ProcedureCallOperations.java
MishaDemianenko, line 97: This interface javadoc describe current interface as: `Specifies procedure call operations for the three types of procedure calls that can be made.`
why functions was moved here? if we distinguish them they should not be part of this class and corresponding interface

community/kernel/src/main/java/org/neo4j/kernel/api/ProcedureCallOperations.java
```diff
@@ -40,15 +39,14 @@
             throws ProcedureException;
 
     /**
-     * Invoke a read-only procedure by name, and override the transaction's access mode with
+     * Invoke a read-only procedure by name, and elevate the transaction's access mode to
```
MishaDemianenko: All java docs claim that access mode will be elevated to the given and given is gone now

community/kernel/src/main/java/org/neo4j/kernel/impl/api/dbms/NonTransactionalDbmsOperations.java
```diff
@@ -47,9 +47,11 @@ public NonTransactionalDbmsOperations( Procedures procedures )
     ) throws ProcedureException
     {
         BasicContext ctx = new BasicContext();
-        if ( mode instanceof AuthSubject )
+
+        AccessMode originalMode = mode.getOriginalAccessMode();
```
MishaDemianenko, line 51: why do we use original mode here? what if we have situation:
original mode (full) -elevated-> mode(read) -try to elevate -> mode(write).
Original mode will allow it, but that mode was restricted by nested procedure. And now procedure nested in nested can actually get more higher access mode then outer.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/AccessModeSnapshot.java
```diff
@@ -0,0 +1,103 @@
...
+    private final boolean allowsWrites;
+    private final boolean allowsSchemaWrites;
+
+    private final AccessMode originalMode;
+
+    public static AccessMode create( AccessMode accessMode )
+    {
+        // TODO: Use flyweight pattern instead of always creating new objects, when we have a proper
+        //       security context and do not need to obtain the original mode through this object
+        return new AccessModeSnapshot( accessMode );
```
MishaDemianenko, line 37: is it not the case already now that our accessMode objects are final and can't be modified?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/ElevatedAccessMode.java
```diff
@@ -0,0 +1,55 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api.security;
+
+import org.neo4j.kernel.api.security.AccessMode;
+
+public class ElevatedAccessMode extends LayeredAccessMode
+{
+    public ElevatedAccessMode( AccessMode originalMode,
```
MishaDemianenko: since we elevate access mode to a new one should be the case that we should only use elevatedMode to check what we can do now?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/ElevatedAccessMode.java
```diff
@@ -0,0 +1,55 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api.security;
+
+import org.neo4j.kernel.api.security.AccessMode;
+
+public class ElevatedAccessMode extends LayeredAccessMode
```
MishaDemianenko: Javadocs missing

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/LayeredAccessMode.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api.security;
+
+import org.neo4j.graphdb.security.AuthorizationViolationException;
+import org.neo4j.kernel.api.security.AccessMode;
+
+abstract class LayeredAccessMode implements AccessMode
```
MishaDemianenko: java docs is missing; what layers are we talking about here?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/RestrictedAccessMode.java
```diff
@@ -0,0 +1,55 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api.security;
+
+import org.neo4j.kernel.api.security.AccessMode;
+
+public class RestrictedAccessMode extends LayeredAccessMode
+{
+    public RestrictedAccessMode( AccessMode originalMode,
```
MishaDemianenko: have the same question as above here if we restrict our mode to some restrictive mode should we just simply use that mode as a source of mode truth?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
```diff
@@ -225,8 +223,8 @@ final void forceClose()
 
     final Optional<String> username()
     {
-        AccessMode mode = transaction.mode();
-        return ( mode instanceof AuthSubject ) ? Optional.of( ((AuthSubject) mode).username() ) : Optional.empty();
+        String username = transaction.mode().username();
```
MishaDemianenko: since mode is always present now, does it make sense still to have Optional result?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -230,7 +229,7 @@ public KernelTransactionImplementation initialize(
         this.lastTransactionTimestampWhenStarted = lastTimeStamp;
         this.transactionEvent = tracer.beginTransaction();
         assert transactionEvent != null : "transactionEvent was null!";
-        this.accessMode = accessMode;
+        this.accessMode = accessMode.getSnapshot();
```
MishaDemianenko, line 232: the same question as before: accessMode should be non modifiable why do we need to snapshot it?

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
```diff
@@ -221,4 +224,22 @@ public void clearAuthCache()
...
+        List<AuthorizationInfo> infoList = new ArrayList<>( 1 );
+        for ( Realm realm : realms )
+        {
+            if ( realm instanceof ShiroAuthorizationInfoProvider )
+            {
+                AuthorizationInfo info = ((ShiroAuthorizationInfoProvider) realm)
+                        .getAuthorizationInfoSnapshot( principalCollection );
+                if ( info != null )
+                {
+                    infoList.add( info );
```
MishaDemianenko, line 239: If we really expect only one instance of AuthorizationInfo to be added into the list we should stop iteration over loop here, otherwise 1 in list constructor is redundant

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ProcedureInteractionTestBase.java
```diff
@@ -357,8 +335,14 @@ void assertPasswordChangeWhenPasswordChangeRequired( S subject, String newPasswo
     void assertFail( S subject, String call, String partOfErrorMsg )
     {
         String err = assertCallEmpty( subject, call );
-        assertThat( err, not( equalTo( "" ) ) );
-        assertThat( err, containsString( partOfErrorMsg ) );
+        if ( partOfErrorMsg == null || partOfErrorMsg == "" )
```
MishaDemianenko: StringUtils.isEmpty( partOfErrorMsg )

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ProcedureInteractionTestBase.java
```diff
@@ -357,8 +335,14 @@ void assertPasswordChangeWhenPasswordChangeRequired( S subject, String newPasswo
...
+        {
+            assertThat( err, not( equalTo( "" ) ) );
```
MishaDemianenko: and since we know it's null or empty here, why just not to use `nullValue` here instead of negation?

MishaDemianenko (commented):
I finish my review round, i think some of the comments popup just because i'm do not really know how that api should be used and in which way, so they are more questions then actual comments. Looks good in general!

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/EmbeddedInteraction.java
Mats-SX: Makes sense.

Mats-SX (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/api/dbms/NonTransactionalDbmsOperations.java
henriknyman: We are only using it to get hold of the `AuthSubject` through a downcast. The only permission checks done in dbms procedures are done directly on auth subject, like `isAdmin()` or has a username.
We have a refactoring planned of separating access mode and auth subject into a security context, but right now you get auth subject through the reference of the access mode.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/AccessModeSnapshot.java
henriknyman: No, because the access mode could be an `AuthSubject` (it almost always is), and we need to be able to get hold of it by a downcast of the original access mode (until we can do the refactoring mentioned above). I guess we could still do a flyweight including the original access mode reference, but it would have much more combinations so it didn't seem worthwhile.

fickludd: well no, in the current architecture the `StandardEnterpriseAuthSubject` (which extends AccessMode) dynamically checks permissions on each allows\* call. This PR is about freezing these permissions in a snapshot on transaction start.

community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/BoltResponseMessageHandler.java
henriknyman: The default implementation was just to avoid having to modify a lot of test files. But in this case I guess it makes sense to remove the default. In the other classes below I would prefer to keep the default implementations to keep the change set down.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/Neo4jError.java
henriknyman: I don't know

community/kernel/src/main/java/org/neo4j/kernel/api/security/AccessMode.java
Mats-SX: We plan on separating the concepts of AuthSubject and AccessMode soon, so we rather make the docs effort then.

community/kernel/src/main/java/org/neo4j/kernel/api/security/AccessMode.java
Mats-SX: Nearly all implementations will do `return this` here. On this instance, we like the code sharing.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ProcedureInteractionTestBase.java
henriknyman: Hmm, yes we had that first, but encountered some weird runtime dependency issue (that I cannot remember exactly what it was now), and didn't feel we wanted to spend more time on it.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/ElevatedAccessMode.java
Mats-SX: Fixed!

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/ElevatedAccessMode.java
Mats-SX: Good catch! Fixed.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/LayeredAccessMode.java
Mats-SX: Renamed and fixed.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/RestrictedAccessMode.java
Mats-SX: Given that we do explicit checks on the original mode before overriding, yes. Change made.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
Mats-SX: Yes, you're right. Changed.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
Mats-SX: See above; it is not immutable generally.

PR #8116, opened 2016-10-09
Avoid neo4j crash when indexing long string

mz24cn:
When string property byte array length is bigger than 32768, lucene will throw Exception due to LUCENE-5472 improvement, which leads neo4j engine stopped to response. This modification change StringField to TextField, and BinaryDocValuesField instead of SortedSetDocValuesField when longer than 10240 (for UTF8 encoding, the String bytes length are close to 10922=32766/3). It is only a workaround solution.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/IndexType.java
+++ b/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/IndexType.java
@@ -140,7 +140,7 @@ public abstract class IndexType
         {
             // TODO We should honor ValueContext instead of doing value.toString() here.
             // if changing it, also change #get to honor ValueContext.
-            document.add( new StringField( exactKey( key ), value.toString(), Store.YES ) );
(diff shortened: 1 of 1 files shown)


mz24cn:
commit the lucene legacy index related code

MishaDemianenko:
Hi @mz24cn,

Thanks a lot for your contribution.

First of all, if you want your code to be part of Neo4j you need to sign our [CLA](https://neo4j.com/developer/cla/).

It's great that you highlight the problem with too long values for a string field, but i'm afraid we can't just change field types like that. 
For example: TextFields are tokenized, while StringFields are not.

Instead I would propose to add validation step that will prevent attempts to store too long values into a StringField(s).
Please do not forget that to confirm that changes are actually fixing something appropriate test cases should be part of PR.

Also please keep your code style as close as possible with one in the repo (comments, brackets, etc).

MishaDemianenko:
@mz24cn issue should be fixed now as part of #8404 and should be included in 3.2. Closing this PR.

PR #8117, opened 2016-10-09
Avoid neo4j crash when indexing long string

mz24cn:
When string property byte array length is longer than 32768, lucene will throw Exception due to LUCENE-5472 improvement, which leads neo4j engine stopped to response. This modification change StringField to TextField, and BinaryDocValuesField instead of SortedSetDocValuesField when longer than 10240 (for UTF8 encoding, the String bytes length are close to 10922=32766/3). It is only a workaround solution.


tinwelint:
@mz24cn Is this simply a duplicate of https://github.com/neo4j/neo4j/pull/8116 ?

MishaDemianenko:
@mz24cn issue should be fixed now as part of #8404 and should be included in 3.2. Closing this PR.

PR #8120, opened 2016-10-10
Bolt enable false when performing recovery on copied store

burqen:
**What?**
Turn of bolt when performing recovery on copied store.

**Why?**
Problems with store copy when bolt is turned on.
Having bolt enabled would cause the "certificates" folder to be created in the "temp-copy" directory. This is the directory on slave to which we write the store that we copy from master before replacing our current store. Then when we move the copied store to replace our current store we suddenly have two "certificates" folders and we break down because we do not want to replace our existing certificates folder.
This will cause slave to retry and fail until TimeoutException.

Test:
New test is identical to TestBranchedData, but with BoltKernelExtension in the class path which automatically loads bolt if not turned off.

### DIFF
--- a/enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyClient.java
+++ b/enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyClient.java
@@ -330,6 +330,7 @@ public class StoreCopyClient
                 .setConfig( GraphDatabaseSettings.keep_logical_logs, Settings.TRUE )
                 .setConfig( GraphDatabaseSettings.allow_store_upgrade,
                         config.get( GraphDatabaseSettings.allow_store_upgrade ).toString() )
+                .setConfig( GraphDatabaseSettings.boltConnector( "bolt" ).enabled, "false" )
                 .newGraphDatabase();
     }
 
--- a/integrationtests/src/test/java/org/neo4j/ha/BranchedDataWithBoltIT.java
+++ b/integrationtests/src/test/java/org/neo4j/ha/BranchedDataWithBoltIT.java
@@ -1,3 +1,22 @@
+/*
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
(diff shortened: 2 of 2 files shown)


integrationtests/src/test/java/org/neo4j/ha/BranchedDataWithBoltIT.java
```diff
@@ -0,0 +1,76 @@
+package org.neo4j.ha;
```
chrisvest, line 20: Get You a License Header for Great Good!

chrisvest:
Looks good. You can rebase if you want a green build, I'm told.

burqen:
Rebased

enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyClient.java
```diff
@@ -330,6 +330,7 @@ private GraphDatabaseService newTempDatabase( File tempStore )
                 .setConfig( GraphDatabaseSettings.keep_logical_logs, Settings.TRUE )
                 .setConfig( GraphDatabaseSettings.allow_store_upgrade,
                         config.get( GraphDatabaseSettings.allow_store_upgrade ).toString() )
+                .setConfig( GraphDatabaseSettings.boltConnector( "bolt" ).enabled, "false" )
```
MishaDemianenko, line 333: we can use `Settings.FALSE` here instead

integrationtests/src/test/java/org/neo4j/ha/BranchedDataWithBoltIT.java
```diff
@@ -0,0 +1,95 @@
...
+public class BranchedDataWithBoltIT
+{
+    @Rule
+    public TestDirectory dir = TestDirectory.testDirectory( getClass() );
+
+    @Test
+    public void mustHandleBranchedDataWithBoltAvailable() throws Throwable
+    {
+        // GIVEN
+        ClusterManager clusterManager = new ClusterManager.Builder( dir.cleanDirectory( "dbs" ) )
```
MishaDemianenko, line 48: can we please use cluster rule to build this cluster? that will make test code cleaner and will make this test ready for other cluster tests, and failures handling of cause

chrisvest:
@burqen Misha has good comments. The changes otherwise look fine.

burqen:
Problem already fixed by https://github.com/neo4j/neo4j/pull/8107
Bolt is now turned off for by default for Embedded databases and thus does not create this problem.
Closing.

PR #8121, opened 2016-10-10
Use a sensible timeout for assertion

spacecowboy:
Fixing a flaky test.

If the method itself can run for up to `timeout`, then the check for
that method should wait for a `time >= timeout`.

### DIFF
--- a/community/kernel/src/test/java/org/neo4j/kernel/builtinprocs/AwaitIndexProcedureTest.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/builtinprocs/AwaitIndexProcedureTest.java
@@ -186,7 +186,7 @@ public class AwaitIndexProcedureTest
 
         state.set( ONLINE );
         assertEventually( "Procedure did not return after index was online",
-                done::get, is( true ), 5, TimeUnit.MILLISECONDS );
+                done::get, is( true ), 10, TimeUnit.SECONDS );
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


community/kernel/src/test/java/org/neo4j/kernel/builtinprocs/AwaitIndexProcedureTest.java
```diff
@@ -186,7 +186,7 @@ public InternalIndexState answer( InvocationOnMock invocationOnMock ) throws Thr
 
         state.set( ONLINE );
         assertEventually( "Procedure did not return after index was online",
-                done::get, is( true ), 5, TimeUnit.MILLISECONDS );
+                done::get, is( true ), timeout * 2, TimeUnit.MILLISECONDS );
```
benbc: The procedure wait timeout is not reached in this test. There is a separate test which does indeed double the timeout for the wait.

However I agree that the problem here is that 5 ms just isn't long enough -- I should never have made it that low in the first place. I don't think we should express it in terms of multiples of `timeout` because the two timeouts are not conceptually related.

I'd set it to something _extremely_ high. Something ludicrous like 10 seconds. It's never going to be hit unless there is a (new) bug in the code, so there is no reason to keep it small.

The same logic applies to the timeout test following, actually -- where it is related to the timeout. Since you're here will you increase that to something much bigger too?

spacecowboy: Done

benbc:
LGTM

PR #8122, opened 2016-10-10
Restore should respect store lock

spacecowboy:
cl: Fixed `neo4j-admin restore` so that `--force` does not overwrite a running database


benbc:
LGTM

PR #8124, opened 2016-10-10
Improved `http.log` output in case of nulls

spacecowboy:
Tries to replicate prior art and prints empty string instead "null" in
case of uninitialized values.

### DIFF
--- a/community/server/src/main/java/org/neo4j/server/web/AsyncRequestLog.java
+++ b/community/server/src/main/java/org/neo4j/server/web/AsyncRequestLog.java
@@ -43,6 +43,8 @@ import org.neo4j.logging.RotatingFileOutputStreamSupplier;
 import org.neo4j.logging.async.AsyncLogEvent;
 import org.neo4j.logging.async.AsyncLogProvider;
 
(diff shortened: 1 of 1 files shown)


community/server/src/main/java/org/neo4j/server/web/AsyncRequestLog.java
```diff
@@ -83,7 +83,24 @@ public void log( Request request, Response response )
...
+                now,
+                emptyIfNull( requestURL ),
+                statusCode,
+                length,
+                emptyIfNull( referer ),
+                emptyIfNull( userAgent ),
+                serviceTime );
+    }
+
+    private String emptyIfNull( String text )
```
lutovich: `StringUtils#defaultString()` can be used instead.

lutovich (commented):
@spacecowboy change look fine but it would be nice to have a test for it to avoid regressions. Also is it fine to change log format in a patch release?

spacecowboy:
@lutovich I wanted to argue that the format hasn't changed, but people might depend on it anyway so I will re-target the PR to 3.1. Thanks.

PR #8127, opened 2016-10-10
Bumped version of Neo4j Browser to 1.1.9

pe4cey:
changelog [3.0]


spacecowboy:
@pe4cey you need to add a `changelog` or `cl` prefix. Fixed that for you.

But I see that you are using it like pros elsewhere :+1:

PR #8133, opened 2016-10-11
Traces calls to free() in EFSA

tinwelint:
and includes that stack trace as cause in exception thrown when accessing
freed buffer later on. This to more easily track down bugs in tests.

### DIFF
--- a/community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
+++ b/community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
@@ -1152,7 +1152,7 @@ public class EphemeralFileSystemAbstraction implements FileSystemAbstraction
 
         synchronized DynamicByteBuffer copy()
         {
-            return new DynamicByteBuffer( buf ); // invoke "copy constructor"
+            return new DynamicByteBuffer( buf() ); // invoke "copy constructor"
         }
 
         static
@@ -1207,7 +1207,6 @@ public class EphemeralFileSystemAbstraction implements FileSystemAbstraction
         void free()
         {
             assertNotFreed();
-
             try
             {
                 clear();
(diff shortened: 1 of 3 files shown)


community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
```diff
@@ -1199,18 +1206,22 @@ private ByteBuffer allocate( int sizeIndex )
...
+            assertNotFreed();
+
             try
             {
                 clear();
             }
             finally
             {
                 buf = null;
+                freeCall = new Exception();
```
chrisvest: I'd add a descriptive message to this exception so it's easy to recognise in the stack traces what these stack frames mean.

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
```diff
@@ -1199,18 +1206,22 @@ private ByteBuffer allocate( int sizeIndex )
...
             }
         }
 
         synchronized void put( int pos, byte[] bytes, int offset, int length )
         {
+            assertNotFreed();
             verifySize( pos + length );
```
chrisvest, line 1225: You can change these explicit asserts into `buf()` calls that stores the result to a local variable.

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
```diff
@@ -1269,11 +1283,24 @@ private void verifySize( int totalAmount )
...
         }
 
+        private void assertNotFreed()
+        {
+            if ( this.buf == null )
+            {
+                String message = "This buffer have been freed";
+                throw this.freeCall != null
+                        ? new IllegalStateException( message, freeCall )
+                        : new IllegalStateException( message );
```
chrisvest: It's perfectly legal for the cause of an exception to be `null`, so this code can be simplified.

chrisvest (commented):
Had a few comments but generally looks good.

chrisvest (approved):

PR #8135, opened 2016-10-11
Fixes a race condition creating input groups in importer

tinwelint:
Now that input may be processed in parallel, different groups of inputs
which have different groups may request to create new groups concurrently.
Groups class wasn't thread safe since there was previously no requirement
for it.

This commit makes Groups#getOrCreate call synchronized to handle concurrent
invocation. It's typically invoked quite rarely so the locking nature of it
is fine.

cl [tools] Fixed a race condition in `neoj4-import` with ID groups which would have some correct relationships considered invalid and logged into bad.log instead

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/InputGroupsDeserializer.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/InputGroupsDeserializer.java
@@ -60,7 +60,7 @@ class InputGroupsDeserializer<ENTITY extends InputEntity>
     }
 
     InputGroupsDeserializer( Iterator<DataFactory<ENTITY>> dataFactory, Header.Factory headerFactory,
-            Configuration config, IdType idType, int maxProcessors, DeserializerFactory<ENTITY> factory,
+            Configuration config, IdType idType, int maxProcessors, int processors, DeserializerFactory<ENTITY> factory,
             Validator<ENTITY> validator, Class<ENTITY> entityClass )
     {
         super( dataFactory );
@@ -68,6 +68,7 @@ class InputGroupsDeserializer<ENTITY extends InputEntity>
         this.config = config;
         this.idType = idType;
         this.maxProcessors = maxProcessors;
+        this.previousInputProcessors = processors;
         this.factory = factory;
         this.validator = validator;
         this.entityClass = entityClass;
@@ -111,8 +112,7 @@ class InputGroupsDeserializer<ENTITY extends InputEntity>
             // complete rows of data and can be parsed individually by multiple threads.
 
             currentInput = new ParallelInputEntityDeserializer<>( data, headerFactory, config, idType,
-                    maxProcessors, factory, validator, entityClass );
-            currentInput.processors( previousInputProcessors );
+                    maxProcessors, previousInputProcessors, factory, validator, entityClass );
             currentInputOpen = true;
         }
 
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/InputGroupsDeserializer.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/InputGroupsDeserializer.java
@@ -45,7 +45,7 @@ class InputGroupsDeserializer<ENTITY extends InputEntity>
     private final IdType idType;
     private InputIterator<ENTITY> currentInput = new InputIterator.Empty<>();
     private long previousInputsCollectivePositions;
-    private int previousInputProcessors = 1;
+    private int previousInputProcessors;
     private boolean currentInputOpen;
     private final int maxProcessors;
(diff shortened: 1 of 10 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/InputGroupsDeserializer.java
```diff
@@ -60,14 +60,15 @@
...
-            Configuration config, IdType idType, int maxProcessors, DeserializerFactory<ENTITY> factory,
+            Configuration config, IdType idType, int maxProcessors, int processors, DeserializerFactory<ENTITY> factory,
             Validator<ENTITY> validator, Class<ENTITY> entityClass )
     {
         super( dataFactory );
         this.headerFactory = headerFactory;
         this.config = config;
         this.idType = idType;
         this.maxProcessors = maxProcessors;
+        this.previousInputProcessors = processors;
```
burqen: Can remove initialisation from declaration if initialised here.

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/input/csv/InputGroupsDeserializerTest.java
```diff
@@ -80,7 +84,52 @@ public void shouldBeAbleToAskForSourceInformationEvenBetweenTwoSources() throws
...
+        Collector badCollector = mock( Collector.class );
+        Configuration config = lowBufferSize( COMMAS, false );
+        try ( InputGroupsDeserializer<InputNode> deserializer = new InputGroupsDeserializer<>(
+                data.iterator(), defaultFormatNodeFileHeader(), config, idType,
+                processors, processors, (header,stream,decorator,validator) ->
+                {
+                    InputNodeDeserialization deserialization =
+                            new InputNodeDeserialization( header, stream, groups, idType.idsAreExternal() );
+                    return new InputEntityDeserializer<>( header, stream, config.delimiter(),
+                            deserialization, decorator, validator, badCollector );
```
burqen: This is the same lamba used in `CsvInput#nodes()` and almost the same as `CsvInput#relationships()`.
I think readability would be improved by extracting them to concrete classes.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/InputGroupsDeserializer.java
```diff
@@ -60,14 +60,15 @@
     }
 
     InputGroupsDeserializer( Iterator<DataFactory<ENTITY>> dataFactory, Header.Factory headerFactory,
-            Configuration config, IdType idType, int maxProcessors, DeserializerFactory<ENTITY> factory,
+            Configuration config, IdType idType, int maxProcessors, int processors, DeserializerFactory<ENTITY> factory,
```
burqen: See over all comment about processors.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/ParallelInputEntityDeserializer.java
```diff
@@ -68,7 +68,7 @@
 
     @SuppressWarnings( "unchecked" )
     public ParallelInputEntityDeserializer( Data<ENTITY> data, Header.Factory headerFactory, Configuration config,
-            IdType idType, int maxProcessors, DeserializerFactory<ENTITY> factory,
+            IdType idType, int maxProcessors, int processors, DeserializerFactory<ENTITY> factory,
```
burqen: See over all comment about processors.

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/input/GroupsTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.Test;
+
+import org.neo4j.test.Race;
+
+import static org.junit.Assert.assertEquals;
+
+public class GroupsTest
+{
+    @Test
+    public void shouldHandleConcurrentGetOrCreate() throws Throwable
```
burqen, line 31: Nice

burqen (changes requested):
Behaviour change looks good.

I'm not entirely sure about why we need to add processors argument to any of the `Deserializers`. Is it only for test purposes? Seems unrelated to the problem that this PR is addressing.

Otherwise, small clean up comments.

tinwelint:
@burqen the change simplified tests. The way these classes are used currently there's really no need to specify initial processor count, currently. At the same time this could be seen as a bug because an invocation on `processors(int)` would have no effect before the first call to `hasNext()`, of course we wouldn't need to add this constructor argument for this, we could have tests call `processors(int)` after construction (which they previously not quite couldn't as easily, with the first `hasNext()` thing), but since I was changing things around here and tests looked unnecessarily complicated I though I'd just add it.

burqen (approved):

PR #8137, opened 2016-10-11
Reverts previous move of bad.log during import

tinwelint:
since it's a surface change. This could instead be done in a major release.


chrisvest:
@tinwelint So, do we null-merge this?

PR #8138, opened 2016-10-11
Heap size no longer hard codes unit to Mb

spacecowboy:
The problem with this is that it provides a non-uniform interface
compared to the page cache settings (which allow the user to specify the
unit, and defaults to bytes). It is also at variance with what
experienced Java users will expect.

The particular motivation for fixing this is that the Docker image
configuration exposes two memory-related settings, one of which takes a
unit and the other doesn't.

To preserve backwards-compatibility, we still assume that the default
unit is megabytes.

Powershell tests can be run with this command:

``` shell
docker run --rm --name=powershell -it -v /path/to/standalone:/standalone centos/powershell /usr/bin/powershell /standalone/src/tests/Neo4j-Management/unit/Get-Java.Tests.ps1
```
## TODO
- [x] Should still treat no suffix as megabytes, for backward compatibility
- [x] Make sure both shell script and windows handle missing suffix correctly
- [x] Default value should not be `512`, it should be `512m`

changelog: Heap size config setting is no longer hardcoded to Mb

It is now possible to use the standard java suffixes as a unit `k|K|m|M|g|G`. For backwards-compatibility reasons, if no unit is specified we will still assume you mean Megabytes.

### DIFF
--- a/manual/kernel/src/docs/ops/performance-guide.asciidoc
+++ b/manual/kernel/src/docs/ops/performance-guide.asciidoc
@@ -51,9 +51,9 @@ For this reason, the heap should ideally be sized and tuned such that transactio
 When using Neo4j Server, JVM configuration goes into the _conf/neo4j.conf_ file.
 --
 
-In server deployments, the heap size is configured with the `dbms.memory.heap.max_size` (in MBs) setting in the _neo4j.conf_ file.
-For embedded, the heap size is specified by giving the `-Xmx???m` command line flag to the `java` process, where the `???` is the maximum heap size in MBs.
-The initial size of the heap is specified by the `dbms.memory.heap.initial_size` setting, or with the `-Xms???m` flag, or chosen heuristically by the JVM itself if left unspecified.
+In server deployments, the heap size is configured with the `dbms.memory.heap.max_size` setting in the _neo4j.conf_ file.
+For embedded, the heap size is specified by giving the `-Xmx???` command line flag to the `java` process, where the `???` is the maximum heap sizes.
+The initial size of the heap is specified by the `dbms.memory.heap.initial_size` setting, or with the `-Xms???` flag, or chosen heuristically by the JVM itself if left unspecified.
 The JVM will automatically grow the heap as needed, up to the maximum size.
 The growing of the heap requires a full GC cycle.
 If you know that you will need all the heap memory, you can set the initial heap size and the maximum heap size to the same value.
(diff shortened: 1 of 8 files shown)


manual/kernel/src/docs/ops/performance-guide.asciidoc
```diff
@@ -51,9 +51,9 @@ For this reason, the heap should ideally be sized and tuned such that transactio
 When using Neo4j Server, JVM configuration goes into the _conf/neo4j.conf_ file.
 --
 
-In server deployments, the heap size is configured with the `dbms.memory.heap.max_size` (in MBs) setting in the _neo4j.conf_ file.
-For embedded, the heap size is specified by giving the `-Xmx???m` command line flag to the `java` process, where the `???` is the maximum heap size in MBs.
-The initial size of the heap is specified by the `dbms.memory.heap.initial_size` setting, or with the `-Xms???m` flag, or chosen heuristically by the JVM itself if left unspecified.
+In server deployments, the heap size is configured with the `dbms.memory.heap.max_size` setting in the _neo4j.conf_ file.
+For embedded, the heap size is specified by giving the `-Xmx???` command line flag to the `java` process, where the `???` is the maximum heap size in MBs.
```
benbc: "in MBs" in this line should go (but I think this doc page is obsolete)

packaging/standalone/src/main/distribution/shell-scripts/bin/neo4j-shared.sh
```diff
@@ -81,6 +81,14 @@ resolve_path() {
     echo "${filename}"
 }
 
+# Ensure that the string ends with a unit [g|G|m|M|k|K] which Java
+# recognizes. Defaults to megabyte if no unit specified.
+ensure_memory_unit() {
+  local mem=$1
+  [[ ${mem} =~ .*[gGmMkK] ]] || mem="${mem}m"
+  echo "${mem}"
```
benbc: I think this adds an unwanted newline, doesn't it?

benbc (commented):
Great. Thank you. You probably need to update the manual as well.

packaging/standalone/src/main/distribution/shell-scripts/bin/neo4j-shared.sh
spacecowboy: Fixed

benbc:
LTGM

PR #8140, opened 2016-10-11
Handle packing and unpacking of maps with null key

pontusmelke:
Bolt didn't properly handle maps containing `null` as key, but failed both on
packing and unpacking.


boggle:
This is a great find; let's pause a second though to discuss what semantics we want from Cypher. Let's perhaps jump on a quick call on this beginning next week. I'm adding "DONT_MERGE" for now.

systay:
We will fix this by giving a nicer error message. Null keys will not be allowed

PR #8151, opened 2016-10-13
Clean up old test threads

tinwelint:
which was left hanging around until JVM ended.
Even found some places in product which forgot to close things.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/util/DebugUtil.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/util/DebugUtil.java
@@ -111,9 +111,26 @@ public class DebugUtil
 
     public static void dumpThreads( PrintStream out )
     {
+        dumpThreads( out, true, true );
+    }
+
+    public static void dumpThreads( PrintStream out, boolean excludeSystemThreads, boolean excludeCallingThread )
+    {
         for ( Map.Entry<Thread,StackTraceElement[]> stack : Thread.getAllStackTraces().entrySet() )
         {
-            out.println( new DebugUtil.CallStack( stack.getValue(), "Thread: " + stack.getKey() ) );
+            Thread thread = stack.getKey();
+            ThreadGroup group = thread.getThreadGroup();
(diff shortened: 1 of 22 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/util/DebugUtil.java
```diff
@@ -111,9 +111,26 @@ public static boolean currentStackTraceContains( Predicate<StackTraceElement> pr
...
+            {
+                continue;
+            }
+            if ( excludeCallingThread && thread == Thread.currentThread() )
+            {
+                continue;
+            }
+
+            out.println( new CallStack( stack.getValue(), "Thread: " + stack.getKey() ) );
+            out.println();
```
chrisvest: We have something similar in `ThreadTestUtil`.

chrisvest:
The build failed to download some dependencies, so I've restarted it.

chrisvest:
@tinwelint When removing `DebugUtil#dumpThreads` you missed a usage site at `BackupServiceStressTestingBuilder.java:[238,30]`

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -122,7 +122,18 @@ public BatchingNeoStores( FileSystemAbstraction fileSystem, File storeDir, Recor
...
+            IllegalStateException ise =
+                    new IllegalStateException( storeDir + " already contains data, cannot do import here" );
+            try
+            {
+                pageCache.close();
+            }
+            catch ( IOException e )
+            {
+                // Oddly enough we can't close the page cache, how to communicate this? Here we add as suppressed
+                ise.addSuppressed( e );
```
chrisvest, line 134: Yeah, I think that's the right way to communicate this.

chrisvest:
Looks pretty good, otherwise.

tinwelint:
@chrisvest I (actually force-) pushed that one-line fix now

tinwelint:
@chrisvest could potentially be ready to be merged now

PR #8154, opened 2016-10-13
Specific handling of upgrade transaction during update pooling

MishaDemianenko:
During store upgrade in case if transaction logs are missing upgrade transaction checksum will be set to predefined constant value
(since we can't get find real value in the logs).
In HA setup upgraded store will be copied by slaves and update pullers will try to fetch updates starting from upgrade transaction id.
Master should be able to recognise that specific upgrade transaction id and use checksum that was assign to it during upgrade.

### DIFF
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigratorIT.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigratorIT.java
@@ -25,12 +25,15 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.function.Function;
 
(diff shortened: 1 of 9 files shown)


community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigratorIT.java
```diff
@@ -25,12 +25,15 @@
 import org.junit.runners.Parameterized;
 
 import java.io.File;
+import java.io.IOException;
```
digitalstain: Are these imports added by accident?

digitalstain (commented):
Apart from the imports, this looks good

community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigratorIT.java
MishaDemianenko: @digitalstain indeed, sorry for that and strangely intelij does not cleaned them up...
should be clean now

PR #8158, opened 2016-10-13
Use secure hasher in ldap realm

henriknyman:
Plugin realm was already using a new credentials hasher (based on Apache Shiro) that also does iterations and uses a secure random number generator, while the ldap realm was using the credentials hasher from the community edition. 

This unifies the implementation on the new hasher in enterprise edition.

### DIFF
--- a/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/LdapRealm.java
+++ b/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/LdapRealm.java
@@ -21,11 +21,12 @@ package org.neo4j.server.security.enterprise.auth;
 
 import org.apache.shiro.authc.AuthenticationInfo;
 import org.apache.shiro.authc.AuthenticationToken;
-import org.apache.shiro.authc.credential.HashedCredentialsMatcher;
+import org.apache.shiro.authc.credential.AllowAllCredentialsMatcher;
 import org.apache.shiro.authz.AuthorizationException;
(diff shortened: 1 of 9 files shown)


enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/LdapRealm.java
```diff
@@ -89,23 +89,29 @@
...
-        setAuthorizationCachingEnabled( true );
-        boolean authenticationCachingEnabled = config.get( SecuritySettings.ldap_authentication_cache_enabled );
-        setAuthenticationCachingEnabled( authenticationCachingEnabled );
+        if ( isAuthenticationCachingEnabled() )
+        {
+            setCredentialsMatcher( secureHasher.getHashedCredentialsMatcher() );
+        }
+        else
+        {
+            setCredentialsMatcher( new AllowAllCredentialsMatcher() );
```
Mats-SX, line 113: This looks like a change in behaviour. Are we allowing all credentials if we have caching turned off?

henriknyman: No, the LdapRealm fails authentication by throwing an exception if the bind to the ldap server fails, so it is only if caching is enabled that the credentials matcher is needed. But since the credentials matcher is always called it has to be disabled like this (this is actually the default in the JndiLdapRealm super class that this is extending).
The difference now is that previously we always hashed credentials even if it was not being cached, which was just unnecessary work.

PR #8163, opened 2016-10-13
Introduce SecurityContext

fickludd:
The `SecurityContext` separates `AuthSubject` and `AccessMode` from eachother, by using composition rather than inheritance. One benefit of this refactoring is the removal of numerous type checks and casts that we do to cast `AccessMode`s to `AuthSubject`s.


community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -229,7 +229,7 @@ public KernelTransactionImplementation initialize(
         this.lastTransactionTimestampWhenStarted = lastTimeStamp;
         this.transactionEvent = tracer.beginTransaction();
         assert transactionEvent != null : "transactionEvent was null!";
-        this.accessMode = accessMode.getSnapshot();
+        this.securityContext = securityContext.freeze();
```
henriknyman, line 232: It would be nicer if the transaction could keep the AccessMode member as it was, and do like this.accessMode = securityContext.getAccessModeSnapshot(). That would also minimize the necessary code changes necessary. If we need to get SecurityContext from transaction it can just be an additional member.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -796,11 +796,11 @@ public long getCommitTime()
     }
 
     @Override
-    public Revertable overrideWith( AccessMode mode )
+    public Revertable overrideWith( SecurityContext context )
```
henriknyman, line 799: ...then this could be unchanged (overrideWith( AccessMode mode ))

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -1494,59 +1495,67 @@ public void unregisterExecutingQuery( ExecutingQuery executingQuery )
...
         }
-        return callProcedure( name, input, new RestrictedAccessMode( tx.mode(), AccessMode.Static.FULL ) );
+        return callProcedure( name, input, new RestrictedAccessMode( tx.securityContext().mode(), AccessMode.Static.FULL ) );
     }
 
     @Override
     public RawIterator<Object[], ProcedureException> procedureCallSchemaOverride( QualifiedName name, Object[] input ) throws ProcedureException
     {
-        return callProcedure( name, input, new OverriddenAccessMode( tx.mode(), AccessMode.Static.FULL ) );
+        return callProcedure( name, input, new OverriddenAccessMode( tx.securityContext().mode(), AccessMode.Static.FULL ) );
```
henriknyman, line 1547: ...all these could remain unchanged tx.mode()

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -1597,7 +1609,7 @@ private Object callFunction(
     {
         statement.assertOpen();
 
-        try ( KernelTransaction.Revertable ignore = tx.overrideWith( mode ) )
+        try ( KernelTransaction.Revertable ignore = tx.overrideWith( tx.securityContext().freeze( mode ) ) )
```
henriknyman: What is freeze() doing here?

community/server/src/main/java/org/neo4j/server/rest/dbms/UserService.java
```diff
@@ -146,15 +146,11 @@ public Response setPassword( @PathParam("username") String username, @Context Ht
         return output.ok();
     }
 
-    private AuthSubject getSubjectFromPrincipal( Principal principal )
+    private SecurityContext getSecurityContextFromPrincipal( Principal principal )
```
henriknyman: This looks like a duplicate of the static method getSecurityContextFromUserPrincipal() above

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/PersonalUserManager.java
```diff
@@ -34,14 +35,16 @@
 
 class PersonalUserManager implements EnterpriseUserManager
 {
-    private final EnterpriseUserManager userManager;
-    private final AuthSubject authSubject;
-    private final SecurityLog securityLog;
+    private EnterpriseUserManager userManager;
+    private SecurityContext securityContext;
+    private AuthSubject authSubject;
+    private SecurityLog securityLog;
```
henriknyman: Why was final removed?

enterprise/security/src/main/resources/simplelogger.properties
```diff
@@ -0,0 +1,34 @@
...
+# Defaults to true.
+#org.slf4j.simpleLogger.showThreadName=true
+
+# Set to true if you want the Logger instance name to be included in output messages.
+# Defaults to true.
+#org.slf4j.simpleLogger.showLogName=true
+
+# Set to true if you want the last component of the name to be included in output messages.
+# Defaults to false.
+#org.slf4j.simpleLogger.showShortLogName=false
```
henriknyman: Was this added by mistake? We are not using ehcache..

community/kernel/src/main/java/org/neo4j/kernel/api/security/SecurityContext.java
```diff
@@ -0,0 +1,105 @@
...
+{
+    AccessMode mode();
+    AuthSubject subject();
+
+    SecurityContext freeze();
+    SecurityContext freeze( AccessMode mode );
+
+    default String defaultString( String name )
+    {
+        return String.format( "%s{ securityContext=%s, allowance=%s }", name, subject().username(), mode() );
```
henriknyman: allowance -> accessMode ?

henriknyman: securityContext -> username ?

henriknyman (changes requested):
Looks good in general

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/ExecutionEngine.scala
```diff
@@ -150,8 +150,8 @@ class ExecutionEngine(val queryService: GraphDatabaseQueryService, logProvider:
         val tc = externalTransactionalContext.getOrBeginNewIfClosed()
 
         // Temporarily change access mode during query planning
-        // NOTE: This will force read access even if the current transaction did not have it
-        val revertable = tc.restrictCurrentTransaction(AccessMode.Static.READ)
+        // NOTE: This will force read allowance if the current transaction did not have it
```
henriknyman: allowance -> access

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
```diff
@@ -590,8 +590,8 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
   private def shouldElevate(allowed: Array[String]): Boolean = {
     // We have to be careful with elevation, since we cannot elevate permissions in a nested procedure call
     // above the original allowed procedure mode. We enforce this by checking if mode is already an overridden mode.
-    val mode = transactionalContext.accessMode
-    allowed.nonEmpty && !mode.isOverridden && mode.getOriginalAccessMode.allowsProcedureWith(allowed)
+    val allowance = transactionalContext.securityContext.mode()
+    allowed.nonEmpty && !allowance.isOverridden && allowance.allowsProcedureWith(allowed)
```
henriknyman: allowance -> mode

community/server/src/main/java/org/neo4j/server/rest/dbms/UserService.java
fickludd: Removed this

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/PersonalUserManager.java
fickludd: Merge mistake. Fixed.

enterprise/security/src/main/resources/simplelogger.properties
fickludd: Oops, removed.

community/kernel/src/main/java/org/neo4j/kernel/api/security/SecurityContext.java
fickludd: yes

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
fickludd: It's just a way to associate the same subject with a new mode (within a SecurityContext). We'll change this to `withMode` to make it more clear.

henriknyman (commented):
Minor traces of allowance left.. :)

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/StandardEnterpriseSecurityContext.java
```diff
@@ -0,0 +1,228 @@
...
+    @Override
+    public AccessMode mode()
+    {
+        return new StandardAccessMode(
+                shiroSubject.isAuthenticated() && shiroSubject.isPermitted( READ ),
+                shiroSubject.isAuthenticated() && shiroSubject.isPermitted( READ_WRITE ),
+                shiroSubject.isAuthenticated() && shiroSubject.isPermitted( SCHEMA_READ_WRITE ),
+                shiroSubject.getAuthenticationResult() == AuthenticationResult.PASSWORD_CHANGE_REQUIRED,
+                authManager.getAuthorizationInfo( shiroSubject.getPrincipals() )
+            );
```
henriknyman: Will this ever be called outside except from freeze()  ? Maybe we could use a different name for the implementation method to make the code more clear.

henriknyman: I guess the place where we have the opportunity to use a flyweight has now moved to here, but now authorization info is also tied to it. I wonder if that should be generalized as a set of roles to make AccessMode more agnostic to Apache Shiro.

community/kernel/src/main/java/org/neo4j/kernel/api/security/AnonymousContext.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public SecurityContext freeze()
+    {
+        return this;
+    }
+
+    @Override
+    public SecurityContext freeze( AccessMode mode )
```
henriknyman: I guess you meant that this will be renamed to `withMode` ? That would make it much more clear.

henriknyman (commented):
After a second look I think my initial concern was more about the immutable design of security context snapshotting and if that would lead to more object creation than if access mode overriding would just mutate the access mode within the security context. I was imagining that the refactoring would open up the possibility that combined with a flyweight pattern for access modes, it would make it possible to have tight loops of procedure calls without generating much garbage. Maybe my concern is exaggerated, and there are other pros of an immutable design. It would be interesting to do some micro benchmarks on procedure calls though.

community/bolt/src/test/java/org/neo4j/bolt/security/auth/BasicAuthenticationTest.java
```diff
@@ -191,42 +208,19 @@ public void shouldThrowWithNoScheme() throws Exception
...
-        // Expect
-        exception.expect( AuthenticationException.class );
-        exception.expect( hasStatus( Status.Security.Unauthorized ) );
-
-        // When
-        authentication.authenticate( map( "scheme", "UNKNOWN", "principal", "bob", "credentials", "secret" ) );
-    }
-
-    @Test
-    public void shouldFailOnMalformedToken() throws Exception
```
craigtaverner: Why was this test removed?

boggle:
This looks good overall. Thanks for picking this up! 

I'd like to suggest to add more extensive docs to `SecurityContext` (class level and methods), possibly to `AccessMode` as well (didn't see that one pass by). Naming wise, isn't `AccessMode` more like `Permissions`? Also perhaps rename `Frozen` to `FrozenSecurityContext`, just calling it `Frozen` feels a bit out of place.

boggle:
Also: Did you benchmark the changes?

Mats-SX:
@boggle On `Frozen`, we figured it being an inner class (`SecurityContext.Frozen`) made the name make sense. Calling it `SecurityContext.FrozenSecurityContext` is a mouthful.

If it moves outside of the interface, I agree the name needs to change.

No specific benchmark has been made for this PR. No performance changes are expected, I think.

boggle:
@Mats-SX ok, didn't spot `Frozen` to be inner. On benchmarking, it looked like we now always instantiate security context objects that used to be enums, did I misread that? Let's perhaps talk later.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/AccessModeSnapshot.java
```diff
@@ -1,103 +0,0 @@
-/*
```
craigtaverner, line 1: I do not find a replacement for AccessModeSnapshot and AccessModeSnapshotTest. I would think at least the test code should exist in some form?

community/kernel/src/test/java/org/neo4j/kernel/impl/api/security/AccessModeSnapshotTest.java
```diff
@@ -1,48 +0,0 @@
...
- */
-package org.neo4j.kernel.impl.api.security;
-
-import org.junit.Test;
-
-import org.neo4j.kernel.api.security.AccessMode;
-
-import static org.junit.Assert.assertEquals;
-
-public class AccessModeSnapshotTest
```
craigtaverner, line 28: I do not find a replacement for AccessModeSnapshot and AccessModeSnapshotTest. I would think at least the test code should exist in some form?

henriknyman:
@Mats-SX I think some performance change is actually expected on nested tight loop procedure calls compared to the access mode snapshot code because of the additional object creation with the immutable design (also generating more garbage). This feels like a small step backwards from a performance point of view and I think we should profile it. It could easily be changed to have access mode mutable within the SecurityContext, so that all overriding is done only on access mode.

manual/cypher/graphgist/src/main/java/org/neo4j/doc/cypherdoc/BlockType.java
```diff
@@ -39,13 +39,15 @@
...
 import org.neo4j.kernel.api.security.AccessMode;
+import org.neo4j.kernel.api.security.AnonymousContext;
 import org.neo4j.kernel.api.KernelTransaction;
 import org.neo4j.kernel.impl.coreapi.InternalTransaction;
 import org.neo4j.visualization.asciidoc.AsciidocHelper;
 import org.neo4j.visualization.graphviz.AsciiDocSimpleStyle;
 import org.neo4j.visualization.graphviz.GraphvizWriter;
 import org.neo4j.walk.Walker;
 
+
```
Mats-SX: This may break checkstyle.

community/bolt/src/test/java/org/neo4j/bolt/security/auth/BasicAuthenticationTest.java
Mats-SX: I don't know. I'm reinstating it.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/security/AccessModeSnapshot.java
Mats-SX: `AccessModeSnapshot` is removed, because `AccessMode` is now immutable. Previously it was mutable because `AuthSubject` inherited from it. This is now handled using `SecurityContext`.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/StandardEnterpriseSecurityContext.java
Mats-SX: I don't know how exactly to act on these comments. Is it OK if we perhaps address them in an upcoming commit?

henriknyman (approved):

PR #8164, opened 2016-10-13
Make it possible to temporarily disable native access checking in UnsafeUtil

chrisvest:
This is useful for speeding up the execution of large numbers of generated tests that make intense use of native memory.

No tests in the main repo currently make use of this, but we would like to add tests to our block device repo that do make use of it.


burqen:
LGTM. Merge once green.

PR #8175, opened 2016-10-14
Bump hazelcast version

davidegrohmann:


lutovich:
@davidegrohmann it would be nice for the second commit message to describe the reason for version update.

PR #8176, opened 2016-10-14
Fixed `neo4j-admin backup` printing to stdout

spacecowboy:
Diagnostic output should go to stderr, with the rest.


benbc:
Looks fine -- assuming that you've tested this better than I evidently did.

spacecowboy:
@benbc here is some output which should prove it:

selection_007

And then for incremental:

selection_008

benbc:
Great

PR #8177, opened 2016-10-14
Better error messages for udf and procedures

pontusmelke:
When providing the wrong number of arguments to a procedure or user-defined function we should print the whole signature making it easier for use it correctly.

changelog: User-defined functions and procedures will now print a more helpful error message, including its signature, when used erroneously.

### DIFF
--- a/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/ResolvedCall.scala
+++ b/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/ResolvedCall.scala
@@ -101,7 +101,7 @@ case class ResolvedCall(signature: ProcedureSignature,
         }.foldLeft(success)(_ chain _)
       } else {
         val msg = (if (signature.inputSignature.isEmpty) "arguments"
-        else if (signature.inputSignature.size == 1) s"argument with type ${signature.inputSignature.head.typ.toNeoTypeString}"
+        else if (signature.inputSignature.size == 1) s"argument of type ${signature.inputSignature.head.typ.toNeoTypeString}"
         else s"arguments of type ${signature.inputSignature.map(_.typ.toNeoTypeString).mkString(", ")}") +
           signature.description.map(d => s"${System.lineSeparator()}Description: $d").getOrElse("")
         error(_: SemanticState, SemanticError(
--- a/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/ResolvedCall.scala
+++ b/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/ResolvedCall.scala
@@ -100,7 +100,15 @@ case class ResolvedCall(signature: ProcedureSignature,
             arg.semanticCheck(SemanticContext.Results) chain arg.expectType(field.typ.covariant)
         }.foldLeft(success)(_ chain _)
       } else {
-        error(_: SemanticState, SemanticError(s"Procedure call does not provide the required number of arguments ($expectedNumArgs)", position))
+        val msg = (if (signature.inputSignature.isEmpty) "arguments"
+        else if (signature.inputSignature.size == 1) s"argument with type ${signature.inputSignature.head.typ.toNeoTypeString}"
+        else s"arguments of type ${signature.inputSignature.map(_.typ.toNeoTypeString).mkString(", ")}") +
+          signature.description.map(d => s"${System.lineSeparator()}Description: $d").getOrElse("")
+        error(_: SemanticState, SemanticError(
+          s"""Procedure call does not provide the required number of arguments: got $actualNumArgs expected $expectedNumArgs.
+             |
+             |Procedure ${signature.name} has signature: $signature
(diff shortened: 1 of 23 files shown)


community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/ResolvedFunctionInvocation.scala
```diff
@@ -80,7 +80,14 @@ case class ResolvedFunctionInvocation(qualifiedName: QualifiedName,
...
         } else {
-          error(_: SemanticState, SemanticError(s"Function call does not provide the required number of arguments ($expectedNumArgs)", position))
+          val msg = (if (signature.inputSignature.isEmpty) "arguments"
+          else if (signature.inputSignature.size == 1) s"argument with type ${signature.inputSignature.head.typ.toNeoTypeString}"
+          else s"arguments of type ${signature.inputSignature.map(_.typ.toNeoTypeString).mkString(", ")}") +
+            signature.description.map(d => s"${System.lineSeparator()}Description: $d").getOrElse("")
+          error(_: SemanticState, SemanticError( s"""Function call does not provide the required number of arguments: expected $expectedNumArgs got $actualNumArgs.
+             |
+             |Function ${signature.name} has signature: $signature
+             |meaning that it expects $expectedNumArgs $msg""".stripMargin, position))
```
Mats-SX, line 90: The code in this block and the procedure block is almost identical. Could we extract a common function for them? I foresee that this error message will be edited in the future.

community/cypher/cypher-compiler-3.1/src/test/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/CallClauseTest.scala
```diff
@@ -229,8 +229,12 @@ class CallClauseTest extends CypherFunSuite with AstConstructionTestSupport {
...
     val resolved = ResolvedCall(_ => signature)(unresolved)
 
-    errorTexts(resolved.semanticCheck(SemanticState.clean)).toList should equal(List(
-      "Procedure call does not provide the required number of arguments (1) (line 1, column 0 (offset: 0))"
+    val toList: List[String] = errorTexts(resolved.semanticCheck(SemanticState.clean)).toList
+    toList should equal(List(
+      """Procedure call does not provide the required number of arguments: got 0 expected 1.
+        |
+        |Procedure my.proc.foo has signature: my.proc.foo(a :: INTEGER?) :: x :: INTEGER?, y :: LIST? OF NODE?
+        |meaning that it expects 1 argument with type INTEGER? (line 1, column 0 (offset: 0))""".stripMargin
```
Mats-SX: Nice!

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/ResolvedCall.scala
```diff
@@ -100,7 +100,15 @@ case class ResolvedCall(signature: ProcedureSignature,
             arg.semanticCheck(SemanticContext.Results) chain arg.expectType(field.typ.covariant)
         }.foldLeft(success)(_ chain _)
       } else {
-        error(_: SemanticState, SemanticError(s"Procedure call does not provide the required number of arguments ($expectedNumArgs)", position))
+        val msg = (if (signature.inputSignature.isEmpty) "arguments"
+        else if (signature.inputSignature.size == 1) s"argument with type ${signature.inputSignature.head.typ.toNeoTypeString}"
```
Mats-SX: ... argument `of` type ...

Mats-SX (commented):
Very nice usability improvement.

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/ResolvedFunctionInvocation.scala
pontusmelke: I am slightly torn on this, it started out as a common thing but after that the functions and procedures have slowly began drifting apart rather than becoming more similar. The code is similar but it actually acts on different types with no common superclass together with we need to make sure we say`Function` and not `Procedure` and vice versa.

Mats-SX: I was thinking a static helper that could take a string parameter (`Function` or `Procedure`). Perhaps an `object` called `SignatureArgumentsWriter`?

pontusmelke: It would have to be a function that would need to take a lot of arguments and when trying it out I think it just makes the code harder to read. I actually prefer the duplication.  Ping me if you have a suggestion, so far I have only been able to make it worse

Mats-SX: I trust your judgement on this one. Merging.

Issue #7814, opened 2016-08-26
Feature Request: Unique constraint across multiple labels


naraesk:
I have a set of nodes with multiple Labels (A, B, C). All nodes have a common property, which is unique across all labels. However, when creating unique constraint it is limited to one label, isn't it?

Documentation says something like:

`CREATE CONSTRAINT ON (n:A) ASSERT n.uid IS UNIQUE`

But it would be useful to do something like this:

`CREATE CONSTRAINT ON (n:A AND n:B AND n:C) ASSERT n.uid IS UNIQUE`

or

`CREATE CONSTRAINT ON (n) ASSERT n.uid IS UNIQUE`

spacecowboy:
Would make it easier, but at the moment you could create a common label D for all nodes and create the constraint on that.

NickyYo:
Hi @spacecowboy,  i was re-directed to this from stack overflow as a top search on Google.
 Do you have an example on how to create common labels?  Struggling a little.

Many thanks, Nick.

spacecowboy:
@NickyYo 

Ideally you can do this from the start. So that when you create nodes in the first place, they have the correct labels.

Say I have two kinds of labels originally, `COWBOY` and `SAMURAI`, and I want an index that spans both of these labels. I then define a third label, let's call it `GIBSONTYPE`.

A create query then looks something like this

```
CREATE (n:GIBSONTYPE:COWBOY {name:"Case"})
```

and

```
CREATE (n:GIBSONTYPE:SAMURAI {name: "Molly"})
```

Both `SAMURAI` nodes and `COWBOY` nodes have a label in common, namely `GIBSONTYPE`. So indices can be created on that label.

In case you need to update an existing graph, you can set new labels on nodes in this way for example:

```
MATCH (n:COWBOY)
SET n:COWBOY:GIBSONTYPE
```

and naturally

```
MATCH (n:SAMURAI)
SET n:SAMURAI:GIBSONTYPE
```

asashour:
Is there a real-life use case for a multi-label constraint, which could be shared with others?

naraesk:
The use case is the same as for any other unique constraint. The question is, do you want to create the super type D (referring to the comment of spacecowboy ) or not? In my case, you would add the label D to tens of thousands of nodes, which would not be used in any query afterwards. This makes the database more complex than it should be.

yarduza:
I vote up the need for this.

NielsonJ:
[...]

PR #8179, opened 2016-10-15
[WIP] - Allow creation of unique constraint across multiple labels

erwanor:
Addressing issue #7814 -  this is a WIP and will later include appropriate testing. My strategy is to handle this at the parsing/AST-generation level. Basically, match a `NodeLabels` pattern and create the consequent nodes in the AST and process the rest to the graph query engine as usual.

Note: I will be a bit busy for the next three weeks so if anyone feels like they can do the job in a shorter timeframe, feel free to push. Otherwise, I will get back to it once my course load gets back to normal.

---

CLA signed.


chrisvest:
There's a lot of compilation failures.

erwanor:
@chrisvest Hence the WIP tag! Besides, do you know where could I find developer documentation about the graph query engine? Online doc on the Neo4j wiki seems to exclusively be user doc; nothing about the GQE. 😢

systay:
Let's re-open when this PR is closer to being mergable.

PR #8180, opened 2016-10-16
Documentation for user-defined functions

pontusmelke:


manual/cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -283,3 +283,24 @@ include::split.asciidoc[leveloffset=+1]
 include::reverse.asciidoc[leveloffset=+1]
 
 include::tostring.asciidoc[leveloffset=+1]
+
+ifndef::asciidoctor[:leveloffset: 1]
+
+[[query-functions-udf]]
+== User-defined functions ==
+User-defined functions are written in Java, deployed into the database and are called in the same way as any other Cypher function.
```
Mats-SX, line 291: This doesn't detail _how_ they are deployed into the database. Perhaps referring to procedures would be enough?

... deployed into the database similarly to procedures (_link to procedure section_), and ...

Mats-SX, line 291: Additionally, I'd like to suggest a second paragraph:

UDFs share many similarities with procedures, but they differ in two key ways: they are read-only and may only return scalar values. These constraints makes it possible for the Cypher planner to treat them just as if they were built-in functions.

manual/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/UserDefinedFunctionTest.scala
```diff
@@ -0,0 +1,53 @@
...
+        val node = db.createNode(Label.label("Member"))
+        node.setProperty("name", name)
+      }
+    }
+  }
+
+  @Test def call_a_udf() {
+    testQuery(
+      title = "Call a user-defined function",
+      text = "This calls the user defined function `org.neo4j.procedure.example.join`",
```
Mats-SX: Here we use two versions of the same name (`user-defined function` vs `user defined function`). Either is fine, but we should try to be consistent. (I prefer the dash'd version.)

manual/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/UserDefinedFunctionTest.scala
```diff
@@ -0,0 +1,53 @@
...
+import org.neo4j.kernel.impl.proc.Procedures
+import org.neo4j.procedure.example.JoinFunction
+
+class UserDefinedFunctionTest extends DocumentingTestBase with QueryStatisticsTestSupport with HardReset {
+
+  override def section = "functions"
+
+  override def hardReset() = {
+    super.hardReset()
+    db.getDependencyResolver.resolveDependency(classOf[Procedures]).registerFunction( classOf[JoinFunction] )
```
Mats-SX: Padding space unnecessary in Scala.

manual/cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
jjaderberg: Let's not put deployment instructions here, this section should primarily explain what UDFs are and how to call them. The linked content (`See also <<user-defined-functions>>`) has details on writing and testing a UDF, and is part of the same section that covers procedures, so how to deploy is also discussed in that context.

manual/cypher/cypher-docs/src/test/scala/org/neo4j/cypher/docgen/UserDefinedFunctionTest.scala
jjaderberg: Use parentheses (but no arguments) for functions in prose. Also terminate sentence with a period: "...the user-defined function `org.neo4j.procedure.example.join().`"
It's confusing that the function namespace/package contains "procedure".

manual/cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
```diff
@@ -283,3 +283,24 @@ include::split.asciidoc[leveloffset=+1]
...
+----
+component=neo4j-cypher-docs
+source=org/neo4j/procedure/example/JoinFunction.java
+classifier=sources
+tag=joinFunction
+----
+
+ifndef::asciidoctor[:leveloffset: 2]
+
+include::call-a-user-defined-function.asciidoc[]
```
jjaderberg: This needs to be `include::call-a-user-defined-function.asciidoc[leveloffset=+1]`

jjaderberg (changes requested):

manual/cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
pontusmelke: I think I disagree: this section is about `functions` and I think I would prefer to keep procedures out of the discussion here. UDFs are compared with procedures in the dedicated section on UDFs.

pontusmelke: If possible I don't want to mention procedures at all here, since this is a section about procedures. @jjaderberg I am not totally clear what you are suggesting, should I remove the sentence, remove the code snippet, or both?

jjaderberg (approved):

manual/cypher/cypher-docs/src/docs/dev/ql/functions/index.asciidoc
Mats-SX: All right.

pontusmelke:
@jjaderberg I think I have addressed your concerns

PR #8184, opened 2016-10-17
Support username and metadata in TransactionData events

craigtaverner:
Users that write transaction events handlers are interested in having access to the information about the user running the transaction. This information can be either the username of the logged in user running the transaction, or the metadata map attached using the setTXMetaData procedure, which is used sometimes to associated 'ultimate user' or 'application user' to transactions run using a system account.

changelog [security] Support username and metadata in TransactionData events

### DIFF
--- a/community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionData.java
+++ b/community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionData.java
@@ -167,16 +167,16 @@ public interface TransactionData
     /**
      * Get the username under which authorization state this transaction is running.
      *
-     * @return name of subject
+     * @return the username of the user who initiated the transaction.
      */
-    String getUsername();
+    String username();
 
     /**
      * Applications that start transactions may attach additional application specific meta-data to each transaction.
      *
-     * @return The application specific meta-data map associated with this transaction
+     * @return The application specific meta-data map associated with this transaction.
      */
-    Map<String,Object> getMetaData();
+    Map<String,Object> metaData();
 
     /**
      * Return transaction id that assigned during transaction commit process.
--- a/community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionData.java
+++ b/community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionData.java
@@ -19,6 +19,9 @@
  */
 package org.neo4j.graphdb.event;
 
+import java.util.Map;
+import java.util.Optional;
+
 import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.Relationship;
 
@@ -160,6 +163,20 @@ public interface TransactionData
      * @return all properties that have been removed from relationships.
      */
     Iterable<PropertyEntry<Relationship>> removedRelationshipProperties();
+
+    /**
+     * Get the username under which authorization state this transaction is running.
+     *
+     * @return name of subject
+     */
+    String getUsername();
+
+    /**
+     * Applications that start transactions may attach additional application specific meta-data to each transaction.
+     *
+     * @return The application specific meta-data map associated with this transaction
+     */
+    Map<String,Object> getMetaData();
 
     /**
      * Return transaction id that assigned during transaction commit process.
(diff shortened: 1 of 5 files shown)


ikwattro:
Thanks @craigtaverner for taking our inputs into account.

Concerning an empty string username, wouldn't it be better to replace it with a public constant like `EMPTY_USER` for eg ? I actually don't see a use case for us where we would need to do something with an empty user though.

This addition is very userful to us though, thanks for that.

community/kernel/src/test/java/org/neo4j/kernel/impl/coreapi/TxStateTransactionDataViewTest.java
```diff
@@ -298,6 +303,48 @@ public void accessTransactionIdAndCommitTime()
...
+
+    @Test
+    public void shouldAccessUsernameFromAuthSubject()
+    {
+        AuthSubject authSubject = mock( AuthSubject.class );
+        when( authSubject.username() ).thenReturn( "Christof" );
+        when( transaction.mode() ).thenReturn( authSubject );
+
+        TxStateTransactionDataSnapshot transactionDataSnapshot = snapshot();
+        assertEquals( Optional.of("Christof"), transactionDataSnapshot.getUsername() );
```
chrisvest: You forgot to correct this when you removed the use of `Optional`.

community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionData.java
```diff
@@ -162,6 +165,20 @@
     Iterable<PropertyEntry<Relationship>> removedRelationshipProperties();
 
     /**
+     * Get the username under which authorization state this transaction is running.
+     *
+     * @return name of subject
+     */
+    String getUsername();
```
Mats-SX: I prefer omitting the prefix `get`, and just call this `username()`.

community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionData.java
```diff
@@ -162,6 +165,20 @@
...
+
+    /**
+     * Applications that start transactions may attach additional application specific meta-data to each transaction.
+     *
+     * @return The application specific meta-data map associated with this transaction
+     */
+    Map<String,Object> getMetaData();
```
Mats-SX: I prefer omitting the prefix `get`, and just call this `metaData()`.

community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionData.java
```diff
@@ -162,6 +165,20 @@
     Iterable<PropertyEntry<Relationship>> removedRelationshipProperties();
 
     /**
+     * Get the username under which authorization state this transaction is running.
```
Mats-SX: I find this a bit confusing. Perhaps `which` could be replaced with `whose`?

community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionData.java
```diff
@@ -162,6 +165,20 @@
...
+     *
+     * @return name of subject
```
Mats-SX: I suggest replacing with

```
@return the username of the user who initiated the transaction.
```

community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionData.java
```diff
@@ -162,6 +165,20 @@
...
+     */
+    String getUsername();
+
+    /**
+     * Applications that start transactions may attach additional application specific meta-data to each transaction.
+     *
+     * @return The application specific meta-data map associated with this transaction
```
Mats-SX: Add full stop.

community/kernel/src/main/java/org/neo4j/kernel/internal/TransactionEventHandlers.java
```diff
@@ -283,6 +286,18 @@ public boolean isDeleted( Relationship relationship )
...
         @Override
+        public String getUsername()
+        {
+            return "";
+        }
+
+        @Override
+        public Map<String,Object> getMetaData()
+        {
+            return Collections.EMPTY_MAP;
```
Mats-SX: I prefer using the generic version (also used above): `Collections.emptyMap()`.

Mats-SX (commented):
Mostly flavour comments.

craigtaverner:
Changed the API to no longer use the java classic `get` prefix on getters, since the rest of the class did not follow that convention either.

lutovich (approved):
Looks good to me.

PR #8188, opened 2016-10-17
Update index writer delete flush settings

MishaDemianenko:
Add setting for max buffered delete terms, max buffered docs;
Introduce couple of additional feature toggle for already defined configuration parameters.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/IndexWriterConfigs.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/IndexWriterConfigs.java
@@ -36,6 +36,10 @@ public final class IndexWriterConfigs
 {
     private static final int MAX_BUFFERED_DOCS =
             FeatureToggles.getInteger( IndexWriterConfigs.class, "max_buffered_docs", 100000 );
+    private static final int POPULATION_MAX_BUFFERED_DOCS =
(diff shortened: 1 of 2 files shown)


community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/IndexWriterConfigs.java
```diff
@@ -36,6 +36,10 @@
 {
     private static final int MAX_BUFFERED_DOCS =
             FeatureToggles.getInteger( IndexWriterConfigs.class, "max_buffered_docs", 100000 );
+    private static final int POPULATION_MAX_BUFFERED_DOCS =
+            FeatureToggles.getInteger( IndexWriterConfigs.class, "population_max_buffered_docs", IndexWriterConfig.DISABLE_AUTO_FLUSH );
```
lutovich, line 40: Maybe use `.` instead of `_` to be consistent with other feature toggle names in this class?

lutovich (approved):
@MishaDemianenko pr looks good, one small comment.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/IndexWriterConfigs.java
MishaDemianenko: @lutovich  max_buffered_docs on line 38 uses `_` as well as `BatchingMultipleIndexPopulator`, thats why i choose to use `_` here as well

PR #8189, opened 2016-10-17
Change the API of FlightMap so that we have separate methods for

mneedham:
truncating multiple entries and removing an individual entry

This will solve the problem where if we have gaps in the map
then some entries will never be removed.


enterprise/core-edge/src/main/java/org/neo4j/coreedge/core/state/InFlightLogEntryReader.java
```diff
@@ -59,10 +59,12 @@ public RaftLogEntry get( long logIndex ) throws IOException
             useInFlightMap = false;
             entry = getUsingCursor( logIndex );
         }
-
-        if ( removeReadIndexFromMap )
+        else
```
davidegrohmann: why?

mneedham: I originally had an assertion inside the `apply` method which would only let you remove the lowest key and it would blow up if you tried to remove a key that wasn't in the map

PR #8192, opened 2016-10-17
Errors raised in unpacking should also be sent back to client

pontusmelke:
After the state-machine refactoring we no longer send back fatal/external errors
to the client but only abruptly closes the connection. In some cases such as invalid
parameters to a run statement we should send back a proper error to the client and not
just silently fail.

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/BoltRequestMessageReader.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/BoltRequestMessageReader.java
@@ -20,13 +20,13 @@
 
 package org.neo4j.bolt.v1.messaging;
 
-import org.neo4j.bolt.v1.packstream.PackStream;
-import org.neo4j.bolt.v1.runtime.Neo4jError;
-import org.neo4j.kernel.api.exceptions.Status;
-
 import java.io.IOException;
 import java.util.Map;
 import java.util.Optional;
+
+import org.neo4j.bolt.v1.packstream.PackStream;
+import org.neo4j.bolt.v1.runtime.Neo4jError;
+import org.neo4j.kernel.api.exceptions.Status;
 
 /**
  * Reader for Bolt request messages made available via a {@link Neo4jPack.Unpacker}.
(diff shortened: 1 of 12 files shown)


community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/BoltRequestMessageReader.java
```diff
@@ -74,7 +76,14 @@ public boolean hasNext() throws IOException
...
                     Map<String, Object> params = unpacker.unpackMap();
-                    handler.onRun( statement, params );
+                    Optional<Neo4jError> error = unpacker.consumeError();
+                    if (error.isPresent())
+                    {
+                        handler.onExternalError( error.get() );
+                    } else {
+                        handler.onRun( statement, params );
+
+                    }
```
thobe: this hunk is horribly formatted.

community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/Neo4jPack.java
```diff
@@ -396,11 +398,16 @@ public Object unpack() throws IOException
...
                             }
                             break;
                         case NULL:
-                            throw new BoltIOException( Status.Request.Invalid, "Value `null` is not supported as key in maps, must be a non-nullable string." );
+                            error = Optional.of( Neo4jError.from( Status.Request.Invalid,
+                                    "Value `null` is not supported as key in maps, must be a non-nullable string." ) );
+                            unpackNull();
+                            val = unpack();
+                            map.put( null, val );
+                            break;
```
thobe, line 411: We are changing from only reacting to the first error (which was done in a bad way), to only reporting (good!) the _last error_. Should we perhaps take into account the possibility of there being more than one error, and package them all up?

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltStateMachine.java
```diff
@@ -257,6 +257,18 @@ public void interrupt()
...
+    public void externalError( Neo4jError error, BoltResponseHandler handler ) throws BoltConnectionFatality
+    {
+        before( handler );
+        try
+        {
+            fail( this, error );
+            this.state = State.FAILED;
+        }
+        finally { after(); }
+
```
thobe: Strangely formatted hunk.

thobe (commented):
I just had some minor comments from a quick read-through.

community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/Neo4jPack.java
pontusmelke: Added a simple mechanism for combining errors.

PR #8193, opened 2016-10-18
Add AuthorizationExpired exception to plugin API

henriknyman:
Plugins that manages their own caching of authorization info that requires
a user to re-authenticate need the possibility to signal this back to the
client.
To the client this will surface in the same way as if authorization
expired in the internal cache of the ldap and plugin auth providers.

changelog: Add `AuthorizationExpired` exception to plugin API so that clients can differentiate between cache expiry requiring a simple reauthentication and more serious failures

### DIFF
--- a/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
+++ b/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
@@ -403,7 +403,7 @@ public interface Status
         }
     }
 
-    enum Security implements Status  // TODO: rework by the Security Team before these are updated
+    enum Security implements Status
     {
         // client
         CredentialsExpired( ClientError, "The credentials have expired and need to be updated." ),
--- a/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginRealm.java
+++ b/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginRealm.java
@@ -46,6 +46,7 @@ import org.neo4j.server.security.enterprise.auth.SecureHasher;
 import org.neo4j.server.security.enterprise.auth.ShiroAuthToken;
 import org.neo4j.server.security.enterprise.auth.ShiroAuthorizationInfoProvider;
 import org.neo4j.server.security.enterprise.auth.plugin.api.AuthToken;
+import org.neo4j.server.security.enterprise.auth.plugin.api.AuthorizationExpired;
 import org.neo4j.server.security.enterprise.auth.plugin.api.RealmOperations;
 import org.neo4j.server.security.enterprise.auth.plugin.spi.AuthInfo;
 import org.neo4j.server.security.enterprise.auth.plugin.spi.AuthPlugin;
@@ -148,8 +149,16 @@ public class PluginRealm extends AuthorizingRealm implements RealmLifecycle, Shi
     {
         if ( authorizationPlugin != null )
         {
-            org.neo4j.server.security.enterprise.auth.plugin.spi.AuthorizationInfo authorizationInfo =
-                    authorizationPlugin.authorize( getPrincipalAndRealmCollection( principals ) );
+            org.neo4j.server.security.enterprise.auth.plugin.spi.AuthorizationInfo authorizationInfo;
+            try
+            {
+                 authorizationInfo = authorizationPlugin.authorize( getPrincipalAndRealmCollection( principals ) );
+            }
+            catch ( AuthorizationExpired e )
+            {
+                throw new AuthExpirationException(
+                        "Plugin '" + getName() + "' authorization info expired: " + e.getMessage() );
+            }
             if ( authorizationInfo != null )
             {
                 return PluginAuthorizationInfo.create( authorizationInfo );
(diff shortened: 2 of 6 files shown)


community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
```diff
@@ -403,7 +403,7 @@ public Code code()
         }
     }
 
-    enum Security implements Status  // TODO: rework by the Security Team before these are updated
```
Mats-SX, line 406: So I guess it's reworked now, then? ;)

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginRealm.java
```diff
@@ -148,8 +149,16 @@ protected AuthorizationInfo doGetAuthorizationInfo( PrincipalCollection principa
     {
         if ( authorizationPlugin != null )
         {
-            org.neo4j.server.security.enterprise.auth.plugin.spi.AuthorizationInfo authorizationInfo =
-                    authorizationPlugin.authorize( getPrincipalAndRealmCollection( principals ) );
+            org.neo4j.server.security.enterprise.auth.plugin.spi.AuthorizationInfo authorizationInfo;
```
Mats-SX, line 152: Why does it have to contain the full qualified name here? Can we rename it to avoid this?

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginRealm.java
```diff
@@ -148,8 +149,16 @@ protected AuthorizationInfo doGetAuthorizationInfo( PrincipalCollection principa
...
+            try
+            {
+                 authorizationInfo = authorizationPlugin.authorize( getPrincipalAndRealmCollection( principals ) );
+            }
+            catch ( AuthorizationExpired e )
+            {
+                throw new AuthExpirationException(
+                        "Plugin '" + getName() + "' authorization info expired: " + e.getMessage() );
```
Mats-SX, line 160: I'm a fan of attaching cause exceptions. Is there a particular reason we do not want to do that here? (To clarify, I mean sending in the caught exception into the constructor of the thrown exception.)

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/api/AuthorizationExpired.java
```diff
@@ -0,0 +1,41 @@
...
+ * {@link org.neo4j.server.security.enterprise.auth.plugin.spi.AuthorizationPlugin#authorize}
+ * method of a combined authentication and authorization plugin (that implements
+ * the two separate interfaces <tt>AuthenticationPlugin</tt> and <tt>AuthorizationPlugin</tt>),
+ * that manages its own caching of auth info.
+ *
+ * @see org.neo4j.server.security.enterprise.auth.plugin.spi.AuthorizationPlugin
+ */
+public class AuthorizationExpired extends RuntimeException
+{
+    public AuthorizationExpired( String message )
```
Mats-SX, line 37: Above comment about attaching cause exception would require a change here.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/api/AuthorizationExpired.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * An exception that can be thrown if authorization has expired and the user needs to re-authenticate
+ * in order to renew authorization.
+ * Throwing this exception will cause the server to disconnect the client.
+ *
+ * <p>This is typically used from the
+ * {@link org.neo4j.server.security.enterprise.auth.plugin.spi.AuthorizationPlugin#authorize}
+ * method of a combined authentication and authorization plugin (that implements
+ * the two separate interfaces <tt>AuthenticationPlugin</tt> and <tt>AuthorizationPlugin</tt>),
```
Mats-SX, line 30: Can we have `@link`s for these two as well?

Mats-SX (changes requested):
Unless there's a good reason, I'd like the exception to have its cause attached.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginRealm.java
henriknyman: Apache Shiro also has AuthorizationInfo. We are using both in this class.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginRealm.java
henriknyman: The primary point of this exception is communicating a status to the client. Only the message and status code will reach the client. Nobody is looking at the stack trace so I don't see any point of attaching the cause.

community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
henriknyman: I think it is an old comment that is not that relevant anymore

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginRealm.java
Mats-SX: OK, fair enough.

Mats-SX (approved):

PR #8194, opened 2016-10-18
Fix lifecycle bug when stopping NeoDataSource

davidegrohmann:
If the checkpointing step fails in NeoDataSource.stop(), we never call
life.shutdown(). Hence we'll never unmap the store files from the page
cache. So later when we try to close the page cache it will fail due
to still mapped files.  This change we'll make sure that we always
shutdown all the lifecycled instances in the internal life in
NeoDataSource

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
@@ -502,6 +502,10 @@ public class NeoStoreDataSource implements Lifecycle, IndexProviders
             throw Exceptions.launderedException( e );
         }
 
+        // NOTE: please make sure this is performed after having added everything to the life, in fact we would like
+        // to perform the checkpointing as first step when the life is shutdown.
+        life.add( lifecycleToTriggerCheckPointOnShutdown() );
+
         try
         {
             life.start();
@@ -873,23 +877,7 @@ public class NeoStoreDataSource implements Lifecycle, IndexProviders
             // will be able to start committing at this point.
             awaitAllTransactionsClosed();
 
-            // Write new checkpoint in the log only if the kernel is healthy.
-            // We cannot throw here since we need to shutdown without exceptions,
-            // so let's make the checkpointing part of the life, so LifeSupport can handle exceptions properly
-            life.add( new LifecycleAdapter()
-            {
-                @Override
-                public void shutdown() throws Throwable
-                {
-                    if ( databaseHealth.isHealthy() )
-                    {
-                        // Flushing of neo stores happens as part of the checkpoint
-                        transactionLogModule.checkPointing()
-                                .forceCheckPoint( new SimpleTriggerInfo( "database shutdown" ) );
-                    }
-                }
-            } );
-
+            // Checkpointing is now triggered as part of life.shutdown see lifecycleToTriggerCheckPointOnShutdown()
             // Shut down all services in here, effectively making the database unusable for anyone who tries.
             life.shutdown();
         }
@@ -909,6 +897,26 @@ public class NeoStoreDataSource implements Lifecycle, IndexProviders
         {
             LockSupport.parkNanos( 10_000_000 ); // 10 ms
         }
+    }
+
+    private Lifecycle lifecycleToTriggerCheckPointOnShutdown()
+    {
+        // Write new checkpoint in the log only if the kernel is healthy.
+        // We cannot throw here since we need to shutdown without exceptions,
+        // so let's make the checkpointing part of the life, so LifeSupport can handle exceptions properly
+        return  new LifecycleAdapter()
+        {
+            @Override
+            public void shutdown() throws Throwable
+            {
+                if ( databaseHealth.isHealthy() )
+                {
+                    // Flushing of neo stores happens as part of the checkpoint
+                    transactionLogModule.checkPointing()
+                            .forceCheckPoint( new SimpleTriggerInfo( "database shutdown" ) );
+                }
+            }
+        };
     }
 
     @Override
--- a/community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
@@ -84,7 +84,6 @@ import org.neo4j.kernel.impl.proc.Procedures;
 import org.neo4j.kernel.impl.storageengine.impl.recordstorage.RecordStorageEngine;
 import org.neo4j.kernel.impl.store.MetaDataStore;
 import org.neo4j.kernel.impl.store.StoreId;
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
```diff
@@ -851,18 +851,6 @@ protected void assertNotClosed()
     @Override
     public void close()
     {
-        if ( idGenerator == null || !storeOk )
-        {
-            try
-            {
-                closeStoreFile();
```
tinwelint, line 858: This is something of an unneccesary change in 3.0, also it's already done in 3.1

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -876,19 +874,21 @@ public synchronized void stop()
             awaitAllTransactionsClosed();
 
             // Write new checkpoint in the log only if the kernel is healthy.
-            // We cannot throw here since we need to shutdown without exceptions.
-            if ( databaseHealth.isHealthy() )
+            // We cannot throw here since we need to shutdown without exceptions,
+            // so let's make the checkpointing part of the life, so LifeSupport can handle exceptions properly
+            life.add( new LifecycleAdapter()
```
tinwelint: Perhaps this is a neat way of solving it, I just want this checkpointer lifecycle instance to be added during startup, not during calling stop().

tinwelint (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
davidegrohmann: Last time I checked, it was in 3.1 as well.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
davidegrohmann: Yes, unfortunately doing it that way you suggest, it is very hard to guarantee it that the check pointer is "shutdown" first. This way it might not elegant but solve the mentioned problem and touches the least amount of code since I am doing this change in a maintenance branch.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
davidegrohmann: Anyhow, what's wrong in doing this change in 3.0 as well?

MishaDemianenko: no reason to touch it in a maintenance branch

community/kernel/src/test/java/org/neo4j/kernel/NeoStoreDataSourceTest.java
```diff
@@ -242,6 +245,34 @@ public void logModuleSetUpError() throws Exception
                 equalTo( openStoresError ) ) );
     }
 
+    @Test
+    public void shouldAlwaysShutdownLifeEvenWhenCheckPointingFails() throws Exception
```
MishaDemianenko, line 249: Is it possible to have also integration test for this?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
davidegrohmann: I see no reason to not remove stupid code in an old branch. Being super religious about this stuff is not very helpful.

community/kernel/src/test/java/org/neo4j/kernel/NeoStoreDataSourceTest.java
davidegrohmann: This is already failing in 2 stress tests.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
tinwelint: I think it was removed today or something. Funny how two devs notice the same thing at the same time after years just sitting there in the code base :) I'm sure it's not a problem to remove this in 3.0 as well though

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
tinwelint: By adding it to life "last" it should be shutdown "first", that order is reversed during stop/shutdown isn't it?

davidegrohmann: yes it is correct, adding it last means shutting down first

community/kernel/src/test/java/org/neo4j/kernel/NeoStoreDataSourceTest.java
MishaDemianenko: that does not remove the need in IT right?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
MishaDemianenko: you can freely do it in an active brunch, no reason to touch something else there in general just because you can, that's why it's called maintenance

tinwelint:
Btw, what's was the reason for checkpointing failing?

davidegrohmann:
@tinwelint I don't know yet, what comes out is a failure in closing the page cache (since we didn't close the store) but the last thing I see in the `debug.log` is a tentative to check point on shutdown.

community/kernel/src/test/java/org/neo4j/kernel/NeoStoreDataSourceTest.java
davidegrohmann: Having 3 tests showing the problem is enough I think.

davidegrohmann: I have added an IT for this scenario

MishaDemianenko: thank you, appreciate!

tinwelint:
@davidegrohmann This is a slightly scary change in 3.0 since it changes behaviour of shutdown in the event of failure during shutdown, a failure which we don't know what it is. If this is a response to some of our own test failures I think it should not go to maintenance until we (a) understand the problem and (b) potentially have users in real scenarios experiencing this. Otherwise risk is that the change itself breaks something else, isn't it?

davidegrohmann:
@tinwelint you have 2 tests in the code base showing the bug! how is that not enough?

PR #8195, opened 2016-10-18
We don't need to bootstrap in RaftDiscoveryServiceConnector anymore as

mneedham:
we already do this after we download a snapshot


davidegrohmann:
@mneedham there are failing tests

mneedham:
@davidegrohmann sorry for taking so long. Have got those tests working locally + done Causal Clustering renames

PR #8196, opened 2016-10-18
Give client error message on all failures

pontusmelke:
Until now we have been obfuscating error messages for
_unexpected database errors_ with the intention that these are
errors more aimed at an administrator than aimed directly at the client.
However since the classification isn't perfect we get too many false negatives,
i.e. in too many case we end up returning obfuscated messages where the client
would have benefited from a detailed message.


thobe:
I don't think this re-classification of error codes is a good thing. It is a breaking incompatible change, since it changes the error codes.

I also don't think `Neo.ReportableDatabaseError.Schema.IndexCreationFailed` is a particularly good error code. The "Reportable" word in there seems out of place.

How about we introduce this differentiation in the logging behaviour without changing the error codes themselves? One way to do this would be to still introduce the `ReportableDatabaseError` classification object, but have it serialise to the same classification as before, i.e. `DatabaseError` (in terms of implementation, this could be achieved by overriding `toString()`).

For similar reasons, I'm not sure it is a good idea to change the classification of `IndexDropFailed` from `DatabaseError` to `ClientError`, even `ClientError` might be a better classification for it.

pontusmelke:
@thobe the PR has been updated after our discussions.

PR #8198, opened 2016-10-18
Fixes importer issue which could hang shutdown during panic

tinwelint:
this would prevent import tool from shutting down, but more importantly
it would prevent the error to be shown.

This fix treats a call to close() before all processing is done in
ParallelInputEntityDeserializer as a panic and delegates that flag to
its executor so that all processors can now abort their processing
accordingly.

### DIFF
--- a/community/import-tool/src/test/java/org/neo4j/tooling/EntityDataGenerator.java
+++ b/community/import-tool/src/test/java/org/neo4j/tooling/EntityDataGenerator.java
@@ -84,7 +84,7 @@ public class EntityDataGenerator<T> extends InputIterator.Adapter<T>
     public void close()
     {
         super.close();
-        processing.shutdown( 0 );
+        processing.shutdown();
     }
 
     @Override
--- a/community/import-tool/src/test/java/org/neo4j/tooling/EntityDataGenerator.java
+++ b/community/import-tool/src/test/java/org/neo4j/tooling/EntityDataGenerator.java
@@ -84,7 +84,7 @@ public class EntityDataGenerator<T> extends InputIterator.Adapter<T>
     public void close()
     {
         super.close();
-        processing.shutdown( false );
+        processing.shutdown( 0 );
     }
 
     @Override
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/executor/DynamicTaskExecutor.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/executor/DynamicTaskExecutor.java
@@ -168,6 +168,12 @@ public class DynamicTaskExecutor<LOCAL> implements TaskExecutor<LOCAL>
     {
         if ( shutDown )
         {
+            // We're already shut down, although take the ABORT_QUEUED flag seriously as to abort all
+            // processors still working on tasks. This looks like a panic.
+            if ( (flags & TaskExecutor.SF_ABORT_QUEUED) != 0 )
+            {
+                this.abortQueued = true;
+            }
             return;
         }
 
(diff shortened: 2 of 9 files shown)


community/import-tool/src/test/java/org/neo4j/tooling/EntityDataGenerator.java
```diff
@@ -84,7 +84,7 @@ public long position()
     public void close()
     {
         super.close();
-        processing.shutdown( false );
+        processing.shutdown( 0 );
```
MishaDemianenko: Would be really nice if we will have plain `shutdown()` in case everything is fine and `shutdown(TaskPolicy taskPolicy)` that take enum value as param instead of int.
Then in simple case you do not need even think what the number you need to pass that and in case you actually need to pass something - you can see enum values instead of trying yo find int value

burqen: +1 on this

burqen (changes requested):
This looks good. Would be cleaner with the suggestion from @MishaDemianenko

tinwelint:
@burqen Simplified shutdown now, feels much better (I was never happy about the shutdown flags)

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/executor/TaskExecutor.java
```diff
@@ -46,23 +43,31 @@
...
+     * being processed will be aborted and currently processing tasks will not be awaited for completion.
+     *
+     * @param panic cause of panic.
      */
-    void shutdown( int flags );
+    void panic( Throwable panic );
 
     /**
-     * @return {@code true} if {@link #shutdown(int)} has been called, otherwise {@code false}.
+     * @return {@code true} if {@link #shutdown()} has been called, otherwise {@code false}.
```
burqen: From  this comment it is not clear to be if `isShutdown` will return true or false if `panic` has been called. Can we please clarify this?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/executor/DynamicTaskExecutor.java
```diff
@@ -164,24 +160,35 @@ private void notifyProcessors()
...
         {
             parkAWhile();
         }
+        this.shutDown = true;
+    }
+
+    @Override
+    public void panic( Throwable panic )
+    {
+        this.panic = panic;
```
burqen: Multiple calls to `panic` will overwrite each other. That means original cause of failure will be overwritten by subsequent failures.

We could chain them together, although we do not _know_ that they are related. Another idea would be to keep a list of panics and report them all in `assertHealthy`.

burqen (changes requested):
Second review done. Please have a look.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/executor/DynamicTaskExecutor.java
chrisvest: … or add subsequent panics as suppressed exceptions to the original.

burqen (approved):

burqen:
@tinwelint Looking good. Merge once green!

burqen:
@tinwelint Had to revert this merge because of test failure in `ParallelInputEntityDeserializerTest.shouldTreatExternalCloseAsPanic`.
Branch should be rebased on 3.0, fixed and put into a new PR.

PR #8202, opened 2016-10-18
Update legacy indexes to preserve documents sort fields on update

MishaDemianenko:
Legacy indexes use stored fields to store user data and sorted doc value typed fields to make sorting smooth.
After retrieval of any saved document from index store only stored fields are visible (not sort fields)
and in case if we will re-save that document we will loose any sorted fields that we had before in the index for that document.
This PR introduce functionality that will restore all 'missing' sort fields to make sure that documents can be always sorted.
Also as result of fix we need to migrate potentially affected old indexes and restore potentially missing sorting fields. To be able to that legacy index store migrators updated and both standard and high level format versions upgraded to be able to use standard migration mechanism.


tinwelint:
It looks like you've bumped the store format and duplicated all those formats, why did you do that? The formats are all the same, aren't they?

tinwelint:
@MishaDemianenko I'd like it if you would squasch the two last commits, just so much more noise in a PR making a lot of changes and having another commit undoing those.

MishaDemianenko:
@tinwelint done

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/IndexType.java
```diff
@@ -403,25 +392,74 @@ private void clearDocument( Document document )
...
+                addNewFieldToDocument( document, key, existingValue );
+            }
+        }
+    }
+
+    private boolean haveSortField( IndexableField[] fields )
+    {
+        for ( IndexableField field : fields )
+        {
+            if ( !DocValuesType.NONE.equals( field.fieldType().docValuesType() ) )
```
tinwelint, line 438: use the other `haveSortedField` method here?

community/consistency-check/src/test/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporterTest.java
```diff
@@ -237,7 +236,7 @@ private void assertConsistent( File storeDir ) throws ConsistencyCheckIncomplete
 
     protected String getFormatName()
     {
-        return StandardV3_0.NAME;
+        return StandardV3_0_7.NAME;
```
tinwelint, line 239: (This comment applies to all similar changes): should we have a notion of "latest" standard format to not have to make wide-spread changes like this and potentially forget some?

community/kernel/src/main/java/org/neo4j/kernel/spi/legacyindex/IndexImplementation.java
```diff
@@ -107,4 +107,11 @@
      */
     @Override
     void shutdown() throws Throwable;
+
+    /**
+     * Provide store migrator for particular index implementation that participate in store migration and
+     * should be able to perform cross version migration of index implementation
+     * @return index implementation store migration
+     */
+    StoreMigrationParticipant getStoreMigrator();
```
tinwelint: Provide default implementation `NOT_PARTICIPATING` to not cause problems for actual custom implementations out there?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/store/format/highlimit/v306/HighLimitFactoryV3_0_6.java
```diff
@@ -0,0 +1,38 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.store.format.highlimit.v306;
+
+import org.neo4j.helpers.Service;
+import org.neo4j.kernel.impl.store.format.RecordFormats;
+
+@Service.Implementation( RecordFormats.Factory.class )
+public class HighLimitFactoryV3_0_6 extends RecordFormats.Factory
```
tinwelint, line 26: Shouldn't this be V3_0_7?

MishaDemianenko: 3.0.6 it's a version where format was last used, while new one is always HighLimit and factory for it is HighLimitFactory. Which is different how it is for standard record format where we use version where it was introduced as part of versioning schema

community/consistency-check/src/test/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporterTest.java
MishaDemianenko: yes that sounds like a good idea, card created

MishaDemianenko:
@tinwelint updated

MishaDemianenko:
Replaced by #8462 closing

PR #8205, opened 2016-10-19
Supports degrees up to 2^35 in importer

tinwelint:
without additional memory overhead. This is done via indirection in
integer count field (which also contains other flags though) if count
surpasses 2^29 into an array which has bigger slots. This indirection
should be extremely rare and only kicks in for nodes with more than
2^29 number of relationships in a single import.


tinwelint:
Will move to 3.2 probably

tinwelint:
Replaced by https://github.com/neo4j/neo4j/pull/8223

PR #8206, opened 2016-10-19
Fix typo in default setting value in neo4j.conf

henriknyman:
The commented-out default value of `dbms.security.ldap.user_dn_template`
had a trailing qoute that would lead to syntax error if uncommented.

changelog: Fixed trailing quote in default value in neo4j.conf which leads to syntax error if uncommented


Mats-SX (approved):
+1

fickludd:
Seems like this caused 1050 tests to fail... serious stuff! Otherwise I think it's fine =)

PR #8207, opened 2016-10-19
Don't expose page-size as option in import-tool

tinwelint:
this option was introduced very recently (in previous maint release),
but shouldn't have been since page size shouldn't be a user concern.
Manually configuring page size is potentially risky and has no functional
effect of the import anyway.


spacecowboy:
@tinwelint but it WAS released. So you can't arbitrarily remove it in the next point release.

We should instead remove it from `neo4j-admin` and that is fine because it has never been released except in beta. `neo4j-import` is deprecated anyway.

PR #8208, opened 2016-10-19
Change security setting names

henriknyman:
This is based on demands from product management to have more consistent naming of settings
- Change the security setting names to have a more consistent differentiation
  between authentication and authorization settings.
- Change the internal setting variable names to be more consistent with
  the real setting names.

Builds on https://github.com/neo4j/neo4j/pull/8206

changelog: Changed security setting names to have more consistent differentiation between authentication and authorization settings


Mats-SX (approved):
Looks pretty good!

PR #8211, opened 2016-10-19
Remove `page-size` option from `neo4j-admin`.

srbaker:
No changelog needed, because this hasn't been in a release.

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/CsvImporter.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/CsvImporter.java
@@ -39,9 +39,7 @@ import org.neo4j.unsafe.impl.batchimport.input.Collector;
 import org.neo4j.unsafe.impl.batchimport.input.csv.CsvInput;
 import org.neo4j.unsafe.impl.batchimport.input.csv.IdType;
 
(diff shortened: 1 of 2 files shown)


community/dbms/src/main/java/org/neo4j/commandline/dbms/CsvImporter.java
```diff
@@ -139,7 +132,7 @@ public void doImport() throws IOException
         OutputStream badOutput = new BufferedOutputStream( fs.openAsOutputStream( reportFile, false ) );
         Collector badCollector = badCollector( badOutput, 1000, collect( true, true, true ) );
 
-        Configuration configuration = importConfiguration( null, false, config, pageSize );
+        Configuration configuration = importConfiguration( null, false, config, -1 );
```
spacecowboy: is the `-1` a sensible default? It won't try and use a -1M page cache right?

chrisvest: You could create an overload of `importConfiguration` that doesn't take a `pageSize` argument, or a `processors` argument for that matter..

spacecowboy (approved):
LGTM

PR #8213, opened 2016-10-19
Don't close transaction from separate thread

pontusmelke:
The side-channel method `terminate` on `BoltStateMachine` shouldn't kill the
transaction directly but just mark it and then we close down the machine from
the session thread on next call instead.

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltStateMachine.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltStateMachine.java
@@ -23,6 +23,7 @@ package org.neo4j.bolt.v1.runtime;
 import java.time.Clock;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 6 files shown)


community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltStateMachine.java
```diff
@@ -290,14 +296,20 @@ public String owner()
...
+        ctx.isTerminated.set( true );
+        ctx.statementProcessor.markCurrentTransactionForTermination();
         spi.onTerminate( this );
     }
 
     @Override
-    public boolean hasTerminated()
+    public boolean willTerminate()
     {
-        return isClosed();
```
Mats-SX, line 300: Does this warrant removing `isClosed()`? I notice it's `public`, but it's not part of an interface and only used by one test matcher, so perhaps?

Mats-SX (approved):

henriknyman (approved):

PR #8218, opened 2016-10-19
Fix flaky LoadCommandTest

spacecowboy:
Since the lock was not assigned to anything the garbage collector was free to GC the lock at any time. This happened infrequently (something like 1/1000 to 1/100 runs as I recreated it).

To encourage better usage of the lock I also made it implement `Closeable` so that try-with-resources could be used. All usages were changed so that the prior art is clear for future tests.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/internal/StoreLocker.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/internal/StoreLocker.java
@@ -19,6 +19,7 @@
  */
 package org.neo4j.kernel.internal;
 
+import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.nio.channels.FileLock;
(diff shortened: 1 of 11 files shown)


community/kernel/src/main/java/org/neo4j/kernel/internal/StoreLocker.java
```diff
@@ -103,4 +107,9 @@ public void release() throws IOException
             storeLockFileChannel = null;
         }
     }
+
+    @Override
+    public void close() throws IOException {
+        release();
```
MishaDemianenko: What if instead of adding new `close` method we will just rename existing `release` to `close`?
Having two ways to do the same thing is always confusing and we do not have any interfaces or contracts to satisfy with current `release` methods

MishaDemianenko (commented):
One small question/comment, otherwise locks really nice

community/kernel/src/main/java/org/neo4j/kernel/internal/StoreLocker.java
spacecowboy: I was just trying to keep the diff as minimal as possible. But since I now have conflicts anyway I an do that.

PR #8220, opened 2016-10-19
Optimization of isGraphKernelResultValue

pontusmelke:
`isGraphKernelResultValue` is used a lot and shows up in profiling.
This change makes it roughly 30 percent faster assuming an even distribution of
types (`Node`, 'Relationship', `Path`, `String`, `Integer`, `Float`, `Boolean`)

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/isGraphKernelResultValue.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/isGraphKernelResultValue.scala
@@ -19,13 +19,9 @@
  */
 package org.neo4j.cypher.internal
 
-import org.neo4j.graphdb.{Path, Node, Relationship}
+import org.neo4j.graphdb.{Path, PropertyContainer}
 
 object isGraphKernelResultValue extends (Any => Boolean) {
-  override def apply(v: Any): Boolean = v match {
-    case node: Node => true
-    case rel: Relationship => true
-    case path: Path => true
-    case _ => false
-  }
+
+  override def apply(v: Any): Boolean = v.isInstanceOf[PropertyContainer] || v.isInstanceOf[Path]
 }


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/isGraphKernelResultValue.scala
```diff
@@ -19,13 +19,9 @@
...
 
 object isGraphKernelResultValue extends (Any => Boolean) {
-  override def apply(v: Any): Boolean = v match {
-    case node: Node => true
-    case rel: Relationship => true
-    case path: Path => true
-    case _ => false
-  }
+
+  override def apply(v: Any): Boolean = v.isInstanceOf[Node] || v.isInstanceOf[Relationship] || v.isInstanceOf[Path]
```
jexp: You could even use PropertyContainer to save one check.

jexp:
Good catch, I used the same approach locally and it worked well. 
I only stored the class references in the object not sure if that makes a big difference.

```
import org.neo4j.graphdb.{Path, PropertyContainer}

object isGraphKernelResultValue extends (Any => Boolean) {
  val PropertyContainerType = classOf[PropertyContainer]
  val PathType = classOf[Path]
  override def apply(v: Any): Boolean = PropertyContainerType.isInstance(v) || PathType.isInstance(v)
}
```

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/isGraphKernelResultValue.scala
pontusmelke: ~~I used `Entity` and that was actually slower than doing the check twice. Not sure why though.~~ No your right, it makes it faster

pontusmelke:
@jexp I tested with that (using `inInstance` instead of `instance of`) but it is actually significantly slower in all my tests. Even slower than the original code, no idea why though.

Mats-SX:
How did you benchmark this?

jexp:
@pontusmelke cool I trust you in this. I just did ad-hoc testing with loading for a graph-algo.

pontusmelke:
@Mats-SX nothing fancy, I created a list of values with all supported types (`Node`, `Path`, `Relationship`, `double`, `long`, `boolean`), i.e. I assume an even distribution. Then I just looped over that calling `isGraphKernelResultValue` some number of million times and measured elapsed times with standard deviation.

It would of course be better to use JMH or something but the difference is big enough that I trust that this is at least making it faster.

boggle:
@pontusmelke Now that this is merged it occurred to me that if we really wanted to optimize this, we should probably introduce a marker interface for all graph values (PropertyContainer, Path) and check against that instead.

PR #8222, opened 2016-10-19
Change AuthToken parameters to not return null

henriknyman:
Return empty map instead of null so that auth plugin writers do not have
to check for null.


fickludd:
Good stuff.

PR #8223, opened 2016-10-20
Supports degrees up to 2^35 in importer

tinwelint:
without additional memory overhead. This is done via indirection in
integer count field (which also contains other flags though) if count
surpasses 2^29 into an array which has bigger slots. This indirection
should be extremely rare and only kicks in for nodes with more than
2^29 number of relationships in a single import.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NodeRelationshipCache.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NodeRelationshipCache.java
@@ -20,6 +20,7 @@
 package org.neo4j.unsafe.impl.batchimport.cache;
 
 import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
 import org.neo4j.graphdb.Direction;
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NodeRelationshipCache.java
```diff
@@ -106,27 +116,101 @@ public NodeRelationshipCache( NumberArrayFactory arrayFactory, int denseNodeThre
...
-    void setCount( long nodeId, int count )
+    /**
+     * Should only be used by tests
+     */
+    void setCount( long nodeId, long count, Direction direction )
+    {
+        if ( isDense( nodeId ) )
+        {
+            long relGroupId = all48Bits( array, nodeId, SPARSE_ID_OFFSET );
+            relGroupCache.getAndSetCount( relGroupId, direction, count );
```
MishaDemianenko, line 132: why do we use relGroupCache here to store the count?

MishaDemianenko:
@tinwelint had a single comment; not sure if we need to have introduced possibilities

MishaDemianenko:
@tinwelint i have changed target branch for this pr to point to 3.2 now, still have open quiestion

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NodeRelationshipCache.java
tinwelint: Because the count we store is per relationship chain head. A dense node has 3 chain heads per type, which is why we need to store them in the relGroupCache for dense nodes.

tinwelint:
@MishaDemianenko cheers, I also replied to your question. Makes sense?

MishaDemianenko:
@tinwelint ok, makes sense.

PR #8224, opened 2016-10-20
Fix: made auth plugins log to security.log instead of debug.log

fickludd:
changelog: Changed auth plugins to log to security.log instead of debug.log and exposed security log level in config as `dbms.logs.security.level`

### DIFF
--- a/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/configuration/SecuritySettings.java
+++ b/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/configuration/SecuritySettings.java
@@ -237,7 +237,7 @@ public class SecuritySettings
             PATH );
 
     @Description("Security log level threshold.")
-    public static final Setting<Level> security_log_level = setting( "dbms.security.log_level",
+    public static final Setting<Level> security_log_level = setting( "dbms.logs.security.level",
             options( Level.class ), "INFO" );
 
     @Description( "Set to log successful authentication events." )
--- a/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/configuration/SecuritySettings.java
+++ b/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/configuration/SecuritySettings.java
@@ -27,6 +27,7 @@ import org.neo4j.graphdb.config.Setting;
 import org.neo4j.graphdb.factory.Description;
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
 import org.neo4j.kernel.configuration.Internal;
+import org.neo4j.logging.Level;
 
 import static org.neo4j.kernel.configuration.Settings.BOOLEAN;
 import static org.neo4j.kernel.configuration.Settings.BYTES;
@@ -39,6 +40,7 @@ import static org.neo4j.kernel.configuration.Settings.STRING_LIST;
 import static org.neo4j.kernel.configuration.Settings.derivedSetting;
 import static org.neo4j.kernel.configuration.Settings.max;
 import static org.neo4j.kernel.configuration.Settings.min;
+import static org.neo4j.kernel.configuration.Settings.options;
 import static org.neo4j.kernel.configuration.Settings.setting;
 
 /**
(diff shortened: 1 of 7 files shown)


enterprise/security/src/main/java/org/neo4j/server/security/enterprise/configuration/SecuritySettings.java
```diff
@@ -226,6 +228,10 @@
             ( logs ) -> new File( logs, "security.log" ),
             PATH );
 
+    @Description("Security log level threshold.")
+    public static final Setting<Level> security_log_level = setting( "dbms.security.log_level",
```
henriknyman: The corresponding setting for debug.log is called `dbms.logs.debug.level`. I wonder if we should aim for consistency with that?

henriknyman: It would be good to have the setting in enterprise neo4j.conf, commented-out with default value  (even if debug log level is not there)

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/configuration/SecuritySettings.java
```diff
@@ -234,6 +236,10 @@
             ( logs ) -> new File( logs, "security.log" ),
             PATH );
 
+    @Description("Security log level threshold.")
```
Mats-SX: Missing wrapping space (code style).

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/plugin/PluginRealmTest.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        @Override
+        public AuthorizationInfo authorize( Collection<PrincipalAndRealm> principals )
+        {
+            return null;
+        }
+    }
+
+    abstract class LoggingAdapter extends RealmLifecycle.Adapter
```
Mats-SX: Not requesting change in this PR, but perhaps it would be nice to have a somewhat sophisticated test `PluginRealm` available to use (even as an abstract class that tests could extend), which would contain implementations for these interfaces? I'm just envisioning that we will have (or have already) several test classes that could use such a test infrastructure instead of having to tailor new private classes. WDYT?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/log/SecurityLogTest.java
```diff
@@ -54,17 +58,61 @@ public void shouldRotateLog() throws IOException
         FileSystemAbstraction fs = fileSystemRule.get();
 
         File activeLogFile = config.get( SecuritySettings.security_log_filename );
-        MatcherAssert.assertThat( fs.fileExists( activeLogFile ), Matchers.equalTo( true ) );
-        MatcherAssert.assertThat( fs.fileExists( archive( 1 ) ), Matchers.equalTo( true ) );
-        MatcherAssert.assertThat( fs.fileExists( archive( 2 ) ), Matchers.equalTo( false ) );
+        assertThat( fs.fileExists( activeLogFile ), Matchers.equalTo( true ) );
```
Mats-SX: Could statically import `equalTo()` to improve readability.

Mats-SX (commented):
Looks useful!

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/log/SecurityLog.java
```diff
@@ -53,10 +53,10 @@ public SecurityLog( Config config, FileSystemAbstraction fileSystem, Executor ex
...
                 config.get( SecuritySettings.store_security_log_max_archives ), executor );
 
-        this.inner = new AsyncLog(
-                event -> executor.execute( event::process ),
-                builder.toOutputStream( rotatingSupplier )
-        );
+        FormattedLog formattedLog = builder.toOutputStream( rotatingSupplier );
+        formattedLog.setLevel( config.get( SecuritySettings.security_log_level ) );
+
+        this.inner = new AsyncLog( event -> executor.execute( event::process ), formattedLog );
```
davidegrohmann, line 59: Why async?

PR #8225, opened 2016-10-20
asPrivateType not handling lists and maps

pontusmelke:
The conversion method for private types wasn't properly handling Java collections properly.
Meaning that whenever lists or maps are used as parameters they weren't converting correctly.


Mats-SX (approved):
Looks great.

PR #8226, opened 2016-10-20
Fix lock test which fails on Windows

spacecowboy:
Problem here was that both `forceShouldRespectStoreLock` and `shouldAllowForcedCopyOverAnExistingDatabase` implicitly defined the same destination database directory. This caused a file locking conflict on Windows which holds on to file mappings in a strict way.


spacecowboy:
the test is now fixed. It failed consistently every time on windows before.

benbc:
LGTM

PR #8227, opened 2016-10-20
Causal Clustering rename.

jimwebber:
First sweep through removing the 'edge' nomenclature from the code and config.


jjaderberg:
Fine as far as I can tell

PR #8230, opened 2016-10-20
Return servers in a shuffled order in the getServers proc.

maxsumrall:


enterprise/core-edge/src/test/java/org/neo4j/coreedge/discovery/procedures/GetServersProcedureTest.java
```diff
@@ -426,6 +422,57 @@ public void shouldReturnNoWriteEndpointsIfThereIsNoAddressForTheLeader() throws
                 containsInAnyOrder( coreAddresses( 0 ).getRaftServer().toString() ) );
     }
 
+    @Test
+    public void shouldReturnReadEndpointsInDifferentOrders() throws Exception
```
davidegrohmann: Shall we add tests for routing and write endpoints as well?

maxsumrall: Yes for routing, no for writing. The current implementation only has one possible write endpoint.

PR #8234, opened 2016-10-21
3.0 biased winner strategy

mneedham:
changelog: Adds configuration settings for ordering branch handling and store copying on slaves


digitalstain (commented):
We should ensure that checkstyle rules for 3.1 are met

Mats-SX:
@digitalstain Do you know what Pontus' commits are doing in this PR?

PR #8240, opened 2016-10-22
Resolve dependencies of Neo4jTransactionContext{Factory} early

boggle:


community/cypher/cypher/src/main/java/org/neo4j/cypher/javacompat/internal/GraphDatabaseCypherService.java
```diff
@@ -19,28 +19,31 @@
...
 
-import java.net.URL;
-import java.util.concurrent.TimeUnit;
-
-import org.neo4j.graphdb.DependencyResolver;
-import org.neo4j.graphdb.GraphDatabaseService;
-import org.neo4j.graphdb.Label;
-import org.neo4j.graphdb.Node;
-import org.neo4j.graphdb.Relationship;
+import org.neo4j.graphdb.*;
```
MishaDemianenko: Please reformat this without start imports :)

community/cypher/cypher/src/main/java/org/neo4j/cypher/javacompat/internal/GraphDatabaseCypherService.java
```diff
@@ -19,28 +19,31 @@
...
-    private GraphDatabaseFacade graph;
+    private final GraphDatabaseFacade graph;
+    private final DbmsOperations dbmsOperations;
+    private final ThreadToStatementContextBridge txBridge;
 
     public GraphDatabaseCypherService( GraphDatabaseService graph )
     {
         this.graph = (GraphDatabaseFacade) graph;
+        DependencyResolver resolver = getDependencyResolver();
+        this.dbmsOperations = resolver.resolveDependency( DbmsOperations.class );
```
MishaDemianenko: Why do we need to use `GraphDatabaseCypherService` as a service provider instead of dependencyResolver?
I have tried to find a place where we call `getDbmsOperations` or `getTxBridge` to get stored services and it looks like we always have dependencyResolver available there.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -196,7 +197,7 @@ public GraphDatabaseFacade()
     /**
      * Create a new Core API facade, backed by the given SPI.
      */
-    public void init( SPI spi, Config config )
+    public void init( SPI spi, ThreadToStatementContextBridge txBridge, Config config )
```
MishaDemianenko: Hm.. what do you think about passing dependency resolver here instead of already resolved txBridge?
So we can make it thing more generic

community/kernel/src/main/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContext.java
```diff
@@ -245,12 +234,6 @@ public Statement statement()
     }
 
     @Override
-    public void check()
```
MishaDemianenko: hm why we are removing this? That thing was recently introduced (that's why it's not used) for cypher land to be able to check if current transaction is still valid while doing time consuming operations (lets imaging huge sorting, etc.)

community/kernel/src/test/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacadeTest.java
```diff
@@ -37,10 +33,10 @@
...
 import org.neo4j.kernel.impl.coreapi.TopLevelTransaction;
 
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import java.util.HashMap;
+import java.util.concurrent.TimeUnit;
+
+import static org.mockito.Mockito.*;
```
MishaDemianenko: unfriendly star again

community/kernel/src/test/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContextTest.java
```diff
@@ -39,46 +36,23 @@
...
 import org.neo4j.kernel.impl.coreapi.PropertyContainerLocker;
-import org.neo4j.kernel.impl.coreapi.TopLevelTransaction;
 
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
-import static org.mockito.Mockito.when;
+import java.util.Collections;
+
+import static org.mockito.Mockito.*;
```
MishaDemianenko: here it goes again

community/kernel/src/test/java/org/neo4j/kernel/GraphDatabaseFacadeFactoryTest.java
```diff
@@ -23,34 +23,24 @@
...
-
 import org.neo4j.graphdb.mockfs.EphemeralFileSystemAbstraction;
 import org.neo4j.helpers.Exceptions;
-import org.neo4j.kernel.impl.factory.DataSourceModule;
-import org.neo4j.kernel.impl.factory.DatabaseInfo;
-import org.neo4j.kernel.impl.factory.EditionModule;
-import org.neo4j.kernel.impl.factory.GraphDatabaseFacade;
-import org.neo4j.kernel.impl.factory.GraphDatabaseFacadeFactory;
-import org.neo4j.kernel.impl.factory.PlatformModule;
+import org.neo4j.kernel.impl.factory.*;
```
MishaDemianenko: ⭐️

MishaDemianenko (changes requested):
Have a couple of small comments also please check your code format settings

community/cypher/cypher/src/main/java/org/neo4j/cypher/javacompat/internal/GraphDatabaseCypherService.java
boggle: The other dependency resolver we got doesn't resolve all the dependencies we need. This change just materializes what was going on already but at least avoids reresolution.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
boggle: Don't mind; will try

community/kernel/src/main/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContext.java
boggle: I just removed it because it was not used. If we want it for going forward I can try to bring it back.

MishaDemianenko: that would be great :)

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
boggle: Actually doing this will require to resolve again per procedure call. I thought the point of this whole exercise was to avoid having to resolve dependencies again and again?

MishaDemianenko: yes indeed then it's fine

PR #8242, opened 2016-10-24
Fixed `neo4j install-service` on Windows 7

spacecowboy:
This would throw a powershell exception on Windows 7. Should be
compatible with all versions of Windows now.


benbc:
LGTM

benbc:
Merging despite failed build because the problems are understood and fixed elsewhere.

PR #8245, opened 2016-10-24
Add new config settings for `allowed`

Mats-SX:
When procedures and UDFs come from a third party, where modifying source code
is not possible, it was previously not possible to assign an `allowed` role to them.
This PR introduces two new config settings that may be used to set specific roles
to any procedures and UDFs that do not have their `allowed` annotation set.

In detail, wildcard mappings between procedure names and allowed roles are supported. All matching mappings are returned as roles. If no roles are found, the default is used.

changelog: Add config options for default and wildcard procedure `allowed` roles


Mats-SX:
Test breakage is unrelated and being fixed by the kernel team. Should be mergable.

PR #8257, opened 2016-10-25
Allow `neo4j-admin` commands to be blocked from running.

srbaker:


benbc:
LGTM

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
```diff
@@ -71,5 +73,30 @@ public String name()
...
+         * @param databaseName  the name of the database that the command applies to.
+         * @param config        a configuration object for the blocker to use when making its decision.
+         * @return Whether the command should be blocked from running or not.
+         */
+        public abstract boolean doesBlock( String databaseName, Config config );
+
+        /**
+         * @return              a list of the commands this blocker applies to.
+         */
+        public abstract List<String> commands();
```
chrisvest: Perhaps this should be a `Set<String>`?

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
```diff
@@ -71,5 +73,30 @@ public String name()
         public abstract AdminCommand create( Path homeDir, Path configDir, OutsideWorld outsideWorld );
     }
 
+    abstract class Blocker extends Service
```
chrisvest: Why a `Service`? Why an abstract class instead of an interface?

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
```diff
@@ -71,5 +73,30 @@ public String name()
...
+    {
+        protected Blocker( String key, String... altKeys )
+        {
+            super( key, altKeys );
+        }
```
chrisvest: We're not doing any lookup with the keys.

community/dbms/src/main/java/org/neo4j/commandline/dbms/DumpCommand.java
```diff
@@ -100,6 +113,15 @@ public void execute( String[] args ) throws IncorrectUsage, CommandFailed
...
         Path databaseDirectory = toDatabaseDirectory( database );
+        Config config = loadNeo4jConfig( homeDir, configDir, database, new HashMap<>() );
+
+        for ( AdminCommand.Blocker blocker : blockerLocator.findBlockersForCommand( "dump" ) )
+        {
+            if ( blocker.doesBlock( database, config ) )
+            {
+                throw new CommandFailed( blocker.explanation() );
+            }
+        }
```
chrisvest: Why is this in the individual commands, and not in `AdminTool`? The `LoadCommand` is in the same situation, isn't it?

community/dbms/src/test/java/org/neo4j/commandline/dbms/DumpCommandTest.java
```diff
@@ -310,6 +317,66 @@ public void shouldWrapIOExceptionsCarefulllyBecauseCriticalInformationIsOftenEnc
...
+    public void shouldBlockDumpIfABlockerSaysSo() throws IncorrectUsage
+    {
+        AdminCommand.Blocker blocker = mock( AdminCommand.Blocker.class );
+        when( blocker.doesBlock( any(), any() ) ).thenReturn( true );
+        when( blocker.explanation() ).thenReturn( "blocked" );
+
+        List<AdminCommand.Blocker> blockers = new ArrayList<AdminCommand.Blocker>()
+        {{
+            add( blocker );
+        }};
```
chrisvest: Whoever told you about this syntax has broken a vow. Also, there's `Arrays.asList(…)`

chrisvest (commented):
At first I thought would only have one comment, but it ended up being many. Please take a look at them.

spacecowboy:
@srbaker I don't think this is something for the changelog. It's strictly an internal implementation detail.

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
benbc: No reason for this to be an abstract class. The `AdminCommand.Provider` is usefully abstract, but this needn't be.

benbc (commented):
Looks good apart from the abstract/interface thing.

PR #8258, opened 2016-10-25
Fix buffer size evaluation in EphemeralFileSystemAbstraction for big stores.

MishaDemianenko:
Fix calculation of store size index for stores that over exceed standard sizes.
Before in some of the cases newly evaluated capacity was actually less then requested.

### DIFF
--- a/community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
+++ b/community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
@@ -1184,8 +1184,7 @@ public class EphemeralFileSystemAbstraction implements FileSystemAbstraction
          */
         private ByteBuffer allocate( int sizeIndex )
         {
-            int capacity = (sizeIndex < SIZES.length) ?
-                           SIZES[sizeIndex] : ((sizeIndex - SIZES.length + 1) * SIZES[SIZES.length - 1]);
+            int capacity = capacity( sizeIndex );
             try
             {
                 return ByteBuffer.allocateDirect( capacity );
@@ -1202,6 +1201,12 @@ public class EphemeralFileSystemAbstraction implements FileSystemAbstraction
                     throw oom;
                 }
             }
+        }
+
+        private int capacity( int sizeIndex )
+        {
+            return (sizeIndex < SIZES.length) ?
+                           SIZES[sizeIndex] : ((sizeIndex - SIZES.length + 1) * SIZES[SIZES.length - 1]);
         }
 
         void free()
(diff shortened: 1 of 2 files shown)


community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
```diff
@@ -1204,6 +1203,12 @@ private ByteBuffer allocate( int sizeIndex )
             }
         }
 
+        private int capacity( int sizeIndex )
+        {
+            return (sizeIndex < SIZES.length) ?
+                           SIZES[sizeIndex] : ((sizeIndex - SIZES.length + 1) * SIZES[SIZES.length - 1]);
```
burqen, line 1209: The real problem is that `capacity( indexSize )` does not scale as expected by this code in `verifySize( total amount )`

```
while ( newSize < totalAmount )
{
    newSize += Math.min( newSize, 1024 * 1024 );
    sizeIndex++;
}
```

Capacity scales with sizeIndex like: 64, 128, 256, 512, `1024, 1024`, 2048, 3072,... k. While `verifySize` does not of course expect the double `1024`.
Correct calculation of capacity when `sizeIndex < SIZES.length` would be

```
SIZES[SIZES.length - 1] + (sizeIndex - SIZES.length + 1) * SIZES[SIZES.length - 1]
```

which simplifies to

```
(sizeIndex - SIZES.length + 2) * SIZES[SIZES.length - 1]
```

I suggest fixing this bug in `capacity`.

burqen (changes requested):
This PR is not fixing the actual problem, rather covering it up. See comment.

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
MishaDemianenko: The root problem is even more then different scale step, i do not see why we need to do sizeIndex <-> actual size dance at all; this change is  minimal change for current release. Lets talk about that as well :)

burqen: I still think that changing how `capacity` is calculated would be an even better minimal change.

MishaDemianenko: @burqen current change try to use what ever method is there for expected size calculation, lets change way how ephemeral fs allocates it buffers in 3.2 since we can make it all together much more simpler and that will not gonna require either size index either conversion between size and index. And we will not gonna have any confusion around that

burqen: Sounds good, merging. Please create a card for it if that's not done already.

burqen (approved):

PR #8261, opened 2016-10-26
Make FileSystemAbstraction Closeable

MishaDemianenko:
Make ephemeral file system to implement AutoCloseable and move resource cleanup into close() instead of relying on finalization.
That will make usage pattern more explicit and will avoid cases when yet opened resources get cleaned up because file system itself get garbage collected.


MishaDemianenko:
@lutovich it actually look like that was not that much work to update all the tests :)

burqen:
LGTM!

burqen (changes requested):
Making `EphemeralFileSystemAbstraction` implement `AutoClosable` will mean further splitting the behaviour and lifecycle from `DefaultFileSystemAbstraction`.

I would argue that we want to work towards unifying behaviour between the implementations of `FileSystemAbstraction` rather than splitting them up further. Especially since we might have to further rely on the behaviour of `FileSystemAbstraction` when cleaning up tech dept in `PageCache` forced by development in `Blockdevice`.

I agree that relying on `finalization` is not a very compelling idea either, though.

community/command-line/src/test/java/org/neo4j/commandline/admin/RealOutsideWorldTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.neo4j.io.fs.FileSystemAbstraction;
+import org.neo4j.test.rule.system.SystemExitRule;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+public class RealOutsideWorldTest
+{
+
```
burqen: Formatting, empty line

community/command-line/src/test/java/org/neo4j/commandline/admin/RealOutsideWorldTest.java
```diff
@@ -0,0 +1,65 @@
...
+        FileSystemAbstraction fileSystemMock = mock( FileSystemAbstraction.class );
+        outsideWorld.fileSystemAbstraction = fileSystemMock;
+
+        systemExitRule.expectExit( 0 );
+
+        outsideWorld.exit( 0 );
+
+        verify( fileSystemMock ).close();
+    }
+
```
burqen: Formatting, empty line

community/io/src/test/java/org/neo4j/graphdb/mockfs/UncloseableDelegatingFileSystemAbstraction.java
```diff
@@ -0,0 +1,42 @@
...
+package org.neo4j.graphdb.mockfs;
+
+import java.io.IOException;
+
+import org.neo4j.io.fs.FileSystemAbstraction;
+
+/**
+ * File system abstraction that wraps real file system and prevent it to be closed.
+ * Useful for cases when we pass file system to db, shutdown it and verify file system content.
+ */
```
burqen: Some grammar edits:

```
/**
 * File system abstraction that wraps real file system and prevent it from being closed.
 * Useful for cases when we pass file system to db, shut it down and verify file system content.
 */
```

community/neo4j-harness/src/main/java/org/neo4j/harness/internal/AbstractInProcessServerBuilder.java
```diff
@@ -113,39 +112,44 @@ public TestServerBuilder copyFrom( File originalStoreDir )
...
+            config.put( ServerSettings.third_party_packages.name(), toStringForThirdPartyPackageProperty( extensions.toList() ) );
 
-        final FormattedLogProvider userLogProvider = FormattedLogProvider.toOutputStream( logOutputStream );
-        GraphDatabaseDependencies dependencies = GraphDatabaseDependencies.newDependencies();
-        dependencies = dependencies.kernelExtensions(
-                append( new Neo4jHarnessExtensions( procedures ), dependencies.kernelExtensions() ) )
-                .userLogProvider( userLogProvider );
+            final FormattedLogProvider userLogProvider = FormattedLogProvider.toOutputStream( logOutputStream );
+            GraphDatabaseDependencies dependencies = GraphDatabaseDependencies.newDependencies();
+            dependencies = dependencies.kernelExtensions( append( new Neo4jHarnessExtensions( procedures ), dependencies.kernelExtensions() ) )
```
burqen: Formatting, some long lines here

tools/src/main/java/org/neo4j/tools/rebuild/RebuildFromLogs.java
```diff
@@ -135,7 +135,10 @@ public static void main( String[] args ) throws Exception
             }
         }
 
-        new RebuildFromLogs( new DefaultFileSystemAbstraction() ).rebuild( source, target, txId );
+        try (FileSystemAbstraction fileSystem = new DefaultFileSystemAbstraction())
```
burqen: Formatting

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -452,7 +450,7 @@ public static void doImport( PrintStream out, PrintStream err, File storeDir, Fi
         LogService logService = life.add( StoreLogService.inLogsDirectory( fs, logsDir ) );
 
         life.start();
-        BatchImporter importer = new ParallelBatchImporter( storeDir,
+        BatchImporter importer = new ParallelBatchImporter( storeDir, fs,
```
burqen: Put fs on new line to follow established pattern.

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
```diff
@@ -234,6 +224,7 @@ public BatchInserterImpl( final File storeDir, final FileSystemAbstraction fileS
 
         life = new LifeSupport();
         this.storeDir = storeDir;
+        life.add( new FileSystemLifecycleAdapter( fileSystem ) );
```
burqen: The pattern so far has been "if you created the FSA you are responsible for closing it". This breaks that pattern.
Example:
`BatchInserterTest` uses a `DefaultFileSystemRule` and provides a FSA from this rule when creating a `BatchInserterImpl` using `BatchInserters`.
Now, they will both try to close the FSA which is fine but it is problematic that you can not trust your FSA not to be closed by someone else when giving it away.

I suggest `BatchInserters` should always take FSA from outside and never create its own, and thus we can remove this line.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/PropertyKeyTest.java
```diff
@@ -49,7 +50,7 @@ public void lazyLoadWithinWriteTransaction() throws Exception
         // Given
         FileSystemAbstraction fileSystem = fs.get();
         File dir = new File( "dir" ).getAbsoluteFile();
-        BatchInserter inserter = BatchInserters.inserter( dir, fileSystem );
+        BatchInserter inserter = BatchInserters.inserter( dir, new UncloseableDelegatingFileSystemAbstraction( fileSystem ) );
```
burqen: With suggested change to BatchInserters and BatchInserterImpl we can get rid of this and let PropertyKeyTest take full responsibility for its fileSystem.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/EmbeddedInteraction.java
```diff
@@ -52,8 +52,13 @@
...
     EmbeddedInteraction( Map<String, String> config ) throws Throwable
     {
+        this (config, EphemeralFileSystemAbstraction::new);
+    }
+
+    EmbeddedInteraction( Map<String, String> config, Supplier<FileSystemAbstraction> fileSystemSupplier ) throws Throwable
+    {
         TestEnterpriseGraphDatabaseFactory factory = new TestEnterpriseGraphDatabaseFactory();
-        factory.setFileSystem( new EphemeralFileSystemAbstraction() );
+        factory.setFileSystem( fileSystemSupplier.get() );
```
burqen: Responsibility for FSA becomes a bit unclear here. We know that the factory will inject the FSA into `PlatformModule` using the provided Supplier and that it will be added to some `life`, which means we don't need to take responsibility for closing. Although we work around this in at some place by wrapping FSA in `UncloseableDelegatingFileSystemAbstraction`.

I see that it's hard to work around this problem because some times we want to be able to keep the FSA around after interaction with a running db and some times we don't. Maybe we can document it in `TestGraphDatabaseFactory` that a provided FSA will be closed by `PlatformModule` if not wrapped in `Uncloseable...`.

burqen (changes requested):
First review half done. Still have last four commits left. Please have a look at the comments.

community/io/src/test/java/org/neo4j/io/pagecache/impl/SingleFilePageSwapperWithRealFileSystemIT.java
```diff
@@ -20,18 +20,27 @@
...
 import java.io.File;
+import java.io.IOException;
 
 import org.neo4j.io.fs.DefaultFileSystemAbstraction;
 import org.neo4j.io.fs.FileSystemAbstraction;
 
 public class SingleFilePageSwapperWithRealFileSystemIT extends SingleFilePageSwapperTest
 {
-    private final DefaultFileSystemAbstraction fs = new DefaultFileSystemAbstraction();
+    private DefaultFileSystemAbstraction fs;
```
burqen: `SingleFilePageSwapperTest` already have a `DefaultFileSystemAbstraction` instance.
Maybe we can use this one instead of creating another.

Let superclass have two methods `realFs` and `ephemeralFs` and delegate to the appropriate one from `getFs`. Then we can get rid of almost everything in this test class.

enterprise/ha/src/test/java/org/neo4j/kernel/ha/management/HighAvailabilityBeanTest.java
```diff
@@ -79,8 +78,8 @@
     private final LastUpdateTime lastUpdateTime = mock( LastUpdateTime.class );
     private final ClusterDatabaseInfoProvider dbInfoProvider =
             new ClusterDatabaseInfoProvider( clusterMembers, lastTxIdGetter, lastUpdateTime );
-    private final KernelData kerneData = new HighlyAvailableKernelData( db, clusterMembers, dbInfoProvider,
-            new DefaultFileSystemAbstraction(), null, new File( "storeDir" ), Config.empty() )
+    private DefaultFileSystemAbstraction fileSystem = new DefaultFileSystemAbstraction();
+    private final KernelData kerneData = new HighlyAvailableKernelData( db, clusterMembers, dbInfoProvider, fileSystem, null, new File( "storeDir" ), Config.empty() )
```
burqen: Formatting, long line

burqen (changes requested):
@MishaDemianenko First review round finished. Two additional comments.

burqen:
@MishaDemianenko Build failure because of
`StoreCopyClientTest.java:246: error: Two or more consecutive empty lines`

MishaDemianenko:
@burqen comments addressed, new line hater!
thx for really good feedback!

community/kernel/src/test/java/org/neo4j/test/TestGraphDatabaseFactory.java
```diff
@@ -47,7 +48,11 @@
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.boltConnector;
 
 /**
- * Test factory for graph databases
+ * Test factory for graph databases.
+ * Please be aware that since it's a database it will close filesystem as part of it lifecycle
+ * a real database.
+ * If you expect your file system to be open in the end use {@link UncloseableDelegatingFileSystemAbstraction}
+ *
```
burqen: Suggestion for comment change:

```
/** 
* Test factory for graph databases.
 * Please be aware that since it's a database it will close filesystem as part of its lifecycle.
 * If you expect your file system to be open after database is closed, use {@link UncloseableDelegatingFileSystemAbstraction}
/*
```

burqen (changes requested):
This is starting to shape up. I started on a list of places where we still don't close FSA but then I realised that you are probably not done with this PR. Anyway, gonna leave the (not finished) list here.

**DefaultFileSystemAbstraction**
`CheckTxLog.main`, line 91.
`GraphStoreFixture.directStoreAccess`, line 133.
`AbstractRESTInteraction.fileSystem`, line 121. Called from `AuthScenariosInteractionTestBase.SecurityLog.load`, line 767. Think it would be enough to wrap fileSystem in try catch.
`PartitionStorageTest.shouldListIndexPartitionsSorted`, line 176.
`HighAvailabilityBeanTest`, DFSA should be created in `@before` and not as part of declaration because it will be closed as in `@after`.

MishaDemianenko:
@burqen updated and rebased

MishaDemianenko:
@burqen ping

community/consistency-check/src/test/java/org/neo4j/consistency/checking/GraphStoreFixture.java
```diff
@@ -121,6 +122,23 @@ public GraphStoreFixture( String formatName )
         this( false, formatName );
     }
 
+    @Override
+    protected void after( boolean success )
+    {
+        super.after( success );
+        if (fileSystem != null)
```
burqen: Formatting

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/AuthScenariosInteractionTestBase.java
```diff
@@ -776,4 +756,26 @@ public void shouldNotTryToCreateTokensWhenReading()
...
     }
+
+    private class SecurityLog
+    {
+        List<String> lines;
+
+        void load() throws IOException
+        {
+            File securityLog = new File( AuthScenariosInteractionTestBase.this.securityLog.getAbsolutePath() );
+            try ( FileSystemAbstraction fileSystem = neo.fileSystem();
```
burqen, line 767: There are three different implementations of `NeoInteractionLevel.fileSystem()`, `AbstractRESTInteraction`, `BoltInteraction` and `EmbeddedInteraction`.
It is only `AbstractRestInteraction` that create a new instance of FSA when called. The other once keep `fileSystem` as a state. Thus, the second time we call `neo.fileSystem()` we will have an instance that is already closed.

I think it is better if `NeoInteractionLevel` itself is responsible for closing the fileSystem, because the embedded one gets the `fileSystem` from a `TestEnterpriseGraphDatabaseFactory` which will also use it to inject into the database itself and eventually ad it to lifecycle, effectively giving away control to database shutdown.

Do you have any other suggestions on how to make this more clear?

burqen (changes requested):
I think these are all of the places where we don't close FSA.

**DefaultFileSystemAbstraction**
`RecoveryTest`, line 81
`LuceneDataSourceTest`, line 67
`DatabaseRebuildToolTest`, line 175
`StoreMigratorFrom20IT`, line 97 (we already have a FileSystemRule in place here but not using it.
`BackupToolIT`, line 64
`BackupService`, line 167

**EphemeralFileSystemAbstraction**
`RandomPageCacheTestHarness`, line 101, 425

**RealOutsideWorld** (that creates a FSA)
`CsvImporterTest`, line 58
`ImportCommandTest`, line 69

**EphemeralFileSystemAbstraction.snapshot()**
A lot of places.
Would it be a good idea to let EFSA itself be responsible for keeping track of other instances that it has spawned with `snapshot` and close them as well during close?

**DelegatingFileSystemAbstraction**
There are quite a few places where we create a `DelegatingFileSystemAbstraction` together with an other FSA that we close but without closing the DFSA. This may be ok but I think following the pattern of letting the delegator delegate the close call would be better.
I'll not list all of the cases here, search project for `new DelegatingFileSystemAbstraction`

I'm finished with the implementations listed here, there are still some implementations left that I haven't checked. To be continued...

enterprise/backup/src/test/java/org/neo4j/backup/BackupToolIT.java
```diff
@@ -70,6 +71,7 @@ public void setUp() throws Exception
     public void teardown() throws Exception
     {
         pageCache.close();
+        fs.close();
```
burqen: This class makes it pretty clear that how we use FSA is not very rigid (making this PR super ligit).
We have three instantiations of `DefaultFileSystemAbstraction` here.
`TestDirectory` will create its own, so will `EmbeddedDatabaseRule` and finally we create our own as part of the test.

It would be nice to clean this up, but maybe that is to much for this PR. Thoughts?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/InternalFlatFileRealmIT.java
```diff
@@ -91,9 +91,10 @@ public void setup() throws Throwable
     }
 
     @After
-    public void teardown() throws Throwable
+    public void tearDown() throws Throwable
     {
         realm.shutdown();
+        fs.close();
```
burqen, line 93: `fs` here is an `EvilFileSystem` wrapping an `EphemeralFileSystem` created by a rule. I think the rule should be removed as this is the only place it is used.

burqen (changes requested):
**`FSA` in combination with `TestDirectory`**
Right now, `TestDirectory` can be provided with a `FSA` from the outside or create its own. This make responsibility a bit unclear on who should close it.

I see two alternatives:
1. Let `TestDirectory` take all the responsibility for closing fs which could be a good idea, seeing as it will create a new fs if it not provided with one. Although there are quite a few places where we use TestDirectory in combination with `DefaultFileSystemRule`. Those places should be cleaned up if we choose 1.
2. Don't leave responsibility of closing fs to `TestDirectory` and instead always provide it with an fs from the outside. There are a lot of places that will be affected by this change.

Maybe there is a better alternative.

**Places not yet fixed**

**DefaultFileSystemAbstraction**
`StoreMigratorFrom20IT`, line 97 (we already have a FileSystemRule in place here but not using it)

**RealOutsideWorld (that creates a FSA)**
`CsvImporterTest`, line 58
`ImportCommandTest`, line 69

**DelegatingFileSystemAbstraction**
There are quite a few places where we create a `DelegatingFileSystemAbstraction` together with an other FSA that we close but without closing the DFSA. This may be ok but I think following the pattern of letting the delegator delegate the close call would be better.
I won't list all of the cases here, search project for `new DelegatingFileSystemAbstraction`

**AdversarialChannelDefaultFileSystemAbstraction**
Created by `SingleFilePageSwapperWithAdversarialFileChannelIT` every time `getFs()` is called in `SingleFilePageSwapperTest`, which is quite a few times.

**AdversarialFileSystemAbstraction**
Has the same problem as `DelegatingFileSystemAbstraction`. The wrapped fileSystem seem to be closed in most cases, but `AdversarialFileSystemAbstraction` is not.

**SelectiveFileSystemAbstraction**
`DurableStateStorageIT`, line 103, 145, 220

burqen:
@MishaDemianenko Another review round finished

enterprise/backup/src/test/java/org/neo4j/backup/BackupToolIT.java
MishaDemianenko: Lets keep that away from this PR there are too much places that will eventually require almost full rewrite as soon as we will be more strict with FS abstraction

MishaDemianenko:
@burqen  Ok and will all rules and delegation/wrapper abstraction i would really like to propose do that work in separate PRs. Current one already grow to a much bigger state then it should be for PR that introduce possibility to close FS.
Tons of code and tests should be updated with new abstraction in mind since most of the places where those kind of rules and commons states are used need to be slightly updated. Current state give is an intermediate state in our trip to ideal state anyway.

MishaDemianenko:
@burqen rebased

burqen (approved):

burqen:
@MishaDemianenko Sounds good! I'll merge this now.

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/BatchInserters.java
```diff
@@ -36,50 +35,24 @@
...
-     *
-     * @param storeDir the store directory
-     * @return a new {@link BatchInserter}
-     * @throws IOException if there is an IO error
-     */
-    public static BatchInserter inserter( File storeDir ) throws IOException
-    {
-        return inserter( storeDir, stringMap() );
-    }
-
```
chrisvest, line 50: Removing these are an incompatible product surface change.

MishaDemianenko: there is already separate card for that

PR #8262, opened 2016-10-26
List allowed roles per procedure in enterprise

Mats-SX:
The `dbms.procedures` procedure will now contain an additional
column `roles` that lists the roles that are allowed to execute it
(enterprise only). This column is only populated if the executing user
is an administrator, otherwise it will be empty.

changelog: Added column to `dbms.procedures` in enterprise, listing roles allowed to execute the procedure

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInDbmsProcedures.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInDbmsProcedures.java
@@ -22,6 +22,7 @@ package org.neo4j.kernel.builtinprocs;
 import java.util.stream.Stream;
 
 import org.neo4j.kernel.api.proc.ProcedureSignature;
+import org.neo4j.kernel.api.proc.UserFunctionSignature;
 import org.neo4j.kernel.impl.proc.Procedures;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
 import org.neo4j.procedure.Context;
@@ -45,7 +46,29 @@ public class BuiltInDbmsProcedures
                 .map( ProcedureResult::new );
     }
 
-    @SuppressWarnings( "WeakerAccess" )
+    @Description( "List all user functions in the DBMS." )
+    @Procedure(name = "dbms.functions", mode = DBMS)
+    public Stream<FunctionResult> listFunctions()
+    {
+        return graph.getDependencyResolver().resolveDependency( Procedures.class ).getAllFunctions().stream()
+                .sorted( ( a, b ) -> a.name().toString().compareTo( b.name().toString() ) )
+                .map( FunctionResult::new );
+    }
+
+    public static class FunctionResult
+    {
+        public final String name;
+        public final String signature;
+        public final String description;
+
+        private FunctionResult( UserFunctionSignature signature )
+        {
+            this.name = signature.name().toString();
+            this.signature = signature.toString();
+            this.description = signature.description().orElse( "" );
+        }
+    }
+
     public static class ProcedureResult
     {
         public final String name;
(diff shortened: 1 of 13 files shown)


community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
```diff
@@ -184,19 +184,6 @@ public void resampleOutdatedIndexes()
...
-            return statement.readOperations().proceduresGetAll()
-                    .stream()
-                    .sorted( ( a, b ) -> a.name().toString().compareTo( b.name().toString() ) )
-                    .map( ProcedureResult::new );
-        }
-    }
-
     @Description( "List all user functions in the DBMS." )
     @Procedure(name = "dbms.functions", mode = READ)
     public Stream<FunctionResult> listFunctions()
```
henriknyman: Shouldn't this thinking be applied to `dbms.functions` as well?

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/BuiltInProcedures.java
```diff
@@ -163,6 +168,51 @@ private KernelTransaction getCurrentTx()
...
+            {
+            case DBMS:
+                roles.add( "admin" );
+                break;
+            case READ_ONLY:
+                roles.add( "reader" );
+            case READ_WRITE:
+                roles.add( "publisher" );
+            case SCHEMA_WRITE:
+                roles.add( "architect" );
```
fickludd: So will this switch add the architect role to a `mode=READ` query? Do we want that?

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInDbmsProcedures.java
```diff
@@ -0,0 +1,62 @@
...
+import org.neo4j.kernel.impl.proc.Procedures;
+import org.neo4j.kernel.internal.GraphDatabaseAPI;
+import org.neo4j.procedure.Context;
+import org.neo4j.procedure.Description;
+import org.neo4j.procedure.Procedure;
+
+import static org.neo4j.procedure.Mode.DBMS;
+
+@SuppressWarnings( "unused" )
+public class BuiltInDbmsProcedures
```
fickludd, line 35: I think I see what you are doing here, but then maybe we could have the same class-name in enterprise for consistency? Atm. enterprise listProcedures is in `BuiltInProcedures`, while community lies in `BuiltInDbmsProcedures`.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ConfiguredProceduresTestBase.java
```diff
@@ -222,4 +233,48 @@ public void shouldHandleMultipleRolesSpecifiedForMapping() throws Throwable
...
+                "db.labels", newSet( "default", READER, PUBLISHER, ARCHITECT, ADMIN ),
+                "dbms.security.changePassword", newSet( ADMIN ),
+                "dbms.procedures", newSet( ADMIN ),
+                "dbms.listQueries", newSet( ADMIN ),
+                "dbms.security.createUser", newSet( ADMIN ) );
+
+        assertListProceduresHasRoles( adminSubject, expected );
+        assertListProceduresHasRoles( schemaSubject, expected );
+        assertListProceduresHasRoles( writeSubject, expected );
+        assertListProceduresHasRoles( writeSubject, expected );
```
fickludd: Over-assertion!

fickludd:
I think this is good.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
Mats-SX: You're right, that would make sense.

PR #8267, opened 2016-10-26
Fixed path calculations in `neo4j-admin dump` and `neo4j-admin load`

spacecowboy:
changelog: Fixed path calculations in `neo4j-admin dump` and `neo4j-admin load`
- specifying `--to=foo` which previously didn't work is now equivalent to `--to=./foo`
- now follows symlinks, in case the database directory is a symlink for example

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/Util.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/Util.java
@@ -51,7 +51,7 @@ public class Util
         }
         catch ( IOException e )
         {
-            throw new IllegalArgumentException( "Unable to parse path: " + file );
+            throw new IllegalArgumentException( "Unable to parse path: " + file, e );
         }
     }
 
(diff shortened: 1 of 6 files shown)


benbc:
LGTM

community/dbms/src/main/java/org/neo4j/commandline/dbms/Util.java
```diff
@@ -31,6 +33,28 @@
...
+
+    public static Path canonicalPath( File file ) throws IllegalArgumentException
+    {
+        try
+        {
+            return Paths.get( file.getCanonicalPath() );
+        }
+        catch ( IOException e )
+        {
+            throw new IllegalArgumentException( "Unable to parse path: " + file );
```
chrisvest: Please include the original exception… the kinds of failures that can pop out of here are platform specific, and we don't want to miss the root cause when customers bring us stack traces asking for an explanation.

srbaker: Consider doing this only when debug=true.

spacecowboy: - Stacktraces are only thrown when `debug=true` @srbaker 
- Will fix it so the original trace is shown in that case, just as @chrisvest suggests

spacecowboy:
Superceded by: https://github.com/neo4j/neo4j/pull/8288

PR #8270, opened 2016-10-27
Fix: bug with streaming of store files that grow during the streaming process

mneedham:
At the moment when a follower/read replica gets sent
store files we tell it how many bytes it should expect
to receive.

If the file size changes during the copy process we can
end up sending more bytes than we promised and the
client doesn't handle that very well and story copy fails.

With this commit we introduce our own implementation of
ReadableByteChannel which will only send the portion of the
file that we promised when we first determined its size.


mneedham:
We had one test failing because of a change of level of logging. The CI run is here: https://build.neohq.net/viewLog.html?buildId=1007632&buildTypeId=PullRequests_PullRequests&tab=buildLog

mneedham:
This was paired on with @apcj & passed the build except one unit test assertion on logging a message which I've now fixed

PR #8273, opened 2016-10-27
Fix buf leak in encoders

davidegrohmann:
Do not allocate buffers by hand but ask netty to allocate them for us,
so they can do the tracking and release them when they are not needed
anymore.


digitalstain (approved):
Yup, looks good.

lutovich:
@davidegrohmann it would also be great to run unit tests and ITs with Netty's leak detector turned on, as described here: http://netty.io/wiki/reference-counted-objects.html#troubleshooting-buffer-leaks

davidegrohmann:
@lutovich will do in another PR

apcj:
Superseded by https://github.com/neo4j/neo4j/pull/8284 so closing

PR #8278, opened 2016-10-28
Backport fix for `isolateAggregation`

Mats-SX:
The `isolateAggregation` rewriter was shaped up in `3.1` when
map projections were introduced. This commit backports the rewriter
updates to fix an issue that was fixed as part of that work.

changelog: Fix a bug where aliasing of aggregations caused queries to fail in semantic checking.


Mats-SX:
Note: the forward merge for this PR _will_ be slightly tricky. The bug exists in `3.0` as well, but _not_ in `3.1` (the changes here are copied from commits made in `3.1`). In case it's scary, ping me for handholding/delegating.

PR #8280, opened 2016-10-28
New stress tests for InternalFlatFileRealm

fickludd:
Also fix of bug in `InternalFlatFileRealm.newRole` were the initial users assigned to the new role were not checked for existance. This bug was never possible to encounter from the product surface.


henriknyman (approved):
This looks good as a basic starting point. It could later be extended with more stuff like weighted probabilities for actions, snapshot consistency checks mid-execution, repeated action patterns etc.

PR #8282, opened 2016-10-31
set-initial-password does nothing if real users exist

fickludd:
Neo4j-admin command `set-initial-password` does nothing if real users exist
- Gives warning to stdout that nothing was changed

### DIFF
--- a/community/security/src/main/java/org/neo4j/commandline/admin/security/SetInitialPasswordCommand.java
+++ b/community/security/src/main/java/org/neo4j/commandline/admin/security/SetInitialPasswordCommand.java
@@ -32,6 +32,7 @@ import org.neo4j.commandline.admin.OutsideWorld;
 import org.neo4j.dbms.DatabaseManagementSystemSettings;
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
 import org.neo4j.helpers.Args;
+import org.neo4j.io.fs.FileSystemAbstraction;
 import org.neo4j.kernel.configuration.Config;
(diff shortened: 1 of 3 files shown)


community/security/src/main/java/org/neo4j/commandline/admin/security/SetInitialPasswordCommand.java
```diff
@@ -125,20 +126,37 @@ private Args validateArgs( String[] args ) throws IncorrectUsage
     private void setPassword( String password ) throws Throwable
     {
         Config config = loadNeo4jConfig();
-        File file = CommunitySecurityModule.getInitialUserRepositoryFile( config );
-        if ( outsideWorld.fileSystem().fileExists( file ) )
+        if ( realUsersExist( config ) )
         {
-            outsideWorld.fileSystem().deleteFile( file );
+            outsideWorld.stdOutLine( "Warning: Initial password was not set because live Neo4j-users were " +
+                    "detected, so the initial password has no effect." );
```
henriknyman, line 132: I think a warning should go to `stderr`.

henriknyman (changes requested):

community/security/src/main/java/org/neo4j/commandline/admin/security/SetInitialPasswordCommand.java
henriknyman: Otherwise it looks good

henriknyman (approved):

PR #8290, opened 2016-11-01
Fail on missing configured auth plugin

fickludd:
If plugin authentication of authorization is enabled, but no respective auth plugin is found during service loading, the server will fail to start and give an appropriate error message.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/EditionModule.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/EditionModule.java
@@ -177,7 +177,7 @@ public abstract class EditionModule
                     } );
                     return;
                 }
-                catch ( KernelException e )
+                catch ( Exception e )
                 {
                     String errorMessage = "Failed to load security module.";
                     log.error( errorMessage );
--- a/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
+++ b/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.neo4j.dbms.DatabaseManagementSystemSettings;
 import org.neo4j.helpers.Service;
@@ -55,6 +56,7 @@ import org.neo4j.server.security.enterprise.configuration.SecuritySettings;
 import org.neo4j.server.security.enterprise.log.SecurityLog;
 import org.neo4j.time.Clocks;
 
+import static java.lang.String.format;
 import static org.neo4j.kernel.api.proc.Context.SECURITY_CONTEXT;
 
 @Service.Implementation( SecurityModule.class )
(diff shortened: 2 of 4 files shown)


enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
```diff
@@ -258,11 +263,30 @@ private static CacheManager createCacheManager( Config config )
...
+        List<PluginRealm> realms =
+                availablePluginRealms.stream()
+                        .filter( realm -> configuredRealms.contains( realm.getName() ) )
+                        .collect( Collectors.toList() );
+        boolean missingAuthenticationRealm = pluginAuthenticationEnabled &&
+                !realms.stream().anyMatch( PluginRealm::canAuthenticate );
+        boolean missingAuthorizingRealm = pluginAuthorizationEnabled &&
+                !realms.stream().anyMatch( PluginRealm::canAuthorize );
+
+        if ( missingAuthenticationRealm || missingAuthorizingRealm )
```
henriknyman, line 280: This is too strict. Since both settings are derived from `auth_providers` I think this will prevent you from combining auth providers.

henriknyman (changes requested):

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
fickludd: no, because this only fails if you explicitly put plugin_authentication_enabled=true, but no plugin authentication realm was loaded.

henriknyman: But plugin authentication enabled is a derived setting, so it will have a value of true if you configured a plugin auth provider

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/EditionModule.java
```diff
@@ -177,7 +177,7 @@ public DependencySatisfier dependencySatisfier()
                     } );
                     return;
                 }
-                catch ( KernelException e )
+                catch ( Exception e )
```
Mats-SX, line 180: If we're being general, why not `Throwable`?

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
```diff
@@ -143,16 +145,18 @@ public EnterpriseAuthAndUserManager newAuthManager( Config config, LogProvider l
             realms.add( new LdapRealm( config, securityLog, secureHasher ) );
         }
 
-        // Load plugin realms if we have any
-        realms.addAll( createPluginRealms( config, securityLog, secureHasher ) );
+        // Load plugin realms if a plugin provider in configured
```
Mats-SX, line 148: Typo in comment: `in` -> `is`

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
```diff
@@ -143,16 +145,18 @@ public EnterpriseAuthAndUserManager newAuthManager( Config config, LogProvider l
...
+        if ( configuredRealms.stream().anyMatch( ( r ) -> r.startsWith( SecuritySettings.PLUGIN_REALM_NAME_PREFIX ) ) )
```
Mats-SX, line 149: Parentheses around `r` in lambda superfluous.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/PluginRealm.java
```diff
@@ -224,6 +224,22 @@ private void cacheAuthorizationInfo( PluginAuthInfo authInfo )
         authorizationCache.put( key, authInfo );
     }
 
+    public boolean canAuthenticate()
+    {
+        return authPlugin != null || authenticationPlugin !=null;
```
Mats-SX, line 229: Bad style: missing space

Mats-SX:
Is this PR abandoned? @fickludd

PR #8292, opened 2016-11-01
More sanity checks on auth config

fickludd:
This PR adds and number of sanity checks to the enterprise security module config parsing.

 - If an auth plugin is present among the `auth_providers`, but that auth plugin is not found on startup (not service loaded), the server will fail to start and give an appropriate error message.
 - Asserts that all listed auth providers have one of `authentication_enabled` and `authorization_enabled` true
 - Asserts that at least on of the 3 `xxx_authentication_enabled` is true
 - Asserts that at least on of the 3 `xxx_authorization_enabled` is true

### DIFF
--- a/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
+++ b/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.neo4j.commandline.admin.security.SetDefaultAdminCommand;
 import org.neo4j.dbms.DatabaseManagementSystemSettings;
@@ -58,6 +59,7 @@ import org.neo4j.server.security.enterprise.configuration.SecuritySettings;
 import org.neo4j.server.security.enterprise.log.SecurityLog;
 import org.neo4j.time.Clocks;
 
+import static java.lang.String.format;
 import static org.neo4j.kernel.api.proc.Context.SECURITY_CONTEXT;
 
 @Service.Implementation( SecurityModule.class )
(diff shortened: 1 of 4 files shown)


enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
```diff
@@ -276,4 +285,70 @@ public static File getRoleRepositoryFile( Config config )
     {
         return new File( config.get( DatabaseManagementSystemSettings.auth_store_directory ), ROLE_STORE_FILENAME );
     }
+
+    private static IllegalArgumentException illegalConf( String message )
```
henriknyman: illegalConfiguration ?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModuleTest.java
```diff
@@ -19,83 +19,155 @@
...
-        );
+        nativeAuth( false, false );
+        ldapAuth( false, false );
+        pluginAuth( true, true );
+        authProviders( SecuritySettings.NATIVE_REALM_NAME, SecuritySettings.LDAP_REALM_NAME );
+
+        // Then
         thrown.expect( IllegalArgumentException.class );
+        thrown.expectMessage( "Illegal configuration: Native auth provider configured, " +
+                                "but both authentication and authorization are disabled." );
```
henriknyman, line 109: Do we need this? What if you just want to flip the enabled setting to override the default without changing the ordered list of auth providers?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModuleTest.java
```diff
@@ -19,83 +19,155 @@
...
+                SecuritySettings.PLUGIN_REALM_NAME_PREFIX + "TestAuthenticationPlugin",
+                SecuritySettings.PLUGIN_REALM_NAME_PREFIX + "IllConfiguredAuthorizationPlugin"
+            );
 
         // Then
-        verify( mockLog, atLeastOnce() ).debug( anyString(),
-                contains( "Illegal configuration: No valid security realm is active." ), anyString() );
+        thrown.expect( IllegalArgumentException.class );
+        thrown.expectMessage( "Illegal configuration: No plugin authorization provider loaded even though required by " +
+                "configuration." );
```
henriknyman: This is related to the comment in the previous PR of being to strict

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModuleTest.java
fickludd: I'm not entirely sure I know what you mean, so bear with me if I'm rumbling in the wrong direction here.

I think this check is good because it could save you a lot time debugging why your auth setup doesn't work if you somehow accidentally ended up such an inconsistent config, maybe after testing more complex setups. Since that wasted debug time could be long (minutes-hours), I prefer guarding against that above making temporary changes in the config slightly easier.

craigtaverner: I also favor easier debugging over the convenience of only changing one setting when re-configuring. I would assume changing config settings is not common in real deployment (only during testing and trying out by plugin authors), and the requirement to enable it in two places (order and boolean) is not so bad.

craigtaverner (approved):

PR #8295, opened 2016-11-02
Fix eagerness bug on property reads in the horizon

Mats-SX:
changelog: Fixes a bug with eagerness on property reads.

Previous to this commit, the eagerness analysis system has been working with
the assumption that the `QueryGraph` encodes all reads being done in a
`PlannerQuery`. This is not correct, the `QueryHorizon` may also encode reads.
This commit fixes eagerness issues related to property reads in the horizon,
when those conflict with writes being done in the `QueryGraph`.

Still left to do are reads made by functions in the horizon (`labels()`, `properties()`).
Additionally, potential conflicts between reads in horizon and subsequent
writes in the tail(s) of the PQs.

Also fixed for the rule planner.

### DIFF
--- a/community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/logical/Eagerness.scala
+++ b/community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/logical/Eagerness.scala
@@ -107,7 +107,8 @@ object Eagerness {
   def headWriteReadEagerize(inputPlan: LogicalPlan, query: PlannerQuery)
                        (implicit context: LogicalPlanningContext): LogicalPlan = {
     val alwaysEager = context.config.updateStrategy.alwaysEager
-    if (alwaysEager || query.tail.isDefined && writeReadConflictInHead(query, query.tail.get))
+    val conflictInHorizon = query.queryGraph.horizonOverlap(query.horizon)
+    if (alwaysEager || conflictInHorizon || query.tail.isDefined && writeReadConflictInHead(query, query.tail.get))
       context.logicalPlanProducer.planEager(inputPlan)
     else
       inputPlan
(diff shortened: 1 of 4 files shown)


community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/logical/Eagerness.scala
```diff
@@ -107,7 +107,8 @@ object Eagerness {
   def headWriteReadEagerize(inputPlan: LogicalPlan, query: PlannerQuery)
                        (implicit context: LogicalPlanningContext): LogicalPlan = {
     val alwaysEager = context.config.updateStrategy.alwaysEager
-    if (alwaysEager || query.tail.isDefined && writeReadConflictInHead(query, query.tail.get))
+    val conflictInHorizon = query.queryGraph.horizonOverlap(query.horizon)
+    if (alwaysEager || conflictInHorizon || query.tail.isDefined && writeReadConflictInHead(query, query.tail.get))
```
henriknyman, line 111: What about `writeReadConflictsInTail` ? Maybe it would be better to try to incorporate the horizon within `setPropertyOverlap` to cover all cases?

community/cypher/cypher/src/test/scala/org/neo4j/cypher/EagerizationAcceptanceTest.scala
```diff
@@ -44,6 +44,23 @@ class EagerizationAcceptanceTest
...
+                  |RETURN n.val AS nv, m.val AS mv
+                """.stripMargin
+
+    val result = updateWithBothPlanners(query)
+
+    result.toList should equal(List(Map("nv" -> 2, "mv" -> 2),
+                                    Map("nv" -> 2, "mv" -> 2)))
+    assertStats(result, propertiesWritten = 2)
+  }
+
```
henriknyman, line 63: I think we are missing a test case with a PlannerQuery tail.

Mats-SX: What do you mean precisely?

community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/logical/Eagerness.scala
Mats-SX: Do you mean when the horizon is in a tail and conflicts with writes made in a `PlannerQuery` one or several hops earlier?

PR #8297, opened 2016-11-02
Fix store copy on block device

chrisvest:
This was broken by a forward merge of a fix of another issue.


chrisvest:
Marking it as not ready for merge, because I don't think it currently moves anything _other_ than store files.

chrisvest:
@burqen @mneedham @digitalstain This should make store copy work on block device storage _and_ preserve the behaviour modifications you introduced in 3.0. Please take a look.

digitalstain:
I like this change, it fixes the known issues with the non default store copy strategy.

chrisvest:
The build is finished, even though the status is still yellow on GitHub. Probably a missed notification of some sort.

burqen (approved):
LGTM
Really nice comments explaining the asymmetric usage of `FileMoveAction` between record files and other store files.

PR #8300, opened 2016-11-03
Inlining of the file contents caused poor formatting.

srbaker:
Because of XML formatting, we were putting 12 blank spaces at the end of the config file.


benbc:
LGTM

PR #8301, opened 2016-11-03
neo4j-backup is gone, so we shouldn't install a wrapper in Debian.

srbaker:


benbc:
This isn't right. Sorry, the card is unclear. `neo4j-backup` still exists. But it has never been part of the Community package.

benbc:
Look good.

PR #8302, opened 2016-11-03
Ephemeral file system file buffer allocation simplification

MishaDemianenko:
Update ephemeral file system to use 1KB buffer for new file by default.
In case if file data size over exceed available buffer capacity it will be doubled till enough space will be available.

### DIFF
--- a/community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
+++ b/community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
@@ -534,7 +534,7 @@ public class EphemeralFileSystemAbstraction implements FileSystemAbstraction
     public long checksum()
     {
         Checksum checksum = new CRC32();
-        byte[] data = new byte[4096];
+        byte[] data = new byte[(int) ByteUnit.kibiBytes( 1 )];
 
         // Go through file name list in sorted order, so that checksum is consistent
         List<File> names = new ArrayList<>( files.size() );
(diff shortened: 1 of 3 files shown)


MishaDemianenko:
@burqen fixed and retarget to 3.2

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
```diff
@@ -1273,16 +1224,14 @@ private void verifySize( int totalAmount )
 
             // Double size each time, but after 1M only increase by 1M at a time, until required amount is reached.
             int newSize = buf.capacity();
-            int sizeIndex = sizeIndexFor( newSize );
-            while ( capacity( sizeIndex ) < totalAmount )
```
chrisvest, line 1277: The comment is now out of date. It might not be a bad idea to restrict the doubling a bit.

burqen: Do you think about any specific scenario for when continue doubling in size would be a bad idea?
I think it is a reasonable trade off for simple code.

burqen (changes requested):
@MishaDemianenko Review round finished. Let's wait for outcome from doubling-in-size-thread.

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstraction.java
MishaDemianenko: @chrisvest you thinking about having some maximum size that if we reach that we just throw an exception?

MishaDemianenko: @chrisvest add some basic check and updated obsolete comment

burqen (approved):

burqen:
@MishaDemianenko I'm happy with this change. How about you @chrisvest ?

chrisvest:
@burqen Yeah, looks good.

PR #8305, opened 2016-11-03
Add a version command to neo4j-admin

spacecowboy:
Help text is as follows:

```
usage: neo4j-admin version --store=<path-to-dir>

Checks the version of a Neo4j database store. Note that this command expects a
path to a store directory, for example --store=data/databases/graph.db.

options:
  --store=<path-to-dir>   Path to database store to check version of.
```

changelog [tools] `neo4j-admin`: added `version` command

This command can be pointed against a store and reports information
about its version. 

```
usage: neo4j-admin version --store=<path-to-dir>
```

Two example outputs are

- If store matches latest known version to the tool:

```
Store format version:    v0.A.7
Introduced in version:   3.0.0
Current version:         3.0.5
```

- If store is older than latest known version to the tool:

```
Store format version:    v0.A.5
Introduced in version:   2.2.0
Superseded in version:   2.3.0
Current version:         3.0.5
```

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
@@ -34,6 +34,7 @@ import org.neo4j.io.fs.DefaultFileSystemAbstraction;
 import org.neo4j.io.pagecache.PageCache;
 import org.neo4j.kernel.impl.pagecache.StandalonePageCacheFactory;
 import org.neo4j.kernel.impl.store.MetaDataStore;
+import org.neo4j.kernel.impl.store.format.FormatFamily;
 import org.neo4j.kernel.impl.store.format.RecordFormatSelector;
 import org.neo4j.kernel.impl.store.format.RecordFormats;
 import org.neo4j.kernel.impl.storemigration.StoreVersionCheck;
@@ -128,7 +129,8 @@ public class VersionCommand implements AdminCommand
 
         for ( RecordFormats candidate : RecordFormatSelector.allFormats() )
         {
-            if ( candidate.generation() <= format.generation() )
+            if ( !(FormatFamily.isSameFamily( format, candidate )) ||
+                    candidate.generation() <= format.generation() )
             {
                 continue;
             }
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
@@ -21,7 +21,6 @@ package org.neo4j.commandline.dbms;
 
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.Optional;
 import java.util.function.Consumer;
 
 import org.neo4j.commandline.admin.AdminCommand;
@@ -38,6 +37,7 @@ import org.neo4j.kernel.impl.store.format.RecordFormatSelector;
 import org.neo4j.kernel.impl.store.format.RecordFormats;
 import org.neo4j.kernel.impl.storemigration.StoreVersionCheck;
 import org.neo4j.kernel.impl.util.Validators;
+import org.neo4j.kernel.internal.Version;
 
 import static org.neo4j.kernel.impl.store.format.RecordFormatSelector.findSuccessor;
 
@@ -100,19 +100,19 @@ public class VersionCommand implements AdminCommand
             final String storeVersion = new StoreVersionCheck( pageCache )
                     .getVersion( storeDir.resolve( MetaDataStore.DEFAULT_NAME ).toFile() )
                     .orElseThrow(
-                            () -> new CommandFailed( String.format( "Could not find a version in '%s'", storeDir ) ) );
+                            () -> new CommandFailed( String.format( "Could not find version metadata in store '%s'",
+                                    storeDir ) ) );
 
             final String fmt = "%-20s%s";
             out.accept( String.format( fmt, "Store version:", storeVersion ) );
 
             RecordFormats format = RecordFormatSelector.selectForVersion( storeVersion );
-            out.accept( String.format( fmt, "Introduced in:", format.neo4jVersion() ) );
+            out.accept( String.format( fmt, "Introduced in:", format.firstNeo4jVersion() ) );
 
-            Optional<RecordFormats> nextFormat = findSuccessor( format );
-            if ( nextFormat.isPresent() )
-            {
-                out.accept( String.format( fmt, "Superceded in:", nextFormat.get().neo4jVersion() ) );
-            }
+            out.accept(
+                    findSuccessor( format )
+                            .map( next -> String.format( fmt, "Superseded in:", next.firstNeo4jVersion() ) )
+                            .orElse( String.format( fmt, "Used in:", Version.getNeo4jVersion() + " (current)" ) ) );
         }
         catch ( IOException e )
         {
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
@@ -103,16 +103,17 @@ public class VersionCommand implements AdminCommand
                             () -> new CommandFailed( String.format( "Could not find version metadata in store '%s'",
                                     storeDir ) ) );
 
-            final String fmt = "%-20s%s";
-            out.accept( String.format( fmt, "Store version:", storeVersion ) );
+            final String fmt = "%-25s%s";
+            out.accept( String.format( fmt, "Store format version:", storeVersion ) );
 
             RecordFormats format = RecordFormatSelector.selectForVersion( storeVersion );
-            out.accept( String.format( fmt, "Introduced in:", format.firstNeo4jVersion() ) );
+            out.accept( String.format( fmt, "Introduced in version:", format.introductionVersion() ) );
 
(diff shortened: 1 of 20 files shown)


spacecowboy:
@MishaDemianenko I have addressed your initial comments

community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
```diff
@@ -0,0 +1,122 @@
...
+            final String fmt = "%-20s%s";
+            out.accept( String.format( fmt, "Store version:", storeVersion ) );
+
+            RecordFormats format = RecordFormatSelector.selectForVersion( storeVersion );
+            out.accept( String.format( fmt, "Introduced in:", format.neo4jVersion() ) );
+
+            Optional<RecordFormats> nextFormat = findSuccessor( format );
+            if ( nextFormat.isPresent() )
+            {
+                out.accept( String.format( fmt, "Superceded in:", nextFormat.get().neo4jVersion() ) );
```
MishaDemianenko: typo: Superseded

community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
```diff
@@ -0,0 +1,122 @@
...
+                            () -> new CommandFailed( String.format( "Could not find a version in '%s'", storeDir ) ) );
+
+            final String fmt = "%-20s%s";
+            out.accept( String.format( fmt, "Store version:", storeVersion ) );
+
+            RecordFormats format = RecordFormatSelector.selectForVersion( storeVersion );
+            out.accept( String.format( fmt, "Introduced in:", format.neo4jVersion() ) );
+
+            Optional<RecordFormats> nextFormat = findSuccessor( format );
+            if ( nextFormat.isPresent() )
```
MishaDemianenko: What use map instead and say that format is superseded or the latest one (still newest)?

community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
```diff
@@ -0,0 +1,122 @@
...
+        final Path storeDir = arguments.parseMandatoryPath( "store", args );
+
+        Validators.CONTAINS_EXISTING_DATABASE.validate( storeDir.toFile() );
+
+        try ( PageCache pageCache = StandalonePageCacheFactory.createPageCache( new DefaultFileSystemAbstraction() ) )
+        {
+            final String storeVersion = new StoreVersionCheck( pageCache )
+                    .getVersion( storeDir.resolve( MetaDataStore.DEFAULT_NAME ).toFile() )
+                    .orElseThrow(
+                            () -> new CommandFailed( String.format( "Could not find a version in '%s'", storeDir ) ) );
```
MishaDemianenko: I find this message a bit cryptic, what about: Could not find version in database metadata file in store _store_ or something similar?

community/dbms/src/test/java/org/neo4j/commandline/dbms/VersionCommandTest.java
```diff
@@ -0,0 +1,192 @@
...
+    private Path databaseDirectory;
+    private ByteArrayOutputStream baos;
+    private VersionCommand command;
+    private DefaultFileSystemAbstraction fs;
+    private PageCache pageCache;
+
+    @Before
+    public void setUp() throws Exception
+    {
+        fs = new DefaultFileSystemAbstraction();
```
MishaDemianenko: We also have junit rules for file system abstraction and page cache: DefaultFileSystemRule and PageCacheRule

community/dbms/src/test/java/org/neo4j/commandline/dbms/VersionCommandTest.java
```diff
@@ -0,0 +1,192 @@
...
+        pageCache = StandalonePageCacheFactory.createPageCache( fs );
+        Path homeDir = testDirectory.directory( "home-dir" ).toPath();
+        databaseDirectory = homeDir.resolve( "data/databases/foo.db" );
+        Files.createDirectories( databaseDirectory );
+        baos = new ByteArrayOutputStream();
+        PrintStream printStream = new PrintStream( baos );
```
MishaDemianenko: Couple of thing here: you creating print stream without auto flushing, use  ByteArrayOutputStream.toString which will use system dependent encoding to produce final comparable string. What do you think if we will create rule for us that will gonna provide assertable string consumer instead where we will fix all those comments?

enterprise/neo4j-enterprise/src/test/java/org/neo4j/commandline/dbms/VersionCommandEnterpriseTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @Before
+    public void setUp() throws Exception
+    {
+        fs = new DefaultFileSystemAbstraction();
+        pageCache = StandalonePageCacheFactory.createPageCache( fs );
+        Path homeDir = testDirectory.directory( "home-dir" ).toPath();
+        databaseDirectory = homeDir.resolve( "data/databases/foo.db" );
+        Files.createDirectories( databaseDirectory );
+        baos = new ByteArrayOutputStream();
```
MishaDemianenko: same comments as above

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/store/format/highlimit/v300/HighLimitV3_0_0.java
```diff
@@ -68,6 +68,12 @@ public HighLimitV3_0_0()
     }
 
     @Override
+    public String neo4jVersion()
+    {
+        return "3.0.0";
```
MishaDemianenko: what instead of store this in each format we will add a field in StoreVersion (introducedIn field) for example and then we will just return that field value from each of the format?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/format/RecordFormats.java
```diff
@@ -49,6 +49,12 @@ public Factory( String key, String... altKeys )
     String storeVersion();
 
     /**
+     * @return the neo4j version where this format was introduced. It is almost certainly NOT the only version of
+     * neo4j where this format is used.
+     */
+    String neo4jVersion();
```
MishaDemianenko: should this be something more like introduced in version?

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreVersionCheck.java
```diff
@@ -38,32 +39,45 @@ public StoreVersionCheck( PageCache pageCache )
...
         {
             // since we cannot read let's assume the file is not there
             return new Result( Outcome.missingStoreFile, null, neostoreFile.getName() );
         }
 
-        if ( record == MetaDataRecordFormat.FIELD_NOT_PRESENT )
+        if ( storeVersion.isPresent() )
         {
-            return new Result( Outcome.storeVersionNotFound, null, neostoreFile.getName() );
+            if ( expectedVersion.equals( storeVersion.get() ) )
```
MishaDemianenko: looks like nice place to use `map`/`filter` operation instead

MishaDemianenko (changes requested):
Couple of comments

spacecowboy:
@MishaDemianenko please have another look at your earliest convenience. I think I addressed all the points.

There is one test left with a`ByteArrayOutputStream` which I want to keep, because I really want to test for the line-endings in that case. But I think that's fine since `flush()` is a `noop` on `ByteArrayOutputStream`.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/format/StoreVersion.java
```diff
@@ -24,25 +24,32 @@
...
     }
 
     public String versionString()
     {
         return versionString;
     }
+
+    public String firstNeo4jVersion()
+    {
+        return firstNeo4jVersion;
```
MishaDemianenko: i would really like if we will call this version `introductionVersion` or `introducedInVersion`. First version is not descriptive by itself and as soon as we use `first` it mean that we have some order in mind which is not clear.

community/dbms/src/main/java/org/neo4j/commandline/dbms/VersionCommand.java
```diff
@@ -0,0 +1,122 @@
...
+                    .getVersion( storeDir.resolve( MetaDataStore.DEFAULT_NAME ).toFile() )
+                    .orElseThrow(
+                            () -> new CommandFailed( String.format( "Could not find version metadata in store '%s'",
+                                    storeDir ) ) );
+
+            final String fmt = "%-20s%s";
+            out.accept( String.format( fmt, "Store version:", storeVersion ) );
+
+            RecordFormats format = RecordFormatSelector.selectForVersion( storeVersion );
+            out.accept( String.format( fmt, "Introduced in:", format.firstNeo4jVersion() ) );
```
MishaDemianenko: i think if we will change messages to: `Introduced in version:`, `Superseded in version:` and `Current version:` that will be more clear. WDYT?

MishaDemianenko (commented):
2 comments, all the rest looks great!

spacecowboy:
@MishaDemianenko I agree, changed according to feedback

MishaDemianenko (approved):

PR #8306, opened 2016-11-03
Improve the wording for advertised address in neo4j.conf.

srbaker:


packaging/standalone/standalone-community/src/main/distribution/text/community/conf/neo4j.conf
```diff
@@ -52,9 +52,8 @@ dbms.directories.import=import
 # You can also choose a specific network interface, and configure a non-default
 # port for each connector, by setting their individual listen_address.
 
-# Neo4j advertises the addresses its connectors.
-# To specify the hostname or IP address of this server used in the advertised addresses,
-# change this to the address of the network adapter you want to advertise.
+# Advertised address of connectors.
+# Change to the address that should be advertised for this server.
```
benbc: I don't think this is enormously helpful. How does someone know what address should be advertised? I'd put something about it being the address for this server which is reachable (or resolvable?) from the various clients that need to contact it. Maybe also not that it can be overridden for individual connectors elsewhere.

PR #8308, opened 2016-11-04
Produce a warning when dbms.pagecache.memory.size is not configured

chrisvest:
changelog The default heuristic value for the `dbms.pagecache.memory.size` setting has been changed to work better in deployments where the configured available heap memory is very small, relative to the available system memory. A warning is now written to the log, when the page cache memory has not been explicitly configured. In embedded deployments, the result of `config.get( GraphDatabaseSettings.pagecache_memory )` will now be `null` instead a the computed heuristic, in order to indicate when the setting has not been explicitly configured.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -37,7 +37,6 @@ import org.neo4j.kernel.configuration.Migrator;
 import org.neo4j.kernel.configuration.Settings;
 import org.neo4j.kernel.configuration.Title;
 import org.neo4j.kernel.impl.cache.MonitorGc;
-import org.neo4j.kernel.impl.util.OsBeanUtil;
 import org.neo4j.logging.Level;
 
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.BoltConnector.EncryptionLevel.OPTIONAL;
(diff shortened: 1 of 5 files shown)


community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -375,47 +374,7 @@
                   "the page cache. The default page cache memory assumes the machine is dedicated to running " +
                   "Neo4j, and is heuristically set to 50% of RAM minus the max Java heap size." )
```
MishaDemianenko: Now with default value as null with evaluation in a factory it's not clear from here what is a default value and where it's get calculated. 
As soon as you moved default value calculation from here to ConfiguringPageCacheFactory that makes that settings special in terms how we calculate default value and make it less clear.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
```diff
@@ -64,7 +66,8 @@ public long pageCacheMemory()
         {
             // Get the upper bound of what we can get from the default config calculation
             // We even want to limit amount of memory a bit more since we don't need very much during import
-            return min( MAX_PAGE_CACHE_MEMORY, Config.defaults().get( pagecache_memory ) );
+            String defaultPageCacheMemory = ConfiguringPageCacheFactory.defaultHeuristicPageCacheMemory();
```
MishaDemianenko: i also do not really like new way of evaluation of default value using `ConfiguringPageCacheFactory` and applying Bytes function to get the correct value, lets talk about it

MishaDemianenko (changes requested):
Couple of comments

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
chrisvest: I can make it return a `long` for bytes. No big deal.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
chrisvest: I'll update the text and make the default case more vague, so as to distance the setting from how the computation of the default page cache memory is implemented. This makes the documentation of the setting, only be about the setting.

chrisvest: The new computation is becoming too complicated to concisely explain anyway.

chrisvest:
@MishaDemianenko Please have another look.

MishaDemianenko (approved):

benbc:
@chrisvest I think this needs a changelog as it is effectively a change in the surface (we no longer support the default calculation).

chrisvest:
Thank you, @benbc. I have updated the description accordingly.

PR #8309, opened 2016-11-04
Bolt address with same origin as http address

maxsumrall:
SharedServerTestBase should ultimately not have those changes.

### DIFF
--- a/community/server-api/src/main/java/org/neo4j/server/rest/repr/Serializer.java
+++ b/community/server-api/src/main/java/org/neo4j/server/rest/repr/Serializer.java
@@ -76,11 +76,6 @@ abstract class Serializer
         list.done();
     }
 
-    final String relativeBoltUri( int boltPort )
-    {
-        return "bolt://" + baseUri.getHost() + ":" + boltPort;
-    }
-
     final String relativeUri(String path)
     {
         return joinBaseWithRelativePath(baseUri, path);
@@ -109,7 +104,7 @@ abstract class Serializer
         return result.append(path).toString();
     }
 
-    protected void checkThatItIsBuiltInType( Object value )
+    void checkThatItIsBuiltInType( Object value )
     {
         if ( !"java.lang".equals( value.getClass().getPackage().getName() ) )
         {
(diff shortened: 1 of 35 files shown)


community/server-api/src/main/java/org/neo4j/server/rest/repr/Serializer.java
```diff
@@ -76,6 +76,11 @@ final void serialize( ListWriter list, ListRepresentation value )
         list.done();
     }
 
+    final String relativeBoltUri(String boltPort){
+        String bolt = baseUri.toString().replaceFirst( baseUri.getScheme(), "bolt" );
```
apcj: This seems a bit backward to me. Why throw away the parsed URI and use string manipulation (replaceFirst, length etc.) when when you can neatly extract the host part out of the URI with a method call?

community/server/src/test/java/org/neo4j/server/rest/DiscoveryServiceIT.java
```diff
@@ -83,14 +83,39 @@ public void shouldRedirectOnHtmlRequest() throws Exception
...
+        // given
+        RestRequest request = new RestRequest( server().baseUri() ).host( "boltprotocol.org" );
+
+        // when
+        JaxRsResponse response = request.get();
+
+        // then
+        Map<String,Object> map = JsonHelper.jsonToMap( response.getEntity() );
+        assertThat( String.valueOf( map.get( "bolt" ) ),
+                containsString( "bolt://" + request.getHeaders().get("host") ) );
```
apcj: This doesn't seem like a very strong assertion. For example, does the URI have the correct port?

maxsumrall: Yes I'll amend it to also check for the port. The main concern was to test that the host part of the address was correct.

community/server-api/src/main/java/org/neo4j/server/rest/repr/Serializer.java
maxsumrall: Yes, I had also found that to be a better way to do this.

chrisvest:
@maxsumrall Looks like some extra commits snuck in by accident somehow.

jimwebber:
Following some conversation with @benbc we're going to affine this with the existing Bolt advertised address stuff such that an explicitly configured address has precedence over an inferred address. Until that's pushed, this remains not ready for merge.

benbc:
Fine by me.

PR #8312, opened 2016-11-04
Fail loudly when doing backup against an old version backup

spacecowboy:
Incremental backup has the following scenarios, when run against an
existing backup B:

1. B is "up to date" (incremental backup possible)
2. B is "too far out of date" (full backup required)
3. B is a backup of an older database store (because you just updated
   to 3.1 and your backup is of 3.0)
4. B is a backup of a different store but the same version

Previously, 3 and 4 were intended to behave differently.

4 would correctly fail with an error to avoid deleting someone's
precious backup files.

3 failed with an error due a bug in the error handling logic, the
intention was to move the old backup and then do a full
backup. However, another bug (masked by the error handling bug) would
have deleted that directory.

Regardless, the correct thing is to fail for 3. The primary reason
being that it is impossible (or unreasonably hard) to determine if
both 3 and 4 apply (e.g., it's a backup of a different store which
also has a different version). This because we start an embedded
database server as part of the backup, and that will always fail if
the store is of a different version.

### DIFF
--- a/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
@@ -262,9 +262,7 @@ class BackupService
         }
         catch ( RuntimeException e )
         {
-            Throwable cause = e.getCause() == null ? e : rootCause( e );
-
-            if ( cause instanceof UpgradeNotAllowedByConfigurationException )
+            if ( rootCause( e ) instanceof UpgradeNotAllowedByConfigurationException )
             {
                 throw new UnexpectedStoreVersionException( "Failed to perform backup because existing backup is from " +
                         "a different version.", e );
--- a/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
@@ -267,7 +267,7 @@ class BackupService
             if ( cause instanceof UpgradeNotAllowedByConfigurationException )
             {
                 throw new UnexpectedStoreVersionException( "Failed to perform backup because existing backup is from " +
-                        "a different version." );
+                        "a different version.", e );
             }
 
             throw e;
--- a/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
@@ -58,7 +58,9 @@ import org.neo4j.kernel.impl.logging.StoreLogService;
 import org.neo4j.kernel.impl.store.MetaDataStore;
 import org.neo4j.kernel.impl.store.MismatchingStoreIdException;
 import org.neo4j.kernel.impl.store.StoreId;
+import org.neo4j.kernel.impl.store.UnexpectedStoreVersionException;
 import org.neo4j.kernel.impl.store.id.IdGeneratorImpl;
+import org.neo4j.kernel.impl.storemigration.UpgradeNotAllowedByConfigurationException;
 import org.neo4j.kernel.impl.transaction.log.MissingLogDataException;
 import org.neo4j.kernel.impl.transaction.log.TransactionIdStore;
 import org.neo4j.kernel.impl.transaction.log.entry.VersionAwareLogEntryReader;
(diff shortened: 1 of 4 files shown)


enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
```diff
@@ -257,6 +260,18 @@ BackupOutcome doIncrementalBackupOrFallbackToFull( String sourceHostNameOrIp, in
                         "but that failed as well: '" + fullBackupFailure.getMessage() + "'.", fullBackupFailure );
             }
         }
+        catch ( RuntimeException e )
+        {
+            Throwable cause = e.getCause() == null ? e : rootCause( e );
```
benbc: `rootCause()` already handles the `null` case.

enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
```diff
@@ -257,6 +260,18 @@ BackupOutcome doIncrementalBackupOrFallbackToFull( String sourceHostNameOrIp, in
...
+
+            if ( cause instanceof UpgradeNotAllowedByConfigurationException )
+            {
+                throw new UnexpectedStoreVersionException( "Failed to perform backup because existing backup is from " +
+                        "a different version." );
```
benbc: Why not include the original exception as a cause to easy debugging?

enterprise/backup/src/main/java/org/neo4j/backup/BackupTool.java
```diff
@@ -243,6 +203,10 @@ private BackupOutcome doBackup( HostnamePort hostnamePort, File to, ConsistencyC
             systemOut.println( "Done" );
             return outcome;
         }
+        catch ( UnexpectedStoreVersionException e )
+        {
+            throw new ToolFailureException( e.getMessage() );
```
benbc: Also pass on underlying exception here.

enterprise/backup/src/test/java/org/neo4j/backup/BackupToolIT.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @After
+    public void teardown() throws Exception
+    {
+        pageCache.close();
+    }
+
+    @Test
+    public void oldIncompatibleBackupsAreMovedToSubDir() throws Exception
```
benbc: Not moved?

enterprise/backup/src/test/java/org/neo4j/backup/BackupToolIT.java
```diff
@@ -0,0 +1,107 @@
...
+        dbRule.getGraphDatabaseAPI();
+
+        expected.expect( BackupTool.ToolFailureException.class );
+        expected.expectMessage( "Failed to perform backup because existing backup is from a different version." );
+
+        // Perform backup
+        backupTool.executeBackup( new HostnamePort( "localhost", 6362 ), backupDir.toFile(),
+                ConsistencyCheck.NONE, Config.defaults(), 20L * 60L * 1000L, false );
+    }
+
```
benbc, line 92: Is it worth it/possible to add more cases here for the error cases that you're fixing?

spacecowboy: Like? As far as I know, it's one case, and I have one test for it.

benbc: Different store; different store + different version.

enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
spacecowboy: No, it explicitly does _not_ handle the null case:

``` java
public static Throwable rootCause( Throwable caughtException )
    {
        if ( null == caughtException )
        {
            throw new IllegalArgumentException( "Cannot obtain rootCause from (null)" );
        }
...
```

enterprise/backup/src/test/java/org/neo4j/backup/BackupToolIT.java
spacecowboy: `BackupServiceIT` has a test: `incrementalBackupShouldFailWhenTargetDirContainsDifferentStore` which covers that IMO.

enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
benbc: That's the wrong `null` case, isn't it? You're checking `e.getCause()`, which it copes with being `null`. The case you point out would be equivalent to `e` being `null`.

enterprise/backup/src/test/java/org/neo4j/backup/BackupToolIT.java
benbc: Okay. As long as you're happy we're testing everything we should.

enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
spacecowboy: true enough!

PR #8313, opened 2016-11-04
Fallback on interpreted runtime on unsupported type

henriknyman:
If the compiled runtime encounters a result type it cannot handle it should
fallback on the interpreted runtime rather than fail hard


henriknyman:
[Fallback on interpreted runtime on unsupported result type](https://trello.com/c/z7vBLMf0/187-fallback-on-interpreted-runtime-on-unsupported-result-type)

PR #8325, opened 2016-11-08
Add a close() method to the PageSwapperFactory interface

chrisvest:
This allows the page swapper factories to initialise internal resources in
`configure`, and release them in `close`.

This is useful for our block device integration.

### DIFF
--- a/community/io/src/main/java/org/neo4j/io/pagecache/PageSwapperFactory.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/PageSwapperFactory.java
@@ -129,4 +129,6 @@ public interface PageSwapperFactory
      * @throws IOException If an I/O error occurs, possibly with the canonicalisation of the paths.
      */
     Stream<FileHandle> streamFilesRecursive( File directory ) throws IOException;
+
+    void close();
 }
--- a/community/io/src/main/java/org/neo4j/io/pagecache/PageSwapperFactory.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/PageSwapperFactory.java
@@ -130,5 +130,12 @@ public interface PageSwapperFactory
      */
     Stream<FileHandle> streamFilesRecursive( File directory ) throws IOException;
 
+    /**
+     * Close and release any resources associated with this PageSwapperFactory, that it may have opened or acquired
+     * during its construction or use.
+     * <p>
+     * This method cannot be called before all of the opened {@link PageSwapper PageSwappers} have been closed,
+     * and it is guaranteed that no new page swappers will be created after this method has been called.
+     */
     void close();
 }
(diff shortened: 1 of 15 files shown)


chrisvest:
Note that there's a blockdevice PR that needs to be merged alongside this one.
/cc @burqen @magnusvejlstrup

community/io/src/main/java/org/neo4j/io/pagecache/PageSwapperFactory.java
```diff
@@ -129,4 +129,6 @@ PageSwapper createPageSwapper(
      * @throws IOException If an I/O error occurs, possibly with the canonicalisation of the paths.
      */
     Stream<FileHandle> streamFilesRecursive( File directory ) throws IOException;
+
+    void close();
```
burqen, line 140: Would be nice with some javadoc for this method and some documentation on the lifecycle of a `PageSwapperFactory`.

community/io/src/test/java/org/neo4j/adversaries/pagecache/AdversarialPageCache.java
```diff
@@ -98,9 +98,8 @@ public void flushAndForce( IOLimiter limiter ) throws IOException
     }
 
     @Override
-    public void close() throws IOException
+    public void close()
     {
-        adversary.injectFailure( FileNotFoundException.class, IOException.class, SecurityException.class );
```
burqen, line 103: Should we not inject `IllegalStateException` since we updated PageCache.close to throw that?

burqen (changes requested):
I have two inline comments and one general one:

There are some tests where we use `PageSwapperFactory` without a `PageCache` and in those cases we do not close it. This is not a problem with current implementation but looking at the code from an outside perspective I would expect an object with a close method to be closed after use and that it would be the responsibility of the user to do so.
My suggestion would be to take care of the cases where we do not close the `PageSwapperFactory`.

chrisvest:
@burqen There you go.

burqen:
@chrisvest Nice! Still wondering about not closing the swapper when we use it without the PageCache.

chrisvest:
@burqen I don't think that ever happens outside of tests, but now that you mention it, it should happen in those tests.

burqen:
@chrisvest You mean _not_ closing the Swapper in those tests is actually what we _want_ to do? I don't see why that would be.

chrisvest:
@burqen Done.

burqen:
@chrisvest In `ConfiguringPageCacheFactory` we create a swapper in constructor. This swapper will only be added to a PageCache instance if we call `createPageCache()`, otherwise swapper will remain open.

Any ideas?

burqen:
After offline discussion with @chrisvest we reached this conclusion:

This is a transitional change and will be re-implemented soon. Therefore we are fine with the technical dept added in `ConfiguringPageCacheFactory`.

PR #8322, opened 2016-11-08
Remove neo4j-backup deprecation, and mark neo4j-admin backup as experimental.

srbaker:
Before:
```
$ bin/neo4j-backup
WARNING: neo4j-backup is deprecated and support for it will be removed in a future version of Neo4j. Please use neo4j-admin backup.
Backup failed.
Specify target location with -to <target-directory>
```

and

```
usage: neo4j-admin backup [--from=<address>] --to=<backup-path>
                          [--check-consistency[=<true|false>]]
                          [--cc-report-dir=<directory>]
                          [--additional-config=<config-file-path>]
                          [--timeout=<timeout>]

Perform a backup, over the network, from a running Neo4j server into a local
copy of the database store (the backup). Neo4j Server must be configured to run
a backup service. See http://neo4j.com/docs/operations-manual/current/backup/
for more details.

options:
```

After:
```
$ bin/neo4j-backup
Backup failed.
Specify target location with -to <target-directory>
```

and

```
$ bin/neo4j-admin help backup
usage: neo4j-admin backup [--from=<address>] --to=<backup-path>
                          [--check-consistency[=<true|false>]]
                          [--cc-report-dir=<directory>]
                          [--additional-config=<config-file-path>]
                          [--timeout=<timeout>]

Perform a backup, over the network, from a running Neo4j server into a local
copy of the database store (the backup). Neo4j Server must be configured to run
a backup service. See http://neo4j.com/docs/operations-manual/current/backup/
for more details.

WARNING: this command is experimental and subject to change.

options:
...
```

PR #8328, opened 2016-11-08
Handle explicit multiline description strings

spacecowboy:
This also restores the previous formatting in the ImportCommand help text.


srbaker:
Merged with #8322

PR #8329, opened 2016-11-08
Respects aligned memory access constraint

tinwelint:
specifically in off-heap versions of primitive-collections and NumberArray.
UnsafeUtil provides means of knowing about aligned memory access constraint,
but doesn't automatically conforms access by it. Need for aligned memory access
is hardware dependent.

Currently the muninn page cache has some logic for this. Other users of
UnsafeUtil are the primitive-collections and NumberArray. They have different
constraints themselves than the page cache do and so they have been changed
to conform to alignment constraints, but can get away with doing less adaptations
than the page cache does. This is the primary reason UnsafeUtil doesn't do
these things automatically, because there are different optimizations that
apply to different use cases.

Original issue came up when noticing that using primitive-collections or
NumberArray on Solaris could crash the JVM.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/OffHeapByteArray.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/OffHeapByteArray.java
@@ -98,28 +98,54 @@ public class OffHeapByteArray extends OffHeapNumberArray<ByteArray> implements B
     @Override
     public short getShort( long index, int offset )
     {
-        return UnsafeUtil.getShort( address( index, offset ) );
+        return putShort( address( index, offset ) );
+    }
+
+    private short putShort( long p )
+    {
+        if ( UnsafeUtil.allowUnalignedMemoryAccess )
+        {
+            return UnsafeUtil.getShort( p );
+        }
+
+        return UnsafeUtil.getShortByteWise( p );
     }
 
     @Override
     public int getInt( long index, int offset )
     {
-        return UnsafeUtil.getInt( address( index, offset ) );
+        return getInt( address( index, offset ) );
+    }
+
+    private int getInt( long p )
+    {
+        if ( UnsafeUtil.allowUnalignedMemoryAccess )
+        {
+            return UnsafeUtil.getInt( p );
+        }
+
+        return UnsafeUtil.getIntByteWise( p );
     }
 
     @Override
     public long get6ByteLong( long index, int offset )
     {
         long address = address( index, offset );
-        long low4b = (UnsafeUtil.getInt( address )) & 0xFFFFFFFFL;
-        long high2b = UnsafeUtil.getShort( address + Integer.BYTES );
+        long low4b = getInt( address ) & 0xFFFFFFFFL;
+        long high2b = putShort( address + Integer.BYTES );
         return low4b | (high2b << 32);
     }
 
     @Override
     public long getLong( long index, int offset )
     {
-        return UnsafeUtil.getLong( address( index, offset ) );
+        long p = address( index, offset );
+        if ( UnsafeUtil.allowUnalignedMemoryAccess )
+        {
+            return UnsafeUtil.getLong( p );
+        }
+
+        return UnsafeUtil.getLongByteWise( p );
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/OffHeapByteArray.java
```diff
@@ -98,28 +98,54 @@ public byte getByte( long index, int offset )
     @Override
     public short getShort( long index, int offset )
     {
-        return UnsafeUtil.getShort( address( index, offset ) );
+        return putShort( address( index, offset ) );
+    }
+
+    private short putShort( long p )
```
chrisvest: Why is this called `putShort`?

community/unsafe/src/main/java/org/neo4j/unsafe/impl/internal/dragons/UnsafeUtil.java
```diff
@@ -333,6 +333,31 @@ public static long allocateMemory( long sizeInBytes )
...
+     * The returned pointer as aligned with {@code alignBy} such that {@code pointer % alignBy == 0}.
+     * The given pointer should be allocated with at least the requested size + {@code alignBy - 1},
+     * where the additional bytes will serve as padding for the worst case where the start of the usable
+     * area of the allocated memory will need to be shifted at most {@code alignBy - 1} bytes to the right.
+     *
+     * 0   4   8   12  16  20        ; 4-byte alignments
+     * |---|---|---|---|---|         ; memory
+     *        --------===            ; allocated memory (-required, =padding)
+     *         ^------^              ; used memory
+     *
```
chrisvest, line 348: Put this diagram in a `<pre><code>` block.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/OffHeapByteArray.java
chrisvest: Looks like it should have been a `getShort` overload.

chrisvest (changes requested):
Looks good, except for two things I'd like to see changed.

community/unsafe/src/main/java/org/neo4j/unsafe/impl/internal/dragons/UnsafeUtil.java
```diff
@@ -896,4 +921,113 @@ public static boolean exchangeNativeAccessCheckEnabled( boolean newSetting )
...
+    /**
+     * Gets a {@code short} at memory address {@code p} by reading byte for byte, instead of the whole value
+     * in one go. This can be useful, even necessary in some scenarios where {@link #allowUnalignedMemoryAccess}
+     * is {@code false} and {@code p} isn't aligned properly. Values read with this method should have been
+     * previously put using {@link #putShortByteWise(long, short)}.
+     *
+     * @param p address pointer to start reading at.
+     * @return the read value, which was read byte for byte.
+     */
+    public static short getShortByteWise( long p )
```
chrisvest: Actually I'd like these method names to also say what endianness they are using. It's good to be explicit about that in case they end up being used together with the non-byte-wise accesses.

chrisvest: Doing so would also highlight why the page cache has its own copies of these methods.

tinwelint: Good idea

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/OffHeapByteArray.java
tinwelint: Yup, I knew I renamed the wrong method at some point, but couldn't find it right then and then forgot about it

tinwelint:
@chrisvest Made an additional change, would be cool to try the `Squash and merge` option when merging because I do not think the second commit is of any use on its own after this review is over, do you?

chrisvest (approved):

PR #8338, opened 2016-11-10
Dump complete file path on shutdown failures

davidegrohmann:
This helps debugging shutdown failures on the page cache


chrisvest (approved):
Looks good.

PR #8340, opened 2016-11-10
Test robustification

tinwelint:


community/kernel/src/test/java/org/neo4j/helpers/HostnamePortTest.java
```diff
@@ -208,8 +208,7 @@ public void testMatchesUnknownHosts() throws Exception
     public void testMatchesKnownHostWithIP() throws Exception
     {
     	// Given
-    	
-    	String hostname1 = InetAddress.getLocalHost().getHostName();
+        String hostname1 = InetAddress.getLocalHost().getHostName().replace( '.', '-' );
```
tinwelint, line 211: Formatting looks odd here, but isn't. This is due to most lines in this file containing tabs and so it's the other lines that are wrong. The changed lines have spaces instead. I didn't want to change this entire file, for conflict reasons and it is most likely already fixed in 3.1 branch.

PR #8341, opened 2016-11-10
Considers page cache mem to heap ratio when calculating default

tinwelint:
A big difference between amount of physical memory and max heap memory
for the jvm could previously create a large page cache, large enough such
that the mere memory footprint of the page cache data structures inside
the heap would occupy so much as to cause space problems.

This added constraint restricts page cache size to 70x the heap size as
to not get into this problem.


tinwelint:
Merged into https://github.com/neo4j/neo4j/pull/8308 instead

PR #8343, opened 2016-11-10
NumberArrayFactory catches all errors in Auto

tinwelint:
so that errors like ArithmeticException due to trying to allocate a HEAP
array which is bigger than 2^32-1 will be caught and the next factory
attempted instead.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayFactory.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayFactory.java
@@ -26,6 +26,7 @@ import org.neo4j.helpers.Exceptions;
 import static java.lang.Math.toIntExact;
 import static java.lang.String.format;
 
+import static org.neo4j.helpers.Exceptions.launderedException;
 import static org.neo4j.helpers.Format.bytes;
 import static org.neo4j.unsafe.impl.batchimport.Utils.safeCastLongToInt;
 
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayFactory.java
```diff
@@ -212,60 +213,60 @@ public Auto( NumberArrayFactory... candidates )
...
                     return candidate.newLongArray( length, defaultValue, base );
                 }
-                catch ( OutOfMemoryError e )
+                catch ( Throwable e )
                 {   // Allright let's try the next one
                     error = e;
                 }
             }
-            throw error( length, 8, error );
+            throw launderedException( error( length, 8, error ) );
```
burqen, line 228: The error message produced by `error` seem to be a bit off with the new behaviour. It claims "Not enough memory available for allocating..." which is a bit misleading in the case of an `ArithmeticException`.
I would also like us to have a more generic message in the event of some other unexpected exception.

burqen (changes requested):
Looks good once fixing up error message.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayFactory.java
tinwelint: Something like: `"%s: ERROR when trying to allocate %s bytes, tried %s"`?

burqen: Yes, something like that. In the event of an `OutOfMemoryException` it would be ok to keep the old message, I don't really have a strong opinion about that, just as long as we have some reasonable message for every case (OutOfMemory, Arithmetic, other).

tinwelint: We do still throw the original throwable, so if it's `OutOfMemoryError` then the message will include that, likewise for the `ArithmeticException`. It's just that the added information to the message is now more generic. Makes sense?

burqen:
Yes, make sense!

burqen:
@tinwelint Friendly reminder about this PR

tinwelint:
@burqen It's rather unclear to me what's the remaining items under discussion, could you enlighten me?

burqen:
@tinwelint If we have an `ArithmeticException` we will still use the error message `Not enough memory available for allocating...`, which has nothing to do with the real error. We will, as you say, include the original exception and from there it will be clear what happened, but it would be nice to not have the error message claiming a different problem than what is actually the case.

tinwelint:
@burqen in a way it has to do with that because it's in and around allocating memory. I think we're splitting hairs here...

burqen (approved):

burqen:
@tinwelint alright, will merge as soon as GA goes out

PR #8344, opened 2016-11-10
Moved 8m surefire config for page cache to root pom.xml

tinwelint:
so that it will apply to all testing. Tests that want bigger page cache
will have to explicitly specify that.

### DIFF
--- a/community/collections/src/main/java/org/neo4j/helpers/collection/Iterators.java
+++ b/community/collections/src/main/java/org/neo4j/helpers/collection/Iterators.java
@@ -724,4 +724,17 @@ public abstract class Iterators
         }
         return sb.toString();
     }
+
+    public static <T> PrefetchingIterator<T> prefetching( Iterator<T> iterator )
+    {
+        return iterator instanceof PrefetchingIterator ? (PrefetchingIterator<T>) iterator :
+            new PrefetchingIterator<T>()
+            {
+                @Override
+                protected T fetchNextOrNull()
+                {
+                    return iterator.hasNext() ? iterator.next() : null;
+                }
+            };
+    }
 }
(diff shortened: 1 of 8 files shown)


chrisvest:
@tinwelint I think we can safely lower the lower-bound on the `GraphDatabaseSettingsTest.mustHaveReasonableDefaultPageCacheMemorySizeInBytes` to make that one pass.

I don't know what's wrong with the other test. I think it should be investigated.

tinwelint:
@chrisvest made that change. The other test was actually a real issue, which is now fixed as part of this PR, see b6d92e25c30e32aff92c0521b2c4da52f109a635

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ProducerStep.java
```diff
@@ -56,6 +56,7 @@ public void run()
                 }
                 catch ( Throwable e )
                 {
+                    e.printStackTrace();
```
chrisvest: Are you sure?

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/RelationshipGroupDefragmenterTest.java
```diff
@@ -213,6 +213,8 @@ private void verifyGroupsAreSequentiallyOrderedByNode()
                 continue;
             }
 
+//            System.out.println( "read:" + groupRecord );
+
```
chrisvest: This can go away

tinwelint:
@chrisvest a bit sloppy, removed those two lines

PR #8346, opened 2016-11-11
DumpLogicalLog can dump transactions relevant to cc report

tinwelint:
after a consisistency checker has run a report is made, potentially including
incosistencies. DumpLogicalLog can be given a --ccfilter which should
point to such a report file. DumpLogicalLog will then only include transactions
which touches any of the records in the cc report and mark the records it
matched with a `<----` arrow to easier navigate.

### DIFF
--- a/tools/src/main/java/org/neo4j/tools/dump/DumpLogicalLog.java
+++ b/tools/src/main/java/org/neo4j/tools/dump/DumpLogicalLog.java
@@ -25,14 +25,21 @@ import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.ByteBuffer;
 import java.util.TimeZone;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
+import org.neo4j.collection.primitive.Primitive;
+import org.neo4j.collection.primitive.PrimitiveLongSet;
 import org.neo4j.cursor.IOCursor;
 import org.neo4j.helpers.Args;
 import org.neo4j.io.fs.DefaultFileSystemAbstraction;
 import org.neo4j.io.fs.FileSystemAbstraction;
 import org.neo4j.io.fs.StoreChannel;
+import org.neo4j.kernel.impl.transaction.command.Command.NodeCommand;
+import org.neo4j.kernel.impl.transaction.command.Command.PropertyCommand;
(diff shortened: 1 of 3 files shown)


tools/src/main/java/org/neo4j/tools/dump/DumpLogicalLog.java
```diff
@@ -172,32 +183,160 @@ public boolean test( LogEntry[] transaction )
...
             {
                 new DumpLogicalLog( new DefaultFileSystemAbstraction() )
-                        .dump( fileAsString, printer.getFor( fileAsString ), timeZone, regex );
+                        .dump( fileAsString, printer.getFor( fileAsString ), filter, serializer );
             }
         }
     }
 
+    @SuppressWarnings( "unchecked" )
+    private static Function<LogEntry,String> parseSerializer( Args arguments, Predicate<LogEntry[]> filter,
```
MishaDemianenko: `arguments` not used here

tools/src/main/java/org/neo4j/tools/dump/InconsistencyReportReader.java
```diff
@@ -0,0 +1,149 @@
...
+            else if ( state == 1 )
+            {
+                String entityType = entityType( line );
+                long id = id( line );
+                if ( entityType != null && id != -1 )
+                {
+                    propagate( entityType, id );
+                }
+            }
+            else if ( state == 1 || state == 2 )
```
MishaDemianenko: state == 1 is impossible here

tools/src/main/java/org/neo4j/tools/dump/InconsistencyReportReader.java
```diff
@@ -0,0 +1,149 @@
...
+            if ( state == 0 )
+            {
+                if ( !line.contains( "ERROR" ) && !line.contains( "WARNING" ) )
+                {
+                    break;
+                }
+            }
+            else if ( state == 1 )
+            {
+                String entityType = entityType( line );
```
MishaDemianenko: entity type load and propagation (line 84- 89) can be extracted since exactly the same code used in next brach

MishaDemianenko (changes requested):
Couple of comments, otherwise really nice idea!

tinwelint:
@MishaDemianenko made some changes

MishaDemianenko (approved):

PR #8356, opened 2016-11-13
More cache friendly traversal iterators

tinwelint:
By changing two things:
- No eager relationship cursor initialization. Previously a CursorRelationshipIterator was always
  one step ahead, i.e. when constructing it the first item was read, when going to the first item
  the second item was read, a.s.o. In a traversal scenario this would mean that many of the
  eagerly read relationships would be read in vain because the paths would never even be explored.
- No eager expansion of traversal branches. Previously a TraversalBranch would be expanded when
  initializing it, i.e. when simply discovering it. Decision about when to actually traverse to it
  would be made by a BranchOrderingPolicy. To immediately expand a discovered path would only be  benefitial if that branch would be selected as the immediate next branch, otherwise it would
  - (a) be potentially an unnecessary read (of one node, but also its first relationship given the
    previously eager relationship cursor as explained above
  - (b) remove some of the benefits of physically co-located relationships, again depending on
    branch ordering, but would remove the benefits in all scenarios

All in all this change results in less node/relationship records read and opportunity to read
relationship chains more consecutively without interleaved "other" records, during a traversal.
Effects of these optimizations depend on the dataset and traversal, but worst case is the same or
slightly better and best case is much better.

As an example in a breadth-first outgoing traversal with outgoing relationships physically co-located,
as can be after e.g. an import, performance was increased by 30%.

### DIFF
--- a/community/graph-algo/src/test/java/org/neo4j/graphalgo/path/TestAStar.java
+++ b/community/graph-algo/src/test/java/org/neo4j/graphalgo/path/TestAStar.java
@@ -120,7 +120,7 @@ public class TestAStar extends Neo4jAlgoTestCase
         graph.makeEdge( "c", "end", "length", 4L );
         graph.makeEdge( "start", "d", "length", (short)2 );
         graph.makeEdge( "d", "e", "length", (byte)3 );
-        graph.makeEdge( "e", "end", "length", (int)2 );
+        graph.makeEdge( "e", "end", "length", 2 );
 
         // WHEN
         WeightedPath path = finder.findSinglePath( start, end );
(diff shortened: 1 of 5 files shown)


MishaDemianenko:
@tinwelint there are couple of failing tests otherwise looks good

tinwelint:
@MishaDemianenko Fixed those, it was an assumption that more branches would be eagerly expanded. Now with this change they aren't. Sorry did a force-push, the tiny change is in `TestAStar` itself.

MishaDemianenko:
@tinwelint hm...can you maybe try to rebase pr to see if build failure get fixed already?

MishaDemianenko:
@tinwelint looks like tools are happy now, will merge on green

PR #8358, opened 2016-11-14
Fix flaky TransactionThroughMasterSwitchStressIT

burqen:
Master go down
Server 2 is waiting for heartbeat
Master go up and re-election happens.
Server 2 will not join cluster because he still think old master is master and he's waiting for his heartbeat.
Because the heartbeat time (40 s) is larger than join cluster timeout (30 s)
it can happen that the cluster fails to form before Server 2 realise that he should re-join cluster.

Setting heartbeat timeout to a lower value should fix this.
Adding additional allSeesAllAsAvailable should make it more clear where we fail in the future, if this fix is not correct.

### DIFF
--- a/enterprise/ha/src/test/java/org/neo4j/kernel/ha/transaction/TransactionThroughMasterSwitchStressIT.java
+++ b/enterprise/ha/src/test/java/org/neo4j/kernel/ha/transaction/TransactionThroughMasterSwitchStressIT.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.neo4j.cluster.ClusterSettings;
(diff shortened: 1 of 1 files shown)


chrisvest:
@burqen Should we reflect on the default timeout settings based on this?

burqen:
@chrisvest Possibly.
But, in the normal case operation will proceed with Master and Server 1 being the only once available in the cluster. Server 2 will eventually fail and re-join the cluster. It's just a matter ClusterRule being a little bit picky about `allSeeAllAsAvailable` when starting up.

To be honest I don't know how `heartbeat_timeout` and `join_timeout` depend on each other and how the defaults was settled upon. Experimenting with new timeouts should wait until 3.1 GA goes out anyway.

Do you have any suggestions on how to go about this?

martinfurmanski:
Have you seen that join_timeout is 60s in ClusterRule?

burqen:
@martinfurmanski No, I have not, but you are right. Back to the drawing board. Closing.

PR #8360, opened 2016-11-14
Handled `UNWIND` in prettifier so that it follows the Cypher style guide

petraselmer:


community/cypher/cypher-compiler-3.0/src/test/scala/org/neo4j/cypher/internal/compiler/v3_0/prettifier/PrettifierTest.scala
```diff
@@ -149,10 +149,18 @@ class PrettifierTest extends CypherFunSuite {
     )
   }
 
-  test("should handle call yield") {
+  test("should handle CALL YIELD") {
     actual("match (n) call db.indexes yield state RETURN *") should equal(expected("MATCH (n)%nCALL db.indexes YIELD state%nRETURN *"))
   }
 
+  test("MERGE should start on a new line") {
```
Mats-SX, line 156: This test is more about `RETURN` starting on a new line, isn't it?

systay: You are right, but that test was wrongly named even before this PR. Let's not complect the two.

petraselmer: @Mats-SX Indeed! Thanks for spotting :) Fixed by AT.

PR #8361, opened 2016-11-14
Make store copy not rely on file size for sending/receiving data

davidegrohmann:
We can't control how files grow/shrink during store copy so better to
not rely on file sizes when streaming files around the cluster.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetStoreRequestHandler.java
```diff
@@ -84,13 +83,58 @@ private void sendFiles( ChannelHandlerContext ctx ) throws IOException
...
-    private void sendFile( ChannelHandlerContext ctx, File file ) throws FileNotFoundException
+    private void sendFile( ChannelHandlerContext ctx, File file ) throws IOException
     {
         ctx.writeAndFlush( ResponseMessageType.FILE );
-        long initialLength = file.length();
-        ctx.writeAndFlush( new FileHeader( file.getName(), initialLength ) );
-        ctx.writeAndFlush( new ChunkedNioStream( new LimitedLengthReadableByteChannel(
-                new FileInputStream( file ).getChannel(), initialLength ) ) );
+        ctx.writeAndFlush( new FileHeader( file.getName() ) );
+        ByteBuffer buffer = ByteBuffer.allocate( FileChunk.DEFAULT_SIZE );
```
chrisvest: If you use `allocateDirect` there will be one less memory copy in `channel.read`.

PR #8367, opened 2016-11-15
Fix foreach with bound node variable

henriknyman:
changelog: Fixes a bug whereby previously-bound nodes were created when using `FOREACH` on a mixed-type collection

Foreach on a mixed/Any type collection containing nodes should not
create nodes already bound by the loop variable.

Switches to use SemanticTable.isNode(_:String) which correctly infers node type.


community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/ast/convert/plannerQuery/ClauseConverters.scala
```diff
@@ -420,7 +420,7 @@ object ClauseConverters {
     val currentlyAvailableVariables = builder.currentlyAvailableVariables
 
     val setOfNodeVariables =
-      if (builder.semanticTable.isNode(clause.variable))
+      if (builder.semanticTable.isNode(clause.variable.name))
```
thobe, line 423: ouch, embarrassing.

thobe:
Looks good. Very good indeed.

PR #8375, opened 2016-11-15
Simplify the way we do transaction polling.

mneedham:
We don't need to have multiple threads competing
to send a TX pull request. Instead lets just have
one thread.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/TxPollingClient.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/TxPollingClient.java
@@ -20,6 +20,7 @@
 package org.neo4j.causalclustering.catchup.tx;
 
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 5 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/TxPollingClient.java
```diff
@@ -66,18 +70,23 @@
...
-    private final RenewableTimeoutService timeoutService;
     private final long txPullIntervalMillis;
     private final BatchingTxApplier applier;
     private final PullRequestMonitor pullRequestMonitor;
 
-    private RenewableTimeout timeout;
+    private volatile JobScheduler.JobHandle handle;
+    private final JobScheduler scheduler;
+    private volatile boolean stopped;
+    private volatile boolean polling;
```
davidegrohmann, line 80: I don't think we need this boolean anymore.

mneedham:
Superseded by #8401

PR #8376, opened 2016-11-16
Fix join failures

martinfurmanski:
A rare occurrence of events could cause a conflict in proposals
and because of a missing join event to the ProperState it wouldn't
retry.


davidegrohmann:
@digitalstain @martinfurmanski are we gonna merge this PR in? or close it?

digitalstain:
The bahaviour change introduced with this patch is not clearly understood. As such, some more testing is required before merging it. Closing for now, we'll reconsider this when we know more about the underlying bug.

PR #8384, opened 2016-11-16
Add more logging to LocalDatabase

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
```diff
@@ -71,13 +75,15 @@ public void init() throws Throwable
     public synchronized void start() throws Throwable
     {
         storeId = readStoreIdFromDisk();
+        log.info( "Starting with storeId: " + storeId );
```
davidegrohmann, line 78: maybe debug?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
```diff
@@ -71,13 +75,15 @@ public void init() throws Throwable
...
         dataSourceManager.start();
         started = true;
     }
 
     @Override
     public synchronized void stop() throws Throwable
     {
+        log.info( "Stopping" );
```
davidegrohmann, line 86: maybe debug?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
martinfurmanski: I think info is good. This is really low volume and useful.

PR #8385, opened 2016-11-17
Make tests safe to run on different OSs

systay:
Before this commit, our tests were safe to run on the same OS as they were built on. 
With this change, you can create the artefacts on one OS, and then run the tests on a different one, something that was not supported earlier.


thobe:
I suspect this is backwards in two ways:
1. Windows line endings are `\r\n`, not `\n\r` - this I know for sure.
2. I think we should convert the line endings to the current platform line ending rather than always converting to unix line endings - this I only suspect.

Mats-SX:
It looks like the test failure is in the module `debian installers`, which I don't think makes sense on Windows anyway. This should probably be mergable?

Mats-SX:
Rebased.

PR #8388, opened 2016-11-17
Fix bug on close in NeoStores

davidegrohmann:
If any store does not close cleanly but throws an exception, the
subsequents stores in store array are not gonna be closed.  This will
make the page cache to fail when shutdown due to still store files
mapped in it.  This change will make sure we attempt to close all the
stores even if exceptions are thrown.

### DIFF
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
@@ -58,10 +58,12 @@ import org.neo4j.kernel.impl.api.RelationshipVisitor;
 import org.neo4j.kernel.impl.core.RelationshipTypeToken;
 import org.neo4j.kernel.impl.storageengine.impl.recordstorage.RecordStorageEngine;
 import org.neo4j.kernel.impl.store.MetaDataStore.Position;
+import org.neo4j.kernel.impl.store.counts.CountsTracker;
 import org.neo4j.kernel.impl.store.format.RecordFormatSelector;
 import org.neo4j.kernel.impl.store.format.standard.DynamicRecordFormat;
 import org.neo4j.kernel.impl.store.id.DefaultIdGeneratorFactory;
 import org.neo4j.kernel.impl.store.id.IdType;
+import org.neo4j.kernel.impl.store.kvstore.RotationTimeoutException;
 import org.neo4j.kernel.impl.store.record.PropertyBlock;
 import org.neo4j.kernel.impl.store.record.PropertyKeyTokenRecord;
 import org.neo4j.kernel.impl.store.record.PropertyRecord;
@@ -84,12 +86,15 @@ import org.neo4j.test.rule.PageCacheRule;
 import org.neo4j.test.rule.TestDirectory;
 import org.neo4j.test.rule.fs.EphemeralFileSystemRule;
 
+import static java.util.Collections.singletonMap;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.neo4j.graphdb.factory.GraphDatabaseSettings.counts_store_rotation_timeout;
 import static org.neo4j.helpers.collection.MapUtil.stringMap;
 import static org.neo4j.kernel.api.security.SecurityContext.AUTH_DISABLED;
 import static org.neo4j.kernel.impl.store.RecordStore.getRecord;
@@ -713,6 +718,46 @@ public class NeoStoresTest
             assertEquals( new TransactionId( 40, 4444, BASE_TX_COMMIT_TIMESTAMP ),
                     store.getLastCommittedTransaction() );
             assertArrayEquals( store.getLastClosedTransaction(), new long[]{40,0,LogHeader.LOG_HEADER_SIZE} );
+        }
+    }
+
+    @Test
+    public void shouldCloseAllTheStoreEvenIfExceptionsAreThrown() throws Exception
+    {
+        // given
+        FileSystemAbstraction fileSystem = fs.get();
+        Config defaults = Config.defaults().with( singletonMap( counts_store_rotation_timeout.name(), "10ms" ) );
+        StoreFactory factory =
+                new StoreFactory( storeDir, defaults, new DefaultIdGeneratorFactory( fileSystem ), pageCache,
+                        fileSystem, NullLogProvider.getInstance() );
+        NeoStores neoStore = factory.openAllNeoStores( true );
+
+        // let's hack the counts store so it fails to rotate and hence it fails to close as well...
+        final CountsTracker counts = neoStore.getCounts();
+        counts.start();
+        try
+        {
+            long nextTxId = neoStore.getMetaDataStore().getLastCommittedTransactionId() + 1;
+            counts.rotate( nextTxId );
+            fail( "should have thrown" );
+        }
+        catch ( RotationTimeoutException ex )
+        {
+            // expected
+        }
+
+        try
+        {
+            // when we close the stores...
+            neoStore.close();
+            fail( "should have thrown" );
+        }
+        catch ( IllegalStateException ex )
+        {
+            // then
+            assertEquals( "Cannot stop in state: rotating", ex.getMessage() );
+            // and the page cache closes with no errors
+            pageCache.close();
         }
     }
 
--- a/community/kernel/src/main/java/org/neo4j/helpers/Exceptions.java
+++ b/community/kernel/src/main/java/org/neo4j/helpers/Exceptions.java
@@ -19,19 +19,14 @@
  */
 package org.neo4j.helpers;
 
-import java.io.ByteArrayOutputStream;
-import java.io.PrintStream;
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.io.UnsupportedEncodingException;
 import java.lang.Thread.State;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.util.function.Predicate;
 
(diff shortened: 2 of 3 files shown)


community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -703,16 +705,54 @@ public void shouldNotSetHighestTransactionIdWhenNeeded() throws Throwable
...
+                        fileSystem, NullLogProvider.getInstance() );
+        NeoStores neoStore = factory.openAllNeoStores( true );
+
+        // let's hack the counts store so it fails to rotate and hence it fails to close as well...
+        final CountsTracker counts = neoStore.getCounts();
+        counts.start();
+        try
+        {
+            long nextTxId = neoStore.getMetaDataStore().getLastCommittedTransactionId() + 1;
+            counts.rotate( nextTxId );
```
tinwelint, line 741: should you have a `fail(...)` here?

tinwelint:
It's difficult to see the actual changes since 90% is checkstyle. For future commits, would it be possible to keep that separate?

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -107,8 +112,8 @@
     private final NeoStoreDataSourceRule dsRule = new NeoStoreDataSourceRule();
 
     @Rule
-    public RuleChain ruleChain = RuleChain.outerRule( exception ).around( pageCacheRule )
-            .around( fs ).around( dir ).around( dsRule );
+    public RuleChain ruleChain =
+            RuleChain.outerRule( exception ).around( pageCacheRule ).around( fs ).around( dir ).around( dsRule );
```
chrisvest: I think the file system rule should be around the page cache rule, so we don't close the file system before we close the files.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -485,14 +485,17 @@ public void setVersion() throws Exception
...
-        assertEquals( 0, MetaDataStore.setRecord( pageCache, new File( storeDir,
-                MetaDataStore.DEFAULT_NAME ).getAbsoluteFile(), Position.LOG_VERSION, 10 ) );
-        assertEquals( 10, MetaDataStore.setRecord( pageCache, new File( storeDir,
-                MetaDataStore.DEFAULT_NAME ).getAbsoluteFile(), Position.LOG_VERSION, 12 ) );
+        assertEquals( 0, MetaDataStore
+                .setRecord( pageCache, new File( storeDir, MetaDataStore.DEFAULT_NAME ).getAbsoluteFile(),
+                        Position.LOG_VERSION, 10 ) );
+        assertEquals( 10, MetaDataStore
+                .setRecord( pageCache, new File( storeDir, MetaDataStore.DEFAULT_NAME ).getAbsoluteFile(),
+                        Position.LOG_VERSION, 12 ) );
```
chrisvest: Not a fan of breaking at the dot so far in. Can you break at the start of the parameter list instead?

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -679,16 +681,16 @@ public void shouldSetHighestTransactionIdWhenNeeded() throws Throwable
         try ( NeoStores neoStore = factory.openAllNeoStores( true ) )
         {
             MetaDataStore store = neoStore.getMetaDataStore();
-            store.setLastCommittedAndClosedTransactionId( 40, 4444, BASE_TX_COMMIT_TIMESTAMP,
-                    LogHeader.LOG_HEADER_SIZE, 0 );
+            store.setLastCommittedAndClosedTransactionId( 40, 4444, BASE_TX_COMMIT_TIMESTAMP, LogHeader.LOG_HEADER_SIZE,
+                    0 );
```
chrisvest: For cases such as this, I'd also prefer breaking at the start of the parameter list, or introduce a static import for the constant.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -703,16 +705,54 @@ public void shouldNotSetHighestTransactionIdWhenNeeded() throws Throwable
...
+        }
+        catch ( RotationTimeoutException ex )
+        {
+            // expected
+        }
+
+        try
+        {
+            // when we close the stores...
+            neoStore.close();
```
chrisvest, line 752: You also need a fail here, since you'd skip the asserts and checks in the catch clause if close doesn't throw.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -880,7 +916,7 @@ private int validateAndCountRelationships( long node, long rel1, long rel2, int
     {
         int count = 0;
         try ( KernelStatement statement = (KernelStatement) tx.acquireStatement();
-              Cursor<NodeItem> nodeCursor = statement.getStoreStatement().acquireSingleNodeCursor( node ) )
+                Cursor<NodeItem> nodeCursor = statement.getStoreStatement().acquireSingleNodeCursor( node ) )
```
chrisvest: This extra indentation looks like a mistake. Resource declarations look better when they are aligned in a try-with-resources statement.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -911,15 +947,15 @@ else if ( rel == rel2 )
         return count;
     }
 
-    private PropertyReceiver<DefinedProperty> newPropertyReceiver( final ArrayMap<Integer,Pair<DefinedProperty,Long>> props )
+    private PropertyReceiver<DefinedProperty> newPropertyReceiver(
+            final ArrayMap<Integer,Pair<DefinedProperty,Long>> props )
     {
-        return ( property, propertyRecordId ) -> props.put( property.propertyKeyId(), Pair.of( property, propertyRecordId ) );
+        return ( property, propertyRecordId ) -> props
+                .put( property.propertyKeyId(), Pair.of( property, propertyRecordId ) );
```
chrisvest: Here I'd prefer to either break before `props`, or at the start of the parameter list.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -962,7 +995,7 @@ else if ( data.propertyKeyId() == prop3.propertyKeyId() )
         count = 0;
 
         try ( KernelStatement statement = (KernelStatement) tx.acquireStatement();
-              Cursor<NodeItem> nodeCursor = statement.getStoreStatement().acquireSingleNodeCursor( node ) )
+                Cursor<NodeItem> nodeCursor = statement.getStoreStatement().acquireSingleNodeCursor( node ) )
```
chrisvest: This one got indented as well.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -1260,16 +1275,15 @@ private void assertHasRelationships( long node )
     {
 
         try ( KernelStatement statement = (KernelStatement) tx.acquireStatement();
-              Cursor<NodeItem> nodeCursor = statement.getStoreStatement().acquireSingleNodeCursor( node ) )
+                Cursor<NodeItem> nodeCursor = statement.getStoreStatement().acquireSingleNodeCursor( node ) )
```
chrisvest: Another extra two-space indent.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -1379,7 +1386,7 @@ private void testGetRels( long[] relIds )
     private void deleteRelationships( long nodeId ) throws Exception
     {
         try ( KernelStatement statement = (KernelStatement) tx.acquireStatement();
-              Cursor<NodeItem> nodeCursor = statement.getStoreStatement().acquireSingleNodeCursor( nodeId ) )
+                Cursor<NodeItem> nodeCursor = statement.getStoreStatement().acquireSingleNodeCursor( nodeId ) )
```
chrisvest: And another one.

chrisvest (changes requested):
The fix itself looks good, but the test can be improved and I think some of the stylistic changes can be improved as well.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
davidegrohmann: we can add that

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
davidegrohmann: that's only formatting

davidegrohmann:
@chrisvest @tinwelint I'll revert the formatting mess

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
davidegrohmann: done

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
davidegrohmann: done

davidegrohmann:
@chrisvest @tinwelint reverted the formatting changes, many comments were on code I didn't touch at all.

davidegrohmann:
@chrisvest @tinwelint any news on this PR?

chrisvest (approved):
Looks good.

PR #8393, opened 2016-11-17
Update allocation tracing mechanics

MishaDemianenko:
Update allocation pojo to contain all allocation related information.
Switch from concurrent skip list map to custom copy on write allocations container because of performance reasons.
By using new allocation container we improve total time of windows build up to 2 times.


chrisvest:
Nice!

MishaDemianenko:
Marking as not ready to merge, catch agent that still slow

MishaDemianenko:
Root cause is not fixed by proposed changes, closing

PR #8396, opened 2016-11-17
Allow for stopping counts store in failed rotation state

davidegrohmann:


davidegrohmann:
@tinwelint any news on this?

tinwelint:
@davidegrohmann question: what's the state of the counts store files on disk at this moment of failure? Does it get worse by allowing to close the counts store, instead of throwing that `IllegalStateException( "Cannot stop in state: rotating" )`? I could imagine this could allow the db as a whole to shut down cleanly and next startup wouldn't recover the counts store properly. Could that happen?

EDIT: we've discussed and tested this and concluded a safe change

tinwelint:
For clarification, this will not affect recovery as the shutdown of the db will still fail (and trigger recovery), it's just that the actual error causing the timeout will be thrown/visible instead of a less useful `IllegalStateException( "Cannot stop in state: rotating" )`, which doesn't really say anything useful.

PR #8397, opened 2016-11-17
Fixing store copy to use netty chunking

mneedham:
We're now wiring our chunking into ChunkedInput so
it does chunking at multiple levels. We were previously
having a problem where we probably flooded Netty on the
server side and then nothing would be sent for 5 minutes.

With this change we let Netty handle the flow control.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/TimeoutLoop.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/TimeoutLoop.java
@@ -25,10 +25,12 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Supplier;
 
+import org.neo4j.logging.Log;
+
 class TimeoutLoop
 {
     static <T> T waitForCompletion( Future<T> future, String operation, Supplier<Long> millisSinceLastResponseSupplier,
-            long inactivityTimeoutMillis ) throws CatchUpClientException
+                                    long inactivityTimeoutMillis, Log log ) throws CatchUpClientException
     {
         long remainingTimeoutMillis = inactivityTimeoutMillis;
         while ( true )
@@ -49,6 +51,7 @@ class TimeoutLoop
             catch ( TimeoutException e )
             {
                 long millisSinceLastResponse = millisSinceLastResponseSupplier.get();
+                log.info( "Request timed out. Time since last response: " + millisSinceLastResponse );
                 if ( millisSinceLastResponse < inactivityTimeoutMillis )
                 {
                     remainingTimeoutMillis = inactivityTimeoutMillis - millisSinceLastResponse;
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/FileSender.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/FileSender.java
@@ -19,64 +19,89 @@
  */
 package org.neo4j.causalclustering.catchup.storecopy;
 
-import io.netty.channel.ChannelHandlerContext;
-
-import java.io.File;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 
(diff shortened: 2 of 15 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/EntryRecord.java
```diff
@@ -64,7 +64,7 @@ public static EntryRecord read( ReadableChannel channel, ChannelMarshal<Replicat
         }
         catch ( ReadPastEndException e )
         {
-            throw EndOfStreamException.INSTANCE;
+            throw new EndOfStreamException( e );
```
martinfurmanski, line 67: Is this really necessary?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/FileSender.java
```diff
@@ -19,64 +19,89 @@
...
-                // and create a new one for the remaining bytes (or an empty byte array in case the file is empty)
-                bytes = createByteArray( buffer );
-            }
-
-            // send out whatever is left to send as the last chunk
-            perhapsWriteAndFlush( ctx, bytes, true );
+            endOfInput = true;
+            return null;
         }
+        return createByteArray( byteBuffer );
```
martinfurmanski, line 104: Can we in future perhaps refer to the ByteBuf instead of byte[] in FileChunk? Not super important now.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/FileSender.java
```diff
@@ -19,64 +19,89 @@
...
-import org.neo4j.causalclustering.catchup.ResponseMessageType;
-import org.neo4j.io.fs.FileSystemAbstraction;
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.stream.ChunkedInput;
+
 import org.neo4j.io.fs.StoreChannel;
 
-class FileSender
+class FileSender implements ChunkedInput<FileChunk>
```
martinfurmanski, line 31: Perhaps rename to ChunkedFileSender?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/TimeoutLoop.java
```diff
@@ -25,10 +25,12 @@
 import java.util.concurrent.TimeoutException;
 import java.util.function.Supplier;
 
+import org.neo4j.logging.Log;
+
 class TimeoutLoop
 {
     static <T> T waitForCompletion( Future<T> future, String operation, Supplier<Long> millisSinceLastResponseSupplier,
-            long inactivityTimeoutMillis ) throws CatchUpClientException
+                                    long inactivityTimeoutMillis, Log log ) throws CatchUpClientException
```
martinfurmanski, line 33: Unnecessary indentation

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/storage/SafeChannelMarshal.java
```diff
@@ -43,7 +43,7 @@ public final STATE unmarshal( ReadableChannel channel ) throws IOException, EndO
         }
         catch ( ReadPastEndException e )
         {
-            throw EndOfStreamException.INSTANCE;
+            throw new EndOfStreamException( e );
```
martinfurmanski, line 46: Same comment as before.

martinfurmanski (approved):
Looks solid to me, few minor comments.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/EntryRecord.java
mneedham: We were losing the underlying exception so it wasn't obvious which line threw the ReadPastEndException

PR #8398, opened 2016-11-17
Start LocalDatabase if it isn't started

mneedham:


mneedham:
We'll replace this with something better in the morning

PR #8402, opened 2016-11-18
Setting dbms.tracer=null will no longer prevent the database from starting when metrics are enabled

chrisvest:
Some default tracer implementations were implementing additional interfaces
that were looked after via the dependency resolver. This made the metrics
kernel extension fail to start when `dbms.tracer=null` was configured - a
setting that effectively turns all tracers off.

The `PlatformModule` now detects this, and publishes null-variants of the
alternative monitor interfaces when the configured tracer does not implement
them.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/PlatformModule.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/PlatformModule.java
@@ -53,8 +53,6 @@ import org.neo4j.kernel.impl.spi.SimpleKernelContext;
 import org.neo4j.kernel.impl.transaction.TransactionCounters;
 import org.neo4j.kernel.impl.transaction.log.checkpoint.CheckPointerMonitor;
 import org.neo4j.kernel.impl.transaction.state.DataSourceManager;
-import org.neo4j.kernel.impl.transaction.tracing.CheckPointTracer;
-import org.neo4j.kernel.impl.transaction.tracing.TransactionTracer;
 import org.neo4j.kernel.impl.util.JobScheduler;
 import org.neo4j.kernel.impl.util.Neo4jJobScheduler;
 import org.neo4j.kernel.info.DiagnosticsManager;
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/factory/PlatformModule.java
```diff
@@ -153,8 +157,12 @@ public DependencyResolver get()
         tracers = dependencies.satisfyDependency(
                 new Tracers( desiredImplementationName, logging.getInternalLog( Tracers.class ) ) );
         dependencies.satisfyDependency( tracers.pageCacheTracer );
-        dependencies.satisfyDependency( tracers.transactionTracer );
-        dependencies.satisfyDependency( tracers.checkPointTracer );
+        TransactionTracer transactionTracer = tracers.transactionTracer;
+        dependencies.satisfyDependency(
```
MishaDemianenko: Lets please introduce a function `defaultInstance(DesiredInterface, instances...)` or smt closed to that, that will select first matching instance and will return it; that will make this kind of implementation selection much more clean

MishaDemianenko (commented):
Looks good, one small stylish comment

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/PlatformModule.java
chrisvest: It's not a very common thing, though.

MishaDemianenko: that's true, so local private helper method should be good enough

chrisvest:
@MishaDemianenko I extracted the logic into a private method instead. I think it's premature to move it into the DependencyResolver class.

MishaDemianenko:
@chrisvest looks good now; my though was also about private helper method it's too early to introduce that thing in dependency resolving mechanics.

PR #8404, opened 2016-11-18
Value validation for indexed properties.

MishaDemianenko:
Introduce validation of property values for legacy and schema indexes.
Create validators for empty/null/allowable cases. Perform validation during property updated and index population.

changelog: Add validation of schema and legacy index values: do not allow nulls, values that over exceed lucene max term size, etc.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/IndexSimpleValueValidator.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/IndexSimpleValueValidator.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api;
+
+import org.neo4j.kernel.impl.util.Validator;
+
+public class IndexSimpleValueValidator implements Validator
+{
+
+    public static IndexSimpleValueValidator INSTANCE = new IndexSimpleValueValidator();
+
+    private IndexSimpleValueValidator()
+    {
+    }
+
+    @Override
+    public void validate( Object value )
+    {
+        if ( value == null )
+        {
+            throw new IllegalArgumentException( "Null value" );
+        }
+        if ( value instanceof String )
+        {
+            IndexValueLengthValidator.INSTANCE.validate( ((String) value).getBytes() );
+        }
+    }
+}
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/api/IndexSimpleValueValidatorTest.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/api/IndexSimpleValueValidatorTest.java
@@ -53,6 +53,16 @@ public class IndexSimpleValueValidatorTest
     }
 
     @Test
+    public void stringOverExceedLimitNotAllowed()
+    {
+        expectedException.expect( IllegalArgumentException.class );
+        expectedException.expectMessage(
+                "Property value bytes length: 32767 is longer then 32766, " +
+                        "which is maximum supported length of indexed property value." );
+        getValidator().validate( RandomStringUtils.randomAlphabetic( IndexValueLengthValidator.MAX_TERM_LENGTH + 1 ) );
+    }
+
+    @Test
     public void numberIsValidValue()
     {
         getValidator().validate( 5 );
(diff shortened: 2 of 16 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/IndexSimpleValueValidator.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    public void validate( Object value )
+    {
+        if ( value == null )
+        {
+            throw new IllegalArgumentException( "Null value" );
+        }
+        if ( value instanceof String )
+        {
+            IndexValueLengthValidator.INSTANCE.validate( ((String) value).getBytes() );
```
lutovich, line 42: `getBytes()` here and in `IndexValueValidator` will cause decoding of chars and copying them in a newly allocated byte array. is this ok performance wise? maybe we can calculate bytes length manually?

MishaDemianenko: you are obviously right, those guys where just automagically moved from LuceneDocumentStructure to a specific validator, maybe it's time to clean that as well even if that is not the goal of this pr...

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneLegacyIndex.java
```diff
@@ -131,16 +132,9 @@ private static void assertValidKey( String key )
         }
     }
 
-    private static void assertValidValue( Object singleValue )
+    protected void assertValidValue( Object value )
```
tinwelint: Looks like this methods isn't actually needed anymore, or it could at least still remain private static, couldn't it?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/IndexSimpleValueValidator.java
tinwelint: So schema indexes already did this, right? Then it's "OK" to add this slight extra garbage to legacy indexes as well imho

tinwelint:
@MishaDemianenko looks like it needs rebase

tinwelint:
In general I think this looks OK

community/kernel/src/main/java/org/neo4j/kernel/impl/api/IndexSimpleValueValidator.java
lutovich: Never mind me. This validation has already been present for schema indexes just deeper in the call stack.

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneLegacyIndex.java
MishaDemianenko: yeah that was sneaked experimentation, cleaned up

community/kernel/src/main/java/org/neo4j/kernel/impl/api/IndexSimpleValueValidator.java
MishaDemianenko: ok, that was the case for schema indexes before and we definitely can take a look into that in future if that will be required, i think for now it should be fine

MishaDemianenko:
@tinwelint cleaned up and rebased

community/kernel/src/test/java/org/neo4j/kernel/impl/api/IndexSimpleValueValidatorTest.java
```diff
@@ -0,0 +1,77 @@
...
+        getValidator().validate( 5L );
+    }
+
+    @Test
+    public void shortStringIsValidValue()
+    {
+        getValidator().validate( RandomStringUtils.randomAlphabetic( 5 ) );
+        getValidator().validate( RandomStringUtils.randomAlphabetic( 10 ) );
+        getValidator().validate( RandomStringUtils.randomAlphabetic( 250 ) );
+        getValidator().validate( RandomStringUtils.randomAlphabetic( 450 ) );
```
burqen, line 80: Would be nice with a test for exactly `MAX_TERM_LENGTH` as well.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/IndexValueLengthValidatorTest.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Test
+    public void shortByteArrayIsValid()
+    {
+        INSTANCE.validate( RandomUtils.nextBytes( 3 ) );
+        INSTANCE.validate( RandomUtils.nextBytes( 30 ) );
+        INSTANCE.validate( RandomUtils.nextBytes( 300 ) );
+        INSTANCE.validate( RandomUtils.nextBytes( 4303 ) );
+        INSTANCE.validate( RandomUtils.nextBytes( 13234 ) );
```
burqen, line 59: Test array of `MAX_TERM_LENGTH` here as well.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/IndexValueValidatorTest.java
```diff
@@ -0,0 +1,52 @@
...
+        getValidator().validate( RandomUtils.nextBytes(39978 ) );
+    }
+
+    @Test
+    public void shortArrayIsAllowed() throws Exception
+    {
+        getValidator().validate( RandomUtils.nextBytes( 3 ) );
+        getValidator().validate( RandomUtils.nextBytes( 30 ) );
+        getValidator().validate( RandomUtils.nextBytes( 450 ) );
+        getValidator().validate( RandomUtils.nextBytes( 4556 ) );
```
burqen, line 44: Again, `MAX_TERM_LENGTH`

burqen (changes requested):
Small comments on test coverage. Looks good otherwise.

burqen:
@MishaDemianenko This would be a candidate for changelog 👍

MishaDemianenko:
@burqen additional tests added for all things except IndexValueValidatorTest since you can directly ask for those number of bytes for array since those will be encoded by array encoder. We can pretend and guess border line but i do not think we need to do that.
And change log label added and message provided

burqen:
@MishaDemianenko Looks good! Approved by me once green.

burqen (approved):

burqen:
@MishaDemianenko triggered another run for Linux build.
The muted test that failed, is that not related?

MishaDemianenko:
@burqen ok, rebased and old failure gone now it's failing with new 3.2 non related failures

PR #8405, opened 2016-11-18
Use label scan store by default if possible for schema index population

MishaDemianenko:
Switch toggle to allow usage of label scan store by default as a source during index population


tinwelint:
@MishaDemianenko test failures look related

MishaDemianenko:
@tinwelint failures fixed and rebased, now it's failing because of yet another non related 3.2 problem

PR #8406, opened 2016-11-21
Solve bug when combining pattern comprehension and map projections

systay:
MapProjection needs the current scope, so it's been recorded in the semantic state.
But since MapProjections can contain other AST elements that get rewritten, it's important to
put this information in a rewrite-safe format and not keep it in the SemanticState, where any
rewrite will make the information hard to get at.

This commit puts the scope information into the MapProjection AST-object.

### DIFF
--- a/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/rewriters/recordScopes.scala
+++ b/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/rewriters/recordScopes.scala
@@ -24,12 +24,12 @@ import org.neo4j.cypher.internal.frontend.v3_1.{Rewriter, SemanticState, topDown
 
 case class recordScopes(semanticState: SemanticState) extends Rewriter {
 
-  def apply(that: AnyRef): AnyRef = topDown(instance).apply(that)
+  def apply(that: AnyRef): AnyRef = instance.apply(that)
 
-  private val instance: Rewriter = Rewriter.lift {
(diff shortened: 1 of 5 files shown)


community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/rewriters/recordScopes.scala
```diff
@@ -19,15 +19,17 @@
...
 
-import org.neo4j.cypher.internal.frontend.v3_1.ast.PatternComprehension
-import org.neo4j.cypher.internal.frontend.v3_1.{Rewriter, SemanticState, bottomUp}
+import org.neo4j.cypher.internal.frontend.v3_1.ast.{MapProjection, PatternComprehension}
+import org.neo4j.cypher.internal.frontend.v3_1.{Rewriter, SemanticState, topDown}
 
 case class recordScopes(semanticState: SemanticState) extends Rewriter {
 
-  def apply(that: AnyRef): AnyRef = bottomUp(instance).apply(that)
+  def apply(that: AnyRef): AnyRef = topDown(instance).apply(that)
```
Mats-SX: Since `topDown()` creates an object, isn't it better to have that be done once, on the instance, instead of at each call to `apply()`? Ie:

```
def apply(that: AnyRef): AnyRef = instance.apply(that)

private val instance: Rewriter = topDown(Rewriter.lift { ...
```

Mats-SX (changes requested):
Looks good, except for a little thing.

systay:
@Mats-SX Your review comment has been addressed. Went ahead and merged since we want this in the next 3.1 release

PR #8408, opened 2016-11-21
Introducting GBPTree, initial step for native indexing

burqen:
**WHAT?**
This is the first step in building a generation-aware B+tree (GB+Tree) implementation directly atop a PageCache with no caching in between. Additionally internal and leaf nodes on same level are linked both left and right with sibling pointers, this to provide correct reading when concurrently modifying the tree.

**WHY?**
Initial prototyping has shown great performance improvement on label scan store writes compared to lucene.

A GB+Tree could potentially be used for a lot of different types of indexes, path index for example.

**DETAILS ON IMPLEMENTATION**
*Non blocking*
Tree is modified by a single writer, multiple concurrent readers are allowed. Both reading and writing is non blocking.

*Garbage*
Implementation aim to make reading and writing garbage free.

*Build order*
The 'index' module is built directly before 'kernel' module, see community package pom.

**GENERATION**
Every node and child/sibling pointer has a generation indicating when it first evolved (was created). During recovery, pointers and nodes can have been flushed to disk or lost from memory. To determine if a pointer or node can be 'trusted' to be stable on disk there is a global STABLE generation of the tree. This generation and every generation before it has been checkpointed (flushed to disk) and can therefore be trusted, even after a crash.

The generation currently under evolution is considered to be UNSTABLE. Data belonging to or pointing to UNSTABLE generation can not be trusted after a crash and must be overwritten during recovery. Note that this is not implemented yet. After checkpoint the STABLE generation will be set to the UNSTABLE generation and UNSTABLE generation will be incremented.

*GenSafePointer, GSP*
By guarding the pointer fields, generation and address, with a checksum we can determine if a pointer is BROKEN (wrong checksum). A pointer can be broken if the page is flushed while writing the pointer and then crash before that page is flushed again.

A pointer with generation and checksum is called a GenSafePointer, GSP.

*GenSafePointerPair, GSPP*
We always want to keep a stable version of every pointer. Therefore all pointers are in fact two GSPs, a GSPP. So when we read or write a child pointer for example, we look at the pair for pointers as a whole and determine which one to read from or write to, look at GenSafePointerPair class.

**LabelScanStore**
There is an implementation of a label scan store built on top of this that will be introduced in a separate PR.

**Note**
This is minimal feature set and among the things missing (but in pipeline) are:
- Implement recovery
- Implement freelist
- Rebalance and merge when removing from tree


community/index/src/main/java/org/neo4j/index/Index.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Returns a {@link Modifier} able to modify the index, i.e. insert and remove keys/values.
+     * After usage the returned modifier must be closed, typically by using try-with-resource clause.
+     *
+     * @param options {@link Modifier.Options} which will apply to all modifications by the returned {@link Modifier}.
+     * @return {@link Modifier} able to modify the index.
+     * @throws IOException on error accessing the index.
+     */
+    Modifier<KEY,VALUE> modifier( Modifier.Options options ) throws IOException;
```
chrisvest: I think `IndexWriter` or `IndexMutator` would be better names. `Modifier` sounds like something that offsets the result of a computation.

community/index/src/main/java/org/neo4j/index/Modifier.java
```diff
@@ -0,0 +1,115 @@
...
+     * Inserts a key/value pair. In the event where {@code key} already exists the {@link ValueAmender}
+     * gets consulted, which can choose e.g. to insert a new key/value pair, overwrite or somehow modify
+     * the existing value.
+     *
+     * @param key key to insert.
+     * @param value value to insert for the {@code key}.
+     * @param amender {@link ValueAmender} to consult if key already exists.
+     * @throws IOException on index access error.
+     */
+    void insert( KEY key, VALUE value, ValueAmender<VALUE> amender ) throws IOException;
```
chrisvest: This is equivalent to the `java.util.Map#merge` method. Perhaps we can borrow nomenclature from there? Also, take a look at the types they are using.

community/index/src/main/java/org/neo4j/index/Index.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Flushes any pending changes to storage.
+     *
+     * @throws IOException on error flushing to storage.
+     */
+    void flush() throws IOException;
```
chrisvest, line 64: Is this really going to flush things, or just update the generations?

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointer.java
```diff
@@ -0,0 +1,132 @@
...
+package org.neo4j.index.gbptree;
+
+import org.neo4j.io.pagecache.PageCursor;
+
+/**
+ * Provides static methods for getting and manipulating GSP (gen-safe pointer) data.
+ * All interaction is made using a {@link PageCursor}. These methods are about a single GSP,
+ * whereas the normal use case of a GSP is in pairs ({@link GenSafePointerPair GSPP}).
+ * <p>
+ * A GSP consists of [generation,pointer,checksum] where checksum is updated
```
chrisvest: Truncated sentence?

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeTest.java
```diff
@@ -0,0 +1,624 @@
...
+    public final RandomRule random = new RandomRule().withSeed( 1 );
+    private PageCache pageCache;
+    private File indexFile;
+    private final Layout<MutableLong,MutableLong> layout = new SimpleLongLayout();
+    private GBPTree<MutableLong,MutableLong> index;
+
+    public GBPTree<MutableLong,MutableLong> createIndex( int pageSize )
+            throws IOException
+    {
+        pageCache = new MuninnPageCache( swapperFactory(), 10_000, pageSize, NULL );
```
chrisvest, line 83: Why don't you use the page cache rule?

chrisvest (commented):
Had a quick glance through, and had a few comments on obvious stuff I happened upon.

community/index/src/main/java/org/neo4j/index/Index.java
burqen: It should also flush, but that part is not implemented yet.
A call to this flush will be the last step when checkpointing the index.

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeTest.java
tinwelint: Probably because it lives in kernel, but it could be moved of course.

community/index/src/main/java/org/neo4j/index/Index.java
chrisvest: Okay, in that case you should make it take, and pass an `IOLimiter` to the `PagedFile#flushAndForce` method.

burqen: Good point! We will add this when we implement checkpointing.

MishaDemianenko (changes requested):
First part of review couple of comments so far

and all my comments gone somewhere...so i will start all over again! feels_like_a_monday! doooh

community/index/src/main/java/org/neo4j/index/Index.java
```diff
@@ -0,0 +1,65 @@
...
+     * @throws IOException on error accessing the index.
+     */
+    IndexWriter<KEY,VALUE> writer( IndexWriter.Options options ) throws IOException;
+
+    /**
+     * Flushes any pending changes to storage.
+     *
+     * @throws IOException on error flushing to storage.
+     */
+    void flush() throws IOException;
```
MishaDemianenko, line 64: It really strange to have flush method ad part of index interface. Since index already have a writer that ` able to modify the index, i.e. insert and remove keys/values` why it's not a writer responsibility then? also from a usage patter we do that only while closing index so why do not use close instead then?

community/index/src/main/java/org/neo4j/index/IndexWriter.java
```diff
@@ -0,0 +1,111 @@
...
+         *
+         * @return a factor between 0..1 where 0 means far to the left, 1 means far to the right and
+         * as an example 0.5 will select the middle item (floor division).
+         */
+        float splitLeftChildSize();
+
+        class Defaults implements Options
+        {
+            @Override
+            public float splitLeftChildSize()
```
MishaDemianenko: it more sounds like a split factor and not a split size because it hard to think about a size of 0.5f and because we already referring to it as a factor in the java docs

community/index/src/main/java/org/neo4j/index/IndexWriter.java
```diff
@@ -0,0 +1,111 @@
...
+     * more efficiently in various scenarios.
+     */
+    interface Options
+    {
+        /**
+         * Decides relatively where a split happens, i.e. which position will be the split key.
+         * Keys (and its values/children) including the split key will go to the right tree node,
+         * everything before it goes into the left.
+         *
+         * @return a factor between 0..1 where 0 means far to the left, 1 means far to the right and
```
MishaDemianenko: It gives a good explanation where we will put value (left or right) depending from supplied value but it does not describe macro effect of that: what impact that will give to the whole index

community/index/src/main/java/org/neo4j/index/ValueMerger.java
```diff
@@ -0,0 +1,39 @@
...
+{
+    /**
+     * Merge an existing value with a new value, returning potentially a combination of the two, or {@code null}
+     * if no merge was done effectively meaning that nothing should be written.
+     *
+     * @param value existing value
+     * @param withValue new value
+     * @return {@code value}, now merged with {@code withValue}, or {@code null} if no merge was done.
+     */
+    VALUE merge( VALUE value, VALUE withValue );
```
MishaDemianenko: Can we rename parameters to oldValue and newValue instead?

community/index/src/main/java/org/neo4j/index/ValueMerger.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public interface ValueMerger<VALUE>
+{
+    /**
+     * Merge an existing value with a new value, returning potentially a combination of the two, or {@code null}
+     * if no merge was done effectively meaning that nothing should be written.
+     *
+     * @param value existing value
+     * @param withValue new value
+     * @return {@code value}, now merged with {@code withValue}, or {@code null} if no merge was done.
```
MishaDemianenko, line 36: why do we need to return null in case if nothing has been change? to check for null afterwards? can we simply return old value as result in case we want to preserve it?

community/index/src/main/java/org/neo4j/index/Index.java
burqen: Flush will later be part of the checkpointing routine which will be
- flush index
- lock index (blocking writers)
- flush index (to make sure we flush data that was written during the previous flush)
- update index state (increment generation)
- flush index
- unlock index

It will of course also flush the page.

Does this make more sense?

community/index/src/main/java/org/neo4j/index/ValueMergers.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.neo4j.index;
+
+/**
+ * Common {@link ValueMerger} implementations.
+ */
+public class ValueMergers
+{
+    @SuppressWarnings( "rawtypes" )
+    private static final ValueMerger OVERWRITE = new ValueMerger()
```
MishaDemianenko: Lambda N1

community/index/src/main/java/org/neo4j/index/ValueMergers.java
```diff
@@ -0,0 +1,66 @@
...
+    {
+        @Override
+        public Object merge( Object value, Object withValue )
+        {
+            return withValue;
+        }
+    };
+
+    @SuppressWarnings( "rawtypes" )
+    private static final ValueMerger KEEP_EXISTING = new ValueMerger()
```
MishaDemianenko: Lambda N2

community/index/src/main/java/org/neo4j/index/IndexWriter.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Able to {@link #merge(Object, Object, ValueMerger)} and {@link #remove(Object)} key/value pairs
+ * into an {@link Index}. After all modifications have taken place the writer must be {@link #close() closed},
+ * typically using try-with-resource clause.
+ *
+ * @param <KEY> type of keys
+ * @param <VALUE> type of values
+ */
+public interface IndexWriter<KEY,VALUE> extends Closeable
```
MishaDemianenko, line 33: Somewhere in indexWriter on in the index javadoc we need to mention what it the visibility of updated/inserted values? when they will be visible to the reader? are updates atomic?

community/index/src/main/java/org/neo4j/index/gbptree/ByteArrayPageCursor.java
```diff
@@ -0,0 +1,278 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+import org.neo4j.io.pagecache.CursorException;
+import org.neo4j.io.pagecache.PageCursor;
+
+/**
+ * This class is bridging something which would otherwise make {@link InternalTreeLogic} code slightly more
```
MishaDemianenko, line 36: I find it super hard to follow this java doc it gives a lot of historical insight and describe challenges that we faced somewhere else, but almost nothing about this class itself

community/index/src/main/java/org/neo4j/index/gbptree/Layout.java
```diff
@@ -0,0 +1,167 @@
...
+     */
+    KEY newKey();
+
+    /**
+     * Populates {@code into} with minimum key that can ever exist in a tree with this layout.
+     *
+     * @param into instance to write minimum key into.
+     * @return the provided {@code into} instance for convenience.
+     */
+    KEY minKey( KEY into );
```
MishaDemianenko: For minKey and maxKey why do we need to supply into param? can we simply return new value with min and max key?

community/index/src/main/java/org/neo4j/index/gbptree/Layout.java
```diff
@@ -0,0 +1,167 @@
...
+     */
+    KEY maxKey( KEY into );
+
+    /**
+     * Copies contents of {@code key} to {@code into}.
+     *
+     * @param key key (left unchanged as part of this call) to copy contents from.
+     * @param into key (changed as part of this call) to copy contents into.
+     */
+    void copyKey( KEY key, KEY into );
```
MishaDemianenko: Why it's index layout responsibility to copy keys?

community/index/src/main/java/org/neo4j/index/ValueMerger.java
burqen: We took inspiration from `Map#merge` where a `null` from provided `BiFunction` means remove entry from map. We can not do that because we have a different path for remove. So we want to explicitly define the expected behaviour for when `ValueMerger` returns null.

We made it like this to avoid confusion, so it actually creates more confusion, it should be changed. Does it?

Note: Returning the old value will result in a write with no effective change of the actual value. We could add a check to avoid writing in that case.

community/index/src/main/java/org/neo4j/index/gbptree/Layout.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+        long upperInt = 0;
+        for ( int i = 0; i < chars.length; i++ )
+        {
+            byte byteValue = (byte) (((byte) chars[i]) ^ ((byte) (chars[i] >> 8)));
+            upperInt <<= 8;
+            upperInt |= (byteValue & 0xFF);
+        }
+
+        return upperInt << Integer.SIZE | (checksum & 0xFFFFFFFF);
```
MishaDemianenko, line 150: Why do we need 0xFFFFFFFF here?

community/index/src/main/java/org/neo4j/index/gbptree/Layout.java
burqen: To avoid object creation. This is the general pattern used in layout. See `readKey` and `readValue`.

community/index/src/main/java/org/neo4j/index/gbptree/Layout.java
burqen: Because he is the only one who knows about how a key looks.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+                    {
+                        pageSize = metaCursor.getInt();
+                        rootId = metaCursor.getLong();
+                        lastId = metaCursor.getLong();
+                        layoutIdentifier = metaCursor.getLong();
+                        majorVersion = metaCursor.getInt();
+                        minorVersion = metaCursor.getInt();
+                        layout.readMetaData( metaCursor );
+                    }
+                    while ( metaCursor.shouldRetry() );
```
MishaDemianenko, line 205: `metaCursor.checkAndClearCursorException();`  should be added here ?

community/index/src/test/java/org/neo4j/index/gbptree/SimpleLongLayout.java
```diff
@@ -0,0 +1,171 @@
...
+
+import org.apache.commons.lang3.mutable.MutableLong;
+
+import java.nio.charset.Charset;
+
+import org.neo4j.io.pagecache.PageCursor;
+
+class SimpleLongLayout implements Layout<MutableLong,MutableLong>
+{
+    private static final Charset UTF_8 = Charset.forName( "UTF-8" );
```
MishaDemianenko: this constant is useless `StandardCharsets.UTF_8` should be used instead

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+     * If the index doesn't exist it will be created and the {@link Layout} and {@code pageSize} will
+     * be written in index header.
+     * If the index exists it will be opened and the {@link Layout} will be matched with the information
+     * in the header. At the very least {@link Layout#identifier()} will be matched, but also if the
+     * index has {@link Layout#writeMetaData(PageCursor)} additional meta data it will be
+     * {@link Layout#readMetaData(PageCursor)}.
+     *
+     * @param pageCache {@link PageCache} to use to map index file
+     * @param indexFile {@link File} containing the actual index
+     * @param tentativePageSize page size, i.e. tree node size. Must be less than or equal to that of the page cache.
```
MishaDemianenko: now i can see that, so please ignore this comment

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+                                + pageCache.pageSize() + " so cannot be opened" );
+                    }
+                    pagedFile.close();
+                    pagedFile = pageCache.map( indexFile, pageSize );
+                }
+                return pagedFile;
+            }
+            catch ( Throwable t )
+            {
+                try
```
MishaDemianenko, line 236: IOUtils.closeAll

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+                    pagedFile.close();
+                }
+                catch ( IOException e )
+                {
+                    t.addSuppressed( e );
+                }
+                throw t;
+            }
+        }
+        catch ( NoSuchFileException e )
```
MishaDemianenko, line 247: instead of catching exception shouldn't we check file existence instead to perform index creation?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+            // Write header
+            try ( PageCursor metaCursor = openMetaPageCursor( pagedFile ) )
+            {
+                metaCursor.putInt( pageSize );
+                metaCursor.putLong( rootId );
+                metaCursor.putLong( lastId );
+                metaCursor.putLong( layout.identifier() );
+                metaCursor.putInt( layout.majorVersion() );
+                metaCursor.putInt( layout.minorVersion() );
+                layout.writeMetaData( metaCursor );
```
MishaDemianenko, line 269: metaCursor retry, reset offset, check for errors stuff?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+                {
+                    t.addSuppressed( e );
+                }
+                throw t;
+            }
+        }
+        catch ( NoSuchFileException e )
+        {
+            pageSize = pageSizeForCreation == 0 ? pageCache.pageSize() : pageSizeForCreation;
+            if ( pageSize > pageCache.pageSize() )
```
MishaDemianenko, line 250: exception handling should be extracted

community/index/src/main/java/org/neo4j/index/ValueMerger.java
tinwelint: Sure, that makes sense. Or even, to be even more specific to the use case, `existingValue` and `newValue`?

community/index/src/main/java/org/neo4j/index/IndexWriter.java
tinwelint: That'll most likely be implementation-specific so I'd say not suitable as interface javadoc. We can however elaborate on that in the `GBPTree` javadocs.

community/index/src/main/java/org/neo4j/index/gbptree/ByteArrayPageCursor.java
tinwelint: OK, thought it was so straight forward that it didn't need that kind of docs. We can add some about what it is and more notably what it can't do (w/ respect to PageCursor inheritance).

community/index/src/main/java/org/neo4j/index/gbptree/Layout.java
tinwelint: To avoid negative values writing a lot of 1's into the checksum. Collision rate is orders of magnitude bigger on random values when not having this mask.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: Not until we, if at all, start setting cursor exceptions in our write execution path. To my knowledge the cursor exception is completely user controlled and serves as a way for application code to transfer an exception through a "shouldRetry" barrier.

community/index/src/test/java/org/neo4j/index/gbptree/SimpleLongLayout.java
tinwelint: Not useless, merely sub-optimal then :) we can certainly change that.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: Don't think that applies. We're closing only one thing and the exception we're potentially adding a suppressed exception to has already been thrown (and caught) and cannot be injected into closeAll, can it?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: Letting the page cache throwing this is actually better because of potentially different hardware configuration the page cache may have. A file system instance has a more limited view of the world, compared to the page cache.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: Afaik not required when having a write lock on the cursor

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: Are you thinking of the two (almost identical) exceptions around page size? Those two messages are phrased slightly different, one relevant for creation and the other for open. I think that's fine to have explicit and "duplicated".

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+
+        boolean isInternal;
+        int keyCount = 0; // initialized to satisfy compiler
+        long childId = 0; // initialized to satisfy compiler
+        int pos;
+        do
+        {
+            do
+            {
+                isInternal = bTreeNode.isInternal( cursor );
```
MishaDemianenko, line 304: in case of shouldRetry cursor offset will not be reseted

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+                // Find the left-most key within from-range
+                keyCount = bTreeNode.keyCount( cursor );
+                pos = 0;
```
MishaDemianenko: it looks like we do not need this assignemnt

community/index/src/main/java/org/neo4j/index/gbptree/KeySearch.java
```diff
@@ -0,0 +1,125 @@
...
+            pos = lower;
+
+            hit = comparator.compare( key, bTreeNode.keyAt( cursor, readKey, pos ) ) == 0;
+        }
+        return searchResult( pos, hit );
+    }
+
+    private static int searchResult( int pos, boolean hit )
+    {
+        return (pos & 0x7FFFFFFF) | ((hit ? 1 : 0) << 31);
```
MishaDemianenko: we should have a constants for all maks

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+        {
+            throw new IllegalStateException( "Couldn't go to meta data page " + META_PAGE_ID );
+        }
+        return metaCursor;
+    }
+
+    @Override
+    public RawCursor<Hit<KEY,VALUE>,IOException> seek( KEY fromInclusive, KEY toExclusive ) throws IOException
+    {
+        long localRootId = rootId;
```
MishaDemianenko: i do not think we need localRootId variable in a for it's used now, it's only brings additional confusion why it's need to be reassigned.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+                continue;
+            }
+            if ( isInternal )
+            {
+                // TODO perhaps detect being too far to the left and so follow right siblings,
+                // to reduce unnecessary seek in leafs (concurrent splits would be the cause)
+
+                PointerChecking.checkChildPointer( childId );
+
+                if ( !cursor.next( childId ) )
```
MishaDemianenko, line 327: what if cursor.next faces io problems that should be retried?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
burqen: cursor will do it himself in case of should retry

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+        {
+            cursor.next();
+            TreePrinter.printTree( cursor, bTreeNode, layout, stableGeneration, unstableGeneration, System.out );
+        }
+    }
+
+    // Utility method
+    public boolean consistencyCheck() throws IOException
+    {
+        try ( PageCursor cursor = pagedFile.io( rootId, PagedFile.PF_SHARED_READ_LOCK ) )
```
MishaDemianenko, line 360: should retry?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+            cursor.next();
+            return new ConsistencyChecker<>( bTreeNode, layout, stableGeneration, unstableGeneration )
+                    .check( cursor );
+        }
+    }
+
+    @Override
+    public void flush() throws IOException
+    {
+        try ( PageCursor cursor = openMetaPageCursor( pagedFile ) )
```
MishaDemianenko, line 371: should retry?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+    /**
+     * @return the single {@link IndexWriter} for this index. The returned writer must be
+     * {@link IndexWriter#close() closed} before another caller can acquire this writer.
+     * @throws IllegalStateException for calls made between a successful call to this method and closing the
+     * returned writer.
+     */
+    @Override
+    public IndexWriter<KEY,VALUE> writer( IndexWriter.Options options ) throws IOException
+    {
+        SingleIndexWriter result = this.writer.getAndSet( null );
```
MishaDemianenko, line 396: mostly just an idea and own preference: what if instead of atomic reference for writer we will have ring buffer of work that writer will pick up instead?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -0,0 +1,481 @@
...
+        if ( result == null )
+        {
+            throw new IllegalStateException( "Only supports one concurrent writer" );
+        }
+        return result.take( rootId, options );
+    }
+
+    class SingleIndexWriter implements IndexWriter<KEY,VALUE>
```
MishaDemianenko: writer does not have at all retry attempts or error handling for cursors problems

community/index/src/main/java/org/neo4j/index/gbptree/RightmostInChain.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    long assertNext( PageCursor cursor )
+    {
+        long pageId = cursor.getCurrentPageId();
+        long leftSibling = node.leftSibling( cursor, stableGeneration, unstableGeneration );
+        long rightSibling = node.rightSibling( cursor, stableGeneration, unstableGeneration );
+
+        // Assert we have reached expected node and that we agree about being siblings
+        assert leftSibling == currentRightmost :
```
MishaDemianenko, line 58: should all those assertion be assertion exception instead of assertion errors?

MishaDemianenko: thats applicable to whole consistency checker

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointer.java
```diff
@@ -0,0 +1,132 @@
...
+    {
+        assert generation >= MIN_GENERATION && generation <= MAX_GENERATION : generation;
+        cursor.putInt( (int) generation );
+        put6BLong( cursor, pointer );
+        cursor.putShort( checksumOf( generation, pointer ) );
+    }
+
+    public static long readGeneration( PageCursor cursor )
+    {
+        return cursor.getInt() & 0xFFFFFFFFL;
```
MishaDemianenko: max generation mask?

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointerPair.java
```diff
@@ -0,0 +1,485 @@
...
+    static final long GEN_EQUAL    = 0x00000000_00000000L;
+    static final long GEN_A_BIG    = 0x10000000_00000000L;
+    static final long GEN_B_BIG    = 0x20000000_00000000L;
+    static final long WRITE_TO_A   = 0x00000000_00000000L;
+    static final long WRITE_TO_B   = 0x00200000_00000000L;
+    static final int STATE_SHIFT_A = 57;
+    static final int STATE_SHIFT_B = 54;
+
+    // Aggregations and masks
+    static final long SUCCESS_WRITE_TO_B = SUCCESS | WRITE | WRITE_TO_B;
```
MishaDemianenko, line 93: you can completely avoid | with SUCCESS since that is 0

MishaDemianenko: same for GEN_EQUAL

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
burqen: Yes, definitely. Suggest postponing this work for a different PR, as we need to refactor entire ConsistencyChecker to not be recursive. Card created.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
burqen: Cursor has write lock, should retry not needed.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
burqen: This could be a nice thing for the future if interaction with neo changes and we can't rely on WorkSync. For now, this implementation is bare minimum of what is needed.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
burqen: Does not need to call `shouldRetry` because all page access is guarded by write locks. `InternalTreeLogic` does not set any cursor exceptions that we need to check. 
`checkAndClearBoundsFlag` should be moved to end of `SingleIndexWriter#merge`.

burqen: Will fix `shouldRetry` for remove. Good catch!

MishaDemianenko:
@burqen @tinwelint i think i finish with a first round for now also general comment in cases when algo is taking more then a page it would be nice to split and extract separate parts from it to make it more easier to follow, not it's can be quite time consuming to get what's going on there

community/index/src/main/java/org/neo4j/index/IndexWriter.java
MishaDemianenko: sure :)

community/index/src/main/java/org/neo4j/index/gbptree/Layout.java
MishaDemianenko: well except before invoking those now we always create new key for them :)

community/index/src/main/java/org/neo4j/index/gbptree/Layout.java
MishaDemianenko: waaait, should key itself knew that?

community/index/src/main/java/org/neo4j/index/gbptree/RightmostInChain.java
burqen: Yes, this should be fixed. Right now only used in tests and we are not sure how we want it to look, therefore we suggest postponing changing this.

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointerPair.java
burqen: Yes, but we wanted to keep it here for code clarity

burqen:
@chrisvest @MishaDemianenko Comments addressed and also cleaned up some additional stuff that we found. Go get 'em tiger 🐯

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointerPair.java
tinwelint: Yeah, I'm pretty sure some level of the compiler or JIT will remove those anyway. It looks great when reading the code.

tinwelint:
For the record: we've set a deadline for addressing new comments (we believe that up to this point all current comments have been addressed) to Monday 27:th. This because further development is moving quickly ahead and the gap between this PR and the further development should be limited as much as possible.

PR #8409, opened 2016-11-21
Only print on actual timeout

martinfurmanski:


mneedham (approved):
LGTM

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/TimeoutLoop.java
```diff
@@ -51,13 +51,13 @@
...
             {
                 long millisSinceLastResponse = millisSinceLastResponseSupplier.get();
-                log.info( "Request timed out. Time since last response: " + millisSinceLastResponse );
                 if ( millisSinceLastResponse < inactivityTimeoutMillis )
                 {
                     remainingTimeoutMillis = inactivityTimeoutMillis - millisSinceLastResponse;
                 }
                 else
                 {
+                    log.info( "Request timed out. Time since last response: " + millisSinceLastResponse );
```
maxsumrall: String.format might have better performance.

PR #8416, opened 2016-11-22
Fix test that started failing

davidegrohmann:


community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -756,9 +774,13 @@ public void shouldCloseAllTheStoreEvenIfExceptionsAreThrown() throws Exception
         {
             // then
             assertEquals( "Cannot stop in state: rotating", ex.getMessage() );
-            // and the page cache closes with no errors
-            pageCache.close();
         }
+
+        thread.interrupt();
+        thread.join();
```
chrisvest, line 778: Could you use an single-threaded `ExecutorService` instead of a `Thread` directly? That way you can get the `RuntimeException` out of the `Future.get()` method. You can use `Future.cancel(true)` to interrupt the thread. (Or perhaps even make it interrupt itself? I don't know)

chrisvest: Well... except a cancelled future will throw `CancellationException` from its `get` methods. Okay, never mind.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
```diff
@@ -735,16 +738,31 @@ public void shouldCloseAllTheStoreEvenIfExceptionsAreThrown() throws Exception
...
+                }
+                catch ( IOException e )
+                {
+                    throw new RuntimeException( e );
+                }
+            }
+        };
+        thread.start();
+        latch.await();
+        Thread.sleep( 500 ); // some extra time to allow the other thread to run
```
chrisvest: You can use `ThreadTestUtils.awaitThreadState` instead.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
chrisvest: It'd just be nice if you could assert that only the InterruptedException was thrown, and not any other exception.

chrisvest (commented):
It looks okay. A couple of things could be improved, if you don't mind.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
davidegrohmann: oh, good point, I'll try

community/kernel/src/test/java/org/neo4j/kernel/impl/store/NeoStoresTest.java
davidegrohmann: I'll try to add that

davidegrohmann: Done

PR #8418, opened 2016-11-22
Fix recovery issue in the code finding the latest check point

davidegrohmann:
In the case there are only log files with versions greater than zero
and no checkpoints in any of those, we mistakenly report that no
recovery is needed.

changelog: Fix a bug that could prevent recovery from finding the latest check point record in the logs, preventing adequate recovery of the store.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
@@ -62,7 +62,7 @@ public class LatestCheckPointFinder
             LogVersionedStoreChannel channel = PhysicalLogFile.tryOpenForVersion( logFiles, fileSystem, version );
             if ( channel == null )
             {
-                return new LatestCheckPoint( null, false, oldestVersionFound );
+                break;
             }
 
             oldestVersionFound = version;
--- a/community/kernel/src/test/java/org/neo4j/kernel/recovery/LatestCheckPointFinderTest.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/recovery/LatestCheckPointFinderTest.java
@@ -19,14 +19,17 @@
  */
 package org.neo4j.kernel.recovery;
 
-import org.junit.Before;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
```diff
@@ -62,7 +62,7 @@ public LatestCheckPoint find( long fromVersionBackwards ) throws IOException
             LogVersionedStoreChannel channel = PhysicalLogFile.tryOpenForVersion( logFiles, fileSystem, version );
```
martinfurmanski, line 62: For the future: Would be better if this actually threw the IOException and we caught only FileNotFoundException specifically.

martinfurmanski (approved):
Looks like an improvement to me.

community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
davidegrohmann: it makes sense, shall we investigate that in a separate PR?

martinfurmanski: Perhaps not at this point in time?

davidegrohmann: whenever we got some time

tinwelint:
Whereas I think this is a good and correct change, one has to wonder how the db can get into a state where there are only tx log version > 0 AND none of them contains a checkpoint. This means that there have been rotations and pruning taking place and that pruning have actually pruned the last checkpont.

Or, is this an issue with store copy/backup only? I could see that happening there perhaps.

In any case this make recovery safer than it was before, whether or not there's an additional log pruning issue somewhere.

davidegrohmann:
@tinwelint This has been seen after store copy when pulling txs. Mistakenly no checkpoint was written into the log, but the recovery code failed to figure out that recovery was needed, which is unsafe.

PR #8423, opened 2016-11-23
On database shutdown make sure to fail txs waiting on a bookmark

davidegrohmann:


community/kernel/src/main/java/org/neo4j/kernel/api/txtracking/TransactionIdTracker.java
```diff
@@ -39,9 +40,11 @@
     private static final TimeUnit POLL_UNIT = TimeUnit.MILLISECONDS;
 
     private final TransactionIdStore transactionIdStore;
+    private final LongPredicate isAvailable;
 
-    public TransactionIdTracker( TransactionIdStore transactionIdStore )
+    public TransactionIdTracker( TransactionIdStore transactionIdStore, LongPredicate isAvailable )
```
lutovich: I think `LongPredicate` is a bit too generic here. Can we maybe instead pass in `AvailabilityGuard` and then use it's `#isAvailable(long)` method? This would result in same functionality but with more descriptive and concrete types.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachineSPI.java
```diff
@@ -70,10 +70,9 @@
         this.db = db;
         this.txBridge = txBridge;
         this.queryExecutionEngine = queryExecutionEngine;
-        this.transactionIdTracker = new TransactionIdTracker( transactionIdStoreSupplier );
+        this.transactionIdTracker = new TransactionIdTracker( transactionIdStoreSupplier, db::isAvailable );
```
lutovich: We can get `AvailabilityGuard` from `db.dependencyResolver` here, I think.

lutovich (commented):
@davidegrohmann changes look fine. I think it is worth changing `LongPredicate` to `AvailabilityGuard` in `TransactionIdTracker`.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachineSPI.java
davidegrohmann: No we cannot, since it is not there.

community/kernel/src/main/java/org/neo4j/kernel/api/txtracking/TransactionIdTracker.java
davidegrohmann: There is no easy way to do so. Since the `AvailabilityGuard` cannot be fetched from the dependency resolver.

lutovich: We can add it to dependency resolver, right?

davidegrohmann: Yes, but I rather do not take on refactors now.

lutovich: I would not call it a refactor. It is a one line change in `PlatformModule`. Just wrap creation of `AvailabilityGuard` with `dependencies.satisfyDependency(...)`.

davidegrohmann:
@lutovich updated as requested

lutovich (approved):

lutovich:
@davidegrohmann build has some failures in `ReadReplicaReplicationIT`

davidegrohmann:
@lutovich They are unrelated and muted in CI. I believe a fix has already been merged.

PR #8427, opened 2016-11-23
do not allow message handling before being started

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/CoreState.java
```diff
@@ -145,54 +148,57 @@ public void prune() throws IOException
...
-    public void start() throws Throwable
+    public synchronized void start() throws Throwable
     {
         // How can state be installed?
         // 1. Already installed (detected by checking on-disk state)
         // 2. Bootstrap (single selected server)
         // 3. Download from someone else (others)
 
         clusterIdentity.bindToCluster( this::installSnapshot );
+        allowMessageHandling = true;
```
davidegrohmann, line 166: Can we really allow messages before having started the db?

martinfurmanski: Yea so we are holding the monitor at this point right so messages can't enter handle without us explicitly wishing for them to do so by the wait(). The RAFT is up and running at this point but the CommandApplicationProcess isn't so nothing will be attempted to be applied until we start it further below. It will just queue up in the RAFT logs. The purpose here is to wait() for the download of a snapshot when we !haveState(). If we do haveState() then we will still recover by starting CommandApplicationProcess before allowing anything to happen due to the incoming messages, and specifically for example a snapshot download can't happen while we are starting up and in the middle of recovering.

PR #8428, opened 2016-11-23
Allow reads on followers by default

davidegrohmann:
We have seen in experiments than cores are idle doing nothing, so it
makes sense to allow reads on followers by default and utilize them
for running queries.


maxsumrall (approved):

martinfurmanski (commented):
Seems like the sanest default to me.

PR #8430, opened 2016-11-24
Do not always prune logs when checkpointing

davidegrohmann:


chrisvest:
Is this an optimisation or is it about correctness?

davidegrohmann:
@chrisvest I am trying to figure out if log pruning before store copy could cause problems or not. This PR is to make sure that all tests are still green.

PR #8434, opened 2016-11-25
Restore removed batch inserters public API

MishaDemianenko:
Add batch inserters package to clirr for automatic detection of future API changes.
Restored removed and updated API methods, add tests that verify that we properly close or not close batch inserter file system.


chrisvest (approved):
Good stuff!

PR #8438, opened 2016-11-25
Enable running checkstyle without breaking pre-3.1 builds

Mats-SX:


Mats-SX:
Build errors are unrelated to change.

PR #8445, opened 2016-11-28
Print a helpful message if JAVA_HOME is incorrectly defined

benbc:


packaging/standalone/src/main/distribution/shell-scripts/bin/neo4j-shared.sh
```diff
@@ -88,6 +88,10 @@ _find_java_cmd() {
 
   if [[ "${JAVA_HOME:-}" ]] ; then
     JAVA_CMD="${JAVA_HOME}/bin/java"
+    if [[ ! -e "${JAVA_CMD}" ]]; then
```
spacecowboy: Maybe it should be `! -f` even? Finding a folder called java won't help us. It's unlikely, but if their path is wrong, who knows how wrong it can be.

benbc:
@spacecowboy Thanks. Done.

spacecowboy (approved):

PR #8446, opened 2016-11-28
Better shutdown code in Neo4jScheduler

davidegrohmann:


MishaDemianenko:
@davidegrohmann in which way this is better, where we can find tests that shows that?

davidegrohmann:
@MishaDemianenko we don't swallow exceptions to begin with

MishaDemianenko:
@davidegrohmann super nice! lets have a test that confirm that! and description of PR that give a descriptive meaning of nice

PR #8447, opened 2016-11-28
Clarify usage of dbms.security.auth_cache_ttl

OliviaYtterbrink:
Clarified that disabling the auth cache while using the LDAP auth provider and not using the LDAP system account is not supported.


fickludd:
This looks good to me!

PR #8450, opened 2016-11-28
Fix bug where network thread worker would stop if inbound channel queue was full

davidegrohmann:
When enqueueing a new message in the inboud channel if the queue is
full an exception is throw which effectively kills the network thread,
hence the test fails.  Instead of throwing we simply drop the incoming
message and allow the network thread to keep running.  This is what
would happen in practice in a real case scenario so the code should
handle it and the test should not fail.


martinfurmanski (approved):
👍

PR #8451, opened 2016-11-28
Update LuceneBatchInserterIndexProviderNewImpl expectation of specific batch inserter

MishaDemianenko:
Introduce new interface IndexConfigStoreProvider that responsible for providing config store
and use it as an expected type in LuceneBatchInserterIndexProviderNewImpl instead of BatchInserterImpl.
Modifications here is quite limited since most of the involved classes are part of public API


jjaderberg:
This impacts the 3.2 docs build, favourably. It is low (very) priority at the moment, but it is (so it seems) the next thing for us to get a green docs build for 3.2. The only reason for us to chase it now is to prevent _this_ failure to obfuscate others, which could potentially have us chasing lots of possible breaking product changes at once. Tl;dr: it would be sort of nice, from a docs perspective, to see this merged.

PR #8453, opened 2016-11-29
Index optimisations

chrisvest:
```
       High selectivity read    Low selectivity read     Medium selectivity read   Write
3.0:   0.581 ± 0.056  ops/ms    0.027 ± 0.003  ops/ms    0.143 ± 0.013  ops/ms      66.295 ± 2.103  ops/ms
3.1:   0.497 ± 0.073  ops/ms    0.022 ± 0.002  ops/ms    0.120 ± 0.011  ops/ms     105.256 ± 10.828 ops/ms
Now:   1.512 ± 0.137  ops/ms    0.030 ± 0.002  ops/ms    0.174 ± 0.019  ops/ms     104.168 ± 1.766  ops/ms
```

I think the removal of the `partition` lock is safe because of the use of a `CopyOnWriteArrayList`, plus I added extra synchronisation in the places that need atomicity at a higher level, but please check the correctness thoroughly.

**Note:** This should not be merged until _after_ 3.1 GA.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/AbstractLuceneIndex.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/AbstractLuceneIndex.java
@@ -31,6 +31,8 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.neo4j.graphdb.ResourceIterator;
 import org.neo4j.helpers.ArrayUtil;
@@ -56,10 +58,9 @@ import static java.util.stream.Collectors.toList;
  */
 public abstract class AbstractLuceneIndex
 {
-
     protected final PartitionedIndexStorage indexStorage;
     private final IndexPartitionFactory partitionFactory;
-    private List<AbstractIndexPartition> partitions = new CopyOnWriteArrayList<>();
+    private CopyOnWriteArrayList<AbstractIndexPartition> partitions = new CopyOnWriteArrayList<>();
     private volatile boolean open;
 
     public AbstractLuceneIndex( PartitionedIndexStorage indexStorage, IndexPartitionFactory partitionFactory )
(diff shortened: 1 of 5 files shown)


community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/AbstractLuceneIndex.java
```diff
@@ -56,10 +58,9 @@
  */
 public abstract class AbstractLuceneIndex
 {
-
     protected final PartitionedIndexStorage indexStorage;
     private final IndexPartitionFactory partitionFactory;
-    private List<AbstractIndexPartition> partitions = new CopyOnWriteArrayList<>();
+    private CopyOnWriteArrayList<AbstractIndexPartition> partitions = new CopyOnWriteArrayList<>();
```
MishaDemianenko, line 64: this should be List and it was before

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/AbstractLuceneIndex.java
```diff
@@ -92,14 +93,33 @@ public void create() throws IOException
...
-        for ( Map.Entry<File,Directory> indexDirectory : indexDirectories.entrySet() )
+        Stream<Map.Entry<File,Directory>> indexDirectories = indexStorage.openIndexDirectories().entrySet().stream();
+        List<AbstractIndexPartition> list = null;
+        try
         {
-            partitions.add( partitionFactory.createPartition( indexDirectory.getKey(), indexDirectory.getValue() ) );
+            list = indexDirectories.map( e -> createPartition( e.getKey(), e.getValue() ) )
+                                       .collect( Collectors.toList() );
         }
+        catch ( UncheckedIOException e )
```
MishaDemianenko: i do not think that map -> and wrapping/catching exceptions should be here since it makes it less readable, it's correct observation that it should use intermediate list instead.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/AbstractLuceneIndex.java
```diff
@@ -200,9 +220,10 @@ public void flush() throws IOException
 
     public void close() throws IOException
     {
-        IOUtils.closeAll( partitions );
-        partitions.clear();
         open = false;
+        List<AbstractIndexPartition> partitionsCopy = new ArrayList<>( partitions );
```
MishaDemianenko: i do not think it's needed since index is already closed at this point and we can operate on partitions itself right?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/WritableDatabaseLabelScanIndex.java
```diff
@@ -44,14 +44,14 @@ public WritableDatabaseLabelScanIndex( BitmapDocumentFormat format, PartitionedI
     @Override
     public LabelScanReader getLabelScanReader()
     {
-        partitionsLock.lock();
+//        partitionsLock.lock();
```
MishaDemianenko: left overs

MishaDemianenko (commented):
Couple of small comments

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/AbstractLuceneIndex.java
chrisvest: I changed it because we are now relying on specifics of the implementation, even if we are not using any methods not already defined in `List`. Giving the field a concrete type helps emphasise this, but I can also add a comment. What do you think?

MishaDemianenko: ok comment there about that is good enough

MishaDemianenko:
@chrisvest looks fine now will merge it after 3.1 release

PR #8454, opened 2016-11-29
GBTree stable generation copy-on-write

tinwelint:
Further implements tree generations by making sure changes aren't applied in stables nodes, but instead copying stable node into unstable generation and making change there. Nodes now each have a generation and stable node gets "new generation" forward-pointer to the new generation node. The new-generation pointer is rarely actually traversed because parent nodes redirect to new nodes directly, they are there for when reader goes to a stable node, which concurrently gets a new version of itself created.

This PR is a reasonably coherent chunk of this generation work and focuses on InternalTreeLogic, whereas there's further work to be done inside GBPTree to complete the generation work.


tinwelint:
Accidentally deleted this branch, recreated at https://github.com/neo4j/neo4j/pull/8474

PR #8455, opened 2016-11-29
Update Indexing service lifecycle to shutdown indexes on shutdown instead of stop

MishaDemianenko:
During database shutdown we stop life components, wait of all running transaction to finish and then shutdown everything.
To be able to finish all transactions that can be potentially still be there we need to keep indexes open till we reach shutdown phase.

Update NeoStoreDataSource to provide test access to internal life.

### DIFF
--- a/community/neo4j/src/test/java/org/neo4j/index/ShutdownOnIndexUpdateIT.java
+++ b/community/neo4j/src/test/java/org/neo4j/index/ShutdownOnIndexUpdateIT.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright (c) 2002-2016 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


community/neo4j/src/test/java/org/neo4j/index/ShutdownOnIndexUpdateIT.java
```diff
@@ -0,0 +1,145 @@
...
+    private static final String UNIQUE_PROPERTY_NAME = "uniquePropertyName";
+    private static final AtomicLong indexProvider = new AtomicLong();
+    private static Label constraintIndexLabel = Label.label( "ConstraintIndexLabel" );
+
+    @Rule
+    public TestDirectory testDirectory = TestDirectory.testDirectory();
+    @Rule
+    public CleanupRule cleanupRule = new CleanupRule();
+
+    private ExecutorService executors;
```
tinwelint: Not sure this is actually used.

community/neo4j/src/test/java/org/neo4j/index/ShutdownOnIndexUpdateIT.java
```diff
@@ -0,0 +1,145 @@
...
+    @After
+    public void tearDown()
+    {
+        executors.shutdown();
+    }
+
+    @Test
+    public void shutdownWhileFinishingTransactionWithIndexUpdates() throws Exception
+    {
+        GraphDatabaseService database = new GraphDatabaseFactory().newEmbeddedDatabase( testDirectory.graphDbDir() );
```
tinwelint: `new TestGraphDatabaseFactory().newImpermanentDatabase()`? and delete directory rule as well?

tinwelint:
Generally looks like a good change

MishaDemianenko:
@tinwelint test cleaned up.

PR #8457, opened 2016-11-29
Refactor configuration

spacecowboy:
The primary goal of this change is to make it possible to validate a
configuration file when it is parsed from disk. This has previously
been impossible because the individual settings have been spread
across modules and files.

Settings continue to live in separate files but the classes in which
they live are loaded via ServiceLoading. This allows the Config object
to have access to all settings at runtime.

To make sure that the Config object is used a sensible way its
constructors have been made private. The only way it build a Config is
via its static factory methods. This was already the primary way of
constructing it previously.

The new structure allows for two clear entry points:

* `embeddedDefaults()`
* `serverDefaults()`

The names should be self-explanatory.

A secondary effect was it was necessary to be able to pass a Config
object directly to the PlatformModule/GraphDatabase factories instead
of doing a Config to Map conversion.

Converting a Config object to a Map is still possible via the `getRaw()` method,
but should only be done when absolutely required. To
compensate, convenience functions have been added to allow safe and
reasonable access to the raw settings. Primary intended usage remains
passing a Settings object which does its own parsing.

A new maven module `community/configuration` was added to allow
modules to have loadable settings without adding a dependency on
Kernel.

changelog: Improved parsing of configuration

- added a significant amount of config validation to the startup process of neo4j. Any errors will prevent neo4j from starting but they will also have very helpful error messages attached to them.
- added a new config option `dbms.config.strict_validation` which enables even stricter validation (false by default, will be true by default in 4.0).
- added new methods to embedded graph database builder APIs which take a Config object instead of a StringMap.
- deprecated several database builder API methods in favor of the new Config methods.


community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -54,57 +72,158 @@
...
+        return embeddedDefaults( Optional.empty() );
+    }
+
+    /**
+     * @return a configuration with embedded defaults
+     */
+    public static Config embeddedDefaults( Map<String,String> additionalConfig )
+    {
+        return embeddedDefaults( Optional.empty(), additionalConfig );
+    }
```
chrisvest, line 106: Would it make sense to also have an `embeddedTestingDefaults` that tests could use? (Ours and our customers tests alike)

spacecowboy: It certainly would but I'm not sure what values that would entail except pagecache size? I'd be happy to take suggestions on the matter

community/configuration/src/main/java/org/neo4j/configuration/ExternalSettings.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * This class holds settings which are used external to the java code. This includes things present in the
+ * configuration which are only read and used by the wrapper scripts. By including them here, we suppress warning
+ * messages about Unknown configuration options, and make it possible to document these options via the normal methods.
+ *
+ * Be aware that values are still validated.
+ */
+@SuppressWarnings( "unused" )
+public class ExternalSettings implements LoadableConfig
```
benbc, line 35: Could this be provided by the `server` module or something else higher up the stack? `command-line`?

community/dbms/src/main/java/org/neo4j/server/configuration/ConfigLoader.java
```diff
@@ -21,53 +21,62 @@
...
 import org.neo4j.kernel.configuration.Settings;
 
 public class ConfigLoader
 {
     public static final String DEFAULT_CONFIG_FILE_NAME = "neo4j.conf";
 
-    private final Function<Map<String, String> ,Iterable<Class<?>>> settingsClassesSupplier;
+    public ConfigLoader()
+    {
+    }
```
benbc: Why do we need a public constructor as well as the static factory methods? If we can't hid the constructor for some reason then this empty no arg public constructor is not necessary.

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -528,7 +528,7 @@ public static void doImport( PrintStream out, PrintStream err, File storeDir, Fi
 
     private static Config loadDbConfig( File file ) throws IOException
     {
-        return file != null && file.exists() ? new Config( MapUtil.load( file ) ) : Config.defaults();
+        return file != null && file.exists() ? Config.embeddedDefaults( MapUtil.load( file ) ) : Config.defaults();
```
benbc, line 531: Should this use the disabled-connectors version just on principle?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseFactory.java
```diff
@@ -95,10 +96,11 @@ protected void configure( GraphDatabaseBuilder builder )
         // Let the default configuration pass through.
     }
 
-    protected GraphDatabaseService newDatabase( File storeDir, Map<String,String> config, GraphDatabaseFacadeFactory.Dependencies dependencies )
+    protected GraphDatabaseService newDatabase( File storeDir, Map<String,String> settings,
+            GraphDatabaseFacadeFactory.Dependencies dependencies )
```
benbc: Does it make sense to deprecate this method that takes settings as a map?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -464,6 +501,31 @@
...
+    public static final Setting<Boolean> archive_failed_index = setting(
+            "unsupported.dbms.index.archive_failed", BOOLEAN, "false" );
+
+    // Needed to validate config, accessed via reflection
+    @SuppressWarnings( "unused" )
+    public static final BoltConnectorValidator boltValidator = new BoltConnectorValidator();
+
+    // Needed to validate config, accessed via reflection
+    @SuppressWarnings( "unused" )
+    public static final HttpConnectorValidator httpValidator = new HttpConnectorValidator();
```
benbc: Gah! I'm deeply traumatised by defining the HTTP config in the kernel. No chance of pushing this up to the server, I suppose?

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -636,12 +637,15 @@ private TransactionLogModule buildTransactionLogs(
...
+                        transactionIdStore::getLastCommittedTransactionId, logInformation );
 
-        String pruningConf = config.get(
-                config.get( GraphDatabaseFacadeFactory.Configuration.ephemeral )
-                ? GraphDatabaseFacadeFactory.Configuration.ephemeral_keep_logical_logs
-                : GraphDatabaseSettings.keep_logical_logs );
+        if ( config.get( GraphDatabaseFacadeFactory.Configuration.ephemeral ) )
+        {
+            config = config.withDefaults(
+                    stringMap( GraphDatabaseSettings.keep_logical_logs.name(), "1 files" ) );
```
benbc, line 615: This seems an odd place to define this default. I don't understand why it's necessary.

community/kernel/src/main/java/org/neo4j/kernel/configuration/BoltConnectorValidator.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @Override
+    @Nonnull
+    public Map<String,String> validate( @Nonnull Map<String,String> rawConfig )
+            throws InvalidSettingException
+    {
+        final HashMap<String,String> result = new HashMap<>();
+
+        ownedEntries( rawConfig ).forEach( s ->
```
benbc: This seems to be basically the same as `values()`, can it not use the same implementation?

community/kernel/src/main/java/org/neo4j/kernel/configuration/Connector.java
```diff
@@ -0,0 +1,64 @@
...
+
+    // Note: Be careful about adding things here that does not apply to all connectors,
+    //       consider future options like non-tcp transports, making `address` a bad choice
+    //       as a setting that applies to every connector, for instance.
+
+    public final GroupSettingSupport group;
+
+    // Note: We no longer use the typeDefault parameter because it made for confusing behaviour;
+    // connectors with unspecified would override settings of other, unrelated connectors.
+    // However, we cannot remove the parameter at this
```
benbc: Incomplete sentence.

benbc: Should this constructor be deprecated in favour of one without the `typeDefault`?

community/kernel/src/main/java/org/neo4j/kernel/configuration/HttpConnectorValidator.java
```diff
@@ -0,0 +1,224 @@
...
+import static org.neo4j.kernel.configuration.HttpConnector.Encryption.NONE;
+import static org.neo4j.kernel.configuration.HttpConnector.Encryption.TLS;
+import static org.neo4j.kernel.configuration.Settings.BOOLEAN;
+import static org.neo4j.kernel.configuration.Settings.NO_DEFAULT;
+import static org.neo4j.kernel.configuration.Settings.advertisedAddress;
+import static org.neo4j.kernel.configuration.Settings.listenAddress;
+import static org.neo4j.kernel.configuration.Settings.options;
+import static org.neo4j.kernel.configuration.Settings.setting;
+
+public class HttpConnectorValidator extends ConnectorValidator
```
benbc, line 40: Is it possible to move the bits of `validate()/values()` that deal with common connector stuff up into `ConnectorValidator` so as to remove duplicate with `BoltConnectorValidator`?

community/kernel/src/main/java/org/neo4j/kernel/configuration/HttpConnectorValidator.java
```diff
@@ -0,0 +1,224 @@
...
+    public HttpConnectorValidator()
+    {
+        super( HTTP );
+    }
+
+    @Override
+    @Nonnull
+    public Map<String,String> validate( @Nonnull Map<String,String> rawConfig )
+            throws InvalidSettingException
+    {
```
benbc: Same as `BoltConnectorValidator`, there is duplication with `values()`.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacadeFactory.java
```diff
@@ -83,15 +85,10 @@
...
-    public static class Configuration
+    public static class Configuration implements LoadableConfig
     {
         public static final Setting<Boolean> ephemeral =
                 setting( "unsupported.dbms.ephemeral", Settings.BOOLEAN, Settings.FALSE );
-        public static final Setting<String> ephemeral_keep_logical_logs =
-                setting( "dbms.tx_log.rotation.retention_policy", STRING, "1 " + "files", illegalValueMessage( "must be `true`/`false` or of format '<number><optional unit> <type>' " +
-                        "for example `100M size` for " +
-                        "limiting logical log space on disk to 100Mb," +
-                        " or `200k txs` for limiting the number of transactions to keep to 200 000", matches( ANY ) ) );
```
benbc, line 94: Why is it okay to remove this?

community/kernel/src/main/java/org/neo4j/kernel/internal/EmbeddedGraphDatabase.java
```diff
@@ -72,6 +73,16 @@ public EmbeddedGraphDatabase( File storeDir,
         create( storeDir, params, dependencies );
     }
 
+    /**
+     * Internal constructor
+     */
+    public EmbeddedGraphDatabase( File storeDir,
+            Config config,
+            GraphDatabaseFacadeFactory.Dependencies dependencies )
```
benbc: Where is this called from?

community/kernel/src/main/java/org/neo4j/kernel/configuration/ServerConfigurationValidator.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.neo4j.graphdb.config.InvalidSettingException;
+import org.neo4j.graphdb.config.SettingValidator;
+import org.neo4j.logging.Log;
+
+import static org.neo4j.kernel.configuration.Connector.ConnectorType.BOLT;
+import static org.neo4j.kernel.configuration.Connector.ConnectorType.HTTP;
+
+
+public class ServerConfigurationValidator implements ConfigurationValidator
```
benbc, line 39: Does this really need to live down in the kernel?

community/server/src/main/java/org/neo4j/server/CommunityNeoServer.java
```diff
@@ -54,7 +54,7 @@
     protected static final GraphFactory COMMUNITY_FACTORY = ( config, dependencies ) -> {
         File storeDir = config.get( DatabaseManagementSystemSettings.database_path );
         return new GraphDatabaseFacadeFactory( DatabaseInfo.COMMUNITY, CommunityEditionModule::new )
-                .newFacade( storeDir, config.getParams(), dependencies );
+                .newFacade( storeDir, config, dependencies );
```
benbc, line 57: Hurray!

enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
```diff
@@ -217,11 +219,20 @@ private BackupOutcome incrementalBackup( FileSystemAbstraction fileSystem, Strin
...
         Map<String,String> temporaryDbConfig = getTemporaryDbConfig();
         config = config.with( temporaryDbConfig );
+
+        Map<String,String> configParams = new HashMap<>();
+        Set<String> keys = config.getConfiguredSettingKeys();
+        for ( String key : keys )
+        {
+            Optional<String> value = config.getRaw( key );
+            value.ifPresent( s -> configParams.put( key, s ) );
+        }
```
benbc, line 229: What's this doing?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CoreGraphDatabase.java
```diff
@@ -47,13 +48,28 @@ public CoreGraphDatabase( File storeDir, Map<String,String> params,
...
     }
 
+    public CoreGraphDatabase( File storeDir, Config config,
+            GraphDatabaseFacadeFactory.Dependencies dependencies )
+    {
+        this( storeDir, config, dependencies, new HazelcastDiscoveryServiceFactory() );
+    }
+
     public CoreGraphDatabase( File storeDir, Map<String,String> params,
             GraphDatabaseFacadeFactory.Dependencies dependencies, DiscoveryServiceFactory discoveryServiceFactory )
```
benbc: What is this still used for?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/ReadReplicaGraphDatabase.java
```diff
@@ -44,12 +45,26 @@ public ReadReplicaGraphDatabase( File storeDir, Map<String,String> params, Depen
         this( storeDir, params, dependencies, new HazelcastDiscoveryServiceFactory() );
     }
 
+    public ReadReplicaGraphDatabase( File storeDir, Config config, Dependencies dependencies )
+    {
+        this( storeDir, config, dependencies, new HazelcastDiscoveryServiceFactory() );
+    }
+
     public ReadReplicaGraphDatabase( File storeDir, Map<String,String> params, Dependencies dependencies,
             DiscoveryServiceFactory discoveryServiceFactory )
```
benbc: What is this still used for (as above)?

enterprise/cluster/src/main/java/org/neo4j/cluster/ClusterSettings.java
```diff
@@ -69,8 +71,22 @@ public String toString()
...
+        HA,
+        ARBITER,
+        CORE,
+        READ_REPLICA
+    }
+
+    @Description( "Configure the operating mode of the database -- 'SINGLE' for stand-alone operation, 'HA' for " +
+            "operating as a member in a cluster, 'ARBITER' for an HA-only cluster member with no database, " +
+            "CORE for a core member of a Causal Clustering cluster, or READ_REPLICA for read replica." )
+    public static final Setting<Mode> mode = setting( "dbms.mode", options( Mode.class ), Mode.SINGLE.name() );
```
benbc, line 86: This seems like a strange place to define this (since it crosses HA and core-edge. Where was it before?

benbc (commented):
Have you considered plugin-writers? I believe that some plugins have config which is written into `neo4j.conf`. How will this work for them? Maybe discus with Stefan Armbruster.

What is the experience like for Neo4j developers adding new settings classes (say a new module)? How will they discover that they have forgotten the service loader stuff?

benbc:
`Config.defaults()`: is this helpful? Why not force people to choose between `serverDefaults()` and `embeddedDefaults()`?

benbc:
Can `Config.serverDefaults()` use `withDefaults()` instead of doing its own `putIfAbsent` stuff?

benbc:
`Config.replaceSettings()`: can the extra validator (`IndividualSettingsValidator`) just be added to the list of validators locally to avoid the call to `.validate()` outside the loop?

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
spacecowboy: Would be nice but that would introduce a cyclic dependency since ConfigLoader is in dbms, and dbms already depends on importtool.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
spacecowboy: I would like to pull all connectors out of the kernel, but bolt being a kernel extension kind of prevents that. There are a lot of stuff in the kernel which shouldn't logically belong there but them being there now forces stuff to be there because of service loading

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
spacecowboy: I didn't write it, I just copied old functionality.

community/kernel/src/main/java/org/neo4j/kernel/configuration/BoltConnectorValidator.java
spacecowboy: No, `values` calls `values()` (logically enough) on the settings. Whereas `validate` calls `validate()` on the settings.

The result of `values` a map of setting-key, setting-value-object. The values have been converted from the raw string form into objects, like `ListendAddress` for example.

The result of 'validate' on the other hand is a map of the raw key, raw value. The result is the sub-map which this validator considers valid.

community/kernel/src/main/java/org/neo4j/kernel/configuration/HttpConnectorValidator.java
spacecowboy: I believe that would make things hard to read for the very marginal benefit of moving `enabled` out of the switch clause. `type` is a shared setting but each validator asserts that it is indeed correct in terms of itself (http vs bolt). It would also mean we'd need to remove the `default` case where unknown settings cause errors (signs of programmer error).

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacadeFactory.java
spacecowboy: It was a duplicate of the same setting present in `GraphDatabaseSettings.java`

community/kernel/src/main/java/org/neo4j/kernel/internal/EmbeddedGraphDatabase.java
spacecowboy: `ImpermanentGraphDatabase`, used in testing. Was able to make it protected.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ServerConfigurationValidator.java
spacecowboy: Yes, because the Config class has a `serverDefaults` method which requires this validator.

enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
spacecowboy: This piece of logic is converting the Config object into a StringMap. Required because GraphDatabaseBuilder does not yet accept a Config object (see comment above about required breaking interface changes)

enterprise/cluster/src/main/java/org/neo4j/cluster/ClusterSettings.java
spacecowboy: [EnterpriseServerSettings](https://github.com/neo4j/neo4j/blob/ddd4cbac3db01f0994fa3f5eefc656af59879861/enterprise/server-enterprise/src/main/java/org/neo4j/server/enterprise/EnterpriseServerSettings.java#L39)

which is a module higher than both HA and CC in the hierarchy (and it must reside in a module which precedes them both).

It's not the ideal location but I'm not sure such a place exists at the moment and I didn't want to create yet another maven module for this single setting.

spacecowboy:
@benbc in answer to your questions

> Have you considered plugin-writers? I believe that some plugins have config which is written into neo4j.conf. How will this work for them? Maybe discus with Stefan Armbruster.

If they want to put stuff in the neo4j config they'll have to have a settings class which is service loadable too. Our docs on plugins seem to be almost non-existing for the 3.x branches. I'm not sure what our stance on this is.

> What is the experience like for Neo4j developers adding new settings classes (say a new module)? How will they discover that they have forgotten the service loader stuff?

They'll get warnings in the log about "Unknown settings".

> Config.defaults(): is this helpful? Why not force people to choose between serverDefaults() and embeddedDefaults()?

I agree it should be done but it would grow the diff a lot, making reviewing even harder. It can be done once other changes are final.

> Can Config.serverDefaults() use withDefaults() instead of doing its own putIfAbsent stuff?

No because the validator would be executed and throw an exception about missing http connector before you get to call `withDefaults()`.

> Config.replaceSettings(): can the extra validator (IndividualSettingsValidator) just be added to the list of validators locally to avoid the call to .validate() outside the loop?

Not in a straight-forward way. Validators are passed on to child config objects when `with` methods are called. Adding the individualValidator to the list would introduce another duplicate every time a `with` method was called.

Pushed an update with changes to address other comments.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
benbc: Github's diff shows that `"1 files"` is new. Where did it come from?

community/kernel/src/main/java/org/neo4j/kernel/configuration/BoltConnectorValidator.java
benbc: Right -- but can't the structural duplication be removed?

community/kernel/src/main/java/org/neo4j/kernel/configuration/ServerConfigurationValidator.java
benbc: Could that method be generalized into something that would allow the actual server-specific stuff to move out of Kernel? I'm worried that this and the move of the HTTP stuff down to sit alongside the Bolt stuff is making things worse in a way that we don't have a plan to improve.

enterprise/cluster/src/main/java/org/neo4j/cluster/ClusterSettings.java
benbc: Sorry -- why can't it still be in `EnterpriseServerSettings`?

benbc:
> > Have you considered plugin-writers? I believe that some plugins have config which is written into neo4j.conf. How will this work for them? Maybe discus with Stefan Armbruster.

> If they want to put stuff in the neo4j config they'll have to have a settings class which is service loadable too. Our docs on plugins seem to be almost non-existing for the 3.x branches. I'm not sure what our stance on this is.

Please do talk to a human about what people actually do in practice, rather than relying on the lack of docs to consider the behaviour unspecified. This is the kind of thing that causes pain for Field and makes them angry with Engineering when we get it wrong.

benbc:
> > What is the experience like for Neo4j developers adding new settings classes (say a new module)? How will they discover that they have forgotten the service loader stuff?

> They'll get warnings in the log about "Unknown settings".

That's very optimistic of you. :-).

Do you think we should consider a breaking change for 4.0 where you can no longer use settings which haven't been defined?

community/kernel/src/main/java/org/neo4j/kernel/configuration/BoltConnectorValidator.java
spacecowboy: Not without adding a great deal of obfuscation in my opinion, which I consider much worse

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
spacecowboy: This piece of code is the logical (and readable) equivalent of the code which it replaced:

```java
        String pruningConf = config.get(
                 config.get( GraphDatabaseFacadeFactory.Configuration.ephemeral )
                 ? GraphDatabaseFacadeFactory.Configuration.ephemeral_keep_logical_logs
                 : GraphDatabaseSettings.keep_logical_logs );
```

That piece of code referenced a piece of duplicated code (which you also commented on) in `GraphDatabaseFacadeFactory` which was:

```java
ephemeral_keep_logical_logs = setting( "dbms.tx_log.rotation.retention_policy", STRING, "1 " + "files", ....)
```

So the code boils down to:

if database is ephemeral
then default for keep_logical_logs is "1 file"
else the regular default defined in GraphDatabaseSettings

enterprise/cluster/src/main/java/org/neo4j/cluster/ClusterSettings.java
spacecowboy: because `EnterpriseServer` depends on HA and CC, not the other way round

community/configuration/src/test/java/org/neo4j/configuration/LoadableConfigTest.java
```diff
@@ -0,0 +1,176 @@
...
+        assertEquals( 1, options.get( 0 ).settingGroup().values( emptyMap() ).get("myInt") );
+        assertEquals( 123, options.get( 0 ).settingGroup().values( config ).get("myInt") );
+        assertEquals( Optional.empty(), options.get( 0 ).description() );
+
+        assertEquals( "bob", options.get( 1 ).settingGroup().values( emptyMap() ).get("myString") );
+        assertEquals( "bah", options.get( 1 ).settingGroup().values( config ).get("myString") );
+        assertEquals( "A string setting", options.get( 1 ).description().get() );
+    }
+
+    private static class Test implements LoadableConfig
```
chrisvest: You could name this `TestConfig` and avoid the conflict with the `org.junit.Test` class.

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
chrisvest: Could that cycle be solved by moving `ConfigLoader` to `configuration`?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseBuilder.java
```diff
@@ -39,8 +39,13 @@
...
 public class GraphDatabaseBuilder
 {
+    @Deprecated
     public interface DatabaseCreator
     {
+        /**
+         * This method will no longer accept a map in 4.0, but a Config object instead.
+         */
+        @Deprecated
         GraphDatabaseService newDatabase( Map<String, String> config );
```
chrisvest: You could already now add the `newDatabase( Config )` method, and turn this old declaration into a `default` method implemented as `newDatabase( Config.empty().augment( config ) );`. Or so I think.

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
chrisvest: The page cache size is the main differentiator, I think.

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -153,28 +271,83 @@ public Config with( Map<String, String> additionalConfig, Class<?>... settingsCl
...
-    public Iterable<Class<?>> getSettingsClasses()
-    {
-        return settingsClasses;
-    }
-
+    /**
+     * Replays possible existing messages into this log
+     *
+     * @param log to use
+     */
```
chrisvest: The main function of this method is assigning the `log` field. The replying of log messages is a side-effect, if the previous value was a `BufferingLog`.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ConnectorValidator.java
```diff
@@ -0,0 +1,127 @@
...
+import javax.annotation.Nonnull;
+
+import org.neo4j.graphdb.config.InvalidSettingException;
+import org.neo4j.graphdb.config.SettingGroup;
+
+import static org.neo4j.kernel.configuration.Settings.options;
+import static org.neo4j.kernel.configuration.Settings.setting;
+
+
+public abstract class ConnectorValidator implements SettingGroup<Object>
```
chrisvest: Two consecutive empty lines might make checkstyle complain.

enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
chrisvest: Can't such support be added without breaking existing API?

enterprise/server-enterprise/src/main/java/org/neo4j/server/enterprise/EnterpriseServerSettings.java
```diff
@@ -29,29 +30,22 @@
...
 @Description("Settings available in the Enterprise server")
-public interface EnterpriseServerSettings
+public class EnterpriseServerSettings implements LoadableConfig
 {
-    @Description( "Configure the operating mode of the database -- 'SINGLE' for stand-alone operation, 'HA' for " +
-            "operating as a member in a cluster, 'ARBITER' for an HA-only cluster member with no database, " +
-            "CORE for a core member of a Causal Clustering cluster, or READ_REPLICA for read replica." +
-            " cluster" +
-            ". " )
-    Setting<String> mode = setting( "dbms.mode", STRING, "SINGLE" );
```
chrisvest, line 39: Why is it okay to remove this?

spacecowboy: It has been moved (and converted to an enum) to [ClusterSettings.java](https://github.com/neo4j/neo4j/blob/c8932e8f5e8b1eda8f905bb8b33cc8059a16121e/enterprise/cluster/src/main/java/org/neo4j/cluster/ClusterSettings.java)

reason being that the mode setting must be accessible by all the cluster database creators, as well as the server endpoint. The cluster module satisfied that requirement, though it is not the perfect place for this setting, I agree.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseBuilder.java
spacecowboy: Not exactly. That would break the implementation of anyone implementing the interface. To be non-breaking the new method would have to have a default implementation which delegates to the old method. But good idea nonetheless! See changes.

spacecowboy:
thanks for you feedback @chrisvest , I have updated accordingly.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -110,6 +107,11 @@
     @Internal
     public static final Setting<Boolean> dump_configuration = setting("unsupported.dbms.report_configuration", BOOLEAN, FALSE );
 
+    @Description( "A strict configuration validation will prevent the database from starting up if unknown " +
+            "configuration options are specified in the neo4j settings namespace (such as dbms., ha., cypher., etc)." )
+    public static final Setting<Boolean> strict_config_validation =
+            setting("dbms.config.strict_validation", BOOLEAN, FALSE );
```
benbc, line 114: Maybe put a note that the default will switch to `true` in a future version.

PR #8459, opened 2016-11-30
Adds PageCursor#clear()

tinwelint:
which sets all bytes in the page to 0. Useful when reusing pages which have had
previous data in them. Clearing makes reused pages look like completely new ones.

### DIFF
--- a/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
@@ -778,7 +778,7 @@ abstract class MuninnPageCursor extends PageCursor
     }
 
     @Override
-    public void clear()
+    public void zapPage()
     {
         if ( pageSize == 0 )
         {
@@ -789,7 +789,6 @@ abstract class MuninnPageCursor extends PageCursor
         else
         {
             UnsafeUtil.setMemory( pointer, pageSize, (byte) 0 );
-            offset = 0;
         }
     }
 }
--- a/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
@@ -776,4 +776,20 @@ abstract class MuninnPageCursor extends PageCursor
             this.cursorException = message;
         }
     }
+
+    @Override
+    public void clear()
+    {
+        if ( pageSize == 0 )
+        {
+            // if this page has been closed then pageSize == 0 and we must adhere to making writes
+            // trigger outOfBounds when closed
+            outOfBounds = true;
+        }
+        else
+        {
+            UnsafeUtil.setMemory( pointer, pageSize, (byte) 0 );
+            offset = 0;
+        }
+    }
 }
--- a/community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
+++ b/community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
@@ -88,6 +88,7 @@ import static org.hamcrest.Matchers.lessThanOrEqualTo;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.Matchers.sameInstance;
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
@@ -1183,6 +1184,7 @@ public abstract class PageCacheTest<T extends PageCache> extends PageCacheTestSu
         testTemplate.accept( ( cursor ) -> cursor.putInt( 0, 1 ) );
         testTemplate.accept( ( cursor ) -> cursor.putLong( 0, 1 ) );
         testTemplate.accept( ( cursor ) -> cursor.putShort( 0, (short) 1 ) );
+        testTemplate.accept( ( cursor ) -> cursor.clear() );
     }
 
     private void checkUnboundReadCursorAccess( PageCursorAction action ) throws IOException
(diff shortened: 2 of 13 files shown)


community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
```diff
@@ -5370,4 +5375,46 @@ public void streamFilesRecursiveRenameMustNotChangeSourceFileContentsWithReplace
...
+            // THEN
+            try ( PageCursor cursor = pagedFile.io( pageId, PF_SHARED_READ_LOCK ) )
+            {
+                assertTrue( cursor.next() );
+
+                byte[] read = new byte[filePageSize];
+                cursor.getBytes( read );
+                assertFalse( cursor.checkAndClearBoundsFlag() );
+                assertArrayEquals( allZeros, read );
+            }
```
chrisvest, line 5420: This test is racy without a `shouldRetry`.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -776,4 +776,20 @@ public void setCursorException( String message )
             this.cursorException = message;
         }
     }
+
+    @Override
+    public void clear()
```
chrisvest: I'd prefer to call it `zapPage` or something like that, to avoid confusing it with `ByteBuffer.clear()`.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -776,4 +776,20 @@ public void setCursorException( String message )
...
+        if ( pageSize == 0 )
+        {
+            // if this page has been closed then pageSize == 0 and we must adhere to making writes
+            // trigger outOfBounds when closed
+            outOfBounds = true;
+        }
+        else
+        {
+            UnsafeUtil.setMemory( pointer, pageSize, (byte) 0 );
+            offset = 0;
```
chrisvest: Why do you need to reset the offset as well? It sounds unrelated.

community/io/src/test/java/org/neo4j/io/pagecache/impl/ByteBufferPage.java
```diff
@@ -135,4 +137,17 @@ public long address()
...
+    {
+        int remaining = buffer.capacity();
+        buffer.position( 0 );
+        while ( remaining >= ZEROS.length )
+        {
+            buffer.put( ZEROS );
+            remaining -= ZEROS.length;
+        }
+        buffer.put( ZEROS, 0, remaining );
+        buffer.position( 0 );
```
chrisvest: No, just call `setMemory()` on `address()` and remove that `ZEROS` constant.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -776,4 +776,20 @@ public void setCursorException( String message )
...
     }
+
+    @Override
+    public void clear()
+    {
+        if ( pageSize == 0 )
+        {
+            // if this page has been closed then pageSize == 0 and we must adhere to making writes
+            // trigger outOfBounds when closed
+            outOfBounds = true;
```
chrisvest, line 787: I'm not seeing a test for this case.

chrisvest (changes requested):

community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
tinwelint: oops

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
tinwelint: sure, why not. Although "zap" could be interpreted as really pulverizing the page, its data is merely being zero:ed out.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
tinwelint: There is, comment it out and you'll see at least one test failing

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
tinwelint: It felt relevant, perhaps it isn't. For the driving use case it isn't necessary, but zapping it really feels like a `reset --hard`, so to speak, which sort of implies offset reset as well. Doesn't matter to me.

community/io/src/test/java/org/neo4j/io/pagecache/impl/ByteBufferPage.java
tinwelint: Oh, so it's _always_ going to be a direct buffer? Is there anything asserting that?

chrisvest: Yes, the constructor is checking this.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
burqen: How about `clean` or `cleanPage`?

burqen:
@chrisvest Comments addressed

chrisvest (approved):

chrisvest:
@burqen @tinwelint You forgot to implement this new method for `org.neo4j.index.gbptree.TestPageCursor`.

burqen:
@chrisvest Yeah, that was... embarrassing.

PR #8460, opened 2016-11-30
Make sure that LockResults always have a message, when the status is not OK_LOCKED

chrisvest:
This should help us investigate any locking related HA test failures that might pop up.


MishaDemianenko:
@chrisvest i think we need a test here to verify that everything works as expected with different kind of messages

chrisvest:
This needs to be done in 3.2 because it requires a protocol version update.

PR #8461, opened 2016-11-30
Fail store initialisation in case when header record can't be read

MishaDemianenko:
Throw StoreNotFoundException in case if we can't read first page of a store file to get header information out of it.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
@@ -284,7 +284,8 @@ public abstract class CommonAbstractStore<RECORD extends AbstractBaseRecord,HEAD
                     }
                     else
                     {
-                        throw new StoreNotFoundException( "Store file not found: " + storageFileName );
+                        throw new StoreNotFoundException( "Fail to read header record of store file: " +
+                                storageFileName);
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


burqen:
@MishaDemianenko Looks good. But I wonder, is this a problems we've seen? I would expect 
```
void initialise( boolean createIfNotExists )
    {
        try
        {
            checkStorage( createIfNotExists );
            loadStorage();
        }
....
}
```
`checkStorage` to either fail fast or create the store file. How can we get to `loadStorage()` without having the store file?

MishaDemianenko:
@burqen we have that check already before, and how that can be possible at all: https://trello.com/c/mlMEhrgm/818-gh-8387-starting-and-immediately-restarting-corrupts-the-empty-database

burqen:
@MishaDemianenko Cool, approved by me. @chrisvest do you have anything?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
```diff
@@ -282,6 +282,10 @@ private void extractHeaderRecord() throws IOException
                                     pageCache.pageSize() + " bytes.");
                         }
                     }
+                    else
+                    {
+                        throw new StoreNotFoundException( "Store file not found: " + storageFileName );
```
chrisvest: I don't think this is quite true, this error message. If the file did not exists, we would not be able to map it. It should instead say something about not finding the header it expected the store file to have. I think this can still be a `StoreNotFoundException`.

chrisvest (changes requested):
I like the general idea, but I want a different error message.

MishaDemianenko:
@chrisvest agree do not know why i came out with that message. Updated.

chrisvest (approved):

PR #8464, opened 2016-12-01
Simplify NeoStoreDataSource stop procedure.

MishaDemianenko:


tinwelint:
High-level feedback: I like the simplification in stop, and the other simplifications around NSDS as well, great work!

Lemme summarize how I would like shutdown to behave: I'd like a call to `db.shutdown()` never result in a non-clean shutdown, i.e. need for recovery on next startup. It's fine to block new transactions and be nice and wait for in-flight transactions (i.e. both pre-commit: those that in the process of gathering changes and committing transactions) for a while before shutting down and then make a hard cut where transactions that are still in pre-commit, but haven't yet gotten to commit will be terminated, but those who have gotten to commit will be allowed to successfully complete their commit. This should result in a clean shutdown no matter what, shouldn't it.

I think this change puts us close to that, perhaps with one concern which is that NeoStoreDataSource#stop() (which comes after DatabaseAvailability has naturally been nice and awaited in-flight transactions for a while), after calling life.stop(), immediately terminates all transactions and waits for them to close. I'd expect calling terminate on transactions that are in commit phase could interrupt them, specifically around their locks instance, which gets a call to stop(), but I'm not sure.

Do my summary of what I'd expect align with yours here? Can we get some more confidence about shutdown guaranteeing this?

MishaDemianenko:
@tinwelint that's true that current implementation is terminating and closing transactions and fast as possible and waits only for transactions that already closing and doing commit, perhaps that is too aggressive and we should be more tolerant here and give a bit more time to finish work that is already in progress.
Let me try to do that

MishaDemianenko:
@tinwelint should be ready now

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -191,9 +224,8 @@
 
     @Description( "The maximum amount of time to wait for running transactions to complete before allowing "
                   + "initiated database shutdown to continue" )
-    @Internal
     public static final Setting<Long> shutdown_transaction_end_timeout =
-            setting( "unsupported.dbms.shutdown_transaction_end_timeout", DURATION, "10s" );
+            setting( "dbms.shutdown_transaction_end_timeout", DURATION, "10s" );
```
tinwelint, line 228: Don't we need migration for this? Feels like we should make things easy for (albeit few) existing users of an unsupported setting when we promote such to a supported setting.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
```diff
@@ -172,11 +171,10 @@ void assertOpen()
             throw new NotInTransactionException( "The statement has been closed." );
         }
 
-        Status terminationReason = transaction.getReasonIfTerminated();
-        if ( terminationReason != null )
+        transaction.getReasonIfTerminated().ifPresent( reason ->
```
tinwelint: `assertOpen` is called very very often. Isn't it a bad idea to create this lambda for every call to it? Just because Option do exist does it mean that it should be used in all places where null was used IMHO. Even when speculating about escape analysis and how the JVM may avoid creating that instance it feels unnecessary to even give it a chance to create it and potentially contribute to garbage creep.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -815,6 +742,179 @@ public String toString()
...
+            return DATA;
+        }
+
+        TransactionWriteState upgradeToSchemaWrites() throws InvalidTransactionTypeKernelException
+        {
+            return SCHEMA;
+        }
+    }
+
+    static class TransactionStatus
```
tinwelint: I wonder why this class (and the changes it involves) was introduced. In frequently accessed methods in KTI, instead of checking a non-volatile boolean the execution path now includes following reference to TransactionStatus object, following reference to AtomicInteger and incur volatile read/write on internal state in it. That in itself is a big difference. Why is this change made in this PR? Why is it made at all? Has his change been properly benchmarked (micro and macro)?

community/kernel/src/main/java/org/neo4j/kernel/impl/core/ThreadToStatementContextBridge.java
```diff
@@ -71,11 +71,10 @@ private void assertInUnterminatedTransaction( KernelTransaction transaction )
         {
             throw new BridgeNotInTransactionException();
         }
-        Status terminationReason = transaction.getReasonIfTerminated();
-        if ( terminationReason != null )
+        transaction.getReasonIfTerminated().ifPresent( reason ->
```
tinwelint: Same concern here as above about garbage creep.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -747,10 +742,9 @@ protected boolean fetchNext()
 
     private void assertTransactionOpen()
     {
-        Status reason = spi.currentTransaction().getReasonIfTerminated();
-        if ( reason != null )
+        spi.currentTransaction().getReasonIfTerminated().ifPresent( reason ->
```
tinwelint: Same concern here as above about garbage creep.

tinwelint (changes requested):
There's a bit more than expected in this PR, also potentially the "other stuff" goes in the wrong direction.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
MishaDemianenko: yep, migration added

MishaDemianenko:
@tinwelint updated

tinwelint (approved):

PR #8465, opened 2016-12-01
Add deferred locks to a list of enterprise edition dependencies

MishaDemianenko:
Add deferred locks to a list of dependencies to publish it into maven repo as part of release.

### DIFF
--- a/enterprise/pom.xml
+++ b/enterprise/pom.xml
@@ -119,6 +119,11 @@ terms of the relevant Commercial Agreement.
           <artifactId>neo4j-harness-enterprise</artifactId>
           <version>${project.version}</version>
         </dependency>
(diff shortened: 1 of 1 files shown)


benbc:
@MishaDemianenko Yes, I think that's correct. We can check we've got it right by looking at the `packaging` build artifacts.

obrienmi:
Hi, guys, I am Michael O'Brien at Nokia - I'll be watching this issue and bringing in the change at the next push of artifacts to maven in our upgrade from 2.3.7 to 3.1.0-RC (and possibly 3.0.8)
Thanks for the quick response.
/michael

obriensystems:
btw - isolation level workaround for us is good in 3.1.1

PR #8468, opened 2016-12-02
Fix flaky TxPushStrategyConfigIT

davidegrohmann:
Since transaction push is on a best effort bases, sometimes there are
missing pushed txs, test assertion should take that into account.

### DIFF
--- a/enterprise/ha/src/test/java/org/neo4j/kernel/ha/TxPushStrategyConfigIT.java
+++ b/enterprise/ha/src/test/java/org/neo4j/kernel/ha/TxPushStrategyConfigIT.java
@@ -23,7 +23,7 @@ import org.junit.Rule;
 import org.junit.Test;
 
 import java.util.ArrayList;
-import java.util.Collections;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


enterprise/ha/src/test/java/org/neo4j/kernel/ha/TxPushStrategyConfigIT.java
```diff
@@ -282,24 +287,29 @@ public void failure( Throwable ex )
...
         {
             this.serverId = serverId;
             this.txId = txId;
+            this.missed = missed;
         }
 
         public void format( StringBuilder failures, long txId )
         {
-            if ( txId != this.txId )
+            if ( txId < this.txId - this.missed )
```
martinfurmanski: Maybe (txId < thix.txId - this.missed || txId > this.txId) to keep the old semantics of failing on excessive amounts of transactions as well? Just an observation that the upper range was changed as well.

martinfurmanski (approved):

enterprise/ha/src/test/java/org/neo4j/kernel/ha/TxPushStrategyConfigIT.java
davidegrohmann: Sure

PR #8471, opened 2016-12-02
Simplify test to use a single thread

davidegrohmann:
This should solve all the races seen in CI.

### DIFF
--- a/enterprise/ha/src/test/java/org/neo4j/kernel/ha/lock/ClusterLocksIT.java
+++ b/enterprise/ha/src/test/java/org/neo4j/kernel/ha/lock/ClusterLocksIT.java
@@ -42,6 +42,7 @@ import org.neo4j.test.ha.ClusterRule;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.neo4j.helpers.Exceptions.stringify;
+import static org.neo4j.kernel.ha.cluster.HighAvailabilityMemberState.PENDING;
 import static org.neo4j.kernel.impl.ha.ClusterManager.allSeesAllAsAvailable;
 import static org.neo4j.kernel.impl.ha.ClusterManager.instanceEvicted;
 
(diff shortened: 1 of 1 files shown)


enterprise/ha/src/test/java/org/neo4j/kernel/ha/lock/ClusterLocksIT.java
```diff
@@ -83,49 +84,26 @@ public void lockCleanupOnModeSwitch() throws Throwable
...
         createNodeOnMaster( testLabel, cluster.getMaster() );
         cluster.sync();
+
         HighlyAvailableGraphDatabase slave = cluster.getAnySlave();
+        cluster.fail( slave, ClusterManager.NetworkFlag.values() );
+        cluster.await( instanceEvicted( slave ) );
 
-        AtomicBoolean run = new AtomicBoolean( true );
-        AtomicReference<Throwable> ref = new AtomicReference<>(  );
+        assertEquals( PENDING, slave.getInstanceState() );
```
martinfurmanski, line 95: Are we sure it should go to PENDING before cluster considers it evicted? Perhaps assertEventually(..) with timing based on some ha/cluster timing setting?

martinfurmanski (approved):
Looks good to me, but minor point raised.

enterprise/ha/src/test/java/org/neo4j/kernel/ha/lock/ClusterLocksIT.java
davidegrohmann: If it is evicted it _is_ in pending.

martinfurmanski: Ok the code / test framework is a bit weird here. It seems instanceEvicted means that the particular member itself thinks that it has the role of UNKNOWN. I thought it would be how the remaining members view the state of the slave and the transitions to evicted and pending would be asynchronous.

davidegrohmann: UNKNOWN==PENDING

PR #8473, opened 2016-12-03
clear the leader if we're stepping down

mneedham:


mneedham:
Actually now I think about this more I realise this isn't quite right. There are some scenarios where we step down as leader but the new leader is set in `Outcome` and this PR wouldn't set that new leader. 

I think we only need to set it to null on stepping down if it's set to us in the Outcome.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMachine.java
```diff
@@ -257,6 +257,10 @@ private boolean leaderChanged( Outcome outcome, MemberId oldLeader )
         {
             return true;
         }
+        else if ( oldLeader != null && outcome.getLeader() == null )
```
martinfurmanski: Wouldn't it be better to simply have it set to null but make the users handle that case? It seems like the correct thing to do so that external components are forced to handle the case of there being no known leader. Otherwise they might keep trying. Today of course that means only a single component, the replicator.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/locks/LeaderOnlyLockManager.java
```diff
@@ -108,7 +108,7 @@ private synchronized int acquireTokenOrThrow()
         {
             future = replicator.replicate( lockTokenRequest, true );
         }
-        catch ( InterruptedException e )
+        catch ( InterruptedException | NoLeaderFoundException e )
         {
             throw new AcquireLockTimeoutException( e, "Interrupted acquiring token.", Interrupted );
```
martinfurmanski: Wrong message!

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/RaftReplicator.java
```diff
@@ -80,13 +74,13 @@ public RaftReplicator( LeaderLocator leaderLocator, MemberId me,
...
             assertDatabaseNotShutdown();
-            outbound.send( leader, new RaftMessages.NewEntry.Request( me, operation ) );
             try
             {
+                outbound.send( leaderLocator.getLeader(), new RaftMessages.NewEntry.Request( me, operation ) );
                 progress.awaitReplication( timeout.getMillis() );
                 timeout.increment();
             }
-            catch ( InterruptedException e )
+            catch ( InterruptedException | NoLeaderFoundException e )
```
martinfurmanski: I think it would be good if we kept retrying here, to keep up exactly-once behaviour as far as possible. So the NoLeader case can be handled by just ignoring or logging it perhaps. You could do this by putting the send() in it's own try-catch. We clearly have no test catching this, so at the very least we should add a comment about the reasoning here.

martinfurmanski (changes requested):

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/locks/LeaderOnlyLockManager.java
mneedham: Oops!

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/RaftReplicator.java
mneedham: Ah yes good call. Have made it log and then carry on in the `do...while()` loop

PR #8474, opened 2016-12-04
GBTree stable generation copy-on-write

tinwelint:
Further implements tree generations by making sure changes aren't applied in stables nodes, but instead copying stable node into unstable generation and making change there. Nodes now each have a generation and stable node gets "new generation" forward-pointer to the new generation node. The new-generation pointer is rarely actually traversed because parent nodes redirect to new nodes directly, they are there for when reader goes to a stable node, which concurrently gets a new version of itself created.

This PR is a reasonably coherent chunk of this generation work and focuses on InternalTreeLogic, whereas there's further work to be done inside GBPTree to complete the generation work.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
@@ -417,12 +417,14 @@ public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
     private class SingleIndexWriter implements IndexWriter<KEY,VALUE>
     {
         private final InternalTreeLogic<KEY,VALUE> treeLogic;
+        private final StructurePropagation<KEY> structurePropagation;
         private PageCursor cursor;
         private IndexWriter.Options options;
         private final byte[] tmp = new byte[0];
 
         SingleIndexWriter( InternalTreeLogic<KEY,VALUE> treeLogic )
         {
+            this.structurePropagation = new StructurePropagation<>( layout.newKey() );
             this.treeLogic = treeLogic;
         }
 
@@ -444,11 +446,12 @@ public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
         {
             goToRoot( cursor );
 
-            SplitResult<KEY> split = treeLogic.insert( cursor, key, value, valueMerger, options,
+            treeLogic.insert( cursor, structurePropagation, key, value, valueMerger, options,
                     stableGeneration, unstableGeneration );
 
-            if ( split != null )
+            if ( structurePropagation.hasSplit )
             {
+                structurePropagation.hasSplit = false;
                 // New root
                 long newRootId = acquireNewId();
                 if ( !cursor.next( newRootId ) )
@@ -457,12 +460,17 @@ public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
                 }
 
                 bTreeNode.initializeInternal( cursor, stableGeneration, unstableGeneration );
-                bTreeNode.insertKeyAt( cursor, split.primKey, 0, 0, tmp );
+                bTreeNode.insertKeyAt( cursor, structurePropagation.primKey, 0, 0, tmp );
                 bTreeNode.setKeyCount( cursor, 1 );
-                bTreeNode.setChildAt( cursor, split.left, 0, stableGeneration, unstableGeneration );
-                bTreeNode.setChildAt( cursor, split.right, 1, stableGeneration, unstableGeneration );
+                bTreeNode.setChildAt( cursor, structurePropagation.left, 0, stableGeneration, unstableGeneration );
+                bTreeNode.setChildAt( cursor, structurePropagation.right, 1, stableGeneration, unstableGeneration );
                 rootId = newRootId;
             }
+            else if ( structurePropagation.hasNewGen )
+            {
+                rootId = structurePropagation.left;
+            }
+            structurePropagation.hasNewGen = false;
 
             checkOutOfBounds( cursor );
         }
@@ -472,7 +480,13 @@ public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
         {
             goToRoot( cursor );
 
-            VALUE result = treeLogic.remove( cursor, key, layout.newValue(), stableGeneration, unstableGeneration );
+            VALUE result = treeLogic.remove( cursor, structurePropagation, key, layout.newValue(),
+                    stableGeneration, unstableGeneration );
+            if ( structurePropagation.hasNewGen )
+            {
+                structurePropagation.hasNewGen = false;
+                rootId = structurePropagation.left;
+            }
 
             checkOutOfBounds( cursor );
             return result;
--- a/community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
@@ -75,7 +75,6 @@ import org.neo4j.io.pagecache.PagedFile;
  */
 public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
 {
-
     /**
      * Paged file in a {@link PageCache} providing the means of storage.
      */
@@ -315,6 +314,7 @@ public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
             }
             while ( cursor.shouldRetry() );
             checkOutOfBounds( cursor );
+
             if ( isInternal )
             {
                 PointerChecking.checkChildPointer( childId );
--- a/community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
@@ -75,11 +75,6 @@ import org.neo4j.io.pagecache.PagedFile;
  */
 public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
 {
-    /**
-     * Page id of the meta page holding information about root id and custom user meta information.
-     * This page id is statically allocated throughout the life of a tree.
-     */
-    private static final int META_PAGE_ID = 0;
 
     /**
      * Paged file in a {@link PageCache} providing the means of storage.
@@ -120,7 +115,7 @@ public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
     /**
      * Current page id which contains the root of the tree.
      */
-    private volatile long rootId = META_PAGE_ID + 1;
+    private volatile long rootId = IdSpace.MIN_TREE_NODE_ID;
 
     /**
      * Last allocated page id, used for allocating new ids as more data gets inserted into the tree.
@@ -277,10 +272,10 @@ public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
 
     private PageCursor openMetaPageCursor( PagedFile pagedFile ) throws IOException
     {
-        PageCursor metaCursor = pagedFile.io( META_PAGE_ID, PagedFile.PF_SHARED_WRITE_LOCK );
+        PageCursor metaCursor = pagedFile.io( IdSpace.META_PAGE_ID, PagedFile.PF_SHARED_WRITE_LOCK );
         if ( !metaCursor.next() )
         {
-            throw new IllegalStateException( "Could not go to meta data page " + META_PAGE_ID );
+            throw new IllegalStateException( "Couldn't go to meta data page " + IdSpace.META_PAGE_ID );
         }
         return metaCursor;
     }
--- a/community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
@@ -321,10 +321,7 @@ public class GBPTree<KEY,VALUE> implements Index<KEY,VALUE>, IdProvider
             {
                 PointerChecking.checkChildPointer( childId );
 
-                if ( !cursor.next( childId ) )
-                {
-                    throw new IllegalStateException( "Couldn't go to child " + childId );
-                }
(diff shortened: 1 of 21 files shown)


community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
```diff
@@ -518,11 +591,81 @@ private VALUE removeFromLeaf( PageCursor cursor, KEY key, VALUE into )
...
+
+        // Do copy
+        long newGenId = idProvider.acquireNewId();
+        try ( PageCursor newGenCursor = cursor.openLinkedCursor( newGenId ) )
+        {
+            if ( !newGenCursor.next() )
+            {
+                throw new IllegalStateException( "Could not go to new node " + newGenId );
+            }
+            cursor.copyTo( 0, newGenCursor, 0, cursor.getCurrentPageSize() );
```
chrisvest, line 629: 👍

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointerPair.java
```diff
@@ -387,9 +387,8 @@ private static byte pointerState( int stableGeneration, int unstableGeneration,
         }
         if ( generation < MIN_GENERATION )
         {
-            // todo Better error message?
             throw new UnsupportedOperationException( "Generation was less than MIN_GENERATION " + MIN_GENERATION +
-                                                     " but checksum was correct. This should not happen." );
+                    " and but checksum was correct. Pointer was " + generation + "," + pointer );
```
chrisvest: `and but`? Needs moar english.

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -124,7 +124,6 @@ public boolean next() throws IOException
                 else
```
chrisvest, line 123: Looks like the `reread` and `resetPosition` fields do not need to be set to `false` in the logic above, because the `while` expression always overwrites them. Small simplification.

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
```diff
@@ -392,7 +392,6 @@ private int readRecordsWithInsertRecordInPosition( PageCursor cursor, Consumer<P
         cursor.getBytes( into, 0, insertPosition * recordSize );
 
         // Read newRecord
-        // TODO: A bit expensive to wrap in a PageCursor tin foil just to write this middle key
         PageCursor buffer = ByteArrayPageCursor.wrap( into, insertPosition * recordSize, recordSize );
         newRecordWriter.accept( buffer );
```
chrisvest, line 430: We could add some `PageCursor.copyTo(ByteBuffer)` and `PageCursor.copyFrom(ByteBuffer)`, and then you would be able to allocate direct ByteBuffers and work with those, instead of these on-heap byte arrays. Something to look at in the future.

community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
```diff
@@ -135,40 +151,45 @@
...
 
-        goTo( cursor, childId );
+        goTo( cursor, childId, stableGeneration, unstableGeneration );
 
-        SplitResult<KEY> split = insert( cursor, key, value, valueMerger, options, stableGeneration,
-                unstableGeneration );
+        insert( cursor, structurePropagation, key, value, valueMerger, options, stableGeneration, unstableGeneration );
 
-        goTo( cursor, currentId );
+        goTo( cursor, currentId, stableGeneration, unstableGeneration );
```
chrisvest: I wonder if opening a linked cursor for the child is faster than moving the given one back and forth.

community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
```diff
@@ -465,16 +486,37 @@ private static int middle( int keyCountAfterInsert, float splitLeftChildSize )
...
+     * Otherwise {@code null}.
+     * @throws IOException on cursor failure
+     */
+    VALUE remove( PageCursor cursor, StructurePropagation<KEY> structurePropagation, KEY key, VALUE into,
+            long stableGeneration, long unstableGeneration ) throws IOException
     {
         if ( bTreeNode.isLeaf( cursor ) )
         {
-            return removeFromLeaf( cursor, key, into );
+            return  removeFromLeaf( cursor, structurePropagation, key, into, stableGeneration, unstableGeneration );
```
chrisvest: Accidental space.

community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
```diff
@@ -486,16 +528,41 @@ public VALUE remove( PageCursor cursor, KEY key, VALUE into, long stableGenerati
...
+     * @param key key to be removed
+     * @param into {@code VALUE} instance to write removed value to
+     * @param stableGeneration stable generation, i.e. generations <= this generation are considered stable.
+     * @param unstableGeneration unstable generation, i.e. generation which is under development right now.
+     * @return Provided {@code into}, populated with removed value for convenience if {@code key} was removed.
+     * Otherwise {@code null}.
+     * @throws IOException on cursor failure
+     */
+    private VALUE removeFromLeaf( PageCursor cursor, StructurePropagation<KEY> structurePropagation,
+            KEY key, VALUE into, long stableGeneration, long unstableGeneration ) throws IOException
```
chrisvest, line 568: Could `StructurePropagation` be used for the out-parameter here, or is that conflating concerns? Value types sounds like they could be of some use here, if we had them.

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
```diff
@@ -370,6 +371,26 @@ int readChildrenWithInsertRecordInPosition( PageCursor cursor, Consumer<PageCurs
                 childSize(), childOffset( 0 ), into );
     }
 
+    void goTo( PageCursor cursor, long childId, long stableGeneration, long unstableGeneration )
+            throws IOException
+    {
+        if ( !cursor.next( childId ) )
+        {
+            throw new IllegalStateException( "Could not go to " + childId );
```
chrisvest, line 379: This error message could be improved. If this pops up in front a user, they'd have no idea what to do. The `childId` should at least have a unit associated with it, and we could qualify the message with a `GBPTree Index Structure Error: ` prefix. Perhaps you should make a custom runtime exception subclass for these runtime index exceptions. It would also be nice to attach which specific index the error relates to, in case people want to try to remedy the situation with a drop+create index command sequence.

chrisvest (changes requested):
I had a number of comments. Please take a look at them.

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
tinwelint: I would like to get rid of the temp copying all together in the end, by using directly moving stuff from cursor to cursor, or in case of plain insert, move around keys/values inside the cursor w/o buffer in between. I think we should defer this to a slightly later optimization stage.

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
tinwelint: Good point

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
chrisvest: That sounds even better!

chrisvest:
The main thing I would like to see improved is the error messages thing.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -447,25 +441,31 @@ public void merge( KEY key, VALUE value, ValueMerger<VALUE> valueMerger ) throws
         {
             goToRoot( cursor );
 
-            SplitResult<KEY> split = treeLogic.insert( cursor, key, value, valueMerger, options,
+            treeLogic.insert( cursor, structurePropagation, key, value, valueMerger, options,
```
MishaDemianenko, line 443: `structurePropagation` is storage of split result that particular insert/remove cause for a tree and depending from results of that we will do some post steps, it also means that if we will forget to clear it or will fail to do that that can potentially influence future step. Can we have block with reset of that context object to clean initial state in finally section or before using it?

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointer.java
```diff
@@ -43,11 +43,11 @@
  */
 class GenSafePointer
 {
-    public static final long MIN_GENERATION = 1L;
+    static final long MIN_GENERATION = 1L;
```
MishaDemianenko, line 46: this class have a lot of static methods that are just helpers and i find it hard to see what it has to do with GenSafePointer (GenerationalPointer), also comment in java doc mentions one return time as a limitation of some sort can we talk about that maybe?

community/index/src/main/java/org/neo4j/index/gbptree/IdSpace.java
```diff
@@ -0,0 +1,46 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.index.gbptree;
+
+public class IdSpace
```
MishaDemianenko: one liner java doc about that is that would be super nice

community/index/src/main/java/org/neo4j/index/gbptree/StructurePropagation.java
```diff
@@ -25,9 +25,16 @@
  *
  * @param <KEY> type of key.
  */
-class SplitResult<KEY>
+class StructurePropagation<KEY>
```
MishaDemianenko, line 28: i would propose to rename it to SplitContext or UpdateContext, and also include in it generational information for example to simplify InternalTreeLogic methods signatures, basically use it as context object instead.

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
```diff
@@ -65,15 +66,16 @@
 {
     private static final int SIZE_PAGE_REFERENCE = GenSafePointerPair.SIZE;
     private static final int BYTE_POS_TYPE = 0;
-    private static final int BYTE_POS_KEYCOUNT = BYTE_POS_TYPE + 1;
+    private static final int BYTE_POS_GEN = BYTE_POS_TYPE + 1;
+    private static final int BYTE_POS_KEYCOUNT = BYTE_POS_GEN + 4;
```
MishaDemianenko: Integer.BYTES ?

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
```diff
@@ -65,15 +66,16 @@
 {
     private static final int SIZE_PAGE_REFERENCE = GenSafePointerPair.SIZE;
     private static final int BYTE_POS_TYPE = 0;
-    private static final int BYTE_POS_KEYCOUNT = BYTE_POS_TYPE + 1;
+    private static final int BYTE_POS_GEN = BYTE_POS_TYPE + 1;
```
MishaDemianenko: Byte.BYTES ?

MishaDemianenko (changes requested):
Couple of additional comments so far:
1. It would be nice to have separation already between public and non public surface for index structure: at least public/private api package to see what we expose and what is completely internal
2. Should we have a put in IndexWriter that always return previous value and remove merge from that interface at all? (actual implementation can still have it if that is required )

tinwelint:
@MishaDemianenko 

1. The API lives in `org.neo4j.index` and the gb+tree in the `.gbptree` sub-package. That's not enough?

Also, we won't change that in this PR since other work would suffer too much from it. We can re-visit in the end.

2. The Native LSS needs proper merge so it'll need to stay.

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointer.java
tinwelint: Yeah, I don't understand your concern, so let's talk about it

community/index/src/main/java/org/neo4j/index/gbptree/IdSpace.java
tinwelint: Definitely

MishaDemianenko:
@tinwelint absolutely even all other comments can be fixed in next PRs if that easier for you guys.
Well we already have *.internal.* in part of the places to represent internal parts, and *.impl.* in some other parts of code, so i would like us to follow one of those conventions for api splitting.

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
tinwelint: Think so, yeah

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
tinwelint: Think so, yeah

community/index/src/main/java/org/neo4j/index/gbptree/StructurePropagation.java
tinwelint: An interesting idea, we still need to pass generation values further down to TreeNode, GenSafePointer and such, so wouldn't reduce that much overall. We're really really trying to avoid objects to cut down object indirections/traversals in heap to close to zero.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: That is a concern we had when making this change as well, perhaps it makes sense to add

PR #8475, opened 2016-12-05
Better exception handling during shutdown in Neo4jJobScheduler

davidegrohmann:
Make sure we don't swallow exceptions.


MishaDemianenko:
@davidegrohmann looks good. Do you need this change now or it can wait to be merged after release?

davidegrohmann:
@MishaDemianenko after release is fine

PR #8476, opened 2016-12-05
don't try and truncate things that don't exist

mneedham:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/segmented/Terms.java
```diff
@@ -88,6 +96,11 @@ private void setSize( int newSize )
      */
     synchronized void truncate( long fromIndex )
     {
+        if ( fromIndex < 0 || fromIndex < min )
```
davidegrohmann, line 99: isn't enough `fromIndex < min`? can't we assume `min >= 0`?

mneedham: When we start up for the first time it's set to `-1`!

PR #8477, opened 2016-12-05
Slightly improved logging

chrisvest:


chrisvest:
Build failed and this is not really necessary.

PR #8479, opened 2016-12-05
Checkpoint tuning

chrisvest:
When a check-point is running, it consumes IO, and this may negatively affect
the performance of the database. Especially on platforms where IO performance
is low (like rotating hard disk drives) or throttled (like AWS EBS volumes).
Therefor, I have tried to modify the defaults to minimase this influence:

* Tripled the check-point inteval so it runs every 15 minutes instead of every
  5 minutes. I have left the 100.000 transaction threshold the same, though.
* Cut the default IOPS limit by two thirds. I observed that we never appeared
  to reach the 1.000 IOPS default on AWS EBS anyway. Instead we were only
  reaching about 600 to 700 IOPS. The new default is half of this.
* Add a random offset to the initial check-point time, so instances that start
  in a cluster simultaneusly will attempt to schedule their check-point cycles
  out of phase.

cl [kernel] Lower default checkpoint system strain by changing checkpoint interval from 5m to 15m and iops limit from 1000 to 300.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -19,22 +19,55 @@
  */
 package org.neo4j.graphdb.factory;
 
-import org.neo4j.graphdb.config.Setting;
-import org.neo4j.helpers.AdvertisedSocketAddress;
-import org.neo4j.helpers.ListenSocketAddress;
-import org.neo4j.io.ByteUnit;
-import org.neo4j.kernel.configuration.*;
-import org.neo4j.kernel.impl.cache.MonitorGc;
-import org.neo4j.logging.Level;
-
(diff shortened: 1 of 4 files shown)


chrisvest:
Not ready for merge until we've soak tested it.

burqen:
@chrisvest Failed test seem to be flaky and unrelated to this change. Will give operability a chance to have a look before merging.

burqen:
@chrisvest Operability suggested to update the default configs to match as well.
https://github.com/neo4j/neo4j/blob/3.1/packaging/standalone/standalone-enterprise/src/main/distribution/text/enterprise/conf/neo4j.conf#L486

chrisvest:
@burqen Pushed.

burqen:
@chrisvest Over 1000 test failures. Seems like most of them is because of "address already in use" issues. Likely unrelated to this fix, but some other problem could be hiding among those failures. I triggered another build to see if it goes green.

PR #8481, opened 2016-12-06
GBPTree recovery

tinwelint:
**This PR contain a lot of stuff**
... but the main focus is on recovery.

**Recovery build on two important sub-parts:**

- *Tree state* (e.g. tree root id, generation etc.) comes in two pages, at least one of which will always be stable and consistent. An equivalent of GSPP, but for the tree itself.
- *Checkpoint algorithm*, creating consistent durable state of the tree between generations. Designed to be integrated with neo4j's checkpointing.

**Some minor changes to allow tree to recover after a crash:**

- Introduce `GBPTree.prepareForRecovery()` that will prepare tree to accept the operations that happened since last checkpoint to be re-applied. Creator of tree is responsible to call this method whenever recovery happens.
- In `GBPTree.acquireNewId()`. Overwrite newly acquired pages with zeroes to not leak state if page was reused.
- In `InternalTreeLogic. createUnstableVersionIfNeeded`. Overwrite `newGen` field in new generation of node.
- Checkpoint tree on first initialisation to be able to recover if crash happens before last checkpoint.

**Some other changes are:**

- Writing to tree and perform checkpointing need to be mutually exclusive to guarantee a consistent state after checkpoint. This is guarded by `writerCheckpointMutex` in GBPTree.
- `SeekCursor` is now responsible for traversal through internal nodes as well as leaf nodes. `GBPTree` was previously the one traversing the internal nodes.
- If a reader survives a checkpoint it now have the possibility to get an updated `stable` and `unstableGeneration`.
- Every `cursor.next` and `cursor.checkAndClearBoundsFlag` is now done using `PageCursorUtil` to unify exception behaviour and make code less cluttered by `if... throw`.

**Improved exceptions**
Introduce `TreeInconsistencyException` and `MetadataMismatchException` together with more detailed messages. What is still missing is some information about tree (file path, etc) for exceptions thrown from the internals of the tree. This is outside the scope of this PR but will be added soon.


community/index/src/main/java/org/neo4j/index/gbptree/Generation.java
```diff
@@ -0,0 +1,52 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.index.gbptree;
+
+/**
+ * Logic for composing and decomposing stable/unstable to/from a single {@code long}.
+ *
+ * <pre>
+ * long: [S,S,S,S,U,U,U,U]
```
ragadeeshu, line 26: Might want to clarify what S and U  is here, unsigned ints representing generation?

tinwelint: Stands for **S**table and **U**nstable as the above comment mentions, you think it needs clarification?

ragadeeshu: Logic for composing and decomposing stable/unstable generation number (uint) to/from a single {@code long}.

community/common/src/test/java/org/neo4j/test/rule/RandomRule.java
```diff
@@ -66,6 +70,11 @@ public Statement apply( final Statement base, Description description )
             public void evaluate() throws Throwable
             {
                 seed = specificSeed == null ? currentTimeMillis() : specificSeed;
+                Seed methodSeed = description.getAnnotation( Seed.class );
+                if ( methodSeed != null )
+                {
+                    seed = methodSeed.value();
```
MishaDemianenko, line 68: now It can be unfortunate that your specificSeed is ignored, maybe we can combine two way of specification of specific seed into one?

community/index/src/main/java/org/neo4j/index/gbptree/ConsistencyChecker.java
```diff
@@ -164,6 +190,47 @@ private void assertKeyOrder( PageCursor cursor, KeyRange<KEY> range )
         }
     }
 
+    static void assertNoCrashOrBrokenPointerInGSPP( PageCursor cursor, long stableGeneration, long unstableGeneration,
+            String pointerFieldName, int offset, TreeNode<?,?> treeNode )
+    {
+        cursor.setOffset( offset );
```
MishaDemianenko, line 196: we should check that we did not have any error during reading all the pointer from cursor and retry it in case we have before we start any interpretation of results especially throwing exceptions

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -242,139 +315,176 @@ private PagedFile openOrCreate( PageCache pageCache, File indexFile,
...
-
-    // Utility method
-    public void printTree() throws IOException
-    {
-        try ( PageCursor cursor = pagedFile.io( rootId, PagedFile.PF_SHARED_READ_LOCK ) )
+        try ( PageCursor cursor = pagedFile.io( lastId, PagedFile.PF_SHARED_WRITE_LOCK ) )
         {
             cursor.next();
-            TreePrinter.printTree( cursor, bTreeNode, layout, stableGeneration, unstableGeneration, System.out );
+            // TODO use cursor.clear() when available
```
MishaDemianenko, line 468: should be already there right?

community/index/src/main/java/org/neo4j/index/gbptree/TreeState.java
```diff
@@ -0,0 +1,123 @@
...
+            long lastId )
+    {
+        GenSafePointer.assertGenerationOnWrite( stableGeneration );
+        GenSafePointer.assertGenerationOnWrite( unstableGeneration );
+
+        writeStateOnce( cursor, stableGeneration, unstableGeneration, rootId, lastId ); // Write state
+        writeStateOnce( cursor, stableGeneration, unstableGeneration, rootId, lastId ); // Write checksum
+    }
+
+    static TreeState read( PageCursor cursor )
```
MishaDemianenko, line 83: we should check that we did not have any errors while reading from cursor before returning TreeState

community/index/src/main/java/org/neo4j/index/gbptree/TreeStatePair.java
```diff
@@ -0,0 +1,96 @@
...
+    {
+        TreeState newestValidState = selectNewestValidStateOrNull( states );
+        if ( newestValidState == null )
+        {
+            return states.getLeft();
+        }
+        return newestValidState == states.getLeft() ? states.getRight() : states.getLeft();
+    }
+
+    private static TreeState selectNewestValidStateOrNull( Pair<TreeState,TreeState> states )
```
MishaDemianenko: Shouldn't method return Optional<TreeState>?

community/index/src/main/java/org/neo4j/index/gbptree/TreeStatePair.java
```diff
@@ -0,0 +1,96 @@
...
+    {
+        TreeState stateA = states.getLeft();
+        TreeState stateB = states.getRight();
+
+        if ( stateA.isValid() != stateB.isValid() )
```
MishaDemianenko, line 64: this looks like a code that should be in a TreeStateComparator class?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -387,27 +497,99 @@ public void close() throws IOException
     @Override
     public IndexWriter<KEY,VALUE> writer( IndexWriter.Options options ) throws IOException
     {
-        SingleIndexWriter result = this.writer.getAndSet( null );
-        if ( result == null )
+        if ( !writerTaken.compareAndSet( false, true ) )
```
MishaDemianenko, line 526: now when we have a lock do we really need this guard?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -387,27 +497,99 @@ public void close() throws IOException
...
+    /**
+     * {@link GBPTree} class-level javadoc mentions how this method interacts with recovery,
+     * it's an essential piece to be able to recover properly and must be called when external party
+     * detects that recovery is required, before re-applying the recovered updates.
+     *
+     * @throws IOException on {@link PageCache} error.
+     */
+    public void prepareForRecovery() throws IOException
+    {
+        writerCheckpointMutex.lock();
```
MishaDemianenko: looks like it's not only writer checkpointer mutex :)

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -242,139 +315,176 @@ private PagedFile openOrCreate( PageCache pageCache, File indexFile,
...
+        return lastId;
     }
 
-    // Utility method
-    boolean consistencyCheck() throws IOException
+    @Override
+    public void checkpoint( IOLimiter ioLimiter ) throws IOException
     {
-        try ( PageCursor cursor = pagedFile.io( rootId, PagedFile.PF_SHARED_READ_LOCK ) )
+        pagedFile.flushAndForce( ioLimiter );
```
MishaDemianenko, line 479: why we doing this twice? (before (is it safe to do so?) and after the lock)

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
```diff
@@ -0,0 +1,360 @@
...
+import static org.neo4j.io.pagecache.IOLimiter.unlimited;
+import static org.neo4j.io.pagecache.tracing.PageCacheTracer.NULL;
+
+public class GBPTreeRecoveryTest
+{
+    private static final int PAGE_SIZE = 256;
+
+    @Rule
+    public final RandomRule random = new RandomRule();
+    private final File directory = new File( "/somewhere" );
```
MishaDemianenko: please still use test directory to generate directory for your test, to guard from future modification like: oh i think we should use default fs here! push!

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
```diff
@@ -0,0 +1,360 @@
...
+    private final File file = new File( directory, "index" );
+    private EphemeralFileSystemAbstraction fs = new EphemeralFileSystemAbstraction();
```
MishaDemianenko: please use corresponding rule

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointer.java
```diff
@@ -47,6 +47,8 @@
     // unsigned int
```
MishaDemianenko, line 47: Can we rename this class to GenerationalPointer. GenSafe bring a feeling of char greediness when readability is the only victim.

MishaDemianenko (changes requested):
Couple of comments,  packages for public-private api are still not clear

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
tinwelint: Oh, right, a very good point!

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
tinwelint: Then we have to move that rule to neo4j-io or something, no biggie

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointer.java
tinwelint: It's not just about this class, GSP is a core concept/term so it's a much bigger rename if that is to be done. We have an action point for taking a naming approach around gen vs generation and the likes later on.

tinwelint:
API/impl package will be a separate action later on, when there's no conflicting future development in another branch, not in this PR.

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
chrisvest: A better home would be `neo4j-io`.

community/common/src/test/java/org/neo4j/test/rule/RandomRule.java
tinwelint: 👍 will go for `@Seed` only

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: This code has moved as part of free-list work and changed to use `zapPage()` so will ignore in this PR.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: Instead of explaining here I'll add comments to each line in `checkpoint()` method to make that clear.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: `writerTaken` is a slightly different thing than the lock. We want an attempt to acquire multiple concurrent writers to fail, not to block and succeed when other writer as done. This to be certain that assumptions about users of this tree are correct.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: Yeah, being a little too safe here, it's not required here.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
MishaDemianenko: ok

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
MishaDemianenko: sure sounds good

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
MishaDemianenko: ok

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointer.java
MishaDemianenko: sure, it's absolutely out of scope of this pr and if you have a note about that, it covers this comment

community/index/src/main/java/org/neo4j/index/gbptree/TreeState.java
tinwelint: Yup correct, I believe GBPTree#read/writeState() should be changed to do that

community/index/src/main/java/org/neo4j/index/gbptree/TreeStatePair.java
tinwelint: As a private method this becomes more of a matter of taste imho, not quite a "should". But with that said, Optional makes the calling methods a bit more readable and shorter.

community/index/src/main/java/org/neo4j/index/gbptree/TreeStatePair.java
tinwelint: What would be gained by doing that?

community/index/src/main/java/org/neo4j/index/gbptree/ConsistencyChecker.java
tinwelint: The structure as a whole of this class as well as TreePrinter needs to be redesigned to better be able to properly read from page cursor. Although this is a test utility and works for what it does for the moment. They are planned to be redesigned by another PR.

MishaDemianenko (approved):

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -254,23 +217,46 @@ private PagedFile openOrCreate( PageCache pageCache, File indexFile,
...
+        {
+            Pair<TreeState,TreeState> states = TreeStatePair.readStatePages(
+                    cursor, IdSpace.STATE_PAGE_A, IdSpace.STATE_PAGE_B );
+
+            TreeState state = TreeStatePair.selectNewestValidState( states );
+            rootId = state.rootId();
+            lastId = state.lastId();
+            stableGeneration = state.stableGeneration();
+            unstableGeneration = state.unstableGeneration();
+        }
```
chrisvest: The reason you are using the write page lock here is because you don't want to deal with retry right? You are okay with this causing the page to be marked as dirty?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -390,17 +432,28 @@ public long acquireNewId()
...
+        pagedFile.flushAndForce( ioLimiter );
+        writerCheckpointMutex.lock();
+        try
+        {
+            pagedFile.flushAndForce( ioLimiter );
+            stableGeneration = unstableGeneration;
+            unstableGeneration++;
+            writeState( pagedFile );
+            pagedFile.flushAndForce( ioLimiter );
+            monitor.checkpointCompleted();
```
chrisvest, line 501: I think we can allow ourselves to use `IOLimiter.unlimited()` within this critical section, since it ought to complete as quickly as possible, in order to not interfere with our write throughput.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -413,12 +466,35 @@ public void close() throws IOException
...
+            writer.take( rootId, options );
+            success = true;
+            return writer;
+        }
+        finally
+        {
+            if ( !success )
+            {
+                releaseWriter();
+            }
```
chrisvest, line 546: If we somehow don't succeed, then we'll never release the `writerCheckpointMutex`. Is that intentional? I'd block all future writers, and check pointing.

community/index/src/test/java/org/neo4j/index/gbptree/InternalTreeLogicTest.java
```diff
@@ -369,23 +426,25 @@ public void modifierMustRemoveFromLeftChild() throws Exception
...
-        node.initializeLeaf( cursor, STABLE_GENERATION, UNSTABLE_GENERATION );
+        node.initializeLeaf( cursor, stableGen, unstableGen );
         for ( int i = 0; !structurePropagation.hasSplit; i++ )
         {
             insert( i, i );
         }
         long rootId = newRootFromSplit( structurePropagation );
 
-        // when key to remove exists in internal
+        // when and key to remove exists in internal
```
chrisvest: `when and` what?

community/index/src/main/java/org/neo4j/index/Hit.java
```diff
@@ -29,7 +29,15 @@
...
 {
+    /**
+     * @return key of this hit. This returned key instance shouldn't be held on to, rather its contents,
+     * because key instances can be immutable and change within a result set to be overwritten with the next hit.
+     */
     KEY key();
 
+    /**
+     * @return value of this hit. This returned value instance shouldn't be held on to, rather its contents,
+     * because value instances can be immutable and change within a result set to be overwritten with the next hit.
```
chrisvest: These two descriptions are hard to parse. Are KEY and VALUE supposed to be containers of immutable objects, or what exactly is mutating?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -465,19 +465,36 @@ public long acquireNewId() throws IOException
...
+
+        // Block writers, or if there's a current writer then wait for it to complete and then block
+        // From this point and till the lock is released we know that the tree won't change.
         writerCheckpointMutex.lock();
         try
         {
+            // Flush dirty pages since that last flush above. This should be a very small set of pages
+            // and should be rather fast. In here writers are blocked and we want to minimize this
+            // windows of time as much as possible, that's why there's an initial flush outside this lock.
             pagedFile.flushAndForce( ioLimiter );
```
chrisvest: To minimise the amount of time in this critical section, use `IOLimiter.unlimited()`.

chrisvest: Or the `flushAndForce()` method that doesn't take an `IOLimiter`.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -465,19 +465,36 @@ public long acquireNewId() throws IOException
...
+
+            // Increment generation, i.e. stable becomes current unstable and unstable increments by one
+            // and write the tree state (rootId, lastId, generation a.s.o.) to state page.
             long unstableGeneration = unstableGeneration( generation );
             generation = Generation.generation( unstableGeneration, unstableGeneration + 1 );
             writeState( pagedFile );
+
+            // Flush the state page.
             pagedFile.flushAndForce( ioLimiter );
```
chrisvest: Same as above.

community/index/src/test/java/org/neo4j/index/gbptree/ThrowingRunnable.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package org.neo4j.index.gbptree;
+
+interface ThrowingRunnable
+{
+    void run() throws Exception;
+
+    static Runnable throwing( ThrowingRunnable callable )
+    {
+        return new Runnable()
```
chrisvest: Lambda?

chrisvest (commented):
Looks good overall. I like that this was broken out into many small commits. I had a few comments, though.

chrisvest:
@burqen @tinwelint Look through my "hidden" comments as well. GitHub incorrectly thinks they've been addressed.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
tinwelint: I think it'll call `releaseWriter` if it doesn't succeed and `releaseWriter` will unlock the mutex, won't it?

community/index/src/test/java/org/neo4j/index/gbptree/InternalTreeLogicTest.java
tinwelint: Probably a comment moved up from a lower part of the test and just put a when in front. Removing it.

community/index/src/test/java/org/neo4j/index/gbptree/ThrowingRunnable.java
tinwelint: Yes

community/index/src/main/java/org/neo4j/index/Hit.java
tinwelint: Typo: immutable --> mutable

tinwelint:
@chrisvest hopefully I found your hidden comments as well. Fixed the issues you highlighted and commented on the one about `releaseWriter`. Do you agree with that comment?

EDIT: apparently I didn't find them. We're working on them now

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
chrisvest: Correct, I misread the code.

PR #8482, opened 2016-12-06
Added support for little-endian pache cache operations

ragadeeshu:
This is in preparation for little-endian store formats. Bytewise access consolidated to UnsafeUtil.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/gbptree/ByteArrayPageCursor.java
+++ b/community/index/src/main/java/org/neo4j/index/gbptree/ByteArrayPageCursor.java
@@ -100,7 +100,19 @@ class ByteArrayPageCursor extends PageCursor
     }
 
     @Override
+    public long getLongLE()
+    {
+        return buffer.getLong();
+    }
+
+    @Override
     public long getLongBE( int offset )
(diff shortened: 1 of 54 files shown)


burqen:
@ragadeeshu A note on `ByteArrayPageCursor`.
It is used like this:
- Read stuff from a `PageCursor` into a byte array.
- Wrap byte array with `ByteArrayPageCursor`
- Write some stuff into that byte array, among that `putLong`
- Write entire byte array back through `PageCursor`.

Now, it looks like `ByteArrayPageCursor` does not take byte ordering into account in this PR and if that is true we could potentially end up mixing different orderings. Sad.
Looks like [ByteBuffer.order()](https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#order()) could be of some use here.

ps. we want to get rid of `ByteArrayPageCursor` eventually but for now we have are stuck with it.

Hmm... or this may not be a problem because we will explicitly use `put...BE()` always. Either way, I think `ByteArrayPageCursor` should implement correct endianess for coherency.

chrisvest:
This needs a rethink.

PR #8499, opened 2016-12-12
Fix a bug where advancing a page cursor on a unmapped file leaves the…

ragadeeshu:
… currently pinned pages locked when they should not be.

changelog: Fix a bug that on rare occasions could lead to leaking paged memory.


tinwelint:
For future PRs, consider using the title/body type of commit/PR messages instead of one long line.

PR #8509, opened 2016-12-13
Fixed flakey test in UsingIndexAcceptanceTest

eebus:
By making sure that index is online.


eebus:
[[Flaky test] UsingAcceptanceTest.should succeed if executing a query using both 'USING SCAN' and 'USING INDEX' if index exists is flaky on several 3.2 builds](https://trello.com/c/pyCr43Wc/64-flaky-test-usingacceptancetest-should-succeed-if-executing-a-query-using-both-using-scan-and-using-index-if-index-exists-is-flak)

PR #8511, opened 2016-12-13
Close reader that used to count number of entries in the label scan index.

MishaDemianenko:
Close all entries reader that used during index population.
Log error in case if reader can't be closed and fallback to all node scan in that case.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/DynamicIndexStoreView.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/DynamicIndexStoreView.java
@@ -25,6 +25,7 @@ import java.util.function.IntPredicate;
 import java.util.stream.IntStream;
 
 import org.neo4j.helpers.collection.Visitor;
+import org.neo4j.kernel.api.labelscan.AllEntriesLabelScanReader;
 import org.neo4j.kernel.api.labelscan.LabelScanStore;
 import org.neo4j.kernel.api.labelscan.NodeLabelUpdate;
 import org.neo4j.kernel.impl.api.index.NodePropertyUpdates;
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/DynamicIndexStoreView.java
```diff
@@ -71,12 +77,24 @@ public DynamicIndexStoreView( LabelScanStore labelScanStore, LockService locks,
...
     {
-        return ArrayUtils.isEmpty( labelIds ) || isEmptyLabelScanStore() || isNumberOfLabeledNodesExceedThreshold( labelIds );
+        try
+        {
+            return ArrayUtils.isEmpty( labelIds ) || isEmptyLabelScanStore() ||
+                    isNumberOfLabeledNodesExceedThreshold( labelIds );
+        }
+        catch ( Exception e )
+        {
+            log.error( "Can not determine number of labeled nodes, falling back to all nodes scan.", e );
```
burqen, line 87: Would be nice if the message mentioned something about not being able to close the opened `AllEntriesLabelScanReader`.

burqen (changes requested):
Small comment about logged error message.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/DynamicIndexStoreView.java
MishaDemianenko: I was thinking that exception is already enough to figured out what went wrong and error itself is more explanation what we are doing now. Do you still think it's better to mention why it's like that and mention scan reader? @burqen

burqen (approved):

PR #8516, opened 2016-12-14
GB+Tree page reuse

burqen:
**What?**
The GB+Tree need to reuse pages, otherwise it would grow very fast. This PR enables that.

**How?**
Pages that are eligible for reuse, tree-node-page with an older generation or a freelist-node-page that has been used up, will be added to the freelist which is made up of freelist-nodes making up a linked list.

Whenever a new page is needed, the FreelistIdProvider check if there are any pages to reuse and pick one of those if there are, otherwise allocate at end of file.

**Concurrency!**
Reuse of pages comes with a lot of concurrency problems around read, see `SeekCursor`. Most of this PR thus revolves around ensuring SeekCursor behaves correctly. The `GBPTreeIT#shouldReadCorrectlyWhenConcurrentlyInsertingOutOfOrder` test this very aggressively with overlapping insert and read with a super-high checkpoint load.

**Is that all?**
Commit messages of individual commits are very detailed so please have a look at those.


community/index/src/main/java/org/neo4j/index/gbptree/FreelistNode.java
```diff
@@ -0,0 +1,98 @@
...
+
+    void write( PageCursor cursor, long unstableGeneration, long pageId, int pos )
+    {
+        if ( pageId == NO_PAGE_ID )
+        {
+            throw new IllegalArgumentException( "Tried to write pageId " + pageId + " which means null" );
+        }
+        assertPos( pos );
+        GenSafePointer.assertGenerationOnWrite( unstableGeneration );
+        cursor.setOffset( HEADER_LENGTH + pos * ENTRY_SIZE );
```
MishaDemianenko: would be nice if we will extract `HEADER_LENGTH + pos * ENTRY_SIZE` into separate `calculateOffset` method or similar

community/index/src/main/java/org/neo4j/index/gbptree/FreelistNode.java
```diff
@@ -0,0 +1,98 @@
...
+
+import org.neo4j.io.pagecache.PageCursor;
+
+import static org.neo4j.index.gbptree.PageCursorUtil.get6BLong;
+import static org.neo4j.index.gbptree.PageCursorUtil.getUnsignedInt;
+import static org.neo4j.index.gbptree.PageCursorUtil.put6BLong;
+
+class FreelistNode
+{
+    static final int PAGE_ID_SIZE = GenSafePointer.POINTER_SIZE;
```
MishaDemianenko: almost all of those flags can be private

community/index/src/main/java/org/neo4j/index/gbptree/FreelistNode.java
```diff
@@ -0,0 +1,98 @@
...
+            throw new IllegalArgumentException( "Negative pos " + pos );
+        }
+    }
+
+    long read( PageCursor cursor, long stableGeneration, int pos )
+    {
+        assertPos( pos );
+        cursor.setOffset( HEADER_LENGTH + pos * ENTRY_SIZE );
+        long generation = getUnsignedInt( cursor );
+        long result = generation <= stableGeneration ? get6BLong( cursor ) : NO_PAGE_ID;
```
MishaDemianenko: result here is redundant

community/index/src/main/java/org/neo4j/index/gbptree/FreelistNode.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+package org.neo4j.index.gbptree;
+
+import org.neo4j.io.pagecache.PageCursor;
+
+import static org.neo4j.index.gbptree.PageCursorUtil.get6BLong;
+import static org.neo4j.index.gbptree.PageCursorUtil.getUnsignedInt;
+import static org.neo4j.index.gbptree.PageCursorUtil.put6BLong;
+
+class FreelistNode
```
MishaDemianenko, line 40: Needs javadocification

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+
+import org.neo4j.io.pagecache.PageCursor;
+import org.neo4j.io.pagecache.PagedFile;
+
+import static org.neo4j.index.gbptree.PageCursorUtil.checkOutOfBounds;
+import static org.neo4j.index.gbptree.PageCursorUtil.goTo;
+
+class FreeListIdProvider implements IdProvider
+{
+    interface Monitor
```
MishaDemianenko, line 33: shouldn't this be abstract class instead of interface with all default methods?

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+        this.lastId = lastId;
+        this.writePageId = writePageId;
+        this.readPageId = readPageId;
+        this.writePos = writePos;
+        this.readPos = readPos;
+    }
+
+    void initializeAfterCreation() throws IOException
+    {
+        writePageId = readPageId = nextLastId();
```
MishaDemianenko: please use multiple assignments on multiple lines

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+        try ( PageCursor cursor = pagedFile.io( writePageId, PagedFile.PF_SHARED_WRITE_LOCK ) )
+        {
+            goTo( cursor, "free-list", writePageId );
+            freelistNode.initialize( cursor );
+            checkOutOfBounds( cursor );
```
MishaDemianenko, line 131: cursor.shouldRetry loop is missing?

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+    {
+        return ++lastId;
+    }
+
+    @Override
+    public void releaseId( long stableGeneration, long unstableGeneration, long id ) throws IOException
+    {
+        try ( PageCursor cursor = pagedFile.io( writePageId, PagedFile.PF_SHARED_WRITE_LOCK ) )
+        {
+            if ( !cursor.next() )
```
MishaDemianenko: cursor.shouldRetry?

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+            writePos++;
+        }
+
+        if ( writePos >= freelistNode.maxEntries() )
+        {
+            // Current free-list write page is full, allocate a new one.
+            long nextFreelistPage = acquireNewId( stableGeneration, unstableGeneration, false );
+            try ( PageCursor cursor = pagedFile.io( writePageId, PagedFile.PF_SHARED_WRITE_LOCK ) )
+            {
+                if ( !cursor.next() )
```
MishaDemianenko: cursor.shouldRetry?

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+                {
+                    resultPageId = freelistNode.read( cursor, stableGeneration, readPos );
+                }
+                while ( cursor.shouldRetry() );
+
+                if ( resultPageId != FreelistNode.NO_PAGE_ID )
+                {
+                    // FreelistNode compares generation and so this means that we have an available
+                    // id in the free list which we can acquire from a stable generation
+                    if ( ++readPos >= freelistNode.maxEntries() )
```
MishaDemianenko: can we please extract `++` from if statement

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+        return acquiredId;
+    }
+
+    private long acquireNewIdFromFreelistOrEnd( long stableGeneration, long unstableGeneration,
+            boolean allowTakeLastFromPage ) throws IOException
+    {
+        if ( (readPageId != writePageId || readPos < writePos) &&
+                (allowTakeLastFromPage || readPos < freelistNode.maxEntries() - 1) )
+        {
+            // It looks like read pos is < write pos so check if we can grab the next id
```
MishaDemianenko: based on previous if condition this observation is not correct

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+        if ( readPageId == FreelistNode.NO_PAGE_ID )
+        {
+            return;
+        }
+
+        long pageId = readPageId;
+        visitor.accept( pageId );
+        try ( PageCursor cursor = pagedFile.io( pageId, PagedFile.PF_SHARED_READ_LOCK ) )
+        {
+            PageCursorUtil.goTo( cursor, "free-list", readPageId );
```
MishaDemianenko: since currently pageId equal to readPageId we should be able to move PageCursorUtil.goTo inside a while loop without almost any changes. You will have it as first statement in your while loop instead of last one that will make it a bit more logical and easy to follow

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+        if ( readPageId == FreelistNode.NO_PAGE_ID )
+        {
+            return;
+        }
+
+        long pageId = readPageId;
+        int pos = readPos;
+        try ( PageCursor cursor = pagedFile.io( pageId, PagedFile.PF_SHARED_READ_LOCK ) )
+        {
+            PageCursorUtil.goTo( cursor, "free-list", readPageId );
```
MishaDemianenko: same thing with goTo and while loop as above

community/index/src/test/java/org/neo4j/index/gbptree/FreeListIdProviderTest.java
```diff
@@ -0,0 +1,286 @@
...
+
+        // WHEN/THEN
+        freelist.visitFreelistPageIds( id ->
+        {
+            assertTrue( expected.remove( id ) );
+        } );
+        assertTrue( expected.isEmpty() );
+    }
+
+    private void fillPageWithCrapData( long releasedId ) throws IOException
```
MishaDemianenko: i think fillPageWithRandomData is a bit more descriptive name

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointerPair.java
```diff
@@ -134,77 +170,52 @@ public static long read( PageCursor cursor, long stableGeneration, long unstable
...
+    }
+
+    private static long buildSuccessfulReadResult( long slot, int logicalPos, int gsppOffset, long pointer )
+    {
+        boolean isLogicalPos = logicalPos != NO_LOGICAL_POS;
+        long offsetType = isLogicalPos ? FLAG_LOGICAL_POS : FLAG_ABS_OFFSET;
+        long genOffset = isLogicalPos ? logicalPos : gsppOffset;
+        if ( (genOffset & ~MAX_GEN_OFFSET_MASK) != 0 )
+        {
+            throw new IllegalArgumentException( "Illegal genOffset:" + genOffset + ", it would be too large, max is " +
```
MishaDemianenko, line 215: here we throw exception inside of cursor.shouldRetry loop and base our checks on possibly non valid data at all, and we should not do that

community/index/src/main/java/org/neo4j/index/gbptree/ConsistencyChecker.java
```diff
@@ -50,34 +56,147 @@
...
+    /**
+     * Checks so that all pages between {@link IdSpace#MIN_TREE_NODE_ID} and highest allocated id
+     * are either in use in the tree, on the free-list or free-list nodes.
+     *
+     * @param cursor {@link PageCursor} to use for reading.
+     * @return {@code true} if all pages are taken, otherwise {@code false}. Also is compatible with java
+     * assert calls.
+     * @throws IOException on {@link PageCursor} error.
+     */
+    public boolean checkSpace( PageCursor cursor, long lastId, PrimitiveLongIterator freelistIds ) throws IOException
```
MishaDemianenko, line 85: can we extract some pieces of this huge method into separate methods? like: addRightSiblings, findLeftMostSibling etc

community/index/src/main/java/org/neo4j/index/gbptree/ConsistencyChecker.java
```diff
@@ -105,6 +228,34 @@ else if ( node.isLeaf( cursor ) )
...
+        assert nodeGen <= expectedGen : "Expected node:" + cursor.getCurrentPageId() + " gen:" + nodeGen +
+                " to be ≤ pointer gen:" + expectedGen;
+    }
+
+    private void checkNewGenPointerGen( PageCursor cursor ) throws IOException
+    {
+        long newGen = node.newGen( cursor, stableGeneration, unstableGeneration );
+        if ( TreeNode.isNode( newGen ) )
+        {
+            System.err.println( "WARNING: we ended up on an old generation " + cursor.getCurrentPageId() +
```
MishaDemianenko, line 257: why we print only error message here and continue but in all other places with fail with assertion error?

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
```diff
@@ -0,0 +1,351 @@
...
+     * Last allocated page id, used for allocating new ids as more data gets inserted into the tree.
+     */
+    private volatile long lastId;
+
+    /**
+     * For monitoring internal free-list activity.
+     */
+    private final Monitor monitor;
+
+    FreeListIdProvider( PagedFile pagedFile, int pageSize, long lastId, Monitor monitor )
```
MishaDemianenko, line 104: why do we need to pass lastId as one of the constructor params while it will be reassign when `initialize` will be called?

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -219,13 +220,33 @@ default void checkpointCompleted()
     private volatile long generation;
 
     /**
+     * Guard for being able to read rootId and rootGen atomically. Read by looping.
+     *
+     * Odd -> being updated
+     * Even -> stable
+     */
+    private volatile int rootUpdateGuard;
```
MishaDemianenko: this is extremely surprising guard and way of using it. Ideally it should be replaced by one long where root generation and root id combined which updated atomically. Otherwise it should be extracted to separate class that will perform guarding etc and will provide consistent values to the tree.
Visibility of rootId and generation is in question now always and hard to figured out.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -134,6 +135,11 @@ default void checkpointCompleted()
         }
     }
 
+    interface RootCatchup
```
MishaDemianenko: Can this be moved to upper level class, seek cursor using it so it looks like it deserve it's own class.

community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -594,10 +647,20 @@ boolean consistencyCheck() throws IOException
...
+            PageCursorUtil.goTo( cursor, "root", rootId );
+            long unstableGeneration = unstableGeneration( generation );
+            ConsistencyChecker<KEY> consistencyChecker = new ConsistencyChecker<>( bTreeNode, layout,
+                    stableGeneration( generation ), unstableGeneration );
+
+            boolean check = consistencyChecker.check( cursor, rootGen );
+            PageCursorUtil.goTo( cursor, "root", rootId );
+
+            PrimitiveLongSet freelistIds = Primitive.longSet();
+            freeList.visitFreelistPageIds( id -> freelistIds.add( id ) );
```
MishaDemianenko: can be method reference

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
```diff
@@ -197,6 +218,20 @@ void setNewGen( PageCursor cursor, long newGenId, long stableGeneration, long un
         GenSafePointerPair.assertSuccess( result );
     }
 
+    long pointerGen( PageCursor cursor, long readResult )
+    {
+        if ( !GenSafePointerPair.isRead( readResult ) )
+        {
+            throw new IllegalArgumentException( "Expected read result, but got " + readResult );
```
MishaDemianenko, line 220: It's not always safe to thrown exception from here since it can be based on data that is incorrect since we read it from cursor that should be retried

MishaDemianenko (changes requested):
Some classes are getting quite big code wise and in terms of their responsibilities and it's quite hard to follow what's going on there: for example SeekCursor when you need to scroll to see the end of a loop that's unmanageable complexity.
Also shortage of naming (gen insteand of generation) for example really helping

MishaDemianenko:
Most of the errors have something to do with cursor misuse we need to add tests that use Adversarial cursors with high probability of errors to check that we do not have a lot of places where we miss corresponding retry and error checks

MishaDemianenko:
Introduced way of page reuse will insert new child node page in any free page in file and that will brake data locality since that place can be potentially be far from siblings, parent or anything.
Not sure that it's something we should change now, but we should be aware about that problem.

tinwelint:
@MishaDemianenko much obliged, I think generally you're right in most things here... we'll start addressing them!

community/index/src/main/java/org/neo4j/index/gbptree/FreelistNode.java
tinwelint: 👍 going for `entryOffset` to match equivalent method(s) in `TreeNode`

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
tinwelint: I like Monitors as interfaces, this way you have more freedom in how to implement them. Many (most) of monitors in neo4j are interfaces, aren't they?)

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
tinwelint: `initialize` isn't called when creating an index, only consecutive "open"s. `FreeListIdProvider` could be changed to also know which one is the "first" lastId, but currently only GBPTree knows that and passes that knowledge to it, which I don't think is a bad idea.

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
tinwelint: No, because that page is write locked

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
tinwelint: No, because that page is write locked

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
tinwelint: No, because that page is write locked

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointerPair.java
tinwelint: Not quite true. The exception is never triggered on data that has been read, only data passed in to the method and so is an illegal argument to that method.

community/index/src/main/java/org/neo4j/index/gbptree/ConsistencyChecker.java
tinwelint: Hmm, perhaps serving as additional debug information. This could be enforced later in the development process instead. Although I can't recall it ever happening.

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
MishaDemianenko: for sure lets keep it as interface, np

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
MishaDemianenko: ok i see i missed the branch with initializeAfterCreation call, its fine then to keep it like it is now

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
MishaDemianenko: Well that behaviour is not described in javadoc of page cursor then i would say java doc should be updated to mention this

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
MishaDemianenko: yeah page cursor java doc then need to be updated

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
MishaDemianenko: yeah page cursor java doc then need to be updated

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointerPair.java
MishaDemianenko: `logicalPos` is evaluated based on a data that we read from a cursor and can be incorrect right?

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
tinwelint: You're right in reality, but not conceptually, actually. The `isRead` is determined by a bit in the `readResult` which is set by the `GSPP#read` method itself, regardless of what it reads.

With this said I can see that the `GSPP#read` method doesn't control the pointer it puts into the `readResult` and so it may very well be huge (due to invalid read from `PageCursor`) and overflows into that bit too. That should be controlled in that `read` method anyhow. And so I think the action point here is to add that control and then the above check is safe to do.

tinwelint:
Note on data locality, I don't think data locality changes from w/o free-list since there's no control in which order data is inserted/changed. Also the very nature of GB+Tree is that new generations create new pages and leave the old, so tree "sortedness" doesn't match physical "sortedness" anyway.

community/index/src/main/java/org/neo4j/index/gbptree/TreeNode.java
MishaDemianenko: ok agree

community/index/src/main/java/org/neo4j/index/gbptree/FreeListIdProvider.java
tinwelint: PageCursor javadoc to be updated separately.

community/index/src/main/java/org/neo4j/index/gbptree/GenSafePointerPair.java
tinwelint: It's not, it's passed in from outside. EDIT: I'm wrong, there's a usage in SeekCursor, let's fix it

tinwelint: hmm, actually, we _are_ safe here because keyCount check above guards for that, `KeySearch.search` cannot go outside 0 and keyCount. SImply extended the keyCount check for negative numbers as well.

tinwelint:
@MishaDemianenko pushed fixes for your comments

community/index/src/test/java/org/neo4j/index/gbptree/SeekCursorTest.java
```diff
@@ -990,6 +995,144 @@ public void shouldFailChildPointerIfReadFailureNotCausedByCheckpoint() throws Ex
...
+                cursor.next( leftChild );
+                cursor.zapPage();
+                node.initializeLeaf( cursor, stableGen, unstableGen );
+
+                cursor.next( rootId );
+                return new Root( rootId, gen );
+            }
+            catch ( IOException e )
+            {
+                throw new RuntimeException( e );
```
MishaDemianenko, line 1061: its not that important here, but in general for future cases like that its better to use UncheckedIOException

MishaDemianenko (approved):

PR #8517, opened 2016-12-15
Fixed that `STARTS WITH`, `ENDS WITH`, or `CONTAINS` on a non-string should returns null

eebus:
changelog: Attempting to apply `STARTS WITH`, `ENDS WITH`, or `CONTAINS` on a non-string now returns `null`

### DIFF
--- a/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/StartsWithAcceptanceTest.scala
+++ b/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/StartsWithAcceptanceTest.scala
@@ -42,6 +42,36 @@ class StartsWithAcceptanceTest extends ExecutionEngineFunSuite with QueryStatist
     fNode = createLabeledNode("LABEL")
   }
 
+  test("should return null when end with is used on none strings"){
+    val result = executeWithAllPlanners("""
+                                          | CREATE ({name: 1})
+                                          | WITH *
+                                          | MATCH (a)
+                                          | WHERE a.name ENDS WITH 'foo'
+                                          | RETURN a.name""".stripMargin)
+    result.columnAs("a.name").toList should be (List())
+  }
+
+  test("should return null when contains is used on none strings"){
+    val result = executeWithAllPlanners("""
+                                          | CREATE ({name: 1})
+                                          | WITH *
+                                          | MATCH (a)
+                                          | WHERE a.name CONTAINS 'foo'
+                                          | RETURN a.name""".stripMargin)
+    result.columnAs("a.name").toList should be (List())
+  }
+
+  test("should return null when starts with is used on none strings"){
+    val result = executeWithAllPlanners("""
+                                          | CREATE ({name: 1})
+                                          | WITH *
+                                          | MATCH (a)
+                                          | WHERE a.name STARTS WITH 'foo'
+                                          | RETURN a.name""".stripMargin)
+    result.columnAs("a.name").toList should be (List())
+  }
+
   // *** TESTS OF PREFIX SEARCH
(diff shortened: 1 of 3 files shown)


eebus:
[CONTAINS, STARTS WITH, ENDS WITH failing](https://trello.com/c/5HmlqPcl/71-contains-starts-with-ends-with-failing)

community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/StartsWithAcceptanceTest.scala
```diff
@@ -42,6 +42,46 @@ class StartsWithAcceptanceTest extends ExecutionEngineFunSuite with QueryStatist
     fNode = createLabeledNode("LABEL")
   }
 
+  test("should return null when end with is used on none strings"){
```
Mats-SX: `end with` => `ENDS WITH`
`none strings` => `non-strings` (also below)

Mats-SX (commented):
Looks good.

~~I would really prefer to see Cypher acceptance tests in Cucumber form, but I realise we haven't communicated that very well yet. I can add some scenarios in a subsequent commit.~~

Just realised this is `3.0`, not `3.1`. Scenarios would have to be added after merging in the appropriate branches.

community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/StartsWithAcceptanceTest.scala
eebus: Thanks

community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/commands/predicates/Predicate.scala
```diff
@@ -175,7 +175,7 @@ trait StringOperator {
     case (null, _) => None
     case (_, null) => None
     case (l: String, r: String) => Some(compare(l,r))
-    case (l, r) => throw new CypherTypeException(s"Expected two strings, but got $l and $r")
+    case (_, _) => None
```
systay, line 176: Why not have the string-string case, and then a default handler that returns None?

community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/StartsWithAcceptanceTest.scala
```diff
@@ -42,6 +42,46 @@ class StartsWithAcceptanceTest extends ExecutionEngineFunSuite with QueryStatist
     fNode = createLabeledNode("LABEL")
   }
 
+  test("should return null when END WITH is used on non-strings"){
```
systay, line 45: Nice tests, but maybe not a perfect fit for this test class. Rename this class maybe?

community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/commands/predicates/Predicate.scala
craigtaverner: We considered `case _ => None` but the choice of  `case (_, _) => None` was visually more symmetrical with the other cases. It was a choice of visual elegance only. Since the matched object is always a tuple, we viewed the two approaches as functionally identical and felt free to make an aesthetic choice.

fickludd:
@Mats-SX we have a separate card for adding Cypher TCK test from 3.1 onwards - for this fix and other places that are stumbled upon

PR #8521, opened 2016-12-16
Handle interruptions during initialization

ragadeeshu:
This PR contains two things, fixing both the way startup and stopping of Server interacts, as well as allowing the database to repair broken or missing headers in store files when opening the database. The latter can be the case when the database is killed during initialization.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
@@ -130,26 +130,30 @@ public abstract class CommonAbstractStore<RECORD extends AbstractBaseRecord,HEAD
     {
         try
         {
-            checkStorage( createIfNotExists );
-            loadStorage();
+            checkAndLoadStorage( createIfNotExists );
         }
         catch ( Exception e )
         {
-            if ( storeFile != null )
-            {
-                try
-                {
-                    closeStoreFile();
-                }
-                catch ( IOException failureToClose )
-                {
-                    // Not really a suppressed exception, but we still want to throw the real exception, e,
-                    // but perhaps also throw this in there or convenience.
-                    e.addSuppressed( failureToClose );
-                }
-            }
-            throw launderedException( e );
+            closeAndThrow( e );
         }
+    }
+
+    private void closeAndThrow( Exception e )
+    {
+        if ( storeFile != null )
+        {
+            try
+            {
+                closeStoreFile();
+            }
+            catch ( IOException failureToClose )
+            {
+                // Not really a suppressed exception, but we still want to throw the real exception, e,
+                // but perhaps also throw this in there or convenience.
+                e.addSuppressed( failureToClose );
+            }
+        }
+        throw launderedException( e );
     }
 
     /**
@@ -163,24 +167,30 @@ public abstract class CommonAbstractStore<RECORD extends AbstractBaseRecord,HEAD
     }
 
     /**
-     * This method is called by constructors.
+     * This method is called by constructors. Checks the header record and loads the store.
+     * <p>
+     * Note: This method will map the file with the page cache. The store file must not
+     * be accessed directly until it has been unmapped - the store file must only be
+     * accessed through the page cache.
      * @param createIfNotExists If true, creates and initialises the store file if it does not exist already. If false,
      * this method will instead throw an exception in that situation.
      */
-    protected void checkStorage( boolean createIfNotExists )
+    protected void checkAndLoadStorage( boolean createIfNotExists )
     {
         int pageSize = pageCache.pageSize();
-        //noinspection EmptyTryBlock
-        try ( PagedFile ignore = pageCache.map( storageFileName, pageSize, ANY_PAGE_SIZE ) )
+        int filePageSize;
+        try ( PagedFile pagedFile = pageCache.map( storageFileName, pageSize, ANY_PAGE_SIZE ) )
         {
+            extractHeaderRecord( pagedFile );
+            filePageSize = pageCache.pageSize() - pageCache.pageSize() % getRecordSize();
         }
-        catch ( NoSuchFileException e )
+        catch ( NoSuchFileException | StoreNotFoundException e )
         {
             if ( createIfNotExists )
             {
-                try ( PagedFile file = pageCache.map( storageFileName, pageSize, StandardOpenOption.CREATE ) )
+                try
                 {
-                    initialiseNewStoreFile( file );
+                    createStore( pageSize );
                     return;
                 }
                 catch ( IOException e1 )
@@ -188,7 +198,34 @@ public abstract class CommonAbstractStore<RECORD extends AbstractBaseRecord,HEAD
                     e.addSuppressed( e1 );
                 }
             }
+            if ( e instanceof StoreNotFoundException )
+            {
+                throw (StoreNotFoundException) e;
+            }
             throw new StoreNotFoundException( "Store file not found: " + storageFileName, e );
+        }
+        catch ( IOException e )
+        {
+            throw new UnderlyingStorageException( "Unable to open store file: " + storageFileName, e );
+        }
+        loadStorage( filePageSize );
+    }
+
+    private void createStore( int pageSize ) throws IOException
+    {
+        try ( PagedFile file = pageCache.map( storageFileName, pageSize, StandardOpenOption.CREATE ) )
+        {
+            initialiseNewStoreFile( file );
+        }
+        checkAndLoadStorage( false );
+    }
+
+    private void loadStorage( int filePageSize )
+    {
+        try
+        {
+            storeFile = pageCache.map( getStorageFileName(), filePageSize, openOptions );
+            loadIdGenerator();
         }
         catch ( IOException e )
         {
@@ -221,7 +258,7 @@ public abstract class CommonAbstractStore<RECORD extends AbstractBaseRecord,HEAD
         recordSize = determineRecordSize();
 
         File idFileName = new File( storageFileName.getPath() + ".id" );
-        idGeneratorFactory.create( idFileName, getNumberOfReservedLowIds(), true );
+        idGeneratorFactory.create( idFileName, getNumberOfReservedLowIds(), false );
     }
 
     protected void createHeaderRecord( PageCursor cursor ) throws IOException
@@ -232,57 +269,32 @@ public abstract class CommonAbstractStore<RECORD extends AbstractBaseRecord,HEAD
         readHeaderAndInitializeRecordFormat( cursor );
     }
 
-    /**
-     * Should do first validation on store validating stuff like version and id
-     * generator. This method is called by constructors.
-     * <p>
-     * Note: This method will map the file with the page cache. The store file must not
-     * be accessed directly until it has been unmapped - the store file must only be
-     * accessed through the page cache.
-     */
-    protected void loadStorage()
-    {
-        try
-        {
-            extractHeaderRecord();
-            int filePageSize = pageCache.pageSize() - pageCache.pageSize() % getRecordSize();
-            storeFile = pageCache.map( getStorageFileName(), filePageSize, openOptions );
-        }
-        catch ( IOException e )
-        {
-            throw new UnderlyingStorageException( e );
-        }
-        loadIdGenerator();
-    }
-
-    private void extractHeaderRecord() throws IOException
+    private void extractHeaderRecord( PagedFile pagedFile ) throws IOException
     {
         if ( getNumberOfReservedLowIds() > 0 )
         {
-            try ( PagedFile pagedFile = pageCache.map( getStorageFileName(), pageCache.pageSize(), ANY_PAGE_SIZE ) )
+            try ( PageCursor pageCursor = pagedFile.io( 0, PF_SHARED_READ_LOCK ) )
             {
-                try ( PageCursor pageCursor = pagedFile.io( 0, PF_SHARED_READ_LOCK ) )
+
+                if ( pageCursor.next() )
                 {
-                    if ( pageCursor.next() )
+                    do
                     {
-                        do
-                        {
-                            pageCursor.setOffset( 0 );
-                            readHeaderAndInitializeRecordFormat( pageCursor );
-                        }
-                        while ( pageCursor.shouldRetry() );
-                        if ( pageCursor.checkAndClearBoundsFlag() )
-                        {
-                            throw new UnderlyingStorageException(
-                                    "Out of page bounds when reading header; page size too small: " +
-                                    pageCache.pageSize() + " bytes.");
-                        }
+                        pageCursor.setOffset( 0 );
+                        readHeaderAndInitializeRecordFormat( pageCursor );
                     }
-                    else
+                    while ( pageCursor.shouldRetry() );
+                    if ( pageCursor.checkAndClearBoundsFlag() )
                     {
-                        throw new StoreNotFoundException( "Fail to read header record of store file: " +
-                                storageFileName);
+                        throw new UnderlyingStorageException(
+                                "Out of page bounds when reading header; page size too small: " +
+                                pageCache.pageSize() + " bytes." );
                     }
+                }
+                else
+                {
+                    throw new StoreNotFoundException( "Fail to read header record of store file: " +
+                                                      storageFileName );
                 }
             }
         }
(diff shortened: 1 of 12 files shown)


MishaDemianenko:
@ragadeeshu still test failures

community/server/src/main/java/org/neo4j/server/AbstractNeoServer.java
```diff
@@ -194,28 +196,43 @@ public void start() throws ServerStartupException
         init();
         try
         {
-            life.start();
-
-            DiagnosticsManager diagnosticsManager = resolveDependency(DiagnosticsManager.class);
+            serverComponents = new LifecycleAdapter()
```
MishaDemianenko: Please extract it into separate class class and move component creation into init phase then you do not need to remove it on life stop.
Also please check your formatting configuration

community/server/src/main/java/org/neo4j/server/ServerBootstrapper.java
```diff
@@ -72,6 +72,7 @@ public static int start( Bootstrapper boot, String... argv )
     @SafeVarargs
     public final int start( File homeDir, Optional<File> configFile, Pair<String, String>... configOverrides )
     {
+        addShutdownHook();
```
MishaDemianenko, line 75: shutdown hook use `server` field  to access currently running server. And that variable is not volatile, ho do you ensure that you can see valid value in `server`?

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
```diff
@@ -0,0 +1,68 @@
...
+        latch.await();
+
+        server = getNeoServer();
+        server.start();
+        server.stop();
+    }
+
+    protected abstract NeoServer getNeoServer() throws IOException;
+
+    private Runnable $waitThenStopServer( NeoServer server, CountDownLatch latch )
```
MishaDemianenko: please use conventional naming

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
```diff
@@ -0,0 +1,68 @@
...
+    {
+        return () -> {
```
MishaDemianenko: please use standard formatting settings

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
```diff
@@ -0,0 +1,68 @@
...
+            try
+            {
+                //Make sure that we have started starting the database. This value is empirically chosen and may need
+                // to be revisited in the future.
+                Thread.sleep( 200 );
```
MishaDemianenko: that is absolutely useless comment and timeout that in best case was working only on your machine that day while it was in a mood, instead please create custom server that will as part of it stop process fire additional thread that will try to stop it.

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
```diff
@@ -0,0 +1,68 @@
...
+                server.stop();
+                latch.countDown();
+            }
+            catch ( Exception e )
+            {
+                //This is ok, it is not what we are testing for.
```
MishaDemianenko: exceptions may indicate that there is a problem during immediate stop: like can't shutdown storage layer of some sort. Why is it safe to ignore them?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
```diff
@@ -163,37 +167,70 @@ public String getTypeDescriptor()
...
      * @param createIfNotExists If true, creates and initialises the store file if it does not exist already. If false,
      * this method will instead throw an exception in that situation.
      */
-    protected void checkStorage( boolean createIfNotExists )
+    protected void checkAndLoadStorage( boolean createIfNotExists )
     {
         int pageSize = pageCache.pageSize();
-        //noinspection EmptyTryBlock
-        try ( PagedFile ignore = pageCache.map( storageFileName, pageSize, ANY_PAGE_SIZE ) )
+        int filePageSize;
```
MishaDemianenko, line 181: there is path when filePageSize will be 0 it looks like it's not safe in case we will create new store?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
```diff
@@ -163,37 +167,70 @@ public String getTypeDescriptor()
...
+        try ( PagedFile pagedFile = pageCache.map( storageFileName, pageSize, ANY_PAGE_SIZE ) )
         {
+            extractHeaderRecord( pagedFile );
+            filePageSize = pageCache.pageSize() - pageCache.pageSize() % getRecordSize();
         }
-        catch ( NoSuchFileException e )
+        catch ( NoSuchFileException | StoreNotFoundException e )
```
MishaDemianenko, line 187: instead of catching StoreNotFoundException here and do instance of later on i would extract store creation handler instead and would separate catch blocks, but that is not really important and more question of taste

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
```diff
@@ -163,37 +167,70 @@ public String getTypeDescriptor()
...
+        loadStorage( filePageSize );
+    }
+
+    private void createStore( int pageSize ) throws IOException
+    {
+        try ( PagedFile file = pageCache.map( storageFileName, pageSize, StandardOpenOption.CREATE ) )
+        {
+            initialiseNewStoreFile( file );
+        }
+        checkAndLoadStorage( false );
```
MishaDemianenko, line 220: since we just create new store why we do additional round with  checkAndLoadStorage instead of calling loadStorage?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
MishaDemianenko, line 187: is it always safe to assume that we can recreate store in case of exception during header reading? it's should be safe to do in case of empty store file no doubts, what if we have store file that for with somehow broken header? what do you guys think about additional check that will ensure that we do not delete file that already have something except header?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/NeoStores.java
```diff
@@ -185,11 +186,15 @@ private void verifyRecordFormat()
         try
         {
             String expectedStoreVersion = recordFormats.storeVersion();
-            String actualStoreVersion = versionLongToString( getRecord( pageCache, neoStoreFileName, STORE_VERSION ) );
-            RecordFormats actualStoreFormat = RecordFormatSelector.selectForVersion( actualStoreVersion );
-            if ( !isCompatibleFormats( actualStoreFormat ) )
+            long record = getRecord( pageCache, neoStoreFileName, STORE_VERSION );
+            if ( record != MetaDataRecordFormat.FIELD_NOT_PRESENT )
```
MishaDemianenko, line 190: when it's the case that we do not have STORE_VERSION in metadata store? and why it's safe to skip format check in that case?

community/kernel/src/test/java/org/neo4j/kernel/impl/store/StoreFactoryTest.java
```diff
@@ -25,10 +25,16 @@
 import org.junit.Test;
 
 import java.io.File;
+import java.io.FileOutputStream;
```
MishaDemianenko: plenty of useless imports

community/kernel/src/test/java/org/neo4j/kernel/impl/store/StoreFactoryTest.java
```diff
@@ -152,4 +159,27 @@ public void shouldDelegateDeletionOptionToStores() throws Exception
...
+        storeFactory.openAllNeoStores( true ).close();
+    }
+
+    @Test
+    public void shouldCompleteInitializationOfStoresWithIncompleteHeaders() throws Exception
+    {
+        StoreFactory storeFactory = storeFactory( Config.empty() );
+        storeFactory.openAllNeoStores( true ).close();
+        FileSystemAbstraction fs = fsRule.get();
+        File[] files = fs.listFiles( storeDir, ( file, s ) -> !s.endsWith( ".id" ) );
```
MishaDemianenko: why do we need `files`?

MishaDemianenko (changes requested):
Please find couple of comments regarding proposed changed.
Also please provide description for your PR

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
ragadeeshu: This follows a naming convention used in e.g. org.neo4j.io.pagecache.PageCacheTestSupport

MishaDemianenko: please do not follow that

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
ragadeeshu: As the comment states, this is not what we are testing for. We just want to test that `shouldBeAbleToRestartWhenStoppedDuringStartup`, will change test name to clarify.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
ragadeeshu: filePageSize is only used in the case where we do not create a new store, and that only happens when we successfully compute it. When we create a new store we use the page size from the page cache.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
ragadeeshu: It is not "recreating the store" as such, as store initialization only does the following:
1. maps the file with a write cursor
2. writes the header record
This means that if there is a file present, but with a broken header record, we will repair the header record, but leave the rest of the file intact. Nothing is deleted.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
ragadeeshu: In order to read the header record, calculate the `filePageSize` and properly handle io errors.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/NeoStores.java
ragadeeshu: For the same reason that it is safe to skip it if the file does not exists.
In this case the file would have been created, but the database was shutdown before it was populated.

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
MishaDemianenko: If you can't stop server - meaning you throw exception during stop all your test is useless since you do no testing restart at all. Aslo since it's separate thread junit will gonna miss exception.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
MishaDemianenko: agree, missed return statement after store creation

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
MishaDemianenko: My question was about different code path. 
What you describing is happy path and my question was about case when we will get any kind of exception and fall back to exception handling that will initialise new store file except the one that present.
Is it safe to do for cases when the file size will be more then header only?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
MishaDemianenko: ok

community/kernel/src/main/java/org/neo4j/kernel/impl/store/NeoStores.java
MishaDemianenko: ok

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
ragadeeshu: The problem won't be missed, since the latch will not be touched and the test will deadlock. Maybe the comment should be changed, and maybe a timeout should be added to the latch.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
ragadeeshu: My point was exactly that, when we "initialize new store file" we only touch the header record, and leave the rest of the file intact. There is no code that could delete or truncate a file. It is safe to do because it will never affect anything other than the header record.

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
MishaDemianenko: ok, please timeout then to that test

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
MishaDemianenko: ah oke

community/server/src/main/java/org/neo4j/server/ServerBootstrapper.java
ragadeeshu: Fixed.

MishaDemianenko:
@ragadeeshu are you done with fixing comments round?

ragadeeshu:
@MishaDemianenko Sorry, yes, I am.

community/server/src/main/java/org/neo4j/server/AbstractNeoServer.java
```diff
@@ -549,4 +536,44 @@ public DependencyResolver get()
             return db.getGraph().getDependencyResolver();
         }
     } );
+
+    private class ServerComponentsLifecycleAdapter extends LifecycleAdapter
+    {
+        @Override
+        public void init() throws Throwable
```
MishaDemianenko: We don't need empty to override `init()` here right?

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.neo4j.io.pagecache.impl.SingleFilePageSwapperFactory;
+import org.neo4j.test.ThreadTestUtils;
+
+import static org.junit.Assert.fail;
+
+public abstract class NeoServerRestartTest
+{
+
+    public static final String CUSTOM_SWAPPER = "CustomSwapper";
+    private static volatile Semaphore semaphore;
```
MishaDemianenko: why this semaphore needs to be volatile?

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    protected abstract NeoServer getNeoServer( String customPageSwapperName ) throws IOException;
+
+    private Runnable stopServerAfterStartingHasStarted( NeoServer server, CountDownLatch latch )
+    {
+        return () ->
+        {
+            try
+            {
+                semaphore.acquire();
```
MishaDemianenko: I'm not quite sure what you are trying to achieve now with all of that. With current flow of things you will only make sure that you will acquire semaphore after the fact that at least one page swapper was created then some time after you will try to stop server that will call synchronised life stop. 
So what we a checking here?
Also if that check is valid please add some description of what and how you achieving with custom page swapper

community/server/src/test/java/org/neo4j/server/NeoServerRestartTest.java
```diff
@@ -0,0 +1,103 @@
...
+                server.stop();
```
MishaDemianenko: if your stop will throw any exception you will not countDown your latch.
Empty catch block claims that any kind of exception is fine - in that case we should not fail test on latch awaiting timeout.

MishaDemianenko:
@ragadeeshu some comments still

MishaDemianenko:
@ragadeeshu please do not amend commit especially when you fixing comments since its really hard to figured out set of changes that you need to verify and requires full verification again

MishaDemianenko (approved):

PR #8525, opened 2016-12-19
More ruthless GB+Tree recovery testing

tinwelint:
Also includes a fix for tree state comparison exposed by the added testing

### DIFF
--- a/community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
+++ b/community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
@@ -27,7 +27,6 @@ import org.junit.rules.RuleChain;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
```diff
@@ -266,11 +317,12 @@ else if ( randomized <= 0.95 )
         }
         else
         {
-            // checkpoint
-            return index -> index.checkpoint( unlimited() );
+            return CHECKPOINT;
         }
     }
 
+    private static final Action CHECKPOINT = index -> index.checkpoint( unlimited() );
```
MishaDemianenko: please move the field to the top of the class

MishaDemianenko:
@tinwelint one small comment, as soon that fixed should be ready to merge

tinwelint:
@MishaDemianenko fixed, oh sorry, didn't think so amended the first commit with it.

PR #8527, opened 2016-12-19
Include CPU active time, lock wait time, and query status in ListQueries

thobe:
Sample output from `CALL dbms.listQueries`:

```
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| queryId   | username | query                         | parameters | startTime                       | elapsedTime    | connectionDetails   | cpuTimeMillis | status                                                                                 | waitTimeMillis | metaData |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| "query-1" | ""       | "MATCH (n) SET n.v = n.v + 1" | {}         | "2017-01-03T16:38:29.213+01:00" | "00:00:01.507" | "embedded-session " | 856           | {waitTimeMillis -> 57, state -> "WAITING", resourceType -> "NODE", resourceIds -> [0]} | 57             | {}       |
| "query-4" | ""       | "CALL dbms.listQueries"       | {}         | "2017-01-03T16:38:30.719+01:00" | "00:00:00.001" | "embedded-session " | 1             | {state -> "RUNNING"}                                                                   | 0              | {}       |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
@@ -22,6 +22,7 @@ package org.neo4j.kernel.api;
 import java.util.Map;
 
 import org.neo4j.kernel.impl.query.QuerySource;
+import org.neo4j.time.CpuClock;
 
 import static java.lang.String.format;
 
@@ -37,6 +38,9 @@ public class ExecutingQuery
     private final String queryText;
     private final Map<String, Object> queryParameters;
     private final long startTime;
+    private final Thread threadExecutingTheQuery;
+    private final CpuClock cpuClock;
+    private final long cpuTimeWhenQueryStarted;
     private Map<String,Object> metaData;
 
     public ExecutingQuery(
@@ -46,7 +50,9 @@ public class ExecutingQuery
             String queryText,
             Map<String,Object> queryParameters,
             long startTime,
-            Map<String,Object> metaData
+            Map<String,Object> metaData,
+            Thread threadExecutingTheQuery,
+            CpuClock cpuClock
     ) {
         this.queryId = queryId;
         this.querySource = querySource;
@@ -55,6 +61,9 @@ public class ExecutingQuery
         this.queryParameters = queryParameters;
         this.startTime = startTime;
         this.metaData = metaData;
+        this.threadExecutingTheQuery = threadExecutingTheQuery;
+        this.cpuClock = cpuClock;
+        this.cpuTimeWhenQueryStarted = cpuClock.cpuTime( threadExecutingTheQuery );
     }
 
     @Override
@@ -109,6 +118,14 @@ public class ExecutingQuery
     public long startTime()
     {
         return startTime;
+    }
+
+    /**
+     * @return the CPU time used by the query, in nanoseconds.
+     */
+    public long cpuTime()
+    {
+        return cpuClock.cpuTime( threadExecutingTheQuery ) - cpuTimeWhenQueryStarted;
     }
 
     @Override
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
@@ -77,7 +77,7 @@ public class ExecutingQuery
         this.metaData = metaData;
         this.threadExecutingTheQuery = threadExecutingTheQuery;
         this.cpuClock = cpuClock;
-        this.cpuTimeNanosWhenQueryStarted = cpuClock.cpuTime( threadExecutingTheQuery );
+        this.cpuTimeNanosWhenQueryStarted = cpuClock.cpuTimeNanos( threadExecutingTheQuery );
     }
 
     @Override
@@ -144,7 +144,7 @@ public class ExecutingQuery
      */
     public long cpuTime()
     {
-        return cpuClock.cpuTime( threadExecutingTheQuery ) - cpuTimeNanosWhenQueryStarted;
+        return cpuClock.cpuTimeNanos( threadExecutingTheQuery ) - cpuTimeNanosWhenQueryStarted;
     }
 
     public long waitTime()
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
@@ -143,9 +143,9 @@ public class ExecutingQuery
     /**
      * @return the CPU time used by the query, in nanoseconds.
      */
-    public long cpuTimeMicros()
+    public long cpuTimeMillis()
     {
-        return NANOSECONDS.toMicros( cpuClock.cpuTimeNanos( threadExecutingTheQuery ) - cpuTimeNanosWhenQueryStarted );
+        return NANOSECONDS.toMillis( cpuClock.cpuTimeNanos( threadExecutingTheQuery ) - cpuTimeNanosWhenQueryStarted );
     }
 
     public long waitTimeMillis()
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
@@ -54,6 +54,8 @@ public class ExecutingQuery
     private final long cpuTimeNanosWhenQueryStarted;
     private Map<String,Object> metaData;
     private volatile ExecutingQueryStatus status = ExecutingQueryStatus.RUNNING;
+    /** Updated through {@link #WAIT_TIME} */
+    @SuppressWarnings( "unused" )
     private volatile long waitTimeNanos;
 
     public ExecutingQuery(
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
@@ -20,7 +20,6 @@
 package org.neo4j.kernel.api;
 
 import java.util.Map;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 
 import org.apache.commons.lang3.builder.ToStringBuilder;
@@ -31,6 +30,7 @@ import org.neo4j.storageengine.api.lock.ResourceType;
 import org.neo4j.time.CpuClock;
 import org.neo4j.time.SystemNanoClock;
 
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.concurrent.atomic.AtomicLongFieldUpdater.newUpdater;
 
 /**
@@ -42,7 +42,6 @@ public class ExecutingQuery
             newUpdater( ExecutingQuery.class, "waitTimeNanos" );
     private final long queryId;
     private final Locks.Tracer lockTracer = ExecutingQuery.this::waitForLock;
-
     private final String username;
     private final QuerySource querySource;
     private final String queryText;
@@ -97,7 +96,6 @@ public class ExecutingQuery
         ExecutingQuery that = (ExecutingQuery) o;
 
         return queryId == that.queryId;
-
     }
 
     @Override
@@ -136,7 +134,7 @@ public class ExecutingQuery
         return startTime;
     }
 
-    public long elapsedTime()
+    public long elapsedTimeMillis()
     {
         return clock.millis() - startTime;
     }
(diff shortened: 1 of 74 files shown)


chrisvest:
What is it a sample output of, and can it include units?

praveenag:
@chrisvest This is the sample output of `dbms.listQueries` procedure. The units are in milliseconds. We chose not to explicitly call out units in the output to gather feedback.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
```diff
@@ -20,41 +20,64 @@
...
     private final Map<String, Object> queryParameters;
-    private final long startTime;
+    private final long startTime; // timestamp in milliseconds
+    private final Thread threadExecutingTheQuery;
+    private final SystemNanoClock clock;
+    private final CpuClock cpuClock;
+    private final long cpuTimeNanosWhenQueryStarted;
     private Map<String,Object> metaData;
+    private volatile ExecutingQueryStatus status = ExecutingQueryStatus.RUNNING;
+    private volatile long waitTimeNanos;
```
chrisvest: It's a small nit, but I would prefer if the new instance fields get assigned in the constructor, so that's all kept together.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
```diff
@@ -111,17 +134,54 @@ public long startTime()
...
+    private Locks.WaitEvent waitForLock( ResourceType resourceType, long[] resourceIds )
+    {
+        ExecutingQueryStatus previous = status;
+        long startTimeNanos = clock.nanos();
+        status = new ExecutingQueryStatus.WaitingOnLock( resourceType, resourceIds, startTimeNanos );
+        return () ->
+        {
+            WAIT_TIME.addAndGet( ExecutingQuery.this, clock.nanos() - startTimeNanos );
+            status = previous;
+        };
```
chrisvest: I'm slightly irked that we allocate for the new status, and the lambda. The allocation for the status might be difficult to avoid, but it ought to be possible to avoid allocating the lambda, I think.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQueryStatus.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import org.neo4j.storageengine.api.lock.ResourceType;
+import org.neo4j.time.SystemNanoClock;
+
+abstract class ExecutingQueryStatus
```
chrisvest: Please add a few lines of javadocs to this, describing what it's used for.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQueryStatus.java
```diff
@@ -0,0 +1,82 @@
...
+        @Override
+        long waitTimeNanos( SystemNanoClock clock )
+        {
+            return 0;
+        }
+
+        @Override
+        Map<String,Object> toMap( SystemNanoClock clock )
+        {
+            return Collections.singletonMap( "state", "RUNNING" );
```
chrisvest: I'm not sure how often this is called, but in principle it could be a constant as well.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ExecutingQueryList.java
```diff
@@ -19,21 +19,51 @@
...
+
+        @Override
+        <T> T reduce( T defaultValue, Function<ExecutingQuery,T> accessor, BiFunction<T,T,T> combinator )
+        {
+            return defaultValue;
+        }
+    };
 
-    class Entry implements ExecutingQueryList
+    abstract <T> T reduce( T defaultValue, Function<ExecutingQuery,T> accessor, BiFunction<T,T,T> combinator );
```
chrisvest: The `reduce` method could be moved above the `EMPTY` declaration.

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/CombinedTracer.java
```diff
@@ -0,0 +1,89 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.locking;
+
+import java.util.Arrays;
+
+import org.neo4j.storageengine.api.lock.ResourceType;
+
+final class CombinedTracer implements Locks.Tracer
```
chrisvest: I would like to see some javadocs on this class as well.

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/CombinedTracer.java
```diff
@@ -0,0 +1,89 @@
...
+            for ( int i = 0; i < those.length; i++ )
+            {
+                tracers[this.tracers.length + i] = those[i];
+            }
+        }
+        else
+        {
+            tracers = Arrays.copyOf( this.tracers, this.tracers.length + 1 );
+            tracers[this.tracers.length] = tracer;
+        }
```
chrisvest, line 71: How big of an `n` do we expect to see here?

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
```diff
@@ -126,4 +126,43 @@ void visit( ResourceType resourceType, long resourceId, String description, long
     void accept(Visitor visitor);
 
     void close();
+
+    interface Tracer
```
chrisvest: I would prefer to call it `LockTracer`. I don't like how the monitors system has caused a proliferation of inner interfaces all called `Monitor`.

enterprise/kernel/src/test/java/org/neo4j/kernel/enterprise/builtinprocs/ListQueriesProcedureTest.java
```diff
@@ -0,0 +1,158 @@
...
+        Map<String,Object> row = result.next();
+        assertFalse( result.hasNext() );
+        //        Object queryId = row1.get( "queryId" );
+        //        Object username = row1.get( "username" );
+        //        Object query = row1.get( "query" );
+        //        Object parameters = row1.get( "parameters" );
+        //        Object startTime = row1.get( "startTime" );
+        //        Object elapsedTime = row1.get( "elapsedTime" );
+        //        Object connectionDetails = row1.get( "connectionDetails" );
+        //        Object metaData = row1.get( "metaData" );
```
chrisvest: What is this commented out stuff?

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
chrisvest: Can we also make it a top-level interface? That would make it shorter to reference in other places in the PR.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
```diff
@@ -20,41 +20,64 @@
...
 
 /**
  * Represents a currently running query.
  */
 public class ExecutingQuery
 {
+    private static final AtomicLongFieldUpdater<ExecutingQuery> WAIT_TIME =
+            newUpdater( ExecutingQuery.class, "waitTimeNanos" );
     private final long queryId;
+    private final Locks.Tracer lockTracer = ExecutingQuery.this::waitForLock;
```
chrisvest: Tracer implementations are meant to be changed via plug-ins, and the concrete implementation decided by `Tracers`. Hard coding the implementation here goes against this. For this reason, tracers should also not be relied upon for correctness; which is another mistake we've made with monitors. I think you should allow `Tracers` to decide what implementation to use, and then you can move the clock state into the default implementation. I would not be against adding a method to the interface to query the accumulated wait time, as long as you can accept that an alternative implementation might not track it and instead always report zero.

chrisvest (changes requested):
The overall idea is good, but please take a look at my comments. There are a couple of things I would like to see changed.

chrisvest:
As for the units, we could either add them as part of the column names in the `listQueries` output, or we could add columns dedicated to the units. Both of these solutions would make it easier for people to see what units are being reported, without compromising the machine readability. Adding the unit at the end of the value would require that the values be returned as strings which, while looking very nice to the human eye, would make it harder for machines to interpret the output.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
```diff
@@ -273,30 +288,46 @@ public void acquireExclusive( ResourceType resourceType, long... resourceIds ) t
...
-                long waitStartMillis = clock.millis();
-                while ( (existingLock = lockMap.putIfAbsent( resourceId, myExclusiveLock )) != null )
-                {
-                    assertValid( waitStartMillis, resourceType, resourceId );
 
-                    // If this is a shared lock:
-                    // Given a grace period of tries (to try and not starve readers), grab an update lock and wait
-                    // for it to convert to an exclusive lock.
-                    if ( tries > 50 && existingLock instanceof SharedLock )
+                Locks.WaitEvent waitEvent = null;
```
chrisvest: There are two empty lines in succession here. They are causing the build to fail on a checkstyle error.

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
thobe: I always think of the names of nested types as being prefixed with the name of the outer type, making the name here `Locks.Tracer`, but when I wrote it I fully anticipated it to be moved to the top level and renamed `LocksTracer` at some point. The main thing that stopped me is the package structure mess.

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/CombinedTracer.java
thobe: if by `n` you mean the length of the `tracers` array, I'd anticipate the size to be(come) `3` if this code ever executes, typically I'd expect the size to be `2` in instances of `CombinedTracer`, do you think we should make it tree structured instead of flat? (since that would optimize for the expected size of 2)

chrisvest:
@thobe Now also needs a rebase and updated license headers.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQueryStatus.java
thobe: It is called when the status is inspected, so not on the hot path, but yes, it could be made constant.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
thobe: Do you have a concrete suggestion for how? Making the status object also implement `Locks.WaitEvent` feels like conflating concerns, but that is the only way I can think of reducing the number of allocations here.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
thobe: for volatile fields that need a default value, I actually prefer it in the field declaration if possible, since it signals the non-null default nature of the field more clearly.

community/kernel/src/main/java/org/neo4j/kernel/api/ExecutingQuery.java
thobe: I do not fully agree.

I agree that it should be possible to have a tracer provided through configuration. But I don't think that excludes being able to use the event based tracer pattern for other purposes as well. The important part of that is just that if one is needed for some feature, it shouldn't disable other features. This is one reason why we included the `CombinedTracer` so that the configured system-tracer can be combined with the query specific one(s).

thobe: The place where a system-configured `Locks.Tracer` should be injected is in `KernelStatement` in the `lockTracer()` method, instead of the hard-coded `Locks.Tracer.NONE` there.

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/CombinedTracer.java
chrisvest: For such small N I don't have a strong opinion. An inlined tree would be one less dependent load. I'm not sure there's enough (or any) measurable difference in performance to justify this, but it may be a cleaner design. I'm fine either way.

chrisvest (approved):

PR #8532, opened 2016-12-20
Kernel transaction state handling update

MishaDemianenko:
Introduce transaction status that encapsulate all the possible transaction state and transitions around them.
Give much more clear understanding in what state transaction is now and what are possible transitions.

Now transaction can be in one of the states: OPEN, CLOSING, CLOSED and SHUTDOWN.
Also transaction can be terminated while it's OPEN or CLOSING.
Typical transaction lifecycle is: OPEN (transaction initialized and user work with it) -> CLOSING (transaction close in progress) -> CLOSED (transaction closed).

SHUTDOWN status used only in case when we shutdown database and transaction can be marked as SHUTDOWN only in case if it's OPEN.
SHUTDOWN transactions can't be closed/committed.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
@@ -26,8 +26,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
 import org.neo4j.collection.pool.Pool;
@@ -67,6 +66,7 @@ import org.neo4j.storageengine.api.StoreReadLayer;
 import org.neo4j.storageengine.api.txstate.TxStateVisitor;
 
 import static org.neo4j.storageengine.api.TransactionApplicationMode.INTERNAL;
+
 
 /**
  * This class should replace the {@link org.neo4j.kernel.api.KernelTransaction} interface, and take its name, as soon
@@ -116,11 +116,9 @@ public class KernelTransactionImplementation implements KernelTransaction, TxSta
     private SecurityContext securityContext;
     private volatile StatementLocks statementLocks;
     private boolean beforeHookInvoked;
-    private volatile boolean closing;
-    private volatile boolean closed;
-    private boolean failure;
+    private TransactionStatus transactionStatus = new TransactionStatus();
+    private volatile boolean failure;
     private boolean success;
-    private volatile Status terminationReason;
     private long startTimeMillis;
     private long timeoutMillis;
     private long lastTransactionIdWhenStarted;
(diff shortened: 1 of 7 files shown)


tinwelint:
Not sure about the tools compilation issue, perhaps a rebase would help, although it probably doesn't matter since this PR doesn't touch those things

MishaDemianenko:
@tinwelint that was because PR was based on 3.2 version with broken ha robustness, rebased and now it should be fine

tinwelint:
Failed on unrelated flaky test, it's fine

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -473,9 +436,22 @@ public long closeTransaction() throws TransactionFailureException
...
+    private void markTransactionAsClosing() throws TransactionFailureException
     {
-        return closing;
+        if ( !transactionStatus.closing() )
+        {
+            assertTransactionOpen();
+            if ( transactionStatus.isShutdown() )
+            {
+                throw new TransactionFailureException( Status.General.DatabaseUnavailable,
+                        "Transaction terminated since database is shutting down." );
```
tinwelint: I wonder if this message is correct. A transaction is shut down in KTI#dispose(), which may be called for various reasons, one of them shutting down the database.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -473,9 +436,22 @@ public long closeTransaction() throws TransactionFailureException
...
+            }
+            else
+            {
+                throw new TransactionFailureException( Status.Transaction.TransactionTerminated,
+                        "Transaction is already closing. Repeated execution of transactions are not allowed." );
```
tinwelint: I wonder if this message is correct. Above we've checked `transactionStatus.closing()`, which at this point must be false, so this transaction isn't in the process of closing.

Also status is set as `TransactionTerminated`, can we be sure of that?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -833,4 +802,123 @@ TransactionWriteState upgradeToSchemaWrites() throws InvalidTransactionTypeKerne
...
+                {
+                    return false;
+                }
+                terminationReason = reason;
+            }
+            while ( !statusUpdater.compareAndSet( this, currentStatus, currentStatus | TERMINATED ) );
+            return true;
+        }
+
+        public boolean closing()
```
tinwelint, line 869: Looks like `closing()` and `shutdown()` have a lot in common, everything except some parameters, no

tinwelint (changes requested):
Just a few comments. I like the nature of TransactionStatus and that it gets rid of many volatile variables and a lock!

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
MishaDemianenko: message updated

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
MishaDemianenko: restore old IllegalStateException for this case

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
MishaDemianenko: yep, method extracted

MishaDemianenko:
@tinwelint updated

PR #8536, opened 2016-12-21
Hardens GB+Tree testing

tinwelint:
By using adversarial page cache and an added access-checking page cache.

Also contains fixes for exposed issues for the added testing.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
+++ b/community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
@@ -126,15 +126,9 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
     private int pos;
 
     /**
-     * Number of keys in the current leaf, this value is cached and only re-read every time there's
-     * a {@link PageCursor#shouldRetry() retry due to concurrent write}.
-     */
-    private int keyCount;
-
-    /**
      * Set if the position of the last returned key need to be found again.
      */
-    private boolean rediscoverKeyPosition;
+    private boolean needToRediscoverKeyPosition;
 
     /**
      * {@link TreeNode#gen(PageCursor) generation} of the current leaf node, read every call to {@link #next()}.
@@ -186,19 +180,22 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
 
     private void traverseDownToFirstLeaf() throws IOException
     {
-        // some variables below are initialized to satisfy the compiler
-        byte nodeType;
-        long newGen = -1;
         boolean isInternal = false;
         int keyCount = -1;
-        long childId = -1;
         int pos = -1;
-        long newGenGen = -1;
-        long childIdGen = -1;
         do
         {
+            // some variables below are initialized to satisfy the compiler
+            byte nodeType;
+            long newGen = -1;
+            long childId = -1;
+            long newGenGen = -1;
+            long childIdGen = -1;
+            int searchResult = -1;
+            boolean expectingShouldRetry;
             do
             {
+                expectingShouldRetry = false;
                 nodeType = TreeNode.nodeType( cursor );
                 if ( nodeType != TreeNode.NODE_TYPE_TREE_NODE )
                 {
@@ -217,16 +214,22 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
                 isInternal = bTreeNode.isInternal( cursor );
                 // Find the left-most key within from-range
                 keyCount = bTreeNode.keyCount( cursor );
-                if ( !keyCountIsSane() )
+                if ( !keyCountIsSane( keyCount ) )
                 {
+                    expectingShouldRetry = true;
                     continue;
                 }
 
-                int search = KeySearch.search( cursor, bTreeNode, fromInclusive, mutableKey, keyCount );
-                pos = KeySearch.positionOf( search );
+                searchResult = KeySearch.search( cursor, bTreeNode, fromInclusive, mutableKey, keyCount );
+                if ( !KeySearch.isSuccess( searchResult ) )
+                {
+                    expectingShouldRetry = true;
+                    continue;
+                }
+                pos = KeySearch.positionOf( searchResult );
 
                 // Assuming unique keys
-                if ( isInternal && KeySearch.isHit( search ) )
+                if ( isInternal && KeySearch.isHit( searchResult ) )
                 {
                     pos++;
                 }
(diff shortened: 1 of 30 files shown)


community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -362,8 +384,17 @@ public boolean next() throws IOException
                     bTreeNode.valueAt( cursor, mutableValue, pos );
                 }
             }
-            while ( rediscoverKeyPosition = cursor.shouldRetry() );
+            while ( needToRediscoverKeyPosition = cursor.shouldRetry() );
             checkOutOfBounds( cursor );
+            if ( expectingShouldRetry )
+            {
+                throw new TreeInconsistencyException( "Read inconsistent tree node %d%n" +
```
ragadeeshu, line 391: This seems very similar to the other loop. Any way to consolidate, at least the exception?

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
```diff
@@ -59,17 +54,20 @@
 import static org.junit.Assert.fail;
 import static org.neo4j.index.IndexWriter.Options.DEFAULTS;
 import static org.neo4j.index.gbptree.GBPTree.NO_MONITOR;
-import static org.neo4j.io.pagecache.tracing.PageCacheTracer.NULL;
+import static org.neo4j.index.gbptree.PageCacheTestUtil.pageCache;
```
ragadeeshu: Would be nice to in the future use the page cache rule here, it would automatically give you the adversarial page cache.

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
```diff
@@ -341,9 +338,10 @@ else if ( randomized <= 0.95 || !allowCheckPoint )
 
     private PageCache createPageCache()
```
ragadeeshu, line 339: See above comment about page cache rule.
There are a lot of similar places in this pr.

community/io/src/test/java/org/neo4j/adversaries/pagecache/AdversarialWritePageCursor.java
```diff
@@ -264,7 +264,11 @@ public boolean shouldRetry() throws IOException
     public int copyTo( int sourceOffset, PageCursor targetCursor, int targetOffset, int lengthInBytes )
     {
         adversary.injectFailure( IndexOutOfBoundsException.class );
-        return delegate.copyTo( sourceOffset, targetCursor, targetOffset, lengthInBytes );
+
+        PageCursor targetCursorDelegate = targetCursor instanceof AdversarialWritePageCursor ?
```
ragadeeshu, line 268: This feels like the wrong place to fix this. The AdversarialWritePageCursor should in some way be a WritePageCursor, and the copyTo should check that the target cursor is a WPC (or that WPC.is assignable from target cursor), and not specifically a Muninn WPC.

community/index/src/test/java/org/neo4j/index/gbptree/PageCacheTestUtil.java
```diff
@@ -95,6 +103,10 @@ PageCache build()
             {
                 pageCache = new AdversarialPageCache( pageCache, adversary );
             }
+            if ( checkAccess )
+            {
+                pageCache = new AccessCheckingPageCache( pageCache );
```
ragadeeshu: If the page cache rule is moved to the IO package, this might belong there?

ragadeeshu (commented):
Some minor comments and concerns. As for the fix for the instanceof check, I feel that this might be the wrong way to fix it. Maybe a concept of a more generic `WritePageCursor` should be introduced.
The `PageCacheRule` etc. needs to be moved to the io package in a seperate PR. This would allow for a lot of cleanup in the index package, as has been made apparent by this PR.
I also recommend to go over your usage of page cursors, and see if there are more cases where the `shouldRetry` loops could and should be moved into the method itself.

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
tinwelint: The PageCacheRule lives in kernel and is somewhat tied to other things in the kernel so we opted for not moving it. Things from it moved to neo4j-io tho, so we're better off than before. The page cache management in these tests I think is minimal and manageable.

community/io/src/test/java/org/neo4j/adversaries/pagecache/AdversarialWritePageCursor.java
tinwelint: I believe this was the recommended fix after discussing briefly with @chrisvest although I do see that there's another PR now which introduces a method that can be used instead of doing the instanceof check inside MuninnWritePageCursor itself. I'm thinking of https://github.com/neo4j/neo4j/pull/8540

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeRecoveryTest.java
tinwelint: Moving the PageCacheRule is a non-trivial operation and I don't see an immediate need for doing that, even if it would be nice.

tinwelint:
Thanks for feedback,

we did go through page cursor access patterns as part of this PR, if you've found some specific place which ain't valid it'd be great to know about them.

Also again the note about PageCacheRule moving to neo4j-io which may be a bigger thing than anticipated, although haven't tried for real so not sure.

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
tinwelint: Hmm, actually no, the two places have different data in their exceptions. I don't think they can be merged.

tinwelint:
@ragadeeshu test failure is in unrelated test

tinwelint:
@ragadeeshu any outstanding issue still unaddressed?

ragadeeshu:
@tinwelint No, it seems that you are very aware of issues and trade-offs, and I agree that this is not the place for a io/kernel refactor.

tinwelint:
Will be reworked and replaced by another PR soon

tinwelint:
@ragadeeshu see #8584 instead, it differs in following ways:

- one commit (first one) which improves on configurability on the PageCacheRule in neo4j-io
- `GB+Tree tests run with adversarial page cache` commit was slightly reworked (its SeekCursor changes) since there were many refactorings made in other PRs that went in before it.

Otherwise it's the same thing so you only really need to re-review those points. Hopefully it won't be too much of an inconvenience for you this way, cheers!

PR #8539, opened 2016-12-21
Implement new neo4j-admin backup surface

spacecowboy:
New help text:

```
usage: neo4j-admin backup --backup-dir=<backup-path> --name=<graph.db-backup>
                          [--from=<address>] [--fallback-to-full[=<true|false>]]
                          [--check-consistency[=<true|false>]]
                          [--cc-report-dir=<directory>]
                          [--additional-config=<config-file-path>]
                          [--timeout=<timeout>]

Perform an online backup from a running Neo4j enterprise server. Neo4j's backup
service must have been configured on the server beforehand. See
http://neo4j.com/docs/operations-manual/current/backup/ for more details.

options:
  --backup-dir=<backup-path>               Directory to place backup in.
  --name=<graph.db-backup>                 Name of backup. If a backup with this
                                           name already exists an incremental
                                           backup will be attempted.
  --from=<address>                         Host and port of Neo4j.
                                           [default:localhost:6362]
  --fallback-to-full=<true|false>          If a failed incremental backup will
                                           move the old backup to <name>.err.<N>
                                           and fallback to a full backup
                                           instead. [default:true]
  --check-consistency=<true|false>         If a consistency check should be
                                           made. [default:true]
  --cc-report-dir=<directory>              Directory where consistency report
                                           will be written. [default:.]
  --additional-config=<config-file-path>   Configuration file to supply
                                           additional configuration in.
                                           [default:]
  --timeout=<timeout>                      Timeout in the form <time>[ms|s|m|h],
                                           where the default unit is seconds.
                                           [default:20m]

```

changelog 	
Implement new `neo4j-admin backup` surface

* consistency-check, if specified, is run for *both* full and
  incremental backup
* Exit with clear error messages on failures with exit code 1
* If incremental backup fails, the old backup is *moved*, and then it
  falls back to full backup
* fallback behavior can be disabled if desired, in which case failures
  are reported directly and old backup is *not* moved

### DIFF
--- a/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
@@ -33,6 +33,7 @@ import org.neo4j.commandline.arguments.Arguments;
 import org.neo4j.commandline.arguments.MandatoryNamedArg;
 import org.neo4j.commandline.arguments.OptionalBooleanArg;
 import org.neo4j.commandline.arguments.OptionalNamedArg;
+import org.neo4j.commandline.arguments.common.MandatoryCanonicalPath;
 import org.neo4j.commandline.arguments.common.OptionalCanonicalPath;
 import org.neo4j.consistency.ConsistencyCheckService;
 import org.neo4j.consistency.ConsistencyCheckSettings;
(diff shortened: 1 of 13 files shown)


enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
```diff
@@ -50,15 +53,20 @@
...
+            .withArgument( new MandatoryNamedArg( "name", "graph.db-backup",
+                    "Name of backup. If a backup with this name already exists an incremental backup will be " +
+                            "attempted." ) )
             .withArgument( new OptionalNamedArg( "from", "address", "localhost:6362",
                     "Host and port of Neo4j." ) )
-            .withArgument( new MandatoryNamedArg( "to", "backup-path",
-                    "Directory where the backup will be made; if there is already a backup present an " +
-                            "incremental backup will be attempted." ) )
+            .withArgument( new OptionalBooleanArg( "fallback-to-full", true,
+                    "If a failed incremental backup will move the old backup to <name>.err.<N> and fallback to a " +
```
srbaker: "If an incremental backup fails backup will move ..."

srbaker (changes requested):

srbaker (approved):

PR #8540, opened 2016-12-22
Simplify write page cursor usage

chrisvest:


tinwelint:
Shouldn't https://github.com/neo4j/neo4j/blob/3.2/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java#L673 also change to use this new `isWriteLocked()` method?

chrisvest:
@tinwelint No, that method relies on the implementation details of the target cursor, namely the native pointer address. I discussed with @burqen about adding a slow-path that would loop and copy byte-by-byte, but we decided instead to just document that the cursors should be from the same page cache implementation instead. The reason is that we only use copyTo with alternative cursor implementations in testing.

tinwelint:
@chrisvest makes sense, I can see that now

burqen (changes requested):
Looks good! Nice javadoc and sweet separation between commits.

Take a look at `PagedWritableByteChannel#write`, looks like `shouldRetry` can be removed from `#write` as well.

chrisvest:
@burqen Done.

burqen (approved):

PR #8541, opened 2016-12-22
Cleanup

davidegrohmann:


community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/AllIdIterator.java
```diff
@@ -0,0 +1,49 @@
...
+        if ( currentId <= highId )
+        {
+            try
+            {
+                return next( currentId );
+            }
+            finally
+            {
+                currentId++;
+            }
```
chrisvest, line 45: If the next fails, the `currentId` will still get incremented so we'll effectively move on to the next one. Is that what we want here?

chrisvest: Okay, I see you just extracted the class verbatim and preserved whatever was there. Fine.

chrisvest (changes requested):
Looks good. Just update the license headers to 2017 and it'll be mergeable. Sorry I didn't get to this sooner.

chrisvest (approved):

PR #8550, opened 2016-12-27
Remove unused code from Cypher compiler

systay:


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/commands/expressions/ShortestPathExpression.scala
```diff
@@ -176,7 +174,5 @@ case class ShortestPathExpression(shortestPathPattern: ShortestPath, predicates:
           }
       }
 
-  private def doesNotDependOnFullPath(predicate: Predicate): Boolean = {
-    (predicate.symbolTableDependencies intersect pathVariables).isEmpty
-  }
+  private def doesNotDependOnFullPath(predicate: Predicate): Boolean = false // TODO: This is not correct!
```
henriknyman, line 171: So how would we determine this now?

Mats-SX: We could extend the `Predicate` class to contain a set of dependencies, but perhaps that's what we aimed to get rid of with this commit?

davidegrohmann:
@systay is this still relevant?

PR #8552, opened 2016-12-27
InternalTreeLogic uses manual stack instead of recursion

tinwelint:
this reduces number of page pins since the whole depth won't need to
be traversed up and down for every insert/remove, e.g:

````
                    [A]
             ------/ | \------
            /        |        \
          [B]       [C]       [D]
         / | \     / | \     / | \
       [E][F][G] [H][I][J] [K][L][M]
````

where previously an insert into F, then G, then L would result in page pins
(every pin also including a binary key search):

- for F: A, B, F, B, A
- for G: A, B, G, B, A
- for L: A, D, L, D, A
= totaling 15 pins

but now instead this would look like:

- for F: A, B, F
- for G: B, G
- for L: A, D, L
= totaling 8 pins

the more co-located updates are within the same session (i.e. for the same writer)
the fewer pins and binary searches are required. This works by keeping a stack of where
the writer is in the tree and the key ranges for every level it has gone down.
For every insert/remove the current location is asked whether or not it covers the key
to insert/remove, if not then stack is popped until finding a level which covers the key.

This stack consists of an array of objects initialized once when the GBPTree
instance is constructed and the stack is just an integer knowing which array item
is the current one, i.e. it's garbage-free. Popping the stack also doesn't involve
page pins since the stack knows the key ranges.

tl;dr
The net result of this change is fewer page pins and binary searches to find
correct leaf to perform inserts/removals at. Worst-case is slightly better than before
and best-case is very much better - sequential batch insert being a couple of times
faster due to magnitudes fewer pins and binary searches.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
+++ b/community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
@@ -20,6 +20,9 @@
 package org.neo4j.index.gbptree;
 
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.Comparator;
+
 import org.neo4j.index.IndexWriter;
 import org.neo4j.index.ValueMerger;
 import org.neo4j.io.pagecache.PageCursor;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java [hunk 1]
```diff
@@ -84,15 +87,230 @@
...
     InternalTreeLogic( IdProvider idProvider, TreeNode<KEY,VALUE> bTreeNode, Layout<KEY,VALUE> layout )
     {
         this.idProvider = idProvider;
         this.bTreeNode = bTreeNode;
         this.layout = layout;
-        int maxKeyCount = max( bTreeNode.internalMaxKeyCount(), bTreeNode.leafMaxKeyCount() );
         this.primKeyPlaceHolder = layout.newKey();
         this.readKey = layout.newKey();
         this.readValue = layout.newValue();
+        ensureStackCapacity( 20 );
```
burqen: 20 levels as default is gonna leave about 13-14 unused for the normal use case with a normal page size.

community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
```diff
@@ -84,15 +87,230 @@
...
+            }
+
+            long childId = bTreeNode.childAt( cursor, childPos, stableGeneration, unstableGeneration );
+            PointerChecking.checkPointer( childId, false );
+
+            bTreeNode.goTo( cursor, "child", childId );
+            level.treeNodeId = cursor.getCurrentPageId();
+        }
+
+        assert bTreeNode.isLeaf( cursor );
```
burqen: Can we have a better error message here

community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
```diff
@@ -129,41 +348,32 @@ void insert( PageCursor cursor, StructurePropagation<KEY> structurePropagation,
             ValueMerger<VALUE> valueMerger, IndexWriter.Options options,
             long stableGeneration, long unstableGeneration ) throws IOException
     {
-        if ( bTreeNode.isLeaf( cursor ) )
-        {
-            insertInLeaf( cursor, structurePropagation, key, value, valueMerger, options,
-                    stableGeneration, unstableGeneration );
-            return;
-        }
+        assert currentLevel >= 0;
```
burqen: Error message.
Maybe also assert that `currentPageId` is equal to that of current level?

community/index/src/test/java/org/neo4j/index/gbptree/InternalTreeLogicTest.java
```diff
@@ -798,17 +741,13 @@ public void shouldCreateNewVersionWhenInsertInStableLeaf() throws Exception
         //       ------root-------
         //      /        |         \
         //     v         v          v
-        //   left <--> middle <--> right
-        node.initializeLeaf( cursor, stableGen, unstableGen );
+        //   left <--> middle <--> right        initialize();
```
burqen: Oops. I guess this was not intentional

burqen (changes requested):
InternalTreeLogic also has two compilation errors on line 336 and 810.

community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
tinwelint: Not quite a problem, it's a small "cost" of having a stack that's capable of even the craziest tree. Although we can probable set it to 10, cool?

community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
tinwelint: Oh sure

community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
tinwelint: A good idea

community/index/src/test/java/org/neo4j/index/gbptree/InternalTreeLogicTest.java
tinwelint: Oh so that's where that `initialize()` went...

community/index/src/main/java/org/neo4j/index/gbptree/InternalTreeLogic.java
burqen: 10 should be sufficient!

burqen (approved):

burqen:
Nice, will merge once green.

PR #8553, opened 2016-12-28
Removes IndexWriter.Options

tinwelint:
since it only contained option for deciding where cut-off point would be done
when splitting nodes. Initially introduced to be able to speed up batch-insertion
where cut-off point would be far to the right so that left nodes would be more filled
and so cause less splits during batch-insert.

However this would later on cause split of almost any change on any node when
switching to normal update mode. Also batch-insert gain much more on other optimizations
that have been done anyway.


burqen:
Sweet! Merge once green.

burqen (approved):

PR #8556, opened 2016-12-29
Make bookmark await timeout configurable

lutovich:
It was previously set to 30 seconds without a possibility to change this value. This PR makes it configurable via `dbms.transaction.bookmark_ready_timeout`. Setting is located in community `GraphDatabaseSettings` because bookmark is exposed in both single instance and Causal Cluster. In single instance bookmark is basically a no-op.


zhenlineo:
LGTM. Let me ask if anyone wants to improve the config name

PR #8568, opened 2017-01-03
Remove unused code around cursors

davidegrohmann:


community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -298,16 +298,6 @@ public boolean nodeExists( long nodeId )
     }
 
     @Override
-    public boolean relationshipExists( long relId )
```
jexp: How would I check that in the future?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -423,13 +423,7 @@ public PrimitiveLongIterator relationshipsGetAll( KernelStatement state )
...
     }
 
-    @Override
-    public Cursor<NodeItem> nodeCursor( KernelStatement statement, long nodeId )
-    {
-        return entityReadOperations.nodeCursor( statement, nodeId );
-    }
-
-    @Override
+        @Override
```
chrisvest: Indentation mishap

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -482,10 +477,9 @@ public NodePropertyExistenceConstraint nodePropertyExistenceConstraintCreate( Ke
     public RelationshipPropertyExistenceConstraint relationshipPropertyExistenceConstraintCreate( KernelStatement state,
             int relTypeId, int propertyKeyId ) throws AlreadyConstrainedException, CreateConstraintFailureException
     {
-        try ( Cursor<RelationshipItem> cursor = relationshipCursorGetAll( state ) )
-        {
-            constraintSemantics.validateRelationshipPropertyExistenceConstraint( cursor, relTypeId, propertyKeyId );
-        }
+        Iterator<Cursor<RelationshipItem>> relationships = new EntityLoadingIterator<>( relationshipsGetAll( state ),
+                (id) -> relationshipCursorById( state, id ) );
+        constraintSemantics.validateRelationshipPropertyExistenceConstraint( relationships, relTypeId, propertyKeyId );
```
chrisvest: `Iterator<Cursor<…>>` seems rather strange. A cursor should be able to move around. I think the new code will load the relationship records twice, while the old code would only load it once.

chrisvest: I think it would make more sense to keep the cursor API, and implement `relationshipsGetAll` in terms of the cursor instead. We probably still need the old method to remain compatible with older Cypher binaries.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
davidegrohmann: why do you use an internal/undocumented API for that?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -466,10 +460,9 @@ public UniquenessConstraint uniquePropertyConstraintCreate( KernelStatement stat
     public NodePropertyExistenceConstraint nodePropertyExistenceConstraintCreate( KernelStatement state, int labelId,
             int propertyKeyId ) throws AlreadyConstrainedException, CreateConstraintFailureException
     {
-        try ( Cursor<NodeItem> cursor = nodeCursorGetForLabel( state, labelId ) )
-        {
-            constraintSemantics.validateNodePropertyExistenceConstraint( cursor, labelId, propertyKeyId );
-        }
+        Iterator<Cursor<NodeItem>> nodes = new EntityLoadingIterator<>( nodesGetForLabel( state, labelId ),
+                ( id ) -> nodeCursorById( state, id ) );
+        constraintSemantics.validateNodePropertyExistenceConstraint( nodes, labelId, propertyKeyId );
```
chrisvest, line 473: Unlike the relationships change, this one is probably fine, because the IDs are fed from an index search, rather than a store scan.

community/kernel/src/main/java/org/neo4j/kernel/impl/constraints/ConstraintSemantics.java
```diff
@@ -36,7 +36,7 @@
  */
 public interface ConstraintSemantics
 {
-    void validateNodePropertyExistenceConstraint( Cursor<NodeItem> allNodes, int label, int propertyKey )
+    void validateNodePropertyExistenceConstraint( Iterator<Cursor<NodeItem>> allNodes, int label, int propertyKey )
```
chrisvest, line 39: I still don't like the `Iterator<Cursor<…>>` meme, though.

chrisvest (changes requested):
Very good to see a PR split into commits in such a nice way. I enjoy seeing code be deleted – especially code that has so little testing – but I'm disagreeing with the `Remove relationshipCursorGetAll` part.

community/kernel/src/main/java/org/neo4j/kernel/impl/constraints/ConstraintSemantics.java
davidegrohmann: Yeah it is a bit silly.. but you need to "close" the "SingleNodeCursor" since they are cached on the statement. I agree this is not nice but we can clean this up later, it is not a small change.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
davidegrohmann: the relationship cursor was loading the record twice as well, there is no change there. See https://github.com/neo4j/neo4j/blob/3.2/community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/StoreStatement.java#L160

I agree the `Iterator<Cursor<...>>` is silly, I plan to clean that up later (see the other comment).

chrisvest: The `AllStoreIdIterator` does not load the record; it only counts through `0` to `highId`.

davidegrohmann: You are right, I got confused.

davidegrohmann: I'll drop this commit for now.

davidegrohmann:
@chrisvest fixed your comment around all relationship cursor

chrisvest:
@davidegrohmann Thanks, this looks good!

chrisvest (approved):

PR #8570, opened 2017-01-03
Fix idle backup seeding

martinfurmanski:
changelog: [causal clustering] Seeding a new member of an idle cluster using a backup from that idle cluster could fail.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloader.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloader.java
@@ -25,10 +25,11 @@ import org.neo4j.causalclustering.catchup.CatchUpClient;
 import org.neo4j.causalclustering.catchup.CatchUpResponseAdaptor;
 import org.neo4j.causalclustering.catchup.CatchupResult;
 import org.neo4j.causalclustering.catchup.storecopy.LocalDatabase;
-import org.neo4j.causalclustering.catchup.storecopy.StoreCopyFailedException;
 import org.neo4j.causalclustering.catchup.storecopy.RemoteStore;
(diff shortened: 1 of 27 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloader.java
```diff
@@ -66,6 +66,8 @@ public CoreStateDownloader( FileSystemAbstraction fs, LocalDatabase localDatabas
     public synchronized void downloadSnapshot( MemberId source, CoreState coreState ) throws StoreCopyFailedException
     {
         // TODO: Think about recovery scenarios.
+        StoreCopyProcess storeCopyProcess = new StoreCopyProcess( fs, localDatabase, copiedStoreRecovery, log );
+
```
davidegrohmann: field?

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/helpers/DataCreator.java
```diff
@@ -0,0 +1,40 @@
...
+        for ( int i = 0; i < numberOfNodes; i++ )
+        {
+            last = cluster.coreTx( ( db, tx ) ->
+            {
+                db.createNode();
+                tx.success();
+            } );
+        }
+        return last;
+    }
```
chrisvest, line 39: There is already a class called `SampleData` that has a similar method.

chrisvest:
The kernel changes look good.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/helpers/DataCreator.java
martinfurmanski: I'll take a look at that separately

PR #8571, opened 2017-01-03
GBPTree - Backwards seek for SeekCursor

burqen:
Introduce backwards seek in SeekCursor. This makes it possible to scan index built on top of GBPTree in descending order.

**Why is this so complicated?**

*General*
To allow for lock free concurrency control the GBPTree agree to only move keys between nodes in 'forward' direction, from left to right. This means when we do normal forward seek we never risk having keys moved 'passed' us and therefore we can always be sure that we never miss any keys when traversing the leaf nodes or end up in the middle of the range when traversing down the internal nodes.
This assumption, of course, does not hold when seeking backwards.

There are two complicated cases where we risk missing keys.
Case 1 - Split in current node
Case 2 - Split in next node while seeker is moving to that node

**Case 1 - Split in current node**
*Forward seek*
Here, the seeker is seeking forward and has read `K0, K1, K2` and is about to read `K3`.
Then `K4` is suddenly inserted and a split happens.
Seeker will now wake up and find that he is now outside the range of `(previously returned key, end of range)`. But he knows that he can continue to read forward until he hits the previously returned key, which is `K2` and then continue to return the next key, `K3`.
```
FORWARD
    Seeker->
       v  
[K0 K1 K2 K3]

    Seeker->
       v  
[K0 K1 __ __]<->[K2 K3 K4 __]
```

*Backward seek*
Here, the seeking is seeking backwards and has only read `K3` and is about to read `K2`.
Again, `K4` is inserted, causing the same split as above.
Seeker now wakes up and find that the next key he can return is `K1`. What he do not see is that `K2` has been moved to the previous sibling and so, because he can not find the previously returned key, `K3`, in the current node and because the next to return, `K1`, is located to the far right in the current node he need to jump back to the previous sibling to find where to start again.
```
BACKWARD
     <-Seeker
          v  
[K0 K1 K2 K3]

     <-Seeker (WRONG)
          v  
[K0 K1 __ __]<->[K2 K3 K4 __]

               <-Seeker (RIGHT)
                    v  
[K0 K1 __ __]<->[K2 K3 K4 __]
```

**Case 2 - Split in next node while seeker is moving to that node**
*Forward seek*
Seeker has read `K0, K1` on node 1  and has just moved to right sibling, node 2.
Now, `K6` is inserted and a split  happens in node 2, right sibling of node 1.
Seeker wakes up and continues to read on node 2. Everything is fine.
```
FORWARD
                  Seeker->
                     v  
1:[K0 K1 __ __]<->2:[K2 K3 K4 K5]

                  Seeker->
                     v  
1:[K0 K1 __ __]<->2:[K2 K3 __ __]<->3:[K4 K5 K6 __]
```
*Backward seek*
Seeker has read `K4, K5` and has just moved to left sibling, node 1.
Insert `K6`, split in node 2. Note that keys are move to node 3, passed our seeker.
Seeker wakes up and see `K1` as next key but misses `K3` and `K2`.
```
BACKWARD
                      <--Seeker
                            v  
                1:[K0 K1 K2 K3]<->2:[K4 K5 __ __]

       <-Seeker
            v  
1:[K0 K1 __ __]<->3:[K2 K3 K4 __]<->2:[K5 K6 __ __]
```
To guard for this, seeker 'scout' next sibling before moving there and read first key that he expect to see, `K3` in this case. By using a linked cursor to 'scout' we create a consistent read over the node gap. If there us suddenly another key when he goes there he knows that he could have missed some keys and he needs to go back until he find the place where he left off, `K4`.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
+++ b/community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
@@ -414,9 +414,10 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
             checkOutOfBounds( cursor );
 
             // Act
-            if ( saneRead() )
+            if ( notSaneRead() )
             {
-                restartSeekFromRoot();
+                prepareToStartFromRoot();
+                isInternal = true;
                 continue;
             }
 
@@ -494,10 +495,11 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
             checkOutOfBounds( cursor );
 
             // Act
-            if ( saneRead() )
+            if ( notSaneRead() )
             {
                 // This node has been reused for something else than a tree node. Restart seek from root.
-                restartSeekFromRoot();
+                prepareToStartFromRoot();
+                traverseDownToFirstLeaf();
                 continue;
             }
 
@@ -634,14 +636,14 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
      */
     private boolean verifyFirstKeyInNodeIsExpectedAfterGoTo()
     {
+        boolean result = true;
         if ( verifyExpectedFirstAfterGoToNext && layout.compare( firstKeyInNode, expectedFirstAfterGoToNext ) != 0 )
         {
             concurrentWriteHappened = true;
-            verifyExpectedFirstAfterGoToNext = false;
-            return false;
+            result = false;
         }
         verifyExpectedFirstAfterGoToNext = false;
-        return true;
+        return result;
     }
 
     /**
@@ -706,12 +708,8 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
         isInternal = bTreeNode.isInternal( cursor );
         // Find the left-most key within from-range
         keyCount = bTreeNode.keyCount( cursor );
-        if ( !keyCountIsSane( keyCount ) )
-        {
-            return false;
-        }
 
-        return true;
+        return keyCountIsSane( keyCount );
     }
 
     /**
@@ -901,14 +899,15 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
     }
 
     /**
-     * Restarts seek from root, i.e. updates current root (and its generation) and traverses down to
-     * leaf containing the key containing the last returned, or first, key.
+     * Perform a generation catchup, updates current root and update range to start from
+     * previously returned key. Should be followed by a call to {@link #traverseDownToFirstLeaf()}
+     * or if already in that method just loop again.
      * <p>
      * Caller should retry most recent read after calling this method.
      *
      * @throws IOException on {@link PageCursor}.
      */
-    private void restartSeekFromRoot() throws IOException
+    private void prepareToStartFromRoot() throws IOException
     {
         generationCatchup();
         lastFollowedPointerGen = rootCatchup.get().goTo( cursor );
@@ -916,8 +915,6 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
         {
             layout.copyKey( prevKey, fromInclusive );
         }
-        isInternal = true;
-        traverseDownToFirstLeaf();
     }
 
     /**
@@ -1002,7 +999,7 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
      * @return {@code true} if read is sane and can be trusted, otherwise {@code false} meaning that
      * seek should be restarted from root.
      */
-    private boolean saneRead()
+    private boolean notSaneRead()
     {
         return nodeType != NODE_TYPE_TREE_NODE || !saneKeyCountRead( keyCount ) || !verifyNodeGenInvariants();
     }
--- a/community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
+++ b/community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
@@ -44,6 +44,14 @@ import static org.neo4j.index.gbptree.TreeNode.NODE_TYPE_TREE_NODE;
  * guards for that by re-reading if change happens underneath, but will not provide a consistent view of
  * the data as it were when the seek starts, i.e. doesn't support MVCC-style.
  * <p>
+ * Seek can be performed forwards or backwards, returning hits in ascending or descending order respectively
+ * (as defined by {@link Layout#compare(Object, Object)}). Direction is decided on relation between
+ * {@link #fromInclusive} and {@link #toExclusive}. Backwards seek is expected to be slower than forwards seek
+ * because extra care needs to be taken to make sure no keys are skipped when keys are move to the right in the tree.
+ * <pre>
+ * If fromInclusive <= toExclusive, then seek forwards, otherwise seek backwards
+ * </pre>
+ * <p>
  * Implementation note: there are assumptions that keys are unique in the tree.
  */
 class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hit<KEY,VALUE>
@@ -64,12 +72,14 @@ class SeekCursor<KEY,VALUE> implements RawCursor<Hit<KEY,VALUE>,IOException>, Hi
     private final VALUE mutableValue;
 
     /**
-     * Provided when constructing the {@link SeekCursor}, marks the low end of the key range to seek.
+     * Provided when constructing the {@link SeekCursor}, marks the start (inclusive) of the key range to seek.
+     * Comparison with {@link #toExclusive} decide if seeking forwards or backwards.
      */
     private final KEY fromInclusive;
 
     /**
-     * Provided when constructing the {@link SeekCursor}, marks the high end (exclusive) of the key range to seek.
+     * Provided when constructing the {@link SeekCursor}, marks the end (exclusive) of the key range to seek.
+     * Comparison with {@link #fromInclusive} decide if seeking forwards or backwards.
      */
     private final KEY toExclusive;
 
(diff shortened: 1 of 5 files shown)


community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -178,299 +363,551 @@
...
+
+        newGen = bTreeNode.newGen( cursor, stableGeneration, unstableGeneration );
+        if ( GenSafePointerPair.isSuccess( newGen ) )
+        {
+            newGenGen = bTreeNode.pointerGen( cursor, newGen );
+        }
+        isInternal = bTreeNode.isInternal( cursor );
+        // Find the left-most key within from-range
+        keyCount = bTreeNode.keyCount( cursor );
+        if ( !keyCountIsSane( keyCount ) )
```
MishaDemianenko: line 709-712 can be simplify to `return keyCountIsSane( keyCount );`

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -545,6 +998,15 @@ private boolean generationCatchup()
         return false;
     }
 
+    /**
+     * @return {@code true} if read is sane and can be trusted, otherwise {@code false} meaning that
+     * seek should be restarted from root.
+     */
+    private boolean saneRead()
```
MishaDemianenko: looks like it should be `insaneRead()` or `notSaneRead()` or `gibberishRead()` instead

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -479,9 +916,18 @@ private void restartSeekFromRoot() throws IOException
         {
             layout.copyKey( prevKey, fromInclusive );
         }
+        isInternal = true;
         traverseDownToFirstLeaf();
```
MishaDemianenko: we do not really need to call `traverseDownToFirstLeaf` as part of `restartSeekFromRoot` when we already in `traverseDownToFirstLeaf` loop we only need to do that when we doing next(). In super extreme cases that can cause stack overflow. Would propose to not call it internally here but do that only in next() call path

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -178,299 +363,551 @@
...
+     *
+     * @return position of the {@code key} in the current tree node, or position of the closest key.
+     */
+    private int searchKey( KEY key )
+    {
+        int search = KeySearch.search( cursor, bTreeNode, key, mutableKey, keyCount );
+        int pos = KeySearch.positionOf( search );
+
+        // Assuming unique keys
+        if ( seekForward && isInternal && KeySearch.isHit( search ) )
```
MishaDemianenko: it looks like we should do pos++ for forward and backwards scans and not only for seekForward case. In backward case we will end up in a wrong position

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -178,299 +363,551 @@
...
-    private boolean saneKeyCountRead( int keyCount )
-    {
-        return keyCount <= maxKeyCount;
-    }
-
-    @Override
-    public Hit<KEY,VALUE> get()
-    {
-        return this;
+        pos -= stride;
```
MishaDemianenko, line 437: position evaluation mechanics should be part of next() call only instead. To be able to achieve that next can be updated to contain finally or do while block to manipulate pos instead.

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -178,299 +363,551 @@
...
                 }
 
-                if ( GenSafePointerPair.isSuccess( newGen ) )
+                if ( concurrentWriteHappened )
                 {
-                    newGenGen = bTreeNode.pointerGen( cursor, newGen );
+                    // Keys could have been moved so we need to make sure we are not missing any keys by
+                    // moving position back until we find previously returned key
+                    pos = searchKey( first ? fromInclusive : prevKey );
+                    if ( !seekForward && pos >= keyCount )
```
MishaDemianenko, line 472: how pos can be >= the keyCount in case of !seekForward?

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -178,299 +363,551 @@
...
+        return -1; // this value doesn't matter
+    }
+
+    /**
+     * @return {@code false} if there was a set expectancy on first key in tree node which weren't met,
+     * otherwise {@code true}. Caller should
+     */
+    private boolean verifyFirstKeyInNodeIsExpectedAfterGoTo()
+    {
+        if ( verifyExpectedFirstAfterGoToNext && layout.compare( firstKeyInNode, expectedFirstAfterGoToNext ) != 0 )
```
MishaDemianenko, line 640: verifyExpectedFirstAfterGoToNext is always false as result of this method.
Can be updated to something like this:
```
        boolean result = true;
        if ( verifyExpectedFirstAfterGoToNext && layout.compare( firstKeyInNode, expectedFirstAfterGoToNext ) != 0 )
        {
            concurrentWriteHappened = true;
            result = false;
        }
        verifyExpectedFirstAfterGoToNext = false;
        return result;
```

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
```diff
@@ -498,6 +498,196 @@ else if ( lastSeenKey > nextToSee )
...
+
+        // given
+        index = createIndex( 256 );
+
+        Runnable reader = () -> {
+            int numberOfLocalReads = 0;
+            try
+            {
+                readerReadySignal.countDown();
+                while ( currentWriteIteration.get() < 1 )
```
MishaDemianenko: not sure i see a reason for this loop

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
```diff
@@ -498,6 +498,196 @@ else if ( lastSeenKey > nextToSee )
...
+                {
+                    startSignal.await( 5, SECONDS );
+                }
+
+                while ( !endSignal.get() && !failHalt.get() )
```
MishaDemianenko: readerError should also terminate all and fail the test?

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
```diff
@@ -498,6 +498,196 @@ else if ( lastSeenKey > nextToSee )
...
+                readerError.set( e );
+            }
+            finally
+            {
+                numberOfReads.addAndGet( numberOfLocalReads );
+            }
+        };
+
+        // WHEN starting the readers
+        Thread[] readerThreads = new Thread[readers];
```
MishaDemianenko: please use poooool instead

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
```diff
@@ -498,6 +498,196 @@ else if ( lastSeenKey > nextToSee )
...
+        for ( int i = 0; i < readers; i++ )
+        {
+            readerThreads[i] = new Thread( reader );
+            readerThreads[i].start();
+        }
+
+        // and starting the checkpointer
+        Thread checkpointer = checkpointerThread( minCheckpointInterval, maxCheckpointInterval, endSignal );
```
MishaDemianenko: please use thread from pool

MishaDemianenko: if checkpointer thread will fail we will not gonna notice that

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
```diff
@@ -498,6 +498,196 @@ else if ( lastSeenKey > nextToSee )
...
+                        lastWrittenKey.set( i );
+                        if ( failHalt.get() )
+                        {
+                            break;
+                        }
+                    }
+                }
+                iteration = currentWriteIteration.incrementAndGet();
+                // Sleep a little in between update groups (transactions, sort of)
+                MILLISECONDS.sleep( random.nextInt( 10 ) + 3 );
```
MishaDemianenko, line 465: maybe use `nextInt( int origin, int bound )` instead?

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
```diff
@@ -498,6 +498,196 @@ else if ( lastSeenKey > nextToSee )
...
+        CountDownLatch startSignal = new CountDownLatch( 1 );
+        AtomicBoolean endSignal = new AtomicBoolean();
+        AtomicBoolean failHalt = new AtomicBoolean(); // Readers signal to writer that there is a failure
+        AtomicReference<Throwable> readerError = new AtomicReference<>();
+        AtomicInteger numberOfReads = new AtomicInteger();
+
+        // given
+        index = createIndex( 256 );
+
+        Runnable reader = () -> {
```
MishaDemianenko: please extract reader to it's own class

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
```diff
@@ -498,6 +498,196 @@ else if ( lastSeenKey > nextToSee )
     }
 
     @Test
+    public void shouldReadCorrectlyWhenConcurrentlyInsertingOutOfOrderAndSeekingBackwards() throws Throwable
```
MishaDemianenko, line 398: comments for this test is applicable to his copy pasted brother bellow as well :)

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
```diff
@@ -498,6 +498,196 @@ else if ( lastSeenKey > nextToSee )
...
+                        {
+                            lastWrittenBeforeStart = lastWrittenWhenFinished;
+                            lastWrittenWhenFinished = lastWrittenKey.get();
+
+                            lastSeenKey = cursor.get().key().longValue();
+                            long lastSeenValue = cursor.get().value().longValue();
+                            if ( lastSeenKey != lastSeenValue )
+                            {
+                                failHalt.set( true );
+                                fail( String.format( "Read mismatching key value pair, key=%d, value=%d%n",
```
MishaDemianenko: failing in separate thread will not gonna fail original test, will it? Maybe save message somewhere and fail main test thread instead?

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
```diff
@@ -44,7 +43,102 @@
...
+ * 1:[K0 K1 K2 K3]<->2:[K4 K5 __ __]
+ *
+ *        <-Seeker
+ *             v
+ * 1:[K0 K1 __ __]<->3:[K2 K3 K4 __]<->2:[K5 K6 __ __]
+ * </pre>
+ * To guard for this, seeker 'scout' next sibling before moving there and read first key that he expect to see, K3
+ * in this case. By using a linked cursor to 'scout' we create a consistent read over the node gap. If there us
+ * suddenly another key when he goes there he knows that he could have missed some keys and he needs to go back until
+ * he find the place where he left off, K4.
```
MishaDemianenko, line 141: Now this cursor is almost impossible to read and understand even when i can grab @burqen and bomb him with unstoppable rain of questions. Have you guys consider possibility to split this cursor at least into seek cursor and then have 2 actual public visible forward and backward cursor classes that do the stuff only for their use case. That at least will gonna simplify logic around everything that responsible for the navigation logic

MishaDemianenko (changes requested):
couple of comments and a big bigger concern in regards of readability of new monster

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
burqen: Node: `[10 20 30 40]`, `keyCount = 4`
`pos = searchKey( 50 )` will result in `pos == 4`

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
burqen: The readers will use last fully applied iteration to predict what he should see in the tree. If we do not wait until the first iteration is complete there is no way for the reader to predict anything about what data he can expect.

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
burqen: It will fail the test because before exiting `readerErrror` is checked and thrown if not `null`.
You are correct in that it should also terminate by setting `failHalt` to `true.`

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
burqen: You are correct. Verified with
```
    @Test
    public void stupidTest() throws Exception
    {
        Thread failThread = new Thread( () -> fail( "with some message"  ) );
        failThread.start();
        failThread.join();
    }
```
that passes. However, fail will throw `assertionError` that will end up in `readerError` that will eventually fail the test. Confirmed by
```
@Test
    public void stupidTest() throws Throwable
    {
        AtomicReference<Throwable> readerError = new AtomicReference<>();
        Thread failThread = new Thread( () ->
        {
            try
            {
                fail( "with some message" );
            }
            catch ( Throwable e )
            {
                readerError.set( e );
            }
        } );
        failThread.start();
        failThread.join();

        if ( readerError.get() != null )
        {
            throw readerError.get();
        }
    }
```

community/index/src/main/java/org/neo4j/index/gbptree/SeekCursor.java
burqen: This is a super valid concern and I agree, SeekCursor is becoming a monster that both me and @tinwelint have a hard time to understand.
I suggest we clean it up in another PR though and leave this one to be about backwards seek (which is already quite large).
Comment about `pos -= stride` would also be covered by suggested clean up.
I've created a card for this task / refactoring.

burqen:
@MishaDemianenko Some comments addressed and some commented on. Please have a look.

community/index/src/test/java/org/neo4j/index/gbptree/GBPTreeIT.java
MishaDemianenko: that what i wanted - to store exception/exception error somewhere and pass back to main thread

MishaDemianenko:
@burqen looks fine now, failure in ha robustness is old and unrelated

MishaDemianenko (approved):

PR #8573, opened 2017-01-03
Make tests use TestGraphDatabaseFactory

chrisvest:
And also delete some unused test code.


chrisvest:
@tinwelint We've confirmed that this is now green on Solaris, so it's good to merge now.

tinwelint:
I can imagine this will be an iterative process, as future versions will have added tests that will have the same problems of not using test graph database. Not sure this is up to the merger (me) to do. Doing the heavily conflicting merge is a handful already :)

chrisvest:
The changes to the rule & test factories have the biggest bang for buck. If there's anything that's too difficult to merge now, you can skip it and we can redo it from scratch on the newer branches later.

PR #8575, opened 2017-01-03
Fix eagerness for labels() function

henriknyman:
We need to be eager when the use of labels() function in a query horizon
conflicts with a label write, which requires new conflict checks.

- For the rule planner a new read effect is introduced
- For read-write conflicts in the cost planner, a new location to plan eagerness
is introduced, with a dedicated check and pass, after planning the event horizon.
- For write-read conflicts in the cost planner, the check is added to the
existing passes.

Also fixes related eagerness bugs for property reads and extends testing.

### DIFF
--- a/community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/UpdateGraph.scala
+++ b/community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/UpdateGraph.scala
@@ -19,7 +19,7 @@
  */
 package org.neo4j.cypher.internal.compiler.v3_0.planner
 
-import org.neo4j.cypher.internal.compiler.v3_0.planner.logical.plans.{PatternRelationship, IdName}
+import org.neo4j.cypher.internal.compiler.v3_0.planner.logical.plans.IdName
 import org.neo4j.cypher.internal.frontend.v3_0.ast._
 
 import scala.annotation.tailrec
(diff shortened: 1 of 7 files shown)


community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/logical/Eagerness.scala
```diff
@@ -195,8 +218,13 @@ object Eagerness {
   @tailrec
   def writeReadConflictInHeadRecursive(head: PlannerQuery, tail: PlannerQuery): Boolean = {
     // TODO:H Refactor: This is same as writeReadConflictInTail, but with different overlaps method. Pass as a parameter
-    val conflict = if (tail.queryGraph.writeOnly) false
-    else head.queryGraph writeOnlyHeadOverlaps tail.queryGraph
+    val conflict =
+      if (tail.queryGraph.writeOnly) false
+      else
+        // NOTE: Here we do not check writeOnlyHeadOverlapsHorizon, because we do not know of any case where a
+        // write-only head could cause problems with reads in future horizons
```
Mats-SX, line 225: That's a useful observation. I don't know any way for a write-only head to increase cardinality above one, which makes eagerness analysis trivial.

community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/UpdateGraph.scala
```diff
@@ -149,26 +150,30 @@ trait UpdateGraph {
...
         case p: Property => p.propertyKey
       }.toSet
 
-      setNodePropertyOverlap(propertiesReadInHorizon)
-    }
+      setNodePropertyOverlap(propertiesReadInHorizon) || setRelPropertyOverlap(propertiesReadInHorizon)
+    } || {
+      (labelsToSet.nonEmpty || removeLabelPatterns.nonEmpty) &&
+        horizon.dependingExpressions.exists {
+          case f: FunctionInvocation if f.function.get == Labels => true
```
Mats-SX, line 164: I'm wondering whether `get` here could ever throw exception?

Mats-SX (approved):
Looks good!

community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/logical/Eagerness.scala
henriknyman: Right, that explanation is more on point

community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/UpdateGraph.scala
henriknyman: I hope not :)

PR #8577, opened 2017-01-04
Small improvements

davidegrohmann:
Builds on #8568


chrisvest:
@davidegrohmann I've merged #8568

davidegrohmann:
@chrisvest rebased

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
```diff
@@ -1006,15 +1006,10 @@ public RECORD newRecord()
     @Override
     public RECORD getRecord( long id, RECORD record, RecordLoad mode )
     {
-        // Mark the record with this id regardless of whether or not we load the contents of it.
-        // This is done in this method since there are multiple call sites and they all want the id
-        // on that record, so it's to ensure it isn't forgotten.
-        record.setId( id );
         long pageId = pageIdForRecord( id );
-        int offset = offsetForId( id );
```
chrisvest: The reason these two were kept together was because they have the same formula, except page id uses divide and offset uses modulo. At the assembly level, the division instruction computes both the division and the modulo at the same time, so by keeping the two together, the optimiser can do both calls with a single division instruction. This is a useful optimisation because division is pretty expensive.

You can actually still keep them together if you move both into the `readIntoRecord` method, and pass `0` as the page id to `storeFile.io()`. This works because `readIntoRecord` calls `cursor.next( pageId )` anyway, so the page id passed to `io` isn't used.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/CommonAbstractStoreBehaviourTest.java
```diff
@@ -257,6 +258,60 @@ public void pageCursorErrorsMustNotLingerInRecordCursor() throws Exception
...
+        assertEquals( record43, cursor.get() );
+
+        IntRecord record = new IntRecord( -1 );
+        assertTrue( cursor.next( idOnAnotherPage, record, NORMAL ) );
+        assertEquals( record43, record );
+
+        // next with id does not affect the old pointer either, so 42 is read now
+        assertTrue( cursor.next() );
+        assertEquals( record42, cursor.get() );
+    }
```
chrisvest, line 286: 👍

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/AllRelationshipIterator.java
```diff
@@ -38,4 +44,25 @@
...
+    {
+        while ( currentId <= highId )
+        {
+            try
+            {
+                store.getRecord( currentId, record, RecordLoad.CHECK );
+                if ( record.inUse() )
+                {
+                    return next( record.getId() );
+                }
```
chrisvest, line 59: Should this not also be implemented in terms of `store.isInUse( currentId )` ?

community/kernel/src/main/java/org/neo4j/kernel/api/txstate/TransactionCountingStateVisitor.java
```diff
@@ -68,35 +69,27 @@ public void visitCreatedNode( long id )
...
+                for ( int labelId : labelIds )
+                {
+                    updateRelationshipsCountsFromDegrees( degree.type(), labelId, -degree.outgoing(),
+                            -degree.incoming() );
                 }
-            }
-        }
+            } );
+
+        } );
```
chrisvest: Stylistically, this might be more approachable if the outer-most lambda was turned into a reference to a private helper method.

community/kernel/src/main/java/org/neo4j/kernel/api/txstate/TransactionCountingStateVisitor.java
```diff
@@ -141,30 +133,18 @@ public void visitNodeLabelChanges( long id, final Set<Integer> added, final Set<
...
-                    }
+                    updateRelationshipsCountsFromDegrees( degree.type(), label, degree.outgoing(), degree.incoming() );
                 }
-            }
+                for ( Integer label : removed )
+                {
+                    updateRelationshipsCountsFromDegrees( degree.type(), label, -degree.outgoing(),
+                            -degree.incoming() );
+                }
+            } ) );
```
chrisvest, line 141: Same here, where two lambda clauses are started on the same line.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
```diff
@@ -26,104 +26,104 @@
...
     {
         this.entityReadOperations = entityReadOperations;
         this.relIdAction = relIdAction;
     }
 
     void lockAllNodesAndConsumeRelationships( long nodeId, final KernelStatement state ) throws KernelException
     {
+        boolean retry = true;
         nodeIds.add( nodeId );
         while ( retry )
```
chrisvest: This could be a do-while, I think.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
```diff
@@ -26,104 +26,104 @@
...
-                        PrimitiveLongIterator iterator = nodeIds.iterator();
-                        while ( iterator.hasNext() )
-                        {
-                            state.locks().optimistic().releaseExclusive( ResourceTypes.NODE, iterator.next() );
-                        }
-                        nodeIds.clear();
-                        break;
+                        retry = performAction( state, relationships.get(), first );
+                        first = false;
                     }
```
chrisvest, line 76: Very nice. These changes makes it much easier to see what's going on.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLockingTest.java
```diff
@@ -204,7 +206,42 @@ public long next()
...
+                {
+
+                }
+            };
+            if ( !cursor.next() )
+            {
+                throw new EntityNotFoundException( EntityType.NODE, nodeId );
+            }
+            return cursor;
+        } );
```
chrisvest, line 244: Too much logic in these mocks :( These tests should have been written with stubs, and in terms of observable behaviour from the start.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/state/LabelTransactionStateTest.java
```diff
@@ -362,23 +356,21 @@ private void commitLabels( Integer... labels ) throws Exception
...
     {
-        try ( Cursor<NodeItem> cursor = txContext.nodeCursorById( state, nodeId ) )
-        {
-            if ( cursor.next() )
-            {
-                assertEquals( asSet( labels ), PrimitiveIntCollections.toSet( cursor.get().getLabels() ) );
-            }
-        }
+        txContext.nodeCursorById( state, nodeId ).forAll( node -> assertEquals( asSet( labels ),
+                node.labels().mapReduce( new HashSet<>(), IntSupplier::getAsInt, this::addToCollection ) ) );
```
chrisvest: Maybe we want a `collect` method to simplify these uses. Looks like this has come up a couple of times now.

Is it intentional that you are now no longer collecting into a primitive collection?

community/primitive-collections/src/main/java/org/neo4j/cursor/RawCursor.java
```diff
@@ -50,11 +52,33 @@
...
+            return current;
+        }
+        finally
+        {
+            close();
+        }
+    }
+
+    default <E> void mapForAll( Function<T,E> function, Consumer<E> consumer ) throws EXCEPTION
+    {
```
chrisvest: I don't know if you have any long term plans for this interface, but if we implemented `Spliterator` for `RawCursor`, we could give it a `stream()` method, and access the power of the streams framework.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -380,15 +383,40 @@ public RelationshipIterator nodeGetRelationships( long nodeId, Direction directi
...
+            { // this has to be done outside the inner loop, otherwise we'd never accept a single one...
+                types[unique++] = types[i];
+            }
+        }
+        if ( unique < types.length )
+        {
+            types = Arrays.copyOf( types, unique );
+        }
+        return types;
+    }
```
chrisvest, line 411: We could alternatively sort the types, and do the uniqueness check in a single pass. _Iff_ it could help us to know that the types are sorted up front, then this would be a good idea. I haven't checked if this premise is true, but it sounds like a logical thing.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/DegreeCounter.java
```diff
@@ -0,0 +1,104 @@
...
+        {
+            long firstOut = groupRecord.getFirstOut();
+            long firstIn = groupRecord.getFirstIn();
+            return countByFirstPrevPointer( firstOut, cursor, nodeRecord, relationshipRecord ) +
+                    countByFirstPrevPointer( firstIn, cursor, nodeRecord, relationshipRecord ) + loopCount;
+        }
+        default:
+            throw new IllegalArgumentException( direction.name() );
+        }
+    }
```
chrisvest, line 107: Maybe you want these methods to be static instead of default? They don't really rely on any methods in this or super interfaces. Default methods are primarily for adding methods to existing interfaces in a backwards compatible way. They're also harder for the JIT to optimise, because of their complicated resolution logic.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/DegreeItemCursor.java
```diff
@@ -0,0 +1,92 @@
...
+    {
+        return incoming;
+    }
+
+    @Override
+    public DegreeItem get()
+    {
+        if ( keys == null )
+        {
+            throw new IllegalStateException();
```
chrisvest: Perhaps mention that the cursor has been accessed after it's been closed.

chrisvest (commented):
Please take a look at the comments I have.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
davidegrohmann: I'll do that

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/AllRelationshipIterator.java
davidegrohmann: That would be nice, but you need the record in `relationshipVisit` which is called on each element in the iterator.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/state/LabelTransactionStateTest.java
davidegrohmann: Yeah maybe a `collect` makes sense.

We can use a primitive collection but I didn't think it was important in a test.

community/primitive-collections/src/main/java/org/neo4j/cursor/RawCursor.java
davidegrohmann: I didn't explore that possibility. I might give it a try.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
davidegrohmann: Yep that might an improvement, I simply moved that code as is from another class. No changes to the logic there.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/DegreeCounter.java
davidegrohmann: I'll do that.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/DegreeItemCursor.java
davidegrohmann: sure

davidegrohmann: Actually key is set to null also when a next value cannot be found.

PR #8578, opened 2017-01-04
Remove wrapping by ConstantScoreQuery, because all MultiTermQuery ins…

uschindler:
…tances are constant score by default. Refactor DocWithId a bit (and use simpler ThreadLocal).

This is just a small optimization. I am still not sure what the strange handling of the empty strings in newRangeSeekByStringQuery should do. There is also no test for it! I just changed this to be ConstantScore, too (to be correct). All inner clauses are now Occur.FILTER and Occur.MUST_NOT (but Lucene would rewrite this anyways).

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
@@ -62,14 +62,7 @@ public class LuceneDocumentStructure
 
     public static final String NODE_ID_KEY = "id";
 
-    private static final ThreadLocal<DocWithId> perThreadDocument = new ThreadLocal<DocWithId>()
-    {
-        @Override
-        protected DocWithId initialValue()
-        {
-            return new DocWithId( NODE_ID_KEY );
-        }
-    };
+    private static final ThreadLocal<DocWithId> perThreadDocument = ThreadLocal.withInitial( DocWithId::new );
 
     private LuceneDocumentStructure()
     {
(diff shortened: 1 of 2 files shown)


uschindler:
What was the problem with tests? On my machine, all Lucene-Module ones passed.

MishaDemianenko:
@uschindler all lucene related tests passed, one of the cluster fluky test failure. I re-trigger corresponding build

MishaDemianenko:
@uschindler All those queries where wrapped before into `ConstantScoreQuery` to avoid score evaluation to make execution as fast as possible.
Also tests for that quite sophisticated newRangeSeekByStringQuery are at DatabaseIndexAccessorTest::indexStringRangeQuery

MishaDemianenko:
Will gonna start perf run on top of this branch will come back as soon as that will finish

uschindler:
MultiTermQuery and all its subclasses (TermRange, NumericRange, Wildcard, Prefix, including your special prefix query) is constant score anyways, as it always rewrites being constant score, ignoring scores:

- https://github.com/apache/lucene-solr/blob/branch_5_5/lucene/core/src/java/org/apache/lucene/search/MultiTermQuery.java#L67
- https://github.com/apache/lucene-solr/blob/branch_5_5/lucene/core/src/java/org/apache/lucene/search/MultiTermQuery.java#L81-L98
- https://github.com/apache/lucene-solr/blob/branch_5_5/lucene/core/src/java/org/apache/lucene/search/MultiTermQueryConstantScoreWrapper.java

Wrapping with a separate CSQ is not needed, as scores are not needed. BTW: There is an easier way to disable scores already on the top level: Just return the correct setting in your collector! This changed around Lucene 5.2, maybe the reason why you think it might still think that it calculate scores comes from that fact.

PR #8579, opened 2017-01-04
Include specific connection details in ListQueries

thobe:
Sample output from `dbms.listQueries`:

```
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| queryId   | username | query                         | parameters | startTime                       | elapsedTime    | elapsedTimeMillis | connectionDetails    | requestScheme | clientAddress | requestURI | cpuTimeMillis | status                                                                                  | waitTimeMillis | metaData |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| "query-4" | ""       | "CALL dbms.listQueries"       | {}         | "2017-01-04T17:50:08.925+01:00" | "00:00:00.001" | 1                 | "embedded-session  " | <null>        | <null>        | <null>     | 0             | {state -> "RUNNING"}                                                                    | 0              | {}       |
| "query-1" | ""       | "MATCH (n) SET n.v = n.v + 1" | {}         | "2017-01-04T17:50:05.271+01:00" | "00:00:03.659" | 3659              | "embedded-session  " | <null>        | <null>        | <null>     | 2515          | {waitTimeMillis -> 247, state -> "WAITING", resourceType -> "NODE", resourceIds -> [0]} | 247            | {}       |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

For `bolt` or `http` sessions: `requestScheme`, `clientAddress` and `requestURI` will contain useful information.

As of this, `connectionDetails` is a deprecated field, since `requestScheme`, `clientAddress` and `requestURI` contains the same information in a more structured way.


enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryStatusResult.java
```diff
@@ -43,10 +43,22 @@
     public final Map<String,Object> parameters;
     public final String startTime;
     public final String elapsedTime;
+    /** EXPERIMENTAL: added in Neo4j 3.2 */
```
spacecowboy: Do these comments add some kind of value? Shouldn't the experimental nature be properly documented elsewhere? Somewhere user facing that is. Git blame can already inform an interested dev that it was added in 3.2.

technige:
Very minor point but capitalisation is a little inconsistent between `queryId` and `requestURI`. Although I personally dislike the form, perhaps `requestUri` would be more consistent?

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/cypher/StatementProcessor.java
```diff
@@ -38,5 +39,8 @@
 
     boolean hasTransaction();
 
-    void setQuerySource( String querySource );
+    // NOTE: if this interface moved to the parent package, BoltQuerySource could be package-private
+    // in fact, this entire `cypher` package does not really seem to server a special purpose, and could be merged
+    // with its parent package.
```
technige: +1

technige:
I have previously removed the `Session` term from server-side code since a Session is (now) a purely client-side concept. The server deals with a `Connection` instead and has no visibility of client-side sessions. This change reintroduces several classes that use the `Session` term.

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/QueryStatusResult.java
praveenag: We will be documenting the experimental fields separately as part of procedure docs. We wanted to release this in a milestone to get some feedback and have the ability to modify based on feedback. We will remove this once we finish with the implementation/ in the next major release.

Dcanzano:
curious what happens with a idle txn/session.   if for example I start a begin-comitt transaction, for example run the following

line1:  Begin
line2:  create (n:Person {id:1})

and then walk away for 10 minutes..   what would listQueries return.   Admittedly the transaction is in an 'open' state since I have yet to commit.   But it does hold a read lock on the :Person node.    and yet call dbms.listQueries does not show it.      If I'm trying to debug why my other query is not making any progress (since its blocked by the read lock) how would i diagnose this.

Any opportunity for listQueries to show idle txns?

spacecowboy:
@Dcanzano I might be mistaken, but `listQueries` is listing active queries, not transactions. Your create statement has already finished, so there is nothing to list here.

Do you see value in a `listTransactions` procedure? What would it display in your opinion?

thobe:
@technige There are plenty of things with the word session left in the server codebase, but for few types introduced by this PR I think substituting the word session for the word connection would work just as well.

On the topic of `requestURI` vs `requestUri` I don't really care, if you think that `requestUri` would be more consistent with existing fields, then that is enough of an opinion to change it.

Does that `+1` you added mean that you think we should nuke the `org.neo4j.bolt.v1.runtime.cypher` package?

thobe:
@Dcanzano just like @spacecowboy already said, what you are asking for is new and different functionality, unrelated to listing queries. I agree with you that it would be beneficial to also have the ability to list transactions, their state and the resources tied to them.

Although, to be picky, the query in your example would not be a problem, since the lock it holds is on a node that is not visible to any other transaction.

thobe:
I believe those updates address the comments above. Except for the comment from @spacecowboy which is to be handled in another repository (maybe? - @jjaderberg what is the best way to document the output of procedures?)

community/shell/src/main/java/org/neo4j/shell/kernel/apps/cypher/Start.java
```diff
@@ -208,9 +209,9 @@ private TransactionalContext createTransactionContext( String queryText, Map<Str
 
     private static class ShellQuerySession
```
praveenag: We can replace this class with `ShellConnectionInfo`

thobe:
The docs repository needs to be updated after this PR has been merged, there is a dependency on `QuerySource` in the docs build, which this PR renames to `ClientConnectionInfo`. (Part of the PR build will be red because of this, that is ok).

@praveenag will take care of updating the docs repo.

PR #8581, opened 2017-01-05
GBPTree test for recovery replaying transactions from before last checkpoint

burqen:
To make sure nothing breaks if more transactions then necessary is replayed.


MishaDemianenko:
@burqen will merge on green

PR #8582, opened 2017-01-05
Fix a bug in the schema index partitioning (maxDoc vs. numDocs)

uschindler:
There is a bug in the schema index partitioning that could lead to problems adding/updating Lucene documents. The problem is the following:

Lucene's docIds are 32 bits (so we need partitioning) and those are assigned using the IndexReader's maxDoc() property. maxDoc() is generally larger than numDocs() which takes deleted documents into account. As deleted documents are not removed asap (it may take very long time until they are deleted, sometimes they are never deleted, especially because you are using the "outdated" LogByteSizeMergePolicy, which never removes deletes from huge segments unless they get merged by other reasons!!! -> TieredMergePolicy merges segments once they have > 10% deletes).

The problem is that a document addition or document update (delete + add) consumes one document ID, so IndexReaders/IndexWriter's maxDoc raises by one. Once maxDoc() reaches IndexWriter.MAX_DOC (slightly smaller than Integer.MAX_VALUE), indexing will break and stop with IllegalState/IllegalArgumentException.

The partitioning code currently uses numDocs() to find a segment with enough space. This is wrong, because it does not count deleted documents, which also consume document IDs. The bug does not affect most users, because there is a buffer of 10% included, but once a segment has more than 10% deletes, the whole thing will break.

This PR uses IndexWriter.maxDoc() for the calculation. It also changes the getIndexWriter() method to require the number of documents to be added (generally 1, but for bulk updates there are more required). 

For concurrency reason it still adds the buffer of 10%, because of incomplete synchronization: At the time when a thread requests an IndexWriter that is able to accept the number of document adds/updates to write, it does not "reserve those document ids", so another thread could do the same and then on addDocument/updateDocument it will break. So I left the buffer of 10%. To fix this correctly (and allow the partitions to get maximum size), one would need to completely synchronize addition of documents (bad idea), or use some AtomicInteger for each partition to "reserve" document ids before the actual document is added.

I figured out that the labelscan index seems to have the same problem, but I don't understand the partitioning there. It also has not 10% buffer, so i don't understand why it did not break at some point. Maybe because it never does deletes or updates (although it uses IndexWriter.updateDocument)?

Further steps would be to use TieredMergePolicy instead of outdated LogByteSizeMergePolicy - but this is unrelated to this issue.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/writer/PartitionedIndexWriter.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/writer/PartitionedIndexWriter.java
@@ -34,7 +34,7 @@ import org.neo4j.kernel.api.impl.schema.WritableDatabaseSchemaIndex;
  * Schema Lucene index writer implementation that supports writing into multiple partitions and creates partitions
  * on-demand if needed.
  * <p>
- * Writer threats partition as writable if partition has number of documents that is less then configured
+ * Writer threats partition as writable if partition has number of live and deleted documents that is less then configured
  * {@link #MAXIMUM_PARTITION_SIZE}.
  * First observable partition that satisfy writer criteria is used for writing.
  */
@@ -42,8 +42,10 @@ public class PartitionedIndexWriter implements LuceneIndexWriter
 {
     private final WritableDatabaseSchemaIndex index;
 
+    // by default we still keep a spare of 10% to the maximum partition size: During concurrent updates
+    // it could happen that 2 threads reserve space in a partition (without claiming it by doing addDocument):
     private final Integer MAXIMUM_PARTITION_SIZE = Integer.getInteger( "luceneSchemaIndex.maxPartitionSize",
-            Integer.MAX_VALUE - (Integer.MAX_VALUE / 10) );
+            IndexWriter.MAX_DOCS - (IndexWriter.MAX_DOCS / 10) );
 
     public PartitionedIndexWriter( WritableDatabaseSchemaIndex index ) throws IOException
     {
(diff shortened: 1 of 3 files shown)


uschindler:
I fixed a violation of the Iterable contract in LuceneIndexPopulator. Nevertheless, cool idea to make use of the thread local Document instances!

uschindler:
What is the problem with this PR?

MishaDemianenko:
So far flaky test failures unrelated to content of this PR, re-trigger another chain of builds

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/writer/PartitionedIndexWriter.java
```diff
@@ -34,16 +34,18 @@
...
- * Writer threats partition as writable if partition has number of documents that is less then configured
+ * Writer threats partition as writable if partition has number of live and deleted documents that is less then configured
  * {@link #MAXIMUM_PARTITION_SIZE}.
  * First observable partition that satisfy writer criteria is used for writing.
  */
 public class PartitionedIndexWriter implements LuceneIndexWriter
 {
     private final WritableDatabaseSchemaIndex index;
 
+    // by default we still keep a spare of 10% to the maximum partition size: During concurrent updates
```
MishaDemianenko, line 45: We keep this big (comparatively) gap to avoid synchronisation in here. It's fine if we will expose a writer even in quota is reached or almost reached. Since population will gonna happen in batches and if couple of them will gonna squeeze is, it's fine. Thats why we can avoid any atomics or complete sync.

MishaDemianenko:
Changes looks reasonable, merging.

uschindler:
How should we proceed with the other partitioning scheme?:

> I figured out that the labelscan index seems to have the same problem, but I don't understand the partitioning there. It also has not 10% buffer, so i don't understand why it did not break at some point. Maybe because it never does deletes or updates (although it uses IndexWriter.updateDocument)?

Can you explain how this works and what it does. The labelscan stuff is a bit complicated to understand for me.

This is also a "bug", so I propose to add the "bug" tag.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/writer/PartitionedIndexWriter.java
uschindler: Yes that's why I left it in.

MishaDemianenko:
As you remember you can find label scan writer at LabelScanWriter and PartitionedLuceneLabelScanWriter is the most important implementation. 
Unlike other schema indexes label scan indexes use completely separate document format represented by BitmapDocumentFormat.
Each document contain range field (based on node id) and number of fields with bitmap values for different node labels in that range.
PartitionedLuceneLabelScanWriter will try to group updates by range and apply them once to whole range (so if you have multiple nodes from the same range affected those updates will be grouped together).
As you already noticed PartitionedLuceneLabelScanWriter::flush will gonna do document update when it will detect that particular document for certain range was not empty before update.
I would assume that the only reason why it's not failing with mentioned problem is because you need to have quite quite big number of nodes/updates to overflow the whole index. 

To get more clear picture how it looks like you can try to use one of the IT tests: lets take LabelScanStoreStartupIT::scanStoreRecreateCorruptedIndexOnStartup for example. Its will use quite small db that can help to understand how that mechanics is working you can put a breakpoint in PartitionedLuceneLabelScanWriter::write( NodeLabelUpdate update ) or PartitionedLuceneLabelScanWriter::flush to see how that is working.

uschindler:
OK, I will look into the LabelScanWriter and look into changing the partitioning schema a bit. Currently it uses some "div/modulo" approach to delegate to the right partition - so it is not dynamic and it may overflow if the partition has too many deletes. Have to debug through this.

PR #8584, opened 2017-01-06
Hardens GB+Tree testing

tinwelint:
By using adversarial page cache and an added access-checking page cache.

Also contains fixes for exposed issues for the added testing.

### DIFF
--- a/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/StandalonePageCacheFactory.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/StandalonePageCacheFactory.java
@@ -34,22 +34,27 @@ import org.neo4j.io.pagecache.tracing.PageCacheTracer;
  */
 public final class StandalonePageCacheFactory
 {
-
     private StandalonePageCacheFactory()
     {
     }
 
     public static PageCache createPageCache( FileSystemAbstraction fileSystem )
     {
-        return createPageCache( fileSystem, PageCacheTracer.NULL );
+        return createPageCache( fileSystem, null, PageCacheTracer.NULL );
     }
 
-    public static PageCache createPageCache( FileSystemAbstraction fileSystem, PageCacheTracer tracer )
+    public static PageCache createPageCache( FileSystemAbstraction fileSystem, Integer pageSize )
+    {
+        return createPageCache( fileSystem, pageSize, PageCacheTracer.NULL );
+    }
+
+    public static PageCache createPageCache( FileSystemAbstraction fileSystem, Integer pageSize,
+            PageCacheTracer tracer )
     {
         SingleFilePageSwapperFactory factory = new SingleFilePageSwapperFactory();
         factory.setFileSystemAbstraction( fileSystem );
 
-        int cachePageSize = factory.getCachePageSizeHint();
+        int cachePageSize = pageSize != null ? pageSize : factory.getCachePageSizeHint();
         long pageCacheMemory = ByteUnit.mebiBytes( 8 );
         long pageCount = pageCacheMemory / cachePageSize;
         return new MuninnPageCache( factory, (int) pageCount, cachePageSize, tracer );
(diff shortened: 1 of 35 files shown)


community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/StandalonePageCacheFactory.java
```diff
@@ -34,22 +34,27 @@
...
+        return createPageCache( fileSystem, null, PageCacheTracer.NULL );
+    }
+
+    public static PageCache createPageCache( FileSystemAbstraction fileSystem, Integer pageSize )
+    {
+        return createPageCache( fileSystem, pageSize, PageCacheTracer.NULL );
     }
 
-    public static PageCache createPageCache( FileSystemAbstraction fileSystem, PageCacheTracer tracer )
+    public static PageCache createPageCache( FileSystemAbstraction fileSystem, Integer pageSize,
```
ragadeeshu, line 51: Not too happy about this public method that does null logic on an Integer argument without any comment, I would like one of these to be fixed; for example add a comment, or add more constructors so that you can hide your null-magic in this class.

ragadeeshu (changes requested):
Looks good overall. I have a comment about the changes made to `StandalonePageCacheFactory`. It would also be great if @chrisvest could take a look at `4135a84416c1ab0b0ae0f8b413ef1d6a80cc0248` and how it fits into the page cache surface.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/StandalonePageCacheFactory.java
chrisvest: I agree with this sentiment. You could add another static constructor that takes a `PageSwapperFactory` and a page size, and then move the construction of the default page swapper factory up one level in the stack. The creation & configuration of this page swapper factory can also be extracted into a static helper method. That way you would not need any null-logic at all.

tinwelint: Very much agree

burqen:
@ragadeeshu Something like this?

ragadeeshu (approved):

tinwelint:
Tools compilation error seems to be unrelated and matches recent changes around configuration refactorings in neo4j

community/kernel/src/test/java/org/neo4j/kernel/impl/store/CommonAbstractStoreTest.java
```diff
@@ -193,7 +193,8 @@ public void recordCursorPinsEachPageItReads() throws Exception
     {
         File storeFile = dir.file( "a" );
         RecordingPageCacheTracer tracer = new RecordingPageCacheTracer( Pin.class );
-        PageCache pageCache = pageCacheRule.getPageCache( fileSystemRule.get(), tracer, Config.empty() );
+        PageCache pageCache = pageCacheRule.getPageCache( fileSystemRule.get(),
+                config().withTracer( tracer ), Config.empty() );
```
spacecowboy, line 197: It is true that the Tools compile failure is unrelated but you still have some merge issues. `Config.empty()` no longer exists and the PR builder won't tell you that because it is building your branch and not the merge result. So a rebase/merge is still required.

spacecowboy (changes requested):

community/kernel/src/test/java/org/neo4j/kernel/impl/store/CommonAbstractStoreTest.java
burqen: Nice catch!

tinwelint: Although it looks like [Config.empty()](https://github.com/neo4j/neo4j/blob/3.2/community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java#L87) still exists, and this rebased branch compiles and is generally happy now. Thanks for the heads up

spacecowboy (approved):

PR #8585, opened 2017-01-09
Adds test for GBPTree which auto-detects format changes

tinwelint:
this test works by having a committed current-format.zip in src/test/resources
and on failure reading and consistency checking that tree notifies developer that
a FORMAT_VERSION needs to be incremented as well. When developer has incremented
this version a new current-format.zip is generated and must be committed together
with the other changes which changed the format.


community/index/src/main/java/org/neo4j/index/gbptree/GBPTree.java
```diff
@@ -288,13 +288,21 @@ public GBPTree( PageCache pageCache, File indexFile, Layout<KEY,VALUE> layout, i
...
+            {
+                loadState( pagedFile );
+            }
         }
-        else
+        catch ( Throwable e )
         {
-            loadState( pagedFile );
+            close();
+            throw e;
```
chrisvest: If `close()` throws, for whatever reason, then that exception should be suppressed instead of shadowing the original exception.

PR #8586, opened 2017-01-10
Store migration support for block devices

ragadeeshu:
Make store migration work for custom IO integrations, by accessing store files through the page cache. Including for file operations such as deletes, copying and moving files.

This work includes a new `StoreMigrationIT` test, which searches for all available store formats, and tests migration for all supported from-to format tuples. This test is used in the block device integration code to verify that migration works there as well.

The page cache based file operations only take effect for formats from 2.3 and onward.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
@@ -32,7 +32,6 @@ import org.neo4j.io.fs.FileSystemAbstraction;
 import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.impl.storemigration.monitoring.MigrationProgressMonitor;
 import org.neo4j.kernel.impl.storemigration.monitoring.MigrationProgressMonitor.Section;
-import org.neo4j.kernel.impl.util.CustomIOConfigValidator;
 import org.neo4j.kernel.internal.Version;
 import org.neo4j.logging.Log;
 import org.neo4j.logging.LogProvider;
@@ -116,7 +115,6 @@ public class StoreUpgrader
             throw new UpgradeNotAllowedByConfigurationException();
         }
 
-        CustomIOConfigValidator.assertCustomIOConfigNotUsed( config, CUSTOM_IO_EXCEPTION_MESSAGE );
 
         // One or more participants would like to do migration
         progressMonitor.started();
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
@@ -24,11 +24,15 @@ import java.io.FilenameFilter;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.regex.Pattern;
 
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
 import org.neo4j.helpers.Exceptions;
 import org.neo4j.io.fs.FileSystemAbstraction;
+import org.neo4j.io.pagecache.FileHandle;
+import org.neo4j.io.pagecache.PageCache;
+import org.neo4j.io.pagecache.PagedFile;
 import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.impl.storemigration.monitoring.MigrationProgressMonitor;
 import org.neo4j.kernel.impl.storemigration.monitoring.MigrationProgressMonitor.Section;
@@ -66,22 +70,23 @@ public class StoreUpgrader
     public static final String MIGRATION_DIRECTORY = "upgrade";
     public static final String MIGRATION_LEFT_OVERS_DIRECTORY = "upgrade_backup";
     private static final String MIGRATION_STATUS_FILE = "_status";
-    public static final String CUSTOM_IO_EXCEPTION_MESSAGE = "Store upgrade not allowed with custom IO integrations";
 
     private final UpgradableDatabase upgradableDatabase;
     private final MigrationProgressMonitor progressMonitor;
     private final List<StoreMigrationParticipant> participants = new ArrayList<>();
     private final Config config;
     private final FileSystemAbstraction fileSystem;
+    private final PageCache pageCache;
     private final Log log;
 
     public StoreUpgrader( UpgradableDatabase upgradableDatabase, MigrationProgressMonitor progressMonitor, Config
-            config, FileSystemAbstraction fileSystem, LogProvider logProvider )
+            config, FileSystemAbstraction fileSystem, PageCache pageCache, LogProvider logProvider )
     {
         this.upgradableDatabase = upgradableDatabase;
         this.progressMonitor = progressMonitor;
         this.fileSystem = fileSystem;
         this.config = config;
+        this.pageCache = pageCache;
         this.log = logProvider.getLog( getClass() );
     }
 
@@ -114,7 +119,6 @@ public class StoreUpgrader
         {
             throw new UpgradeNotAllowedByConfigurationException();
         }
-
 
         // One or more participants would like to do migration
         progressMonitor.started();
@@ -254,6 +258,18 @@ public class StoreUpgrader
 
     private void cleanMigrationDirectory( File migrationDirectory )
     {
+        try
+        {
+            Iterable<FileHandle> fileHandles = pageCache.streamFilesRecursive( migrationDirectory )::iterator;
+            for ( FileHandle fh : fileHandles )
+            {
+                fh.delete();
+            }
+        }
+        catch ( IOException e )
+        {
+            // This means that we had no files to clean, this is fine.
+        }
         if ( migrationDirectory.exists() )
         {
             try
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
@@ -24,7 +24,6 @@ import java.io.FilenameFilter;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Optional;
 import java.util.regex.Pattern;
 
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
@@ -32,7 +31,6 @@ import org.neo4j.helpers.Exceptions;
 import org.neo4j.io.fs.FileSystemAbstraction;
 import org.neo4j.io.pagecache.FileHandle;
 import org.neo4j.io.pagecache.PageCache;
-import org.neo4j.io.pagecache.PagedFile;
 import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.impl.storemigration.monitoring.MigrationProgressMonitor;
 import org.neo4j.kernel.impl.storemigration.monitoring.MigrationProgressMonitor.Section;
@@ -258,6 +256,8 @@ public class StoreUpgrader
 
     private void cleanMigrationDirectory( File migrationDirectory )
     {
+        // We use the page cache here to make sure that the migration directory is clean even if we are using a block
+        // device.
         try
         {
             Iterable<FileHandle> fileHandles = pageCache.streamFilesRecursive( migrationDirectory )::iterator;
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
@@ -260,7 +260,7 @@ public class StoreUpgrader
     {
         try
         {
-            if ( migrationDirectory.exists() )
+            if ( fileSystem.fileExists( migrationDirectory ) )
             {
                 fileSystem.deleteRecursively( migrationDirectory );
             }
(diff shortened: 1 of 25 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
```diff
@@ -466,6 +477,29 @@ private void migrateWithBatchImporter( File storeDir, File migrationDir, long la
...
+            try
+            {
+                Iterable<FileHandle> fileHandles = pageCache.streamFilesRecursive( migrationDir )::iterator;
+                for ( FileHandle fh : fileHandles )
+                {
+                    Predicate<StoreFile> predicate =
+                            storeFile -> storeFile.fileName( StoreFileType.STORE ).equals( fh.getFile().getName() );
+                    if ( storesToDeleteFromMigratedDirectory.stream().anyMatch( predicate ) )
+                    {
+                        final Optional<PagedFile> optionalPagedFile = pageCache.getExistingMapping( fh.getFile() );
```
burqen: I don't see how there can still be a mapping on any file here. All of the paged files should be closed once the import finishes.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
```diff
@@ -256,6 +256,20 @@ private void migrateToIsolatedDirectory( File storeDir, File migrationDirectory,
 
     private void cleanMigrationDirectory( File migrationDirectory )
     {
+        // We use the page cache here to make sure that the migration directory is clean even if we are using a block
+        // device.
+        try
```
burqen, line 261: Here we delete files through `pageCache` first and `fileSystem` second.
In `StoreMigrator` it is the other way around. 
I think it make sense to use the same pattern between the two. Especially since most difficulties around adapting Neo to behave with blockdevice is around file operations.

Maybe consider using `FileHandle.HANDLE_DELETE`?

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
```diff
@@ -256,6 +256,20 @@ private void migrateToIsolatedDirectory( File storeDir, File migrationDirectory,
...
+            Iterable<FileHandle> fileHandles = pageCache.streamFilesRecursive( migrationDirectory )::iterator;
+            for ( FileHandle fh : fileHandles )
+            {
+                fh.delete();
+            }
+        }
+        catch ( IOException e )
+        {
+            // This means that we had no files to clean, this is fine.
+        }
```
burqen: We get this `IOException` from `SingleFilePageSwapperFactory.streamFilesRecursiveInner(...)` if `migrationDirectory` does not exist. I would prefer if we specifically caught the `NoSuchFileException` that it throws instead to not hide other possible `IOExceptions`, for example if some file is already mapped we get `FileIsMappedException`.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
```diff
@@ -466,6 +477,29 @@ private void migrateWithBatchImporter( File storeDir, File migrationDir, long la
...
+                        {
+                            optionalPagedFile.get().close();
+                        }
+                        fh.delete();
+                    }
+                }
+            }
+            catch ( IOException e )
+            {
+                // This means that we had no files only present in the page cache, this is fine.
```
burqen, line 499: Same comment as for `StoreUpgrader`, even though we explicitly `close` mapped files, there could be other `IOException` that we do not want to hide.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
```diff
@@ -520,6 +582,23 @@ private void prepareBatchImportMigration( File storeDir, File migrationDir, Reco
...
+                NullLogProvider.getInstance() );
+        try ( NeoStores neoStores = storeFactory.openAllNeoStores( true ) )
+        {
+            neoStores.getMetaDataStore();
+            neoStores.getLabelTokenStore();
+            neoStores.getNodeStore();
+            neoStores.getPropertyStore();
+            neoStores.getRelationshipGroupStore();
+            neoStores.getRelationshipStore();
+            neoStores.getSchemaStore();
```
burqen, line 595: Are all of these calls here to make us fail fast if some store is present? In that case 👍

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
```diff
@@ -634,6 +714,30 @@ public void moveMigratedFiles( File migrationDir, File storeDir, String versionT
...
+                    {
+                        optionalPagedFile.get().close();
+                    }
+                    fh.rename( new File( storeDir, fh.getFile().getName() ), StandardCopyOption.REPLACE_EXISTING );
+                }
+            }
+        }
+        catch ( IOException e )
+        {
+            //This means that we had no files only present in the page cache, this is fine.
```
burqen, line 736: Same `IOException` comment again.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
```diff
@@ -634,6 +714,30 @@ public void moveMigratedFiles( File migrationDir, File storeDir, String versionT
                 true, // allow to skip non existent source files
                 ExistingTargetStrategy.OVERWRITE, // allow to overwrite target files
                 StoreFileType.values() );
+        // Since some of the files might only be accessible through the page cache (i.e. block devices), we also try to
+        // move the files with the page cache.
```
burqen, line 715: I like this pattern of first trying to finish the job with `FileSystemAbstraction` and then "clean up" what's left with `PageCache`.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
```diff
@@ -634,6 +714,30 @@ public void moveMigratedFiles( File migrationDir, File storeDir, String versionT
...
+        try
+        {
+            Iterable<FileHandle> fileHandles = pageCache.streamFilesRecursive( migrationDir )::iterator;
+            for ( FileHandle fh : fileHandles )
```
burqen, line 719: This block has a lot in common with same block in `migrateWithBatchImporter` when deleting files from `migrationDir`. Maybe the behaviour could be extracted into a helper method by extracting the action (delete, rename) and a `Supplier<Stream<StoreFile>>` to test the predicate for.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
```diff
@@ -357,6 +367,7 @@ private void removeDuplicateEntityProperties( File storeDir, File migrationDir,
             SchemaIndexProvider schemaIndexProvider, RecordFormats oldFormat )
             throws IOException
     {
+        CustomIOConfigValidator.assertCustomIOConfigNotUsed( config, CUSTOM_IO_EXCEPTION_MESSAGE );
```
burqen: Can we please move this assertion to before calling this method and add a comment about it, like we do in `rebuildCounts` so it becomes:
```
if ( versionToMigrateFrom.equals( StandardV2_1.STORE_VERSION ) )
{
    // These versions are not supported for block devices.
    CustomIOConfigValidator.assertCustomIOConfigNotUsed( config, CUSTOM_IO_EXCEPTION_MESSAGE );
    removeDuplicateEntityProperties( storeDir, migrationDir, pageCache, schemaIndexProvider, oldFormat );
}
```

Also, why do we not check for `StandardV2_2.STORE_VERSION` like we do in `rebuildCounts`?

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
```diff
@@ -116,8 +118,6 @@ public void migrateIfNeeded( File storeDirectory)
             throw new UpgradeNotAllowedByConfigurationException();
         }
 
-        CustomIOConfigValidator.assertCustomIOConfigNotUsed( config, CUSTOM_IO_EXCEPTION_MESSAGE );
-
```
burqen, line 120: Instead of removing this from here and check it later in `StoreMigrator`, does it make sense to instead fail as fast as possible here as soon and `versionToMigrateFrom` is known?
I mean inside the `if ( MigrationStatus.X.isNeededFor( migration status ) )` blocks.

community/neo4j/src/test/java/upgrade/PlatformConstraintStoreUpgradeTest.java
```diff
@@ -68,7 +68,7 @@ public void shouldFailToStartWithCustomIOConfigurationTest() throws IOException
         }
         catch ( RuntimeException ex )
         {
-            assertEquals( StoreUpgrader.CUSTOM_IO_EXCEPTION_MESSAGE, ex.getCause().getCause().getMessage() );
+            assertEquals( StoreMigrator.CUSTOM_IO_EXCEPTION_MESSAGE, ex.getCause().getCause().getMessage() );
```
burqen, line 93: Should we not also check for `StandardV2_1.STORE_VERSION` here?

burqen (changes requested):
First review round finished. Have some concerns about Exception handling and some small comments to improve readability.

burqen:
@ragadeeshu Comments ready for you 👍

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
ragadeeshu: These calls are there in order to explicitly create the stores, I think. This method has just been migrated from `BatchingNeoStores` since only this is the only class that calls it.

burqen: Alright, that was my first thought as well but i couldn't see any creation being done in those code paths.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
ragadeeshu: One argument against this is that the `StoreUpgrader` would need to become aware of store versions, instead of that being isolated to `StoreMigrator`

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
ragadeeshu: This, along with consolidating all/most of our file helper classes, should probably be done separately.

community/neo4j/src/test/java/upgrade/PlatformConstraintStoreUpgradeTest.java
ragadeeshu: Fixed, also added 2.0

ragadeeshu:
@burqen Please have a look again :)

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
burqen: I was thinking of extracting it only for this class, but a larger refactoring might be a good idea. If you plan to do this, then nevermind, otherwise feel free to address it for this class in this PR.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
burqen: It's actually `StoreUpgrader` that reads store version and passes it to `StoreMigrator`.

ragadeeshu: @burqen I meant checking against specific versions, importing `StandardV` classes, etc. `StoreUpgrader` is currently free of such logic, while `StoreMigrator` is full of it already. `StoreMigrater` is also the only `Migrator` implementation that is incompatible with custom IO configurations.

burqen: Ah, then would agree that it makes more sense keeping it as it is.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/StoreUpgrader.java
```diff
@@ -256,31 +258,27 @@ private void migrateToIsolatedDirectory( File storeDir, File migrationDirectory,
 
     private void cleanMigrationDirectory( File migrationDirectory )
     {
-        // We use the page cache here to make sure that the migration directory is clean even if we are using a block
-        // device.
         try
         {
-            Iterable<FileHandle> fileHandles = pageCache.streamFilesRecursive( migrationDirectory )::iterator;
-            for ( FileHandle fh : fileHandles )
+            if ( migrationDirectory.exists() )
```
burqen: I missed this during my first review apparently, sorry.
We should work towards using `FileSystemAbstraction` whenever possible, rather than mixing between `FSA`, `FileUtils` and `File#...`. Feel free to use `fileSystem.fileExists( migrationDirectory )` here.

burqen (changes requested):
Second review finished. Changes look good. Approved once I get an answer on the remaining questions.

burqen:
@ragadeeshu Second review finished. There are some test failures that seem related to these changes as well.

burqen (approved):

PR #8588, opened 2017-01-10
Page cache counters for separate transaction 

MishaDemianenko:
Previously page cache counters and metrics like number of flushes, bytes read etc was global and there was no possibility to find out in what way particular transaction or query influence it.

This PR will splits tracing into page cache global and thread local tracing of page cursors.
Also making thread local information available for consuming by any interesting parties to be able to report any transaction based page cache counters.
In opposed to previous behaviour when all events where reported directly to global tracer,  in new model  events will be reported first to thread local tracer and then eventually those will be propagated to global tracer.


chrisvest:
@MishaDemianenko Needs a rebase.

MishaDemianenko:
@chrisvest done

tinwelint:
You could rebase this PR to hopefully get it green.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -72,16 +71,18 @@
     // offending code.
     private Object cursorException;
 
-    MuninnPageCursor( long victimPage )
+    private final PageCursorTracer pageCursorTracer;
```
chrisvest: Put it next to the `victimPage` field.

chrisvest: It can also still be called `tracer`, for a shorter name.

community/io/src/main/java/org/neo4j/io/pagecache/tracing/PageCacheTracer.java
```diff
@@ -170,4 +199,20 @@ public String toString()
...
+
+    void bytesRead( long bytesRead );
+
+    void evictions( long evictions );
+
+    void evictionExceptions( long evictionExceptions );
+
+    void bytesWritten( long bytesWritten );
+
+    void flushes( long flushes );
```
chrisvest: Give them a bit of javadocs.

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/DefaultPageCursorTracer.java
```diff
@@ -0,0 +1,295 @@
...
+    private long faults;
+    private long bytesRead;
+    private long bytesWritten;
+    private long evictions;
+    private long evictionExceptions;
+    private long flushes;
+
+    private PageCacheTracer pageCacheTracer;
+
+    public void init( PageCacheTracer pageCacheTracer )
```
chrisvest: This needs an `@Override` doesn't it?

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/DefaultPageCursorTracer.java
```diff
@@ -0,0 +1,295 @@
...
+    {
+        this.pageCacheTracer = pageCacheTracer;
+    }
+
+    public void reportEvents()
```
chrisvest, line 51: Some javadocs would be nice, to clarify what this does and doesn't do.

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/PageCursorTracer.java
```diff
@@ -0,0 +1,112 @@
...
+        {
+
+        }
+    };
+
+    PinEvent beginPin( boolean writeLock, long filePageId, PageSwapper swapper );
+
+    void init( PageCacheTracer tracer );
+
+    void reportEvents();
```
chrisvest: I see, _this_ is where the javadoc is supposed to be.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -131,6 +132,7 @@ public final void close()
             if ( cursor.pagedFile != null )
             {
                 cursor.unpinCurrentPage();
+                cursor.pageCursorTracer.reportEvents();
```
chrisvest: I'm not entirely convinced that this is the right place to put it. Cypher wants to get numbers for a whole Statement, in which case it makes sense to collect and report the data when a statement is closed. Linked cursors would presumably also be attached to the same thread-local tracer, in which case there is no point to do the reporting several times.

chrisvest (changes requested):
Had a few comments, but in particular I think we need to take a closer look at where the cursor tracer reporting is done.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
MishaDemianenko: There are couple of reasons why i did it like that:
- if statement is quite big and long running we will not gonna be able to see any page cache counters update and that will make tracing and understanding current situation non clear and hard to trace since all counters will be reported only once in the end when statement will be closed.
- currently code is really bad on closing statement (that is a some code path problem and should be fixed)

Agree on the point of linked cursors: if we will assume that our page cursors tracers are thread local and all of the cursors have the same tracer we should avoid doing reporting multiple times.
My thinking was a bit different - local counters in cursor remember what was done so far in  each separate tracer, so we can get statement dependent info now.

But agree your proposal will make code a bit simpler and faster even if two mentioned above points will not gonna be solved, but we can come back to them later if that will be a problem. Will do like you suggest.

MishaDemianenko:
@chrisvest updated and another benchmark run triggered to see where we in terms of performance with this update.

chrisvest (approved):

chrisvest:
@MishaDemianenko Very nice.

PR #8589, opened 2017-01-10
List all locks held by a transaction/query

thobe:
Adds a procedure called dbms.listActiveLocks(queryId) that lists all locks held by the transaction that executes the query with the given id.

Sample output from `CALL dbms.listQueries() YIELD queryId CALL dbms.listActiveLocks(queryId) YIELD mode, resourceType, resourceId RETURN *`:
```
+-----------------------------------------------------+
| mode        | queryId   | resourceId | resourceType |
+-----------------------------------------------------+
| "EXCLUSIVE" | "query-1" | 0          | "NODE"       |
| "EXCLUSIVE" | "query-1" | 1          | "NODE"       |
| "EXCLUSIVE" | "query-1" | 2          | "NODE"       |
| "EXCLUSIVE" | "query-1" | 3          | "NODE"       |
| "EXCLUSIVE" | "query-1" | 4          | "NODE"       |
| "SHARED"    | "query-1" | 0          | "SCHEMA"     |
| "SHARED"    | "query-3" | 0          | "SCHEMA"     |
+-----------------------------------------------------+
```

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
@@ -793,6 +793,12 @@ public class KernelTransactionImplementation implements KernelTransaction, TxSta
         storageStatement.close();
     }
 
+    public Collection<Locks.ActiveLock> activeLocks()
+    {
+        StatementLocks locks = this.statementLocks;
+        return locks == null ? Collections.emptyList() : locks.activeLocks();
+    }
+
     /**
      * It is not allowed for the same transaction to perform database writes as well as schema writes.
      * This enum tracks the current write transactionStatus of the transaction, allowing it to transition from
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
@@ -29,6 +29,7 @@ import java.util.Optional;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import org.neo4j.collection.pool.Pool;
 import org.neo4j.graphdb.TransactionTerminatedException;
@@ -50,6 +51,7 @@ import org.neo4j.kernel.api.txstate.TxStateHolder;
 import org.neo4j.kernel.impl.api.state.ConstraintIndexCreator;
 import org.neo4j.kernel.impl.api.state.TxState;
 import org.neo4j.kernel.impl.factory.AccessCapability;
+import org.neo4j.kernel.impl.locking.ActiveLock;
 import org.neo4j.kernel.impl.locking.LockTracer;
 import org.neo4j.kernel.impl.locking.Locks;
 import org.neo4j.kernel.impl.locking.StatementLocks;
@@ -793,10 +795,18 @@ public class KernelTransactionImplementation implements KernelTransaction, TxSta
         storageStatement.close();
     }
 
-    public Collection<Locks.ActiveLock> activeLocks()
+    /**
+     * This method will be invoked by concurrent threads for inspecting the locks held by this transaction.
+     * <p>
+     * The fact that {@link #statementLocks} is a volatile fields, grants us enough of a read barrier to get a good
+     * enough snapshot of the lock state (as long as the underlying methods give us such guarantees).
+     *
+     * @return the locks held by this transaction.
+     */
+    public Stream<? extends ActiveLock> activeLocks()
     {
         StatementLocks locks = this.statementLocks;
-        return locks == null ? Collections.emptyList() : locks.activeLocks();
+        return locks == null ? Stream.empty() : locks.activeLocks();
     }
 
     /**
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
@@ -20,8 +20,7 @@
 package org.neo4j.kernel.impl.locking;
 
 import java.time.Clock;
-import java.util.Collection;
-import java.util.Objects;
+import java.util.stream.Stream;
 
 import org.neo4j.helpers.Service;
 import org.neo4j.kernel.configuration.Config;
@@ -115,7 +114,7 @@ public interface Locks
         /** For slave transactions, this tracks an identifier for the lock session running on the master */
         int getLockSessionId();
 
-        Collection<ActiveLock> activeLocks();
+        Stream<? extends ActiveLock> activeLocks();
     }
 
     /**
@@ -130,67 +129,4 @@ public interface Locks
     void accept(Visitor visitor);
 
     void close();
-
-    abstract class ActiveLock
-    {
-        public final ResourceType resourceType;
-        public final long resourceId;
-
-        private ActiveLock( ResourceType resourceType, long resourceId )
-        {
-            this.resourceType = resourceType;
-            this.resourceId = resourceId;
-        }
-
-        @Override
-        public boolean equals( Object o )
-        {
-            if ( this == o )
-            {
-                return true;
-            }
-            if ( o.getClass() != this.getClass() )
-            {
-                return false;
-            }
-            ActiveLock activeLock = (ActiveLock) o;
-            return resourceId == activeLock.resourceId &&
-                    Objects.equals( resourceType, activeLock.resourceType );
-        }
-
-        @Override
-        public int hashCode()
-        {
-            return Objects.hash( resourceType, resourceId );
-        }
-
-        public abstract String mode();
-    }
-
-    final class ActiveExclusiveLock extends ActiveLock
-    {
-        public ActiveExclusiveLock( ResourceType resourceType, long resourceId )
-        {
-            super( resourceType, resourceId );
-        }
-
-        @Override
-        public String mode()
-        {
-            return "EXCLUSIVE";
-        }
-    }
-    final class ActiveSharedLock extends ActiveLock
-    {
-        public ActiveSharedLock( ResourceType resourceType, long resourceId )
-        {
-            super( resourceType, resourceId );
-        }
-
-        @Override
-        public String mode()
-        {
-            return "SHARED";
-        }
-    }
 }
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
@@ -115,6 +115,8 @@ public interface Locks
         int getLockSessionId();
 
         Stream<? extends ActiveLock> activeLocks();
+
+        long activeLockCount();
     }
 
     /**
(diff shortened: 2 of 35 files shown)


thobe:
Merging this should probably wait until #8579 has been merged, since it touches the same test files.

chrisvest:
Are they _query ids_ or _transaction ids_?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -793,6 +793,12 @@ public void dispose()
         storageStatement.close();
     }
 
+    public Collection<Locks.ActiveLock> activeLocks()
+    {
+        StatementLocks locks = this.statementLocks;
+        return locks == null ? Collections.emptyList() : locks.activeLocks();
+    }
```
chrisvest: I think all the methods that can be accessed concurrently with the running transaction should have javadocs that clearly calls out this fact. The `markForTermination()` method is a good example.

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
```diff
@@ -126,4 +130,67 @@ void visit( ResourceType resourceType, long resourceId, String description, long
...
+        {
+            super( resourceType, resourceId );
+        }
+
+        @Override
+        public String mode()
+        {
+            return "SHARED";
+        }
+    }
```
chrisvest: I guess you use this pattern to make the active locks collection use less heap memory?

community/kernel/src/test/java/org/neo4j/kernel/impl/locking/ActiveLocksListingCompatibility.java
```diff
@@ -0,0 +1,60 @@
...
+        // then
+        assertEquals(
+                asList(
+                        new Locks.ActiveExclusiveLock( NODE, 1 ),
+                        new Locks.ActiveExclusiveLock( NODE, 2 ),
+                        new Locks.ActiveExclusiveLock( NODE, 3 ),
+                        new Locks.ActiveSharedLock( NODE, 3 ),
+                        new Locks.ActiveSharedLock( NODE, 4 ),
+                        new Locks.ActiveSharedLock( NODE, 5 ) ),
+                locks );
```
chrisvest: Do we really want to assert on the order? A lock manager implementation may be using hashed sets internally, and requiring an order would force a sort to happen.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/locks/LeaderOnlyLockManager.java
```diff
@@ -257,5 +258,11 @@ public int getLockSessionId()
         {
             return lockTokenId;
         }
+
+        @Override
+        public Collection<ActiveLock> activeLocks()
+        {
+            throw new UnsupportedOperationException( "not implemented" );
+        }
```
chrisvest: You should still be able to delegate this to the `localLocks`.

enterprise/ha/src/main/java/org/neo4j/kernel/ha/lock/SlaveLocksClient.java
```diff
@@ -240,6 +241,12 @@ public int getLockSessionId()
         return initialized ? client.getLockSessionId() : -1;
     }
 
+    @Override
+    public Collection<Locks.ActiveLock> activeLocks()
+    {
+        throw new UnsupportedOperationException( "not implemented" );
```
chrisvest: This one should also be able to use the `sharedLocks` and `exclusiveLocks` maps.

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
```diff
@@ -245,6 +245,24 @@ public ProcedureResult( ProcedureSignature signature )
...
+        {
+            long id = fromExternalString( queryId ).kernelQueryId();
+            return getActiveTransactions( tx -> executingQueriesWithId( id, tx ) )
+                    .flatMap( pair -> pair.first().activeLocks().stream().map( ActiveLocksQueryResult::new ) );
+        }
+        catch ( UncaughtCheckedException uncaught )
+        {
+            throwIfPresent( uncaught.getCauseIfOfType( InvalidArgumentsException.class ) );
+            throw uncaught;
+        }
```
chrisvest, line 263: Do we need `activeLocks()` to snapshot its state in a collection, or could it directly return a stream?

Why do we need to transform them to `ActiveLocksQueryResult` objects?

enterprise/kernel/src/test/java/org/neo4j/kernel/enterprise/builtinprocs/ListQueriesProcedureTest.java
```diff
@@ -71,36 +77,11 @@ public void shouldContainTheQueryItself() throws Exception
...
-        nodeLocked.await();
-
-        threads.executeAndAwait( parameter ->
-        {
-            db.execute( "MATCH (n) SET n.v = n.v + 1" ).close();
-            return null;
-        }, null, waitingWhileIn( GraphDatabaseFacade.class, "execute" ), 5, SECONDS );
-
-        try
+        String QUERY = "MATCH (n) SET n.v = n.v + 1";
```
chrisvest: It looks odd to me when local variables have all upper case names.

chrisvest (commented):
@thobe I had a number of questions and comments that I'd like you to take a look at.

community/kernel/src/test/java/org/neo4j/kernel/impl/locking/ActiveLocksListingCompatibility.java
thobe: good point, thank you. I know I did unordered testing elsewhere. will fix.

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
thobe: correct.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
thobe: ok, I'll add that.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/locks/LeaderOnlyLockManager.java
thobe: ok.

enterprise/ha/src/main/java/org/neo4j/kernel/ha/lock/SlaveLocksClient.java
thobe: ok.

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
thobe: I implemented it as snapshots since that was easier than streaming on the implementation level. Initially I had streams on the API level but then asked myself "who am I fooling", and changed the API to collections like it was in the implementation.

We need to convert to something with the appropriate public fields, not sure we want to have the internal API use such style of object, but maybe we do. Or perhaps we can turn this into some sort of visitor based (internal) API then have that directly construct the stream of the appropriate objects. That would be a more complicated API, but would imply less allocation.

enterprise/kernel/src/test/java/org/neo4j/kernel/enterprise/builtinprocs/ListQueriesProcedureTest.java
thobe: I used uppercase as a highlight of the central aspect of the test, but I don't mind changing it.

thobe:
Yes, these are Query IDs, not Transaction IDs, since we don't have any IDs fore live transactions. I think it would be better to introduce a Transaction ID, but that was deemed out of scope for the feature request that prompted this work.

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/Locks.java
thobe: ... and also as a way to keep these string constants encapsulated.

chrisvest:
Okay. Would it make sense to name the procedure `listActiveLocksForQuery` so we have naming room for a future `listActiveLocks` that takes transaction ids, in case we want to add that later?

chrisvest:
Alternatively, if we add transaction ids later, I suppose the `dbms.listQueries()` could start returning those instead and we can keep the names and queries would keep working?

thobe:
if we add transaction ids later, `dbms.listQueries()` would return _both_ the transaction id and the query id. What we could do though is have the `listActiveLocks` procedure look at the input and do one thing if it is a transaction id, and another thing if it's a query id...

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
thobe: With it now changed to Steams a few places became simpler and could avoid allocation / copying.

enterprise/ha/src/main/java/org/neo4j/kernel/ha/lock/SlaveLocksClient.java
```diff
@@ -242,9 +247,21 @@ public int getLockSessionId()
...
 
     @Override
-    public Collection<Locks.ActiveLock> activeLocks()
+    public Stream<? extends ActiveLock> activeLocks()
     {
-        throw new UnsupportedOperationException( "not implemented" );
+        return Stream.concat( // TODO: do we need to make these maps of locks ConcurrentHashMaps?
+                exclusiveLocks.entrySet().stream().flatMap( EXCLUSIVE_ACTIVE_LOCKS ),
+                sharedLocks.entrySet().stream().flatMap( SHARED_ACTIVE_LOCKS ) );
+    }
```
chrisvest: Ouch, that's a good point. HashMap is not safe for reads concurrent with writes, because a rehash will reverse the linked list collision buckets, which a concurrent reader than observe as a cycle in the chain, and thus end up in an infinite loop. `ConcurrentHashMap` has slower writes and uses more memory, but I think we probably need it for the lock maps. The maps that segregate by resource type could be turned into arrays and indexed by the resource types ordinal. The probability of seeing an old value in these arrays are probably negligible. It's unfortunate that there's no `AtomicReferenceArrayFieldUpdater`.

chrisvest (changes requested):
I had one comment about the use of `HashMap`s and thread safety in `SlaveLocksClient`. It also needs a rebase, which might also fix the failing test. Otherwise it looks good.

thobe:
@chrisvest yes, that was what I thought would be the right way to do it. I added those TODO notes so that someone would spot it in review and weigh in with another opinion on how to do that. Since we both have the same idea for how to structure it, I'll go ahead and make that change.

thobe:
@chrisvest the rebasing I was taking care of while you were writing your comment 😉

thobe:
@chrisvest fixed.

In a much simpler way. Since the local `Locks.Client` has all the lock information we need (except maybe re-entry counts), and is capable of supporting this API, we can just delegate to that and avoid any concurrency concern in `SlaveLocksClient`.

chrisvest (approved):

chrisvest:
Beautiful.

PR #8590, opened 2017-01-10
Provide information from query planning through dbms.listQueries()

thobe:
The information provided is:

 * Which planner was used.
 * Which runtime is used.
 * Which indexes are used from the plan.

The index usage output currently does not include legacy indexes.


thobe:
Merging this should probably wait for #8589, since it touches the same test files. Most likely a rebase will be required.

henriknyman (approved):

PR #8592, opened 2017-01-10
Initial version of a NativeLabelScanStore

tinwelint:
built using our GB+Tree as storage. Storage structure-wise it's very similar
to the current Lucene label scan store in that nodes are grouped into small
bit sets of 64 as the unit of storage.

Initial benchmarks show big write performance improvements (within and up to
an order of magnitude, even for heavily concurrent write load). Read performance
is equal to or better than that of Lucene label scan store.

Disk space-wise it occupies a bit more than Lucene lss, within 2x.
Memory-wise it's much less because there's no caching of the index,
only the page cache itself.
Cost of opening a native lss is small and constant compared to lucene lss,
or rather Lucene in general where cost of opening index is a function of the size
of the index. As an example it could take a minute to open a Lucene lss on
a store with 10's of billions of labelled nodes, sub-second with the native lss.
Additionally there's no background processing or file-merging on native lss.

Currently this lss implementation is disabled by default, as implmentation
gets more complete over time this will be the default lss for Neo4j builds.

Testing of NativeLabelScanStore contains unit tests for its units as well as some
integration testing. A broader compliance test suite will come in a later PR.


tinwelint:
#8597 changes locations of GB+Tree classes, but merged that PR will not cause conflicts in this one, rather compilation errors when this PR gets merged in. Be wary of that.

MishaDemianenko:
@tinwelint i will focus on this one at first, since other one in super worst case quite easy to just redo.

tinwelint:
@MishaDemianenko I think that's the better choice, yeah

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -0,0 +1,265 @@
...
+     *
+     * @param limiter {@link IOLimiter}.
+     * @throws UnderlyingStorageException on failure writing changes to {@link PageCache}.
+     */
+    @Override
+    public void force( IOLimiter limiter ) throws UnderlyingStorageException
+    {
+        try
+        {
+            index.checkpoint( limiter );
```
MishaDemianenko, line 186: Since index is based on pagedFile and that one of the mappedFiles in PageCache we will gonna checkpoint that pageFile yet another time as part of pageCache flushing. It does not really brake anything as far as i can see except actually doing yet another flush and force.

community/kernel/src/main/java/org/neo4j/kernel/api/labelscan/LabelScanStore.java
```diff
@@ -91,8 +92,22 @@ public void rebuilt( long roughNodeCount )
 
     /**
      * Acquire a writer for updating the store.
+     *
+     * @return {@link LabelScanWriter} which can modify the {@link LabelScanStore}.
+     */
+    default LabelScanWriter newWriter()
```
MishaDemianenko: I do not really like `newWriter()` and `newWriter( boolean batching )` since that give a confusion how those guys relate to each other?, is `newWriter(false)` the same as `newWriter()`? and you can answer that without looking into the code. I would prefer if we will have 2 methods like: `newWriter()` and `newBatchedWriter()` or similar instead.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/scan/LabelScanStoreProvider.java
```diff
@@ -91,13 +91,23 @@ public String toString()
         long applyTo( LabelScanWriter writer ) throws IOException;
     }
 
+    public static final FullStoreChangeStream EMPTY = writer -> 0;
```
MishaDemianenko: this should be part of `FullStoreChangeStream` interface instead. And i think that would be good idea to move this interface out of this class

community/kernel/src/main/java/org/neo4j/kernel/impl/api/scan/NativeLabelScanStoreExtension.java
```diff
@@ -0,0 +1,54 @@
...
+    public NativeLabelScanStoreExtension()
+    {
+        super( "native" );
+    }
+
+    @Override
+    public Lifecycle newInstance( KernelContext context, Dependencies dependencies ) throws Throwable
+    {
+        return new LabelScanStoreProvider( new NativeLabelScanStore( dependencies.pageCache(),
+                context.storeDir(), Long.SIZE, 0, fullStoreLabelUpdateStream( dependencies.indexStoreView() ) ), 0 );
```
MishaDemianenko: can we avoid passing page size as constructor parameter here and use one from page cache instead?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanKey.java
```diff
@@ -0,0 +1,50 @@
...
+    long idRange = -1;
+
+    /**
+     * Sets this key.
+     *
+     * @param labelId labelId for this key.
+     * @param idRange node idRange for this key.
+     * @return this key instance, for convenience.
+     */
+    LabelScanKey set( int labelId, long idRange )
```
MishaDemianenko, line 48: Can we transform it into constructor instead? (There is one place in NativeLabelScanWriter where set is actually used to reset value and reuse existing instance, if that is important we can keep set, but also please provide constructor since all other places will benefit form that)

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanKey.java
```diff
@@ -0,0 +1,50 @@
...
+    {
+        this.labelId = labelId;
+        this.idRange = idRange;
+        return this;
+    }
+
+    @Override
+    public String toString()
+    {
+        return "[lbl:" + labelId + ",range:" + idRange + "]";
```
MishaDemianenko: please make it `label` instead of `lbl`

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanLayout.java
```diff
@@ -0,0 +1,227 @@
...
+    public void writeKey( PageCursor cursor, LabelScanKey key )
+    {
+        cursor.putInt( key.labelId );
+        put6ByteLong( cursor, key.idRange );
+    }
+
+    private static void put6ByteLong( PageCursor cursor, long value )
+    {
+        cursor.putInt( (int) value );
+        cursor.putShort( (short) (value >>> 32) );
```
MishaDemianenko: looks like Integer.SIZE

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanLayout.java
```diff
@@ -0,0 +1,227 @@
...
+    {
+        into.labelId = cursor.getInt();
+        into.idRange = get6ByteLong( cursor );
+    }
+
+    private static long get6ByteLong( PageCursor cursor )
+    {
+        long low4b = cursor.getInt() & 0xFFFFFFFFL;
+        long high2b = cursor.getShort();
+        return low4b | (high2b << 32);
```
MishaDemianenko: also looks like Integer.SIZE

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanLayout.java
```diff
@@ -0,0 +1,227 @@
...
+     * otherwise verified against index meta data on open.
+     */
+    private final int rangeSize;
+
+    /**
+     * {@link #rangeSize}, as number of bytes (instead of number of bits).
+     */
+    private final int rangeSizeBytes;
+
+    LabelScanLayout( int rangeSize )
```
MishaDemianenko: do we really need to support various rangeSizes? looks like if that will be only one it would be so much more easier and exactly what we need right now

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIterator.java
```diff
@@ -0,0 +1,127 @@
...
+            try
+            {
+                if ( !cursor.next() )
+                {
+                    return false;
+                }
+            }
+            catch ( IOException e )
+            {
+                throw new RuntimeException( e );
```
MishaDemianenko: UncheckedIOException looks more suitable here

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIterator.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private boolean keysInOrder( LabelScanKey key )
+    {
+        assert key.labelId >= prevLabel : "Expected to get ordered results, got " + key +
+                " where previous label was " + prevLabel;
+        assert key.idRange > prevRange : "Expected to get ordered results, got " + key +
+                " where previous range was " + prevRange;
+        prevLabel = key.labelId;
+        prevRange = key.idRange;
+        // Made as a method returning boolean so that it can participate in an assert call.
```
MishaDemianenko, line 112: i do not really get idea why this should return true to just participate in assertion ....

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
```diff
@@ -0,0 +1,165 @@
...
+     * This information can be had by asking the node store directly instead.
+     *
+     * @param nodeId node id to get labels for.
+     * @return {@link PrimitiveLongIterator} with labels for {@code nodeId}.
+     * @throws UnsupportedOperationException since not supported by this implementation.
+     */
+    @Override
+    public PrimitiveLongIterator labelsForNode( long nodeId )
+    {
+        throw new UnsupportedOperationException( "Use your db..." );
```
MishaDemianenko, line 158: Message is not really descriptive and surprising mostly, lets add some more explanation what is expected to do to fix this

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    @Override
+    public PrimitiveLongIterator nodesWithLabel( int labelId )
+    {
+        RawCursor<Hit<LabelScanKey,LabelScanValue>,IOException> cursor;
+        try
+        {
+            ensureOpenCursorsClosed();
```
MishaDemianenko, line 85: I find it strange that calling to `nodesWithLabel('1')` and then `nodesWithLabel('2')` will suddenly close your first reader. Maybe lets change instead type of those return iterators to be PrimitiveLongResourceIterator at least?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -0,0 +1,265 @@
...
+     */
+    private boolean recoveryStarted;
+
+    public NativeLabelScanStore( PageCache pageCache, File storeDir, int rangeSize, int pageSize,
+            FullStoreChangeStream fullStoreChangeStream )
+    {
+        this.pageCache = pageCache;
+        this.pageSize = pageSize;
+        this.fullStoreChangeStream = fullStoreChangeStream;
+        this.storeFile = new File( storeDir, DEFAULT_NAME + ".labelscanstore.db" );
```
MishaDemianenko, line 128: hm it looks like a new `StoreFile` type should be introduced instead and maybe name in store factory? WDYT?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * The index which backs this label scan store. Instantiated in {@link #init()} and considered
+     * started after call to {@link #start()}.
+     */
+    private GBPTree<LabelScanKey,LabelScanValue> index;
+
+    /**
+     * Whether or not {@link #start()} has been called.
+     */
+    private boolean started;
```
MishaDemianenko: looks like it should be volatile variable since writers can be started from other threads?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -0,0 +1,265 @@
...
+            if ( !started && !recoveryStarted )
+            {
+                // Let's notify our index that recovery is about to commence, we do this once before
+                // the first recovered transaction gets applied.
+                index.prepareForRecovery();
+                recoveryStarted = true;
+            }
+
+            IndexWriter<LabelScanKey,LabelScanValue> inserter = index.writer();
+            return new NativeLabelScanWriter( inserter, rangeSize, batching ? 10_000 : 1_000 );
```
MishaDemianenko: thats confusing that batching false is still batching but with different size

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -0,0 +1,265 @@
...
+                new LabelScanKey().set( 0, 0 ),
+                new LabelScanKey().set( Integer.MAX_VALUE, Long.MAX_VALUE ) ) )
+        {
+            return !cursor.next();
+        }
+    }
+
+    @Override
+    public void stop() throws IOException
+    {   // Not needed
```
MishaDemianenko, line 255: it looks like we should make started=false here to prevent creation of new writers?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanWriter.java
```diff
@@ -0,0 +1,272 @@
...
+    private static final ValueMerger<LabelScanValue> ADD_MERGER = (value,withValue) -> value.add( withValue );
+
+    /**
+     * {@link ValueMerger} used for removing label->node mappings, see {@link LabelScanValue#remove(LabelScanValue)}.
+     */
+    private static final ValueMerger<LabelScanValue> REMOVE_MERGER = (value,withValue) -> value.remove( withValue );
+
+    /**
+     * {@link IndexWriter} acquired when acquiring this {@link NativeLabelScanWriter},
+     * acquired from {@link GBPTree#writer(org.neo4j.index.IndexWriter.Options)}.
```
MishaDemianenko: javadoc is broken here

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanWriter.java
```diff
@@ -0,0 +1,272 @@
...
+ * applied in sorted order (by label and node id) when reaching batch size and or on {@link #close()}.
+ * <p>
+ * Updates aren't visible to {@link LabelScanReader readers} immediately, rather when queue happens to be applied,
+ * and (simples guarantee to reason about) on {@link #close()}.
+ * <p>
+ * Incoming {@link NodeLabelUpdate updates} are actually modified from representing physical before/after
+ * state to represent to-add/to-remove. These changes are done directly inside the provided
+ * {@link NodeLabelUpdate#getLabelsAfter()} and {@link NodeLabelUpdate#getLabelsBefore()} arrays,
+ * relying on the fact that those arrays are returned in its essential form, instead of copies.
+ * This conversion is done like so mostly to reduce garbage. See NodeLa
```
MishaDemianenko: see section is a bit interrupted in the middle :)

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanWriter.java
```diff
@@ -0,0 +1,272 @@
...
+    /**
+     * {@link Comparator} for sorting the node id ranges, used in batches to apply updates in sorted order.
+     */
+    private static final Comparator<NodeLabelUpdate> UPDATE_SORTER =
+            (o1, o2) -> Long.compare( o1.getNodeId(), o2.getNodeId() );
+
+    /**
+     * {@link ValueMerger} used for adding label->node mappings, see {@link LabelScanValue#add(LabelScanValue)}.
+     */
+    private static final ValueMerger<LabelScanValue> ADD_MERGER = (value,withValue) -> value.add( withValue );
```
MishaDemianenko: LabelScanValue::add

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanWriter.java
```diff
@@ -0,0 +1,272 @@
...
+
+    /**
+     * {@link ValueMerger} used for removing label->node mappings, see {@link LabelScanValue#remove(LabelScanValue)}.
+     */
+    private static final ValueMerger<LabelScanValue> REMOVE_MERGER = (value,withValue) -> value.remove( withValue );
```
MishaDemianenko: LabelScanValue::remove;

community/kernel/src/test/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanWriterTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+        @Override
+        public void merge( LabelScanKey key, LabelScanValue value, ValueMerger<LabelScanValue> amender )
+                throws IOException
+        {
+            // Clone since these instances are reused between calls, internally in the writer
+            key = clone( key );
+            value = clone( value );
+
+            Map<LabelScanKey,LabelScanValue> forLabel = data.get( key.labelId );
```
MishaDemianenko: `Map<LabelScanKey,LabelScanValue> forLabel = data.computeIfAbsent( key.labelId, k -> new TreeMap<>( KEY_COMPARATOR ) );` can be used instead of null check

MishaDemianenko (changes requested):
Looks good in general, but have couple of comments still

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
tinwelint: We talked about it and we thought we could defer it to reduce the PR size. This will be done as part of further integration ahead. OK?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
tinwelint: Initially this `batching` meant more, but now it changes only the internal batch size. We'll measure if this really matters at all, otherwise revert this argument addition.

EDIT: it makes a couple % difference. Bumping to 100k for batching improves performance up to almost 10% (but increases memory overhead of keeping those NodeLabelUpdate objects). I don't think this is worth it at this point, will revert.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
tinwelint: flushing a PagedFile that has already been flushed should be really fast and checkpointing is a "rare" thing. I think this is fine.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
tinwelint: Maybe, but `started==false` doesn't prevent new writers and LuceneLabelScanStore doesn't do anything in `stop()` either

community/kernel/src/main/java/org/neo4j/kernel/api/labelscan/LabelScanStore.java
tinwelint: As above comment mentions, this will be removed and so it looks like it used to do, with only a `newWriter()` method.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanLayout.java
tinwelint: Initially it was there so that we could compare different range sizes against each other, although I'd guess landing at the 64-bit variant for the Lucene LSS tells the tale that this investigation has already been done and we could simply just do that.

We almost removed rangeSize as part of this PR, but narrowly decided to keep it. This may have tipped the scale. Removing it.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIterator.java
tinwelint: inside `fetchNext` this method is used as an assertion, it's simply a helpful fail-fast debug thing we've added.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
tinwelint: That would be a bigger change to the neo4j integration. The current behaviour is what's expected of the `LabelScanReader` afaict. Let's revisit this behaviour in another context if need be, shall we?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
tinwelint: This is actually not a problem because the next PR we have coming will remove that method from LabelScanReader since it's redundant anyway :)

tinwelint:
@MishaDemianenko great comments, have addressed or replied on all. Please have another go

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
MishaDemianenko: Sure, ok

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
MishaDemianenko: ok, true

community/kernel/src/main/java/org/neo4j/kernel/api/labelscan/LabelScanStore.java
MishaDemianenko: ok

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
MishaDemianenko: ok agree

MishaDemianenko:
@tinwelint looks great! will merge on green!

tinwelint:
Cool, there looks to be some unrelated flaky clustering tests in this current build fyi

MishaDemianenko (approved):

PR #8593, opened 2017-01-10
Added procedure to list currently active config

spacecowboy:
Example outputs:

- default output

default output

- with `showHidden=true` shows unsupported stuff (e.g. all of the config)

includes unsupported settings when true

changelog [procedures]

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInDbmsProcedures.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInDbmsProcedures.java
@@ -22,7 +22,6 @@ package org.neo4j.kernel.builtinprocs;
 import java.util.Comparator;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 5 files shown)


community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInDbmsProcedures.java
```diff
@@ -55,6 +70,20 @@
                 .map( FunctionResult::new );
     }
 
+    public static class ConfigResult
```
praveenag: If it is public we might as well have a file of it's own.

spacecowboy: It is public, but it's only ever used in this file (for the procedure). And I was only following the style previously set in this file (which is not necessarily ideal).

praveenag (approved):

praveenag (approved):

PR #8597, opened 2017-01-10
Make GB+Tree purely internal data structure

burqen:
Treat GB+Tree as purely internal data structure by removing `Index` interface and move
`o.n.i.gbptree -> o.n.i.internal.gbptree`.


MishaDemianenko:
@burqen it looks like there is a test failure related to new index

burqen:
@MishaDemianenko Addressed

MishaDemianenko:
@burqen can you please rebase

tinwelint:
Rebased now

community/index/src/main/java/org/neo4j/index/internal/gbptree/ConsistencyChecker.java
```diff
@@ -17,22 +17,23 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-package org.neo4j.index.gbptree;
```
MishaDemianenko, line 20: @tinwelint @burqen i'm surprised a bit to see that all classes are moved to internal package now; was excepting to see there only things that should be internal to this module and in public package something that should be visible and used outside of this module

MishaDemianenko (approved):

community/index/src/main/java/org/neo4j/index/internal/gbptree/ConsistencyChecker.java
tinwelint: Mostly this is to not commit to a public surface since this is purely internal stuff a.t.m.

PR #8604, opened 2017-01-11
Added a way to override default values in config for docs

spacecowboy:
Some settings are calculated at runtime and depend on the current
machine. This is problematic for documentation which uses the default
values and prints them in the manual. This change adds an Annotation
which can specify what to  put in the manual if the actual default
value is not suitable.

Also added the necessary plumbing to access the information in the
annotations via the Config class.


praveenag (approved):

community/configuration/src/main/java/org/neo4j/configuration/ConfigOptions.java
```diff
@@ -24,7 +24,9 @@
 import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
+import javax.swing.text.html.Option;
```
jjaderberg: Over-zealous IDE? Both new imports are unused.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ConnectorValidator.java
```diff
@@ -20,6 +20,7 @@
 package org.neo4j.kernel.configuration;
 
 import java.util.Arrays;
+import java.util.Collections;
```
jjaderberg: unused

jjaderberg (changes requested):
This is a nice way to handle defaults that can't be documented. Can we expose whether a setting is deprecated and what its valid values restrictions are in a similar way?

spacecowboy:
@jjaderberg corrected the unused imports. I will do the deprecated work separately

jjaderberg (approved):

PR #8605, opened 2017-01-11
Read replica startup

martinfurmanski:
changelog: [causal clustering] Improve error reporting for read replica startup.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/ReadReplicaStartupProcess.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/ReadReplicaStartupProcess.java
@@ -19,6 +19,7 @@
  */
 package org.neo4j.causalclustering.readreplica;
 
+import java.io.IOException;
 import java.util.concurrent.locks.LockSupport;
 
 import org.neo4j.causalclustering.catchup.storecopy.CopiedStoreRecovery;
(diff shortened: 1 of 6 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/ReadReplicaStartupProcess.java
```diff
@@ -41,18 +42,23 @@
 
 class ReadReplicaStartupProcess implements Lifecycle
 {
+    private static final int MAX_ATTEMPTS = 5;
```
jimwebber, line 45: Maybe a `GraphDatabaseSettings` setting instead?

martinfurmanski: To be honest I think this should just be redesigned, so I would rather not consider that point for this PR. If this was exposed in settings it should rather be something like a maximum time to keep trying before giving up.

jimwebber (commented):
Just a minor tweak to put a hardcoded setting into a config.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/ReadReplicaStartupProcess.java
jimwebber: Since this MAX_ATTEMPTS was introduced in this PR, I think it'd be nice to compute MAX_ATTEMPTS from some timeout set in config.

martinfurmanski: MAX_ATTEMPTS was just extracted from the hard-coded inline value

PR #8607, opened 2017-01-11
Reintroduce Cypher test comparisons with rule planner

henriknyman:
The rule planner still exists in Cypher 3.1, so we can run the Cypher tests
with assertions that the results are the same between the latest and the
old rule planner.


systay:
@henriknyman This needs some love

PR #8608, opened 2017-01-12
Added native systemd service

spacecowboy:
and moved the sysvinit service so it is installed by dh_installinit.

dh_installinit will do the right thing depending on the system you are
installing on. If the system has systemd, it will enable the native
systemd unit. If it has sysvinit, it will enable the old sysvinit
service.

I have also made sure that the service will be called `neo4j` for both
the neo4j and neo4j-enterprise packages.

I have tested locally (Debian Stretch) that the systemd service is installed correctly.

I tested on Ubuntu 14.04 (which does not have systemd) in VirtualBox and verified that it installs and enables the correct service there:

- [x] Make sure an existing `/etc/default/neo4j` is parsed on upgrade and put into `/etc/systemd/system/neo4j.service.d/override.conf`


benbc:
@spacecowboy This looks great. Thanks.

I think this will (or at least should!) involve documentation changes for the timeout and file handle limit settings.

I believe that the changes you've made so that "the service will be called neo4j for both
the neo4j and neo4j-enterprise packages" is effectively just an implementation detail of using `dh_installinit`. You haven't changed the name of the service from 3.1, have you?

The Quality team are expanding the matrix of compatibility builds. Please talk to them to ensure that there will be (at some point soon) two different distros/versions in tyre-kicking with the Debian package so that we hit both service definitions.

spacecowboy:
@benbc added code to handle migration of defaults file

You are correct in that the service name is not a change - it's just that we used to copy the service and defaults file manually and not let debian do it for us.

Examples of the upgrade process

## User has modified /etc/default/neo4j int their 3.1.0 install and is upgrading to 3.2.0

selection_012

Migration is performed and then the user is asked by debian what to do with the changed defaults file

## User views diff between files

selection_013

And can choose whatever he wants (we have already migrated the file)

## After install, resulting service file

selection_014

new-packaging/src/debian/skeleton/debian/neo4j.default
```diff
@@ -7,10 +7,11 @@
...
 #
+# This is automatically done upon upgrade.
 ######################
 #
 # Number of seconds to wait for shutdown before killing the process.
 #NEO4J_SHUTDOWN_TIMEOUT=120
 #
 # Max number of open file descriptors
-NEO4J_ULIMIT_NOFILE=40000
+#NEO4J_ULIMIT_NOFILE=60000
```
benbc: Why this change?

new-packaging/src/debian/skeleton/debian/neo4j.default
```diff
@@ -7,10 +7,11 @@
 #
 #     systemctl edit neo4j.service
 #
+# This is automatically done upon upgrade.
```
benbc: I think this would be more helpful if it was a bit more verbose. Something like:

```
# Any values that were present in this file have been automatically migrated into
# SystemD configuration as part of the upgrade.
```

benbc (commented):
Looks good other than these two things.

new-packaging/src/debian/skeleton/debian/neo4j.default
spacecowboy: Value: 
I thought I was being consistent with what we shipped in the release 3.1.0 package but looking at the [source](https://github.com/neo4j/neo4j/commit/1fc3c0ab9f352f883a60cb126d25496a6f61d9e6) it seems I was mistaken.

Commented:
Being commented out by default is to prevent an unnecessary migration to happen when you go from 3.2.0 to 3.2.0+, it was also commented out before, so in the final diff this is not a change.

benbc (approved):

PR #8610, opened 2017-01-12
NativeLabelScanStore works with consistency checker

tinwelint:
Which means reducing LabelScanStore (and friends) interface and to some extent consistency checker to consistency check label scan store slightly differently.


tinwelint:
Hmm, not all builds were run on this one

chrisvest:
@tinwelint That's a new feature of our PR builds; to only run tests deemed relevant by some dependency analysis.

MishaDemianenko:
@chrisvest tests and builds are different things 🦄  😋

tinwelint:
@chrisvest @MishaDemianenko I know, but to run no tests at all.... doooon't think so. Actually there was a hickup in CI where other PRs didn't run properly either. Triggered them manually, now the `Run Linux Tests` is there, which is wasn't at all before

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
```diff
@@ -137,6 +137,12 @@ public void start() throws IOException
     }
 
     @Override
+    public boolean isEmpty() throws IOException
+    {
+        return allNodeLabelRanges().maxCount() == 0;
```
tinwelint: Hmm wait, the `allNodeLabelRanges()` should be in a try-with-resource, will fix that

tinwelint:
Failed on unrelated flaky test, it's fine.

community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
```diff
@@ -173,53 +182,48 @@ public Result runFullConsistencyCheck( final File storeDir, Config tuningConfigu
...
-            LabelScanStore labelScanStore = null;
-            try
-            {
-                IndexStoreView indexStoreView = new NeoStoreIndexStoreView( LockService.NO_LOCK_SERVICE, neoStores );
-                OperationalMode operationalMode = OperationalMode.single;
-                labelScanStore = new LuceneLabelScanStoreBuilder( storeDir, indexStoreView, fileSystem,
-                        consistencyCheckerConfig, operationalMode, logProvider )
-                        .build();
+            life.start();
+            IndexStoreView indexStoreView = new NeoStoreIndexStoreView( LockService.NO_LOCK_SERVICE, neoStores );
```
MishaDemianenko, line 188: it looks like we do not need IndexStoreView and operationalMode anymore here?

community/kernel/src/main/java/org/neo4j/helpers/progress/ProgressMonitorFactory.java
```diff
@@ -104,17 +104,37 @@ private MultiPartBuilder( ProgressMonitorFactory factory, String process )
...
-            }
+            assertNotBuilt();
+            assertUniquePart( part );
+            ProgressListener.MultiPartProgressListener progress =
+                    new ProgressListener.MultiPartProgressListener( aggregator, part, totalCount );
+            aggregator.add( progress, totalCount );
+            return progress;
+        }
+
+        public ProgressListener progressForUnknownPart( String part )
```
MishaDemianenko, line 115: that progress will be completely invisible on a output correct?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValue.java
```diff
@@ -32,6 +32,16 @@
      */
     long bits;
 
+    LabelScanValue( long bits )
```
MishaDemianenko: it looks like we do not really need to have this constructor it's used only internally in this class

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeAllEntriesLabelScanReader.java
```diff
@@ -0,0 +1,258 @@
...
+        private long idRange;
+        private long[] nodes;
+        private long[][] labels;
+
+        ReusableNodeLabelRange()
+        {
+            this.nodes = new long[RANGE_SIZE];
+        }
+
+        void setdRange( long idRange, long[][] labels, int slots )
```
MishaDemianenko: `setdRange` have or missing `I` or additional `d`

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeAllEntriesLabelScanReader.java
```diff
@@ -0,0 +1,258 @@
...
+                return null;
+            }
+
+            fill( labelsForEachNode, null );
+            long nextLowestRange = Long.MAX_VALUE;
+            int slots = 0;
+            try
+            {
+                // One "rangeSize" range at a time
+                for ( RawCursor<Hit<LabelScanKey,LabelScanValue>,IOException> cursor : cursors )
```
MishaDemianenko, line 139: hm...is it really safe to share cursors between iterator and reader? next call to iterator will gonna have influence on the first one?

community/kernel/src/test/java/org/neo4j/kernel/impl/index/labelscan/NativeAllEntriesLabelScanReaderTest.java
```diff
@@ -0,0 +1,322 @@
...
+        // GIVEN
+        try ( AllEntriesLabelScanReader reader = new NativeAllEntriesLabelScanReader(
+                store( data ), highestLabelId( data ) ) )
+        {
+            // WHEN/THEN
+            assertRanges( reader, data );
+        }
+    }
+
+    private static long count( Labels[] data )
```
MishaDemianenko: this method is not used

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveIntCollections.java
```diff
@@ -826,4 +827,32 @@ public static boolean contains( int[] values, int candidate )
...
         }
     }
+
+    /**
+     * Deduplicates values in the sorted {@code values} array.
+     *
+     * @param values sorted array of int values.
+     * @return the provided array if no duplicates were found, otherwise a new shorter array w/o duplicates.
+     */
+    public static int[] dedup( int[] values )
```
MishaDemianenko: surprisingly OperationsFacade have method that is almost duplicate of this one

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveIntCollections.java
```diff
@@ -826,4 +827,32 @@ public static boolean contains( int[] values, int candidate )
...
+    {
+        int unique = 0;
+        for ( int i = 0; i < values.length; i++ )
+        {
+            int value = values[i];
+            for ( int j = 0; j < unique; j++ )
```
MishaDemianenko, line 843: i think we can make second index start from i to remove duplicates?

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveIntCollections.java
MishaDemianenko: also please give it a full name `deduplicate` or `unique` or soemthing else :)

MishaDemianenko: this and next comment are also applicable for PrimitiveLongCollections

MishaDemianenko (changes requested):
Couple of mostly minor comments, otherwise looks really good!

community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
tinwelint: Actually this is a mistake, the IndexStoreView should be used as the dependency into the kernel extensions. However it's not reasonable that CC should be able build the label scan store from store on startup to check consistency of it later, but that's apparently how the code did. I don't think we have a single test which starts a CC on a store w/o label scan store though.

I can restore the old behaviour, i.e. using those dependencies again.

community/kernel/src/main/java/org/neo4j/helpers/progress/ProgressMonitorFactory.java
tinwelint: Yes, we'll have to see if we can do better later on, OK

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeAllEntriesLabelScanReader.java
tinwelint: Don't think cursors are shared. Iterator is responsible for moving cursors (those that are on the current range) forwards and cram their range data into a NodeLabelRange, which is returned out from the iterator.

Although this comment reminded me about a fix we did after this PR branch, which is that the NodeLabelRange instances can be processed by multiple threads once they leave the iterator, so changed ReusableNodeLabelRange to not be reused, instead creating new instances every time.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeAllEntriesLabelScanReader.java
tinwelint: Gone now, see comment above.

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveIntCollections.java
tinwelint: Oh, then that was duplicated from NodeItemHelper, which this `dedup` method was moved from. Will deduplicate that method as well :)

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveIntCollections.java
tinwelint: It goes the other way around, doesn't it, from 0 to `unique` which starts at 0 and grows.

MishaDemianenko:
@tinwelint it's ready for second round right?

tinwelint:
@MishaDemianenko yup, sorry didn't mention that

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeAllEntriesLabelScanReader.java
MishaDemianenko: ok

MishaDemianenko (approved):

Issue #8587, opened 2017-01-10
Node returns incorrect empty array property type


bHacklv:
When I create Node(s) with property that has *empty array value* of *any type* (long[], Long[], double[], etc.), when I read this(these) Node(s) from the store - the above properties type is *ALWAYS* empty String[] array. I would expect to receive correct initial array type...

**Neo4j Version:** 3.1
**Operating System:** Win7

### Steps to reproduce
1. Create Node with Empty Array Property Value Long[] type
2. Read Node from store
3. Read above property
4. Assert Property's type

### Expected behavior
instanceof Long[].class

### Actual behavior
instanceof String[].class

### 'Pseudo' test code
```java
final Transaction transaction = beginTransaction();
final Map<String, Object> queryParam = new HashMap<>(1);
queryParam.put("val", new Long[0]);
transaction.run("CREATE (n:`Default Label`) SET n.key = {val}", queryParam).close()
final Node n = (Node) execute.next().get("n");
assertThat(((ClientNodeWrapper) n).getProperty("key")).isInstanceOf(Long[].class);

===

Expecting:<[]>to be an instance of: <java.lang.Long[]>but was instance of:<java.lang.String[]>
```

### My investigation
I've investigated how the property value is being read:
```java
433: NodeProxy#getAllProperties -> propertyCursor.get().value()
220: StorePropertyPayloadCursor#value() -> type()
337: PropertyType#getPropertyTypeOrNull() -> typeIdentifier( propBlock );
-> returns 10 (Array) // correct!
245: StorePropertyPayloadCursor#arrayValue()
215: StorePropertyPayloadCursor#arrayValue() -> readArrayFromBuffer( buffer )
316: StorePropertyPayloadCursor#readArrayFromBuffer() -> buffer.get() -> returns 9
320: StorePropertyPayloadCursor -> if ( typeId == PropertyType.STRING.intValue() ) -> true....
```
So I assume that either the property type is incorrectly read from the store, or when empty array is written to the store, it defaults(?) or incorrectly written as empty String array...

While investigating writes, I've got down till ``org.neo4j.cypher.internal.ExecutionEngine`` and I was not able to dl source code for it.

### BOLT
**API:** neo4j-java-driver 1.1.0

When using Driver (BOLT) the problem is the same, but the return type is different: *java.util.Collections$EmptyList*.

I found that the type is being 'lost' in:
```
182: ExplicitTransaction#run -> value(statementParameters)
[...]

PR #8611, opened 2017-01-12
Preserve type of empty primitive arrays

Mats-SX:
All empty arrays were stored as String arrays, surprising users as they
would send f.e. an empty long array as parameter, but retrieve the value
and find that it was in fact a String array.

Addresses #8587

Still unclear: should we convert byte/short/int arrays to long arrays only? Same question for float arrays to double arrays. Should we move away from supporting array parameters, and only support proper list objects?

Changelog: Fixes [#8587](https://github.com/neo4j/neo4j/issues/8587) maintaining empty array type when storing and retrieving from properties

### DIFF
--- a/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ParameterValuesAcceptanceTest.scala
+++ b/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ParameterValuesAcceptanceTest.scala
@@ -19,11 +19,43 @@
  */
 package org.neo4j.internal.cypher.acceptance
 
-import org.neo4j.cypher.{ExecutionEngineFunSuite, NewPlannerTestSupport}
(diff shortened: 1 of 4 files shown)


systay:
@Mats-SX this needs some love

Mats-SX:
@systay Are we convinced that this is what we want Neo4j to do? Do you have any comments on the 'Still unclear' points in the PR message?

thobe:
I actually think we should move away from using built in java types at all (in the long run), and use our own types for transferring values in through parameters and out through results.

This makes the matter of mapping to and from the built in types of the language a driver concern.

Mats-SX:
@thobe I think that sounds sensible. That statement also relates (heavily) to #8740.

Are you envisioning that also the Core API would be migrated to use the custom transporter types?

henriknyman (approved):
Note that it may not be possible to send some of these types through Cypher parameters from 3.2 anymore if we decide to merge https://github.com/neo4j/neo4j/pull/8740

thobe:
@Mats-SX No, I think of the Core API as a driver/client API, so it should do the translation to Java.

PR #8613, opened 2017-01-13
Make sure CC recovers copied store with configured record format

chrisvest:
Also fixes the assertion in ReadReplicaReplicationIT test so that it is now
actually checking that this functionality works.


mneedham (commented):
Looks good to me.

chrisvest:
@mneedham Any news?

PR #8617, opened 2017-01-13
First draft of block device support for Causal Cluster

chrisvest:
* For doing store copy, store files are now read and written through the page
  cache.
* File operations - copy, move, delete - now go through the page cache for
  store files.
* Tests have been changed to not rely on the file system to e.g. copy store
  directories around, and also to allow the tests to be configured with block
  device storage in the blockdevice repository.
* The test changes primarily involve CoreToCoreCopySnapshotIT and
  ReadReplicaReplicationIT.
* We now send an alignment requirement over with the `FileHeader` message.
  _This is a protocol change!_ This is only a preliminary approach, and an
  invitation to discuss the design and what the best way is to solve this
  problem.
* The checks that mutually excludes Causal Clustering and Custom IO (block
  device) have been removed.

A few things are still left to do:

* Make sure TemporaryStoreDirectory cleans up after itself not only via the
  file system, but also via the page cache.
* Closer looks at LocalDatabase and catch-up code to make sure there are no
  latent bugs in there.
* Possibly extract some duplication that has been discovered between HA and CC?

These changes require #8613 to be merged before they work, so this PR has at least one rebase ahead of it already.

Please take a look, @digitalstain.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/FileSender.java
```diff
@@ -67,7 +66,16 @@ public FileChunk readChunk( ByteBufAllocator allocator ) throws Exception
...
-        byte[] next = prefetch();
+        byte[] next;
+        if ( !endOfInput )
+        {
+            next = prefetch();
+        }
+        else
+        {
+            next = null;
+        }
```
chrisvest, line 77: This probably needs to be removed, so it's just `next = prefetch();`

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/tx/CatchupPollingProcessTest.java
```diff
@@ -25,6 +25,7 @@
 import java.util.concurrent.CompletableFuture;
 
 import java.util.concurrent.Future;
+import java.util.stream.Stream;
```
chrisvest, line 28: This looks redundant

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloaderTest.java
```diff
@@ -23,7 +23,9 @@
 import org.junit.Test;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.UUID;
+import java.util.stream.Stream;
```
chrisvest, line 28: This one as well.

PR #8618, opened 2017-01-13
Remove Cypher dependency on kernel IndexDescriptor

craigtaverner:
changelog [3.1] Remove cypher dependency on kernel IndexDescriptor

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundGraphStatistics.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundGraphStatistics.scala
@@ -23,13 +23,13 @@ import org.neo4j.cypher.internal.compiler.v3_1.planner.logical.{Cardinality, Sel
 import org.neo4j.cypher.internal.compiler.v3_1.spi.{GraphStatistics, StatisticsCompletingGraphStatistics}
 import org.neo4j.cypher.internal.frontend.v3_1.{LabelId, NameId, PropertyKeyId, RelTypeId}
 import org.neo4j.kernel.api.exceptions.index.IndexNotFoundKernelException
-import org.neo4j.kernel.api.index.IndexDescriptor
+import org.neo4j.cypher.internal.compiler.v3_1.IndexDescriptor
 import org.neo4j.kernel.api.{ReadOperations, Statement => KernelStatement}
 
 object TransactionBoundGraphStatistics {
   def apply(ops: ReadOperations) = new StatisticsCompletingGraphStatistics(new BaseTransactionBoundGraphStatistics(ops))
 
-  private class BaseTransactionBoundGraphStatistics(operations: ReadOperations) extends GraphStatistics {
+  private class BaseTransactionBoundGraphStatistics(operations: ReadOperations) extends GraphStatistics with IndexDescriptorCompatibility {
 
     import NameId._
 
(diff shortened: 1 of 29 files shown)


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundGraphStatistics.scala
```diff
@@ -23,13 +23,13 @@ import org.neo4j.cypher.internal.compiler.v3_1.planner.logical.{Cardinality, Sel
...
 import org.neo4j.kernel.api.exceptions.index.IndexNotFoundKernelException
-import org.neo4j.kernel.api.index.IndexDescriptor
+import org.neo4j.cypher.internal.compiler.v3_1.IndexDescriptor
 import org.neo4j.kernel.api.{ReadOperations, Statement => KernelStatement}
 
 object TransactionBoundGraphStatistics {
   def apply(ops: ReadOperations) = new StatisticsCompletingGraphStatistics(new BaseTransactionBoundGraphStatistics(ops))
 
-  private class BaseTransactionBoundGraphStatistics(operations: ReadOperations) extends GraphStatistics {
+  private class BaseTransactionBoundGraphStatistics(operations: ReadOperations) extends GraphStatistics with IndexDescriptorCompatibility {
```
systay, line 32: Do we really need to extend this trait here?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
```diff
@@ -59,7 +59,7 @@ import scala.collection.Iterator
 import scala.collection.JavaConverters._
 
 final class TransactionBoundQueryContext(val transactionalContext: TransactionalContextWrapperv3_1)(implicit indexSearchMonitor: IndexSearchMonitor)
-  extends TransactionBoundTokenContext(transactionalContext.statement) with QueryContext {
+  extends TransactionBoundTokenContext(transactionalContext.statement) with QueryContext with IndexDescriptorCompatibility {
```
systay, line 62: Same question here - do we need to extend this trait?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundGraphStatistics.scala
craigtaverner: We could create and use kernel index descriptors within this class, but more lines would be changed, so I went for the minimum change approach, simplifying (slightly) code comparisons (and merges) with 3.2.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
craigtaverner: Making explicit conversions in this class would involve changes in many places. Are you suggesting instead that we extend QueryContext with IndexDescriptorCompatibility? I prefer them being seperate because they are separate concerns.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundGraphStatistics.scala
systay: Oh, `implicit` is the reason that I didn't see any use of the introduced trait in the diff... Ok.

PR #8619, opened 2017-01-14
Package the sources necessary to produce Javadocs

jjaderberg:
When the fullBuild profile was removed some modules no longer produced
the sources-jar necessary in the downstream Javadocs build.
This adds `maven-source-plugin` to these modules directly so that
a sources-jar is produced by default as part of the `package` phase.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,9 +69,27 @@
           </systemPropertyVariables>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-source-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


lassewesth:
@jjaderberg Why is this not just one single declaration at the root level?

jjaderberg:
@lassewesth Because we don't need a `sources-jar` from every module

lassewesth:
@jjaderberg Let's just build a sources jar for every module and keep the poms simple. Doing it per-module seems brittle

jjaderberg:
@lassewesth OK, I've moved it to the root pom.

PR #8621, opened 2017-01-16
Added permission for token creation

OliviaYtterbrink:
Added check after we know that a token needs to be created.
Default role needed to create tokens is `Architect`, can change `dbms.security.allow_publisher_create_token` to allow `Publisher` as well.

changelog: [security] Added new permission level for handling creation of tokens

* Default role given this permission is `Architect` and `Admin`
* setting `dbms.security.allow_publisher_create_token` to true will give this permission to `Publisher` as well


community/kernel/src/main/java/org/neo4j/kernel/api/security/AccessMode.java
```diff
@@ -173,6 +197,46 @@ public boolean allowsWrites()
...
+                        return true;
+                    }
+
+                    @Override
+                    public boolean allowsTokenCreates()
+                    {
+                        return true;
+                    }
+
+                    @Override
```
chrisvest: A note on style: I prefer to pass values through the constructor to fields of the enums, over the proliferation of overridden abstract methods. It makes the class shorter, since the getters only have to be implemented once.

community/kernel/src/main/java/org/neo4j/kernel/api/proc/Mode.java
```diff
@@ -29,6 +29,8 @@
     READ_ONLY,
     /** This procedure may perform both read and write operations against the graph */
     READ_WRITE,
+    /** This procedure may perform both read and write operations against the graph and create new tokens */
+    TOKEN,
```
chrisvest: I find it strange that the ability to create tokens goes beyond write. At first glance, I thought this was about allowing read-only transactions to also create tokens, but apparently not? Can you explain why TOKEN is a super-set of READ_WRITE?

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/TokenProcedures.java
```diff
@@ -0,0 +1,60 @@
...
+            throws IllegalTokenNameException    {
+        tx.acquireStatement().tokenWriteOperations().relationshipTypeGetOrCreateForName( newRelationshipType );
+    }
+
+    @Description( "Create a Property" )
+    @Procedure( name = "db.createProperty", mode = TOKEN )
+    public void createProperty( @Name( "newProperty" ) String newProperty ) throws IllegalTokenNameException
+    {
+        tx.acquireStatement().tokenWriteOperations().propertyKeyGetOrCreateForName( newProperty );
+    }
```
chrisvest, line 58: We need this so a DBA can prepare the schema items in advance (for people with READ and READ_WRITE modes, presumably) without adding any data to the database?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -135,6 +135,7 @@ final KeyReadOperations tokenRead()
 
     final KeyWriteOperations tokenWrite()
     {
+        statement.assertAllows( AccessMode::allowsTokenCreates, "Token create" );
         return operations.keyWriteOperations();
```
chrisvest, line 139: This `"Token create"` shouldn't really be hard-coded like this, should it?

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/Procedures.java
```diff
@@ -217,4 +218,14 @@ public void start() throws Throwable
         // And register built-in procedures
         builtin.accept( this );
     }
+
+    public void setAllowWriteTokenCreate( boolean allowWriteTokenCreate )
```
chrisvest: Could this (field, parameter, methods...) be called `allowCreateToken` or `tokenCreatePermitted` instead? The current name is a bit of a mouthful.

community/kernel/src/main/java/org/neo4j/procedure/Mode.java
```diff
@@ -25,5 +25,6 @@
     WRITE,
     SCHEMA,
     DBMS,
-    DEFAULT
+    DEFAULT,
+    TOKEN
```
chrisvest: This is the second enum called `Mode`, third if we include `AccessMode`. Could these be collapsed?

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/ReflectiveProcedureCompiler.java
```diff
@@ -242,6 +242,8 @@ private Mode mode( org.neo4j.procedure.Mode incoming )
             return Mode.DBMS;
         case SCHEMA:
             return Mode.SCHEMA_WRITE;
+        case TOKEN:
+            return Mode.TOKEN;
         case WRITE:
             return Mode.READ_WRITE;
         default:
```
chrisvest: This mapping smells.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -1524,16 +1541,40 @@ public void unregisterExecutingQuery( ExecutingQuery executingQuery )
...
+        return callProcedure( name, input, new RestrictedAccessMode( tx.securityContext().mode(), write ) );
     }
 
     @Override
     public RawIterator<Object[],ProcedureException> procedureCallWriteOverride( QualifiedName name, Object[] input )
             throws ProcedureException
     {
+        AccessMode write =
+                procedures.isAllowWriteTokenCreate() ? AccessMode.Static.TOKEN_WRITE : AccessMode.Static.WRITE;
+        return callProcedure( name, input, new OverriddenAccessMode( tx.securityContext().mode(), write ) );
```
chrisvest: Could `isAllowWriteTokenCreate()` directly return the `AccessMode` we need here? As a "baseline access mode"?

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/Procedures.java
chrisvest: Also, `Procedures` is exposed through `DepedencyResolver` in `DataSourceModule`, so user provided procedures could potentially change this setting.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
```diff
@@ -91,6 +91,9 @@ public void setup( Dependencies dependencies ) throws KernelException
             );
         life.add( securityLog );
 
+        Boolean allowTokenCreate = config.get( SecuritySettings.allow_publisher_create_token );
```
chrisvest: I'd prefer to see a lower-case `boolean` here, to make it clear that we really don't expect to see a `null` pop out of this setting. Also, it makes auto-unboxing only happen once.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/PredefinedRolesBuilder.java
```diff
@@ -28,14 +28,15 @@
...
 import static org.neo4j.server.security.enterprise.auth.plugin.api.PredefinedRoles.*;
 
 public class PredefinedRolesBuilder implements RolesBuilder
 {
     public static final Map<String,SimpleRole> roles = staticBuildRoles();
 
-    public static Map<String,SimpleRole> staticBuildRoles()
+    private static Map<String,SimpleRole> staticBuildRoles()
     {
         Map<String, SimpleRole> roles = new LinkedHashMap<>( 4 );
```
chrisvest: Why `LinkedHashMap`? Why does this method put a _mutable_ map in a `public static final` field of a `public` class? I would recommend wrapping the returned map in `Collections.unmodifiableMap`, and have the inner map be a `ConcurrentHashMap` because `setAllowPublisherTokenCreate` can modify it at runtime.

Also, how does this being a static resource interact when multiple databases, with different configurations, are started up in the same Java process?

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/PredefinedRolesBuilder.java
```diff
@@ -46,6 +47,7 @@
         SimpleRole architect = new SimpleRole( ARCHITECT );
         architect.add( new WildcardPermission( "schema:*" ) );
         architect.add( new WildcardPermission( "data:*" ) );
+        architect.add( new WildcardPermission( "token:*" ) );
```
chrisvest: These tokens could be constants. Would slightly speed up the Set operations inside `SimpleRole` when things are added and deleted.

chrisvest (changes requested):
I have done my first review round. It may be that I'm not terribly familiar with the security code, but I have a number of questions. And one or two things I think need more work.

chrisvest (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/Procedures.java
```diff
@@ -217,4 +219,25 @@ public void start() throws Throwable
         // And register built-in procedures
         builtin.accept( this );
     }
+
+    private boolean changed = false;
+
+    public void writerCreateToken( boolean allow )
+    {
+        if ( !changed )
```
craigtaverner, line 227: This seems to be protecting against people abusing the dependency resolver to enable token writes from within procedures. However, that assumes access to GraphDatabaseAPI which already allows them to bypass security entirely, so we are perhaps protecting unnecessarily. Better to remove access to GraphDatabaseAPI. Although that can be part of another PR.

craigtaverner (approved):

PR #8626, opened 2017-01-16
Full stack refactor of IndexDescriptor usage

fickludd:
The main part of this PR refactors method calls across the stack from directly passing two `int`s representing the labelId and propertyId of and index, to instead pass objects that encapsulate these values, the new `NodePropertyDescriptor` (and `RelationshipPropertyDescriptor` for relationship property existence constraints). These are similar to the `IndexDescriptor` which has been moved and refactored to now only represent existing indexes. This work has been done for all indexes and constraints and is in preparation for composite indexes and composite constraints, which are due for 3.2.

To maintain backwards compatibility for running Cypher 2.3 and 3.1 within Neo4j 3.2, we have to keep a very limited version of the old `org.neo4j.kernel.api.index.IndexDescriptor`. This is because of a mistake in the Cypher 2.3 implementation where this IndexDescriptor is used directly instead of mapped in the cypher-kernel SPI. We are making patches for 2.3 and 3.1 that will allow us to remove this old class once merged.


community/kernel/src/main/java/org/neo4j/kernel/api/schema/IndexDescriptorFactory.java
```diff
@@ -0,0 +1,78 @@
...
+
+import static org.neo4j.kernel.impl.coreapi.schema.PropertyNameUtils.getOrCreatePropertyKeyIds;
+import static org.neo4j.kernel.impl.coreapi.schema.PropertyNameUtils.getPropertyKeyIds;
+
+/**
+ * Factory methods for creating IndexDescriptors that are either single property or composite.
+ */
+public class IndexDescriptorFactory
+{
+    public static IndexDescriptor from( NodePropertyDescriptor descriptor )
```
MishaDemianenko: Just a reminder note about from factory method that we have talked about: name and place for optimisations

community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
```diff
@@ -153,7 +155,7 @@ void checkRelationshipPropertyExistenceRule( RelationshipPropertyExistenceConstr
         public void checkIndexRule( IndexRule rule, DynamicRecord record, RecordAccess records,
                 CheckerEngine<DynamicRecord,ConsistencyReport.SchemaConsistencyReport> engine )
         {
-            checkLabelAndPropertyRule( rule, rule.getPropertyKey(), record, records, engine );
+            checkLabelAndPropertyRule( rule, rule.descriptor(), record, records, engine );
```
MishaDemianenko: it's not quite related to current set of changes, just a note since you already here: it's quite strange that we passing rule as first parameter and as a second we give it a descriptor that we just take from rule, looks like second param can be simply removed here

community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
```diff
@@ -256,11 +258,22 @@ public void checkRelationshipPropertyExistenceRule( RelationshipPropertyExistenc
...
 
-    private void checkLabelAndPropertyRule( SchemaRule rule, int propertyKey, DynamicRecord record,
+    private void checkLabelAndPropertyRule( SchemaRule rule, EntityPropertyDescriptor descriptor, DynamicRecord record,
             RecordAccess records, CheckerEngine<DynamicRecord,ConsistencyReport.SchemaConsistencyReport> engine )
     {
         engine.comparativeCheck( records.label( rule.getLabel() ), VALID_LABEL );
-        engine.comparativeCheck( records.propertyKey( propertyKey ), VALID_PROPERTY_KEY );
+        if ( descriptor instanceof NodePropertyDescriptor && ((NodePropertyDescriptor) descriptor).isComposite() )
+        {
+            int[] propertyKeys = descriptor.getPropertyKeyIds();
```
MishaDemianenko: looking into usage of `getPropertyKeyIds ` what if instead of throwing exception in case of single property we will return array of with one element then those kind of if statements that depending from composition of property descriptors will be simplified.
WDYT?

community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
```diff
@@ -256,11 +258,22 @@ public void checkRelationshipPropertyExistenceRule( RelationshipPropertyExistenc
         }
     }
 
-    private void checkLabelAndPropertyRule( SchemaRule rule, int propertyKey, DynamicRecord record,
+    private void checkLabelAndPropertyRule( SchemaRule rule, EntityPropertyDescriptor descriptor, DynamicRecord record,
             RecordAccess records, CheckerEngine<DynamicRecord,ConsistencyReport.SchemaConsistencyReport> engine )
     {
         engine.comparativeCheck( records.label( rule.getLabel() ), VALID_LABEL );
-        engine.comparativeCheck( records.propertyKey( propertyKey ), VALID_PROPERTY_KEY );
+        if ( descriptor instanceof NodePropertyDescriptor && ((NodePropertyDescriptor) descriptor).isComposite() )
```
tinwelint: Could this be done in a less instanceof-y way? Like letting `EntityPropertyDescriptor` have `isComposite()`, and perhaps `getPropertyKeys()` on it directly?

community/kernel/src/main/java/org/neo4j/kernel/api/constraints/IndexBackedConstraint.java
```diff
@@ -0,0 +1,27 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api.constraints;
+
+import org.neo4j.kernel.api.schema.IndexDescriptor;
+
+public interface IndexBackedConstraint
```
tinwelint: It's always nice with at least class-level javadoc

community/kernel/src/main/java/org/neo4j/kernel/api/schema/EntityPropertyDescriptor.java
```diff
@@ -0,0 +1,131 @@
...
+    {
+        return propertyKeyId;
+    }
+
+    /**
+     * @return property key token ids this descriptor is for.
+     */
+    public int[] getPropertyKeyIds()
+    {
+        throw new UnsupportedOperationException( "Cannot get multiple property Ids of single property descriptor" );
```
tinwelint, line 90: Perhaps let this method be implemented by stuffing the `propertyKeyId` into an array. This would get rid of the `propertyIdText` method, which I'm not a fan of.

community/graphdb-api/src/main/java/org/neo4j/graphdb/schema/DefinitionWithProperties.java
```diff
@@ -0,0 +1,25 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.graphdb.schema;
+
+public interface DefinitionWithProperties
```
tinwelint: Some class-level javadoc would be nice

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
```diff
@@ -131,11 +131,17 @@ private void validateNode( long nodeId ) throws ConstraintValidationKernelExcept
...
 
                         // Check if this node has the mandatory property set
-                        if ( !propertyKeyIds.contains( constraint.propertyKey() ) )
+                        NodePropertyDescriptor descriptor = ((NodePropertyExistenceConstraint) constraint).descriptor();
+                        if ( descriptor.isComposite() )
                         {
-                            throw new NodePropertyExistenceConstraintViolationKernelException(
-                                    ((NodePropertyExistenceConstraint) constraint).label(),
-                                    constraint.propertyKey(), nodeId );
+                            for ( int propertyKey : descriptor.getPropertyKeyIds() )
```
MishaDemianenko, line 137: please see my comment about getPropertyKeyIds, in case if we still what to have separate branch for case of multiple properties that can be optimised to work on top of array existing keys and keys from descriptor and avoid contains for each separate key.

community/kernel/src/main/java/org/neo4j/kernel/api/schema/CompositeIndexDescriptor.java
```diff
@@ -0,0 +1,37 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api.schema;
+
+import org.neo4j.storageengine.api.schema.SchemaRule;
+
+import static java.lang.String.format;
+
+/**
+ * Description of a single index.
```
tinwelint: Funny how `single` and `composite` trips over each other when reading this. Also it says nothing about multiple property keys, is it required that a composite index descriptor has multiple keys or can it also handle one? If it can do we need to have multiple classes for this?

community/graphdb-api/src/main/java/org/neo4j/graphdb/schema/ConstraintDefinition.java
```diff
@@ -28,7 +28,7 @@
  * <b>Note:</b> This interface is going to be changed/removed in next major release to better cope with node and
```
MishaDemianenko: java doc is wrong here

community/graphdb-api/src/main/java/org/neo4j/graphdb/schema/DefinitionWithProperties.java
```diff
@@ -0,0 +1,25 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.graphdb.schema;
+
```
MishaDemianenko: java docs is missing, also it looks like strange interface to have - there is nothing here about definitions of some sort, only property keys - it looks like a PropertyContainer or something, but not a definition.
Both of child interfaces: `ConstraintDefinition` and `IndexDefinition` override `Iterable<String> getPropertyKeys()`why?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
```diff
@@ -174,23 +180,23 @@ public void load( List<SchemaRule> schemaRuleIterator )
...
         public IndexDescriptor getDescriptor()
         {
             return descriptor;
         }
 
-        public long getId()
-        {
-            return id;
-        }
+//        public long getId()
```
tinwelint: Can be removed?

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/MultiPropertyConstraintDefinition.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.List;
+
+import org.neo4j.graphdb.schema.ConstraintDefinition;
+import org.neo4j.graphdb.schema.ConstraintType;
+import org.neo4j.graphdb.schema.IndexDefinition;
+
+import static java.util.Arrays.asList;
+import static java.util.Objects.requireNonNull;
+
+abstract class MultiPropertyConstraintDefinition extends PropertyConstraintDefinition
```
tinwelint, line 33: Javadoc

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
```diff
@@ -0,0 +1,173 @@
...
+import org.neo4j.kernel.api.exceptions.PropertyKeyIdNotFoundKernelException;
+import org.neo4j.kernel.api.exceptions.schema.IllegalTokenNameException;
+import org.neo4j.kernel.api.exceptions.schema.SchemaRuleNotFoundException;
+import org.neo4j.kernel.api.schema.IndexDescriptor;
+import org.neo4j.kernel.impl.api.operations.KeyReadOperations;
+
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+
+public class PropertyNameUtils
```
tinwelint, line 42: Javadoc, even would be nice to take some time to javadoc the utility methods in here

community/kernel/src/main/java/org/neo4j/kernel/api/constraints/IndexBackedConstraint.java
MishaDemianenko: java doc is missing and it's really hard to see why do we need to have this interface and in this particular form

community/kernel/src/main/java/org/neo4j/storageengine/api/EntityType.java
```diff
@@ -24,5 +24,19 @@
...
+    GRAPH( "graph" );
+
+    private final String typeDescriptor;
+
+    EntityType( String typeDescriptor )
+    {
+        this.typeDescriptor = typeDescriptor;
+    }
+
+    public String getTypeDescriptor()
```
tinwelint: Not sure `typeDescriptor` is a good name. Better name and some javadoc too?

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
```diff
@@ -1105,46 +1112,63 @@ void forceFlushChanges()
...
+        {
+            ArrayList<Integer> propertyKeyIds = new ArrayList<>();
+            properties.forEach( index -> propertyKeyIds.add( getOrCreatePropertyKeyId( index ) ) );
+            return propertyKeyIds.stream().mapToInt( i -> i ).toArray();
+
+        }
+
+        private int[] getOrCreatePropertyKeyIds( String[] propertyKeys )
+        {
+            return getOrCreatePropertyKeyIds( Arrays.asList( propertyKeys ) );
```
tinwelint: Even if it perhaps doesn't matter since not very many rules are created, this is production code, after all and there's some hefty garbage making going on here: String[] --> List<String> --> List<String> (again, also creating a lambda instance) --> Stream --> String[]. We should be able to do better, it's becomes a bad habit to litter too much.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/EnterpriseConstraintSemantics.java
```diff
@@ -59,48 +61,67 @@ protected PropertyConstraint readNonStandardConstraint( PropertyConstraintRule r
...
         {
             try ( Cursor<NodeItem> cursor = allNodes.next() )
             {
                 NodeItem node = cursor.get();
-                if ( !node.hasProperty( propertyKey ) )
+                if ( descriptor.isComposite() )
                 {
-                    throw createConstraintFailure( new NodePropertyExistenceConstraintVerificationFailedKernelException(
-                            new NodePropertyExistenceConstraint( label, propertyKey ), node.id() ) );
+                    for ( int propertyKey : descriptor.getPropertyKeyIds() )
```
tinwelint, line 86: Was just thinking... so many places that now need to do this checking whether or not a descriptor is composite and loop if true, otherwise do single statement (copied code).

Would it be better if the descriptor accepted a visitor instead? Something like:

```
descriptor.visitPropertyKeys( propertyKey -> validateNodePropertyExistenceConstraint( node, propertyKey, descriptor ) );
```

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDescriptor.java
```diff
@@ -19,25 +19,28 @@
...
- * @see SchemaRule
+ * This is a new version of the old IndexDescriptor interface created here only to allow
+ * older versions of Cypher to run on newer kernels. This is because Cypher supports older
+ * cypher-compiler modules to be instantiated and run within newer versions of Neo4j, and they
+ * assume certain kernel API's remain unchanged. However, the IndexDescriptor was refactored
+ * considerably during the 3.2 development process to allow for composite indexes and constraints.
+ * We should remove this class as soon as older Cypher compilers are re-released without any
+ * kernel dependency on this.
+ * <p>
+ * //TODO: Delete this class!
```
MishaDemianenko: when is the plan to do that? we have card for that somewhere?

Class should be marked as deprecated at least.

community/kernel/src/main/java/org/neo4j/kernel/api/schema/CompositeIndexDescriptor.java
MishaDemianenko: this javadoc does not give any highlights what makes this class composite and what it brings with that

tinwelint (changes requested):
As you mentioned it's pretty hard to properly review this big change and surely a single reviewer will surely miss things, both big and small. I added some comments though.

I think it looks OK, perhaps if number of classes needed to convey label/propertyKey information around in the stack could be reduced somehow it would be easier to understand. Also in that vein if usage of the descriptors could be more unified so that users of them could avoid checking instanceof or always interact with them in two ways, composite or non-composite everywhere, that would be great.

community/kernel/src/main/java/org/neo4j/kernel/api/schema/EntityPropertyDescriptor.java
```diff
@@ -0,0 +1,131 @@
...
+package org.neo4j.kernel.api.schema;
+
+import org.neo4j.kernel.api.TokenNameLookup;
+import org.neo4j.storageengine.api.EntityType;
+import org.neo4j.storageengine.api.schema.SchemaRule;
+
+import static java.lang.String.format;
+
+/**
+ * Description of an object that contains one entity defined by an int, and one or more properties defined by an int
```
MishaDemianenko, line 29: those are not really properties itself but property keys

community/kernel/src/main/java/org/neo4j/kernel/api/schema/EntityPropertyDescriptor.java
```diff
@@ -0,0 +1,131 @@
...
+        throw new UnsupportedOperationException( "Cannot get multiple property Ids of single property descriptor" );
+    }
+
+    @Override
+    public String toString()
+    {
+        return format( ":%s[%d](property[%d])", entityType().getTypeDescriptor(), entityId, propertyKeyId );
+    }
+
+    public String propertyIdText()
```
MishaDemianenko: lets rename it to `propertyKeyIdText` to be consistent

community/kernel/src/main/java/org/neo4j/kernel/api/schema/IndexDescriptor.java
```diff
@@ -0,0 +1,39 @@
...
+
+import static java.lang.String.format;
+
+/**
+ * Description of a single index based on one label and one or more properties.
+ * @see SchemaRule
+ */
+public interface IndexDescriptor
+{
+    int getLabelId();
```
MishaDemianenko: it would be nice to have a javadoc explaining what methods of this interface should do, especially taking into account presence of confusing  `getPropertyKeyId()` and `getPropertyKeyIds()`

community/kernel/src/main/java/org/neo4j/kernel/api/schema/IndexDescriptorFactory.java
```diff
@@ -0,0 +1,78 @@
...
+        return propertyKeyIds.length > 1 ? new NodeMultiPropertyDescriptor( labelId, propertyKeyIds )
+                                         : new NodePropertyDescriptor( labelId, propertyKeyIds[0] );
+    }
+
+    public static NodePropertyDescriptor getTokens( ReadOperations readOperations, IndexDefinition indexDefinition )
+    {
+        int labelId = readOperations.labelGetForName( indexDefinition.getLabel().name() );
+        int[] propertyKeyIds = getPropertyKeyIds( readOperations, indexDefinition.getPropertyKeys() );
+
+        return propertyKeyIds.length > 1 ? new NodeMultiPropertyDescriptor( labelId, propertyKeyIds )
```
MishaDemianenko: that is `getNodePropertyDescriptor` isn't it?

community/kernel/src/main/java/org/neo4j/kernel/api/schema/IndexDescriptorFactory.java
```diff
@@ -0,0 +1,78 @@
...
+        return new NodePropertyDescriptor( labelId, propertyKeyId );
+    }
+
+    public static NodePropertyDescriptor getOrCreateTokens( SchemaWriteOperations schemaWriteOperations,
+            IndexDefinition indexDefinition ) throws IllegalTokenNameException, TooManyLabelsException
+    {
+        int labelId = schemaWriteOperations.labelGetOrCreateForName( indexDefinition.getLabel().name() );
+        int[] propertyKeyIds = getOrCreatePropertyKeyIds( schemaWriteOperations, indexDefinition );
+
+        return propertyKeyIds.length > 1 ? new NodeMultiPropertyDescriptor( labelId, propertyKeyIds )
```
MishaDemianenko: that is `getNodePropertyDescriptor` isn't it?

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
```diff
@@ -0,0 +1,140 @@
...
+     */
+    public int[] getPropertyKeyIds()
+    {
+        return propertyKeyIds;
+    }
+
+    @Override
+    public String propertyIdText()
+    {
+        return Arrays.stream( propertyKeyIds ).mapToObj( id -> Integer.toString( id ) )
```
MishaDemianenko: `return Arrays.toString( propertyKeyIds );`

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
```diff
@@ -0,0 +1,140 @@
...
+                .collect( Collectors.joining( "," ) );
+    }
+
+    public String propertyNameText( TokenNameLookup tokenNameLookup )
+    {
+        return Arrays.stream( propertyKeyIds ).mapToObj( id -> tokenNameLookup.propertyKeyGetName( id ) )
```
MishaDemianenko: `tokenNameLookup::propertyKeyGetName` should be used instead of lamda

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
```diff
@@ -0,0 +1,140 @@
...
+     * @return a user friendly description of what this index indexes.
+     */
+    public String userDescription( TokenNameLookup tokenNameLookup )
+    {
+        return format( ":%s(%s)", tokenNameLookup.labelGetName( getLabelId() ),
+                propertyNameText( tokenNameLookup ) );
+    }
+
+    @Override
+    public int compareTo( EntityPropertyDescriptor other )
```
MishaDemianenko, line 115: compareTo should return -1,0 or 1 as result it looks like results here can be a bit different

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Description of a single label combined with multiple properties.
+ */
+public class NodeMultiPropertyDescriptor extends NodePropertyDescriptor
+{
+    private final int[] propertyKeyIds;
+
+    public NodeMultiPropertyDescriptor( int labelId, int[] propertyKeyIds )
+    {
+        super( labelId, -1 );
```
MishaDemianenko, line 38: this looks a bit strange that we need to pass -1 since we do not have single property key id. Maybe hierarchy can be changed to avoid that?

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
```diff
@@ -0,0 +1,140 @@
...
+                        if ( cmp != 0 )
+                        {
+                            return cmp;
+                        }
+                    }
+                }
+            }
+            return cmp;
+        }
+        return -1;  //TODO: We can compare single and composite indexes, by sorting one type always above the other
```
MishaDemianenko: TODO part: do we really need to do that?

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -79,8 +81,9 @@ public GraphResult buildSchemaGraph()
                 ArrayList<String> indexes = new ArrayList<>();
                 while ( indexDescriptorIterator.hasNext() )
                 {
-                    indexes.add( statementTokenNameLookup
-                            .propertyKeyGetName( indexDescriptorIterator.next().getPropertyKeyId() ) );
+                    IndexDescriptor index = indexDescriptorIterator.next();
+                    String[] propertyNames = PropertyNameUtils.getPropertyKeys( statementTokenNameLookup, index.descriptor() );
+                    indexes.add( Arrays.stream( propertyNames ).collect( Collectors.joining( "," ) ) );
```
MishaDemianenko: Arrays.toString? 
Please avoid using streams just because it's possible to that using them, there are tons of handy and shiny methods around lest use them instead!

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexUpdaterMap.java
```diff
@@ -43,20 +48,35 @@
...
     }
 
-    public IndexUpdater getUpdater( IndexDescriptor descriptor )
+    List<IndexUpdater> getUpdaters( int labelId, int propertyKeyId )
+    {
+        IndexDescriptor key = IndexDescriptorFactory.from( new NodePropertyDescriptor( labelId, propertyKeyId ) );
+        IndexUpdater updater = getUpdater( key );
+        if ( updater != null )
+        {
+            return Arrays.asList(updater);
```
MishaDemianenko: that will create list of updater for any update that will come to indexing service, is that for future modification mainly?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
```diff
@@ -174,23 +180,23 @@ public void load( List<SchemaRule> schemaRuleIterator )
     private static class CommittedIndexDescriptor
     {
         private final IndexDescriptor descriptor;
-        private final long id;
+//        private final long id;
```
MishaDemianenko: commented out lines

community/kernel/src/main/java/org/neo4j/kernel/impl/api/DataIntegrityValidatingStatementOperations.java
```diff
@@ -227,22 +230,26 @@ public void constraintDrop( KernelStatement state, RelationshipPropertyConstrain
         schemaWriteDelegate.constraintDrop( state, constraint );
     }
 
-    private void checkIndexExistence( KernelStatement state, OperationContext context, int labelId, int propertyKey )
+    private void checkIndexExistence( KernelStatement state, OperationContext context,
+            NodePropertyDescriptor descriptor )
             throws AlreadyIndexedException, AlreadyConstrainedException
     {
-        for ( IndexDescriptor descriptor : loop( schemaReadDelegate.indexesGetForLabel( state, labelId ) ) )
+        //TODO: Consider using schemaReadDelegate.indexGetForLabelAndPropertyKey(state, indexDescriptor)
```
MishaDemianenko: TODO should be addressed

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/IndexDefinitionImpl.java
```diff
@@ -103,17 +108,30 @@ public boolean equals( Object obj )
...
+        return "IndexDefinition[label:" + label + ", on:" +
+               Arrays.stream( propertyKeys ).collect( Collectors.joining( "," ) ) + "]";
     }
 
     protected void assertInUnterminatedTransaction()
     {
         actions.assertInOpenTransaction();
     }
+
+    public static int hashCode(Label label, String[] propertyKeys)
```
MishaDemianenko: don't think you need this, you can use handy methods from Objects and Arrays to calculate hashCode more easily.

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/InternalSchemaActions.java
```diff
@@ -36,24 +36,24 @@
  */
 public interface InternalSchemaActions
 {
-    IndexDefinition createIndexDefinition( Label label, String propertyKey );
+    IndexDefinition createIndexDefinition( Label label, String[] propertyKey );
```
MishaDemianenko: i think propertyKey should be String... instead

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/InternalSchemaActions.java
```diff
@@ -36,24 +36,24 @@
...
-    void dropIndexDefinitions( Label label, String propertyKey );
+    void dropIndexDefinitions( IndexDefinition indexDefinition );
 
-    ConstraintDefinition createPropertyUniquenessConstraint( Label label, String propertyKey )
+    ConstraintDefinition createPropertyUniquenessConstraint( IndexDefinition indexDefinition )
             throws IllegalTokenNameException, TooManyLabelsException, CreateConstraintFailureException,
                    AlreadyConstrainedException, AlreadyIndexedException;
 
-    ConstraintDefinition createPropertyExistenceConstraint( Label label, String propertyKey )
+    ConstraintDefinition createPropertyExistenceConstraint( Label label, String[] propertyKey )
```
MishaDemianenko: i think propertyKey should be String... instead

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/MultiPropertyConstraintDefinition.java
```diff
@@ -0,0 +1,74 @@
...
+    {
+        requireNonNull( array );
+        if ( array.length < 1 )
+        {
+            throw new IllegalArgumentException( "Property constraint must have at least one property" );
+        }
+        for ( String field : array )
+        {
+            if ( field == null )
+            { throw new NullPointerException(); }
```
MishaDemianenko: i think that is not really helpful exception to get if you are the user of this class, would be so much nicer to get some additional description and explanation what went wrong

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/MultiPropertyConstraintDefinition.java
```diff
@@ -0,0 +1,74 @@
...
+        this.propertyKeys = requireNonEmpty( propertyKeys );
+    }
+
+    protected MultiPropertyConstraintDefinition( InternalSchemaActions actions, IndexDefinition indexDefinition )
+    {
+        super( actions );
+
+        List<String> indexList = new ArrayList<>();
+        indexDefinition.getPropertyKeys().forEach( index -> indexList.add( index ) );
+        String[] propertyKeys = indexList.toArray( new String[indexList.size()] );
```
MishaDemianenko: looks like this line will gonna make the trick: `String[] propertyKeys = Iterables.asArray( String.class, indexDefinition.getPropertyKeys() );` ?

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/NodeConstraintDefinition.java
```diff
@@ -60,13 +72,27 @@ public boolean equals( Object o )
...
+            return "(" + Arrays.stream( propertyKeys ).map( p -> nodeVariable + "." + p )
+                    .collect( Collectors.joining( "," ) ) + ")";
+        }
     }
 
     @Override
     public int hashCode()
     {
-        return 31 * label.name().hashCode() + propertyKey.hashCode();
+        return IndexDefinitionImpl.hashCode( label, propertyKeys );
```
MishaDemianenko: no please do not do that, if you need a helper class for hashCode evaluation, which i do not think is the case here, we should introduce helper class that will do that and not using something from `IndexDefinitionImpl`

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/NodePropertyExistenceConstraintDefinition.java
```diff
@@ -48,7 +54,15 @@ public ConstraintType getConstraintType()
     @Override
     public String toString()
     {
-        return format( "ON (%1$s:%2$s) ASSERT exists(%1$s.%3$s)",
-                label.name().toLowerCase(), label.name(), propertyKey );
+        if ( propertyKeys.length == 1 )
+        {
+            return format( "ON (%1$s:%2$s) ASSERT exists(%3$s)",
```
MishaDemianenko, line 59: i would assume that hierarchy that we introduce should allow us to avoid those kind of code that depends from propertyKeys.length, isn't it?

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
MishaDemianenko, line 42: private constructor is missing

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
```diff
@@ -0,0 +1,173 @@
...
+{
+    public static String getPropertyKeyNameAt( DefinitionWithProperties index, int propertyKeyIndex )
```
MishaDemianenko: method can be private

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
```diff
@@ -0,0 +1,173 @@
...
+            if ( propertyKeyIndex == 0 )
+            {
+                return propertyKey;
+            }
+            propertyKeyIndex--;
+        }
+        return null;
+    }
+
+    public static IndexDescriptor getIndexDescriptor( ReadOperations readOperations, IndexDefinition index )
```
MishaDemianenko: method can be package visible

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
```diff
@@ -0,0 +1,173 @@
...
+
+        for ( int i = 0; i < propertyKeyIds.length; i++ )
+        {
+            if ( propertyKeyIds[i] == KeyReadOperations.NO_SUCH_PROPERTY_KEY )
+            {
+                throw new NotFoundException(
+                        format( "Property key %s not found", getPropertyKeyNameAt( index, i ) ) );
+            }
+        }
+        return propertyKeyIds.length > 1 ? new NodeMultiPropertyDescriptor( labelId, propertyKeyIds )
```
MishaDemianenko: Why not to use IndexDescriptorFactory in here?

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
```diff
@@ -0,0 +1,173 @@
...
+                                         : new NodePropertyDescriptor( labelId, propertyKeyIds[0] );
+    }
+
+    public static IndexDescriptor getIndexDescriptor( ReadOperations readOperations, Label label,
```
MishaDemianenko: method can be package visible

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
```diff
@@ -0,0 +1,173 @@
...
+        for ( int i = 0; i < propertyKeys.length; i++ )
+        {
+            propertyKeyIds[i] = statement.propertyKeyGetForName( propertyKeys[i] );
+        }
+        return propertyKeyIds;
+    }
+
+    public static int[] getPropertyKeyIds( ReadOperations statement, Iterable<String> propertyKeys )
+    {
+        ArrayList<Integer> propertyKeyIds = new ArrayList<>();
```
MishaDemianenko: isn't it all what you need : 
`return Iterables.stream( propertyKeys )
                         .mapToInt( statement::propertyKeyGetForName )
                        .toArray();
`
?

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
MishaDemianenko, line 42: this class in general have a lot of duplicated code for case when `IndexDefinition` or `String[] propertyKeys` and any methods that rely on dependent parameters from them are duplicated now, that should be cleaned up

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/SchemaImpl.java
```diff
@@ -122,21 +125,26 @@ public IndexCreator indexFor( Label label )
         }
     }
 
+    //TODO: Consider moving to PropertyNameUtils or IndexDescriptorFactory or similar
```
MishaDemianenko: todo should be fixed

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/SinglePropertyConstraintDefinition.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.coreapi.schema;
+
+import static java.util.Collections.singleton;
+import static java.util.Objects.requireNonNull;
+
+abstract class SinglePropertyConstraintDefinition extends PropertyConstraintDefinition
```
MishaDemianenko, line 25: java doc

community/kernel/src/main/java/org/neo4j/kernel/impl/store/counts/keys/IndexKey.java
```diff
@@ -79,6 +70,17 @@ public boolean equals( Object other )
...
+
+    @Override
+    public int compareTo( CountsKey other )
+    {
+        if ( other instanceof IndexKey )
+        {
+            IndexKey that = (IndexKey) other;
+            return this.descriptor.descriptor().compareTo( that.descriptor.descriptor() );
+        }
+        return recordType().ordinal() - other.recordType().ordinal();
```
MishaDemianenko, line 84: compareTo contract is broken here

community/kernel/src/main/java/org/neo4j/kernel/impl/store/record/AbstractSchemaRule.java
```diff
@@ -22,21 +22,24 @@
...
+    public AbstractSchemaRule( long id, Kind kind, EntityPropertyDescriptor descriptor )
     {
         this.id = id;
         this.kind = kind;
+        this.descriptor = descriptor;
     }
 
     @Override
-    public long getId()
+    public final long getId()
```
MishaDemianenko, line 46: why those methods are final now? javadoc maybe?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/record/AbstractSchemaRule.java
```diff
@@ -54,24 +57,24 @@ public final Kind getKind()
...
     @Override
-    public boolean equals( Object o )
+    public final boolean equals( Object o )
     {
         if ( this == o )
         {
             return true;
         }
-        if ( o == null || getClass() != o.getClass() )
+        if ( o == null || o instanceof AbstractSchemaRule )
```
MishaDemianenko: looks like wrong condition here

community/kernel/src/main/java/org/neo4j/kernel/impl/store/record/UniquePropertyConstraintRule.java
```diff
@@ -20,58 +20,61 @@
...
 
-    private UniquePropertyConstraintRule( long id, int labelId, int[] propertyKeyIds, long ownedIndexRule )
+    private UniquePropertyConstraintRule( long id, NodePropertyDescriptor descriptor, long ownedIndexRule )
     {
-        super( id, labelId, Kind.UNIQUENESS_CONSTRAINT );
+        super( id, descriptor, Kind.UNIQUENESS_CONSTRAINT );
         this.ownedIndexRule = ownedIndexRule;
-        assert propertyKeyIds.length == 1; // Only uniqueness of a single property supported for now
-        this.propertyKeyIds = propertyKeyIds;
+        //TODO: Find a better way of asserting this
```
MishaDemianenko: todo and commented out code

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
```diff
@@ -1105,46 +1112,63 @@ void forceFlushChanges()
 
     private class BatchSchemaActions implements InternalSchemaActions
     {
+        private int[] getOrCreatePropertyKeyIds( Iterable<String> properties )
+        {
+            ArrayList<Integer> propertyKeyIds = new ArrayList<>();
```
MishaDemianenko: too much dances around, there some examples where variants proposed how to do that kind of conversion

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
```diff
@@ -78,10 +78,22 @@ private static DocWithId reuseDocument( long nodeId )
     public static Document documentRepresentingProperty( long nodeId, Object value )
     {
         DocWithId document = reuseDocument( nodeId );
+        document.removeAllValueFields();
```
MishaDemianenko: why this was moved? some tests for that?

MishaDemianenko (changes requested):
Hooh, please find some comments!
For now it looks like that some work still to be done to avoid all cases when if branches are done based on number of keys or based on `instanceof` of passed definition

community/kernel/src/main/java/org/neo4j/kernel/api/schema/EntityPropertyDescriptor.java
craigtaverner: We actually wanted a runtime exception to help track down cases of when composite was used as non-composite and vice versa. Other models (which we see elsewhere in the code) rely too much on code correctness, so we avoided them.

However, I agree with your objection to this kind of solution. My position is only that it is certainly better than before, and while we could do even better, we'd prefer to do that in a second round. We are prototyping some other alternatives to this model that do a better job of removing these concerns, and will propose those later.

community/kernel/src/main/java/org/neo4j/kernel/api/schema/CompositeIndexDescriptor.java
craigtaverner: Currently we require that a composite index has multiple keys. There is even a check or two that will throw an exception if this is not the case. It might not be the best model, but it is a good intermediate state where it forces test failures for situations where we mix them up. I think we could improve this further, and have ideas for that, but suggest we do it in later PRs.

An earlier version of this work back in December simply generalized all indexes to composite, and the old behaviour was implied by the property array having length 1. We moved away from that model because it was harder to be sure that code that expected single property indexes didn't inadvertently use only one property of the list. Hence the current model which is stricter. And I think while we can improve the current model, going back to the previous one does not feel like an improvement.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
craigtaverner: Yes!

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
craigtaverner: Actually, these utility methods we are expecting to change quite a bit as we implement composite indexes. I think we should javadoc them once they stabilize. The same can be said of the IndexDescriptorFactory static methods.

community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
craigtaverner: Yes. We have ideas for that. Interestingly the current code represents the state after already refactoring away quite a few of these kinds of checks. We know have (I believe) few checks than stock 3.2. But this process can be taken further. We have some simple ideas that will improve this directly, and are prototyping some more fundamental changes to the descriptor approach that will reduce these checks a lot. I'd rather that work be done in later pull requests.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/EnterpriseConstraintSemantics.java
fickludd: I agree completely, and this is exactly what I've been exploring these last days. We are solving that in the next PR.

community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
craigtaverner: We actually did it that way first, and I commented above to Mattias about this. The current approach has the advantage of forcing runtime errors for when single and composite indexes get mixed up. It is just a way to enforce correct usage. However, I also see this a temporary or interim step while developing composite indexes to be sure single property indexes are tested as thoroughly as before. I think that as composite index features and tests for those features are created, we will simplify the descriptor framework further and aim to remove a lot of _if_ statements as a result.

community/graphdb-api/src/main/java/org/neo4j/graphdb/schema/ConstraintDefinition.java
craigtaverner: Aside from the fact that the comment is not introduced by our code, I actually think there is a reasonable chance that our next wave of refactoring might cause the comment to come true, or at least partly true. We could remove this, or just wait for the composite index work to deal with it.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
fickludd: We intend to remove the double `getPropertyKeyId` / `getPropertyKeyIds` in the next PR.

community/kernel/src/main/java/org/neo4j/kernel/api/schema/EntityPropertyDescriptor.java
fickludd: We are making property / propertyId distinctions better in the next PR.

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
fickludd: We're changing this class hierarchy in the next PR.

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
fickludd: That's nice! We prefer to keep it similar to the method below though, so we maintain the same string output.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
craigtaverner: Arrays.toString adds extra `[]` characters, but we will try `String.join()` instead which should be better.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/DataIntegrityValidatingStatementOperations.java
craigtaverner: This pattern of looking up be label first and then by properties is used in many places in the code. I thought it better to address those in a separate refactoring, so that the current PR does not go too far out of scope. I added the TODO as a reminder for that work.

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/MultiPropertyConstraintDefinition.java
craigtaverner: This was the previous behaviour

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/NodePropertyExistenceConstraintDefinition.java
fickludd: This will be addressed in an upcoming PR.

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
craigtaverner: We have cleaned up some of this, and have plans to cleanup more of this and the related IndexDescriptorFactory, but want the work on composite index behaviour to start because that will have a strong influence on the design.

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/SinglePropertyConstraintDefinition.java
craigtaverner: All the other classes that exist in this structure, and pre-exist our changes, do not have javadoc. We are also planning to restructure this more with the later composite index work, so it would be better to only document the final version.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/counts/keys/IndexKey.java
craigtaverner: We checked the docs at https://docs.oracle.com/javase/7/docs/api/java/lang/Comparable.html#compareTo(T) and it is clear that compareTo is allowed to return values other than -1, 0 and +1. They explicitly describe rules that show that non-normalized values can be normalized after the return from compareTo to achieve certain properties. Also the code we used here was pre-existing code in the code-base, so we are not introducing new behaviour.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/record/AbstractSchemaRule.java
fickludd: The id of a `SchemaRule` is a fixed property that should never be modified of overridden by subclasses, just like the `getKind()` method, which was already final.

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
craigtaverner: The compareTo contract allows any negative and positive numbers. https://docs.oracle.com/javase/7/docs/api/java/lang/Comparable.html#compareTo(T)

This code is based on previously existing code for IndexDescriptor which also did not require -1m 0 or 1.

community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
MishaDemianenko: ok agree

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
MishaDemianenko: ok

community/graphdb-api/src/main/java/org/neo4j/graphdb/schema/ConstraintDefinition.java
MishaDemianenko: please remove that or change, interface should be or `@Deprecated` or not, that javadoc part is just super confusing and wrong

community/kernel/src/main/java/org/neo4j/kernel/api/schema/EntityPropertyDescriptor.java
MishaDemianenko: ok

community/kernel/src/main/java/org/neo4j/kernel/api/schema/EntityPropertyDescriptor.java
MishaDemianenko: Comment still valid

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
MishaDemianenko: ok

community/kernel/src/main/java/org/neo4j/kernel/api/schema/NodeMultiPropertyDescriptor.java
MishaDemianenko: you are right, sorry misread Comparable java doc

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
MishaDemianenko: yes that's much better thx!

community/kernel/src/main/java/org/neo4j/kernel/impl/api/DataIntegrityValidatingStatementOperations.java
MishaDemianenko: do you have card for that somewhere?

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/NodePropertyExistenceConstraintDefinition.java
MishaDemianenko: ok

community/kernel/src/main/java/org/neo4j/kernel/impl/store/counts/keys/IndexKey.java
MishaDemianenko: yep, correct; same mistake from me as above sorry

community/kernel/src/main/java/org/neo4j/kernel/impl/store/record/AbstractSchemaRule.java
MishaDemianenko: Please add that to the java doc, so future maintainers aware about that!

community/collections/src/main/java/org/neo4j/helpers/collection/CastingIterator.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.Iterator;
+
+/**
+ * An iterator which filters another iterator, only letting items with certain
+ * criteria pass through. All iteration/filtering is done lazily.
+ *
+ * @param <T> the type of items in the iteration.
+ */
+public class CastingIterator<T extends A, A> extends PrefetchingIterator<T>
```
MishaDemianenko: We already have filtering iterator available in Iterators utility class that should be able to help to solve problem that this iterator suppose to solve right?

MishaDemianenko:
@fickludd @craigtaverner one new comment and couple of old, also as a general question how do we track all items that should be fixed as separate PR, just to be sure we do not loose any since there are quite a few of them?

community/collections/src/main/java/org/neo4j/helpers/collection/CastingIterator.java
fickludd: The `FilteringIterator` is almost the same, but it does not cast to the subtype, which is really nice for the calling code. Fixed javadoc (which was not correct) to clarify the difference.

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/PropertyNameUtils.java
fickludd: That has been added, but it didn't update github because it's just below these lines =).

fickludd: cardified this in the card `Remainers from index API change` on the security wall

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/SinglePropertyConstraintDefinition.java
fickludd: cardified this in the card `Remainers from index API change` on the security wall

community/kernel/src/main/java/org/neo4j/kernel/impl/api/DataIntegrityValidatingStatementOperations.java
fickludd: yup: https://trello.com/c/uLi338Tt/126-remainers-from-index-api-change

tinwelint:
About the requested changes that are due for other PRs, I can relate to this strategy in the native label scan store work and is the only reasonable thing to do if work has moved on beyond this PR already.

I also second the importance of keeping track of all these TODOs, perhaps adding proper TODOs in the code _in this PR_ would keep me and Misha happy :)

fickludd:
@MishaDemianenko added some javadocs, and linked to card on our wall where we track the remaining items. Everything directly related to improving the `IndexDescriptor`/`NodeDescriptor` classes is already been fixed in #8655, other things are tracked in the card `Remainers from index API change` on the security wall.

community/collections/src/main/java/org/neo4j/helpers/collection/CastingIterator.java
MishaDemianenko: @fickludd should we also add some method to `Iterators` that will provide usage of newly introduced instance-of-filtering-casting-Iterator to have one place where you should look for possible existent iterators?

MishaDemianenko:
@fickludd that card so far consist only from 3 items? are those all leftovers 😮  ?

fickludd:
@MishaDemianenko I think so, did I miss something? I did not put stuff directly related to descriptor classes as we have a separate PR for that which is almost ready (#8655), so we won't forget.

tinwelint:
@fickludd @craigtaverner have you looked into the (could-be-related) failing ` DumpCountsStoreTest`?

tinwelint (approved):

MishaDemianenko (approved):

PR #8632, opened 2017-01-16
Don't create test dirs in file tree

spacecowboy:
Avoids creating a `dbms/community/data/graph.db` directory which pollutes the build space.


MishaDemianenko (approved):
🥇 👍

PR #8643, opened 2017-01-17
Kernel transaction state handling update v2

MishaDemianenko:
Restore https://github.com/neo4j/neo4j/pull/8532 with following updates:
* update handling of termination status reset on transaction init
* prevent exception from client close cause transaction leaking and preventing shutdown
* update transaction event failure status handling

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
@@ -664,7 +664,7 @@ public class KernelTransactionImplementation implements KernelTransaction, TxSta
      */
     private void release()
     {
-        statementLocks.close();
+        Throwable lockCloseException = closeLocksSilently();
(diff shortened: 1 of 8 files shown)


tinwelint:
@MishaDemianenko The test failure ` ClusterTest.givenClusterWhenMasterGoesDownAndTxIsRunningThenDontWaitToSwitch` looks related, also reproduced locally. It looks like you've changed the surface (exception-wise), which may be a big deal becasue transient exceptions are documented to be specifically caught and retried certain ways. Please look through this.

tinwelint (changes requested):
@MishaDemianenko The test failure ClusterTest.givenClusterWhenMasterGoesDownAndTxIsRunningThenDontWaitToSwitch looks related, also reproduced locally. It looks like you've changed the surface (exception-wise), which may be a big deal becasue transient exceptions are documented to be specifically caught and retried certain ways. Please look through this.

MishaDemianenko:
@tinwelint updated

PR #8647, opened 2017-01-18
Added RPM package

### LINKED ISSUE #5100: RPM packages incorrectly depend on a specific version of Java
I am trying to installing neo4j RPMS from http://yum.neo4j.org
I am running Fedora 21

The install fails because of an explicit dependency on OpenJDK 1.7. It should depend on any JDK greater than or equal to 1.7.

The shell transcript makes it clear.

```
[root@richs:~]% cat <<EOF>  /etc/yum.repos.d/neo4j.repo
[neo4j]
name=Neo4j Yum Repo
baseurl=http://yum.neo4j.org
enabled=1
gpgcheck=1
EOF
[root@richs:~]% yum install neo4j
<snip yum cache refresh>
Resolving Dependencies
--> Running transaction check
---> Package neo4j.noarch 0:2.3.0-M02_1 will be installed
--> Processing Dependency: java-1.7.0-openjdk for package: neo4j-2.3.0-M02_1.noarch
--> Finished Dependency Resolution
Error: Package: neo4j-2.3.0-M02_1.noarch (neo4j)
           Requires: java-1.7.0-openjdk
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
[root@richs:~]% yum search openjdk
Loaded plugins: langpacks
================================== N/S matched: openjdk ==================================
java-1.8.0-openjdk.i686 : OpenJDK Runtime Environment
<snip other packages>
```

### LINKED ISSUE #5389: RPM does not mark `/etc/sysconfig/neo4j` as noreplace config file
The RPM spec file does not mark `/etc/sysconfig/neo4j` as a configuration file, so RPM makes no attempt to preserve the contents on upgrade. Since this is where the tuning for the JVM goes, it's kinda important that it not get wacked on upgrade.

Ideally, it will be marked as "noreplace" like the files in `/etc/neo4j`, so that, when modified, new versions are created as `/etc/sysconfig/neo4j.rpmnew`.

I can see where in `packaging/installer-linux/installer-rpm/pom.xml` this should happen approximately, but not having a build environment and not knowing Maven, I regret to say that I do not have a patch.

### LINKED ISSUE #5433: RPM incorrectly removes user/group on upgrade
The RPMs incorrectly removes the _neo4j_ user & group on upgrade, which breaks upgrades by leaving all of the files without an owner and group:

```
$ ls -l /var/lib/neo4j/
total 20340
-rw-r--r--. 1 496 497       11 Sep  8  2014 active_tx_log
drwxr-xr-x. 2 496 497     4096 Mar 11  2015 index/
-rw-r--r--. 1 496 497 20661455 Mar 26 11:53 messages.log
-rw-r--r--. 1 496 497       96 Mar 11  2015 neostore
-rw-r--r--. 1 496 497        9 Mar 11  2015 neostore.id
...
```

This is due to unconditionally removing the user & group in the _postuninstall_ scriptlet:

```
$ rpm -qp --scripts neo4j-2.2.5-1.noarch.rpm
...
postuninstall scriptlet (using /bin/sh):
/usr/bin/getent group neo4j > /dev/null 2>&1 && /usr/sbin/userdel neo4j
```

This should be conditional upon `$1 == 0`, which indicates a complete removal; see
https://fedoraproject.org/wiki/Packaging:ScriptletSnippets#Syntax

Alternatively, just don't remove the user & group automatically, as advised here:
https://fedoraproject.org/wiki/Packaging:UsersAndGroups?rd=Packaging/UsersAndGroups#Allocation_Strategies

### LINKED ISSUE #5484: RPM Java dependencies
I'm aware that RPMs are considered experimental (as per discussion in #5388). They should be the default way to install Neo4j outside of a container (as I said there).

Currently the Neo4j Java dependencies are for a specific 1.7 version of Java. However, 1.7 is obsolescent and the newest Linux distros only supply 1.8. If the RPM dependency were on the java capability, then it would accept any package that provided the java capability. There is a good discussion of this here: http://stevenengelhardt.com/2013/05/29/creating-a-virtual-java-rpm/ - where he explains that it's a little more complicated than I explained it. But still, it seems worthwhile to consider.

Specifically, I can't install Neo4j RPMs on Fedora 21 or 22 because of this issue. Fedora 21 was the basis for RHEL7.

When I install tar balls, the installer explains that this is not the optimal way to install the code, and I shouldn't be using it either. As far as I know, it's either tar balls or RPMs for a big chunk of Linux ;-).

### LINKED ISSUE #6079: RPM-provided init script should look for run-as user from sysconfig/neo4j
Please update the /etc/init.d/neo4j script to look for another user specified, e.g., in /etc/sysconfig/neo4j.

For example:
[/etc/sysconfig/neo4j]

```
export NEO4J_USER='runasuser'
```

[/etc/init.d/neo4j]

```
...
+ neo4j_user="${NEO4J_USER:-neo4j}";
...
start() {
...
- daemon --user neo4j --pidfile $pidfile $exec start
+ daemon --user ${neo4j_user} --pidfile $pidfile $exec start
```

### LINKED ISSUE #6142: neo4j RPM does not work with SuSE Linux
Hi,

The RPM for a Java process ought to work with SuSE Linux - except that the init script includes and uses functions from the RedHat-only script _/etc/rc.d/init.d/functions_. There are also issues with the repository which I'll file a separate bug report on.

### LINKED ISSUE #6148: There is *no* proper way to install Neo4j on RPM-based systems.
If you install from the tar ball, it says "this is not the right way to install Neo4j on your platform". If you use the RPM repository it gives several very serious (and problematic) warnings about it being experimental.

So, there is **no** proper/supported way to install Neo4j on RPM-based systems. 

95+% of all cloud instances are Linux-based. More than half of those run an RPM-based distribution. The most mission-critical organizations gravitate towards RHEL or SUSE Linux (both of which are RPM-based).

Professional, supportable products on these platforms _require_ high-quality RPMs, and a properly-managed RPM repository generated automatically from your production build system. Without these, your customers cannot properly manage Neo4j installations. It brands the product as amateurish.

Neo4j has neither of these. I've put references to a number of the related bugs below.

Moreover, you should (at least minimally) test your product from the RPMs on a number of platforms. Personally, I do this on my product using Docker - it makes it very straightforward.

The upshot of this for me is that I have potential customers who are reluctant to use _my_ product because of _your_ poor RPM support. I certainly understand why they feel this way. I should not have to package your product and create my own RPM repository just to get past your issue.

I appreciate that this may be a set of skills that you may not have in-house. The good news is that these skills are available out there in the real world.  But you need to spend the money and time to make this happen if you want to have production (meaning people-who-pay-you-money) Linux customers.
## Related issues
- neo4j RPM repository has digital signature issues #6143 
- neo4j RPM does not work with SuSE Linux #6142
- RPM-provided init script should look for run-as user from sysconfig/neo4j feature operability #6079 
- RPM Java dependencies feature operability #5484 
- RPM incorrectly removes user/group on upgrade bug operability #5433
- RPM does not mark `/etc/sysconfig/neo4j` as noreplace config file feature operability #5389 
- packages incorrectly depend on a specific version of Java 2.3 bug #5100

spacecowboy:
Initial RPM support. There might be one or two things left to fix but this will allow us to start tyre-kicking it.

Fixes #6148
Fixes #6142 (no platform specific functions are currently used)
Fixes #5433 (user is left alone on upgrade/uninstall)
Fixes #5389 (file no longer exists)
Fixes #5100 (neo4j requires java8 atm, it does not work with java9 or java7)
Fixes #6079 (no longer an issue with systemd service script)
Fixes #5484 (RPM depends on `java-1.8.0-headless` which should hopefully work with any version of java (openjdk/oracle). note that neo4j does not support java9 at this time)


new-packaging/Makefile
```diff
@@ -1,5 +1,5 @@
 .DEFAULT: all
-.PHONY: all clean debian-community debian-enterprise neo4j-desktop-mac neo4j-desktop-windows tarball-community tarball-community-extracted tarball-enterprise tarball-enterprise-extracted
+.PHONY: all clean debian-community debian-enterprise neo4j-desktop-mac neo4j-desktop-windows tarball-community tarball-community-extracted tarball-enterprise tarball-enterprise-extracted rpm-community rpm-enterprise
```
lassewesth: Alphabetical ordering!

lassewesth:
This is stunningly beautiful!

PR #8648, opened 2017-01-18
Check local lock client in SlaveLockClient instead of duplicating state

thobe:
By adding explicit methods for re-entry on the Locks.Client interface we can use the local lock client for checking if the lock is held locally, and achieve the same behaviour of locking on the master first, and not talking to the master on re-entry, while lowering the overhead of lock maintenance on the slave.

The lock maps in the SlaveLockClient completely disappear, and instead the state of the local lock client is used, since that already contains the same state.

Not only does this lower the memory footprint (for slaves that execute a write load), it also simplifies the code of the SlaveLockClient.


chrisvest (approved):
I think this looks good. 👍

chrisvest:
`TransactionConstraintsIT` was failing. I've seen other PRs fail on that, so I've restarted the build to see if it's flaky and decides to pass, or if it looks like a real problem.

chrisvest:
Failing test is flaky and caused by something else.

PR #8652, opened 2017-01-19
Allow all ASCII characters between '!' and '~' except ':' and ',' for usernames

eebus:


community/security/src/test/java/org/neo4j/server/security/auth/FileUserRepositoryTest.java
```diff
@@ -157,12 +157,14 @@ public void shouldNotAllowComplexNames() throws Exception
                 "The provided username is empty." );
         assertException( () -> users.assertValidUsername( "" ), InvalidArgumentsException.class,
                 "The provided username is empty." );
-        assertException( () -> users.assertValidUsername( ":" ), InvalidArgumentsException.class,
-                "Username ':' contains illegal characters. Use simple ascii characters and numbers." );
+        assertException( () -> users.assertValidUsername( "," ), InvalidArgumentsException.class,
+                "Username ',' contains illegal characters. Use simple ascii characters and numbers." );
```
fickludd: "... Use simple ascii characters and numbers." <- this part of the error message should maybe be changed?

community/security/src/test/java/org/neo4j/server/security/auth/FileUserRepositoryTest.java
```diff
@@ -157,12 +157,14 @@ public void shouldNotAllowComplexNames() throws Exception
...
         assertException( () -> users.assertValidUsername( "with space" ), InvalidArgumentsException.class,
                 "Username 'with space' contains illegal characters. Use simple ascii characters and numbers." );
         assertException( () -> users.assertValidUsername( "with:colon" ), InvalidArgumentsException.class,
                 "Username 'with:colon' contains illegal characters. Use simple ascii characters and numbers." );
+        assertException( () -> users.assertValidUsername( "withå" ), InvalidArgumentsException.class,
+                "Username 'withå' contains illegal characters. Use simple ascii characters and numbers." );
```
fickludd: Eh, why do we not allow 'å' btw... we could take UTF-8 alpha-numerics + a nice number of separators. Nice for internationalization. But maybe in the next iteration.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/AuthProceduresInteractionTestBase.java
```diff
@@ -227,10 +227,10 @@ public void shouldNotCreateUserIfInvalidUsername() throws Exception
...
         assertFail( adminSubject, "CALL dbms.security.createUser('', '1234', true)",
                 "The provided username is empty." );
-        assertFail( adminSubject, "CALL dbms.security.createUser('&%ss!', '1234', true)",
-                "Username '&%ss!' contains illegal characters." );
-        assertFail( adminSubject, "CALL dbms.security.createUser('&%ss!', '', true)",
-                "Username '&%ss!' contains illegal characters." );
+        assertFail( adminSubject, "CALL dbms.security.createUser(',ss!', '1234', true)",
+                "Username ',ss!' contains illegal characters." );
+        assertFail( adminSubject, "CALL dbms.security.createUser(':åss!', '', true)",
+                "Username ':åss!' contains illegal characters." );
```
fickludd: I think it slightly better if the usernames here and in the assert above are exactly the same, as the difference is supposed to be in valid vs. invalid password. We test the different illegal characters in the `FileUserRepositoryTest`.

community/security/src/test/java/org/neo4j/server/security/auth/FileUserRepositoryTest.java
fickludd: actually, let's leave that for now.

fickludd (changes requested):

eebus:
Opening new PR against 3.1

PR #8653, opened 2017-01-20
LabelScanStore compliance testing

tinwelint:
There were a couple of test classes testing LabelScanStore functionality as well as its integration with the rest of Neo4j. They have been refactored to be more abstract, only working with LabelScanStore instead of LuceneLabelScanStore and all those tests now also have a "Native" counterpart to them.

This PR also introduces a configuration option `unsupported.dbms.label_scan_store` (perhaps renamed before a GA release) to allow user to control which LabelScanStore implementation to use. LuceneLabelScanStore is still the default a little while longer.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/HighestPrioritizedLabelScanStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/HighestPrioritizedLabelScanStore.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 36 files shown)


tinwelint:
Failed on a seemingly unrelated ` ListQueriesProcedureTest.shouldListUsedIndexes`

community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/HighestPrioritizedLabelScanStore.java
```diff
@@ -0,0 +1,92 @@
...
+import org.neo4j.helpers.collection.Iterables;
+import org.neo4j.kernel.extension.KernelExtensions;
+import org.neo4j.kernel.impl.api.scan.LabelScanStoreProvider;
+
+/**
+ * Select the "correct" {@link LabelScanStoreProvider} among potentially several implementations
+ * loaded as {@link KernelExtensions}.
+ * <p>
+ * Two parameters affects which {@link LabelScanStoreProvider} is selected:
+ * <ol>
```
MishaDemianenko, line 33: java doc ordered list is a bit strange here, first item ends with `:` and then item number 2 goes

tinwelint: It was sort of intentional because point one refers to, or rather transitions into point two

MishaDemianenko: ok, fine :)

MishaDemianenko (approved):

PR #8655, opened 2017-01-20
Partial refactoring of internal index and constraint descriptors

fickludd:
The current way that indexes and constraints are modeled is quite rigid, and very hard to extend with additional types of constraints and indexes. This PR introduces a new more flexible model, where indexes and constraints are represented by `IndexDescriptor`s and `ConstraintDescriptor`s, which internally carry a `SchemaDescriptor` that specifies what part of the database the index or constraint applies to. 

To test the model, it has been applied to the schema store. As this means that only parts of the code base uses this new model, some temporary boundary classes were introduces to map between new and old types. 

I think further simplifications can be made to the `StoreReadLayer` interface concerning indexes and constraints, but I did not want to commit more time to that before the general model is approved. I also suspect that the SchemaCache and SchemaStorage could be improved. 

The usage of `SchemaRule.Kind` has been mostly restricted to code related to serialization of schemas, as the general program logic should not be tied to the disk representation. The role of `SchemaRule.Kind` in other part can now be filled by the new descriptors.


community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/index/NewIndexDescriptor.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return a user friendly description of what this index indexes.
+     */
+    String userDescription( TokenNameLookup tokenNameLookup );
+
+    /**
+     * Checks whether a constraint descriptor Supplier supplies this constraint descriptor.
+     * @param supplier supplier to get a constraint descriptor from
+     * @return true if the supplied constraint descriptor equals this constraint descriptor
+     */
+    default boolean isSame( Supplier supplier )
```
tinwelint: Not sure this is used, also the javadoc mentions constraints

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/index/NewIndexDescriptorImpl.java
```diff
@@ -0,0 +1,72 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api.schema_new.index;
+
+import org.neo4j.kernel.api.TokenNameLookup;
+import org.neo4j.kernel.api.schema_new.LabelSchemaDescriptor;
+
+import static java.lang.String.format;
+
+class NewIndexDescriptorImpl implements NewIndexDescriptor
```
tinwelint: Do we expected more implementations of `NewIndexDescriptor`? Could we just call this class `NewIndexDescriptor` directly and remove the interface? (Same with ConstraintDesciptor).

tinwelint (commented):
I'm having difficulties reviewing these changes as there are multiple aspects and they aren't confined to commits very well. Found some minor things, but likely missing others.

How long do we plan to have two frameworks for schema descriptors? I'd be happy if we completely replaced the old one.

tinwelint:
One more concern I had: some methods for accessing index/constraint rules or descriptors, are very frequently called, i.e. when changing properties or creating relationships or so. Can you just look that through to see if you're being extra liberal with garbage creation in terms of Optionals and lambdas and so on? Those most often amount to real objects and would just increase our garbage creation.

EDIT: scratch this, thought I saw something about this, but I don't think there's anything wrong there

fickludd:
I complete agree that the two index descriptor model should be removed ASAP. We plan on doing this in chunks over the coming weeks (as we touch other parts of the code base for composite index work). All use of the old model should be gone before 3.2 for sure.

Regarding avoiding use of Optionals and lambdas, I think we are actually using less of that following the clean-ups in `SchemaCache` and `SchemaStorage`.

PR #8657, opened 2017-01-20
Block Device support in Causal Clustering

chrisvest:
* For doing store copy, store files are now read and written through the page
  cache.
* File operations - copy, move, delete - now go through the page cache for
  store files.
* Tests have been changed to not rely on the file system to e.g. copy store
  directories around, and also to allow the tests to be configured with block
  device storage in the blockdevice repository.
* The test changes primarily involve CoreToCoreCopySnapshotIT and
  ReadReplicaReplicationIT.
* We now send an alignment requirement over with the `FileHeader` message.
  _This is a protocol change!_ This is only a preliminary approach, and an
  invitation to discuss the design and what the best way is to solve this
  problem.
* The checks that mutually excludes Causal Clustering and Custom IO (block
  device) have been removed.
* TemporaryStoreDirectory now cleans up after itself not only via the
  file system, but also via the page cache.
* The relationship between LocalDatabase and StoreFiles has been cut more clear.

These changes require #8654 to be merged before they work, and a rebase will have to remove the similar but conflicting changes in this PR.
This PR obsoletes #8617.

Please take a look, @digitalstain.


digitalstain:
Looks good.

PR #8660, opened 2017-01-20
Fix flaky ListQueriesProcedureTest

chrisvest:


chrisvest:
@thobe Does this one still make sense given #8672?

thobe:
@chrisvest you fix a different thing here. #8672 fixes what has been failing the build occasionally, but the change you proposed here still looks good to me.

thobe:
The test failure is unrelated. So I could merge this. But since it touches the same file, I would prefer to see #8672 merged first to minimise the risk from conflicts.

chrisvest:
@thobe Merged it.

PR #8661, opened 2017-01-20
Clarified and reworded magic safe-next method

fickludd:
Tried to explain the magic `safeHasNext` method, which was not very magic, put just a hack to robustly read schema rules while ignoring malformed entries. Since the `SchemaStorage` is about to be reworked, I opted for simply clarifying the method and usage. Improving better support in `SchemaStorage` (and making this hack un-needed) will be done soon.

### DIFF
--- a/community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/MandatoryProperties.java
+++ b/community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/MandatoryProperties.java
@@ -30,6 +30,7 @@ import org.neo4j.collection.primitive.PrimitiveIntSet;
 import org.neo4j.consistency.RecordType;
 import org.neo4j.consistency.report.ConsistencyReport;
 import org.neo4j.consistency.report.ConsistencyReporter;
+import org.neo4j.kernel.api.exceptions.schema.MalformedSchemaRuleException;
 import org.neo4j.kernel.impl.store.SchemaStorage;
 import org.neo4j.kernel.impl.store.StoreAccess;
 import org.neo4j.kernel.impl.store.record.NodePropertyExistenceConstraintRule;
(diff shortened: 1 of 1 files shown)


tinwelint (approved):
Looks good to me

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/MandatoryProperties.java
```diff
@@ -131,19 +121,43 @@ public MandatoryProperties( StoreAccess storeAccess )
...
+                {
+                    while ( true )
+                    {
+                        try
+                        {
+                            return ruleIterator.hasNext();
+                        }
+                        catch ( Exception e )
+                        {
+                            if ( !MalformedSchemaRuleException.class.isInstance( e.getCause() ) )
```
thobe, line 147: why not `!(e.getCause() instanceof MalformedSchemaRuleException)`?

thobe:
@fickludd general question around the rework of `SchemaStorage` that you are referring to. What will the new behaviour be in the event that the schema store is broken (which is the case this code guards for)?

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/MandatoryProperties.java
fickludd: No particular reason... I just took whatever came first. I had some a though telling me that getCause() gives `null` every now and then which would be a reason not to call method on it, but `instanceof` is null-safe so that's not really a valid argument. Do you see any reason to change it now apart from code consistency?

fickludd:
I think the behaviour is fine, but the I would rather implement it as a read option (`throwOnException`, default true), or at least get an iterator that can throw typed exceptions and preferably on `next()` instead of `hasNext()`. As we are removing some Rule classes the SchemaStore can become less about generics (fun code to write but hard to read) and more about explicit functionality.

PR #8662, opened 2017-01-21
Log message handling errors in Bolt

lutovich:
Malformed messages caused Bolt server to close the connection but the underlying error was never logged. This PR adds missing log.

Related to https://github.com/neo4j/neo4j-javascript-driver/issues/187.


zhenlineo:
LGFM

PR #8664, opened 2017-01-22
Improve makefile for debian and rpm

spacecowboy:
Now properly handles several tarballs present in the target directory of various versions

See https://github.com/neo4j/neo4j/pull/8688, https://github.com/neo4j/neo4j/pull/8689


spacecowboy:
Should split this up and backport the debian stuff to 3.0, while keeping the RPM stuff in 3.2

PR #8665, opened 2017-01-22
Generally faster HA tests

tinwelint:
A set of changes, _mostly_ in and around ClusterManager/ClusterRule which makes the whole HA component tests run in less than half the time (currently 66 min --> 28 min). High level change view:

- Lowered heartbeat interval/timeout in ClusterManager
- NetworkSender#stop() waits a more reasonable amount of time before closing when unsent messages
- Neo4jJobScheduler cancels uncancelled jobs on shutdown
- ClusterManager starts/stops databases and finds free ports in parallel

### DIFF
--- a/enterprise/cluster/src/test/java/org/neo4j/cluster/FreePorts.java
+++ b/enterprise/cluster/src/test/java/org/neo4j/cluster/FreePorts.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cluster;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Utility for discovering free ports within a given range, also for remembering already discovered ports
+ * so that next allocation will be faster. This utility is designed to be shared among many tests,
+ * preferably all tests in an entire component.
+ *
+ * One test should call {@link #newSession()} and allocate its ports there and {@link Session#close() close}
+ * when test is completed. Sessions share discovered ports for faster port allocation.
+ */
+public class FreePorts
+{
+    private static final int NOT_FOUND = -1;
+
+    private final Set<Integer> discoveredPorts = new HashSet<>();
+    private final Set<Integer> deniedPorts = new HashSet<>();
+    private final Set<Integer> currentlyOccupiedPorts = new HashSet<>();
+
+    public Session newSession()
+    {
+        return new Session();
+    }
+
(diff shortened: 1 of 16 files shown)


tinwelint:
Failed on `TestBranchedData.shouldCopyStoreFromMasterIfBranchedInLiveScenario` which seems to fail now and then with the same cause as this one, so unrelated. However I'll try to fix that test in this PR as well. Hang on.

enterprise/cluster/src/test/java/org/neo4j/cluster/FreePorts.java
```diff
@@ -0,0 +1,165 @@
...
+            {
+                // try binding it at wildcard
+                ServerSocket ss = new ServerSocket();
+                ss.setReuseAddress( false );
+                ss.bind( new InetSocketAddress( port ) );
+                ss.close();
+            }
+            catch ( IOException e )
+            {
+                deniedPorts.add( port );
```
MishaDemianenko, line 91: i see that it was copied and not really introduced here, but i think that all those code paths with lots of try/catches can be combined into one block with catch block that tracks denied ports; that will also remove most of the continue statements that will be redundant.

enterprise/ha/src/test/java/org/neo4j/kernel/impl/ha/ParallelLifecycle.java
```diff
@@ -0,0 +1,108 @@
...
+    public <T extends Lifecycle> T add( T lifecycle )
+    {
+        lifecycles.add( lifecycle );
+        return lifecycle;
+    }
+
+    @Override
+    public void init() throws Throwable
+    {
+        perform( lifecycle -> lifecycle.init() );
```
MishaDemianenko: all perform actions in this class can be method references

enterprise/ha/src/test/java/org/neo4j/kernel/impl/ha/ParallelLifecycle.java
```diff
@@ -0,0 +1,108 @@
...
+        {
+            service.submit( () ->
+            {
+                try
+                {
+                    action.act( lifecycle );
+                }
+                catch ( Throwable e )
+                {
+                    throw new RuntimeException( e );
```
MishaDemianenko: Is it correct that lifecycle failure will just throw exception in separate thread but will not cause test failure?

MishaDemianenko:
@tinwelint looks fine in general couple of small comments; since some of the modifications updating timeouts - do me have a TC windows run with integration tests included on top of this branch so far (windows build is the slowest that's why i'm curious in that )?

tinwelint:
Good point, I'll run a Windows build of these changes too and will let you know here.

tinwelint:
@MishaDemianenko ran a Windows build and it was successful. Time to run HA tests on Windows was already faster than on Linux for some reason, but went down from 45 --> 40 min.

enterprise/cluster/src/test/java/org/neo4j/cluster/FreePorts.java
tinwelint: I didn't want to complicate the PR by changing how free ports are discovered since new failures would be more difficult to pin point. Knowing that this is merely a test utility I'd opt for not changing unless we need to.

enterprise/ha/src/test/java/org/neo4j/kernel/impl/ha/ParallelLifecycle.java
tinwelint: Oh, you're right. I'll make it check the futures as well

tinwelint:
@MishaDemianenko fixed two comments, OK to skip the refactoring of free ports finding?

MishaDemianenko:
@tinwelint sure lets skip it, will merge on green

enterprise/ha/src/test/java/org/neo4j/kernel/impl/ha/ParallelLifecycle.java
```diff
@@ -80,7 +82,9 @@ public void shutdown() throws Throwable
         perform( Lifecycle::shutdown );
     }
 
-    private void perform( Action action ) throws InterruptedException
+    private static final AtomicBoolean FAILING = new AtomicBoolean();
```
MishaDemianenko: why do we need this field? and it looks it like to be in the center on attention and find a good place in the middle of the class file :)

tinwelint: a little test thing, it's already gone (pushed by mistake)

PR #8666, opened 2017-01-22
Test dbs gets null log and skips dump startup diagnostics

tinwelint:
this to improve test performance. The biggest benefit of having the null logger
is that startup diagnostics aren't dumped at all, due to most diagnostics being
dumped using Logger#bulk, which is now a no-op in null logger.

It turns out that quite a large amount of time starting a db is spent dumping
startup diagnostics.

Overall most tests should benefit from this speedup, e.g. kernel tests run
roughly 10% faster.


chrisvest (approved):
Looks good.

MishaDemianenko:
@tinwelint do you know if that is particular diagnostic provider that is slow or just general dumping of diagnostics consume that time?

chrisvest:
The build failure is not relevant. Merged.

tinwelint:
@MishaDemianenko no haven't looked into it, but at the very least it is (were) a lot of string building and printing those strings somewhere

PR #8675, opened 2017-01-23
NativeLabelScanStore as default

tinwelint:
Contains some changes which hindered a full green build with NativeLabelScanStore as the default label scan store.

The biggest change would be that import-tool no longer builds the label scan store as part of the import, which affects users by delaying the next startup of the db (where the label scan store automatically gets built). Read more about why in that specific commit.


tinwelint:
Closing since there's one more thing to do first

PR #8678, opened 2017-01-23
2.3 create unique direction

craigtaverner:
changelog: Fixed bug with relationship direction in create unique with path variable


eebus (approved):
Looks good, like the right/left name change

PR #8679, opened 2017-01-23
Splits meta pages out from main GBPTree paged file

tinwelint:
So that there will be a <file>.meta alongside <file> storing the meta/state pages.
There are two reasons for this, in order of importance:
- Ability to stream a GB+Tree with concurrent changes AND checkpoints. The problem
  is otherwise that tree state pages are copied first, since they are in the beginning
  of the file and if there's a checkpoint while stream the rest of the pages then the
  receiver of the copy will miss the stage change. After a store copy the changes that
  happened while copying are streamed in the form of transaction logs, but checkpoints
  aren't.
- The checkpoint algorithm (and to some extent performance) is clarified into flushing
  data file and meta/state file. Previously the one and the same paged file was flushed
  in both cases, but actually they were logically split into data and meta flushes.
  Now they are split physically as well.


tinwelint:
Not a viable solution, it turns out

PR #8693, opened 2017-01-24
Fix flaky BoltFailuresIT

lutovich:
Setup of a monitor that injects failures for `session#run()` was performed after the actual `session#run()` call. This was wrong because driver sends queries eagerly.


zhenlineo:
LG

PR #8694, opened 2017-01-24
Avoid closing bolt state machine multiple times

lutovich:
Each bolt worker owns a state machine. This state machine was closed when worker exited (both cleanly and by throwing an exception) and when worker was halted. This was problematic because workers can be halted from different threads that react to Netty callbacks (`channelInactive`, `handlerRemoved` and `exceptionCaught`). In such cases state machine could be closed multiple times.

This PR makes it safe to call `RunnableBoltWorker#halt()` multiple times. This method will only signal to the worker that it needs to stop and close it's state machine.


zhenlineo:
LG

PR #8695, opened 2017-01-24
move cluster state to data directory

martinfurmanski:
It was previously placed in the store directory which is not optimal
considering how the store directory usually is handled and because
the cluster state among other things contains unique IDs for the
member as well as for the bound cluster and should not be replicated
around haphazardly.

changelog: Move cluster state to data directory, was previously under graph.db.


mneedham:
Looks good to me. Seems to be some weirdness with the docs build but it's not related to this PR so dunno what's up with that!

PR #8696, opened 2017-01-24
Start and stop store watcher when replace database store

MishaDemianenko:
Store often can be replaced in clustering environment when store get copied from another machine
and then original database store get replaced.
If we will not gonna stop watcher we will get useless notifications that store files where deleted.
Configure to not notify about set of files that determine based on edition:
ignore branch data on clustered db, ignore temp copy etc.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
```diff
@@ -193,6 +200,15 @@ public void registerEditionSpecificProcedures( Procedures procedures ) throws Ke
         life.add( coreServerModule.membershipWaiterLifecycle );
     }
 
+    static Predicate<String> fileWatcherFileNameFilter()
+    {
+        return Predicates.all(
+                fileName -> !fileName.startsWith( PhysicalLogFile.DEFAULT_NAME ),
+                filename -> !filename.startsWith( StoreUtil.BRANCH_SUBDIRECTORY ),
```
martinfurmanski: Branched is not really necessary for CC, but harmless. You could perhaps also move this to EnterpriseEditionModule, although I don't quite get the role of that class.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
```diff
@@ -84,6 +87,7 @@ public synchronized void start() throws Throwable
         log.info( "Starting with storeId: " + storeId );
 
         dataSourceManager.start();
+        watcherService.start();
```
martinfurmanski, line 100: Could watcher-service be owned by the data source manager?

martinfurmanski (approved):

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
MishaDemianenko: @martinfurmanski  EnterpriseEditionModule can't have `branched` or `temp-copy` directories since it will not gonna copy it from anywhere, it's only pure single enterprise database and because of that is will inherit all the filters from community edition.

On the other part of comment: it's impossible for this edition to have branched directory right? (if so -  will remove that filter)

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
MishaDemianenko: so far DS manager only responsible for neo data source and this new service is not really related to that. Like database health, for example, it's more service on the side and do not have anything to do to neo data source itself, so i would say that so far it's more related to local db abstraction.
(But maybe in future if worth to introduce service that will wrap all the db related services together to restart them in cases like this)

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
martinfurmanski: Yes branched is impossible, but we still use temp-copy. See TemporaryStoreDirectory (which has it hardcoded but the same as the one in StoreUtil).

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
MishaDemianenko: updated :)

martinfurmanski (approved):

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
```diff
@@ -193,6 +200,14 @@ public void registerEditionSpecificProcedures( Procedures procedures ) throws Ke
         life.add( coreServerModule.membershipWaiterLifecycle );
     }
 
+    static Predicate<String> fileWatcherFileNameFilter()
+    {
+        return Predicates.all(
+                fileName -> !fileName.startsWith( PhysicalLogFile.DEFAULT_NAME ),
+                filename -> !filename.startsWith( StoreUtil.TEMP_COPY_DIRECTORY_NAME )
```
tinwelint: Could we invert this predicate to say what we watch instead? If so the perhaps community/enterprise could have same predicate.

tinwelint (changes requested):
It seems that still the file list is a bit wrong, I think it should be inclusive rather than exclusive. F.ex. the legacy index storage manages files like `index.db`, `index.db.old` and `index.db.tmp` and they're expected to be created and deleted from session to session.

MishaDemianenko:
@tinwelint updated, thx totally forgot about those!

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
```diff
@@ -201,6 +211,15 @@ public void registerEditionSpecificProcedures( Procedures procedures ) throws Ke
         life.add( coreServerModule.membershipWaiterLifecycle );
     }
 
+    static Predicate<String> fileWatcherFileNameFilter()
+    {
+        return Predicates.any(
+                fileName -> fileName.startsWith( PhysicalLogFile.DEFAULT_NAME ),
+                fileName -> fileName.startsWith( IndexConfigStore.INDEX_DB_FILE_NAME ),
+                filename -> filename.startsWith( StoreUtil.TEMP_COPY_DIRECTORY_NAME )
```
tinwelint, line 239: This doesn't look right, we're actively watching `temp-copy`? Can we just remove this specific filter? Isn't it enough with the single one watching neostore files and index.db?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/EnterpriseReadReplicaEditionModule.java
```diff
@@ -266,6 +270,23 @@ private OnlineBackupKernelExtension pickBackupExtension( NeoStoreDataSource data
         dependencies.satisfyDependency( createSessionTracker() );
     }
 
+    static Predicate<String> fileWatcherFileNameFilter()
+    {
+        return Predicates.any(
+                fileName -> fileName.startsWith( PhysicalLogFile.DEFAULT_NAME ),
+                fileName -> fileName.startsWith( IndexConfigStore.INDEX_DB_FILE_NAME ),
+                filename -> filename.startsWith( StoreUtil.BRANCH_SUBDIRECTORY ),
+                filename -> filename.startsWith( StoreUtil.TEMP_COPY_DIRECTORY_NAME )
```
tinwelint, line 301: Same as previous comment. I think this was just forgotten when the filters were inversed.

tinwelint (changes requested):

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
MishaDemianenko: we are watching for a directory content: our graph db directory and it's parent to see when graph db directory will be removed by itself. We can't just watch for specific files. That's why we need to supply list of exception filters that we will use to filter out file names that we should not notify about.
That's we we need to pass `temp-copy` to filter out all possible notification about it that can come alter.

tinwelint:
You could rebase this PR to hopefully get it green.

tinwelint:
Marking as not ready for merge because (to be clear from overview page):

- unaddressed comments
- needs rebase
- author away for a couple of weeks

MishaDemianenko:
@tinwelint should be ready now

tinwelint (approved):

PR #8697, opened 2017-01-25
Native user management procedures produce warnings

OliviaYtterbrink:
changelog: When using more than one auth provider, calling native user management procedures will produce a warning.

### DIFF
--- a/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
+++ b/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
@@ -416,7 +416,8 @@ public interface Status
         Forbidden( ClientError, "An attempt was made to perform an unauthorized action." ),
         AuthorizationExpired( ClientError, "The stored authorization info has expired. Please reconnect." ),
         AuthProviderTimeout( TransientError, "An auth provider request timed out." ),
-        AuthProviderFailed( TransientError, "An auth provider request failed." );
+        AuthProviderFailed( TransientError, "An auth provider request failed." ),
+        NativeProcedureWarning( ClientNotification, "Native user management procedures will not affect non-native users." );
 
         private final Code code;
 
--- a/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
+++ b/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
@@ -389,7 +389,8 @@ public interface Status
                 "A procedure is using or receiving a value of an invalid type." ),
         ProcedureTimedOut( ClientError,
                 "The procedure has not completed within the specified timeout. You may want to retry with a longer " +
-                "timeout." );
+                "timeout." ),
+        ProcedureWarning( ClientNotification, "The query used a procedure that generate a warning." );
 
         private final Code code;
 
@@ -416,8 +417,7 @@ public interface Status
         Forbidden( ClientError, "An attempt was made to perform an unauthorized action." ),
         AuthorizationExpired( ClientError, "The stored authorization info has expired. Please reconnect." ),
         AuthProviderTimeout( TransientError, "An auth provider request timed out." ),
-        AuthProviderFailed( TransientError, "An auth provider request failed." ),
-        NativeProcedureWarning( ClientNotification, "Native user management procedures will not affect non-native users." );
+        AuthProviderFailed( TransientError, "An auth provider request failed." );
 
         private final Code code;
 
--- a/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
+++ b/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
@@ -390,7 +390,7 @@ public interface Status
         ProcedureTimedOut( ClientError,
                 "The procedure has not completed within the specified timeout. You may want to retry with a longer " +
                 "timeout." ),
-        ProcedureWarning( ClientNotification, "The query used a procedure that generate a warning." );
+        ProcedureWarning( ClientNotification, "The query used a procedure that generated a warning." );
 
         private final Code code;
 
--- a/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/phases/DeprecationWarnings.scala
+++ b/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/phases/DeprecationWarnings.scala
@@ -75,8 +75,8 @@ object ProcedureWarnings extends VisitorPhase {
 
   private def findWarnings(statement: Statement): Set[InternalNotification] =
     statement.treeFold(Set.empty[InternalNotification]) {
-      case f@ResolvedCall(ProcedureSignature(name, _, _, _, _, _, true), _, _, _, _) =>
-        (seq) => (seq + ProcedureWarningNotification(f.position, name.toString), None)
+      case f@ResolvedCall(ProcedureSignature(name, _, _, _, _, _, Some(warning)), _, _, _, _) =>
+        (seq) => (seq + ProcedureWarningNotification(f.position, name.toString, warning), None)
       case _:UnresolvedCall =>
         throw new InternalException("Expected procedures to have been resolved already")
     }
--- a/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/phases/DeprecationWarnings.scala
+++ b/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/phases/DeprecationWarnings.scala
@@ -75,7 +75,7 @@ object ProcedureWarnings extends VisitorPhase {
 
   private def findWarnings(statement: Statement): Set[InternalNotification] =
     statement.treeFold(Set.empty[InternalNotification]) {
-      case f@ResolvedCall(ProcedureSignature(name, _, _, _, _, _, Some(warning)), _, _, _, _) =>
+      case f@ResolvedCall(ProcedureSignature(name, _, _, _, _, _, true), _, _, _, _) =>
         (seq) => (seq + ProcedureWarningNotification(f.position, name.toString), None)
       case _:UnresolvedCall =>
         throw new InternalException("Expected procedures to have been resolved already")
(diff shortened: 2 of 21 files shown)


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/spi/ProcedureSignature.scala
```diff
@@ -27,13 +27,14 @@ case class ProcedureSignature(name: QualifiedName,
...
-                              description: Option[String] = None) {
+                              description: Option[String] = None,
+                              warning: Boolean = false) {
 
   def outputFields = outputSignature.getOrElse(Seq.empty)
 
   def isVoid = outputSignature.isEmpty
 
-  override def toString = {
+  override def toString = {0
```
fickludd: 0?

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/Procedures.java
```diff
@@ -136,10 +136,22 @@ public void registerProcedure( Class<?> proc ) throws KernelException
...
     {
-        for ( CallableProcedure procedure : compiler.compileProcedure( proc ) )
+        registerProcedure( proc, overrideCurrentImplementation, false );
+    }
+
+    /**
+     * Register a new procedure defined with annotations on a java class.
+     * @param proc the procedure class
+     * @param overrideCurrentImplementation set to true if procedures within this class should override older procedures with the same name
+     * @param warn the warning the procedure should generate when called
```
fickludd: How can a boolean be "the warning the procedure should generate when called"? Maybe the javadocs was made and then the code changed? Really nice otherwise that you added a javadoc =).

community/kernel/src/main/java/org/neo4j/graphdb/impl/notification/NotificationCode.java
```diff
@@ -101,6 +101,11 @@
             Status.Statement.FeatureDeprecationWarning,
             "The query used a deprecated procedure."
     ),
+    PROCEDURE_WARNING(
```
fickludd, line 104: `AMBIGUOUS_NATIVE_USER_MANAGEMENT_WARNING`?

fickludd (changes requested):
I think this is behaving correclty, but have an issue with the implementation. I'm a bit weirded out by the fact that setting a boolean called `warn` on procedure compilation, will cause the very specific warning on ambiguous native user management to be generated. How about passing in a list of `Status` or something similar instead? Otherwise I think this is nice!

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ConfiguredAuthScenariosInteractionTestBase.java
```diff
@@ -80,4 +97,60 @@ public void shouldAllowRoleCallCreateNewTokensProceduresWhenConfigured() throws
...
+                localGraph.execute( transaction, "EXPLAIN CALL dbms.security.listUsers", Collections.emptyMap() );
+        assertThat(
+                containsNotification( result, "Native user management procedures will not affect non-native users." ),
+                equalTo( true ) );
+        transaction.success();
+        transaction.close();
+    }
+
+    @Test
+    public void shouldNotWarnWhenUsingNativeAndOtherProvider() throws Throwable
```
craigtaverner: Test name wrong?

craigtaverner: And what about a test for no warnings when we have no native configured (eg. LDAP together with plugins)

craigtaverner (changes requested):

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ConfiguredAuthScenariosInteractionTestBase.java
OliviaYtterbrink: Yes, wrong name on test.
There is no reason to test for when we don't have native configured, since the procedure is only available when one of the authproviders is the native.

community/kernel/src/main/java/org/neo4j/graphdb/impl/notification/NotificationCode.java
OliviaYtterbrink: Since it's more general now, it wouldn't make sense to be more specific.

community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
```diff
@@ -389,7 +389,8 @@ public Code code()
                 "A procedure is using or receiving a value of an invalid type." ),
         ProcedureTimedOut( ClientError,
                 "The procedure has not completed within the specified timeout. You may want to retry with a longer " +
-                "timeout." );
+                "timeout." ),
+        ProcedureWarning( ClientNotification, "The query used a procedure that generate a warning." );
```
craigtaverner: `generated`

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/phases/DeprecationWarnings.scala
```diff
@@ -75,8 +75,8 @@ object ProcedureWarnings extends VisitorPhase {
 
   private def findWarnings(statement: Statement): Set[InternalNotification] =
     statement.treeFold(Set.empty[InternalNotification]) {
-      case f@ResolvedCall(ProcedureSignature(name, _, _, _, _, _, true), _, _, _, _) =>
-        (seq) => (seq + ProcedureWarningNotification(f.position, name.toString), None)
+      case f@ResolvedCall(ProcedureSignature(name, _, _, _, _, _, Some(warning)), _, _, _, _) =>
```
craigtaverner: Might have been slightly nicer to generalize to a List of warnings. Not a requirement though.

community/kernel/src/main/java/org/neo4j/graphdb/impl/notification/NotificationCode.java
```diff
@@ -103,8 +103,8 @@
     ),
     PROCEDURE_WARNING(
             SeverityLevel.WARNING,
-            Status.Security.NativeProcedureWarning,
-            "Native user management procedures will not affect non-native users."
+            Status.Procedure.ProcedureWarning,
+            "The query used a procedure that generate a warning."
```
craigtaverner: `generated`

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
```diff
@@ -110,7 +110,7 @@ public void setup( Dependencies dependencies ) throws KernelException
                     ctx -> authManager.getUserManager( asEnterprise( ctx.get( SECURITY_CONTEXT ) ) ) );
             if ( config.get( SecuritySettings.auth_providers ).size() > 1 )
             {
-                procedures.registerProcedure( UserManagementProcedures.class, true, true );
+                procedures.registerProcedure( UserManagementProcedures.class, true, "%s only affect native users." );
```
craigtaverner: `affects`

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ConfiguredAuthScenariosInteractionTestBase.java
```diff
@@ -108,27 +109,27 @@ public void shouldWarnWhenUsingNativeAndOtherProvider() throws Throwable
                 .beginTransaction( KernelTransaction.Type.explicit, StandardEnterpriseSecurityContext.AUTH_DISABLED );
         Result result =
                 localGraph.execute( transaction, "EXPLAIN CALL dbms.security.listUsers", Collections.emptyMap() );
-        assertThat(
-                containsNotification( result, "Native user management procedures will not affect non-native users." ),
-                equalTo( true ) );
+        String description = String.format( "%s (%s)", Status.Procedure.ProcedureWarning.code().description(),
+                "dbms.security.listUsers only affect native users." );
```
craigtaverner: `affects`

craigtaverner: Actually, a better message would be `dbms.security.listUsers only applies to native users.`

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ConfiguredAuthScenariosInteractionTestBase.java
```diff
@@ -108,27 +109,27 @@ public void shouldWarnWhenUsingNativeAndOtherProvider() throws Throwable
...
         GraphDatabaseFacade localGraph = neo.getLocalGraph();
         InternalTransaction transaction = localGraph
                 .beginTransaction( KernelTransaction.Type.explicit, StandardEnterpriseSecurityContext.AUTH_DISABLED );
         Result result =
                 localGraph.execute( transaction, "EXPLAIN CALL dbms.security.listUsers", Collections.emptyMap() );
-        assertThat(
-                containsNotification( result, "Native user management procedures will not affect non-native users." ),
-                equalTo( false ) );
+        String description = String.format( "%s (%s)", Status.Procedure.ProcedureWarning.code().description(),
+                "dbms.security.listUsers only affect native users." );
```
craigtaverner: `applies to`

craigtaverner (changes requested):

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/phases/DeprecationWarnings.scala
OliviaYtterbrink: There's no benefit from using lists currently, if and when it's something we want/need it's easy to fix then.

fickludd (approved):

craigtaverner (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/Procedures.java
```diff
@@ -136,10 +136,22 @@ public void registerProcedure( Class<?> proc ) throws KernelException
     /**
      * Register a new procedure defined with annotations on a java class.
      * @param proc the procedure class
+     * @param overrideCurrentImplementation set to true if procedures within this class should override older procedures with the same name
      */
     public void registerProcedure( Class<?> proc, boolean overrideCurrentImplementation ) throws KernelException
     {
-        for ( CallableProcedure procedure : compiler.compileProcedure( proc ) )
+        registerProcedure( proc, overrideCurrentImplementation, "" );
```
craigtaverner: I'd prefer not to give empty string special meaning. Better to use optional everywhere.

PR #8698, opened 2017-01-25
Mutex between check-point and store-copy

tinwelint:
This to greatly simplify, even at all make possible to copy a GB+Tree with
concurrent changes.

Even currently, without the GB+Tree concurrent check-point during store-copy
may impose problems where a long store-copy might have concurrent check-points
that pruned the logs beyond where it started. The store-copy, after
taking all that time, would in the end fail due to server not being able to
serve the transactions which happened during the copy.

The trade-off to pause check-points while store-copy is running isn't all that big
because generally store-copy implies first making a checkpoint and check-pointing
can be made later instead w/o affecting behaviour of the system all that much.

Javadocs in StoreCopyCheckPointMutex further outlines how the mutex works and how
concurrent store-copy requests can run concurrently even though each store-copy request
performs a check-point first.

cl: [ha,causal clustering] Check-points can no longer run while store-copy is running, and vice versa. This to prevent a class of failures e.g. check-point pruning logs beyond starting point of store-copy and to better support native label indexing.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/StoreCopyCheckPointMutex.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/StoreCopyCheckPointMutex.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.transaction.log.checkpoint;
+
+import java.io.IOException;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.LockSupport;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import org.neo4j.function.ThrowingAction;
+import org.neo4j.graphdb.Resource;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+import static org.neo4j.helpers.Exceptions.launderedException;
+
+/**
+ * Mutex between {@link #storeCopy(ThrowingAction) store-copy} and {@link #checkPoint() check-point}.
+ * This to prevent those two running concurrently.
+ * <p>
+ * Normally a check-point implies first doing a check-point and so this relationships is somewhat intricate.
+ * In addition to having {@link #storeCopy(ThrowingAction)} as the "read lock" and {@link #checkPoint()} as the
+ * "write lock", {@link #storeCopy(ThrowingAction)} also accepts a code snippet to run before first concurrent
+ * store-copy grabs the lock, a snippet which can include a check-point.
+ *
+ * <pre>
+ *                                  <-WAIT--------------------|-CHECKPOINT--------->
+ *                                  |                                              |
+ * |----------|-----|---------------|---|-----------|---------|-------------|------|--------------------------|-> TIME
+ *            |     |                   |           |         |             |                                 |
+ *            |     |                   <-----STORE-|-COPY---->             <-WAIT-|-CHECKPOINT-|-STORE-COPY-->
+ *            |     |                               |
+ *            |     <-WAIT--|-STORE-COPY------------>
+ *            |                                   |
+ *            <-CHECKPOINT--|-STORE-COPY---------->
+ * </pre>
+ *
+ * In the image above there are three "events":
+ * <ol>
+ * <li>Store-copy 1, where there are three concurrent store-copies going on.
+ * Only the first one performs check-point</li>
+ * <li>External check-point, which waits for the ongoing store-copies to complete and then performs it</li>
+ * <li>Store-copy 2, which waits for the external check-point to complete and then starts its own
+ * check-point, which is part of the store-copy algorithm to then perform the store-copy.</li>
+ * </ol>
+ *
+ * Status changes are made in synchronized as opposed to atomic CAS operations, since this results
+ * in simpler code and since this mutex is normally called a couple of times per hour it's not an issue.
+ */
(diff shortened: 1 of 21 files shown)


tinwelint:
@digitalstain I recognized that failure handling wasn't properly taken care of, state-wise in the mutex so pushed another commit to handle that, please have a look at it as well.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/StoreCopyCheckPointMutex.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    public Resource storeCopy( ThrowingAction<IOException> beforeFirstConcurrentStoreCopy ) throws IOException
+    {
+        Lock readLock = lock.readLock();
+        boolean firstConcurrentRead = incrementCount() == 0;
+        boolean success = false;
+        try
+        {
+            if ( firstConcurrentRead )
```
MishaDemianenko, line 108: in both cases we use first action to try to perform checkpoint, maybe instead of do manual handling of that situation we can rely on our lock: if we can do writelock.tryLock() - then we do checkpoint before store copy otherwise skip it, as example?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/StoreCopyCheckPointMutex.java
```diff
@@ -0,0 +1,199 @@
...
+                // Wait for the "before" first store copy to complete
+                waitForFirstStoreCopyActionToComplete();
+            }
+            success = true;
+        }
+        finally
+        {
+            if ( success )
+            {
+                readLock.lock();
```
MishaDemianenko, line 132: with this handling model we also have temporary situation when storeCopyCount > 0, but somebody can grab a write lock

enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyServer.java
```diff
@@ -148,15 +152,20 @@ public RequestContext flushStoresAndStreamStoreFiles( String triggerName, StoreW
...
             // Copy the store files
-            monitor.startStreamingStoreFiles();
-            try ( ResourceIterator<StoreFileMetadata> files = dataSource.listStoreFiles( includeLogs ) )
+            long lastAppliedTransaction;
+            try ( Resource lock = mutex.storeCopy( checkPointAction );
+                    ResourceIterator<StoreFileMetadata> files = dataSource.listStoreFiles( includeLogs ) )
             {
+                lastAppliedTransaction = checkPointer.lastCheckPointedTransactionId();
+                monitor.startStreamingStoreFiles();
+                ByteBuffer temporaryBuffer = ByteBuffer.allocateDirect( (int) ByteUnit.mebiBytes( 1 ) );
```
MishaDemianenko, line 168: i think that would be super nice if we will release memory from this buffer as fast as we know that buffer can be clear

MishaDemianenko (changes requested):
Couple of minor comments, also as general comment i think that would be easier code wise and design wise to have a mutex that contain checkpointer and store copier inside and fully in change of invoking them - by doing that we will hide all the locking code inside that component and will avoid passing common lock across the components. But if others do not have any objections regarding that i will be fine with current design as well

enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyServer.java
tinwelint: Yes, although unrelated, right?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/StoreCopyCheckPointMutex.java
tinwelint: Yes, and I think that's fine because either the store-copy check-point will get first and the external check-point will have to wait for store-copy session to finish, or the external check-point will win and the store-copy check-point will come after it. In both cases the second check-point should be quite fast. I don't think functionality this is a problem.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/StoreCopyCheckPointMutex.java
tinwelint: Not sure what you mean. It's only the first one that performs the action, the others will wait for it to complete.

tinwelint:
About turning the thing inside-out... that's not straight forward since store-copy doesn't live in kernel. Also store-copy looks different in HA and CC

enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyServer.java
MishaDemianenko: correct, only saw that only because that line was moved a bit; completely unrelated

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/StoreCopyCheckPointMutex.java
MishaDemianenko: ok

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/StoreCopyCheckPointMutex.java
MishaDemianenko: both actions trying to do checkpoint before doing store copy, so maybe it's reasonable to move it inside and do something like: 
```
if (writeLock.tryLock()) {
  was able to do that so do checkpoint
} else {
  can't take write lock so it mean that we have copier doing copy so just skipping it.
}
```

MishaDemianenko:
that's true that will require common parent or interface to be introduced for store copier, and edition based construction of that thing

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/StoreCopyCheckPointMutex.java
tinwelint: Oh now I see your point. Right, that's a good idea, I'll try it out.

tinwelint: Although, it ties the before-action to doing a check-point, then we could might as well not have a before-action and just let the mutex have the CheckPointer directly.

MishaDemianenko: yes i think so

chrisvest:
Instead of sharing that mutex, the `CheckPointer` interface could expose `pause` and `unpause` methods.

tinwelint:
Nice idea. Store-copiers still do need to share the mutex since the mutex supports multiple concurrent store-copies to run concurrently, even though they all would like to perform check-point before starting the copy. This detail makes matters a bit more complicated than a simple pause/unpause _I think_.

PR #8700, opened 2017-01-26
SysVinit script for RPM

spacecowboy:
This adds RPM compatibility for the SysVinit script and includes it in the RPM package as well as the existing SystemD unit.

The main motivation is to support Amazon Linux (which does not have SystemD) but it will also be a way to allow us to consolidate your RPM tyrekicking builds (old 2.3 has it's own special stuff right now and now it won't have to)


new-packaging/bin/build-rpm-package
```diff
@@ -82,6 +84,7 @@ then
 fi
 
 # Copy manpages
+rm -rf ${sources_directory}/manpages
```
benbc, line 87: How does this change relate?

new-packaging/src/common/neo4j.init
```diff
@@ -34,29 +38,51 @@
...
   do_ulimit
   [ -d "${PIDDIR}" ] || mkdir -p "${PIDDIR}"
   chown "${NEO_USER}:" "${PIDDIR}"
 
-  start-stop-daemon --chuid ${NEO_USER} --start --quiet --oknodo --pidfile $PIDFILE --exec $DAEMON -- start
+  if [ -z ${LSB} ]; then
+    daemon --user="${NEO_USER}" --pidfile="${PIDFILE}" "${DAEMON} start > /dev/null 2>&1 &"
+  else
+    start-stop-daemon --chuid ${NEO_USER} --start --quiet --oknodo --pidfile ${PIDFILE} --exec ${DAEMON} -- start
+  fi
```
benbc, line 82: I really like that you're doing feature detection rather than platform detection. Would it be even cleaner to do that feature detection with `if which start-stop-daemon` and get rid of the variable?

new-packaging/src/common/neo4j.init
```diff
@@ -34,29 +38,51 @@
...
+  ${DAEMON} stop
+}
+
+do_status()
+{
+  if [ -z ${LSB} ]; then
+    status -p "${PIDFILE}" "${NAME}"
+  else
+    status_of_proc -p "${PIDFILE}" "${DAEMON}" "${NAME}"
+  fi
```
benbc, line 96: Ditto for feature detection.

new-packaging/src/common/neo4j.init
```diff
@@ -34,29 +38,51 @@
 
 PATH=/sbin:/usr/sbin:/bin:/usr/bin
 NAME=neo4j
-DAEMON=/usr/bin/$NAME
+DAEMON=/usr/bin/${NAME}
 PIDDIR=/var/run/${NAME}
 PIDFILE=${PIDDIR}/neo4j.pid
-SCRIPTNAME=/etc/init.d/$NAME-service
+SCRIPTNAME=/etc/init.d/${NAME}
```
benbc, line 44: This looks surprising. Shouldn't this change have been made ages ago when we switched the service name from `neo4j-service`?

new-packaging/src/common/neo4j.init
benbc: (I should say that I'm in two minds about this. Up to you.)

benbc (commented):
Looks good, thanks. A few comments.

new-packaging/bin/build-rpm-package
spacecowboy: Just a nice-to-have. If you already have built something then the files' presence means `gzip` will stop on an interactive prompt asking if you want to override. Couldn't see a force/yes argument so I added this to avoid the prompt.

new-packaging/src/common/neo4j.init
spacecowboy: correct. This should be backported naturally.

new-packaging/src/common/neo4j.init
spacecowboy: I prefer to not use `which` because that might not exist. surprising I know, but this is the case for some minimal docker images for example. I can use the directory check as I do before the sourcing above though.

spacecowboy:
@benbc updated after feedback

spacecowboy:
Backport of scriptname fix: https://github.com/neo4j/neo4j/pull/8703

benbc (approved):

PR #8704, opened 2017-01-26
Close statement in built in procedures

MishaDemianenko:


community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -64,83 +64,83 @@ public GraphResult buildSchemaGraph()
...
                     {
-                        startNodes.add( node );
+                        IndexDescriptor index = indexDescriptorIterator.next();
+                        String[] propertyNames = PropertyNameUtils.getPropertyKeys( statementTokenNameLookup, index.descriptor() );
+                        indexes.add( String.join( ",", propertyNames ) );
                     }
-                    if ( readOperations.countsForRelationship( ReadOperations.ANY_LABEL, relId, labelId ) > 0 )
+                    properties.put( "indexes", indexes );
+
+                    Iterator<NodePropertyConstraint> nodePropertyConstraintIterator = readOperations.constraintsGetForLabel( readOperations.labelGetForName( label.name() ) );
```
ragadeeshu: Some very long lines here

ragadeeshu:
@MishaDemianenko Some formatting, looks good otherwise!

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
MishaDemianenko: was trying to avoid additional formatting, but here we go should be fine now

MishaDemianenko:
@ragadeeshu updated

PR #8709, opened 2017-01-26
Use dependency resolver to resolve FileSystem from KernelContext.

MishaDemianenko:
Remove specific method to provide file system from kernel context.
All component dependencies should be provided by dependency resolver.


chrisvest:
Build failures are from unrelated known-flaky tests.

PR #8711, opened 2017-01-27
Native label scan store as default

tinwelint:
Contains some changes which hindered a full green build with NativeLabelScanStore as the default label scan store:

- One noticeable change is that import-tool no longer builds the label scan store as part of the import, which affects users by delaying the next startup of the db (where the label scan store automatically gets built), but also making import faster. Read more about why this was done in that specific commit.
- Another noticeable change is the setting for users to override which label scan store to use for the db, `dbms.label_index`. Setting is required and defaults to `NATIVE` Most users will never need to touch this setting though. 
- The presence of NativeLabelScanStore made the InMemoryLabelScanStore (used in testing) redundant and so it was also removed.

This PR depends on https://github.com/neo4j/neo4j/pull/8698 since that PR fixes an issue where a checkpoint in the middle of a copy of the native label scan store may corrupt it. It would be wise to merge that PR first.

cl: [kernel] Storage of labels are generally faster and uses less memory due to a new and native label index. This index is set as the default, but can be changed using `dbms.label_index` setting. No migration is required, but the new label index will be automatically built on the next start-up of a db.


community/neo4j/src/test/java/upgrade/StoreUpgraderTest.java
```diff
@@ -154,6 +153,13 @@ public StoreUpgraderTest( String version )
     }
 
     @Before
```
MishaDemianenko: now we have two `@Before` methods :)

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -480,13 +481,15 @@
...
             "10000" );
 
-    @Description( "Overrides otherwise automatically selected label scan store to use. " +
-            "If this setting is specified then this configured name will be matched with one of the loaded " +
-            "label scan stores, or fail if not found. If this setting isn't specified then the loaded " +
-            "label scan store with highest priority will be selected. Both name and priority of each " +
-            "label scan store is set by the kernel extension loading it." )
-    @Internal
-    public static final Setting<String> label_scan_store = setting( "unsupported.dbms.label_scan_store", STRING, NO_DEFAULT );
+    public static enum LabelIndex
```
MishaDemianenko: static modifier is redundant

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -480,13 +481,15 @@
...
+    {
+        NATIVE,
+        LUCENE;
+    }
+
+    @Description( "Backend to use for label --> nodes index" )
```
MishaDemianenko, line 490: do we really want this setting to be public? why users should have possibility to choose implementation? maybe we can make it internal/unsupported?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanIndexBuilder.java
```diff
@@ -71,4 +71,9 @@ public LabelScanIndex build()
         return isReadOnly() ? new ReadOnlyDatabaseLabelScanIndex( format, storageBuilder.build() )
                             : new WritableDatabaseLabelScanIndex( format, storageBuilder.build() );
```
MishaDemianenko: this line probably should invoke` buildWritable()` now instead

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
```diff
@@ -114,9 +116,22 @@ else if ( !luceneIndex.isValid() )
...
+            luceneIndex.create();
+            switchBackToReadOnly = true;
+            // We'll switch back in start() later
+        }
+    }
+
     @Override
     public void start() throws IOException
     {
+        System.out.println( "LUCENE!!" );
```
MishaDemianenko: debug output

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
```diff
@@ -36,16 +35,19 @@
 
 public class LuceneLabelScanStore implements LabelScanStore
 {
-    private final LabelScanIndex luceneIndex;
+    private final LuceneLabelScanIndexBuilder indexBuilder;
+    private volatile LabelScanIndex luceneIndex;
```
MishaDemianenko, line 39: maybe we should rename this field now :)

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
```diff
@@ -114,9 +116,22 @@ else if ( !luceneIndex.isValid() )
         }
     }
 
+    private void create() throws IOException
```
MishaDemianenko, line 119: The whole idea before was to have index that are fully responsible for creation, flushing, closing etc to avoid index volatile variable and necessity of multiple index builders and state handling. With new way it's gone and we need to handle possible multiple indexes and always use volatile variable to read something from index. Maybe we can avoid doing that and come back to old way of doing things?

MishaDemianenko (changes requested):
Couple of comments, otherwise looks great!

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
tinwelint: Typically users won't need to care since it will "just work", but at this point in time it's slightly optimistic. Perhaps we could still have it as `@Internal` (which makes it not show up publically, right?)

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
tinwelint: I'm not sure how to do that though

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
tinwelint: Uhm, sure... to what and why?

MishaDemianenko: it's not a lucene index anymore right? maybe just to `index` or `labeIndex`?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
MishaDemianenko: yes i think so, vote for `@Internal`

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
MishaDemianenko: We currently already have `ReadOnlyAbstractDatabaseIndex/WritableAbstractDatabaseIndex` but those expect AbstractLuceneIndex maybe those can be updated to expect something else - but those are really lucene oriented currently so that can be some additional work; 
or instead we can just delegate creation to index always - in that case index will replace it's internal structures, not sure how that will work with current new index

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
tinwelint: It's a field in the `LuceneLabelScanStore` so I hope it has to do with lucene :)

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
tinwelint: It'll just be a volatile reference somewhere else, won't it?

Reading a volatile which no one ever writes to, except ones at startup, aint so bad (knowing details about the java memory model), is it?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
MishaDemianenko: aaaaaaa, true :)

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/LuceneLabelScanStore.java
MishaDemianenko: yeah thats true, in previous indexes we have a partitions couple of levels below that where guarded anyway;
ok then it lets keep it for now like it is know; that's true - do not think we will have any problems with it

community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/NamedLabelScanStoreSelectionStrategy.java
```diff
@@ -31,20 +35,26 @@
  * <p>
  * Two parameters affects which {@link LabelScanStoreProvider} is selected:
  * <ol>
- * <li>Specifically configured label scan store name, see {@link GraphDatabaseSettings#label_scan_store},
+ * <li>Specifically configured label scan store name, see {@link GraphDatabaseSettings#label_index},
  * each {@link LabelScanStoreProvider} is instantiated with a name, which is matched with the configured name.
  * The setting can also be left unspecified, at which point the next parameter kicks in:</li>
  * <li>Highest prioritized instance, actually using {@link HighestSelectionStrategy}</li>
```
burqen: This javadoc does not look correct anymore. As I understand it, an exception will be thrown if the setting is left unspecified and priority will never be taken into account. Correct?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/scan/NativeLabelScanStoreExtension.java
```diff
@@ -55,7 +56,7 @@
 
     public NativeLabelScanStoreExtension()
     {
-        this( 0 /*disabled by default*/, LabelScanStore.Monitor.EMPTY );
+        this( 100, LabelScanStore.Monitor.EMPTY );
```
burqen: If I've understood it correctly priority is never taken into account when picking `LabelScanStore` and so priority could be removed from both Native and Lucene `...LabelScanStoreExtension`.

community/kernel/src/test/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStoreChaosIT.java
```diff
@@ -50,6 +50,6 @@ private static File storeFile( File directory )
     @Override
     protected void addSpecificConfig( GraphDatabaseBuilder builder )
     {
-        builder.setConfig( label_scan_store, NativeLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
+        builder.setConfig( label_index, NativeLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
```
burqen: Would be nice to use the `LabelIndex` enum from `GraphDatabaseSettings` explicitly.

community/kernel/src/test/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStoreStartupIT.java
```diff
@@ -28,14 +28,14 @@
...
-import static org.neo4j.graphdb.factory.GraphDatabaseSettings.label_scan_store;
+import static org.neo4j.graphdb.factory.GraphDatabaseSettings.label_index;
 
 public class NativeLabelScanStoreStartupIT extends LabelScanStoreStartupIT
 {
     @Override
     protected void addSpecificConfig( GraphDatabaseBuilder builder )
     {
-        builder.setConfig( label_scan_store, NativeLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
+        builder.setConfig( label_index, NativeLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
```
burqen: `LabelIndex` enum

community/kernel/src/test/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStoreUpdateIT.java
```diff
@@ -38,7 +38,7 @@
         @Override
         protected void configure( GraphDatabaseBuilder builder )
         {
-            builder.setConfig( label_scan_store, NativeLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
+            builder.setConfig( label_index, NativeLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
```
burqen: `LabelIndex` enum

community/lucene-index/src/test/java/org/neo4j/graphdb/LuceneLabelScanStoreChaosIT.java
```diff
@@ -82,6 +82,6 @@
     @Override
     protected void addSpecificConfig( GraphDatabaseBuilder builder )
     {
-        builder.setConfig( label_scan_store, LuceneLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
+        builder.setConfig( label_index, LuceneLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
```
burqen: `LabelIndex` enum

community/lucene-index/src/test/java/org/neo4j/graphdb/LuceneLabelScanStoreStartupIT.java
```diff
@@ -33,15 +33,15 @@
...
+import static org.neo4j.graphdb.factory.GraphDatabaseSettings.label_index;
 import static org.neo4j.io.fs.FileUtils.deleteRecursively;
 
 public class LuceneLabelScanStoreStartupIT extends LabelScanStoreStartupIT
 {
     @Override
     protected void addSpecificConfig( GraphDatabaseBuilder builder )
     {
-        builder.setConfig( label_scan_store, LuceneLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
+        builder.setConfig( label_index, LuceneLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
```
burqen: `LabelIndex` enum

community/lucene-index/src/test/java/org/neo4j/graphdb/LuceneLabelScanStoreUpdateIT.java
```diff
@@ -36,7 +36,7 @@
         @Override
         protected void configure( GraphDatabaseBuilder builder )
         {
-            builder.setConfig( label_scan_store, LuceneLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
+            builder.setConfig( label_index, LuceneLabelScanStoreExtension.LABEL_SCAN_STORE_NAME );
```
burqen: `LabelIndex` enum

community/lucene-index/src/test/java/org/neo4j/index/recovery/UniqueIndexRecoveryTests.java
```diff
@@ -92,7 +94,15 @@ public void before()
...
-        db = (GraphDatabaseAPI) factory.newEmbeddedDatabase( storeDir.absolutePath() );
+        db = newDb();
+    }
+
+    private GraphDatabaseAPI newDb()
+    {
+        return (GraphDatabaseAPI) factory
+                .newEmbeddedDatabaseBuilder( storeDir.absolutePath() )
+                .setConfig( GraphDatabaseSettings.label_index, LabelIndex.LUCENE.name() )
+                .newGraphDatabase();
```
burqen, line 105: Should this test also work with `Native`? It looks like it. Maybe make it abstract and have one implementation for `Lucene` and one for `Native`, like with the `LabelScanStore` specific tests.

burqen (changes requested):
Looks good in general, got some small comments.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/scan/NativeLabelScanStoreExtension.java
tinwelint: True, we can probably remove this altogether. I'll try that!

community/lucene-index/src/test/java/org/neo4j/index/recovery/UniqueIndexRecoveryTests.java
tinwelint: That's really a test of recovery of indexes and uniqueness constraints for the lucene schema index provider. Adding a native variant wouldn't really add anything. It's however unclear why the lucene label index is used specifically, but doesn't seem to matter.

burqen (approved):

burqen:
@tinwelint Approved! I'll wait for feedback from @MishaDemianenko and CI before merging.

PR #8717, opened 2017-01-27
Added a validator for dbms.security.auth_provider(s)

eebus:
changelog: Now dbms.security.auth_provider` and `dbms.security.auth_providers` can not be set to different things at the same time.


OliviaYtterbrink:
This won't work because there's no way to differentiate between `auth_provider` being set or is using the default value.

PR #8720, opened 2017-01-27
Refactored count-store to use indexId

craigtaverner:
The count store used to store the label+property key ids in the key fields for index statistics. However, this made it dependent on the definition of the index, and made it hard to support alternative definitions, including the concept of a composite index with multiple properties. This PR changes it to use only the indexId value and so it can support any index definition, including those we have not yet imagined. It also frees up a few bytes for future use in case we want them for things like histograms.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/counts/CountsTracker.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/counts/CountsTracker.java
@@ -77,7 +77,7 @@ public class CountsTracker extends AbstractKeyValueStore<CountsKey>
 {
     /** The format specifier for the current version of the store file format. */
     private static final byte[] FORMAT = {'N', 'e', 'o', 'C', 'o', 'u', 'n', 't',
-                                          'S', 't', 'o', 'r', 'e', /**/0, 1, 'V'};
+                                          'S', 't', 'o', 'r', 'e', /**/0, 2, 'V'};
     @SuppressWarnings("unchecked")
     private static final HeaderField<?>[] HEADER_FIELDS = new HeaderField[]{FileVersion.FILE_VERSION};
     public static final String LEFT = ".a", RIGHT = ".b";
@@ -184,16 +184,16 @@ public class CountsTracker extends AbstractKeyValueStore<CountsKey>
     }
 
     @Override
-    public Register.DoubleLongRegister indexUpdatesAndSize( IndexDescriptor descriptor,
+    public Register.DoubleLongRegister indexUpdatesAndSize( long indexId,
                                                             Register.DoubleLongRegister target )
     {
-        return get( indexStatisticsKey( descriptor ), target );
+        return get( indexStatisticsKey( indexId ), target );
     }
 
     @Override
-    public Register.DoubleLongRegister indexSample( IndexDescriptor descriptor, Register.DoubleLongRegister target )
+    public Register.DoubleLongRegister indexSample( long indexId, Register.DoubleLongRegister target )
     {
-        return get( indexSampleKey( descriptor ), target );
+        return get( indexSampleKey( indexId ), target );
     }
 
     public Optional<CountsAccessor.Updater> apply( long txId )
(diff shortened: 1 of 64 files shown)


community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobTest.java
```diff
@@ -700,9 +695,17 @@ private DoubleLongRegister indexUpdatesAndSize( Label label, String propertyKey
...
+            DoubleLongRegister result = newDoubleLongRegister(0, 0);
+            try
+            {
+                result = statement.readOperations()
+                        .indexUpdatesAndSize( IndexDescriptorFactory.of( labelId, propertyKeyId ),
+                                newDoubleLongRegister() );
+            }
+            catch (IndexNotFoundKernelException e)
+            {
+                // Tests ignore this
```
burqen: If we reach this catch, `result` will be null and we will fail later on. It may be better to just fail fast by not catching the exception?

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobTest.java
```diff
@@ -713,10 +716,18 @@ private DoubleLongRegister indexSample( Label label, String propertyKey ) throws
...
-            statement.readOperations().indexSample( IndexDescriptorFactory.of( labelId, propertyKeyId ), result );
+            DoubleLongRegister result = newDoubleLongRegister(0, 0);
+            try
+            {
+                result = statement.readOperations()
+                        .indexSample( IndexDescriptorFactory.of( labelId, propertyKeyId ), newDoubleLongRegister() );
+            }
+            catch (IndexNotFoundKernelException e)
+            {
+                // Tests ignore this
```
burqen: Same as above

community/kernel/src/main/java/org/neo4j/kernel/impl/store/counts/CountsTracker.java
```diff
@@ -184,16 +184,16 @@ public long minorVersion()
     }
 
     @Override
-    public Register.DoubleLongRegister indexUpdatesAndSize( IndexDescriptor descriptor,
+    public Register.DoubleLongRegister indexUpdatesAndSize( long indexId,
                                                             Register.DoubleLongRegister target )
```
burqen: One line can now fit all arguments.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexStatisticsIT.java
```diff
@@ -189,12 +192,24 @@ private IndexDefinition indexAliensBySpecimen()
         }
     }
 
-    private void resetIndexCounts( IndexDescriptor index )
+    private long indexId(IndexDescriptor index)
```
burqen: Code style

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexStatisticsTest.java
```diff
@@ -496,6 +499,24 @@ private IndexDescriptor createIndex( String labelName, String propertyKeyName )
         }
     }
 
+    private long indexId(IndexDescriptor index)
```
burqen: Code style

burqen (changes requested):
Looks good! Got some small comments.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobTest.java
burqen: Oops, that comment was apparently a bit to fast. The return value will be "zero" if no index exist. This hides the expected behaviour (introduced in this PR), namely seeing the `IndexNotFoundKernelException`. Let's adapt this test to this behaviour change.

burqen: Oops, that comment was apparently a bit to fast. The return value will be "zero" if no index exist. This hides the expected behaviour (introduced in this PR), namely seeing the `IndexNotFoundKernelException`. Let's adapt this test to this behaviour change.

burqen:
@craigtaverner Looks good! I triggered a new tools compile build, let's see what it says.

burqen:
@craigtaverner Looking at the Tools Compile failure. Seems like `SchemaRule` has changed and this branch needs a rebase.

burqen (approved):

PR #8722, opened 2017-01-30
Added a test which verifies neo4j-admin commands are present

spacecowboy:
See also https://github.com/neo4j/neo4j/pull/8723

### DIFF
--- a/community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
+++ b/community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
@@ -39,7 +39,7 @@ public class AdminTool
                 debug ).execute( homeDir, configDir, args );
     }
 
-    private final String scriptName = "neo4j-admin";
+    public static final String scriptName = "neo4j-admin";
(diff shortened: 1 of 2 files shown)


benbc:
@spacecowboy I don't like the fact that it matches the entire help text, so it will break if the text is tweaked. Could you just match the command names instead? (For example, asserting that each of them appears on its own on a line with just whitespace around it.)

I wonder also about getting it to fail if there are commands it doesn't recognize - as a mechanism to ensure that new commands are added to the test as they are created.

spacecowboy:
@benbc I think it's a good thing to have a failing test if someone tweaks the help text. It forces them to look at the complete output as it will look to a user.

It's also the easiest way to verify the list of commands. It becomes needlessly complicated if something like `verify` is used to make sure each command gets printed because that will by design not fail if you forget to add your command to the verifications. So you can add a command (with proper service loading) and the test won't fail, even though an extra command has appeared.

On the flipside, no solution exists which produces a failing test if someone adds a Command but forgets the service loading bit. It is probable that the dev would also forget to add the command to the list of commands in the test in your suggestion.

PR #8726, opened 2017-01-30
Partially written Raft log entries no longer cause recovery to fail

digitalstain:
In scenarios where a Raft log segment had a partial entry appended,
 recovery would complete normally, ignoring it as it should, but then
 subsequent entries would be appended after that, causing recoveries
 that happened before a rotation to fail. This is fixed by rotating
 at startup and after recovery but before any entries are appended,
 ensuring that no misalignment happens while avoiding introducing
 delete code for the Raft log segment files.

cl: [causal clustering] Fixes problem where running out of disk space when appending to the Raft log would cause recovery to fail.


martinfurmanski (approved):

martinfurmanski:
@digitalstain Looks good apart from the failing test

PR #8729, opened 2017-01-30
Single instance backup stress test update

MishaDemianenko:
Avoid log pollution for cases when backup is consistent and no error occur.
Update timeouts to use TimeUnits
Use standard exception stringification.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
@@ -22,6 +22,7 @@ package org.neo4j.kernel.impl.api;
 import java.time.Clock;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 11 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
```diff
@@ -143,20 +143,31 @@ public KernelTransaction newInstance( KernelTransaction.Type type, SecurityConte
...
+            {
+                newTransactionsLock.readLock().unlock();
+            }
         }
-        finally
+        catch ( InterruptedException ie )
         {
-            newTransactionsLock.readLock().unlock();
+            Thread.interrupted();
+            throw new RuntimeException( ie );
```
chrisvest: You could throw `org.neo4j.graphdb.TransactionFailureException` instead, since this `RuntimeException` will otherwise bubble straight out through things like `GraphDatabaseService.beginTx()` etc.

chrisvest (changes requested):
Looks good apart for that runtime exception, which I really don't like.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
MishaDemianenko: sure, updated

MishaDemianenko:
@chrisvest updated

chrisvest (approved):

PR #8731, opened 2017-01-31
Fixes id duplication detection issue in batch importer

tinwelint:
from a simple mistake incrementing a loop variable in the wrong place

cl: Fix an issue in batch importer where id duplication detection would miss some 1 in 10.000 ids

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
@@ -87,6 +87,8 @@ import static org.neo4j.unsafe.impl.batchimport.cache.idmapping.string.SourceInf
  */
 public class EncodingIdMapper implements IdMapper
 {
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
```diff
@@ -558,10 +560,10 @@ public long dataValue( long nodeId )
         SourceInformation source = new SourceInformation();
         SameInputIdDetector detector = new SameInputIdDetector();
         progress.started( "DEDUPLICATE" );
-        for ( int i = 0; i < numberOfCollisions; i++ )
+        for ( int i = 0; i < numberOfCollisions; )
```
tinwelint, line 563: This is _the_ fix

burqen (approved):
Looking good! Let's wait for the build to go green, I triggered another Tools Compile.
Edit: Oh right, it's docsbuild that is failing because it does not (and is not supposed to) build against 2.3. Merging.

PR #8733, opened 2017-01-31
Unflake ClusterPartitionIT

tinwelint:


burqen (changes requested):
Looking at the latest test failures in `ClusterPartitionIT` they seem to all have failed on some `RuntimeException`, e.g. `TransactionTerminatedException`.
So I don't think retry on `TransientFailureException` will help. Or am I missing something?

Otherwise I think this change looks good!

tinwelint:
@burqen `TransactionTerminatedException` is a `TransientFailureException` no?

burqen:
@tinwelint Not from what I can see.
Extends `TransactionFailureException` 
https://github.com/neo4j/neo4j/blob/3.1/community/graphdb-api/src/main/java/org/neo4j/graphdb/TransactionTerminatedException.java#L28
which extends `RuntimeException`
https://github.com/neo4j/neo4j/blob/3.1/community/graphdb-api/src/main/java/org/neo4j/graphdb/TransactionFailureException.java#L25

tinwelint:
You're probably misreading the code and my comment. `TransactionFailureException` != `TransientFailureException`

PR #8734, opened 2017-01-31
Add deprecation warning about connectors with arbitrary names

spacecowboy:


benbc:
@spacecowboy It seems odd to put this in the shell script rather than Java code. What about Windows?

spacecowboy:
@benbc that's a point, I just copied what we did for Java Heap warning - which can also be done in Java (at least now in 3.2).

You get the issue that people don't see warnings when they do `neo4j start` though, because it ends up in the log file if we do it in java. I'm fine with changing it if you consider that a reasonable trade-off.

benbc:
I think we have to believe that people are reading the logs for warnings. Otherwise all that earlier work you did on the config warnings is wasted. And of course it's immediately visible in console mode. I'd treat this similarly to the unrecognized setting stuff.

PR #8735, opened 2017-01-31
Ensured queries executed through `neo4j shell` are logged

thobe:
By ensuring that `shell` closes the query result, we get the query executed by shell in the log, and prevent resource leak.


spacecowboy (approved):
:+1:

spacecowboy:
Just add a note of this for the changelog

PR #8738, opened 2017-01-31
Warn and fallback to CYPHER 3.1 if planner=rule is selected

davidegrohmann:


henriknyman (approved):
With this we could also re-enable the `CYPHER planner=rule` combination in `NewPlannerTestSupport`.

PR #8740, opened 2017-02-01
Convert Cypher parameter types

### LINKED ISSUE #7115: [3.0][Procedure] Can't use Long safely as procedure argument type

henriknyman:
Fixes #7115 - Can't use Long safely as procedure argument type

- Converts Float to Double
- Converts all other Number types to Long
- Converts Character to String (this breaks backward compatibility tests)
- Converts values of Array, List and Map recursively
(including primitive arrays)
- Fix issue with materialization of Map results in compiled runtime

NOTE: This means that you can no longer use Cypher in embedded Neo4j to
store non-Cypher type property values.

### DIFF
--- a/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ParameterValuesAcceptanceTest.scala
+++ b/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ParameterValuesAcceptanceTest.scala
@@ -67,4 +67,113 @@ class ParameterValuesAcceptanceTest extends ExecutionEngineFunSuite with NewPlan
       r.hasProperty("name") should equal(false)
     }
   }
+
+  test("parameters should be transformed to cypher types") {
+    def echo(any: Any): AnyRef = {
+      executeScalarWithAllPlannersAndCompatibilityMode[AnyRef]("RETURN {p} AS p", "p" -> any)
+    }
+
+    echo(42.asInstanceOf[Byte]) shouldBe a[java.lang.Long]
+    echo(42.asInstanceOf[Short]) shouldBe a[java.lang.Long]
+    echo(42) shouldBe a[java.lang.Long]
+    echo(42L) shouldBe a[java.lang.Long]
+    echo(42.0f) shouldBe a[java.lang.Double]
+  }
+
+  test("array parameters should be transformed to cypher types") {
+    def echo(any: Any): Array[_] = {
+      executeWithAllPlannersAndRuntimesAndCompatibilityMode("RETURN {p} AS p", "p" -> any)
+        .columnAs[Array[_]]("p").next()
+    }
+
+    def echoNoCompatibility(any: Any): Array[_] = {
+      executeWithAllPlanners("RETURN {p} AS p", "p" -> any)
+        .columnAs[Array[_]]("p").next()
+    }
+
+    // Arrays of primitives
+    echo(Array[Byte](42.asInstanceOf[Byte])).foreach(_ shouldBe a[java.lang.Long])
+    echo(Array[Short](42.asInstanceOf[Short])).foreach(_ shouldBe a[java.lang.Long])
+    echo(Array[Int](42)).foreach(_ shouldBe a[java.lang.Long])
+    echo(Array[Long](42L)).foreach(_ shouldBe a[java.lang.Long])
+    echo(Array[Float](42.0F)).foreach(_ shouldBe a[java.lang.Double])
+    echo(Array[Double](42.0D)).foreach(_ shouldBe a[java.lang.Double])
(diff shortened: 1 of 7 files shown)


henriknyman:
It is still unclear to me wether we really want to do the full conversion without having a replacement mechanism in place (e.g. something like user defined value types).

community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ParameterValuesAcceptanceTest.scala
```diff
@@ -67,4 +67,113 @@ class ParameterValuesAcceptanceTest extends ExecutionEngineFunSuite with NewPlan
...
+  }
+
+  test("array parameters should be transformed to cypher types") {
+    def echo(any: Any): Array[_] = {
+      executeWithAllPlannersAndRuntimesAndCompatibilityMode("RETURN {p} AS p", "p" -> any)
+        .columnAs[Array[_]]("p").next()
+    }
+
+    def echoNoCompatibility(any: Any): Array[_] = {
+      executeWithAllPlanners("RETURN {p} AS p", "p" -> any)
```
Mats-SX, line 90: You can use the new parameter syntax for this one: `$p`

community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ParameterValuesAcceptanceTest.scala
```diff
@@ -67,4 +67,113 @@ class ParameterValuesAcceptanceTest extends ExecutionEngineFunSuite with NewPlan
...
+    echoNoCompatibility(Array[java.lang.Character]('a')).foreach(_ shouldBe a[java.lang.String])
+
+    // Nested array
+    echo(Array[Array[Byte]](Array(42.asInstanceOf[Byte]))).foreach { x =>
+      x shouldBe a[Array[AnyRef]]
+      x.asInstanceOf[Array[_]].foreach(_ shouldBe a[java.lang.Long])
+    }
+  }
+
+  test("list parameters should be transformed to cypher types") {
```
Mats-SX, line 120: List _elements_, right?

community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ParameterValuesAcceptanceTest.scala
```diff
@@ -67,4 +67,113 @@ class ParameterValuesAcceptanceTest extends ExecutionEngineFunSuite with NewPlan
...
+    def echo(any: Any): Iterable[_] = {
+      executeWithAllPlannersAndRuntimesAndCompatibilityMode("RETURN {p} AS p", "p" -> any)
+        .columnAs[Iterable[_]]("p").next()
+    }
+
+    def echoNoCompatibility(any: Any): Iterable[_] = {
+      executeWithAllPlanners("RETURN {p} AS p", "p" -> any)
```
Mats-SX, line 127: Also here

Mats-SX (approved):
Looks good I think

PR #8741, opened 2017-02-01
Remove ability to configure page size.

ragadeeshu:
Also adds a way to specify it via the PageCahceRule, so that it can be used for
testing.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/ConfiguringPageCacheFactory.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/ConfiguringPageCacheFactory.java
@@ -35,6 +35,7 @@ import static org.neo4j.graphdb.factory.GraphDatabaseSettings.mapped_memory_page
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.pagecache_memory;
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.pagecache_swapper;
 import static org.neo4j.kernel.configuration.Settings.BYTES;
+import static org.neo4j.unsafe.impl.internal.dragons.FeatureToggles.getInteger;
 
 public class ConfiguringPageCacheFactory
 {
@@ -43,6 +44,9 @@ public class ConfiguringPageCacheFactory
     private final PageCacheTracer tracer;
     private final Log log;
     private PageCache pageCache;
+    private int configuredPageSize;
+
+    private static final int pageSize = getInteger( ConfiguringPageCacheFactory.class, "pageSize", 8192 );
 
     public ConfiguringPageCacheFactory(
             FileSystemAbstraction fs, Config config, PageCacheTracer tracer, Log log )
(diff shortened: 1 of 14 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/ConfiguringPageCacheFactory.java
```diff
@@ -43,6 +44,9 @@
     private final PageCacheTracer tracer;
     private final Log log;
     private PageCache pageCache;
+    private int configuredPageSize;
+
+    private static final int pageSize = getInteger( ConfiguringPageCacheFactory.class, "pageSize", 8192 );
```
chrisvest: `static finals` should be in the top of the class.

community/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/ConfiguringPageCacheFactory.java
```diff
@@ -51,6 +55,16 @@ public ConfiguringPageCacheFactory(
...
+    }
+
+    /**
+     * Sets the page size, used by the {@link StandalonePageCacheFactory} for testing purposes.
+     * @param pageSize The new page size.
+     */
+    public void setPageSize( int pageSize )
+    {
+        configuredPageSize = pageSize;
     }
```
chrisvest: I'd do this juggling in a `TestConfiguringPageCacheFactory` that extends and overrides the original `ConfiguringPageCacheFactory`. The `PageCacheRule` can then use that when configuring its page cache.

chrisvest (changes requested):
I'd like greater separation of product and testing features.

ragadeeshu:
@chrisvest Like this?

chrisvest (approved):

chrisvest:
Yep.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -202,8 +201,7 @@ protected static Config getNeo4jConfig( Configuration config, Config dbConfig )
     {
         return new Config( stringMap( dbConfig.getParams(),
                 dense_node_threshold.name(), valueOf( config.denseNodeThreshold() ),
-                pagecache_memory.name(), valueOf( config.pageCacheMemory() ),
-                mapped_memory_page_size.name(), valueOf( config.pageSize() ) ),
+                pagecache_memory.name(), valueOf( config.pageCacheMemory() ) ),
```
chrisvest, line 204: Just noticed the `config` object still has its `pageSize()` method. Can that be removed?

ragadeeshu: Nice catch, will look into it.

ragadeeshu: @chrisvest removed

PR #8745, opened 2017-02-01
Update TCK to milestone 5

Mats-SX:
Also changes behaviour for comparability (operators `<`, `<=`, `>=`, and `>`) so that it never throws an exception (in the interpreted runtime). This is in line with [Cypher comparability](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc).

changelog: Cypher will no longer throw an exception when comparing values using the comparability operators `<`, `<=`, `>=`, and `>`


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/Comparer.scala
```diff
@@ -40,6 +40,18 @@ trait Comparer extends CypherSerializer {
         throw new IncomparableValuesException(serializeWithType(l), serializeWithType(r))
```
henriknyman: If `l` or `r` is `null` this will throw null pointer exception (from serializeWithType) instead of the intended exception.

henriknyman (commented):
Looks good. I just saw one issue that was pre-existing, so you can leave it for another PR if you like.

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/Comparer.scala
Mats-SX: Good point. I didn't want to touch orderability in this particular PR, because I know you guys are working on it, but the specified behaviour is to _never_ throw, so this code is most definitely wrong.

PR #8749, opened 2017-02-01
Provide more detailed error messages when ordering by mixed types

thobe:
Also for when aggregating with MIN or MAX on mixed types, or when ordering by list values.

### DIFF
--- a/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/AggregationAcceptanceTest.scala
+++ b/community/cypher/acceptance/src/test/scala/org/neo4j/internal/cypher/acceptance/AggregationAcceptanceTest.scala
@@ -19,7 +19,7 @@
  */
 package org.neo4j.internal.cypher.acceptance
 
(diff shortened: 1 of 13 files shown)


Mats-SX:
According to the [orderability CIP](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc#331-orderability), ordering between types should not be an error. Ordering should be defined across any set of values. I think that instead of curating these error messages, we should adapt behaviour so that no errors are thrown at all.

Related: #8745

thobe:
@Mats-SX while that is true for _new_ versions of Cypher, I don't think we should be changing the behaviour of old versions. What we _should_ do though, is make sure that the error messages our users are faced with are as helpful as they can be.

thobe:
Build failure is due to the docs build not covering `2.3`.

thobe:
@Mats-SX Your observation is very much relevant for the forward-merging process of this pull request though, so thank you for noting it.

PR #8750, opened 2017-02-02
Restrict access to unsecured components

OliviaYtterbrink:
Internal procedures should have access to all components while
external procedures should not be allowed to access components
that could let them bypass security.


OliviaYtterbrink:
This work is incorporated in #8753

PR #8751, opened 2017-02-02
Throw exception when calling shortestPath with same start and end node

craigtaverner:
The shortest path algorithm does not work when the start and end nodes are the same. This can happen if you perform a shortestPath search after a cartesian product that might have the same start and end nodes for some of the rows passed to shortestPath. If you would rather not experience this exception, and can accept the possibility of missing results for those rows, disable this in the Neo4j configuration. If you cannot accept missing results, and really want the shortestPath between two common nodes, then re-write the query using a normal cypher variable length pattern expression followed by sorting by path length and limiting to one result.

### DIFF
--- a/community/cypher/frontend-3.0/src/main/scala/org/neo4j/cypher/internal/frontend/v3_0/CypherException.scala
+++ b/community/cypher/frontend-3.0/src/main/scala/org/neo4j/cypher/internal/frontend/v3_0/CypherException.scala
@@ -155,10 +155,10 @@ class ShortestPathCommonEndNodesForbiddenException extends CypherExecutionExcept
 object ShortestPathCommonEndNodesForbiddenException {
   val ERROR_MSG: String =
     s"""The shortest path algorithm does not work when the start and end nodes are the same. This can happen if you
-        |perform a shortestPath search after a cartesian product that might have the same start and end nodes for some
-        |of the rows passed to shortestPath. If would rather not experience this exception, and can accept the
-        |possibility of missing results for those rows, disable this in the Neo4j configuration. If you cannot accept
-        |missing results, and really want the shortestPath between two common nodes, then re-write the query using
-        |a normal cypher variable length pattern expression followed by sorting by path length and limiting to one
-        |result.""".stripMargin
+       |perform a shortestPath search after a cartesian product that might have the same start and end nodes for some
+       |of the rows passed to shortestPath. If you would rather not experience this exception, and can accept the
+       |possibility of missing results for those rows, disable this in the Neo4j configuration by setting
+       |`cypher.forbid_shortestpath_common_node` to false. If you cannot accept missing results, and really want the
+       |shortestPath between two common nodes, then re-write the query using a standard Cypher variable length pattern
+       |expression followed by ordering by path length and limiting to one result.""".stripMargin
 }
--- a/community/cypher/frontend-3.0/src/main/scala/org/neo4j/cypher/internal/frontend/v3_0/CypherException.scala
+++ b/community/cypher/frontend-3.0/src/main/scala/org/neo4j/cypher/internal/frontend/v3_0/CypherException.scala
@@ -146,3 +146,19 @@ object ExhaustiveShortestPathForbiddenException {
        |start filtering.""".stripMargin
 }
 
+class ShortestPathCommonEndNodesForbiddenException extends CypherExecutionException(
+  ShortestPathCommonEndNodesForbiddenException.ERROR_MSG, null) {
+  override def mapToPublic[T <: Throwable](mapper: MapToPublicExceptions[T]): T =
+    mapper.shortestPathCommonEndNodesForbiddenException(message, this)
+}
+
+object ShortestPathCommonEndNodesForbiddenException {
+  val ERROR_MSG: String =
+    s"""The shortest path algorithm does not work when the start and end nodes are the same. This can happen if you
+        |perform a shortestPath search after a cartesian product that might have the same start and end nodes for some
+        |of the rows passed to shortestPath. If would rather not experience this exception, and can accept the
+        |possibility of missing results for those rows, disable this in the Neo4j configuration. If you cannot accept
+        |missing results, and really want the shortestPath between two common nodes, then re-write the query using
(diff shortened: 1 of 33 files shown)


community/cypher/frontend-3.0/src/main/scala/org/neo4j/cypher/internal/frontend/v3_0/CypherException.scala
```diff
@@ -146,3 +146,19 @@ object ExhaustiveShortestPathForbiddenException {
...
+  ShortestPathCommonEndNodesForbiddenException.ERROR_MSG, null) {
+  override def mapToPublic[T <: Throwable](mapper: MapToPublicExceptions[T]): T =
+    mapper.shortestPathCommonEndNodesForbiddenException(message, this)
+}
+
+object ShortestPathCommonEndNodesForbiddenException {
+  val ERROR_MSG: String =
+    s"""The shortest path algorithm does not work when the start and end nodes are the same. This can happen if you
+        |perform a shortestPath search after a cartesian product that might have the same start and end nodes for some
+        |of the rows passed to shortestPath. If would rather not experience this exception, and can accept the
```
Mats-SX: If `you` would rather ...

community/cypher/frontend-3.0/src/main/scala/org/neo4j/cypher/internal/frontend/v3_0/CypherException.scala
```diff
@@ -146,3 +146,19 @@ object ExhaustiveShortestPathForbiddenException {
...
+        |possibility of missing results for those rows, disable this in the Neo4j configuration. If you cannot accept
```
Mats-SX: Please mention the configuration key when referring here: `cypher.forbid_shortestpath_common_nodes`, and that setting it to `false` will disable the exception.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -145,6 +145,18 @@
     public static final Setting<Boolean> forbid_exhaustive_shortestpath = setting(
             "cypher.forbid_exhaustive_shortestpath", BOOLEAN, FALSE );
 
+    @Description( "This setting is associated with performance optimization. The shortest path algorithm does not " +
+                  "work when the start and end nodes are the same. With this setting set to `false` no path will " +
+                  "by returned when that happens. The default value of `true` will instead throw an exception. " +
```
Mats-SX: by -> be

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -145,6 +145,18 @@
...
+                  "This can happen if you perform a shortestPath search after a cartesian product that might have " +
+                  "the same start and end nodes for some of the rows passed to shortestPath. If is it preferable " +
```
Mats-SX: If is it -> If it is

community/cypher/frontend-3.0/src/main/scala/org/neo4j/cypher/internal/frontend/v3_0/CypherException.scala
```diff
@@ -146,3 +146,19 @@ object ExhaustiveShortestPathForbiddenException {
...
+        |missing results, and really want the shortestPath between two common nodes, then re-write the query using
+        |a normal cypher variable length pattern expression followed by sorting by path length and limiting to one
```
Mats-SX: cypher -> Cypher
normal -> standard (in my opinion)
sorting -> ordering

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -145,6 +145,18 @@
...
+                  "to not experience this exception, and acceptable for results to be missing for those rows, then " +
+                  "set this to `false`. If you cannot accept missing results, and really want the shortestPath " +
+                  "between two common nodes, then re-write the query using a normal cypher variable length pattern " +
```
Mats-SX: cypher -> Cypher
normal -> standard (in my opinion)

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -145,6 +145,18 @@
...
+                  "expression followed by sorting by path length and limiting to one result." )
```
Mats-SX: sorting -> ordering

Mats-SX (changes requested):
Some small spelling errors in the error message

Mats-SX (approved):

community/cypher/frontend-3.0/src/main/scala/org/neo4j/cypher/internal/frontend/v3_0/CypherException.scala
xx603: my neo4j is 3.5.4  and i don't have tthis line  in  my conf file

craigtaverner: It is in 3.5, see https://github.com/neo4j/neo4j/blob/3.5/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java#L227

The fact that it is not shown in the default neo4j.conf file is because only the most commonly edited settings are shown there. 

It is also in the Neo4j documentation at https://neo4j.com/docs/operations-manual/current/reference/configuration-settings/#config_cypher.forbid_shortestpath_common_nodes

xx603: I get it .And I solved this shortestpath problem by adding：`match paths = shortestpath((source)-[*0..2]-(target)) where source<>target` This can return right result.

PR #8753, opened 2017-02-02
Added configuration to grant full access to procedures

OliviaYtterbrink:
Procedures that need to access internal APIs such as `DependencyResolver` and `GraphDatabaseAPI` need to be added in the configuration, otherwise they will fail to compile.

changelog: Restrict access to unsecured components and add a configuration option to allow trusted procedures full access.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacadeFactory.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacadeFactory.java
@@ -25,7 +25,6 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
-import org.neo4j.configuration.Description;
 import org.neo4j.configuration.LoadableConfig;
 import org.neo4j.graphdb.config.Setting;
 import org.neo4j.graphdb.security.URLAccessRule;
@@ -100,10 +99,6 @@ public class GraphDatabaseFacadeFactory
         @Internal
         public static final Setting<String> editionName =
                 setting( "unsupported.dbms.edition", Settings.STRING, Edition.unknown.toString() );
-
-        @Description("A comma separated list of procedures that are allowed full access to the database, note that this" +
-                " will enable them to bypass security. Use with care.")
-        public static final Setting<String> procedure_full_access = setting( "dbms.security.procedures.full_access", Settings.STRING, "" );
     }
 
     protected final DatabaseInfo databaseInfo;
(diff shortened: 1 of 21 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacadeFactory.java
```diff
@@ -99,6 +100,10 @@
         @Internal
         public static final Setting<String> editionName =
                 setting( "unsupported.dbms.edition", Settings.STRING, Edition.unknown.toString() );
+
+        @Description("A comma separated list of procedures that are allowed full access to the database, note that this" +
+                " will enable them to bypass security. Use with care.")
+        public static final Setting<String> procedure_full_access = setting( "dbms.security.procedures.full_access", Settings.STRING, "" );
```
fickludd: I would suggest to put this in `GraphDatabaseSettings` instead, unless you have a particular reason to have it here?

Also, how about `dbms.security.procedures.unrestricted`?

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/ProcedureConfig.java
```diff
@@ -69,6 +73,17 @@ public ProcedureAllowedConfig( Config config )
...
+                config.getValue( Configuration.procedure_full_access.name() ).map( Object::toString ).orElse( "" );
+        if ( fullAccessProcedures.isEmpty() )
+        {
+            this.accessPatterns = Collections.emptyList();
+        }
+        else
+        {
+            this.accessPatterns = Stream.of( fullAccessProcedures.split( PROCEDURE_DELIMITER ) )
+                    .map( ProcedureConfig::compilePattern ).collect( Collectors.toList() );
+        }
```
fickludd: This new functionality and parsing is very similar to the `allowedRoles`. Maybe the parsing and compilation of procedure matching regex patterns can be unified?

community/kernel/src/test/java/org/neo4j/kernel/impl/proc/ProcedureJarLoaderTest.java
```diff
@@ -199,6 +204,40 @@ public void shouldGiveHelpfulErrorOnGenericStreamProcedure() throws Throwable
...
+    public void shouldGiveHelpfulErrorOnUnsafeRestrictedProcedure() throws Throwable
+    {
+        // Given
+        URL jar = createJarFor( ClassWithUnsafeComponent.class );
+
+        // Expect
+        exception.expect( ProcedureException.class );
+        exception.expectMessage( "Unable to set up injection for procedure `ClassWithUnsafeComponent`, the field" +
+                " `api` has type `class org.neo4j.kernel.impl.proc.ProcedureJarLoaderTest$UnsafeAPI` which is not a" +
+                " known injectable component.");
```
fickludd, line 217: Could there be different error messages for components that actually exists but are unsafe, and those that do not exists at all? That would be even more helpful.

fickludd (commented):
Strong PR, good test coverage. Some implementation details to address or rebut.

community/kernel/src/test/java/org/neo4j/kernel/impl/proc/ProcedureJarLoaderTest.java
OliviaYtterbrink: This will be fixed in PR #8772

fickludd (approved):

PR #8755, opened 2017-02-02
Kernel API changes for labels and relationship types

davidegrohmann:


tinwelint:
@davidegrohmann please sum up the driving reason for these changes in the PR comment, thanks. It's always helpful to understand that before reviewing the changes.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/Cursors.java
```diff
@@ -45,13 +43,31 @@ public boolean next()
         @Override
         public Object get()
         {
-            return null;
+            throw new IllegalStateException( "no elements" );
```
tinwelint, line 37: Is this in line with Cursor contract? If we change it we should definitely change (or in this case _add_) javadoc to Cursor mentioning this.

tinwelint (approved):
PRs with more deletions than additions is always 👍 
Nice simplifications and removal of code. The only "cost" is that "relationshipTypes" and hence "degrees" will create an intermediary/accidental PrimitiveIntSet, but it's not the most used call there is and I think the trade-offs are OK.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/Cursors.java
davidegrohmann: We claim so here: https://github.com/neo4j/neo4j/blob/3.2/community/primitive-collections/src/main/java/org/neo4j/cursor/Cursor.java#L29

davidegrohmann:
@tinwelint I can try to summarize the meaning of the changes, but this is a first step or a bigger effort in simplifying and cleaning up the cursor kernel api in order to make them suitable for Cypher.

tinwelint:
No harm in letting reviewers know the end goal, is there? "Cleanup" is very contextual.

PR #8756, opened 2017-02-02
Preserve parameter type information in compiled runtime

henriknyman:
This is needed for micro-benchmarking with lists of primitive types.

Also
- Fixes some more orderability cases
- Fixes unnecessary projection of variables

### DIFF
--- a/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
+++ b/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
@@ -364,4 +364,23 @@ public abstract class CompiledConversionUtils
         }
         throw new IllegalArgumentException( format( "Can not be converted to stream: %s", list.getClass().getName() ) );
     }
+
+    public static long unboxNodeOrNull( NodeIdWrapper value )
+    {
+        if ( value == null )
+        {
+            return -1L;
+        }
+        return value.id();
+    }
+
+    public static long unboxRelationshipOrNull( RelationshipIdWrapper value )
+    {
+        if ( value == null )
+        {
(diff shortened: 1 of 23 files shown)


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/codegen/ir/expressions/Parameter.scala
```diff
@@ -22,15 +22,15 @@ package org.neo4j.cypher.internal.compiler.v3_2.codegen.ir.expressions
...
 
   override def init[E](generator: MethodStructure[E])(implicit context: CodeGenContext) =
-    generator.expectParameter(key, variableName)
+    generator.expectParameter(key, variableName, cType)
 
   override def generateExpression[E](structure: MethodStructure[E])(implicit context: CodeGenContext): E =
     structure.loadVariable(variableName)
 
-  override def nullable(implicit context: CodeGenContext) = true
+  override def nullable(implicit context: CodeGenContext) = cType.canBeNullable
```
systay: Does this mean that params can now be non-nullable?

systay:
Looks good. Missing acceptance tests that make sure the parameter types are preserved

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/codegen/ir/expressions/Parameter.scala
alexaverbuch: AFAIK, yes, but only when assembling logical plans programmatically.

PR #8762, opened 2017-02-03
cleanup around GetServers V1 and make room for V2

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/GetServersProcedureV1.java
```diff
@@ -50,47 +50,56 @@
...
-S: SUCCESS {"fields": ["ttl", "servers"]}
-   RECORD [9223372036854775807, [
-{"role": "WRITE", "addresses": ["127.0.0.1:9001"]},
-{"role": "READ", "addresses": ["127.0.0.1:9002", "127.0.0.1:9003"]},
-{"role": "ROUTE", "addresses": ["127.0.0.1:9001", "127.0.0.1:9002", "127.0.0.1:9003"]}
-]]
+/**
+ * Returns endpoints and their capabilities.
+ *
+ * TODO: Detail signature (input and output).
```
mneedham, line 64: Won't the output format stay the same for V1 and any changes would be for V2? If so then we can just reuse the explanation in the comments from before.

mneedham (approved):
LGTM. Have added one comment on output comment explanation for V1

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/GetServersProcedureV1.java
martinfurmanski: I removed it on purpose because it was very messy and didn't seem to fit as a javadoc for the class. I will introduce a better description when I get around to it for the development of the V2. This does not imply a change of the results format itself.

PR #8763, opened 2017-02-03
Revert Convert Cypher parameter types

henriknyman:
Revert PR "Convert Cypher parameter types" #8740

#8740 was merged by mistake. We decided that we instead want to implement a less intrusive type conversion only on the specific arguments passed to procedure calls.

This should also fix the failures of the new tests introduced around the same time by https://github.com/neo4j/neo4j/pull/8611

This reverts commit 578592bb203fbd14320e7503eb63f4d23f0fc66c, reversing
changes made to 2384b0fa907f44b705f6d7371d639f0d71179d83.


Mats-SX:
The build error is unrelated. Merging.

PR #8764, opened 2017-02-03
Cleanup a little around cluster binding and identity

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/CoreState.java
```diff
@@ -100,8 +102,8 @@ public synchronized void handle( RaftMessages.ClusterIdAwareMessage clusterIdAwa
         }
         else
         {
-            log.info( "Discarding message[%s] owing to mismatched storeId. Expected: %s, Encountered: %s",
-                    clusterIdAwareMessage.message(), clusterId, clusterIdentity.clusterId() );
+            log.info( "Discarding message[%s] owing to mismatched clusterId. Expected: %s, Encountered: %s",
+                    clusterIdAwareMessage.message(), boundClusterId, msgClusterId );
```
mneedham, line 107: I think here should be `boundClusterId.get()` as well?

mneedham (approved):
One comment on what we're logging but otherwise LGTM.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/CoreState.java
martinfurmanski: Yea good spot. I fixed it by extracting it further up now.

PR #8768, opened 2017-02-06
Fix wrong return value from visit keys iterator

davidegrohmann:
We should return false to visit all labels in the set rather than
breaking after the first one.


chrisvest (approved):
👍

martinfurmanski (approved):

PR #8770, opened 2017-02-06
 Extend GraphDatabaseService to allow procedures to start transactions in new threads

OliviaYtterbrink:
This was previously not allowed since it meant that procedure authors could bypass security. Now all new transactions will be limited with the same access level as the procedure.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
@@ -1542,6 +1542,7 @@ public class OperationsFacade
             BasicContext ctx = new BasicContext();
             ctx.put( Context.KERNEL_TRANSACTION, tx );
             ctx.put( Context.THREAD, Thread.currentThread() );
+            ctx.put( Context.SECURITY_CONTEXT, procedureSecurityContext );
             procedureCall = procedures.callProcedure( ctx, name, input );
         }
         return new RawIterator<Object[],ProcedureException>()
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/DataSourceModule.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/DataSourceModule.java
@@ -371,8 +371,6 @@ public class DataSourceModule
         Guard guard = platform.dependencies.resolveDependency( Guard.class );
         procedures.registerComponent( TerminationGuard.class, new TerminationGuardProvider( guard ), true );
 
-        // Register injected private API components: useful to have available in procedures to access the kernel etc.
-
         // Below components are not public API, but are made available for internal
         // procedures to call, and to provide temporary workarounds for the following
         // patterns:
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/DataSourceModule.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/DataSourceModule.java
@@ -235,6 +235,10 @@ public class DataSourceModule
 
         this.storeId = neoStoreDataSource::getStoreId;
         this.kernelAPI = neoStoreDataSource::getKernel;
+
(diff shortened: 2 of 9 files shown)


fickludd (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/DataSourceModule.java
```diff
@@ -368,10 +372,6 @@ private Procedures setupProcedures( PlatformModule platform, EditionModule editi
         procedures.registerComponent( TerminationGuard.class, new TerminationGuardProvider( guard ), true );
 
         // Register injected private API components: useful to have available in procedures to access the kernel etc.
```
craigtaverner: This comment should move as well, I assume

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/ProcedureGDBFacadeSPI.java
```diff
@@ -205,6 +184,18 @@ public void shutdown()
     @Override
     public KernelTransaction beginTransaction( KernelTransaction.Type type, SecurityContext securityContext, long timeout )
```
craigtaverner: Should we really provide an API to create transactions with alternative security context?

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/ProcedureGDBFacadeSPI.java
```diff
@@ -205,6 +184,18 @@ public void shutdown()
...
     {
-        throw new UnsupportedOperationException();
+        try
+        {
+            availability.assertDatabaseAvailable();
+            KernelTransaction kernelTx = sourceModule.kernelAPI.get().newTransaction( type, this.securityContext, timeout );
```
craigtaverner, line 190: Perhaps this makes it safe? Using this.securityContext. What about also renaming the method parameter to 'ignoredSecurityContext' to make the code look a bit safer?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -1541,6 +1541,7 @@ public void unregisterExecutingQuery( ExecutingQuery executingQuery )
             BasicContext ctx = new BasicContext();
             ctx.put( Context.KERNEL_TRANSACTION, tx );
```
craigtaverner, line 1543: Since ProcedureGDBFacadeSPI has a currentTransaction method, do we need the kernel transaction in the context also?

integrationtests/src/test/java/org/neo4j/procedure/ProcedureIT.java
```diff
@@ -820,7 +820,7 @@ public void shouldBeAbleToWriteAfterCallingReadOnlyProcedure()
     private static List<Exception> exceptionsInProcedure = Collections.<Exception>synchronizedList( new ArrayList<>() );
 
     @Test
-    public void shouldGracefullyFailWhenSpawningThreadsCreatingTransactionInProcedures() throws Throwable
+    public void shouldBeAbleToSpawnThreadsCreatingTransactionInProcedures() throws Throwable
     {
         // given
         Runnable doIt = () -> {
```
craigtaverner, line 822: Perhaps rename the procedure, now that it is actually supported. Perhaps `createNodeInNewTransaction()`?

craigtaverner (changes requested):
Perhaps worth reviewing just how apoc creates threads and transactions and make a test that mimics that? Hopefully the existing tests are close enough, but if not, we could add more.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
OliviaYtterbrink: `currentTransaction` is only available inside GraphDatabaseFacade. Builtin procedures that need access to kernel transaction depend on it being in the context.

craigtaverner (approved):

PR #8771, opened 2017-02-06
PropertyRecord.getType() now returns PropertyType instead of int

digitalstain:
Reading the property type and converting to a PropertyType value
 requires bit shifting and masking that is not achieved by simply
 using the type int value as read from disk. This can lead to
 improper identification of the record type during recovery, among
 other scenarios.
This patch changes the PropertyRecord.getType() method to return
 PropertyType, properly parsed. It adds a getTypeAsInt() method
 to allow for access to the raw int value for serialization
 purposes.


tinwelint (approved):

tinwelint:
Replaced by #8780

PR #8772, opened 2017-02-06
Load fail procedure added

eebus:
When a procedure fail to load, a `LoadFailProcedure` will be loaded instead.

### DIFF
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/proc/ResourceInjectionTest.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/proc/ResourceInjectionTest.java
@@ -103,9 +103,8 @@ public class ResourceInjectionTest
         //When
         List<CallableProcedure> procList = compile( procedureWithUnsafeAPI.class, false );
         verify( log )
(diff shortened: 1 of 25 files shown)


community/kernel/src/test/java/org/neo4j/kernel/impl/proc/ResourceInjectionTest.java
```diff
@@ -71,11 +74,22 @@ public void shouldFailNicelyWhenUnknownAPI() throws Throwable
...
+        assertThat( procList.size(), equalTo( 1 ) );
+        try
+        {
+            procList.get( 0 ).apply( new BasicContext(), new Object[0] );
+            fail();
+        }
+        catch ( ProcedureException e )
+        {
+            assertThat( e.getMessage(), containsString(
+                    "org.neo4j.kernel.impl.proc.listCoolPeople is not available due to not having access to one or more components." ) );
```
fickludd: As mentioned in a comment on an earlier PR, I would like to get separate messages for the cases 1) where the component exists but the procedure lacks access, and 2) when the component simply does not exist.

fickludd (changes requested):
Look good, I really like this procedure mocking approach. Small request for error msg change.

fickludd (approved):

PR #8752, opened 2017-02-02
New schema rule serialization

fickludd:
Update to the Schema Rule serialization to support composite indexes and constraints.

PR #8775, opened 2017-02-07
Push index descriptor boundary

fickludd:
This PR pushes the boundary between the old and new Index Descriptors (so the the new ones are used more). The double models remain, but this is a step in the right direction.

This PR has no functional changes.

OBS: builds on #8752, review compared to that.


chrisvest:
@fickludd Needs a rebase.

chrisvest:
Replaced by #8812

PR #8776, opened 2017-02-07
Unflake TestBranchedData

tinwelint:
Problem being that after a role switch there were 3 big transactions w/ indexing and everything
taking place which, on some JVMs, caused full GC pauses on up to 2s. This was a problem because
the heartbeat timeout in ClusterManager is 2s and so the cluster would go into an unstable state
and try to switch role right when the transaction would commit, causing the test to fail.

This commit changes this test so that it no longer performs these big transactions.
There were no apparent reason for those transactions being so big, potentially it could be
(as the comment in the test suggests) that it wanted the lucene index files to change underneath.
Although there were no checking of those files, rather it would most likely use the suggested
file changes to verify that a branch + store-copy was actually performed.

Now instead the test installs a monitor to assert that a store-copy is performed on the
correct instances after the repair has been done.

### DIFF
--- a/enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/SwitchToSlave.java
+++ b/enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/SwitchToSlave.java
@@ -537,18 +537,58 @@ public abstract class SwitchToSlave
         }
     }
 
+    /**
+     * Monitors events in {@link SwitchToSlave}
+     */
     public interface Monitor
     {
-        void switchToSlaveStarted();
+        /**
+         * Called before any other slave-switching code is executed.
+         */
+        default void switchToSlaveStarted()
+        {   // no-op by default
+        }
 
-        void switchToSlaveCompleted( boolean wasSuccessful );
+        /**
+         * Called after all slave-switching code has been executed, regardless of whether it was successful or not.
+         *
+         * @param wasSuccessful whether or not the slave switch was successful. Depending on the type of failure
+         * other failure handling outside this class kicks in and there may be a switch retry later.
+         */
+        default void switchToSlaveCompleted( boolean wasSuccessful )
+        {   // no-op by default
+        }
 
-        void storeCopyStarted();
+        /**
+         * A full store-copy is required, either if this is the first time this db starts up or if this
+         * store has branched and needs to fetch a new copy from master.
+         */
+        default void storeCopyStarted()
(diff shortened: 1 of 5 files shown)


enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/SwitchToSlave.java
```diff
@@ -537,18 +537,58 @@ public void done()
         }
     }
 
+    /**
+     * Monitors events in {@link SwitchToSlave}
+     */
     public interface Monitor
```
burqen, line 543: Nice docs and defaults. Anonymous `SwitchToSlave.Monitor` in `HighAvailabilityMemberStateMachineTest`  can be updated to not override with no-op methods.

enterprise/ha/src/test/java/org/neo4j/ha/TestBranchedData.java
```diff
@@ -200,6 +206,13 @@ public void shouldCopyStoreFromMasterIfBranchedInLiveScenario() throws Throwable
...
     }
 
+    private BranchMonitor installBranchedDataMonitor( HighlyAvailableGraphDatabase odin )
+    {
+        BranchMonitor monitor = new BranchMonitor();
+        odin.getDependencyResolver().resolveDependency( Monitors.class ).addMonitorListener( monitor );
+        return monitor;
+    }
+
     private void retryOnTransactionFailure( ThrowingAction<Exception> transaction ) throws Exception
```
burqen: We could use the new `RetryHandler` here. Outside PR scope though, so feel free to ignore.

burqen (changes requested):
Changes look good. 1.5 comments

enterprise/ha/src/test/java/org/neo4j/ha/TestBranchedData.java
tinwelint: Ah yes we could, that thought may have flown through my head, but was ignored before it exited!

enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/SwitchToSlave.java
tinwelint: Good catch

tinwelint:
@burqen comments fixed

burqen (approved):

PR #8777, opened 2017-02-07
add more to the GetServersV2 framework

martinfurmanski:
Extract common functionality from the V1 code, split
procedure functionality from load balancing logic and
add an injectable strategy.

There is a concrete AllServersStrategy hooked in, but it
is merely a plot device for driving out the framework.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/procedure/GetServersProcedureV1.java
```diff
@@ -117,29 +111,28 @@ public ProcedureSignature signature()
         }
 
         return discoveryService.coreServers().find( leader )
-                .map( server -> server.getClientConnectorAddresses().getBoltAddress() );
+                .map( server -> server.connectors().boltAddress() );
```
lutovich: Maybe use `Util#extractBoltAddress()` here and in all other places in this class that do `connectors().boltAddress()`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/procedure/ResultFormatV1.java
```diff
@@ -0,0 +1,84 @@
...
+        Object[] readers = readEndpoints.stream().map( EndPoint::address ).toArray();
+        Object[] writers = writeEndpoints.stream().map( EndPoint::address ).toArray();
+
+        List<Map<String,Object>> servers = new ArrayList<>();
+
+        if ( writers.length > 0 )
+        {
+            Map<String,Object> map = new TreeMap<>();
+
+            map.put( "role", WRITE.name() );
```
lutovich: Please turn `role` and `addresses` into constants

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/procedure/ResultFormatV1.java
```diff
@@ -0,0 +1,84 @@
...
+import static org.neo4j.causalclustering.load_balancing.Role.READ;
+import static org.neo4j.causalclustering.load_balancing.Role.ROUTE;
+import static org.neo4j.causalclustering.load_balancing.Role.WRITE;
+
+/**
+ * The result format of GetServersV1 and GetServersV2 procedures.
+ */
+class ResultFormatV1
+{
+    static RawIterator<Object[],ProcedureException> build( List<EndPoint> routeEndpoints,
```
lutovich: Would be nice to have a test for this method.

lutovich (commented):
@martinfurmanski looks good, made couple comments

PR #8778, opened 2017-02-07
Fixes count issue in rel group defragmenter

tinwelint:
where the count was stored as unsigned short, but retrieved as
signed short in some places, which may have it seen as negative.


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/RelationshipGroupDefragmenter.java
```diff
@@ -57,44 +57,36 @@ public void run( long memoryWeCanHoldForCertain, BatchingNeoStores neoStore, lon
...
+
+                // Make adjustments for the next iteration
+                fromNodeId = toNodeId;
             }
+
+            // Now update nodes to point to the new groups
+            ByteArray groupCountCache = groupCache.getGroupCountCache();
+            groupCountCache.clear();
+            Configuration nodeConfig = withBatchSize( config, neoStore.getNodeStore().getRecordsPerPage() );
+            executeStage( new NodeFirstGroupStage( nodeConfig, toStore, neoStore.getNodeStore(), groupCountCache ) );
```
chrisvest, line 89: You just removed the `try-catch` here?

PR #8781, opened 2017-02-07
More consistent DynamicIndexStoreViewIT test time

tinwelint:
This test takes a couple of seconds to run normally, but now and then takes
up to minutes or hours!

Problem was that there was a race where five threads were created, threads which
busy-looped on making transactions revolving nodes and properties.
Such transactions acquire the schema read lock. The main test thread would,
after starting these other threads, go on to creating an index, a transaction
which needs to acquire the schema write lock. If those other threads had started
before the main thread would have came as far as acquiring the schema write lock
the test would get itself into a situation where the main thread could face
starvation and be waiting for an unknown amount of time since the other threads
would constantly acquire the schema read lock, overlapping with each other
and basically never let the main thread create the index.

This commit starts the other threads, but let them wait for a count down latch
and let them loose as soon as the index has been created, letting them race
with the population of that index. Number of concurrent updates fed into the
index populator has been verified to be the same as before this change.

### DIFF
--- a/community/neo4j/src/test/java/schema/DynamicIndexStoreViewIT.java
+++ b/community/neo4j/src/test/java/schema/DynamicIndexStoreViewIT.java
@@ -24,12 +24,13 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.neo4j.consistency.ConsistencyCheckService;
 import org.neo4j.graphdb.GraphDatabaseService;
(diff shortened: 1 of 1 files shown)


community/neo4j/src/test/java/schema/DynamicIndexStoreViewIT.java
```diff
@@ -76,9 +77,10 @@ public void populateDbWithConcurrentUpdates() throws Exception
             }
 
             List<Populator> populators = new ArrayList<>();
+            CountDownLatch startSignal = new CountDownLatch( 1 );
             for ( int i = 0; i < 5; i++ )
             {
-                Populator populator = new Populator( database, counter );
+                Populator populator = new Populator( database, counter, startSignal );
```
burqen: A `ThreadPool` to manage `populators` would be nice, even though slightly outside scope of this PR.

burqen (changes requested):
Looks good. One small comment.

community/neo4j/src/test/java/schema/DynamicIndexStoreViewIT.java
tinwelint: Sure, could do it while I'm here anyway

tinwelint:
@burqen fixed

burqen (approved):

PR #8785, opened 2017-02-07
Removed dependency on kernel internal schema types

fickludd:
Introduced own case classes in SchemaTypes, and map back and forth in
the PlanContext and QueryContext


eebus (approved):
Looks good

PR #8786, opened 2017-02-08
Removed cypher dependency on internal kernel schema types

fickludd:
Introduced own case classes in SchemaTypes, and map back and forth in
the PlanContext and QueryContext.


OliviaYtterbrink (approved):
Looking good

PR #8789, opened 2017-02-08
Makes CSV backslash-escaped quotes parsing optional

### LINKED ISSUE #8472: neo4j-import does not correctly recognize Excel-type CSV wrt. double quotes and backslash
neo4j-import does not correctly recognize Excel-type CSV wrt. double quotes and backslash

- Neo4j version: 3.1
- Operating system: Arch Linux
- API/Driver: None
- **Steps to reproduce**

The following CSV parsed correctly under MongoDB and MySQL, but it is failing in Neo4j:

http://neuromancer.inf.um.es:8080/es.stackoverflow/Posts.csv.gz

- Expected behavior

CSV read correctly with default parameters, as neo4-import is supposed to convert "" to " within a field.

- Actual behavior

neo4j-import fails reporing early termination of a field. The example error is the following:

For an imput line with the following characters:

abcdef\\""ghi

The parser interprets the first \" as *one* double quotes, and then the next double quote is used to end the field early. I think quoting of backslash should be avoided if backslask is not specified explicitly as the quote character. The above text should produce the following complete field (using the conventions of Excel-CSV):

abcdef\\"ghi

(that is, one backslash and one quote).

tinwelint:
This because it prevents normally expected parsing behaviour of some CSV data.
This commit doesn't introduce any surface configuration though and the default
keeps the behaviour unchanged, to be changed in next major version perhaps).

Fixes #8472

### DIFF
--- a/community/csv/src/main/java/org/neo4j/csv/reader/Configuration.java
+++ b/community/csv/src/main/java/org/neo4j/csv/reader/Configuration.java
@@ -25,6 +25,13 @@ package org.neo4j.csv.reader;
 public interface Configuration
 {
     /**
+     * TODO: Default value is false here, but it is our intention to flip this to true at some point
+     * because of how it better complies with common expectancy of behavior. It may be least disruptive
+     * to do this when changing major version of the product.
+     */
+    boolean DEFAULT_LEGACY_STYLE_QUOTING = true;
+
+    /**
      * Character to regard as quotes. Quoted values can contain newline characters and even delimiters.
      */
     char quotationCharacter();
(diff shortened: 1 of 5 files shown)


RXminuS:
Great work @tinwelint !, should we add something to the docs as well about how LOAD CSV can be configured?

RXminuS:
Dumb question...how/where would you change the configuration setting?

tinwelint:
@RXminuS good question, this PR doesn't expose this setting in LOAD CSV, but there's an internal discussion on how to do that and for which version a.s.o. Will try to keep the GH issue updated on that as well!

community/csv/src/main/java/org/neo4j/csv/reader/Configuration.java
```diff
@@ -25,6 +25,13 @@
 public interface Configuration
 {
     /**
+     * TODO: Default value is false here, but it is our intention to flip this to true at some point
+     * because of how it better complies with common expectancy of behavior. It may be least disruptive
+     * to do this when changing major version of the product.
+     */
+    boolean DEFAULT_LEGACY_STYLE_QUOTING = true;
```
craigtaverner: The comment says this is currently `false`, but this line claims otherwise. Which version conforms the the RFC, the legacy style or the not legacy style?

craigtaverner: Having read the tests, it looks like legacy-style is the existing neo4j behaviour, and not-legacy is the RFC behaviour. So then I guess only the comment above should say the default is true, but we plan to flip to false.

tinwelint: Oh that's right, the comment is wrong, fixing

tinwelint:
@craigtaverner fixed

spacecowboy (approved):
Looks very nice. This just leaves the question of how a user would do it.

From `neo4j-admin import` it could naturally be a command-line option in `3.2`. Not sure how to do it for LOAD CSV though..

burqen:
@spacecowboy @tinwelint Do you think the question about how a used would do it need an answer before merging? Otherwise I will go ahead and click the button.

tinwelint:
I think that can be a separate PR @burqen

PR #8792, opened 2017-02-08
Move degree() and relationships() to entity read operations

davidegrohmann:


davidegrohmann:
@tinwelint any progress here?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -366,14 +365,15 @@ public Object nodeGetProperty( long nodeId, int propertyKeyId ) throws EntityNot
...
             throws EntityNotFoundException
     {
         statement.assertOpen();
         try ( Cursor<NodeItem> node = dataRead().nodeCursorById( statement, nodeId ) )
         {
-            return new CursorRelationshipIterator(
-                    node.get().relationships( direction( direction ), deduplicate( relTypes ) ) );
+            return new CursorRelationshipIterator( dataRead()
+                    .nodeGetRelationships( statement, node.get(), direction( direction ),
+                            asSet( relTypes, t -> t >= 0 ) ) );
```
tinwelint, line 376: So it'll silently filter negative ids, why do we want to hide such illegal arguments?

Have you measured this change going from int[] to PrimitiveIntSet? Quick comparison locally shows at least 10x slower, on a very low level. Would be great to measure this first.

Also later on we'd like to pass in sorted type ids to be able to take advantage of the store format better and this would go against that, I feel.

This comment applies to the execution stack beyond this point of course as well.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
```diff
@@ -1615,6 +1657,67 @@ else if ( txState.nodeIsAddedInThisTx( id ) )
         return storeLayer.nodeExists( id );
     }
 
+    @Override
+    public PrimitiveIntSet relationshipTypes( KernelStatement statement, NodeItem node )
```
tinwelint, line 1661: I think having PrimitiveIntSet for relationship types here is OK though, because it's generally not a nearly as frequently called operation as getting relationships.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
```diff
@@ -338,7 +338,7 @@ private void importRelationships( NodeRelationshipCache nodeRelationshipCache,
 
         String topic = " Sparse";
         nodeRelationshipCache.setForwardScan( true );
-        // Stage 4b -- set node nextRel fields for sparse nodes
+        // Stage 4b -- set node nextRe fields for sparse nodes
```
tinwelint, line 341: Bad keypress?

tinwelint (changes requested):
If we can just show that ths int[] -> PrimitiveIntSet change doesn't affect performance negatively I'm happy with this

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
davidegrohmann: oh yeah, sorry about that

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
davidegrohmann: We had the same before see for example: https://github.com/neo4j/neo4j/pull/8792/files#diff-da5810feb6cba4b1db0a00086846f4c8L170 The return type didn't really changed.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
davidegrohmann: We were silently filtering negative ids before (I am not sure if on purpose or not): https://github.com/neo4j/neo4j/pull/8792/files#diff-35e966ebb9c89256d6078481e6170e09L865 (without the filtering there are tests failures). I am happy to reconsider that, but I think this PR is not the right place for that.

I'll have a look at the performance, but mind that it might be slower to build the set, but the actual check if a type is in the set or not, could be faster than scanning the array. So we might need to look at the bigger picture.

It didn't seem to me that we were taking any advantage of the fact that the array was sorted. E.g., https://github.com/neo4j/neo4j/pull/8792/files#diff-962abdeb972f25d4dc93b35378804fdeL232 Am I missing something?

tinwelint: Oh we did? Alright, then keeping it is alright, agreed.

Keep in mind that normally the number of types is very few and so scanning the array will most likely be faster than asking the set too.

No we didn't/don't, I meant that we'd quite like to do that in the near future.

PR #8804, opened 2017-02-13
Fixes issue with lonely check point in log file during recovery

tinwelint:
LatestCheckPointFinderTest didn't test what it claimed to test and so this issue
slipped through the cracks. This whole test class have been simplified and its
use of mocking heavily reduced. Issue was this:

Given an older log file with some transaction in and a newer log file with only
a check point in, targeting an entry in the older log file ->
LatestCheckPointFinder would report that there were no transactions after
the last check point, while in fact there were. This would result in all those
transactions to not be recovered on that startup.

This commit also adds more detailed logging on finding last checkpoint at startup,
this to better be able to reason about recovery issues by reading debug.log.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
@@ -141,7 +141,7 @@ public class LatestCheckPointFinder
             {
                 // This check point entry targets a previous log file.
                 // Go there and see if there's a transaction. Reader is capped to that log version.
-                startEntryAfterCheckPoint = extractFirstTxIdAfterPosition( target, target.getLogVersion() ) !=
+                startEntryAfterCheckPoint = extractFirstTxIdAfterPosition( target, version ) !=
                         LatestCheckPoint.NO_TRANSACTION_ID;
             }
         }
--- a/community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
@@ -58,6 +58,7 @@ public class LatestCheckPointFinder
         long version = fromVersionBackwards;
         long versionToSearchForCommits = fromVersionBackwards;
         LogEntryStart latestStartEntry = null;
+        LogEntryStart oldestStartEntry = null;
         long oldestVersionFound = -1;
         while ( version >= INITIAL_LOG_VERSION )
         {
@@ -73,6 +74,7 @@ public class LatestCheckPointFinder
             CheckPoint latestCheckPoint = null;
             ReadableLogChannel recoveredDataChannel =
                     new ReadAheadLogChannel( channel, NO_MORE_CHANNELS );
+            boolean firstStartEntry = true;
 
             try ( LogEntryCursor cursor = new LogEntryCursor( logEntryReader, recoveredDataChannel ) )
             {
@@ -84,37 +86,29 @@ public class LatestCheckPointFinder
                     {
                         latestCheckPoint = entry.as();
                     }
-                    if ( entry instanceof LogEntryStart && ( version == versionToSearchForCommits ) )
+                    if ( entry instanceof LogEntryStart )
                     {
-                        latestStartEntry = entry.as();
+                        LogEntryStart startEntry = entry.as();
+                        if ( version == versionToSearchForCommits )
+                        {
+                            latestStartEntry = startEntry;
+                        }
+
+                        // The scan goes backwards by log version, although forward per log version
+                        // Oldest start entry will be the first in the last log version scanned.
+                        if ( firstStartEntry )
+                        {
+                            oldestStartEntry = startEntry;
+                            firstStartEntry = false;
+                        }
                     }
                 }
             }
 
             if ( latestCheckPoint != null )
             {
-                // Is the latest start entry in this log file version later than what the latest check point targets?
-                LogPosition target = latestCheckPoint.getLogPosition();
-                boolean startEntryAfterCheckPoint = latestStartEntry != null &&
-                        latestStartEntry.getStartPosition().compareTo( target ) >= 0;
-                if ( !startEntryAfterCheckPoint )
-                {
-                    if ( target.getLogVersion() < version )
-                    {
-                        // This check point entry targets a previous log file.
-                        // Go there and see if there's a transaction. Reader is capped to that log version.
-                        startEntryAfterCheckPoint = extractFirstTxIdAfterPosition( target, target.getLogVersion() ) !=
-                                LatestCheckPoint.NO_TRANSACTION_ID;
-                    }
-                }
-
-                // Extract first transaction id after check point target position.
-                // Reader may continue into log files after the initial version.
-                long firstTxIdAfterCheckPoint = startEntryAfterCheckPoint
-                        ? extractFirstTxIdAfterPosition( target, fromVersionBackwards )
-                        : LatestCheckPoint.NO_TRANSACTION_ID;
-                return new LatestCheckPoint( latestCheckPoint, startEntryAfterCheckPoint,
-                        firstTxIdAfterCheckPoint, oldestVersionFound );
+                return latestCheckPoint( fromVersionBackwards, version, latestStartEntry, oldestVersionFound,
+                        latestCheckPoint );
             }
 
             version--;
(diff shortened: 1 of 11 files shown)


burqen:
@tinwelint If this address an issue reported by used then I think this would be a good fit the changelog.

chrisvest:
@tinwelint ClusterOverviewIT timed out. I've restarted the Linux build.

community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
```diff
@@ -105,19 +126,77 @@ public LatestCheckPoint find( long fromVersionBackwards ) throws IOException
             }
         }
 
-        return new LatestCheckPoint( null, latestStartEntry != null, oldestVersionFound );
+        boolean commitsAfterCheckPoint = latestStartEntry != null;
```
burqen: `commitsAfterCheckPoint` is a pretty confusing name here as we didn't find any checkpoint. How about `anyStartEntryFound` or something similar?

community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
```diff
@@ -91,9 +93,28 @@ public LatestCheckPoint find( long fromVersionBackwards ) throws IOException
...
-                        latestStartEntry.getStartPosition().compareTo( latestCheckPoint.getLogPosition() ) >= 0;
-                return new LatestCheckPoint( latestCheckPoint, commitsAfterCheckPoint , oldestVersionFound );
+                        latestStartEntry.getStartPosition().compareTo( target ) >= 0;
+                if ( !commitsAfterCheckPoint )
+                {
+                    if ( target.getLogVersion() < version )
+                    {
+                        // This check point entry targets a previous log file.
+                        // Go there and see if there's a transaction. Reader is capped to that log version.
+                        commitsAfterCheckPoint = extractFirstTxIdAfterPosition( target, target.getLogVersion() ) !=
```
burqen: If we cap the reader to targets log version and target transaction is last transaction in that log version, then it will not find any later transactions.
For example:

```
tx1
--- new log version
tx2
tx3
...
--- new log version
Checkpoint targeting tx 1
```

I guess we cap the reader to not have it read through the log versions that we already went over. But should we not cap the reader to the version previous to the one where we found the checkpoint entry?

community/kernel/src/test/java/org/neo4j/kernel/recovery/LatestCheckPointFinderTest.java
```diff
@@ -78,327 +79,415 @@ public LatestCheckPointFinderTest( Integer startLogVersion, Integer endLogVersio
...
-        when( reader.readLogEntry( any( ReadableLogChannel.class ) ) ).thenReturn( start, checkPoint, null );
+        StartEntry start = start();
+        setupLogFiles( logFile( start, checkPoint( start ) ) );
 
         // when
         LatestCheckPoint latestCheckPoint = finder.find( endLogVersion );
 
         // then
-        assertEquals( new LatestCheckPoint( checkPoint, true, endLogVersion ), latestCheckPoint );
+        assertLatestCheckPoint( true, true, NO_TRANSACTION_ID, endLogVersion, latestCheckPoint );
```
burqen, line 258: Is this test case really correct? (`latestLogFileContainingACheckPointAndAStartWithoutCommitAfter`)
There is no commit entry in the log but we still expect `commitsAfterLastCheckPoint` to be true.

tinwelint:
@burqen it doesn't per se, it was discovered by accident while investigating something in that area. Perhaps worth mentioning anyway, I don't know.

community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
```diff
@@ -91,9 +93,28 @@ public LatestCheckPoint find( long fromVersionBackwards ) throws IOException
...
+                    }
+                }
+
+                // Extract first transaction id after check point target position.
+                // Reader may continue into log files after the initial version.
+                long firstTxIdAfterCheckPoint = commitsAfterCheckPoint
+                        ? extractFirstTxIdAfterPosition( target, fromVersionBackwards )
+                        : LatestCheckPoint.NO_TRANSACTION_ID;
+                return new LatestCheckPoint( latestCheckPoint, commitsAfterCheckPoint,
+                        firstTxIdAfterCheckPoint, oldestVersionFound );
```
burqen: Can we simplify the logic in this whole block to something like:

```
if ( latestCheckPoint != null )
{
    // Is the latest start entry in this log file version later than what the latest check point targets?
    LogPosition target = latestCheckPoint.getLogPosition();

    long firstTxAfterCheckPoint = LatestCheckPoint.NO_TRANSACTION_ID;
    if ( latestStartEntry != null )
    {
        // We found a start entry in some of the log versions from current (including) or later
        firstTxAfterCheckPoint = extractFirstTxIdAfterPosition( target, fromVersionBackwards );
    }
    else if ( target.getLogVersion() < version )
    {
        // We did not find any start entries in current version (including) or later
        // We can safely cap reader to only read up to version previous to current
        firstTxAfterCheckPoint = extractFirstTxIdAfterPosition( target, version - 1 );
    }
    boolean commitsAfterCheckPoint = firstTxAfterCheckPoint != LatestCheckPoint.NO_TRANSACTION_ID;
    return new LatestCheckPoint( latestCheckPoint, commitsAfterCheckPoint, firstTxAfterCheckPoint, oldestVersionFound );
}
```

I tried it but it breaks `LatestCheckPointFinderTest.latestLogFileContainingACheckPointAndAStartWithoutCommitAfter` which I suspect is wrong, see comment below.

In any case, it would be super nice to simplify this block.

community/kernel/src/test/java/org/neo4j/kernel/recovery/LatestCheckPointFinderTest.java
```diff
@@ -19,50 +19,51 @@
...
-import static org.mockito.Mockito.when;
-
-import static org.neo4j.kernel.impl.transaction.log.entry.LogHeader.LOG_HEADER_SIZE;
-import static org.neo4j.kernel.impl.transaction.log.entry.LogHeaderWriter.encodeLogVersion;
+import static org.neo4j.io.ByteUnit.mebiBytes;
+import static org.neo4j.kernel.impl.transaction.log.PhysicalLogFile.NO_MONITOR;
+import static org.neo4j.kernel.recovery.LatestCheckPointFinder.LatestCheckPoint.NO_TRANSACTION_ID;
 
 @RunWith( Parameterized.class )
 public class LatestCheckPointFinderTest
```
burqen, line 59: Really nice clean up of this test! 👍

burqen (changes requested):
Behaviour change looks good over all but I think I've found an edge case where we will still slip.
There are also some comments on readability.

burqen:
@tinwelint About changelog: If no one is anticipating a fix for it, then it's just a bug like any other so let's not push it to changelog.

tinwelint:
@burqen mmh, think so

community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
tinwelint: The intention of the code is to start at the target (in your example tx1) and search from there up until max where it is now (including). Isn't that what you're seeing in the code and its tests?

community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
tinwelint: That block has two parts:

- decide whether there are commits after the last checkpoint
- which is the first transaction after the last checkpoint

my version tries to keep those apart. Your suggestions blends them together a bit, doesn't it? I too had some problems with it and found to separate them to be easier to reason about.

community/kernel/src/test/java/org/neo4j/kernel/recovery/LatestCheckPointFinderTest.java
tinwelint: This is related to your previous comment about that variable name, isn't it? If we rename it it's good?

burqen: If that is what we mean to test, then yes.

community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
burqen: I think this code
```
if ( !startEntryAfterCheckPoint )
{
    if ( target.getLogVersion() < version )
    {
        // This check point entry targets a previous log file.
        // Go there and see if there's a transaction. Reader is capped to that log version.
        startEntryAfterCheckPoint = extractFirstTxIdAfterPosition( target, target.getLogVersion() ) !=
                LatestCheckPoint.NO_TRANSACTION_ID;
    }
}
```
when `extractFirstTxIdAfterPosition` will start at log version of `target` and then stop at `target.getLogVersion()` which means it will not search any intermediate log versions between `target.getLogVersion` and wherever we found the checkpoint.

To be clear
```
--- Log version 1
tx1
--- Log version 2
tx2
tx3
...
--- Log version 3
Checkpoint targeting tx 1
```
Here, `target` is `tx1` which live in log version 1. This is where we will start our search when trying to `extractFirstTxIdAfterPosition`. What we want to find is `tx2` which is the first tx id after position, right? Oooor, is our `target` position the start of `tx1` and what we want to find is the commit entry of `tx1`? If so, is it safe for us to assume start and commit of `tx1` is in the same log version?

community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
burqen: Alright, that make sense but I still had quite some trouble understanding what was going on. Can we maybe make it clearer by extracting the logic into methods with some useful names?
I think that would make the logic separation clearer.

burqen:
@tinwelint Changes look good. I still have two points I'm not yet completely satisfied with though. Please have a look at the comment threads (which may be hidden because of latest commit).

community/kernel/src/main/java/org/neo4j/kernel/recovery/LatestCheckPointFinder.java
tinwelint: Transactions are written fully in one or the other tx log, never cut off in the middle. A check point can be made to target the log position either before or after tx1 and it doesn't matter, because `extractFirstTxIdAfterPosition` will go to the target log position and find the first commit entry, crossing log file boundaries if it needs to.

burqen:
@tinwelint Test failures are relevant. One case that does not seem to be covered is if you have to sequential checkpoint entries targeting same start entry.

tinwelint:
@burqen reverted some changes, but kept the extraction of latest check point

burqen (approved):

burqen:
@tinwelint I'll hold merge until we see a green build.

burqen:
@tinwelint Pushed a rebase to see if that fixes (seemingly unrelated) problem with doctest-compile.

PR #8805, opened 2017-02-13
Package installation now creates a neo4j group

spacecowboy:
* Use `useradd` instead of `adduser` in debian package as well for
  consistency's sake
* Set the systemd service to run in the neo4j group (as well as user)
* On upgrade, set the primary group of the neo4j user to neo4j. This
  is safe because the neo4j user has no group currently

Example for user on my current machine

```
$ groups neo4j
neo4j : nogroup
```

Leaving existing `chown` calls in place for now since we make no use
of the neo4j group at the moment and to make sure we don't break
anyone's setup.

Primary motivation was to change to `useradd` to be compatible with Suse. This prompted the group change because I figured it was nice to make the environment consistent on all platforms. It's also what all other services do.

### DIFF
--- a/new-packaging/src/common/neo4j.service
+++ b/new-packaging/src/common/neo4j.service
@@ -6,6 +6,7 @@ After=network.target
 ExecStart=/usr/share/neo4j/bin/neo4j console
 Restart=on-failure
 User=neo4j
+Group=neo4j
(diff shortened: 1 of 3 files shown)


spacecowboy:
Tested manually on OpenSuse and it installed correctly with these changes

benbc:
Looks good.

PR #8807, opened 2017-02-13
Better error message when not able to fullfill hints

systay:


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/planner/logical/steps/verifyBestPlan.scala
```diff
@@ -39,10 +39,15 @@ object verifyBestPlan extends PlanTransformer[PlannerQuery] {
...
           // unknown planner issue failed to find plan (without regard for differences in hints)
-          throw new CantHandleQueryException(s"Expected \n$expected \n\n\nInstead, got: \n$constructed")
+          throw new InvalidArgumentException(s"Expected \n$expected \n\n\nInstead, got: \n$constructed")
         } else {
           // unknown planner issue failed to find plan matching hints (i.e. "implicit hints")
-          throw new CantHandleQueryException(s"Expected \n${expected.allHints} \n\n\nInstead, got: \n${constructed.allHints}")
+          val expectedHints = expected.allHints.mkString(System.lineSeparator())
+          val actualHints = constructed.allHints.mkString(System.lineSeparator())
+          val message =
+            s"Something went when trying to fulfill the index hints of your query.\n" +
```
thobe: _Something went ..._ where did it go? I think you meant to say _Something went *wrong*..._, I would actually prefer a slightly different wording that doesn't sound as much like we don't know what is going on.

Furthermore we have hints that are not index hints, is this part of the code only dealing with index hints as the message says, or should the message change to talk about hints in general. (I'm thinking of how a scan hint is not technically an index hint, and can be used in the same place).

How about `Failed to fulfil the hints of the query.`?

I'm also wondering if we instead should be using a more specific exception type, rather than encoding all of this information in a message only.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/UsingAcceptanceTest.scala
```diff
@@ -702,6 +702,19 @@ class UsingAcceptanceTest extends ExecutionEngineFunSuite with NewPlannerTestSup
...
+    graph.createIndex("PERSON", "id")
+
+    val result = intercept[RuntimeException](graph.execute(
+      """MATCH (a:PERSON {id: 1})-->(b:PERSON {id: 1})
+        |USING INDEX a:PERSON(id)
+        |USING INDEX b:PERSON(id)
+        |RETURN *""".stripMargin)
+    )
+
+    result.getMessage should startWith("Something went when trying to fulfill the index hints of your query")
```
thobe: I would like it if we also asserted on the actual hints here as well. The information about which hints the engine was unable to fulfil being formatted in a readable way is one of the most important pieces of information IMO.

thobe (changes requested):
Seems like a simple change, but I believe we can do better.

systay:
@thobe Pushed updates

thobe (approved):

PR #8808, opened 2017-02-13
GraphDatabaseFactory can be shared with other threads

tinwelint:
its state was non-volatile and so other threads which may receive a
GraphDatabaseFactory would not read its state correctly when creating
databases from it.

This fixes an issue which SchemaIndexHaIT ran into

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseFactoryState.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseFactoryState.java
@@ -19,11 +19,11 @@
  */
 package org.neo4j.graphdb.factory;
 
-import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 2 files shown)


chrisvest (approved):

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseFactoryState.java
```diff
@@ -37,29 +37,32 @@
...
+    private volatile LogProvider userLogProvider;
     private final Map<String,URLAccessRule> urlAccessRules;
 
     public GraphDatabaseFactoryState()
     {
-        settingsClasses = new ArrayList<>();
+        settingsClasses = new CopyOnWriteArrayList<>();
         settingsClasses.add( GraphDatabaseSettings.class );
-        kernelExtensions = new ArrayList<>();
+        kernelExtensions = new CopyOnWriteArrayList<>();
```
chrisvest, line 52: The `removeKernelExtensions` method now needs to be implemented differently, because you cannot remove from a `CopyOnWriteArrayList` via its iterator.

chrisvest: See the test failures.

chrisvest:
@tinwelint Fixed it for you :)

tinwelint:
@chrisvest brilliant, thanks!

PR #8809, opened 2017-02-13
Appends to FailureStorage on multiple failures

tinwelint:
Noticed that first exception in inex FailureStorage could be overwritten
by second exception logged on e.g. closing populator. Since both could be
interesting failures are now appended instead of written at the beginning.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/FailureStorage.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/FailureStorage.java
@@ -115,6 +115,10 @@ public class FailureStorage
         File failureFile = failureFile();
         try ( StoreChannel channel = fs.open( failureFile, "rw" ) )
         {
+            byte[] existingData = new byte[(int) channel.size()];
+            channel.read( ByteBuffer.wrap( existingData ) );
+            channel.position( lengthOf( existingData ) );
+
             byte[] data = UTF8.encode( failure );
             channel.write( ByteBuffer.wrap( data, 0, Math.min( data.length, MAX_FAILURE_SIZE ) ) );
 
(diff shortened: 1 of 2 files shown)


community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/FailureStorage.java
```diff
@@ -115,6 +115,10 @@ public synchronized void storeIndexFailure( String failure ) throws IOException
         File failureFile = failureFile();
         try ( StoreChannel channel = fs.open( failureFile, "rw" ) )
         {
+            byte[] existingData = new byte[(int) channel.size()];
+            channel.read( ByteBuffer.wrap( existingData ) );
+            channel.position( lengthOf( existingData ) );
```
chrisvest, line 120: Why do you need to read the existing data? Can't you just `channel.position( channel.size() )`? Also, the read might not read all of the data in the first go. You also have a `channel.close` further down in this method.

tinwelint: Because a failure file is pre-allocated with all-zeros, so that a failure can be written even in an disk-space-full scenario. Sure, OK correct read (also in existing method `loadIndexFailure()` should be correct.

chrisvest: We can still run into `no space left on device` failures with a file that's preallocated with zeros. File systems that support sparse files will likely not actually write the zeros down, but just note how many there are. Similar problem with file systems that use compression. Even worse with file systems that do copy-on-write, since then even overwriting a file filled with random data can cause out of space failures. But I guess it's better than nothing; I don't think we can reasonably guard for all possible scenarios that can lead to running out of space.

tinwelint: Yes, I think you're right. And better than nothing is at least something. The fundamentals of the design of this class is a bit out of scope for that change tho, isn't it?

PR #8813, opened 2017-02-13
Close bolt state machine only from worker thread

lutovich:
Bolt state machine was closed as part of `RunnableBoltWorker#halt()` which could be called from different thread (by Netty via callbacks `channelInactive`, `handlerRemoved` and `exceptionCaught`).

This commit makes `#halt()` only mark machine for termination and raise a stop signal for the worker. So state machine is only closed by the worker and only in it's thread.


zhenlineo:
LGTM

PR #8814, opened 2017-02-14
GB+Tree rebalance and merge

burqen:
**What?**
Implementation of rebalance and merge.

**Why?**
When removing keys from GB+Tree we want to avoid having leaves with key count less than half of max key count. This is done by rebalancing keys among sibling nodes if they between themselves have enough to at least get `maxKeyCount/2` keys each. Otherwise we will merge them together, effectively deleting the node that we merge from.

**Problems?**
Seek in GB+Tree rely on no keys ever being moved from right to left in the tree. Therefore we are not allowed to rebalance or merge from right to left. This creates a somewhat asymmetric implementation.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/InternalTreeLogic.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/InternalTreeLogic.java
@@ -1077,30 +1077,30 @@ class InternalTreeLogic<KEY,VALUE>
     private void updateRightmostChildInLeftSibling( PageCursor cursor, long childPointer, long stableGeneration,
             long unstableGeneration ) throws IOException
     {
-        long currentPageId = cursor.getCurrentPageId();
         long leftSibling = bTreeNode.leftSibling( cursor, stableGeneration, unstableGeneration );
         // Left sibling is not allowed to be NO_NODE here because that means there is a child node with no parent
         PointerChecking.checkPointer( leftSibling, false );
 
-        bTreeNode.goTo( cursor, "left sibling", leftSibling );
-        int keyCount = bTreeNode.keyCount( cursor );
-        bTreeNode.setChildAt( cursor, childPointer, keyCount, stableGeneration, unstableGeneration );
-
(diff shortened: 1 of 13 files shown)


community/index/src/main/java/org/neo4j/index/internal/gbptree/InternalTreeLogic.java
```diff
@@ -849,11 +854,55 @@ VALUE remove( PageCursor cursor, StructurePropagation<KEY> structurePropagation,
...
 
+    private void updateRightMostChildInLeftSibling( PageCursor cursor, long child, long stableGeneration,
+            long unstableGeneration ) throws IOException
+    {
+        long currentPageId = cursor.getCurrentPageId();
+        long leftSibling = bTreeNode.leftSibling( cursor, stableGeneration, unstableGeneration );
+        // Left sibling is not allowed to be NO_NODE here because that means there is a child node with no parent
+        PointerChecking.checkPointer( leftSibling, false );
+
+        bTreeNode.goTo( cursor, "left sibling", leftSibling );
```
tinwelint: In other places where we temporarily go to another node and then back we usually use a linked cursor, do that here too?

community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeIT.java
```diff
@@ -77,78 +76,80 @@
                 layout, 0/*use whatever page cache says*/, monitor );
     }
 
-    @After
-    public void consistencyCheckAndClose() throws IOException
-    {
-        threadPool.shutdownNow();
-        index.consistencyCheck();
```
tinwelint: The need to merge this "After" code into the test seems to be a bug in RandomRule, doesn't it? It's fine, but it should be fixed so that tests generally don't need to go through this process too.

tinwelint (approved):
Looks great. Much of the code is documented or self-documenting in a way. I did some QA of alternating inserting,removing,checkpoint and can see that file size stays flat.

PR #8817, opened 2017-02-14
introduce filters and tags for server policy load balancing

martinfurmanski:
This is the first step in building up the server-side policy based
load balancing strategy. The idea is that servers are tagged
with one or more tags and these are used in a server-side configured
scheme of filters to select those which match. Different policies can
select different sets of tags and there can be other filters attached apart
from tag-based filters.

Every policy has a name which the client can use to bind to a particular
policy by supplying that name in the client supplied context, as supported
by the V2 version of GetServers.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
@@ -21,7 +21,9 @@ package org.neo4j.causalclustering.discovery;
 
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
+import com.hazelcast.core.MultiMap;
 
+import java.util.List;
 import java.util.function.Function;
 
 import org.neo4j.causalclustering.core.CausalClusteringSettings;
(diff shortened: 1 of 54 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/ClusterTopology.java
```diff
@@ -36,8 +36,8 @@ public ClusterTopology( CoreTopology coreTopology, ReadReplicaTopology readRepli
 
     public Optional<CatchupServerAddress> find( MemberId upstream )
     {
-        Optional<CoreAddresses> coreAddresses = coreTopology.find( upstream );
-        Optional<ReadReplicaAddresses> readReplicaAddresses = readReplicaTopology.findAddressFor( upstream );
+        Optional<CoreServerInfo> coreAddresses = coreTopology.find( upstream );
```
jimwebber: rename the local variables too?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
```diff
@@ -37,6 +39,9 @@
 import static org.neo4j.causalclustering.discovery.HazelcastClusterTopology.READ_REPLICA_BOLT_ADDRESS_MAP_NAME;
 import static org.neo4j.causalclustering.discovery.HazelcastClusterTopology.READ_REPLICA_MEMBER_ID_MAP_NAME;
 import static org.neo4j.causalclustering.discovery.HazelcastClusterTopology.READ_REPLICA_TRANSACTION_SERVER_ADDRESS_MAP_NAME;
+import static org.neo4j.causalclustering.discovery.HazelcastClusterTopology.SERVER_TAGS_MULTIMAP_NAME;
```
jimwebber, line 42: Doesn't this leak information from the plugin into the HZ instance which should be unaware of plugins?

jimwebber: Should the plugins get a  `ReadReplicaTopologyService`instance injected instead?

martinfurmanski: I think tagging could be seen as a general concept for identifying servers or sets of servers, and you could imagine building several things on top of that, e.g. load balancing and replication policies, some future security policy, general affinity, etc. So from that perspective I don't think the layering is bad.

If you look at the setting I simply named it `causal_clustering.server_tags` which is inline with this way of looking at it (e.g. no load_balancing in the name).

Is tagging really the most appropriate design for identification meant for other purposes? Future will have to show.

martinfurmanski: On the second point, yes, and I will rename that to TopologyService.

PR #8823, opened 2017-02-15
Run acceptance tests in source code mode

henriknyman:
Run the cucumber acceptance and compatibility spec suite tests also
with source code generation mode for the compiled runtime.

This adds a cucumber plugin for prepending Cypher pre-parser options to the executing queries.

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/java/cypher/AcceptanceSpecSuiteTest.java
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/java/cypher/AcceptanceSpecSuiteTest.java
@@ -74,6 +74,24 @@ public class AcceptanceSpecSuiteTest
     @RunWith( Cucumber.class )
     @CucumberOptions(
             plugin = {
+                    DB_CONFIG + "cost-compiled.json",
+                    HTML_REPORT + SUITE_NAME + "/cost-compiled",
+                    JSON_REPORT + SUITE_NAME + "/cost-compiled",
+                    BLACKLIST_PLUGIN + "cost-compiled.txt",
+                    CYPHER_OPTION_PLUGIN + "cost-compiled-sourcecode.txt"
+            },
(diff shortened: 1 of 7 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/java/cypher/AcceptanceSpecSuiteTest.java
```diff
@@ -74,6 +74,24 @@
     @RunWith( Cucumber.class )
     @CucumberOptions(
             plugin = {
+                    DB_CONFIG + "cost-compiled.json",
+                    HTML_REPORT + SUITE_NAME + "/cost-compiled",
+                    JSON_REPORT + SUITE_NAME + "/cost-compiled",
```
davidegrohmann: Shouldn't reports have different names between the 2 cost-compiled runs?

enterprise/cypher/compatibility-spec-suite/src/test/java/cypher/CompatibilitySpecSuiteTest.java
```diff
@@ -74,6 +74,24 @@
     @RunWith( Cucumber.class )
     @CucumberOptions(
             plugin = {
+                    DB_CONFIG + "cost-compiled.json",
+                    HTML_REPORT + SUITE_NAME + "/cost-compiled",
+                    JSON_REPORT + SUITE_NAME + "/cost-compiled",
```
davidegrohmann: Shouldn't reports have different names between the 2 cost-compiled runs?

enterprise/cypher/acceptance-spec-suite/src/test/java/cypher/AcceptanceSpecSuiteTest.java
henriknyman: Oh, good catch

PR #8829, opened 2017-02-15
Re-enable relationship counts from count store

systay:


jexp:
Nice!

Von meinem iPhone gesendet

henriknyman (commented):
Looks good. We just need to change some tests to run with all runtimes.

PR #8841, opened 2017-02-16
Raise availability guard when core or read replica copy store

lutovich:
PR consists of two commits:

* _Raise availability guard when core or read replica copy store_
Instance might be forced to copy store from another instance if it requires to pull a transaction log that has already been pruned away. This could happen when transaction pulling (catch up) experience problems like network outage. Database is not operational during store copy, most components including the datasource are shutdown so no transactions can be served.
However it was possible for users (bolt or embedded) to attempt start new transactions which would hang because of kernel components being stopped. In similar situations availability guard must be raised to block new transactions.
This commit makes `LocalDatabase` raise availability guard when it is stopped. `LocalDatabase` is the component used by cores and read replicas to manage the lifecycle of the datasource. It is restarted during store copying.

 * _Better error message when db stopped for store copy_
This commit improves error message when users (bolt or embedded) try to start a transaction while cluster member is copying store. It is done using a special `AvailabilityRequirement` for the `AvailabilityGuard` when local database is stopped to perform a store copy.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
@@ -47,7 +47,10 @@ import static org.neo4j.kernel.AvailabilityGuard.availabilityRequirement;
 
 public class LocalDatabase implements Lifecycle
 {
-    private static final AvailabilityRequirement NOT_STOPPED = availabilityRequirement( "Database is stopped" );
+    private static final AvailabilityRequirement NOT_STOPPED =
+            availabilityRequirement( "Database is stopped" );
+    private static final AvailabilityRequirement NOT_COPYING_STORE =
+            availabilityRequirement( "Database is stopped to copy store from another cluster member" );
 
     private final File storeDir;
 
@@ -59,7 +62,7 @@ public class LocalDatabase implements Lifecycle
 
     private volatile StoreId storeId;
     private volatile DatabaseHealth databaseHealth;
-    private boolean started = false;
+    private AvailabilityRequirement currentRequirement;
 
     private volatile TransactionCommitProcess localCommit;
 
@@ -74,7 +77,7 @@ public class LocalDatabase implements Lifecycle
         this.availabilityGuard = availabilityGuard;
         this.log = logProvider.getLog( getClass() );
 
-        raiseAvailabilityGuard();
+        raiseAvailabilityGuard( NOT_STOPPED );
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
```diff
@@ -80,23 +94,29 @@ public synchronized void start() throws Throwable
...
+     */
+    public void stopForStoreCopy() throws Throwable
+    {
+        stopWithRequirement( NOT_COPYING_STORE );
     }
 
     public boolean isAvailable()
     {
-        return started;
+        return currentRequirement == null;
```
davidegrohmann, line 119: should be this read through a synchronize?

lutovich: Right, this method should be synchronized. Will fix.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
```diff
@@ -193,4 +213,32 @@ public TransactionCommitProcess getCommitProcess()
...
     }
+
+    private synchronized void stopWithRequirement( AvailabilityRequirement requirement ) throws Throwable
+    {
+        log.info( "Stopping, reason: " + requirement.description() );
+        databaseHealth = null;
+        localCommit = null;
+        dataSourceManager.stop();
+
+        raiseAvailabilityGuard( requirement );
```
martinfurmanski: Isn't it better to raise this first?

martinfurmanski: I am also concerned that a transaction might still be in progress running, just about to start or already blocking on something before this was raised, but either way this should take care of a a few cases. Perhaps the intention is for internal components to continuously poll on the availability guard for all blocking matters and thus eventually get released. Let's revisit in the future if necessary and perhaps add that if missing somewhere.

martinfurmanski (approved):

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
lutovich: Yeah, it is definitely better to raise the guard before stopping the datasource. Fixed now.

I think if transaction gets in-between guard and `NeoStoreDataSource#stop()` it'll be asked to terminate during `NSDS#stop()`. Locks are termination aware so transactions should not just block somewhere.

martinfurmanski:
@lutovich Sorry to be a pain, but just realised we probably want this starting from 3.1.

lutovich:
@martinfurmanski I'll open a new PR for 3.1 branch.

PR #8843, opened 2017-02-17
Support Top N in the compiled runtime

henriknyman:
Adds specialized support for the combination of ORDER BY and LIMIT that
only needs to keep the LIMIT number of rows sorted (also known as Top).

Also uses estimated cardinality for initial size of full sort table.


systay:
Moved this commit over to #8842

PR #8848, opened 2017-02-17
defend on shutdown after failed startup

mneedham:


enterprise/com/src/main/java/org/neo4j/com/Server.java
```diff
@@ -142,7 +143,8 @@ public Server( T requestTarget, Configuration config, LogProvider logProvider, i
         this.txVerifier = txVerifier;
         this.byteCounterMonitor = byteCounterMonitor;
         this.requestMonitor = requestMonitor;
-        this.connectedSlaveChannels = new IdleChannelReaper( this, logProvider, clock, config.getOldChannelThreshold() );
+        this.connectedSlaveChannels = new IdleChannelReaper( this, logProvider, clock, config.getOldChannelThreshold
+                () );
```
martinfurmanski: odd line break

enterprise/com/src/main/java/org/neo4j/com/Server.java
```diff
@@ -200,17 +202,27 @@ public void start() throws Throwable
...
+
+        successfullyStarted = true;
     }
 
     @Override
     public void stop() throws Throwable
     {
         String name = getClass().getSimpleName();
-        msgLog.info( name + " communication server shutting down and unbinding from  " + socketAddress );
+        if ( successfullyStarted )
```
martinfurmanski, line 212: To my knowledge LifeSupport should not call stop unless start was successful, so this behaviour should already be in place.

martinfurmanski (changes requested):
I think LifeSupport should already protect against this. Can you try proving it with a test if you think this behaviour is possible?

The branch is also named 3.0 but PR is against 3.2. Is 3.0 the version where some issue was seen? Perhaps expand the commit message with an explanation of what spurred this.

mneedham:
@martinfurmanski  I agree, but this is the stack trace which suggests that isn't happening:

```
2017-02-12 15:38:37.346+0000 INFO  [o.n.b.BackupServer] BackupServer communication server shutting down and unbinding from  null
2017-02-12 15:38:37.347+0000 ERROR [o.n.k.h.c.m.HighAvailabilityModeSwitcher] Error while trying to switch to slave
java.lang.NullPointerException
	at org.neo4j.com.Server.stop(Server.java:212)
	at org.neo4j.backup.OnlineBackupKernelExtension.stop(OnlineBackupKernelExtension.java:163)
	at org.neo4j.kernel.ha.cluster.SwitchToSlave.stopServices(SwitchToSlave.java:490)
	at org.neo4j.kernel.ha.cluster.SwitchToSlaveBranchThenCopy.stopServicesAndHandleBranchedStore(SwitchToSlaveBranchThenCopy.java:181)
	at org.neo4j.kernel.ha.cluster.SwitchToSlaveBranchThenCopy.checkDataConsistency(SwitchToSlaveBranchThenCopy.java:153)
	at org.neo4j.kernel.ha.cluster.SwitchToSlave.executeConsistencyChecks(SwitchToSlave.java:465)
	at org.neo4j.kernel.ha.cluster.SwitchToSlave.switchToSlave(SwitchToSlave.java:220)
	at org.neo4j.kernel.ha.cluster.modeswitch.HighAvailabilityModeSwitcher$1.run(HighAvailabilityModeSwitcher.java:354)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
	at org.neo4j.helpers.NamedThreadFactory$2.run(NamedThreadFactory.java:104)
```

`channelGroup` is presumably the object that is throwing the NPE here. That's only set during the `start` method and isn't set to `null` explicitly anywhere in the code as far as I can see. 

We're also unbinding from `null` which is the value of `socketAddress`which is only set in the `start` method as well. 

So it seems like lifecycle believes we can be stopped which makes sense as we should have been started up via `KernelExtensions`. 

Unless you can think of another way we can get this behaviour...

martinfurmanski (approved):

systay:
@martinfurmanski @mneedham Should this be merged or closed?

systay:
Please re-open if still valid

martinfurmanski:
@mneedham Right... there is no LifeSupport even.

PR #8849, opened 2017-02-17
Don't hold schema lock when creating uniqueness constraint

tinwelint:


enterprise/ha/src/main/java/org/neo4j/kernel/ha/MasterTransactionCommitProcess.java
```diff
@@ -19,46 +19,72 @@
...
  */
 public class MasterTransactionCommitProcess implements TransactionCommitProcess
 {
+    /**
+     * Detector of transactions coming in from slaves which should acquire the shared schema lock before
+     * being applied (and validated for application).
+     */
+    private static final Visitor<StorageCommand,IOException> REQUIRES_SHARED_SCHEMA_LOCK = command ->
+            command instanceof NodeCommand && mayResultInIndexUpdates( (NodeCommand) command ) ||
+            command instanceof PropertyCommand && mayResultInIndexUpdates( (PropertyCommand) command );
```
fickludd, line 53: This will happen a lot, resulting in schema lock grabs. Would it make sense to be more clever about whether there is an index/constraint for the actually changed label / property? I guess not in 3.0 HA...

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneSchemaIndex.java
```diff
@@ -97,6 +97,7 @@ public IndexReader getIndexReader() throws IOException
     public void verifyUniqueness( PropertyAccessor accessor, int propertyKeyId )
             throws IOException, IndexEntryConflictException
     {
+        flush( true );
```
fickludd, line 100: This I did not understand, why do we need to merge partitions before verifying uniqueness? To force the SimpleUniquenessVerifier?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/ConstraintIndexCreator.java
```diff
@@ -36,27 +36,52 @@
...
+     * which doesn't acquire any locking, merely adds tx state and commits so that the index rule is applied
+     * to the store, which triggers the index population</li>
+     * <li>Release the SCHEMA WRITE lock</li>
+     * <li>Await index population to complete</li>
+     * <li>Acquire the SCHEMA WRITE lock (effectively blocking concurrent transactions changing
+     * data related to this constraint, and it so happens, most other transactions as well) and verify
+     * the uniqueness of the built index</li>
+     * <li>Leave this method, knowing that the uniqueness constraint rule will be added to tx state
+     * and this tx committed, which will create the uniqueness constraint</li>
+     * </ol>
```
fickludd, line 84: Nice docs!

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexAccessor.java
```diff
@@ -90,6 +91,17 @@
...
+    /**
+     * Verifies that each value in this index is unique.
+     * Index is guaranteed to not change while this call executes.
+     *
+     * @param propertyAccessor {@link PropertyAccessor} for accessing properties from database storage
+     * in the event of conflicting values.
+     * @throws IndexEntryConflictException for first detected uniqueness conflict, if any.
+     * @throws IOException on error reading from source files.
+     */
+    void verifyDeferredConstraints( PropertyAccessor propertyAccessor ) throws IndexEntryConflictException, IOException;
```
fickludd, line 103: It is a little bit strange that the `verifyDeferredConstraints` always verifies that the index has unique entries, while the IndexPopulator variant only does so if the Index is actually supposed to be unique. The way these are used, that makes sense, but for understanding it would be nicer with wither consistent behaviour or different method names.

fickludd (approved):
I think this looks good and cannot find any logical problems with the new algorithm for releasing the SCHEMA WRITE lock during uniqueness constraint index population. In fact, shouldn't this prove better protection against concurrent uniqueness-violating updates during index population?

Got some comments that can be addressed or not depending on ambition. For what it's worth, I'm fine with this.

tinwelint:
Yes, I think with the local (on master) re-acquire of the shared SCHEMA lock should be a safer and more simplistic approach, which could potentially remove the need for other hack-ish guards put in place around constraint creation in HA.

enterprise/ha/src/main/java/org/neo4j/kernel/ha/MasterTransactionCommitProcess.java
tinwelint: Exactly, there have been some work towards more fine-grained checks in 3.2, haven't there?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneSchemaIndex.java
tinwelint: It turns out that this was a problem previously too, it just happened very rarely. If the index wasn't flushed and merged properly then the uniqueness check, which checks term frequencies on all "leaves" would sometimes see multiple leaves for the same key and hence miss non-uniqueness. Merging all segments together has lucene report correct counts.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/ConstraintIndexCreator.java
tinwelint: Cheers

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexAccessor.java
tinwelint: Oh, that's right. It just so happens to be correct since this method is only called when creating a uniqueness constraint. I'll put this into the uniqueness variant only.

tinwelint: Hmm, it seems that a LuceneIndexAccessor doesn't quite know whether it's a uniqueness index or not. Would it be useful to pass it in only to check that and throw exception if that's the case? I don't feel it would add much, would it?

enterprise/ha/src/main/java/org/neo4j/kernel/ha/MasterTransactionCommitProcess.java
fickludd: On way, but we have not considered HA yet. Maybe when the functionality exists, we can use it here too. Not in scope for this review, but we cannot lock on individual indexes?

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexAccessor.java
fickludd: The easiest thing would be to rename the IndexAccessor method to `verifyUniqueEntries` or something :). That would but the responsibility with the caller.

enterprise/ha/src/main/java/org/neo4j/kernel/ha/MasterTransactionCommitProcess.java
tinwelint: We cannot no, but I think there's work happening in the near future to that end

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexAccessor.java
tinwelint: Yes, although I opted to keep those two methods the same to not introduce confusion. I donät think it makes sense to name them differently and I wanted to keep changes small due to being in a maintenance branch.

PR #8850, opened 2017-02-18
Fix behaviour of LIMIT 0

systay:
Using LIMIT to batch updates is a common pattern. LIMIT 0 should mean that nothing is changed in the graph, but this was not the case before this change.

changelog: Using `LIMIT` with zero or a negative argument will never pull any rows.


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/pipes/QueryState.scala
```diff
@@ -45,6 +45,8 @@ class QueryState(val query: QueryContext,
                    new SingleThreadedLRUCache(maxSize = 16)) {
   private var _pathValueBuilder: PathValueBuilder = null
 
+  def createEmptyExecutionContext(): ExecutionContext = initialContext.getOrElse(ExecutionContext.empty)
```
Mats-SX: This name seems a bit misleading. Perhaps `createOrGetInitialContext()`?

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/pipes/SkipPipe.scala
```diff
@@ -34,27 +34,8 @@ case class SkipPipe(source: Pipe, exp: Expression)
     if(input.isEmpty)
       return Iterator.empty
 
-    implicit val s = state
+    val skip = asInt(exp(state.createEmptyExecutionContext())(state))
```
Mats-SX: Nice simplification!

systay:
Your kind feedback has been acted on, @Mats-SX

Mats-SX:
@systay I'm not convinced this is where we will end up with `LIMIT 0` on updating queries, but this does fix the weird pull-one behaviour of read-only queries. I'll merge it.

PR #8853, opened 2017-02-18
Fixed tests that were passing as a side effect of planner ordering

craigtaverner:
Fixed tests that were passing as a side effect of planner ordering, not actual cost comparisons. In later versions of Neo4j, notably 3.2, enhancements to the planner changed the generation order of plans to consider, and caused these tests to fail. It seems reasonable to fix this as far back as 2.3 so that they do not complicate any potential bug-fixing that might affect planner ordering.


Mats-SX:
Doctests failure expected: not supported for `2.3`.

Mats-SX (approved):

PR #8854, opened 2017-02-20
3.2 composite lucene initial support

ragadeeshu:
Support for population and seek.


community/kernel/src/test/java/org/neo4j/kernel/api/index/CompositeIndexAccessorCompatibility.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Test
+    public void testIndexSeekAndScan() throws Exception
+    {
+        updateAndCommit( asList( IndexEntryUpdate.add( 1L, descriptor, "a", "a" ),
+                IndexEntryUpdate.add( 2L, descriptor, "a", "a" ), IndexEntryUpdate.add( 3L, descriptor, "b", "b" ) ) );
+
+        assertThat( getAllNodesWithProperties( "a", "a" ), equalTo( asList( 1L, 2L ) ) );
+        assertThat( getAllNodes(), equalTo( asList( 1L, 2L, 3L ) ) );
```
fickludd: To make it clear that the implementation considers both the properties, I would add `IndexEntryUpdate.add( 1L, descriptor, "a", "b" )`

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/schema/DatabaseCompositeIndexAccessorTest.java
```diff
@@ -0,0 +1,298 @@
...
+
+    @Parameterized.Parameter
+    public IOFunction<DirectoryFactory,LuceneIndexAccessor> accessorFactory;
+
+    private LuceneIndexAccessor accessor;
+    private final long nodeId = 1, nodeId2 = 2;
+    private final Object[] values = {"value1", "value2"}, value2 = {40, 42};
+    private DirectoryFactory.InMemoryDirectoryFactory dirFactory;
+    private static final NewIndexDescriptor indexDescriptor = NewIndexDescriptorFactory.forLabel( 0, 0 );
+    private static final NewIndexDescriptor uniqueIndexDescriptor = NewIndexDescriptorFactory.uniqueForLabel( 0, 0 );
```
fickludd: These descriptors have only one property each, that's probably wrong?

fickludd (commented):
Some minor things:

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
```diff
@@ -265,13 +274,25 @@ private void setId( long id )
...
+        private void setValues( Object... values )
+        {
+            removeAllValueFields();
+            for ( int i = 0; i < values.length; i++ )
+            {
+                ValueEncoding encoding = ValueEncoding.forValue( values[i] );
+                Field reusableField = getFieldWithValue( i, "composite", encoding, values[i] );
+                document.add( reusableField );
+            }
+        }
```
chrisvest: Keeping existing indexes working? 👍

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
```diff
@@ -250,6 +256,7 @@ protected AcceptStatus accept( BytesRef term ) throws IOException
         private final Field idValueField;
 
         private final Map<ValueEncoding,Field> valueFields = new EnumMap<>( ValueEncoding.class );
+        private final ArrayList<Map<ValueEncoding,Field>> compositeValueFields = new ArrayList<>();
```
chrisvest: The `valueFields` field is now never used.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneIndexAccessor.java
```diff
@@ -167,22 +167,22 @@ public void remove( PrimitiveLongSet nodeIds ) throws IOException
...
             writer.updateDocument( LuceneDocumentStructure.newTermForChangeOrRemove( nodeId ),
-                    LuceneDocumentStructure.documentRepresentingProperty( nodeId, value ) );
+                    LuceneDocumentStructure.documentRepresentingProperties( nodeId, values ) );
         }
 
-        private void add( long nodeId, Object value ) throws IOException
+        private void add( long nodeId, Object[] values ) throws IOException
         {
-            writer.addDocument( LuceneDocumentStructure.documentRepresentingProperty( nodeId, value ) );
+            writer.addDocument( LuceneDocumentStructure.documentRepresentingProperties( nodeId, values ) );
```
chrisvest, line 191: Does this mean that we will always write in a composite encoding? Do we do a migration of old indexes on upgrade, or does the querying part make sure to query in a way that finds both old and new encodings?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
```diff
@@ -284,6 +277,11 @@ private void setValue( Object value )
 
         private void setValues( Object... values )
         {
+            if( values.length == 1 )
+            {
+                setValue( values[0] );
+                return;
+            }
```
chrisvest: Aha, _this_ is how single-property indexes retain their format. Very good.

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/LuceneSchemaIndexUniquenessVerificationIT.java
```diff
@@ -298,7 +298,7 @@ private void insert( Collection<PropertyValue> data ) throws IOException
         PropertyValue[] dataArray = data.toArray( new PropertyValue[data.size()] );
         for ( int i = 0; i < dataArray.length; i++ )
         {
-            Document doc = LuceneDocumentStructure.documentRepresentingProperty( i, dataArray[i].value );
+            Document doc = LuceneDocumentStructure.documentRepresentingProperties( (long) i, dataArray[i].value );
```
chrisvest: Why this cast to `long`? It should automatically coerce. Similarly in `SimpleUniquenessVerifierTest`. Not a big issue, though.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
```diff
@@ -267,26 +267,13 @@ private void setId( long id )
...
-            document.add( reusableField );
-        }
-
         private void setValues( Object... values )
         {
-            if( values.length == 1 )
-            {
-                setValue( values[0] );
-                return;
-            }
```
chrisvest: Oh...

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
```diff
@@ -143,22 +143,22 @@ public static Query newRangeSeekByStringQuery( String lower, boolean includeLowe
...
+        Term term = new Term( ValueEncoding.String.key( 0 ), prefix );
         return USE_LUCENE_STANDARD_PREFIX_QUERY ? new PrefixQuery( term ) :
                                      new PrefixMultiTermsQuery( term );
     }
 
     public static Query newSuffixStringQuery( String suffix )
     {
         String searchTerm = QueryParser.escape( suffix );
-        Term term = new Term( ValueEncoding.String.key(), "*" + searchTerm );
+        Term term = new Term( ValueEncoding.String.key( 0 ), "*" + searchTerm );
```
chrisvest, line 169: This still needs to be updated so we get the field position right, doesn't it?

Likewise for the `NonUniqueLucenePopulator` and `NonUniqueLucenePopulatingUpdater`, right?

Otherwise it looks like we are only populating our composite indexes with the first property.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
chrisvest: So now we're back to all indexes being composite.

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructureTest.java
```diff
@@ -210,7 +210,7 @@ public void shouldBuildWildcardQueries() throws Exception
         WildcardQuery query = (WildcardQuery) LuceneDocumentStructure.newWildCardStringQuery( "foo" );
 
         // then
-        assertEquals( "string", query.getField() );
+        assertEquals( "0string", query.getField() );
```
chrisvest: I'm still inclined on keeping the existing index structure for single-property indexes.

community/kernel/src/test/java/org/neo4j/kernel/api/index/CompositeIndexAccessorCompatibility.java
chrisvest: That, and maybe the tests on composite indexes should be added to the existing `*Compatibility` tests.

chrisvest (changes requested):
I had a number of comments. Mostly, I'm not a fan of changing the index format for the single-property indexes, and I think there's a problem with `NonUniqueLucenePopulator` and `NonUniqueLucenePopulatingUpdater`, and how they build & encode index updates for Lucene. Specifically, it looks like they only add the first property to indexes, regardless of it being composite or not.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
ragadeeshu: Not for queries like these - currently only seek is supported. Suffix etc. still only support working with one property.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
ragadeeshu: Yes, all indexes being composite reduces complexity in a lot of places.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
ragadeeshu: This is leftovers, but not in the way you would think, I have a hunch that it actually is compositeVauleFields that should go.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneIndexAccessor.java
ragadeeshu: Upgrade is probably broken right now. We should test that.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
ragadeeshu: However, this is a real issue in encodedStringValue. This needs testing.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
ragadeeshu: Sorry :-1: They probably went :bomb: :boom:

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/schema/DatabaseCompositeIndexAccessorTest.java
```diff
@@ -152,24 +152,6 @@ public void indexReaderShouldSupportScan() throws Exception
...
-
-        // WHEN
-        PrimitiveLongIterator results = reader.query( IndexQuery.exists( PROP_ID1 ), IndexQuery.exists( PROP_ID2 ) );
-
-        // THEN
-        assertEquals( asSet( nodeId, nodeId2 ), PrimitiveLongCollections.toSet( results ) );
-        assertEquals( asSet( nodeId ), PrimitiveLongCollections.toSet( reader
-                .query( IndexQuery.exact( PROP_ID1, values[0] ), IndexQuery.exact( PROP_ID2, values[1] ) ) ) );
-        reader.close();
-    }
```
chrisvest: What's wrong with this test? Didn't it find a bug?

ragadeeshu: No, not as such. This test does not work. A test written in this way can never test what this test is trying to test, as this works directly with `IndexEntryUpdates`. A test for this would need to use `NodeUpdates`.

chrisvest (approved):

chrisvest:
Test looks flaky. Passed locally. Merged.

PR #8857, opened 2017-02-20
Fixed location of import dir in debian package

spacecowboy:
* Set the correct NEO4J_HOME destination

  Package installations have home located in `/var/lib/neo4j`

* Remove brittle and pointless `sed` invocations

  With the correct NEO4J_HOME set then the default config behaves as
  expected and there is no need to set these. Note that the `sed` for
  the import-entry was broken because the default config ships with an
  un-commented import entry, meaning the `sed` did nothing and ended
  up pointing to the wrong directory.


spacecowboy:
Closing, I have a better plan.

PR #8858, opened 2017-02-20
Fixed failure during init resulting in no admin

OliviaYtterbrink:
When starting database first time default roles get created and a user is made admin.
If there is a failure due to multiple users existing and no default admin is specified
subsequent database start would not try to promote user to admin, resulting in
there being no admin.

### DIFF
--- a/enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/InternalFlatFileRealmTest.java
+++ b/enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/InternalFlatFileRealmTest.java
@@ -27,30 +27,39 @@ import org.apache.shiro.cache.MemoryConstrainedCacheManager;
 import org.apache.shiro.realm.Realm;
 import org.apache.shiro.subject.PrincipalCollection;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 import java.time.Clock;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import org.neo4j.commandline.admin.security.SetDefaultAdminCommand;
+import org.neo4j.kernel.api.exceptions.InvalidArgumentsException;
 import org.neo4j.kernel.api.security.AuthenticationResult;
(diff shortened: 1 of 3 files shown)


enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/InternalFlatFileRealmTest.java
```diff
@@ -131,6 +143,174 @@ public void shouldOnlyReloadUsersOrRolesIfNeeded() throws Throwable
...
+    public void shouldNotAssignAdminToNonExistentUser() throws Throwable
+    {
+        // Given
+        InternalFlatFileRealm realm = internalTestRealmWithUsers( Collections.singletonList( "neo4j" ),
+                Collections.singletonList( "morpheus" ) );
+
+        // Expect
+        exception.expect( InvalidArgumentsException.class );
+        exception.expectMessage( "No roles defined, and default admin user 'morpheus' does not exist. Please use `" +
+                SetDefaultAdminCommand.COMMAND_NAME + "` to select a valid admin." );
```
fickludd: Now it says to use `set-default-admin`. Maybe we should also mention that this command lives in `neo4j-admin`?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/InternalFlatFileRealmTest.java
```diff
@@ -131,6 +143,174 @@ public void shouldOnlyReloadUsersOrRolesIfNeeded() throws Throwable
...
+        exception.expectMessage( "No roles defined, and cannot determine which user should be admin. Please use `" +
+                SetDefaultAdminCommand.COMMAND_NAME + "` to select an admin." );
+
+        // When
+        realm.initialize();
+        realm.start();
+    }
+
+    @Test
+    public void shouldAssignAdminMultipleDefinedUsers() throws Throwable
```
fickludd: This test name could be `shouldFailtoAssignMultipleDefaultAdmins`?

fickludd (approved):
Looks good, lot's of nice tests!

PR #8860, opened 2017-02-20
Made certificate generation atomic

spacecowboy:
It's possible that a race-condition occurs if the server is restarted
in the middle of certificate generation because it takes some time.

To prevent half-generation, a shutdown hook will delete generated
files unless the creation process gets to run to the end.


benbc:
It would be lovely to make this really atomic rather than just less likely to hit the race condition, wouldn't it? You don't have any inspiration?

spacecowboy:
No, nothing except writing another temporary file like `.certfilegeneration` at the start and removing at the end.

It's existance would indicate a startup which "failed". But I'm not sure if it's a reasonable enough approach

spacecowboy:
@benbc made the operation actually atomic

benbc:
This looks great. There is one remaining race where we write both files but don't clear the flag before getting an interrupt. But that's fine because all that will happen is that we have to regenerate the key again at start-up. Thanks Jonas.

PR #8863, opened 2017-02-21
Fixes auto index issue removing one out of many indexed keys

tinwelint:
Given e.g. a node having multiple auto-indexed keys, removing one of them
would previously remove all indexed keys for the node from the auto-index.
This fix will have only the specific key to be removed from the auto index.

cl [lucene-index] Fixes an issue where removing one key from a node or relationship which had multiple keys auto-indexed would remove all other keys for that entity from the auto-index too by mistake.


chrisvest (approved):
Good descriptions! 👍

PR #8865, opened 2017-02-21
3.1 neo4j admin help improvements

praveenag:
`neo4j-admin` help text is now displayed under different sections in the alphabetical order.

```
usage: neo4j-admin <command>

available commands:
General
    check-consistency
    help
        This help text, or help for the command specified in <command>.
    import
Authentication
    set-default-admin
    set-initial-password
Clustering
    unbind
        Removes cluster state data for the specified database.
Manage offline backup
    dump
    load
Manage online backup
    backup
        Perform an online backup from a running Neo4j enterprise server.
    restore
        Restore a backed up database.

```


community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
```diff
@@ -81,6 +81,11 @@ public String name()
         public abstract String summary();
 
         /**
+         * @return AdminCommandSection the command using the provider is grouped under
+         */
+        public abstract AdminCommandSection commandSection();
```
chrisvest: Could this by default return a "miscellaneous" section?

spacecowboy: Agreed, but the default should be equivalent to `General` section.

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommandSegment.java
```diff
@@ -0,0 +1,87 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.commandline.admin;
+
+public abstract class AdminCommandSegment
```
spacecowboy: First naming: `AdminCommandSegment` -> `AdminCommandCategory`.

Second:
My first thought here was that it should have been an `interface` and not an `abstract class`, however it has single method equivalent to a `toString` method. So now I'm wondering why it's a class at all when it could just be a String.

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
```diff
@@ -81,9 +81,9 @@ public String name()
         public abstract String summary();
 
         /**
-         * @return AdminCommandSegment the command using the provider is grouped under
+         * @return AdminCommandSection the command using the provider is grouped under
          */
-        public abstract AdminCommandSegment segment();
+        public abstract AdminCommandSection commandSection();
```
spacecowboy: I think you're making this more complex than it has to be. Why not just have this method return a straight string? And have a default implementation returning `General`?

spacecowboy (changes requested):
I think the `AdminCommandSection` class should be replaced with a simple string.

spacecowboy:
In addition, I don't think we should change the help text output in `3.1`. That's not a bug fix.
The internal refactoring however is fine as it doesn't change behavior and that might even be good to put in `3.0` to avoid merge conflicts for future bug fixes.

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
praveenag: @chrisvest I thought it might be useful to make people think about sections when they add new commands. We have a `General` section for commands that can't be grouped to a specific section.

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminCommand.java
praveenag: I initially started with a simple `String` implementation. But the idea of sections added a lot of behaviour to the simple `String` object for instance: Having to sort and map the `String` to different `AdminCommand.Provider` which makes me think that there's more to the sections than it being a simple `String`

praveenag:
changing the output is not a bug fix and wont be fixed in `3.1` but in `3.2`. I am closing this PR and splitting the refactoring work in `3.1` and the actual ordering in `3.2`

PR #8868, opened 2017-02-22
Check point safety

tinwelint:
A couple of commits which makes handling of check pointing failures more visible and safer.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
@@ -317,7 +317,7 @@ public abstract class AbstractKeyValueStore<Key> extends LifecycleAdapter
                     // won't close the state as it was before rotation began, which we're reverting to right here.
                     try ( LockWrapper ignored = writeLock( updateLock ) )
                     {
-                        state = rotation.preState();
+                        state = rotation.markAsFailed();
                     }
                     throw t;
                 }
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
@@ -322,9 +322,11 @@ public abstract class AbstractKeyValueStore<Key> extends LifecycleAdapter
                     // won't close the state as it was before rotation began, which we're reverting to right here.
                     try ( LockWrapper ignored = writeLock( updateLock ) )
                     {
-                        // Only mark as failed if we're still running, because if this store has been stopped
-                        // this rotation task will be the only option to close the preState and marking
-                        // this rotation as failed will prevent preState to be closed.
+                        // Only mark as failed if we're still running.
+                        // If shutdown has been called while being in rotation state then shutdown will fail
+                        // without closing the store. This means that rotation takes over that responsibility.
+                        // Therefore avoid marking rotation state as failed in this case and let the store
+                        // be naturally closed before leaving this method.
                         if ( !stopped )
                         {
                             state = rotation.markAsFailed();
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
@@ -55,6 +55,7 @@ public abstract class AbstractKeyValueStore<Key> extends LifecycleAdapter
     private final FileSystemAbstraction fs;
     final int keySize;
     final int valueSize;
+    private volatile boolean stopped;
 
     public AbstractKeyValueStore( FileSystemAbstraction fs, PageCache pages, File base, RotationMonitor monitor,
             RotationTimerFactory timerFactory, int keySize, int valueSize, HeaderField<?>... headerFields )
@@ -243,7 +244,11 @@ public abstract class AbstractKeyValueStore<Key> extends LifecycleAdapter
     @Override
     public final void shutdown() throws IOException
     {
-        state = state.stop();
+        try ( LockWrapper ignored = writeLock( updateLock ) )
+        {
+            stopped = true;
+            state = state.stop();
+        }
     }
 
     private boolean transfer( EntryVisitor<WritableBuffer> producer, EntryVisitor<ReadableBuffer> consumer )
@@ -317,7 +322,13 @@ public abstract class AbstractKeyValueStore<Key> extends LifecycleAdapter
                     // won't close the state as it was before rotation began, which we're reverting to right here.
                     try ( LockWrapper ignored = writeLock( updateLock ) )
                     {
-                        state = rotation.markAsFailed();
+                        // Only mark as failed if we're still running, because if this store has been stopped
+                        // this rotation task will be the only option to close the preState and marking
+                        // this rotation as failed will prevent preState to be closed.
+                        if ( !stopped )
+                        {
+                            state = rotation.markAsFailed();
+                        }
                     }
                     throw t;
                 }
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
@@ -49,7 +49,7 @@ public abstract class AbstractKeyValueStore<Key> extends LifecycleAdapter
     private final ReadWriteLock updateLock = new ReentrantReadWriteLock( /*fair=*/true );
     private final Format format;
     final RotationStrategy rotationStrategy;
-    private RotationTimerFactory rotationTimerFactory;
+    private final RotationTimerFactory rotationTimerFactory;
     volatile ProgressiveState<Key> state;
     private DataInitializer<EntryUpdater<Key>> stateInitializer;
     private final FileSystemAbstraction fs;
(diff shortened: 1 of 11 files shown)


davidegrohmann (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
```diff
@@ -300,14 +300,27 @@ private long rotate( boolean force ) throws IOException
...
+                    try ( LockWrapper ignored = writeLock( updateLock ) )
+                    {
+                        state = next;
+                    }
+                    return version;
+                }
+                catch ( Throwable t )
+                {
+                    // Rotation failed. Here we assume that rotation state remembers this so that closing it
+                    // won't close the state as it was before rotation began, which we're reverting to right here.
```
thobe, line 322: Looking at the `failed` flag in `RotationState.Rotation`, it is *not* set for all exceptions that can be thrown by the code in this catch-block, so the assumption here does not always hold.

I believe we should either:

1. Should not rely on it for this purpose, but instead introduce a specific flag for avoiding closing the `preState` (in that case the newly introduced accessor method for the `preState` could trigger this behaviour. Or,
2. Make sure that the `failed` flag is set for all exceptions ( @davidegrohmann would it be safe to set that flag for all exception cases, or is the other behaviour of that flag specific to the timeout case?), or
3. Document that we only care about this specific exception case (preferably by changing `Throwable` above to `RotationTimeoutException`), and why we only care about this case.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointScheduler.java
```diff
@@ -78,12 +103,16 @@ public boolean getAsBoolean()
             return !checkPointing;
         }
     };
+    private volatile Throwable[] failures = new Throwable[MAX_CONSECUTIVE_FAILURES_TOLERANCE];
+    private volatile int consecutiveFailures;
```
thobe: I would prefer for these fields to be declared above their usage.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointScheduler.java
```diff
@@ -78,12 +103,16 @@ public boolean getAsBoolean()
             return !checkPointing;
         }
     };
+    private volatile Throwable[] failures = new Throwable[MAX_CONSECUTIVE_FAILURES_TOLERANCE];
```
thobe: Since this field is never re-assigned, I believe it could be made `final` instead of `volatile`, we still have the volatile `consecutiveFailures` field for the barriers we need.

On the topic of the volatile `consecutiveFailures`, if this can be accessed concurrently, we should make sure that we use atomic operators for incrementing it (through `AtomicIntFieldUpdater`).

On the topic of the `failures` field never being re-assigned: who is responsible for clearing the contents of this field when there is a successful checkpoint? I don't think we want to leave old exceptions around in this array.

thobe (commented):
Overall this looks like a good step forward, but I do have some comments that I would like to see addressed or at least replied to.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
tinwelint: You're right. I like (1). How about renaming `preState` to something like `markAsFailed` and only setting that `failure` flag in there, as per your idea? I like it because we avoid having one `catch Throwable` in `RotationState` too.

EDIT: It was actually (2), yeah :)

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointScheduler.java
tinwelint: On the topic of `failures` and `consecutiveFailures` this check point job is run by the job scheduler and so can potentially be run by a different thread each time. It's not designed to be run concurrently, nor is it run concurrently in practice. I set the `failures` to `volatile` because this task may have been created after a scheduler thread set to run it in the future was created. Although you may be right that `consecutiveFailures` guards for that with its volatility. Then again, this is a task that is run once every couple of minutes so micro performance concerns aren't of uttermost importance.

On the topic of clearing `failures`, I'd have thought that leaving those exceptions there would not hurt, but you're right it's unnecessary. I'll make that change.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
thobe: You said you like option `1`, but you have implemented option `2`.

There are test failures that look related, so I'm not so sure it is safe to use the same flag for two purposes.

Before your changes this flag was used to ensure that we could still close the store even if the final rotation failed.

In your changes you want to ensure that we don't close the previous state if non-final rotation fails.

These two are quite opposing objectives, so it seems like it would be a bad idea to use the same flag (`RotationState.Rotation.failed`) for both purposes.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointScheduler.java
```diff
@@ -19,22 +19,32 @@
...
+     * The max number of consecutive check point failures that can be tolerated before treating
+     * check point failures more seriously, with a panic.
+     */
+    static final int MAX_CONSECUTIVE_FAILURES_TOLERANCE = 3;
+
     private final CheckPointer checkPointer;
     private final JobScheduler scheduler;
     private final long recurringPeriodMillis;
+    private final DatabaseHealth health;
+    private volatile Throwable[] failures = new Throwable[MAX_CONSECUTIVE_FAILURES_TOLERANCE];
```
thobe: If your reason for using `volatile` is safe publication, `final` would take care of that just as well (as long as you don't publish the `CheckPointScheduler` to other threads before the constructor has completed).

I don't think using `volatile` as a _better-safe-than-sorry_ directive is helpful, since it does not only mean something to the runtime, it also means something to the developers that will maintain this code, and then a `volatile` field that is never updated is really confusing. The fact that the type of the field is an array makes it even more confusing, especially to someone who doesn't know Java really well, since it might make you think that it is an array of volatile references to `Throwable` objects, it is not, it is a volatile reference to a single array of normal references to `Throwable` objects. The `volatile` does nothing (or at least very little) to help you ensure that the actual references in the array are safely published.

thobe: You said you were going to ensure that the contents of this array got cleared when it was no longer needed, I can't see that this is being done. This could be a reason for actually having this field `volatile` (probably the only good reason to do so): in order to be able to clear it by re-assigning a new array.

thobe (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
tinwelint: The test failure was due to shutdown not being able to close the state (since it was in rotating state when shutdown() was called. This leaves the rotation task the only guy that can close it. Made a change there so that it knows when it's left with this responsibility.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointScheduler.java
tinwelint: That array is being clear, filled with nulls. Alright, perhaps it's my misconception of final and visibility for threads created before it in this case. I'll make it final.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
```diff
@@ -317,7 +322,13 @@ private long rotate( boolean force ) throws IOException
...
                     try ( LockWrapper ignored = writeLock( updateLock ) )
                     {
-                        state = rotation.markAsFailed();
+                        // Only mark as failed if we're still running, because if this store has been stopped
+                        // this rotation task will be the only option to close the preState and marking
+                        // this rotation as failed will prevent preState to be closed.
+                        if ( !stopped )
+                        {
+                            state = rotation.markAsFailed();
+                        }
```
thobe: I'm not quite seeing how this helps, and the comment isn't quite helping me either.

It seems to me as if the store would already be in a stopped state if the `stopped` flag is `true`, since we both assigned the `stopped` flag and invoked `stop()` on the state under the same lock. Here we are also under the `updateLock`, so we should atomically see that view of the world. In that case, we should already be in the `Stopped` state, in which case rotation is not allowed, so we would not reach here.

The changes in this commit are thus pointless, and only add confusion.

thobe (changes requested):
This most recent commit adds nothing of benefit, only complexity and confusion.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/kvstore/AbstractKeyValueStore.java
tinwelint: That won't happen since stopping while in Rotation state will throw exception and not reassign the state. My first thought was to do just that too, but it turns out to not work that way.

thobe: Right. We should update the comment to say `This happens if we tried to shutdown while rotation was in progress.` as the intention of seeing that the `stopped` flag is true here is for the rotation to _take over_ the shutting down of the store.

thobe (approved):
Looks good.

PR #8870, opened 2017-02-22
Properly closes IndexHits

tinwelint:
This happened in a test, but also discovered that the IndexHits exposed to
users would not delegate close call back to internal lucene searcher,
which should currently cause problems with growing number of open files.


chrisvest:
@tinwelint This is not a problem in 3.0?

tinwelint:
Hmm, it's likely a problem as far back as 2.2. Only the method delegation part is applicable backwards. Tell you what... this can be merged and then I can fire up a new PR with the delegation fix in.

chrisvest:
@tinwelint The `DocToIdIterator` does not appear to close the `IndexHits` it is given. Is that going to be a problem?

tinwelint:
I don't think it is... it's really the searcher which it also gets passed in which is important. A bit confusing, this structure.

PR #8871, opened 2017-02-22
Fixed log rotation on Windows

spacecowboy:
Synchronizes the operations on logs to make sure that we can safely
close the file stream before performing rotation.

Previously we did not close the stream which prevented log rotation
from ever working on Windows.

### DIFF
--- a/community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
+++ b/community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
@@ -52,15 +52,15 @@ public class RotatingFileOutputStreamSupplier implements Supplier<OutputStream>,
      */
     public static class RotationListener
     {
-        public void outputFileCreated( OutputStream newStream, OutputStream oldStream )
+        public void outputFileCreated( OutputStream out )
         {
         }
 
-        public void rotationCompleted( OutputStream newStream, OutputStream oldStream )
+        public void rotationCompleted( OutputStream out )
         {
         }
 
-        public void rotationError( @SuppressWarnings("unused") Exception e, @SuppressWarnings("unused") OutputStream out )
+        public void rotationError( Exception e, OutputStream out )
         {
         }
     }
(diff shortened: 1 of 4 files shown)


maxsumrall (approved):

community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
```diff
@@ -129,10 +130,49 @@ public RotatingFileOutputStreamSupplier( FileSystemAbstraction fileSystem, File
         this.rotationListener = rotationListener;
         this.rotationExecutor = rotationExecutor;
         this.outRef.set( openOutputFile() );
+        // Wrap the actual reference to prevent race conditions during log rotation
+        this.streamWrapper = new OutputStream()
+        {
+            @Override
+            public void write( int i ) throws IOException
+            {
+                synchronized ( outRef )
```
lutovich: Now we always synchronize on `outRef`. Do we still need it to be an `AtomicReference`?

community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
```diff
@@ -190,40 +231,58 @@ private void rotate()
             return;
         }
 
-        Runnable runnable = () -> {
-            OutputStream newStream;
-            try
+        Runnable runnable = () ->
+        {
+            synchronized ( outRef )
```
lutovich: Maybe we can use `ReadWriteLock` instead of synchronized, write lock here and in `#close()` and read lock in all other places?

spacecowboy: We never read from the log files though, we only write to them. So it would be write locks everywhere.

community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
spacecowboy: Probably not, but it doesn't negatively affect anything.

lutovich: Well, it makes code more complex.

community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
lutovich: `outRef` can be read-locked for everything except rotation and closing to make sure all those operations can run concurrently. It should be write-locked for rotation and close to ensure that we only rotate or close at a point in time. It does not really matter what we do with the file, some operations just can run concurrently and some can't.

spacecowboy: Sure, I can buy that reasoning

spacecowboy:
@lutovich please have a look at the change to ReadWriteLock.

lutovich:
@spacecowboy looks good!

PR #8872, opened 2017-02-22
Make error message tests resilient to platform specific newlines

craigtaverner:


integrationtests/src/test/java/org/neo4j/procedure/StringMatcherIgnoresNewlines.java
```diff
@@ -0,0 +1,47 @@
...
+package org.neo4j.procedure;
+
+import org.hamcrest.Matcher;
+import org.hamcrest.core.StringContains;
+
+import java.util.regex.Pattern;
+
+public class StringMatcherIgnoresNewlines
+{
+    public static Matcher<String> containsStringIgnoreNewlines( String substring )
```
lutovich, line 29: We could use built-in `org.hamcrest.Matchers.equalToIgnoringWhiteSpace` matcher instead but it is even more permissive.

craigtaverner:
This PR was made against 3.2 instead of 3.1, so I made a new one at https://github.com/neo4j/neo4j/pull/8880

PR #8875, opened 2017-02-22
Distinguish kinds of settings

jjaderberg:
To generate correct settings docs we need to be able to tell the difference between kinds of settings. This allows us to do so with minimal changes to the settings themselves.

Also use the correct default unit for `DURATION` settings in `#toString`.


spacecowboy (approved):
Approved

PR #8885, opened 2017-02-23
Replacing neo4j.version with project.version for pom consistency

bradnussbaum:
Currently there is a property neo4j.version which is being assigned to project.version. This adversely effects certain maven tools such as maven-release-plugin release:update-versions which fails when trying to update dependencies because of this linked property chain.

The update here is to replace neo4j.version with project.version in dependency sections to bring poms into a working structure for the maven-release-plugin.

This command fails without these changes:
mvn release:update-versions -DautoVersionSubmodules=true -DdevelopmentVersion=3.1.3-SNAPSHOT

### DIFF
--- a/community/consistency-check/pom.xml
+++ b/community/consistency-check/pom.xml
@@ -56,12 +56,12 @@ the relevant Commercial Agreement.
     <dependency>
       <groupId>org.neo4j</groupId>
       <artifactId>neo4j-kernel</artifactId>
-      <version>${neo4j.version}</version>
+      <version>${project.version}</version>
     </dependency>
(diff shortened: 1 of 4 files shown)


spacecowboy:
Odd since that is defined as

```
<neo4j.version>${project.version}</neo4j.version>
```

Internally we use `mvn versions:set -DnewVersion=3.1.3-SNAPSHOT` and that works fine.

bradnussbaum:
Yes I should note that other tools like the codehaus version plugin have worked okay. I only had issues when using the maven-release-plugin.

PR #8891, opened 2017-02-24
extracted providers to separate files. 

praveenag:
After speaking with @spacecowboy changing the output is not a bug fix and wont be fixed in `3.1` but in `3.2`
I've finished the refactoring work on `3.1` since it will help in forward merging with 3.2 with minimum conflicts as possible.


spacecowboy (approved):
Looks good. The test of the output didn't change so that's proof that the interface won't change. Go ahead and merge when green.

spacecowboy (approved):

PR #8893, opened 2017-02-24
Add utilities for debugging through generated code

thobe:
How to use:

1. Add this trait to the test class containing your test.
2. Prefix your query with `CYPHER debug=generate_java_source`
3. When running the test make sure that the Working Directory is set to the directory of the maven module containing your test.
4. Mark `[your-maven-module]/target/generated-test-sources/cypher` as "Generated Sources Root".
5. Make sure you have a breakpoint set to somewhere before execution enters the generated code, but after the code has been generated.
6. When you run you test, as the first breakpoint triggers, find the directory `[your-maven-module]/target/generated-test-sources/cypher`, right click and select "Synchronize 'cypher'"
  _(If you have not done so before, this is a good time to "Mark Directory as" "Generated Sources Root".)_
7. Now you should see the source file for the generated query, and be able to set breakpoints in that code, as well as stepping through it.
8. Note that every time you re-run your test, you will have to repeat steps 5 to 7, since new code will be generated each time.

### DIFF
--- a/enterprise/cypher/cypher-compiled-runtime-3.2/src/test/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/SaveGeneratedSource.scala
+++ b/enterprise/cypher/cypher-compiled-runtime-3.2/src/test/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/SaveGeneratedSource.scala
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher.internal.compiled_runtime.v3_2
+
+import java.nio.file.FileVisitResult.CONTINUE
+import java.nio.file._
+import java.nio.file.attribute.BasicFileAttributes
+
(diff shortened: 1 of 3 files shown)


enterprise/cypher/cypher-compiled-runtime-3.2/src/test/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/SaveGeneratedSource.scala
```diff
@@ -0,0 +1,86 @@
...
+    // If CWD is set up correctly, we assign the generated source location
+    if(Files.isRegularFile(cwd.resolve("src/test/scala").resolve(getClass.getName.replace('.', '/') + ".scala"))
+      && Files.isDirectory(cwd.resolve("target"))) {
+      setLocation(cwd.resolve("target").resolve("generated-test-sources").resolve("cypher"))
+    }
+  }
+
+
+  private def setLocation(location: Path) = {
+    System.err.println(s"Will save generated sources to $location")
```
davidegrohmann, line 64: leftover?

thobe: No, it's actually there on purpose, and for two reasons:

1. To remind you of where the source files are saved to - to help with steps 4-6, and
2. To allow you to verify that you've set up your Working Directory property, if you see that printout - you know that you did it right.

thobe: Also, I don't expect any usage of the `SaveGeneratedSources` trait to be left in actual code, I imagine it being something you add while you are debugging, then you remove it when you are done. So if we ever were to see this from the build system, we know that someone left the debug mode on...

davidegrohmann: perhaps we can add a scalacheck rule to make sure we don't forget usages of `SaveGeneratedSources`?

thobe: @davidegrohmann if you know how to do that and think it would be valuable, I think that sounds like a great idea.

I've tried to make `SaveGeneratedSources` as non-invasive as possible:

* It is only in effect for the duration of each test, and will not affect other tests.
* It will only affect queries marked with `debug=generate_java_source` (which we shouldn't leave in the code in most places either)
* I'm not sure which Working Directory maven runs tests with, so quite possibly `SaveGeneratedSources` will do nothing during a standard maven run.

PR #8902, opened 2017-02-27
Config support for legacy quote escaping in LOAD CSV

craigtaverner:
Added a new config setting `dbms.import.csv.legacy_quote_escaping` to select whether to conform to the standard https://tools.ietf.org/html/rfc4180 for interpreting escaped quotation characters in CSV files loaded using `LOAD CSV`. Setting this to `false` will use the standard, interpreting repeated quotes '\"\"' as a single in-lined quote, while `true` will use the legacy convention originally supported in Neo4j 3.0 and 3.1, allowing a backslash to include quotes in-lined in fields.


craigtaverner:
Closed because it was opened against the wrong branch

PR #8903, opened 2017-02-27
Config support for legacy quote escaping in LOAD CSV

craigtaverner:
changelog: Added a new config setting `dbms.import.csv.legacy_quote_escaping` to select whether to conform to the standard https://tools.ietf.org/html/rfc4180 for interpreting escaped quotation characters in CSV files loaded using `LOAD CSV`. Setting this to `false` will use the standard, interpreting repeated quotes '\"\"' as a single in-lined quote, while `true` will use the legacy convention originally supported in Neo4j 3.0 and 3.1, allowing a backslash to include quotes in-lined in fields.

### DIFF
--- a/community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/logical/LogicalPlanningContext.scala
+++ b/community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/logical/LogicalPlanningContext.scala
@@ -19,6 +19,7 @@
  */
 package org.neo4j.cypher.internal.compiler.v3_0.planner.logical
 
+import org.neo4j.csv.reader.Configuration.DEFAULT_LEGACY_STYLE_QUOTING
 import org.neo4j.cypher.internal.compiler.v3_0.planner.logical.Metrics.QueryGraphSolverInput
 import org.neo4j.cypher.internal.compiler.v3_0.planner.logical.plans.{IdName, LogicalPlan, StrictnessMode}
 import org.neo4j.cypher.internal.compiler.v3_0.planner.logical.steps.LogicalPlanProducer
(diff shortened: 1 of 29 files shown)


community/cypher/cypher-compiler-3.0/src/main/scala/org/neo4j/cypher/internal/compiler/v3_0/planner/logical/LogicalPlanningContext.scala
```diff
@@ -37,6 +37,7 @@ case class LogicalPlanningContext(planContext: PlanContext,
                                   useErrorsOverWarnings: Boolean = false,
                                   errorIfShortestPathFallbackUsedAtRuntime: Boolean = false,
                                   errorIfShortestPathHasCommonNodesAtRuntime: Boolean = true,
+                                  legacyCsvQuoteEscaping: Boolean = true,
```
tinwelint: There are a couple of places where there's a default value specified here. Could that be collected so that when if flips, which it should do at some point, it can be flipped in one place?

tinwelint (changes requested):

tinwelint (approved):
Grand!

xiatian90s:
Hi, I am using the Neo4j Windows desktop version 3.1.4 and I tried to set the dbms.import.csv.legacy_quote_escaping to false. However, I can not find this variable in my C:\Users\txia\AppData\Roaming\Neo4j Community Edition\neo4j.conf.

Here is what my neo4j.conf looks like(to attach it, I change it to .txt):
[neo4j.txt](https://github.com/neo4j/neo4j/files/1005568/neo4j.txt)

Any idea? Thanks!

craigtaverner:
@xiatian90s you should be able to just add it as a line to the end of your neo4j.conf file.

    dbms.import.csv.legacy_quote_escaping=false

aamanlamba:
Hi, what is the default value for dbms.import.csv.legacy_quote_escaping if it is not specified in the config file?

tinwelint:
It's true by default, but may be flipped to false as default in a later version

upidea:
This is a bad solution. I think don't ask data to change for data is always big.  
http://grokbase.com/t/gg/neo4j/14b230yqm3/csv-import-problems-w-double-quotes

*Proposed Solution:*
Include a *TEXTQUALIFIER* *true/false *option so that the default rule for
assuming double quotes around values can be disabled more easily and also
assume that content between the FIELDTERMINIATOR are qualified by the
FIELDTERMINIATOR that is specified by the user in the header of the CSV
import cypher.

PR #8904, opened 2017-02-27
`neo4j-admin` help now has better formatting

praveenag:
`neo4j-admin help` now has alphabetised sections.

```
usage: neo4j-admin <command>

Manage your Neo4j instance.

environment variables:

available commands:
General
    check-consistency
    import
    version
        Check the version of a Neo4j database store.
Authentication
    set-default-admin
    set-initial-password
Clustering
    unbind
        Removes cluster state data for the specified database.
Manage offline backup
    dump
    load
Manage online backup
    backup
        Perform an online backup from a running Neo4j enterprise server.
    restore
        Restore a backed up database.

```


community/command-line/src/main/java/org/neo4j/commandline/admin/CommandUsage.java
```diff
@@ -0,0 +1,52 @@
...
+    {
+        this.command = command;
+        this.scriptName = scriptName;
+    }
+
+    void printDetailed( Consumer<String> output )
+    {
+        for ( Arguments arguments : command.possibleArguments() )
+        {
+            //Arguments arguments = command.arguments();
```
spacecowboy: Left over comment

community/command-line/src/main/java/org/neo4j/commandline/admin/Usage.java
```diff
@@ -40,75 +39,53 @@ public Usage( String scriptName, CommandLocator commands )
...
-        {
-            output.accept( format( "%s", command.name() ) );
-            output.accept( "    " + command.summary() );
-        }
-
-        public void printIndentedSummary( Consumer<String> output )
-        {
-            printSummary( s -> output.accept( "    " + s ) );
-        }
+        Map<AdminCommandSection,List<AdminCommand.Provider>> groupedProviders = groupProvidersBySegment();
```
spacecowboy: There seem to be a mix of terminology. Could we keep it to `Section` everywhere?

community/command-line/src/main/java/org/neo4j/commandline/admin/Usage.java
```diff
@@ -40,75 +39,53 @@ public Usage( String scriptName, CommandLocator commands )
...
-                output.accept( Arguments.rightColumnFormatted( left, arguments.usage(), left.length() + 1 ) );
-            }
-            output.accept( "" );
-            output.accept( command.allArguments().description( command.description() ) );
-        }
+    private Map<AdminCommandSection,List<AdminCommand.Provider>> groupProvidersBySegment()
+    {
+        List<AdminCommand.Provider> providers = new ArrayList<>();
+        commands.getAllProviders().forEach( providers::add );
+        return providers.stream().collect( Collectors.groupingBy( AdminCommand.Provider::commandSection ) );
```
spacecowboy: This introduces an unnecessary possibility for duplication.

By grouping on `AdminCommandSection` objects, two sections with the same names will be listed twice.

For commands bundled in Neo4j you can (and should) point to the constant sections such as `AddminCommandSection.GENERAL` but for external commands which define their own sections, this is not possible.

As an example, consider the capiflash commands. They will have to define their own `AdminCommandSection` object.

Now it becomes impossible for any other command to be listed under the capi flash section unless it shares source with the other capiflash commands.

If instead the code does this then this issue doesn't exist:

```
return providers.stream().collect( Collectors.groupingBy( (p) -> p.commandSection().printable() ) );
```

But it would be much cleaner if `AdminCommandProvider.commandSection()` simply returned a String which would also solve it.

community/dbms/src/main/java/org/neo4j/commandline/dbms/ManageOffineBackupCommandSection.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public static AdminCommandSection instance()
+    {
+        return manageOffineBackupCommandSection;
+    }
+
+    @Override
+    public String printable()
+    {
+        return "Manage offline backup";
```
spacecowboy: I think this would be better as just `Offline backup`

enterprise/backup/src/main/java/org/neo4j/ManageOnlineBackupCommandSection.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public static AdminCommandSection instance()
+    {
+        return manageOnlineBackupCommandSection;
+    }
+
+    @Override
+    public String printable()
+    {
+        return "Manage online backup";
```
spacecowboy: This would be better as just `Online backup`

spacecowboy (changes requested):

community/command-line/src/main/java/org/neo4j/commandline/admin/Usage.java
praveenag: > Now it becomes impossible for any other command to be listed under the capi flash section unless it shares source with the other capiflash commands.

Shouldn't this be expected of them anyway if you are developing stuff related to capi flash it should depend on capi flash module thereby inherit the CapiFlashCommandSection that comes along with it. 

> But it would be much cleaner if AdminCommandProvider.commandSection() simply returned a String which would also solve it.

`AdminCommandSection` is much more than a simple `String` Object. It guides the means by which the `AdminCommand.Provider` are grouped and displayed. Yes they can be expressed by a `String` but we would leak `printAllCommandsUnderSection` implementation outside of the `String` object.

community/dbms/src/main/java/org/neo4j/commandline/dbms/ManageOffineBackupCommandSection.java
praveenag: Do you want to change `Manage Online backup` to `Online backup` too?

spacecowboy (approved):

PR #8905, opened 2017-02-28
Suppress output from test

burqen:


martinfurmanski:
@burqen Please don't. It is surprising behaviour to suppress system outs since that is quite commonly used for debugging and hard to track down in general when they go missing. Let me offer you an alternative: https://github.com/neo4j/neo4j/pull/8911

burqen:
@martinfurmanski Looks better! Closing

PR #8906, opened 2017-02-28
Make IndexEntryConflictException handle composite indexes

fickludd:
Make IndexEntryConflictException handle composite indexes. Also introduces new class OrderedPropertyValues for managing a set of property values that maps to one schema descriptor.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
@@ -19,9 +19,6 @@
  */
 package org.neo4j.kernel.api.exceptions.index;
 
-import java.util.Arrays;
-
-import org.neo4j.helpers.Strings;
 import org.neo4j.kernel.api.TokenNameLookup;
 import org.neo4j.kernel.api.exceptions.KernelException;
 import org.neo4j.kernel.api.schema_new.LabelSchemaDescriptor;
@@ -30,7 +27,6 @@ import org.neo4j.kernel.api.schema_new.SchemaUtil;
 import org.neo4j.kernel.api.schema_new.index.NewIndexDescriptor;
 
 import static java.lang.String.format;
-import static java.lang.String.valueOf;
 import static org.neo4j.kernel.api.StatementConstants.NO_SUCH_NODE;
 
 /**
@@ -42,10 +38,15 @@ public class IndexEntryConflictException extends Exception
     private final long addedNodeId;
     private final long existingNodeId;
 
+    public IndexEntryConflictException( long existingNodeId, long addedNodeId, Object propertyValue )
+    {
+        this( existingNodeId, addedNodeId, OrderedPropertyValues.of( propertyValue ) );
+    }
+
     public IndexEntryConflictException( long existingNodeId, long addedNodeId, OrderedPropertyValues propertyValues )
     {
         super( format( "Both node %d and node %d share the property value %s",
-                existingNodeId, addedNodeId, quote( propertyValues ) ) );
+                existingNodeId, addedNodeId, propertyValues ) );
         this.existingNodeId = existingNodeId;
         this.addedNodeId = addedNodeId;
         this.propertyValues = propertyValues;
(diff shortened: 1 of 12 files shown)


community/kernel/src/test/java/org/neo4j/kernel/api/index/UniqueIndexPopulatorCompatibility.java
```diff
@@ -81,7 +81,7 @@ public void shouldProvidePopulatorThatEnforcesUniqueConstraints() throws Excepti
         catch ( IndexEntryConflictException conflict )
         {
             assertEquals( nodeId1, conflict.getExistingNodeId() );
-            assertEquals( value, conflict.getPropertyValue() );
+            assertEquals( new OrderedPropertyValues( value ), conflict.getPropertyValues() );
```
MishaDemianenko: Would be super nice to have set of `OrderedPropertyValues.of` static factory methods maybe

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
```diff
@@ -61,23 +63,32 @@ public RuntimeException notAllowed( NewIndexDescriptor descriptor )
                 descriptor.userDescription( SchemaUtil.idTokenNameLookup ) ), this );
     }
 
-    public String evidenceMessage( String labelName, String propertyKey )
+    public String evidenceMessage( TokenNameLookup tokenNameLookup, LabelSchemaDescriptor schema )
     {
+        assert schema.getPropertyIds().length == propertyValues.values().length;
```
MishaDemianenko: since we already have a specific value object for property values i think it's super strange that we fall back to backing array to get that would propose to have propertyValues.size() method instead

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
```diff
@@ -61,23 +63,32 @@ public RuntimeException notAllowed( NewIndexDescriptor descriptor )
...
+    public OrderedPropertyValues getPropertyValues()
+    {
+        return propertyValues;
+    }
+
+    public Object getSinglePropertyValue()
     {
-        return propertyValue;
+        assert propertyValues.values().length == 1;
+        return propertyValues.values()[0];
```
MishaDemianenko: also here why not provide method to get single method in helper class

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
```diff
@@ -122,13 +133,43 @@ public int hashCode()
...
+            sep = ", ";
+            sb.append( '`' );
+            sb.append( tokenNameLookup.propertyKeyGetName( propertyIds[i] ) );
+            sb.append( "` = " );
+            sb.append( quote( propertyValues.values()[i] ) );
+        }
+        return sb.toString();
+    }
+
+    private static String quote( OrderedPropertyValues propertyValues )
```
MishaDemianenko: make it OrderedPropertyValues.toString() instead maybe?

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/OrderedPropertyValues.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api.schema_new;
+
+import java.util.Arrays;
+
+/**
+ * Holder for n property values, ordered according to a schema descriptor property id order
```
MishaDemianenko: can't find where this ordering actually enforced, can you please help me to find that place

MishaDemianenko (changes requested):
Couple of minor comments, fine in general

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/OrderedPropertyValues.java
fickludd: The property value order in not enforced yet in this PR, and the places that throw IndexEntryConflictExceptions (`UniqueInMemoryIndex`, `DuplicateCheckingCollector` and `ConstraintEnforcingEntityOperations`) do not yet support composite properties. We broke this exception part out into a separate PR from need in implementing these parts. Enforcement of the OrderedPropertyValues invariant will be performed by callers in that work.

MishaDemianenko: ok

MishaDemianenko (approved):

MishaDemianenko:
@fickludd will merge on green

PR #8908, opened 2017-02-28
Server tags etc

martinfurmanski:
changelog: Support tagging of servers and display it using the cluster overview procedure.

### DIFF
--- a/enterprise/neo4j-harness-enterprise/src/main/java/org/neo4j/harness/CausalClusterInProcessRunner.java
+++ b/enterprise/neo4j-harness-enterprise/src/main/java/org/neo4j/harness/CausalClusterInProcessRunner.java
@@ -1,0 +1,229 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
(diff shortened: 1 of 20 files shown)


enterprise/neo4j-harness-enterprise/src/main/java/org/neo4j/harness/CausalClusterInProcessRunner.java
```diff
@@ -0,0 +1,229 @@
...
+                replicaThread.start();
+            }
+
+            for ( Thread replicaThread : replicaThreads )
+            {
+                replicaThread.join();
+            }
+        }
+
+        private static String specifyPortOnly( int port )
```
jimwebber, line 189: Seems to reduce readability. Inline?

martinfurmanski: @jimwebber Well, I specifically factored it out to increase understandability. The syntax of :port is not really commonplace I believe and would look something like ":" + port inline, so I found this better. This at least tells you something about what that syntax means. It would of course better if this was a supported setting type with helper functions.

PR #8909, opened 2017-02-28
Faster refresh of read replica topology

martinfurmanski:
The default was 1 minute for the product and this can be a surprisingly long
time. There should be no harm of lowering this to 5 seconds.

For tests the default was lowered to 1 second to speed up tests
relying on this timing. An extra 1 minute is unnecessarily long.

changelog: Decrease default topology refresh timeout from 1 minute to 5 seconds.


mneedham:
I'll merge this one in after I've done this one which reduces the amount of logging around replicas: https://github.com/neo4j/neo4j/pull/9024

PR #8915, opened 2017-03-01
Added functions sandboxing

eebus:
changelog: As a extension of [#8753](https://github.com/neo4j/neo4j/pull/8753) we have now added sandboxing to user defined functions and user defined aggregation functions.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/proc/LoadFailFunction.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/proc/LoadFailFunction.java
@@ -19,36 +19,21 @@
  */
 package org.neo4j.kernel.api.proc;
 
+import org.neo4j.collection.RawIterator;
 import org.neo4j.kernel.api.exceptions.ProcedureException;
+import org.neo4j.kernel.api.exceptions.Status;
 
-public interface CallableUserAggregationFunction
+public class LoadFailFunction extends CallableUserFunction.BasicUserFunction
 {
-    UserFunctionSignature signature();
-    Aggregator create( Context ctx ) throws ProcedureException;
-
-    interface Aggregator
+    public LoadFailFunction( UserFunctionSignature signature )
     {
-        void update( Object[] input ) throws ProcedureException;
-
-        Object result() throws ProcedureException;
+        super( signature );
     }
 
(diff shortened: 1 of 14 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/proc/ReflectiveProcedureCompiler.java
```diff
@@ -376,19 +392,39 @@ private ReflectiveUserAggregationFunction compileAggregationFunction( Class<?> d
...
+        if ( !config.fullAccessFor( funcName.toString() ) )
+        {
+            try
+            {
+                setters = safeFieldInjections.setters( definition );
+            }
+            catch ( ComponentInjectionException e )
+            {
+                description = Optional.of( funcName.toString() +
+                        " is not available due to not having unrestricted access rights, check configuration." );
```
fickludd: how about avoiding double-negation here, and go for `is not available due to having restricted access rights, check configuration.`

community/kernel/src/main/java/org/neo4j/kernel/api/proc/LoadFailFunction.java
```diff
@@ -0,0 +1,39 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api.proc;
+
+import org.neo4j.collection.RawIterator;
+import org.neo4j.kernel.api.exceptions.ProcedureException;
+import org.neo4j.kernel.api.exceptions.Status;
+
+public class LoadFailFunction extends CallableUserFunction.BasicUserFunction
```
fickludd: Not completely convinced by this name... ´FailedLoadFunctionMock´ or something?

community/kernel/src/test/java/org/neo4j/kernel/impl/proc/ProcedureJarLoaderTest.java
```diff
@@ -19,6 +19,7 @@
  */
 package org.neo4j.kernel.impl.proc;
 
+import jdk.nashorn.internal.codegen.FunctionSignature;
```
fickludd: import miss?

fickludd (approved):
Good test coverage, some optional comments

PR #8916, opened 2017-03-01
`neo4j start` once again prints correct address on startup

spacecowboy:
This fixes the previous generic message and prints where the server is
actually listening.


benbc (approved):
Brilliant. What a relief.

PR #8917, opened 2017-03-01
Exclude replica from topology if it doesn't have a UUID

mneedham:


mneedham:
I think I have a cleaner way of fixing this. Will close this for now.

PR #8918, opened 2017-03-01
Introduce page cache hit metric

MishaDemianenko:
Introduce hit metric that will show count of how many requested pages are
retrieved from cache.
Default implementation count hits only if page where retrieved from page cache itself
without page faults. In case if there is already page fault in progress
we will gonna count this page as hit.


chrisvest:
Couldn't hits be computed as `pins - faults = hits`?

MishaDemianenko:
@chrisvest since we can have a failures during page faults not really;
 also it can be quite strange to observe number of hits that decrease next time when you check just because page fault happen after pin started.

chrisvest:
@MishaDemianenko Good point. I think this is good to merge once green.

PR #8920, opened 2017-03-01
Added user-defined function whitelisting.

eebus:
changelog: Added support for whitelisting user-defined function. This allows a user to load only a specific user-defined function from a package of functions.


craigtaverner (changes requested):
I did not see tests for wildcards and tests that show regex characters are blocked.

craigtaverner (approved):

PR #8921, opened 2017-03-01
Postpone logging of log rotation progress until release logFileLock

burqen:
Fixing deadlock problem between flushing log (T1) and rotating log (T2).

Problem is the order in which flushing and rotating grab locks.
Flush first grab PrintWriter lock, then log file guard lock.
Rotate first grab log file guard lock then PrintWriter lock.

Solution is to not log rotation progress while holding the log file guard lock
and instead buffer the log messages until lock is released.

Also add a verbose timeout to CoreReplicationIT which suffered from this deadlock.

co-author: @MishaDemianenko

### DIFF
--- a/community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
+++ b/community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
@@ -19,6 +19,7 @@
  */
 package org.neo4j.logging;
 
+import java.io.ByteArrayOutputStream;
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
@@ -243,7 +244,7 @@ public class RotatingFileOutputStreamSupplier implements Supplier<OutputStream>,
         return earliestRotationTimeRef.get() <= currentTimeSupplier.getAsLong();
     }
 
-    private void rotate()
+    void rotate()
     {
         if ( rotating.getAndSet( true ) )
         {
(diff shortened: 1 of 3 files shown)


spacecowboy (approved):
Seems reasonable. I have triggered a Windows build to make sure that also passes.

community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
```diff
@@ -292,26 +294,34 @@ private void rotate()
...
                             outRef = openOutputFile();
-                            rotationListener.outputFileCreated( streamWrapper );
+                            rotationListener.outputFileCreated( bufferingOutputStream );
                         }
                     }
                     catch ( IOException e )
                     {
                         System.err.println( "Failed to open log file after log rotation: " + e.getMessage() );
-                        rotationListener.rotationError( e, streamWrapper );
+                        rotationListener.rotationError( e, bufferingOutputStream );
```
chrisvest, line 303: Not a fan of that `System.err.println` – it implies a lock on stderr, an we are already capturing the error in the rotation listener.

chrisvest (changes requested):
One comment

community/logging/src/main/java/org/neo4j/logging/RotatingFileOutputStreamSupplier.java
spacecowboy: That print was put there in case no log file could be opened, in which case the log will be sent to the `nullStream`. The user would have no indication that something is wrong.

chrisvest:
@burqen 
> Flush first grab PrintWriter lock, then log file guard lock.

Where is this flush?

chrisvest (approved):

chrisvest:
@spacecowboy Do you know anything about that PowerShell test failure? It looks unrelated.

spacecowboy:
@chrisvest it is and would be fixed by a rebase I believe so it can be safely ignored here.

PR #8922, opened 2017-03-01
added @NonNull annotations for command sections.

praveenag:


spacecowboy:
LGTM. Do you think we should also add `NonNull` / `Nullable` annotations to the rest of the methods in that class?

spacecowboy (approved):

PR #8924, opened 2017-03-01
Log warnings about deprecated config settings

spacecowboy:
Output with a deprecated config value added to the config:

```
$ bin/neo4j console
Starting Neo4j.
WARNING: Max 1024 open files allowed, minimum of 40000 recommended. See the Neo4j manual.
2017-03-02 10:55:25.140+0000 WARN  WARNING! Deprecated configuration options used. See manual for details
2017-03-02 10:55:25.141+0000 WARN  dbms.index_sampling.buffer_size has been replaced with dbms.index_sampling.sample_size_limit.
2017-03-02 10:55:25.381+0000 INFO  Starting...
2017-03-02 10:55:26.131+0000 INFO  Bolt enabled on localhost:7687.
2017-03-02 10:55:26.145+0000 INFO  Initiating metrics...
2017-03-02 10:55:28.673+0000 INFO  Started.
2017-03-02 10:55:28.988+0000 INFO  Mounted REST API at: /db/manage
2017-03-02 10:55:29.773+0000 INFO  Remote interface available at http://localhost:7474/
```

Edit: updated example output

### DIFF
--- a/community/configuration/src/main/java/org/neo4j/configuration/ReplacedBy.java
+++ b/community/configuration/src/main/java/org/neo4j/configuration/ReplacedBy.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.neo4j.configuration;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used to indicate what setting have/will replace a deprecated setting.
+ */
+@Retention( RetentionPolicy.RUNTIME )
+@Target( {ElementType.TYPE, ElementType.FIELD} )
+public @interface ReplacedBy
+{
+    String value();
+}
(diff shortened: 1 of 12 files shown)


chrisvest:
@spacecowboy Why is `unsupported.dbms.memory.pagecache.pagesize` being warned about? It's not in any of our default config files as far as I can see.

chrisvest:
Also, `@Internal` settings are ignored by the manual, so if they are in our distribution `neo4j.conf` files, then they are probably not really internal. So if we had separated internal and non-internal settings properly, then I don't think we should warn about any deprecation of internal settings, but we can discuss if we should warn about internal settings being used in the first place.

spacecowboy:
@chrisvest they are warned about because we insert them into the config file at Runtime during the first config file parsing. We probably shouldn't warn about internal stuff but we could also stop using internal things which are deprecated.

chrisvest:
@spacecowboy We are not using `unsupported.dbms.memory.pagecache.pagesize` other than to print a separate warning if it is set to anything but its default value. Hence why I'm puzzled about it showing up.

spacecowboy:
@chrisvest ah. the mistake is in the warning code. It goes through all settings, defined or not. Will fix

spacecowboy:
@jjaderberg how do you feel about the deprication API here from a docs standpoint?

jjaderberg:
LGTM, I think it gives us what we need for deprecations. I don't see a way to filter internal/unsupported settings however, so we can't quite use this yet.

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -416,14 +416,36 @@ private synchronized void replaceSettings( Map<String,String> newSettings, boole
...
         for ( ConfigurationValidator validator : validators )
         {
             validSettings = validator.validate( settingValidators, validSettings, log, configFile.isPresent() );
         }
+
         params.clear();
         params.putAll( validSettings );
+
+        // Warn about deprecations
+        if ( configFile.isPresent() ) {
```
praveenag, line 431: Few suggestions here,
```
    configFile.ifPresent( file -> warnAboutDeprecations() );
    }

    private void warnAboutDeprecations()
    {
        configOptions.stream()
                .flatMap( it -> it.asConfigValues( params ).stream() )
                .filter( config -> params.containsKey( config.name() ) && config.deprecated() )
                .forEach(
                        config ->
                            {
                                if ( config.replacement().isPresent() )
                                {
                                    log.warn( "%s is deprecated. Replaced by %s", config.name(), config.replacement().get() );
                                }
                                else
                                {
                                    log.warn( "%s is deprecated.", config.name() );
                                }
                            }
                );
    }
```

community/configuration/src/main/java/org/neo4j/configuration/ReplacedBy.java
```diff
@@ -0,0 +1,36 @@
...
+package org.neo4j.configuration;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used to indicate what setting have/will replace a deprecated setting.
+ */
```
praveenag, line 30: Loved this

PR #8926, opened 2017-03-02
Fix racy ProcedureIT.shouldBeAbleToSpawnThreadsCreatingTransactionInProcedures

chrisvest:
Though we join the threads that calls the `supportedProcedure`, that procedure itself did not wait for the tasks _it_ spawned to complete
Therefor the test could observe a node count that was less than the tasks it had started

This is no longer possible, because the procedure now blocks until the tasks it submits have finished

@davidegrohmann @OliviaYtterbrink @boggle Can one of you take a look at this?


OliviaYtterbrink (approved):

boggle:
Seems reasonable to me.

PR #7558, opened 2016-07-12
Introduce new parameter syntax

boggle:
Note: The old syntax is not going away for a long time

The link to the CIP for this change is [here](https://github.com/opencypher/openCypher/blob/master/cip/CIP2016-07-07-Parameter-syntax.adoc)

changelog: Introduce new syntax for parameters in Cypher ($param and $`param`)

PR #8931, opened 2017-03-02
Support new parameter syntax on 3.0

Mats-SX:
This PR is a backport of #7558

changelog: Cypher now supports an updated parameter syntax `$param` and it is no longer possible to start symbolic names (variables, property keys, parameter names) with a currency symbol


Mats-SX:
This should be null-forward merged.

technige (approved):
LGTM :+1:

PR #8934, opened 2017-03-02
dbms.listQueries would now report requestScheme as protocol.

praveenag:


praveenag:
Merging after getting approval from @thobe

PR #8935, opened 2017-03-02
Do not start multiple file watchers for core state

MishaDemianenko:
Prevent starting of multiple file watchers from for LocalDatabase that can be started multiple times,
add additional assertion that will prevent multiple watcher threads in future.


MishaDemianenko:
@martinfurmanski can you please take a look? thx in advance!

burqen (approved):
Looks good!

martinfurmanski (approved):

PR #8942, opened 2017-03-03
Revert "Doesn't build label scan store as part of import"

tinwelint:
This reverts commit 5bbc486aee3ff73ae78da90eae6c28648088803f.

Which means that label scan store will be built as part of neo4j-import again.
This is possible since page size is fixed now.


burqen (approved):
Looks good!

PR #8944, opened 2017-03-03
Fixes issue with tx event handler data for deleted relationships

tinwelint:
Where delted relationships wouldn't have their RelationshipProxy instances
that were returned to the user in TransactionEventHandler#afterCommit
filled with data about the relationship. This would have accessor methods like
Relationship#toString fail reporting to not being a transaction when it wanted
to load that data.


chrisvest (approved):
Looks good. Just waiting for a green build.

tinwelint:
@chrisvest the wait is over

PR #8947, opened 2017-03-06
GBPTree checkpoint/close can write header data

tinwelint:
Which is stored in the selected state page, after the other state information.
Checkpoint/close methods come in two variants, one with header writer and one w/o.
The one with header writer will replace the header data with new data,
whereas the variants w/o will keep/carry over the previous header data to
the new state page.

When opening a GBPTree the (newly introduced) supplied header reader can access
the header data.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
@@ -446,34 +446,39 @@ public class GBPTree<KEY,VALUE> implements Closeable
                     freeList.lastId(), freeList.writePageId(), freeList.readPageId(),
                     freeList.writePos(), freeList.readPos() );
 
-            // Write/carry over header
-            int headerOffset = cursor.getOffset();
-            int headerDataOffset = headerOffset + Integer.BYTES; // will contain length of written header data (below)
-            TreeState otherState = other( states, oldestState );
-            if ( otherState.isValid() )
-            {
-                PageCursor previousCursor = pagedFile.io( otherState.pageId(), PagedFile.PF_SHARED_READ_LOCK );
-                PageCursorUtil.goTo( previousCursor, "previous state page", otherState.pageId() );
-                do
-                {
-                    // Place the previous state cursor after state data
-                    TreeState.read( previousCursor );
-                    // Read length of previous header
-                    int previousLength = previousCursor.getInt();
-                    // Reserve space to store length
-                    cursor.setOffset( headerDataOffset );
-                    // Write
-                    headerWriter.write( previousCursor, previousLength, cursor );
-                }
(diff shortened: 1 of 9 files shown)


community/index/src/main/java/org/neo4j/index/internal/gbptree/Header.java
```diff
@@ -0,0 +1,79 @@
...
+         * bytes in size.
+         *
+         * @param from {@link PageCursor} positioned at the header data written in the previous check point.
+         * @param length size in bytes of the previous header data.
+         * @param to {@link PageCursor} to write new header into.
+         */
+        void write( PageCursor from, int length, PageCursor to );
+    }
+
+    final Consumer<PageCursor> CARRY_OVER = cursor -> {};
```
burqen: This guy is never used

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -424,10 +445,43 @@ private void writeState( PagedFile pagedFile ) throws IOException
                     root.id(), root.generation(),
                     freeList.lastId(), freeList.writePageId(), freeList.readPageId(),
                     freeList.writePos(), freeList.readPos() );
+
+            // Write/carry over header
```
burqen: Nice! This can easily be extracted into a helper method for improved readability.

community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
```diff
@@ -436,6 +449,66 @@ public void shouldAllowClosingWriterMultipleTimes() throws Exception
         // THEN that should be OK
     }
 
+    @Test
```
burqen, line 452: Is it in the contract that close should always perform a checkpoint? Otherwise we should also test close with header here, if implementation change in future.

burqen (changes requested):
Some small cleanup comments. Looks good otherwise.

community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
tinwelint: I think it's by contract, at least the javadoc says so. I can add a test regardless.

tinwelint: Actually, the three added tests does test both checkpoint and close, don't they?

burqen: Yes, now that I look a little closer 👍

burqen (approved):

burqen:
@tinwelint Will merge once green

PR #8950, opened 2017-03-06
Enhance set of checks in kernel transactions for starting new transaction.

MishaDemianenko:
Enhance range of checks in kernel transaction to cover scenario when kernel transactions component is stopped.
This changes will allow us to avoid blocking of threads that try to start transaction during:
- store copy;
- different sorts of instance restarts when new kernel transactions component created.


MishaDemianenko:
@lutovich can you also take a look on this

community/kernel/src/main/java/org/neo4j/kernel/impl/api/DatabaseUnavailableException.java
```diff
@@ -0,0 +1,29 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api;
+
+class DatabaseUnavailableException extends RuntimeException
+{
+    DatabaseUnavailableException()
+    {
+        super( "This database is unavailable." );
+    }
```
chrisvest: What do you think about moving this to the `org.neo4j.graphdb` directory and make it extend `TransientDatabaseFailureException`? This would let existing code (with thorough catch clauses) handle it correctly.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
```diff
@@ -219,13 +226,15 @@ public boolean haveClosingTransaction()
...
     {
+        stopped = false;
         unblockNewTransactions();
     }
 
     @Override
     public void stop() throws Throwable
     {
         blockNewTransactions();
+        stopped = true;
```
chrisvest, line 237: Why is this volatile necessary when we already have the `newTransactionsLock`? This lock, by the way, could be changed to a `StampedLock` – it's not reentrant, so there would not be a need for the `assertCurrentThreadIsNotBlockingNewTransactions()` check.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionsTest.java
```diff
@@ -413,12 +414,62 @@ public void exceptionWhenStartingNewTransactionOnShutdownInstance() throws Throw
...
         availabilityGuard.shutdown();
-        Executors.newSingleThreadExecutor().submit( () -> stopKernelTransactions( kernelTransactions ) ).get();
+        t2.execute( (OtherThreadExecutor.WorkerCommand<Void,Void>) state ->
+        {
+            stopKernelTransactions( kernelTransactions );
+            return null;
+        } );
 
         expectedException.expect( DatabaseShutdownException.class );
         kernelTransactions.newInstance( KernelTransaction.Type.explicit, securityContext, 0L );
```
chrisvest: Won't this exception now be racing with the worker thread started above, now that you removed the `get` on the future?

chrisvest: Similar thing with the next two tests.

chrisvest (commented):
Looks okay, but I have a few suggestions and questions.

MishaDemianenko:
@chrisvest turns out that ha starting internal transaction before clear availability requirements, so some part of PR get removed and cleaned, will reply to comments that are on lines that are still relevant.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
MishaDemianenko: newTransactionsLock and assertCurrentThreadIsNotBlockingNewTransactions is not relevant for this kind of case because this assertion will not gonna help you to terminate threads that want to start new transaction, while KernelTransactions are stopped (it can be the case that this particular KernelTransactions instance will not gonna be started ever again and we want to release those threads in that case), in the same time we do not want transactions that are started during safe zone marking phase to be terminated for non related reason.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionsTest.java
MishaDemianenko: and this is miserable mistake! was thinking for some reason that OtherThreadRule doing that for me, will clean this up!

MishaDemianenko:
@chrisvest updated

PR #8951, opened 2017-03-06
Drops label indexes from unselected providers

tinwelint:
This helps keeping a correct and up to date label index even when
potentially flipping between different providers, even tho it's generally
very uncommon.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/DeleteStoresFromOtherLabelScanStoreProviders.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/DeleteStoresFromOtherLabelScanStoreProviders.java
@@ -53,7 +53,10 @@ public class DeleteStoresFromOtherLabelScanStoreProviders implements DependencyR
             {
                 try
                 {
-                    provider.drop();
+                    if ( !provider.isReadOnly() )
+                    {
+                        provider.drop();
+                    }
                 }
                 catch ( IOException e )
                 {
--- a/community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/DeleteStoresFromOtherLabelScanStoreProviders.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/DeleteStoresFromOtherLabelScanStoreProviders.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.extension.dependency;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+
+import org.neo4j.graphdb.DependencyResolver;
+import org.neo4j.kernel.impl.api.scan.LabelScanStoreProvider;
+
+/**
+ * Given an already selected {@link LabelScanStoreProvider}, tell all the other {@link LabelScanStoreProvider}
+ * to {@link LabelScanStoreProvider#drop() drop} their stores, so that there's only one authoritative provider.
+ */
+public class DeleteStoresFromOtherLabelScanStoreProviders implements DependencyResolver.SelectionStrategy
+{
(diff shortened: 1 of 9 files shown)


community/neo4j/src/test/java/org/neo4j/store/label/DropOtherLabelIndexesIT.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Rule
+    public final TestDirectory directory = TestDirectory.testDirectory();
+
+    /**
+     * Flip between different label index providers and assert that the unselected label indexes
+     * gets deleted so that the selected index will not miss updates if it wasn't the selected index in all sessions.
+     */
+    @Test
+    public void shouldDropUnselectedLabelIndexes() throws Exception
```
burqen, line 57: Can we also have a test for `LUCENE -> NATIVE -> LUCENE`

burqen (changes requested):
Approved except for missing test case

davidegrohmann:
@tinwelint how is this change gonna effect backups?

tinwelint:
@davidegrohmann I wouldn't think it would affect backups. As long as all uses of a given db has the same label index config, i.e. the same label index provider selected this thing about dropping "the other" label indexes doesn't come into play. Have you spotted something I've missed perhaps?

community/neo4j/src/test/java/org/neo4j/store/label/DropOtherLabelIndexesIT.java
tinwelint: Added, or rather a bit of randomness. That exposed an issue with trying to delete non-existent native lss too!

burqen:
@tinwelint Edited test looks good. Will wait for backup discussion with @davidegrohmann before merging.

burqen (approved):

davidegrohmann:
@tinwelint what happens if you backup first by using an index provider and then do another after habing switched to another index provider?

tinwelint:
@davidegrohmann that only becomes interesting once you then start up from the backup, at which point this logic will kick in, right?

tinwelint:
@davidegrohmann and the one you're pulling the backup from also has this logic.

With that said there may be an edge-case where starting up w/o all label index providers on the classpath and then there may be a problem, but that's additional checks. This PR still makes things better.

davidegrohmann:
@tinwelint sounds good. Do you plan to have tests ensuring that expected behaviour?

community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/DeleteStoresFromOtherLabelScanStoreProviders.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.UncheckedIOException;
+
+import org.neo4j.graphdb.DependencyResolver;
+import org.neo4j.kernel.impl.api.scan.LabelScanStoreProvider;
+
+/**
+ * Given an already selected {@link LabelScanStoreProvider}, tell all the other {@link LabelScanStoreProvider}
+ * to {@link LabelScanStoreProvider#drop() drop} their stores, so that there's only one authoritative provider.
+ */
+public class DeleteStoresFromOtherLabelScanStoreProviders implements DependencyResolver.SelectionStrategy
```
MishaDemianenko, line 32: i would say this does not look like a selection strategy :) would say that we need to expose method on dependency resolver instead to do things like that

MishaDemianenko: but if you strongly disagree i can live with this thingy as well

tinwelint: I know, I sort of crinched a bit when implementing it. but It's the best/only means of accessing these extensions. Dunno, man

tinwelint:
@davidegrohmann that'd be a good addition, sure

burqen:
@tinwelint Test failures could be relevant

community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/DeleteStoresFromOtherLabelScanStoreProviders.java
MishaDemianenko: I would say if you have some time now - we do that now and i would prefer that, otherwise lets add a tech debt card and fix it during next tech debt cleanup work.

tinwelint: The alternative is to increase the surface of DependencyResolver, I think. Is that a good thing? Do you expect this to be debt that would be a hinderance to other development in this area. I wouldn't think so.

MishaDemianenko: I would say if we will provide a way to get all dependencies of requested type it should be fine from point of view of DependencyResolver API.
No, thats definitely not a problem to anything else then that

tinwelint: FTR: We decided to not change this

MishaDemianenko (approved):

PR #8952, opened 2017-03-06
Handling of composite index in transaction state

fickludd:
This PR adds support for composite indexes in the transaction state. This means that updates in the same transaction will be correctly handled when doing index queries.

The TxState managed a map of changes to the relevant indexes, that is updated on label and proprerty changes. This makes writes more complicated, by reads should be unaffected. The new updating code is placed in the new class `IndexTxStateUpdater`. 

This PR also extends the IndexQuery object to be a predicate, so it can be used directly to verify that the results returned by an index actually fulfill the query. This is used to defend against rounding errors in `LookupFilter`. 

To support the special merge seek pathway, the index entry resource id mechamism is extended to support composite indexes. (`ResourceTypes.indexEntryResourceId`)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -131,6 +131,7 @@ public Property nodeSetProperty( KernelStatement state, long nodeId, DefinedProp
     {
         try ( Cursor<NodeItem> cursor = nodeCursorById( state, nodeId ) )
         {
+            // TODO: support composite indexes
             NodeItem node = cursor.get();
```
chrisvest, line 137: This will be in a separate PR?

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/IndexQuery.java
```diff
@@ -278,6 +325,38 @@ public IndexQueryType type()
             return IndexQueryType.rangeString;
         }
 
+        @Override
+        public boolean test( Object value )
+        {
+            assert false : "StringRangePredicate.test() has not been tested yet... Add tests and remove assert";
```
chrisvest: This.

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/IndexQuery.java
```diff
@@ -315,6 +394,13 @@ public IndexQueryType type()
             return IndexQueryType.stringPrefix;
         }
 
+        @Override
+        public boolean test( Object value )
+        {
+            assert false : "StringPrefixPredicate.test() has not been tested yet... Add tests and remove assert";
+            return value != null && value instanceof String && ((String)value).startsWith( prefix );
```
chrisvest: This.

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/IndexQuery.java
```diff
@@ -337,6 +423,13 @@ public IndexQueryType type()
             return IndexQueryType.stringContains;
         }
 
+        @Override
+        public boolean test( Object value )
+        {
+            assert false : "StringContainsPredicate.test() has not been tested yet... Add tests and remove assert";
+            return value != null && value instanceof String && ((String)value).contains( contains );
```
chrisvest: This.

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/IndexQuery.java
```diff
@@ -359,6 +452,13 @@ public IndexQueryType type()
             return IndexQueryType.stringSuffix;
         }
 
+        @Override
+        public boolean test( Object value )
+        {
+            assert false : "StringSuffixPredicate.test() has not been tested yet... Add tests and remove assert";
+            return value != null && value instanceof String && ((String)value).endsWith( suffix );
```
chrisvest: And this.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
chrisvest: Or is this actually done?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -154,32 +158,47 @@ public Property nodeSetProperty( KernelStatement state, long nodeId, DefinedProp
...
     }
 
-    private void validateNoExistingNodeWithLabelAndProperty(
+    private ExactPredicate[] getAllPropertyValues( KernelStatement state, SchemaDescriptor schema, NodeItem node )
+    {
+        int[] propertyIds = schema.getPropertyIds();
+        ExactPredicate[] values = new ExactPredicate[propertyIds.length];
+        for ( int i = 0; i < values.length; i++ )
+        {
+            Object value = nodeGetProperty( state, node, propertyIds[i] );
```
chrisvest: I think it'll be faster to implement this in terms of the `Cursor<PropertyItem>` from `nodeGetProperties`.

chrisvest (commented):
I haven't fully reviewed the transaction state specific parts, because that's fairly intense, but I have a few comments that I thought would be good to get across sooner rather than later.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
fickludd: It's not done, but part of another card

chrisvest:
@fickludd @ragadeeshu There are lot of test failures. It looks like type coercion semantics are not preserved on index lookups into your transaction state.

community/codegen/src/test/java/org/neo4j/codegen/CodeGenerationTest.java
```diff
@@ -223,6 +225,13 @@ public void shouldGenerateMethodReturningFieldValue() throws Throwable
         assertMethodReturningField( double.class, 42D );
         assertMethodReturningField( String.class, "42" );
         assertMethodReturningField( int[].class, new int[]{42} );
+        assertMethodReturningField( Map.Entry[].class, Collections.singletonMap( 42, "42" ).entrySet().toArray( new Map.Entry[0] ) );
+    }
+
+    @Test
+    public void temp() throws Throwable
```
chrisvest: `temp`?

fickludd: oops

PR #8953, opened 2017-03-06
Do not use PrimitiveIntSet for relTypes

davidegrohmann:
Tentative fix for a performance regression.

### DIFF
--- a/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveIntCollections.java
+++ b/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveIntCollections.java
@@ -20,6 +20,7 @@
 package org.neo4j.collection.primitive;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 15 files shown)


davidegrohmann:
Pending benchmarks results.

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveIntCollections.java
```diff
@@ -848,4 +840,33 @@ public static boolean contains( int[] values, int candidate )
...
                     " when adding items uniquely to a collection:" + collection );
         }
-    }}
+    }
+
+    /**
+     * Deduplicates values in the {@code values} array.
+     *
+     * @param values sorted array of int values.
+     * @return the provided array if no duplicates were found, otherwise a new shorter array w/o duplicates.
```
chrisvest, line 849: The implementation does not appear to require the input array to be sorted. And the test is also not using a sorted array. If it _did_ require the input array to be sorted, then this could be implemented in something akin to O(N) time. And if you don't need stable ordering, you could probably implement it by copying into a set in O(N log N) time. Not sure how much it matters, though, since N is probably often quite small here. You could do a sorted-ness check and, if sorted, a duplicate existence check in a single pass; each element must be strictly greater than the one before it.

chrisvest: If you add a check like this, you can extract the actual deduplication to a helper method, to give the JIT more options for inlining and optimisation.

davidegrohmann: The comment on the old code was lying, indeed it didn't need the array to be sorted and we never sorted that array anywhere. This code is exactly the same code we had before I have deleted it, so I am gonna readd it as it was. The sorting problem can be dealt with later in a separate PR if we think it is important.

PR #8955, opened 2017-03-06
SchemaStore fix-ups

chrisvest:
- [x] Add missing support for creating composite indexes through the Core API.
- [x] Fix serialisation of schema records that take up more than one dynamic record.
- [x] Add computed unique names to SchemaRule.
- [x] Add a name field to the new schema records for indexes and constraints, so that they in the future can be addressed by name on the surface – this will be used instead of the computed SchemaRule name, if it exists (and can initially be set to the computed SchemaRule name).


community/kernel/src/test/java/org/neo4j/test/mockito/matcher/Neo4jMatchers.java
```diff
@@ -618,20 +619,25 @@ public void describeTo( Description description )
         };
     }
 
-    public static IndexDefinition createIndex( GraphDatabaseService beansAPI, Label label, String property )
+    public static IndexDefinition createIndex( GraphDatabaseService beansAPI, Label label, String... property )
```
ragadeeshu: `String... properties` to match `createIndexNoWait`?

community/kernel/src/test/java/org/neo4j/kernel/impl/store/SchemaStorageTest.java
```diff
@@ -115,6 +119,109 @@ public void shouldReturnIndexRuleForLabelAndProperty()
...
+        assertNotNull( rule );
+        assertTrue( SchemaDescriptorPredicates.hasLabel( rule, labelId( LABEL1 ) ) );
+        for ( String prop : props )
+        {
+            assertTrue( SchemaDescriptorPredicates.hasProperty( rule, propId( prop ) ) );
+        }
+        assertEquals( NewIndexDescriptor.Type.GENERAL, rule.getIndexDescriptor().type() );
+    }
+
+    @Ignore("it is currently disputed whether constraints should be supported via Core API or not.")
```
ragadeeshu: Wow

community/kernel/src/main/java/org/neo4j/kernel/impl/store/record/RecordSerializer.java
```diff
@@ -25,7 +25,7 @@
 
 public class RecordSerializer
 {
-    private final List<RecordSerializable> serializables = new ArrayList<RecordSerializable>();
+    private final List<RecordSerializable> serializables = new ArrayList<>();
 
     public RecordSerializer append( RecordSerializable serializable )
```
fickludd: I think we could remove this class completely, but let's wait. Good simplifications.

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/SchemaRule.java
```diff
@@ -26,23 +26,97 @@
...
+    }
+
+    /**
+     * Set the name of this rule, if it has not been set already.
+     *
+     * Note that once a name has been observed - either via this method or via {@link #getName()} – for this rule,
+     * it cannot be changed.
+     * @param name The name desired for this rule.
+     * @return {@code true} if the name was set, {@code false} if this rule already has a name.
+     */
```
fickludd: I don't know, I find this to be a bit too complicated? How about making the name a final field instead, and pass it into the constructor of the SchemaRules. Then we can make a factory method without the name, which would auto-generate the name. This would also avoid the instance of checks in `generateName`.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/SchemaStorageTest.java
```diff
@@ -115,6 +119,109 @@ public void shouldReturnIndexRuleForLabelAndProperty()
...
+    {
+        String a = "a", b = "b", c = "c", d = "d", e = "e", f = "f";
+        createSchema( db ->
+        {
+            db.schema().constraintFor( Label.label( LABEL1 ) )
+              .assertPropertyIsUnique( a ).assertPropertyIsUnique( b ).assertPropertyIsUnique( c )
+              .assertPropertyIsUnique( d ).assertPropertyIsUnique( e ).assertPropertyIsUnique( f ).create();
+        } );
+
+        IndexRule rule = storage.indexGetForSchema( NewIndexDescriptorFactory.forLabel(
```
fickludd: this should be `NewIndexDescriptorFactory.uniqueForLabel` I think

community/kernel/src/test/java/org/neo4j/kernel/impl/store/SchemaStorageTest.java
```diff
@@ -115,6 +119,109 @@ public void shouldReturnIndexRuleForLabelAndProperty()
...
+
+        assertNotNull( rule );
+        assertTrue( SchemaDescriptorPredicates.hasLabel( rule, labelId( LABEL1 ) ) );
+        assertTrue( SchemaDescriptorPredicates.hasProperty( rule, propId( a ) ) );
+        assertTrue( SchemaDescriptorPredicates.hasProperty( rule, propId( b ) ) );
+        assertTrue( SchemaDescriptorPredicates.hasProperty( rule, propId( c ) ) );
+        assertTrue( SchemaDescriptorPredicates.hasProperty( rule, propId( d ) ) );
+        assertTrue( SchemaDescriptorPredicates.hasProperty( rule, propId( e ) ) );
+        assertTrue( SchemaDescriptorPredicates.hasProperty( rule, propId( f ) ) );
+        assertEquals( NewIndexDescriptor.Type.UNIQUE, rule.getIndexDescriptor().type() );
```
fickludd: Also you can do assert directly on `NewIndexDescriptor` equality if you don't care to see exactly what was wrong. Maybe we do in this case.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/record/SchemaRuleSerializationTest.java
```diff
@@ -131,47 +230,75 @@ public void shouldReturnCorrectLengthForConstraintRules() throws MalformedSchema
...
 
     @Test
     public void shouldParseRelationshipPropertyExistsRule() throws Exception
     {
-        assertParseRelationshipPropertyExistsRule( "/////ww9XAAAIUAAAQAAF+c=" ); // LEGACY6
-        assertParseRelationshipPropertyExistsRule( "AAAhQAUAABfn" ); // LEGACY6
+        assertParseRelationshipPropertyExistsRule( "/////ww9XAAAIUAAAQAAF+c=", "constraint_51" );
+        assertParseRelationshipPropertyExistsRule( "AAAhQAUAABfn", "constraint_51" ); // LEGACY6
+        assertParseRelationshipPropertyExistsRule( "/////ww9XAAAIUAAAQAAF+djdXN0b21fbmFtZQA=", "custom_name" ); // named rule
+        assertParseRelationshipPropertyExistsRule( addNullByte( "/////ww9XAAAIUAAAQAAF+c=" ), "constraint_51" ); // empty name
```
fickludd: Lot's of testing, nice!

fickludd (commented):
I like it, let's just iterate the schema name implementation, and then I'm happy.

chrisvest:
I want to remove the null-terminated strings as well, and use the regular length-prefixed strings instead.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/record/ConstraintRule.java
```diff
@@ -43,9 +43,26 @@ public static ConstraintRule constraintRule(
         return new ConstraintRule( id, descriptor, ownedIndexRule );
     }
 
+    public static ConstraintRule constraintRule(
+            long id, ConstraintDescriptor descriptor, String name )
+    {
+        return new ConstraintRule( id, descriptor, null, name );
```
fickludd, line 49: I was thinking to generate the default name here, but whatever. Having it in SchemaRule brings the functionality together, and that's always nice.

fickludd (approved):
Nice, will merge when green.

PR #8957, opened 2017-03-06
Fail rebuild from logs if final target store is inconsistent.

MishaDemianenko:


tinwelint:
Sooo... issue for something like 3.0 instead?

MishaDemianenko:
@tinwelint since i have change test a bit also, can we merge this PR into current 3.2 at it is and i will open another pr against 3.0 and will merge it into 3.1 as well?

tinwelint:
sure, absolutely fine

PR #8960, opened 2017-03-07
Rule chain order in CoreReplicationIT.

MishaDemianenko:
Cluster rule should be around timeout rule to be able to shutdown
cluster in case of timeout and normal termination.


MishaDemianenko:
@burqen please take a look :) thx

burqen (approved):

PR #8963, opened 2017-03-07
Change vocabulary label scan store -> label index

mariascharin:
Also removing wording about cost, as label indexes will be native in 3.2.


jjaderberg (approved):
LGTM

PR #8965, opened 2017-03-07
Modify FileWatchIT to perform modifications till change will be observed.

MishaDemianenko:
Because of file last modification precision we can't be sure that we will observe
modifications, perform them in test till we will see them.
Use highest available sensitivity for file system watcher.


chrisvest (approved):
Looks good!

PR #8966, opened 2017-03-07
Simplify import panic

tinwelint:


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/DetailedExecutionMonitor.java
```diff
@@ -50,34 +50,21 @@ public DetailedExecutionMonitor( PrintStream out, long intervalSeconds )
     }
```
MishaDemianenko, line 50: this whole class is not used atm, but i will assume it's here debugging/investigation purposes

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/DetailedExecutionMonitor.java
```diff
@@ -50,34 +50,21 @@ public DetailedExecutionMonitor( PrintStream out, long intervalSeconds )
...
     @Override
-    public void start( StageExecution[] executions )
+    public void start( StageExecution execution )
     {
         StringBuilder names = new StringBuilder();
-        for ( StageExecution execution : executions )
-        {
-            names.append( names.length() > 0 ? ", " : "" ).append( execution.getStageName() );
-        }
+        names.append( names.length() > 0 ? ", " : "" ).append( execution.getStageName() );
```
MishaDemianenko: This whole message construction can be simplified since we have only one stage here always now, now reason to even have names variable,  also now we always `EXECUTING STAGE` and never `STAGES`

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/StageControl.java
```diff
@@ -26,4 +26,6 @@
 public interface StageControl
 {
     void panic( Throwable cause );
```
MishaDemianenko, line 28: What do you thin if `StageControl` will also be `Panicable` instead of having its own panic method?

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/ImportPanicIT.java
```diff
@@ -0,0 +1,168 @@
...
+        };
+    }
+
+    private Supplier<CharReadable> fileAsCharReadable( File file )
+    {
+        return () ->
+        {
+            try
+            {
+                return Readables.wrap( fs.openAsReader( file, Charset.defaultCharset() ) );
```
MishaDemianenko: why do we want system default charset and not UTF-8 for example always?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/executor/DynamicTaskExecutor.java
```diff
@@ -164,28 +164,32 @@ private void notifyProcessors()
     }
```
MishaDemianenko, line 159: abortQueued field is never assigned

MishaDemianenko (commented):
Couple of small comments

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/DetailedExecutionMonitor.java
tinwelint: We actually used it while developing this, yes

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/StageControl.java
tinwelint: We discussed it, but opted out because Panicable instances report and receive panic whereas  StageControl spreads panic, managing the Panicables, sort of. To have even StageControl be Panicable would make code navigation hard because you'd find yourself navigating around in loops without knowing where to really go.

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/ImportPanicIT.java
tinwelint: Since only ASCII characters is used it doesn't matter. But sure using UTF-8 would have this test run with same charset on all environments. Is that good or bad?

MishaDemianenko: i would say it's pure gold good :)

tinwelint: Gold is good. I couldn't decide, thanks for deciding for me :) fixing

PR #8967, opened 2017-03-07
Ability to set unlimited bad tolerance in import tool

tinwelint:
by using --bad-tolerance -

### DIFF
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -100,7 +100,7 @@ import static org.neo4j.unsafe.impl.batchimport.input.csv.DataFactories.defaultF
  */
 public class ImportTool
 {
-    private static final String UNLIMITED = "-";
+    private static final String UNLIMITED = "true";
     private static final int UNSPECIFIED = -1;
 
     enum Options
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -60,6 +60,7 @@ import org.neo4j.kernel.lifecycle.LifeSupport;
 import org.neo4j.unsafe.impl.batchimport.BatchImporter;
 import org.neo4j.unsafe.impl.batchimport.ParallelBatchImporter;
 import org.neo4j.unsafe.impl.batchimport.cache.idmapping.string.DuplicateInputIdException;
+import org.neo4j.unsafe.impl.batchimport.input.BadCollector;
 import org.neo4j.unsafe.impl.batchimport.input.Collector;
 import org.neo4j.unsafe.impl.batchimport.input.Input;
 import org.neo4j.unsafe.impl.batchimport.input.InputException;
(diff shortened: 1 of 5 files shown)


spacecowboy:
Nice, then I didn't have to do it all myself (see https://github.com/neo4j/neo4j/pull/8968)

But why did you pick `-` as the infinity value? I think that might be confusing. A naked dash is typically used to indicate that stdin should be read.

for example:
http://unix.stackexchange.com/questions/41828/what-does-dash-at-the-end-of-a-command-mean

which means some users could be confused by `neo4j-import --bad-tolerance -`

tinwelint:
@spacecowboy right you are, what would be better?

spacecowboy:
@tinwelint `-1` comes to mind, it also matches what the constant is. otherwise, a string matching a positive action like `true` perhaps

tinwelint:
Good idea, I'd say `true` in that case. `-1` feels like a developer thing, not a user thing to me.

burqen:
@tinwelint Can we have some labels on this guy please.

tinwelint:
@spacecowboy did you already merge related stuff to this commit?

spacecowboy:
@tinwelint I have made the almost equivalent changes for `neo4j-admin` on 3.2. It contains a subset of your commit and should not conflict (git will complain about a conflict but the code should be identical).

Ref: https://github.com/neo4j/neo4j/pull/8968
subset of this pr: https://github.com/neo4j/neo4j/pull/8968/commits/1e313023128d8459169251b38c7f052c1d30cb03

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -99,6 +100,7 @@
  */
 public class ImportTool
 {
+    private static final String UNLIMITED = "-";
```
MishaDemianenko: As @spacecowboy already mentioned that would be much more understandable to use `-1` as representation of unlimited value

community/import-tool/src/test/java/org/neo4j/tooling/ImportToolTest.java
```diff
@@ -1138,6 +1138,27 @@ public void shouldPrintReferenceLinkOnDataImportErrors() throws Exception
                 "/import/import-tool-header-format/#import-tool-id-spaces" );
     }
 
+    @Test
+    public void shouldCollectUnlimitedNumberOfBadEntries() throws Exception
+    {
+        // GIVEN
+        List<String> nodeIds = new ArrayList<>();
```
MishaDemianenko: will `Collections.nCopies( 10_000, "A" )` do the trick instead?

MishaDemianenko (changes requested):
One usability and minor comment, good otherwise

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
tinwelint: I do think -1 isn't particularly user friendly, is it? I'd vote for `true`, as discussed above

MishaDemianenko: do you mean `neo4j-import --bad-tolerance -unlimited=true` or smt else?

tinwelint: `--bad-tolerance=true`

I think `--bad-tolerance=-1` is so developer:y

MishaDemianenko (approved):

PR #8968, opened 2017-03-07
Added ability to ignore bad data in `neo4j-admin import`

spacecowboy:
## Updated help text

```
usage: neo4j-admin import [--mode=csv] [--database=<name>]
                          [--additional-config=<config-file-path>]
                          [--report-file=<filename>]
                          [--nodes[:Label1:Label2]=<"file1,file2,...">]
                          [--relationships[:RELATIONSHIP_TYPE]=<"file1,file2,...">]
                          [--id-type=<STRING|INTEGER|ACTUAL>]
                          [--input-encoding=<character-set>]
                          [--ignore-extra-columns[=<true|false>]]
                          [--ignore-duplicate-nodes[=<true|false>]]
                          [--ignore-missing-nodes[=<true|false>]]
usage: neo4j-admin import --mode=database [--database=<name>]
                          [--additional-config=<config-file-path>]
                          [--from=<source-directory>]

Import a collection of CSV files with --mode=csv (default), or a database from a
pre-3.0 installation with --mode=database.

options:
  --database=<name>
      Name of database. [default:graph.db]
  --additional-config=<config-file-path>
      Configuration file to supply additional configuration in. [default:]
  --mode=<database|csv>
      Import a collection of CSV files or a pre-3.0 installation. [default:csv]
  --from=<source-directory>
      The location of the pre-3.0 database (e.g. <neo4j-root>/data/graph.db).
      [default:]
  --report-file=<filename>
      File in which to store the report of the csv-import.
      [default:import.report]
  --nodes[:Label1:Label2]=<"file1,file2,...">
      Node CSV header and data. Multiple files will be logically seen as one big
      file from the perspective of the importer. The first line must contain the
      header. Multiple data sources like these can be specified in one import,
      where each data source has its own header. Note that file groups must be
      enclosed in quotation marks. [default:]
  --relationships[:RELATIONSHIP_TYPE]=<"file1,file2,...">
      Relationship CSV header and data. Multiple files will be logically seen as
      one big file from the perspective of the importer. The first line must
      contain the header. Multiple data sources like these can be specified in
      one import, where each data source has its own header. Note that file
      groups must be enclosed in quotation marks. [default:]
  --id-type=<STRING|INTEGER|ACTUAL>
      Each node must provide a unique id. This is used to find the correct nodes
      when creating relationships. Possible values are:
        STRING: arbitrary strings for identifying nodes,
        INTEGER: arbitrary integer values for identifying nodes,
        ACTUAL: (advanced) actual node ids.
      For more information on id handling, please see the Neo4j Manual:
      https://neo4j.com/docs/operations-manual/current/tools/import/
      [default:STRING]
  --input-encoding=<character-set>
      Character set that input data is encoded in. [default:UTF-8]
  --ignore-extra-columns=<true|false>
      If un-specified columns should be ignored during the import.
      [default:false]
  --ignore-duplicate-nodes=<true|false>
      If duplicate nodes should be ignored during the import. [default:false]
  --ignore-missing-nodes=<true|false>
      If relationships referring to missing nodes should be ignored during the
      import. [default:false]
```

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/CsvImporter.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/CsvImporter.java
@@ -66,57 +66,6 @@ class CsvImporter implements Importer
     private final boolean ignoreDuplicateNodes;
     private final boolean ignoreExtraColumns;
 
-    public static String description()
-    {
-        return "--mode=csv\n" +
-                "        Import a database from a collection of CSV files.\n" +
-                "--report-file=<filename>\n" +
-                "        File name in which to store the report of the import.\n" +
-                "        Defaults to " + ImportCommand.DEFAULT_REPORT_FILE_NAME + " in the current directory.\n" +
-                "--nodes[:Label1:Label2]=\"<file1>,<file2>,...\"\n" +
-                "        Node CSV header and data. Multiple files will be logically seen as\n" +
-                "        one big file from the perspective of the importer. The first line\n" +
-                "        must contain the header. Multiple data sources like these can be\n" +
-                "        specified in one import, where each data source has its own header.\n" +
-                "        Note that file groups must be enclosed in quotation marks.\n" +
-                "--relationships[:RELATIONSHIP_TYPE]=\"<file1>,<file2>,...\"\n" +
-                "        Relationship CSV header and data. Multiple files will be logically\n" +
-                "        seen as one big file from the perspective of the importer. The first\n" +
-                "        line must contain the header. Multiple data sources like these can be\n" +
-                "        specified in one import, where each data source has its own header.\n" +
-                "        Note that file groups must be enclosed in quotation marks.\n" +
-                "--id-type=<id-type>\n" +
-                "        Each node must provide a unique id. This is used to find the correct\n" +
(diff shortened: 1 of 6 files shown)


spacecowboy:
Based on https://github.com/neo4j/neo4j/pull/8967

community/dbms/src/main/java/org/neo4j/commandline/dbms/CsvImporter.java
```diff
@@ -145,4 +112,9 @@ public void doImport() throws IOException
         ImportTool.doImport( outsideWorld.errorStream(), outsideWorld.errorStream(), storeDir, logsDir, reportFile, fs,
                 nodesFiles, relationshipsFiles, false, input, config, badOutput, configuration );
     }
+
+    private boolean isIgnoringSomething()
+    {
+        return ignoreBadRelationships | ignoreDuplicateNodes | ignoreExtraColumns;
```
praveenag: Is there a reason why you want to do bitwise OR here instead of `||` Did you want to explicitly avoid short-circuiting?

spacecowboy: Ah good point! Fixed.

PR #8969, opened 2017-03-07
Faster HardKillIT

tinwelint:
by removing arbitrary and pretty defensively sized Thread.sleeps
and instead waiting for specific conditions.

Reduces execution time from 1min 30sec --> 20sec


tinwelint:
Unrelated flaky test

burqen (approved):

PR #8972, opened 2017-03-07
Update QueryLoggerIT queries log file check.

MishaDemianenko:


spacecowboy (changes requested):
This no longer actually tests rotation.

The test needs to fail if there is no `query.log.N` file(s). It's fine to only check for one rotated file but you need to check for it.

MishaDemianenko:
@spacecowboy can i check that i have multiple query log files instead?

spacecowboy:
@MishaDemianenko yes. the problem in windows was that a new file could not be created because it's a rename operation and the file was still open. So if you just check that there is more than one file present with name `query.log*` that should be fine.

MishaDemianenko:
@spacecowboy updated

enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerIT.java
```diff
@@ -312,14 +313,15 @@ public void queryLogRotation() throws Exception
...
-        assertTrue( "Expected shifted log file to have at least one log entry", lines.size() >= 1 );
+        assertThat( "Expect to have more then one query log file.", queryLogs.length, greaterThanOrEqualTo( 2 ) );
 
-        lines = readAllLines( shiftedLogFilename2 );
-        assertTrue( "Expected second shifted log file to have at least one log entry", lines.size() >= 1 );
+        List<String> loggedQueries = Arrays.stream( queryLogs )
+                                        .map( this::readAllLinesSilent )
+                                        .flatMap( Collection::stream )
+                                        .collect( Collectors.toList() );
+        assertTrue( "Expected log file to have at least one log entry", loggedQueries.size() >= 1 );
```
spacecowboy: Hmm. Do we miss any log statements due to the rotation here? Wouldn't they all end up in the buffer and then sent to the new log file? In that case we could assert that `loggedQueries.size() == 100` (the number of queries sent).

spacecowboy (approved):

MishaDemianenko:
ping @ragadeeshu

MishaDemianenko:
@ragadeeshu updated with additional check

PR #8976, opened 2017-03-08
Add a flag for verifying generated bytecode

thobe:
This flag is enabled by default in tests through inspecting the assertion status on the GeneratedQueryStructure class. If assertions are enabled, the byte code verification will be enabled, otherwise it is not enabled. When running tests, assertions are enabled.

For example, if compiling a simple method that tries to use a primitive `int` where a boxed `java.lang.Integer` is expected, you will now get the following error message from the codegen library:

```
org.neo4j.codegen.CompilationFailureException: Compilation failed with 1 reported issues.
		Error at instruction 1: Expected an object reference, but found I
		in box(I)Ljava/lang/Integer;
		    00000 [SimpleClass I : -] :     ILOAD 1
		>>> 00001 [SimpleClass I : I] :     ARETURN
```

Where `>>>` marks the location where the error was detected.

### DIFF
--- a/community/codegen/pom.xml
+++ b/community/codegen/pom.xml
@@ -12,6 +12,7 @@
     <bundle.namespace>org.neo4j.codegen</bundle.namespace>
     <license-text.header>GPL-3-header.txt</license-text.header>
     <licensing.prepend.text>notice-gpl-prefix.txt</licensing.prepend.text>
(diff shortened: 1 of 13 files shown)


thobe:
Waiting on #9006 to fix flaky test in target branch.

PR #8984, opened 2017-03-09
Composite uniqueness constraint enforcement on nodeSetProperty

fickludd:
Enforce composite uniqueness constrains on nodeSetProperty. 

This PR introduces the new class `NodeSchemaMatcher` that finds indexes or constraints that are valid for a node. This class is now used to find the relevant uniqueness constraints when setting a property, and the relevant indexes to update in the TxState.

Composite uniqueness constraints were already enforced on nodeAddLabel, but this PR adds some tests to verify this behaviour.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
@@ -167,15 +167,22 @@ public class ConstraintEnforcingEntityOperations implements EntityOperations, Sc
         return getAllPropertyValues( state, schema, node, DefinedProperty.NO_SUCH_PROPERTY );
     }
 
+    /**
+     * Fetch the property values for all properties in schema for a given node. Return these as an exact predicate
+     * array.
+     *
+     * The changedProperty is used to override the store/txState value of the property. This is used when we intend
+     * to change a property, and that to verify that the post-change values do not validate some constraint.
+     */
     private ExactPredicate[] getAllPropertyValues( KernelStatement state, SchemaDescriptor schema, NodeItem node,
-            DefinedProperty specialProperty )
+            DefinedProperty changedProperty )
     {
         int[] schemaPropertyIds = schema.getPropertyIds();
         ExactPredicate[] values = new ExactPredicate[schemaPropertyIds.length];
 
         int nMatched = 0;
         Cursor<PropertyItem> nodePropertyCursor = nodeGetProperties( state, node );
-        int specialPropId = specialProperty.propertyKeyId();
+        int changedPropId = changedProperty.propertyKeyId();
         while ( nodePropertyCursor.next() )
         {
             PropertyItem property = nodePropertyCursor.get();
@@ -184,7 +191,7 @@ public class ConstraintEnforcingEntityOperations implements EntityOperations, Sc
             int k = ArrayUtils.indexOf( schemaPropertyIds, nodePropertyId );
             if ( k >= 0 )
             {
-                if ( nodePropertyId != specialPropId )
+                if ( nodePropertyId != changedPropId )
                 {
                     values[k] = IndexQuery.exact( nodePropertyId, property.value() );
                 }
@@ -192,12 +199,12 @@ public class ConstraintEnforcingEntityOperations implements EntityOperations, Sc
             }
         }
 
-        if ( specialPropId != NO_SUCH_PROPERTY_KEY )
+        if ( changedPropId != NO_SUCH_PROPERTY_KEY )
         {
-            int k = ArrayUtils.indexOf( schemaPropertyIds, specialPropId );
+            int k = ArrayUtils.indexOf( schemaPropertyIds, changedPropId );
             if ( k >= 0 )
             {
-                values[k] = IndexQuery.exact( specialPropId, specialProperty.value() );
+                values[k] = IndexQuery.exact( changedPropId, changedProperty.value() );
                 nMatched++;
             }
         }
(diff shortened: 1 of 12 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -133,47 +139,65 @@ public Property nodeSetProperty( KernelStatement state, long nodeId, DefinedProp
...
-                }
-                return false;
-            } );
+                    } );
         }
 
         return entityWriteOperations.nodeSetProperty( state, nodeId, property );
     }
 
+    private Iterator<ConstraintDescriptor> getConstraintsForProperty( KernelStatement state, int propertyId )
```
craigtaverner: At first I interpreted this method name as implying constraints matching exactly one property, but I can see now that it matches all constraints that involve at least the one property. I wonder if another method name might reduce the chance of this misunderstanding?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -133,47 +139,65 @@ public Property nodeSetProperty( KernelStatement state, long nodeId, DefinedProp
...
+        }
+
+        if ( specialPropId != NO_SUCH_PROPERTY_KEY )
+        {
+            int k = ArrayUtils.indexOf( schemaPropertyIds, specialPropId );
             if ( k >= 0 )
             {
-                values[k] = IndexQuery.exact( property.propertyKeyId(), property.value() );
+                values[k] = IndexQuery.exact( specialPropId, specialProperty.value() );
                 nMatched++;
```
craigtaverner: Would it be safer/easier to check if values does not contain null at the end, instead of tracking nMatched?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/IndexTxStateUpdater.java
```diff
@@ -85,75 +91,44 @@ public void onLabelChange( KernelStatement state, int labelId, NodeItem node, La
...
-            public int propertyId()
-            {
-                return after.propertyKeyId();
-            }
-        };
+        Iterator<NewIndexDescriptor> indexes = getIndexesForProperty( state, after.propertyKeyId() );
+        nodeIndexMatcher.onMatchingSchema( state, indexes, node, after.propertyKeyId(),
+                index -> {
+                    OrderedPropertyValues values = getOrderedPropertyValues( state, node, after, index.schema().getPropertyIds() );
+                    if ( values != null )
```
craigtaverner: This check leads me to expect that `getOrderedPropertyValues` will return null if the node does not have all the properties required by the index, but no such check seemed to exist in the `getOrderedPropertyValues` method.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/IndexTxStateUpdater.java
```diff
@@ -173,84 +148,51 @@ public void updateIndexIfApplicable( KernelStatement state, NodeItem node,
...
-                        ? changedProperty
-                        : Property.property( indexPropertyId, readOps.nodeGetProperty( state, node, indexPropertyId ) );
+            PropertyItem property = propertyCursor.get();
+            int k = ArrayUtils.indexOf( indexPropertyIds, property.propertyKeyId() );
+            if ( k >= 0 )
+            {
+                values[k] = indexPropertyIds[k] == changedProperty.propertyKeyId()
+                            ? changedProperty
+                            : Property.property( indexPropertyIds[k], property.value() );
+            }
```
craigtaverner, line 174: Feels like we should have a check for any nulls in values array, and return null array, since calling code seems to rely on that, right?

craigtaverner (changes requested):

craigtaverner (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -133,47 +139,65 @@ public Property nodeSetProperty( KernelStatement state, long nodeId, DefinedProp
...
+
     private ExactPredicate[] getAllPropertyValues( KernelStatement state, SchemaDescriptor schema, NodeItem node )
     {
-        int[] propertyIds = schema.getPropertyIds();
-        ExactPredicate[] values = new ExactPredicate[propertyIds.length];
+        return getAllPropertyValues( state, schema, node, DefinedProperty.NO_SUCH_PROPERTY );
+    }
+
+    private ExactPredicate[] getAllPropertyValues( KernelStatement state, SchemaDescriptor schema, NodeItem node,
+            DefinedProperty specialProperty )
```
chrisvest: `specialProperty` is not a very clarifying name, and it's hard to track the consequences of this value around the code.

Could you add some comments to help future selves? Is there a better name we could use for it?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/schema/NodeSchemaMatcher.java
```diff
@@ -0,0 +1,109 @@
...
+ * This class holds functionality to match LabelSchemaDescriptors to nodes
+ * @param <SUPPLIER> the type to match. Must implement LabelSchemaDescriptor.Supplier
+ */
+public class NodeSchemaMatcher<SUPPLIER extends LabelSchemaDescriptor.Supplier>
+{
+    private final EntityReadOperations readOps;
+
+    public NodeSchemaMatcher( EntityReadOperations readOps )
+    {
+        this.readOps = readOps;
```
chrisvest, line 42: You are avoiding the security access checks and statement.assertOpen checks by grabbing this as a field up front?

chrisvest (approved):
It looks okay to me. Just a few questions.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/schema/NodeSchemaMatcher.java
fickludd: I didn't think much about that, as far as I can tell this follows the same path as before this PR. If that is broken we should look at it more and fix in separate PR.

PR #8987, opened 2017-03-09
Add page cache hit metric into query profile report

MishaDemianenko:
Using newly introduced page cache hit metric it should be much more easier
to see how particular query use data that are already in page cache.
Each page cache `hit` indicate that data from cache was used and there was no page fault
when particular price of data request was served.
Now when hit metric added into query execution profile statistic we have possibility to
investigate how particular query utilize page cache.

### DIFF
--- a/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/spi/QueryContext.scala
+++ b/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/spi/QueryContext.scala
@@ -26,7 +26,6 @@ import org.neo4j.cypher.internal.compiler.v3_2.pipes.matching.PatternNode
 import org.neo4j.cypher.internal.compiler.v3_2.{IndexDescriptor, InternalQueryStatistics}
 import org.neo4j.cypher.internal.frontend.v3_2.SemanticDirection
 import org.neo4j.graphdb.{Node, Path, PropertyContainer, Relationship}
-import org.neo4j.io.pagecache.tracing.cursor.PageCursorTracer
 
 import scala.collection.Iterator
 
@@ -180,7 +179,7 @@ trait QueryContext extends TokenContext {
 
   def assertSchemaWritesAllowed(): Unit
 
-  def pageCursorTracer() : PageCursorTracer
+  def kernelStatisticProvider() : KernelStatisticProvider
 }
 
 trait Operations[T <: PropertyContainer] {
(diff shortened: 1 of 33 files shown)


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/spi/QueryContext.scala
```diff
@@ -21,12 +21,12 @@ package org.neo4j.cypher.internal.compiler.v3_2.spi
...
 import java.net.URL
 
-import org.neo4j.cypher.internal.compiler.v3_2.InternalQueryStatistics
+import org.neo4j.cypher.internal.compiler.v3_2.{IndexDescriptor, InternalQueryStatistics}
 import org.neo4j.cypher.internal.compiler.v3_2.commands.expressions.{Expander, KernelPredicate, UserDefinedAggregator}
 import org.neo4j.cypher.internal.compiler.v3_2.pipes.matching.PatternNode
 import org.neo4j.cypher.internal.frontend.v3_2.SemanticDirection
 import org.neo4j.graphdb.{Node, Path, PropertyContainer, Relationship}
-import org.neo4j.cypher.internal.compiler.v3_2.IndexDescriptor
+import org.neo4j.io.pagecache.tracing.cursor.PageCursorTracer
```
systay: We should not introduce any new store dependencies here. Instead, create a trait in the compiler module, and implement it in the compatibility package. There you can have your kernel dependencies.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/TransactionBoundQueryContext.scala
```diff
@@ -745,6 +746,10 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
 
   override def assertSchemaWritesAllowed(): Unit =
     transactionalContext.statement.schemaWriteOperations()
+
+  override def pageCursorTracer(): PageCursorTracer = {
+    transactionalContext.statement.asInstanceOf[KernelStatement].getPageCursorTracer
```
systay: Shouldn't `getPageCursorTracer` be in the `Statement` interface?

systay (changes requested):
Looks good, except for the introduction of a kernel dependency in the SPI

MishaDemianenko:
@systay updated!

thobe:
In `CompositeIndexAcceptanceTest`: _Cannot get single property of multi-property index_.

Based on a bad base revision?

systay (approved):

MishaDemianenko:
@thobe yep, since i will need to rebase anyway will wait for green build to do that :)

MishaDemianenko:
@thobe rebased

PR #8992, opened 2017-03-09
Make sure asking for compiled runtime does not fail by default

systay:


enterprise/cypher/cypher/src/test/scala/org/neo4j/cypher/ExecutionEngineTest.scala
```diff
@@ -40,7 +40,7 @@ import org.neo4j.kernel.impl.coreapi.TopLevelTransaction
 import org.neo4j.test.{TestEnterpriseGraphDatabaseFactory, TestGraphDatabaseFactory}
 
 import scala.collection.JavaConverters._
-import scala.collection.mutable
+import scala.collection.{Map, mutable}
```
thobe: This appears to change the type used for the overriding `databaseConfig()` method, such that this class no longer compiles.

thobe (changes requested):
The logical change looks ok, but the actual code changes don't compile.

thobe (approved):

PR #8993, opened 2017-03-09
Composite consistency checker

fickludd:
Adds checks of composite indexes and uniqueness constraints to the consistency checker. This in essence means duplicating existing tests but applied on composite alternative schema objects.

Looks huge, but this is because it sits on #8984. Only commits https://github.com/neo4j/neo4j/pull/8993/commits/127392591c21f0b8558d63ed5ec10cb2f9c67194, https://github.com/neo4j/neo4j/pull/8993/commits/50512cffd7fa618fd8cacb96eda9404ac48d0d6b and https://github.com/neo4j/neo4j/pull/8993/commits/50a03bbd6f5f847a45e0a1c17ec89c2f1deb9168 are relevant.

### DIFF
--- a/community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
+++ b/community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
@@ -238,15 +238,15 @@ public class PropertyAndNodeIndexedCheck implements RecordCheck<NodeRecord, Cons
         catch ( IndexNotApplicableKernelException e )
         {
             throw new RuntimeException( format(
-                    "Cannot continue %s, as index provider does not support exact composite query: %s",
-                    this.getClass().getSimpleName(), Arrays.toString( query ) ), e );
+                    "Consistency checking error: index provider does not support exact query %s",
+                    Arrays.toString( query ) ), e );
         }
 
         indexedNodeIds = LookupFilter.exactIndexMatches( propertyReader, indexedNodeIds, query );
         return indexedNodeIds;
     }
 
-    public static boolean nodeHasSchemaProperties(
+    private static boolean nodeHasSchemaProperties(
             PrimitiveIntObjectMap<PropertyBlock> nodePropertyMap, int[] indexPropertyIds )
     {
         for ( int indexPropertyId : indexPropertyIds )
--- a/community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
+++ b/community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Set;
 
 import org.neo4j.collection.primitive.Primitive;
+import org.neo4j.collection.primitive.PrimitiveIntObjectMap;
 import org.neo4j.collection.primitive.PrimitiveIntSet;
 import org.neo4j.collection.primitive.PrimitiveLongCollections;
 import org.neo4j.collection.primitive.PrimitiveLongIterator;
@@ -45,7 +46,6 @@ import org.neo4j.kernel.impl.store.record.PropertyRecord;
 import org.neo4j.kernel.impl.store.record.Record;
 import org.neo4j.storageengine.api.schema.IndexReader;
 
-import static org.neo4j.kernel.api.StatementConstants.NO_SUCH_PROPERTY_KEY;
 import static org.neo4j.kernel.impl.api.schema.NodeSchemaMatcher.nodeHasSchemaProperties;
 
 /**
@@ -90,23 +90,21 @@ public class PropertyAndNodeIndexedCheck implements RecordCheck<NodeRecord, Cons
             Collection<PropertyRecord> propertyRecs )
     {
         Set<Long> labels = NodeLabelReader.getListOfLabels( record, records, engine );
-        List<PropertyBlock> properties = null;
-        PrimitiveIntSet nodePropertyIds = null;
+        PrimitiveIntObjectMap<PropertyBlock> nodePropertyMap = null;
         for ( IndexRule indexRule : indexes.rules() )
         {
             long labelId = indexRule.schema().getLabelId();
             if ( labels.contains( labelId ) )
             {
-                if ( properties == null )
+                if ( nodePropertyMap == null )
                 {
-                    properties = propertyReader.propertyBlocks( propertyRecs );
-                    nodePropertyIds = propertyIds( properties );
+                    nodePropertyMap = properties( propertyReader.propertyBlocks( propertyRecs ) );
                 }
 
                 int[] indexPropertyIds = indexRule.schema().getPropertyIds();
-                if ( nodeHasSchemaProperties( nodePropertyIds, indexPropertyIds, NO_SUCH_PROPERTY_KEY ) )
+                if ( nodeHasSchemaProperties( nodePropertyMap, indexPropertyIds ) )
                 {
-                    Object[] values = getPropertyValues( properties, indexPropertyIds );
+                    Object[] values = getPropertyValues( nodePropertyMap, indexPropertyIds );
                     try ( IndexReader reader = indexes.accessorFor( indexRule ).newReader() )
                     {
                         long nodeId = record.getId();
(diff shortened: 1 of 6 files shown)


community/consistency-check/src/main/java/org/neo4j/consistency/report/ConsistencyReport.java
```diff
@@ -184,13 +184,13 @@ void forCounts( CountsEntry countsEntry,
...
 
         @Documented( "This node was not found in the expected index." )
-        void notIndexed( IndexRule index, Object propertyValue );
+        void notIndexed( IndexRule index, Object[] propertyValues );
 
         @Documented( "This node was found in the expected index, although multiple times" )
-        void indexedMultipleTimes( IndexRule index, Object propertyValue, long count );
+        void indexedMultipleTimes( IndexRule index, Object[] propertyValues, long count );
 
         @Documented( "There is another node in the unique index with the same property value." )
```
MishaDemianenko: this message is a bit wrong now

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
```diff
@@ -233,4 +195,69 @@ private void checkProperty( NodeRecord record,
...
+
+    private Object[] getPropertyValues( List<PropertyBlock> properties, int[] indexPropertyIds )
+    {
+        Object[] values = new Object[indexPropertyIds.length];
+        for ( int i = 0; i < indexPropertyIds.length; i++ )
+        {
+            PropertyBlock propertyBlock = propertyWithKey( properties, indexPropertyIds[i] );
+            if ( propertyBlock == null )
+            {
+                throw new IllegalStateException( "We should have checked that the index and node match before coming " +
```
MishaDemianenko: Exception message should be more descriptive in case if enduser will get it and should also include set of known property ids and requested id to give some possible insight what went wrong

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
```diff
@@ -233,4 +195,69 @@ private void checkProperty( NodeRecord record,
...
+    }
+
+    private PrimitiveLongIterator queryIndexOrEmpty( IndexReader reader, IndexQuery[] query )
+    {
+        PrimitiveLongIterator indexedNodeIds;
+        try
+        {
+            indexedNodeIds = reader.query( query );
+        }
+        catch ( IndexNotApplicableKernelException e )
```
MishaDemianenko, line 238: What exactly IndexNotApplicableKernelException mean here? shouldn't we report that something went wrong instead of assuming that we do not have matches?

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
```diff
@@ -70,139 +73,98 @@ public void check( NodeRecord record,
...
+                    properties = propertyReader.propertyBlocks( propertyRecs );
+                    nodePropertyIds = propertyIds( properties );
                 }
-                else
+
+                int[] indexPropertyIds = indexRule.schema().getPropertyIds();
+                if ( nodeHasSchemaProperties( nodePropertyIds, indexPropertyIds, NO_SUCH_PROPERTY_KEY ) )
                 {
-                    verifyNodeCorrectlyIndexed( nodeId, propertyValue, engine, indexRule, reader );
+                    Object[] values = getPropertyValues( properties, indexPropertyIds );
```
MishaDemianenko: if we will make properties map instead (id -> block) we can avoid here searching over property block again by ids that we have already extracted and check before

community/kernel/src/main/java/org/neo4j/kernel/impl/api/schema/NodeSchemaMatcher.java
```diff
@@ -0,0 +1,109 @@
...
+        {
+            if ( indexPropertyId != changedPropertyId && !nodeProperties.contains( indexPropertyId ) )
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public interface SchemaMatchAction<SUPPLIER extends SchemaDescriptor.Supplier, EXCEPTION extends Exception>
```
MishaDemianenko: Why this can't be a `Consumer` interface instead of custom `Action` thingy?

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
```diff
@@ -70,139 +73,98 @@ public void check( NodeRecord record,
...
                 {
-                    verifyNodeCorrectlyIndexedUniquely( nodeId, property.getKeyIndexId(), propertyValue, engine,
-                            indexRule, reader );
+                    properties = propertyReader.propertyBlocks( propertyRecs );
+                    nodePropertyIds = propertyIds( properties );
                 }
-                else
+
+                int[] indexPropertyIds = indexRule.schema().getPropertyIds();
+                if ( nodeHasSchemaProperties( nodePropertyIds, indexPropertyIds, NO_SUCH_PROPERTY_KEY ) )
```
MishaDemianenko: why do we need to pass NO_SUCH_PROPERTY_KEY here?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/LookupFilter.java
```diff
@@ -45,13 +45,40 @@
...
+                Arrays.stream( predicates )
+                        .filter( LookupFilter::isNumericPredicate )
+                        .toArray( IndexQuery[]::new );
+
+        if ( numericPredicates.length > 0 )
         {
-            return PrimitiveLongCollections.filter( indexedNodeIds,
-                    new LookupBasedExactMatchPredicate( accessor, propertyKeyId,
-                            value ) );
+            LongPredicate combinedPredicate = nodeId ->
```
MishaDemianenko, line 64: `Predicates.all`?

MishaDemianenko: the same in other place in this class

MishaDemianenko (changes requested):
Couple of comments, also would be interesting to know what performance impact does this change brings

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
fickludd: Ah, you're right! If we get that exception someone messed up the index implementation, and the only thing to do is likely to crash.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/LookupFilter.java
fickludd: I tried to solve this with `Predicates.all` as you suggest, but I couldn't make it better. This is because we need to fetch the property value for every predicate which means that we would need to map them and create lambdas or something.

MishaDemianenko:
@fickludd what is the status of this PR?

fickludd:
@MishaDemianenko Oh, you should have another go at it, I tried to address all comments. It keeps failing on some docs-thing that I don't think I caused.

community/consistency-check/src/main/java/org/neo4j/consistency/report/ConsistencyReport.java
```diff
@@ -184,13 +184,13 @@ void forCounts( CountsEntry countsEntry,
...
-        void notIndexed( IndexRule index, Object propertyValue );
+        void notIndexed( IndexRule index, Object[] propertyValues );
 
         @Documented( "This node was found in the expected index, although multiple times" )
-        void indexedMultipleTimes( IndexRule index, Object propertyValue, long count );
+        void indexedMultipleTimes( IndexRule index, Object[] propertyValues, long count );
 
-        @Documented( "There is another node in the unique index with the same property value." )
-        void uniqueIndexNotUnique( IndexRule index, Object propertyValue, long duplicateNodeId );
+        @Documented( "There is another node in the unique index with the same property values." )
```
MishaDemianenko: what if we will say `value(s)` instead since it will work for single property index as well?

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
```diff
@@ -235,4 +195,67 @@ private void checkProperty( NodeRecord record,
...
+        PrimitiveLongIterator indexedNodeIds;
+        try
+        {
+            indexedNodeIds = reader.query( query );
+        }
+        catch ( IndexNotApplicableKernelException e )
+        {
+            throw new RuntimeException( format(
+                    "Cannot continue %s, as index provider does not support exact composite query: %s",
+                    this.getClass().getSimpleName(), Arrays.toString( query ) ), e );
```
MishaDemianenko: why do we need to print this.getClass().getSimpleName() ? it's not like we can't continue only this check right?
would it be more reasonable to continue check when we are in situation like that and report the failure?

community/consistency-check/src/test/java/org/neo4j/consistency/checking/full/FullCheckIntegrationTest.java
```diff
@@ -1057,12 +1092,30 @@ protected void transactionData( GraphStoreFixture.TransactionDataBuilder tx,
...
         byte[] data = rule.serialize();
-        ReusableRecordsAllocator dynamicRecordAllocator = new ReusableRecordsAllocator( data.length, records );
+        DynamicRecordAllocator dynamicRecordAllocator =
+                new ReusableRecordsCompositeAllocator( records, schemaAllocator );
         Collection<DynamicRecord> result = new ArrayList<>();
         AbstractDynamicStore.allocateRecordsFromBytes( result, data, dynamicRecordAllocator );
         return result;
     }
 
+    private static DynamicRecordAllocator schemaAllocator = new DynamicRecordAllocator()
```
MishaDemianenko: please do not declare fields in a first suitable place Move it to the top to other fields

community/kernel/src/main/java/org/neo4j/kernel/impl/api/LookupFilter.java
```diff
@@ -45,13 +47,43 @@
...
+        {
+            LongPredicate combinedPredicate = nodeId ->
+            {
+                try
+                {
+                    for ( IndexQuery predicate : numericPredicates )
+                    {
+                        int propertyKeyId = predicate.propertyKeyId();
+                        Property property = accessor.getProperty( nodeId, propertyKeyId );
+                        if ( property.isDefined() && !predicate.test( ((DefinedProperty)property).value() ) )
```
MishaDemianenko: why do we need this check for defined property?

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
```diff
@@ -235,4 +195,67 @@ private void checkProperty( NodeRecord record,
...
+        }
+
+        indexedNodeIds = LookupFilter.exactIndexMatches( propertyReader, indexedNodeIds, query );
+        return indexedNodeIds;
+    }
+
+    public static boolean nodeHasSchemaProperties(
```
MishaDemianenko: this method can be private now right?

MishaDemianenko (changes requested):
Have couple of comments, also would be  nice to know how introduced changes affect performance of CC, since that is already quite critical part and we should not do that worse then it is now.
Another point would be super nice if you will include in PR description all introduced checks for composite indexes(basically what we are checking for them)

fickludd:
@MishaDemianenko addressed comments, it's yours again. Also manually verified that the time taken to run `time neo4j-admin check-consistency ...` on a 400 Mb LDBC database is the same on this branch as a recent `3.2.0-alpha07` build.

MishaDemianenko:
@fickludd looks good, only need rebase , will merge it on green

MishaDemianenko (approved):

PR #9000, opened 2017-03-10
Introduce CrashGenCleaner

burqen:
For the time being there's an issue where update that come in before a crash
may be applied in a different order than when they get recovered.
This may have structural changes be different during recovery than what they were
during normal application. The result of this is that there may be CRASH pointers left
hanging in the tree. A problem with CRASH pointers is that they can only be recognized
while doing recovery, where they are recognized from their generation being STABLE < gen < UNSTABLE,
i.e. recovery creates this gap to be able to distinguish crashed from fixed pointers.
After recovery is completed and the next checkpoint/close happens that gap is closed
and any remaining CRASH pointers cannot be recognized as such anymore.

The temporary solution right now is to scan through all tree nodes, as performant as possible,
and zero out all CRASH pointers.

NOTE: Github display commits in wrong order, real order is
```
1cbaaba Introduce CrashGenCleaner
18d3c9c Add test for CrashGenCleaner
c11b59b Add randomized CrashGenCleaner test
e75e181 Different key contention in GBPTreeRecoveryTest
4999ae7 Fix complete recovery being triggered to late because of intermediate force
```

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
@@ -1,0 +1,245 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.index.internal.gbptree;
+
+import java.io.IOException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.neo4j.index.internal.gbptree.GBPTree.Monitor;
+import org.neo4j.io.pagecache.PageCursor;
+import org.neo4j.io.pagecache.PagedFile;
+
+import static java.lang.System.currentTimeMillis;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+import static org.neo4j.helpers.Exceptions.launderedException;
+
+/**
+ * Scans the entire tree and checks all GSPPs, replacing all CRASH gen GSPs with zeros.
+ */
+class CrashGenCleaner
+{
+    private final PagedFile pagedFile;
+    private final TreeNode<?,?> treeNode;
+    private final long lowTreeNodeId;
+    private final long highTreeNodeId;
+    private final long stableGeneration;
+    private final long unstableGeneration;
+    private final Monitor monitor;
+    private final long maxKeyCount;
+
+    public CrashGenCleaner( PagedFile pagedFile, TreeNode<?,?> treeNode, long lowTreeNodeId, long highTreeNodeId,
+            long stableGeneration, long unstableGeneration, Monitor monitor )
+    {
+        this.pagedFile = pagedFile;
+        this.treeNode = treeNode;
+        this.lowTreeNodeId = lowTreeNodeId;
+        this.highTreeNodeId = highTreeNodeId;
+        this.stableGeneration = stableGeneration;
+        this.unstableGeneration = unstableGeneration;
+        this.monitor = monitor;
+        this.maxKeyCount = treeNode.internalMaxKeyCount();
+    }
+
+    // === Methods about the execution and threading ===
+
+    public void clean() throws IOException
+    {
+        assert unstableGeneration > stableGeneration;
+        assert unstableGeneration - stableGeneration > 1;
+
+        long startTime = currentTimeMillis();
+        int threads = Runtime.getRuntime().availableProcessors();
+        ExecutorService executor = Executors.newFixedThreadPool( threads );
+        AtomicLong nextId = new AtomicLong( lowTreeNodeId);
+        AtomicReference<Throwable> error = new AtomicReference<>();
+        AtomicInteger cleanedPointers = new AtomicInteger();
+        for ( int i = 0; i < threads; i++ )
+        {
+            executor.submit( cleaner( nextId, error, cleanedPointers ) );
+        }
+        executor.shutdown();
+
+        try
+        {
+            long lastProgression = nextId.get();
+            // Have max no-progress-timeout quite high to be able to cope with huge
+            // I/O congestion spikes w/o failing in vain.
+            while ( !executor.awaitTermination( 30, SECONDS ) )
+            {
+                if ( lastProgression == nextId.get() )
+                {
+                    // No progression at all, abort?
+                    error.compareAndSet( null, new IOException( "No progress, so forcing abort" ) );
+                }
+                lastProgression = nextId.get();
+            }
+        }
+        catch ( InterruptedException e )
+        {
+            Thread.currentThread().interrupt();
+        }
+
+        Throwable finalError = error.get();
+        if ( finalError != null )
+        {
+            throw launderedException( IOException.class, finalError );
+        }
+
+        long endTime = currentTimeMillis();
+        monitor.recoveryCompleted( highTreeNodeId - lowTreeNodeId, cleanedPointers.get(), endTime - startTime );
+    }
+
+    private Runnable cleaner( AtomicLong nextId, AtomicReference<Throwable> error, AtomicInteger cleanedPointers )
+    {
+        return () ->
+        {
+            try ( PageCursor cursor = pagedFile.io( 0, PagedFile.PF_SHARED_READ_LOCK );
+                    PageCursor writeCursor = pagedFile.io( 0, PagedFile.PF_SHARED_WRITE_LOCK ) )
+            {
+                while ( nextId.get() < highTreeNodeId )
+                {
+                    // Do a batch of max 1000
+                    int batchCount = 0;
+                    for ( ; batchCount < 1_000; batchCount++ )
+                    {
+                        long id = nextId.getAndIncrement();
+                        if ( id >= highTreeNodeId )
+                        {
+                            break;
+                        }
+
+                        // TODO: we should try to figure out if the disk backing this paged file
+                        // is good at concurrent random reads (i.e. if it's an SSD).
+                        // If it's not it'd likely be beneficial to let other threads wait until
+                        // goTo completes before letting others read. This will have reads be
+                        // 100% sequential.
+                        PageCursorUtil.goTo( cursor, "clean", id );
+
+                        if ( hasCrashedGSPP( treeNode, cursor ) )
+                        {
+                            writeCursor.next( cursor.getCurrentPageId() );
+                            cleanTreeNode( treeNode, writeCursor, cleanedPointers );
+                        }
+                    }
+
+                    // Check error status after a batch, to reduce volatility overhead.
+                    // Is this over thinking things? Perhaps
+                    if ( error.get() != null )
+                    {
+                        break;
+                    }
+                }
+            }
+            catch ( Throwable e )
+            {
+                error.compareAndSet( null, e );
+            }
+        };
+    }
+
+    // === Methods about checking if a tree node has crashed pointers ===
+
+    private boolean hasCrashedGSPP( TreeNode<?,?> treeNode, PageCursor cursor ) throws IOException
+    {
+        boolean isTreeNode;
+        int keyCount;
+        do
+        {
+            isTreeNode = TreeNode.nodeType( cursor ) == TreeNode.NODE_TYPE_TREE_NODE;
+            keyCount = treeNode.keyCount( cursor );
+        }
+        while ( cursor.shouldRetry() );
+
+        if ( !isTreeNode )
+        {
+            return false;
+        }
+
+        boolean hasCrashed;
+        do
+        {
+            hasCrashed =
+                    hasCrashedGSPP( cursor, TreeNode.BYTE_POS_NEWGEN ) ||
+                    hasCrashedGSPP( cursor, TreeNode.BYTE_POS_LEFTSIBLING ) ||
+                    hasCrashedGSPP( cursor, TreeNode.BYTE_POS_RIGHTSIBLING );
+
+            if ( !hasCrashed && TreeNode.isInternal( cursor ) )
+            {
+                for ( int i = 0; i <= keyCount && i <= maxKeyCount && !hasCrashed; i++ )
+                {
+                    hasCrashed |= hasCrashedGSPP( cursor, treeNode.childOffset( i ) );
+                }
+            }
+        }
+        while ( cursor.shouldRetry() );
+        return hasCrashed;
+    }
+
+    private boolean hasCrashedGSPP( PageCursor cursor, int gsppOffset )
+    {
+        return hasCrashedGSP( cursor, gsppOffset ) || hasCrashedGSP( cursor, gsppOffset + GenSafePointer.SIZE );
+    }
+
+    private boolean hasCrashedGSP( PageCursor cursor, int offset )
+    {
+        cursor.setOffset( offset );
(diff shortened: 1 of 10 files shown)


community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
```diff
@@ -0,0 +1,245 @@
...
+    private final PagedFile pagedFile;
+    private final TreeNode<?,?> treeNode;
+    private final long lowTreeNodeId;
+    private final long highTreeNodeId;
+    private final long stableGeneration;
+    private final long unstableGeneration;
+    private final Monitor monitor;
+    private final long maxKeyCount;
+
+    public CrashGenCleaner( PagedFile pagedFile, TreeNode<?,?> treeNode, long lowTreeNodeId, long highTreeNodeId,
```
MishaDemianenko: this constructor can be package local since class is package local

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
```diff
@@ -0,0 +1,245 @@
...
+            hasCrashed =
+                    hasCrashedGSPP( cursor, TreeNode.BYTE_POS_NEWGEN ) ||
+                    hasCrashedGSPP( cursor, TreeNode.BYTE_POS_LEFTSIBLING ) ||
+                    hasCrashedGSPP( cursor, TreeNode.BYTE_POS_RIGHTSIBLING );
+
+            if ( !hasCrashed && TreeNode.isInternal( cursor ) )
+            {
+                for ( int i = 0; i <= keyCount && i <= maxKeyCount && !hasCrashed; i++ )
+                {
+                    hasCrashed |= hasCrashedGSPP( cursor, treeNode.childOffset( i ) );
```
MishaDemianenko: line `hasCrashed |= hasCrashedGSPP( cursor, treeNode.childOffset( i ) );` can be changed to `hasCrashed = hasCrashedGSPP( cursor, treeNode.childOffset( i ) );` since we continue only while `!hasCrashed`

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
```diff
@@ -0,0 +1,245 @@
...
+                    int batchCount = 0;
+                    for ( ; batchCount < 1_000; batchCount++ )
+                    {
+                        long id = nextId.getAndIncrement();
+                        if ( id >= highTreeNodeId )
+                        {
+                            break;
+                        }
+
+                        // TODO: we should try to figure out if the disk backing this paged file
```
MishaDemianenko: i don't think this comment is useful so far as a comment since we will never gonna come to this place again and introduce check for SSD or anything else like that, if we want to introduce something like that we should have a card instead

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
```diff
@@ -0,0 +1,245 @@
...
+    private boolean hasCrashedGSPP( TreeNode<?,?> treeNode, PageCursor cursor ) throws IOException
+    {
+        boolean isTreeNode;
+        int keyCount;
+        do
+        {
+            isTreeNode = TreeNode.nodeType( cursor ) == TreeNode.NODE_TYPE_TREE_NODE;
+            keyCount = treeNode.keyCount( cursor );
+        }
+        while ( cursor.shouldRetry() );
```
MishaDemianenko, line 167: while this will be correct still in case of should retry would prefer if we will explicitly set cursor position on each iteration

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
```diff
@@ -0,0 +1,245 @@
...
+
+        if ( !isTreeNode )
+        {
+            return false;
+        }
+
+        boolean hasCrashed;
+        do
+        {
+            hasCrashed =
```
MishaDemianenko, line 178: in this look we will lose correct cursor position in case of shouldRetry

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private void cleanCrashedGSPP( PageCursor cursor, int gsppOffset, AtomicInteger cleanedPointers )
+    {
+        cleanCrashedGSP( cursor, gsppOffset, cleanedPointers );
+        cleanCrashedGSP( cursor, gsppOffset + GenSafePointer.SIZE, cleanedPointers );
+    }
+
+    private void cleanCrashedGSP( PageCursor cursor, int gspOffset, AtomicInteger cleanedPointers )
+    {
+        if ( hasCrashedGSP( cursor, gspOffset ) )
```
MishaDemianenko: this place missing should retry cursor loop

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    private Runnable cleaner( AtomicLong nextId, AtomicReference<Throwable> error, AtomicInteger cleanedPointers )
+    {
+        return () ->
+        {
+            try ( PageCursor cursor = pagedFile.io( 0, PagedFile.PF_SHARED_READ_LOCK );
+                    PageCursor writeCursor = pagedFile.io( 0, PagedFile.PF_SHARED_WRITE_LOCK ) )
+            {
+                while ( nextId.get() < highTreeNodeId )
```
MishaDemianenko: it looks like if we will pass ranges if ids to check we can speed it up, since we will remove common contention point which is nextId currently, plus we will have ranges of ids in one thread that will potentially make it faster to process those because of locality

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -830,6 +842,32 @@ public void prepareForRecovery() throws IOException
         pagedFile.flushAndForce();
     }
 
+    /**
+     * Must be called after all recovered updates have been applied and before doing further work,
+     * be it reading or writing.
+     *
+     * @throws IOException on {@link PageCache} error.
+     */
+    public void completeRecovery() throws IOException
```
MishaDemianenko: this method name is a bit misleading since it's invoked from 'maybeCompleteRecovery' that invoked on start or during flush. So it's not always have something to do with recovery. What if we will call it clean for example?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -830,6 +842,32 @@ public void prepareForRecovery() throws IOException
...
+        // i.e. recovery creates this gap to be able to distinguish crashed from fixed pointers.
+        // After recovery is completed and the next checkpoint/close happens that gap is closed
+        // and any remaining CRASH pointers cannot be recognized as such anymore.
+        //
+        // The temporary solution right now is to scan through all tree nodes, as performant as possible,
+        // and zero out all CRASH pointers.
+
+        long generation = this.generation;
+        new CrashGenCleaner( pagedFile, bTreeNode, IdSpace.MIN_TREE_NODE_ID,freeList.lastId() + 1,
+                stableGeneration( generation ), unstableGeneration( generation ), monitor ).clean();
```
MishaDemianenko, line 868: what will happen with recovery if we will fail on phase of tree cleaning?

community/index/src/test/java/org/neo4j/index/internal/gbptree/CrashGenCleanerTest.java
```diff
@@ -0,0 +1,464 @@
...
+    private final Layout<MutableLong,MutableLong> layout = new SimpleLongLayout();
+    private final CorruptableTreeNode corruptableTreeNode = new CorruptableTreeNode( PAGE_SIZE, layout );
+    private final int oldStableGen = 9;
+    private final int stableGen = 10;
+    private final int unstableGen = 12;
+    private final int crashGen = 11;
+    private final int firstChildPos = 0;
+    private final int middleChildPos = corruptableTreeNode.internalMaxKeyCount() / 2;
+    private final int lastChildPos = corruptableTreeNode.internalMaxKeyCount();
+    private final List<PageCorruption> possibleCorruptionsInInternal = new ArrayList<>( Arrays.asList(
```
MishaDemianenko: why do we need to wrap Arrays.asList to ArrayList?

community/index/src/test/java/org/neo4j/index/internal/gbptree/CrashGenCleanerTest.java
```diff
@@ -0,0 +1,464 @@
...
+            crashed( leftSibling() ),
+            crashed( rightSibling() ),
+            crashed( newGen() ),
+            crashed( child( firstChildPos ) ),
+            crashed( child( middleChildPos ) ),
+            crashed( child( lastChildPos ) )
+    ) );
+    private final List<PageCorruption> possibleCorruptionsInLeaf = new ArrayList<>( Arrays.asList(
```
MishaDemianenko: the same here as above

community/index/src/test/java/org/neo4j/index/internal/gbptree/CrashGenCleanerTest.java
```diff
@@ -0,0 +1,464 @@
...
+import org.neo4j.test.rule.fs.FileSystemRule;
+
+import static java.nio.file.StandardOpenOption.CREATE;
+import static java.nio.file.StandardOpenOption.DELETE_ON_CLOSE;
+import static org.junit.Assert.assertEquals;
+import static org.neo4j.test.rule.PageCacheRule.config;
+
+public class CrashGenCleanerTest
+{
+    private FileSystemRule fileSystemRule = new DefaultFileSystemRule();
```
MishaDemianenko, line 51: what if we will use file system and page cache that are adversarial?

MishaDemianenko (changes requested):
Looks good in general, but have couple of comments

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
burqen: 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
burqen: 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
burqen: 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
burqen: `hasCrashedGSPP` takes exact offset as argument and given offsets are absolute so we don't rely on position from previous call when progressing with next read.

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
burqen: 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
burqen: The cursor is always a write cursor here, so the `shouldRetry` is not needed.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: We are left with a tree that still has crash-pointers in it. As long as we do not checkpoint the tree (making the crash pointers impossible to separate from stable pointers) we can just try to do recovery again. And it should actually be fine to retry only the clean phase if we already replayed the log.

What actually will though is a `UnderlyingStorageException` which will result in kernel panic (I think?). On next startup it should be fine to retry recovery.
This is under the assumption that no checkpoint is performed if we panic(?).

community/index/src/test/java/org/neo4j/index/internal/gbptree/CrashGenCleanerTest.java
burqen: `PageCacheRule` default behaviour is to use random inconsistent reads
https://github.com/neo4j/neo4j/blob/3.2/community/io/src/test/java/org/neo4j/test/rule/PageCacheRule.java#L186

I'm not sure about `AdversarialFileSystemAbstraction`, I don't think we should try to take care of any file system failure since there is nothing we can do other than try to start over again. Right?

community/index/src/test/java/org/neo4j/index/internal/gbptree/CrashGenCleanerTest.java
burqen: That is a very good question 👍

community/index/src/test/java/org/neo4j/index/internal/gbptree/CrashGenCleanerTest.java
burqen: 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
tinwelint: Definitely true about reducing contention on `nextId`. Each id represent one page so I wouldn't expect improvement in access locality. Anyways 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: I think it should signal start and end of recovery though.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: Actually, checkpoint is performed in the event of kernel panic. This is a problem, but lets consider it to be outside the scope of this PR.
A card has been created.

burqen:
@MishaDemianenko Comments addressed

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
```diff
@@ -0,0 +1,243 @@
...
+            long stableGeneration, long unstableGeneration, Monitor monitor )
+    {
+        this.pagedFile = pagedFile;
+        this.treeNode = treeNode;
+        this.lowTreeNodeId = lowTreeNodeId;
+        this.highTreeNodeId = highTreeNodeId;
+        this.availableProcessors = Runtime.getRuntime().availableProcessors();
+        this.batchSize = // Each processor will get roughly 100 batches each
+                min( 1000, max( 10, (highTreeNodeId - lowTreeNodeId) / (100 * availableProcessors) ) );
+        System.out.println( "Number of processors: " + availableProcessors + ", number of pages: " +
```
MishaDemianenko: wow wow wow system.out?

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
MishaDemianenko: lets add a comment about that, for future maintainers to know about that assumption

community/index/src/test/java/org/neo4j/index/internal/gbptree/CrashGenCleanerTest.java
MishaDemianenko: ok, agree

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
MishaDemianenko: that was my question basically, so perfect - addressed!

MishaDemianenko:
@burqen 2 small notes

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
burqen: 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/CrashGenCleaner.java
burqen: Good catch, thank you

MishaDemianenko (approved):

PR #9001, opened 2017-03-10
Made ReadOperations safe to inject into procedures.

eebus:


community/kernel/src/main/java/org/neo4j/kernel/impl/factory/DataSourceModule.java
```diff
@@ -373,6 +374,8 @@ private Procedures setupProcedures( PlatformModule platform, EditionModule editi
         Guard guard = platform.dependencies.resolveDependency( Guard.class );
         procedures.registerComponent( ProcedureTransaction.class, new ProcedureTransactionProvider(), true );
         procedures.registerComponent( TerminationGuard.class, new TerminationGuardProvider( guard ), true );
+        procedures.registerComponent( ReadOperations.class,
+                ( ctx ) -> ctx.get( KERNEL_TRANSACTION ).acquireStatement().readOperations(), true );
```
lutovich, line 378: How will the acquired statement be closed?

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/DataSourceModule.java
```diff
@@ -373,6 +374,8 @@ private Procedures setupProcedures( PlatformModule platform, EditionModule editi
         Guard guard = platform.dependencies.resolveDependency( Guard.class );
         procedures.registerComponent( ProcedureTransaction.class, new ProcedureTransactionProvider(), true );
         procedures.registerComponent( TerminationGuard.class, new TerminationGuardProvider( guard ), true );
+        procedures.registerComponent( ReadOperations.class,
```
MishaDemianenko, line 377: Why do we injecting read operations here?

PR #9007, opened 2017-03-10
Cluster formation decision depends on initial_hosts contents

digitalstain:
This change makes it so only instances listed in initial_hosts are
 considered for purposes of forming a cluster. This allows mismatching
 initial_hosts contents to be manipulated to make it harder for
 instances to start a cluster, for example slave_only instances to
 depend on the presence of a master but the master to be allowed to
 form the cluster regardless of the presence of other instances.
The functionality is achieved with the introduction of a
 (backwards compatible) discovery header in the configurationRequest
 messages which keeps track of which instance has contacted which
 during the initial discovery round. Since sending of these messages
 depends only on the initial_hosts content, this effectively allows
 for filtering of configurationRequests based on initial_host content.

### DIFF
--- a/enterprise/cluster/src/main/java/org/neo4j/cluster/com/message/Message.java
+++ b/enterprise/cluster/src/main/java/org/neo4j/cluster/com/message/Message.java
@@ -83,7 +83,6 @@ public class Message<MESSAGETYPE extends MessageType>
         return timeout;
     }
 
-
     // Standard headers
     public static final String CONVERSATION_ID = "conversation-id";
     public static final String CREATED_BY = "created-by";
@@ -91,6 +90,9 @@ public class Message<MESSAGETYPE extends MessageType>
     public static final String FROM = "from";
     public static final String TO = "to";
     public static final String INSTANCE_ID = "instance-id";
+    // Should be present only in configurationRequest messages. Value is a comma separated list of instance ids.
+    // Added in 3.0.9.
+    public static final String DISCOVERED = "discovered";
 
     private MESSAGETYPE messageType;
     private Object payload;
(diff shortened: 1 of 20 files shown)


enterprise/cluster/src/main/java/org/neo4j/cluster/com/message/Message.java
```diff
@@ -136,16 +138,26 @@ public boolean isInternal()
...
         }
         return value;
     }
 
-    public <MESSAGETYPE extends MessageType> Message<MESSAGETYPE> copyHeadersTo( Message<MESSAGETYPE> message,
-                                                                                 String... names )
+    public String getHeader( String name, String defaultValue )
+    {
+        String value = headers.get( name );
+        if ( value == null )
```
digitalstain, line 152: use getOrDefault perhaps?

enterprise/cluster/src/main/java/org/neo4j/cluster/protocol/atomicbroadcast/multipaxos/context/ClusterContextImpl.java
```diff
@@ -334,9 +352,64 @@ public ObjectInputStreamFactory getObjectInputStreamFactory()
...
+    }
+
+    @Override
+    public String generateDiscoveryHeader()
+    {
+        /*
+         * Maps the keyset of contacting instances from InstanceId to strings, collects them in a Set and joins them
+         * in a string with the appropriate separator
+         */
+        return String.join( DISCOVERY_HEADER_SEPARATOR, contactingInstances.keySet().stream().map( InstanceId::toString ).collect( Collectors.toSet() ) );
```
digitalstain, line 389: use a joining collector instead?

lutovich (approved):

PR #9011, opened 2017-03-13
Switch on/off Multi Data Center with a single config line

jimwebber:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
```diff
@@ -45,8 +45,9 @@
 import org.neo4j.causalclustering.identity.MemberId;
 import org.neo4j.causalclustering.load_balancing.LoadBalancingPluginLoader;
 import org.neo4j.causalclustering.load_balancing.LoadBalancingProcessor;
-import org.neo4j.causalclustering.load_balancing.procedure.GetServersProcedureV1;
-import org.neo4j.causalclustering.load_balancing.procedure.GetServersProcedureV2;
+import org.neo4j.causalclustering.load_balancing.procedure.GetServersProcedureForSingleDC;
+import org.neo4j.causalclustering.load_balancing.procedure.LegacyGetServersProcedure;
+import org.neo4j.causalclustering.load_balancing.procedure.GetServersProcedureForMultiDC;
```
martinfurmanski, line 50: I don't like the renames from GetServersV1/V2 to specialized versions. The V1 and V2 are external interfaces which shall not change and they are exposed under exactly those names. They represent the contract and one could imagine a V3 in the future, for example to include physical rule tables or metrics. I would rather just keep that scheme for clarity.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/procedure/GetServersProcedureForSingleDC.java
```diff
@@ -0,0 +1,172 @@
...
+import static org.neo4j.kernel.api.proc.ProcedureSignature.procedureSignature;
+
+/**
+ * Returns endpoints and their capabilities.
+ *
+ * GetServersV2 extends upon V1 by allowing a client context consisting of
+ * key-value pairs to be supplied to and used by the concrete load
+ * balancing strategies.
+ */
+public class GetServersProcedureForSingleDC implements CallableProcedure
```
martinfurmanski, line 68: This duplicates a lot of code unnecessarily. Perhaps the toggle should instead affect which LoadBalancingPlugin gets instantiated? The question probably begins with answering what the single behaviour we want is, e.g. is it server_policies with simple/default default policy? Is it another plugin? Is server_policies the multi-dc? Do we even allow other plugins without the toggle enabled?

martinfurmanski (dismissed):

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/procedure/GetServersProcedureForSingleDC.java
jimwebber: We preferred the toggle to be at the edition level since that's where everything gets wired together. Ultimately the Legacy get servers will be replaced by get servers for single DC and the code duplication is removed.

martinfurmanski: Perhaps at some point in time the Legacy/V1 will no longer be supported, but until then it is in fact supported, and used by 1.1 and 1.2 java drivers for example. Naming it with legacy seems very premature and also not really helpful. V1/V2/... should be the way to keep track of this externally published interface.

PR #9012, opened 2017-03-13
Added Data Center-Local Catchup Strategy for Multi-DC Deployments.

jimwebber:
Adding a Data Center-Local upstream strategy.
A sensible out of the box plugin for allowing read replicas to catch up from named others within their locality.
Log the loaded strategies and their order of precedence.
Does not catchup from self.


mneedham:
Test failures look unrelated

PR #9013, opened 2017-03-13
Ability to auto-select label index based on existing store

tinwelint:
Introducing LabelIndex.AUTO which is intended to be used for tooling
where it's convenient to not have a say in which label index to use,
but let database select which ever label index is present already.

This was driven by a backup scenario where the backup client would
select a different label index than the source. This would work
conceptually (because it would have been rebuilt), but proved to be
troublesome by means of page cache actions for moving files after
temporary store had been started and stopped. In this case the
original label index store would have been deleted and making the
following move action invalid.

Letting the backup client internally use AUTO label index selection
solves this. It's still of course possible to configure another
label index when restoring from this backup, which is separate from
this problem of doing the backup.

In this vein, even consistency checker uses AUTO.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -508,8 +508,20 @@ public class GraphDatabaseSettings implements LoadableConfig
 
     public enum LabelIndex
     {
+        /**
+         * Native label index. Generally the best option.
+         */
         NATIVE,
-        LUCENE;
+
+        /**
+         * Label index backed by Lucene.
+         */
+        LUCENE,
+
+        /**
+         * Selects which ever label index is present in a store, or the default (NATIVE) if no label index present.
+         */
+        AUTO;
     }
 
     @Description( "Backend to use for label --> nodes index" )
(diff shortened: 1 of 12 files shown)


community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -508,8 +508,20 @@
 
     public enum LabelIndex
     {
+        /**
```
MishaDemianenko, line 511: Why our label_index setting is `Setting<String>` instead of `Setting<LabelIndex>`?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -508,8 +508,20 @@
...
+
+        /**
+         * Label index backed by Lucene.
+         */
+        LUCENE,
+
+        /**
+         * Selects which ever label index is present in a store, or the default (NATIVE) if no label index present.
+         */
+        AUTO;
```
MishaDemianenko, line 524: Why not to use AUTO as default property value and then do fallback to NATIVE in case if none of the indexes exist?

MishaDemianenko (commented):
Couple of small comments

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
tinwelint: That would keep existing databases on LUCENE, which isn't desirable.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
tinwelint: I don't understand your concern

tinwelint:
Merging because it's causing build problems, which this PR fixes

PR #9014, opened 2017-03-13
Backport hz robustness

martinfurmanski:
changelog: [causal clustering] Improve robustness of hazelcast usage.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/RobustHazelcastWrapper.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/RobustHazelcastWrapper.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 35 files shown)


martinfurmanski:
This is a backport, forward merge should be minimal or null.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/RobustHazelcastWrapper.java
```diff
@@ -0,0 +1,101 @@
...
+        catch ( com.hazelcast.core.HazelcastInstanceNotActiveException e )
+        {
+            invalidateConnection();
+            throw new HazelcastInstanceNotActiveException( e );
+        }
+    }
+
+    synchronized void perform( Consumer<HazelcastInstance> operation ) throws HazelcastInstanceNotActiveException
+    {
+        HazelcastInstance hzInstance = tryEnsureConnection();
```
mneedham, line 89: Is this bit intentionally not within the try block?

mneedham:
All looks good. One comment about a `try {}` block but that's all

martinfurmanski:
@mneedham I don't see any comment about a try block

mneedham (changes requested):
blah

mneedham:
@martinfurmanski  weird, it made me finish my review before it'd be visible. Hopefully you can see it now

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/RobustHazelcastWrapper.java
martinfurmanski: Yes it is intentional, because I don't believe HazelcastClient.newHazelcastClient (in HazelcastConnector) can throw com.hazelcast.core.HazelcastInstanceNotActiveException. It should only be thrown when you use the instance for an operation and not on acquirement. So even though it would be harmless to include it in the try-block, it would also be "unclean" from the perspective of expectations for the code.

If in the future we find that in fact it can throw com.hazelcast.core.HazelcastInstanceNotActiveException then HazelcastConnector#connectToHazelcast should catch and declare the checked HazelcastInstanceNotActiveException.

Note that all the usage of Hazelcast is anyway wrapped in robust periodic jobs which catch all exceptions, but I prefer to separate the expectations on hazelcast into its own wrapper for clarity.

mneedham (approved):

PR #9015, opened 2017-03-13
Improved query logging

OliviaYtterbrink:
When changing password through procedures the password should not be logged in the query log.


enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerIT.java
```diff
@@ -324,6 +328,202 @@ public void queryLogRotation() throws Exception
...
+        finally
+        {
+            database.shutdown();
+        }
+
+        List<String> logLines = readAllLines( logFilename );
+        assertEquals( 1, logLines.size() );
+        assertThat( logLines.get( 0 ),
+                containsString(  "CALL dbms.security.changePassword(******)") ) ;
+        assertThat( logLines.get( 0 ), containsString( neo.subject().username() ) );
```
fickludd: Missing assert that the new password is not there! Also for the other tests.

enterprise/query-logging/src/main/java/org/neo4j/kernel/impl/query/QueryLoggerKernelExtension.java
```diff
@@ -183,9 +213,13 @@ private String logEntry( long time, ExecutingQuery query )
             }
         }
 
-        @SuppressWarnings( "unchecked" )
         private static String mapAsString( Map<String,Object> params )
         {
+            return mapAsString( params, "" );
+        }
+
+        private static String mapAsString( Map<String, Object> params, String obfuscate )
```
fickludd: What about multiple passwords in params?

enterprise/query-logging/src/main/java/org/neo4j/kernel/impl/query/QueryLoggerKernelExtension.java
```diff
@@ -137,6 +139,11 @@ private Lifecycle createEmptyAdapter()
         private final long thresholdMillis;
         private final boolean logQueryParameters;
 
+        private static final Pattern PASSWORD_PATTERN = Pattern.compile(
+                "(?:(?i)call)\\s+dbms(?:\\.security)?\\.change(?:User)?Password\\(" +
+                        "(?:\\s*(?:'(?:(?<=\\\\)'|[^'])*'|\"(?:(?<=\\\\)\"|[^\"])*\"|\\$\\w*|\\{\\w*\\})\\s*,)?" +
+                        "\\s*('(?:(?<=\\\\)'|[^'])*'|\"(?:(?<=\\\\)\"|[^\"])*\"|\\$\\w*|\\{\\w*\\})\\s*\\)" );
```
fickludd: I have not attempted to understand this... I would hope that it was possible to split this into separate paths with `|` or something, and maybe put a comment indicating happy path match?

fickludd (changes requested):
With the very complicated regex matching, I definitely think there should be more unit tests on the QueryLogger. Different white space combinations, `null`s, multiple change passwords with params could all be tested.

Maybe one/some of the ITs could be moved to unit test level also. That would help build times. 

If 3.1.3 is hanging on this, we can consider merging this for 3.1.3 and then increasing test coverage and iron out eventual corner cases. It looks like this should work for all likely cases, so it's of course already a big improvement over 3.1.2.

enterprise/query-logging/src/main/java/org/neo4j/kernel/impl/query/QueryLoggerKernelExtension.java
OliviaYtterbrink: The problem is that the last part, the part that is hard to understand can be mixed and matched in any way you'd like, parameters with literals, literals with `"` or `'`, so splitting into different paths would result in somewhere of 20 or so paths...

enterprise/query-logging/src/main/java/org/neo4j/kernel/impl/query/QueryLoggerKernelExtension.java
```diff
@@ -137,6 +143,14 @@ private Lifecycle createEmptyAdapter()
         private final long thresholdMillis;
         private final boolean logQueryParameters;
 
+        private static final Pattern PASSWORD_PATTERN = Pattern.compile(
+                // call signature
+                "(?:(?i)call)\\s+dbms(?:\\.security)?\\.change(?:User)?Password\\(" +
+                // optional username parameter, in single, double quotes, or parametrized
+                "(?:\\s*(?:'(?:(?<=\\\\)'|[^'])*'|\"(?:(?<=\\\\)\"|[^\"])*\"|[^,]*)\\s*,)?" +
```
fickludd, line 150: Is the final `?` part of the white space on the next line?

enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerTest.java
```diff
@@ -260,6 +276,143 @@ public void shouldLogMetaData() throws Exception
...
+        String inputQuery = "CALL dbms.security.changeUserPassword('neo4j','.changePassword(silly)') " +
+                "CALL dbms.security.changeUserPassword('smith','other$silly') RETURN 1";
+        String outputQuery = "CALL dbms.security.changeUserPassword('neo4j',******) " +
+                "CALL dbms.security.changeUserPassword('smith',******) RETURN 1";
+
+        runAndCheck( inputQuery, outputQuery, emptyMap(), "" );
+    }
+
+    @Test
+    public void shouldNotLogPasswordEvenIfYouDoTwoThingsAtTheSameTimeWithSeveralParms() throws Exception
```
fickludd, line 345: Nitpick: I would put this after the happy path params test :)

fickludd (approved):
Great improvement in testing, this I can merge

enterprise/query-logging/src/main/java/org/neo4j/kernel/impl/query/QueryLoggerKernelExtension.java
OliviaYtterbrink: no, that is just to mark this group as optional

PR #9016, opened 2017-03-13
Fix safe zone id reuse for properties

burqen:
Make sure termination of transactions is checked during lazy
property evaluation by pushing down the statement is open assertion
to cursor level.

During slave pull updates the `TransactionBatchCommitter` will mark transactions falling outside of safe zone for termination. 

Consider this scenario:
```
try ( Transaction tx = db.beginTx )
{
    Object property = db.getNodeById( id ).getProperty( propKey );
    // Act on information from property
    tx.success();
}
```
Until we do `tx.success` we don't know if our transaction has been terminated or not and we could potentially act on inconsistent data (from the dynamic store).

The current workaround for this is to assert that statement is still open before returning. We do this for 
`nodeGetProperty( long nodeId, int propertyKeyId )`
`relationshipGetProperty( long nodeId, int propertyKeyId )`
`nodeGetPropertyKeys( long nodeId )`
`relationshipGetPropertyKeys( long nodeId )`
but not for any other `getProperty/Properties`.
The updated `TerminationOfSlavesDuringPullUpdatesTest` shows that this is a problem and this PR tries to solve it by pushing down the assert open check to the cursor level.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api;
+
+/**
+ * Used to verify that some source of operation is still open when operation is performed.
+ * E.g. verify that transaction is still open when lazily reading property values from store.
+ */
+public interface AssertOpen
+{
+    /**
+     * Assert that source tied to instance is open.
+     * Should throw exception with reason if source is not open.
+     */
+    void assertOpen();
+
+    AssertOpen ALWAYS_OPEN = AlwaysOpen.INSTANCE;
+
+    enum AlwaysOpen implements AssertOpen
+    {
+        INSTANCE
+        {
+            @Override
+            public void assertOpen()
+            {   // no-op
+            }
+        }
+    }
+}
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
@@ -31,16 +31,8 @@ public interface AssertOpen
      */
     void assertOpen();
 
-    AssertOpen ALWAYS_OPEN = AlwaysOpen.INSTANCE;
-
-    enum AlwaysOpen implements AssertOpen
+    AssertOpen ALWAYS_OPEN = () ->
     {
-        INSTANCE
-        {
-            @Override
-            public void assertOpen()
-            {   // no-op
-            }
-        }
-    }
+        // Always open
+    };
 }
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
@@ -147,7 +147,7 @@ public class StateHandlingStatementOperations implements
     @Override
     public Cursor<NodeItem> nodeCursor( KernelStatement statement, long nodeId )
     {
-        Cursor<NodeItem> cursor = statement.getStoreStatement().acquireSingleNodeCursor( nodeId, statement::assertOpen );
+        Cursor<NodeItem> cursor = statement.getStoreStatement().acquireSingleNodeCursor( nodeId, statement );
         if ( statement.hasTxStateWithChanges() )
         {
             return statement.txState().augmentSingleNodeCursor( cursor, nodeId );
@@ -172,7 +172,7 @@ public class StateHandlingStatementOperations implements
     public Cursor<RelationshipItem> relationshipCursor( KernelStatement statement, long relationshipId )
     {
         Cursor<RelationshipItem> cursor = statement.getStoreStatement().acquireSingleRelationshipCursor(
-                relationshipId, statement::assertOpen );
+                relationshipId, statement );
         if ( statement.hasTxStateWithChanges() )
         {
             return statement.txState().augmentSingleRelationshipCursor( cursor, relationshipId );
@@ -183,7 +183,7 @@ public class StateHandlingStatementOperations implements
     @Override
     public Cursor<NodeItem> nodeCursorGetAll( KernelStatement statement )
     {
-        Cursor<NodeItem> cursor = statement.getStoreStatement().nodesGetAllCursor( statement::assertOpen );
+        Cursor<NodeItem> cursor = statement.getStoreStatement().nodesGetAllCursor( statement );
         if ( statement.hasTxStateWithChanges() )
         {
             return statement.txState().augmentNodesGetAllCursor( cursor );
@@ -195,7 +195,7 @@ public class StateHandlingStatementOperations implements
     public Cursor<RelationshipItem> relationshipCursorGetAll( KernelStatement statement )
     {
         Cursor<RelationshipItem> cursor =
-                statement.getStoreStatement().relationshipsGetAllCursor( statement::assertOpen );
+                statement.getStoreStatement().relationshipsGetAllCursor( statement );
         if ( statement.hasTxStateWithChanges() )
         {
             return statement.txState().augmentRelationshipsGetAllCursor( cursor );
@@ -209,7 +209,7 @@ public class StateHandlingStatementOperations implements
         // TODO Filter this properly
         StorageStatement storeStatement = statement.getStoreStatement();
         return storeStatement.acquireIteratorNodeCursor(
-                storeLayer.nodesGetForLabel( storeStatement, labelId ), statement::assertOpen );
+                storeLayer.nodesGetForLabel( storeStatement, labelId ), statement );
     }
 
     @Override
@@ -219,7 +219,7 @@ public class StateHandlingStatementOperations implements
         // TODO Filter this properly
         StorageStatement storeStatement = statement.getStoreStatement();
         IndexReader reader = storeStatement.getIndexReader( index );
-        return storeStatement.acquireIteratorNodeCursor( reader.seek( value ), statement::assertOpen );
+        return storeStatement.acquireIteratorNodeCursor( reader.seek( value ), statement );
     }
 
     @Override
@@ -229,7 +229,7 @@ public class StateHandlingStatementOperations implements
         // TODO Filter this properly
         StorageStatement storeStatement = statement.getStoreStatement();
         IndexReader reader = storeStatement.getIndexReader( index );
-        return storeStatement.acquireIteratorNodeCursor( reader.scan(), statement::assertOpen );
+        return storeStatement.acquireIteratorNodeCursor( reader.scan(), statement );
     }
 
     @Override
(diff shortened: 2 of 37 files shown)


burqen:
Forward merge here: https://github.com/burqen/neo4j/commit/8f12f2bb7198b6d7ae83b9f8cf620d623a126759

chrisvest:
@burqen Does this need a change log label?

burqen:
This is essentially a bug fix and as far as I know no one has run into the problem, so my first thought is 'no'.

tinwelint:
I'd like a better description of exactly what this PR fixes and what the problem was.

tinwelint (changes requested):
I'd like a better description of exactly what this PR fixes and what the problem was.

burqen:
@tinwelint Updated the description. Does it make sense?

tinwelint:
Here's my take on why this solution isn't the "correct" one:

- Injecting Runnable (even some other interface that does something similar, doesn't matter) like this is somewhat hacky, mixes responsibilities into different layers and uses abstraction to hide details that should really be made explicit on correct levels.
- From its inception, property loading inside kernel API has been eager and very simple to reason about because of that. When getAllProperties was implemented this was broken, by mistake I think. A lazy property cursor is exposed through the kernel API just so that the core API can take it and delegate it back into a method making it eager (converting it into a Map).

My proposed solution, which I discussed with @burqen is that kernel API should expose eager nodeGetAllProperties/relationshipGetAllProperties which returns Map directly and then, like magic, this whole problem goes away. So instead of working around the problem, in this case we can just move code to its right place.

davidegrohmann:
@tinwelint The lazy fetching properties was done on purpose and explicitly designed that way and that has _not_ being introduced by mistake. Making the loading of properties eager again is moving us backwards not forwards. I am happy to discuss a better way of implementing this fix, but I am afraid we can simply rely on eagerness.

tinwelint:
@davidegrohmann the property loading is eager, it was just lazy in a small and insignificant part of the execution stack to no benefit. Why should it be lazy?

MishaDemianenko:
@burqen @tinwelint @davidegrohmann 
Fully agree with @tinwelint: implementation is a hack and workaround, if we can remove lazy property cursors at all that should be a way to go, my believe that cursors that do low level things should do just that and transaction/statement should do what ever they do and we need to try to avoid mixing responsibilitites

MishaDemianenko (changes requested):
@burqen @tinwelint @davidegrohmann
Fully agree with @tinwelint: implementation is a hack and workaround, if we can remove lazy property cursors at all that should be a way to go, my believe that cursors that do low level things should do just that and transaction/statement should do what ever they do and we need to try to avoid mixing responsibilities

davidegrohmann:
@MishaDemianenko we only use lazy cursors to load properties, that has been the case since 2.3. What we expose through the OperationsFacade are "eagerized" operations on the property cursors.

tinwelint:
To clarify my view: Lazy cursors _inside_ the kernel API is fine and should definitely be kept, that was not in question. We just have to make sure that we don't leak the lazyness _outside_ the kernel API.

burqen:
@davidegrohmann @tinwelint @MishaDemianenko 
The problem is that `nodeGetProperties` and `relationshipGetProperties` is _not_ "eagerized" in OperationsFacade, but rather one level above in `NodeProxy` which makes it impossible for to use the `finally { statement.assertOpen }` solution that we use in other places in `OperationsFacade`.

EDIT: I'm thinking of 3.2. NVM

tinwelint:
@burqen yup, that is exactly what I'd like us to change

burqen:
@tinwelint Although, that solution will not work for 3.1, because NodeProxy gets property cursor from node cursor and not through OperationsFacade.

tinwelint:
@burqen uhm, we'll of course change NodeProxy together with OperationsFacade, won't we?

On the other hand, it would be bad to break procedures using kernel API in 3.1 and 3.2, so let's keep the solution as is and we can fix this in a later version to not let this change spiral out of control with amount of "other" work.

tinwelint (approved):

tinwelint:
@MishaDemianenko any pending comments?

MishaDemianenko:
@tinwelint was never pinged back have no idea what happening in here, @burqen is it ready to review round?

community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public interface AssertOpen
+{
+    /**
+     * Assert that source tied to instance is open.
+     * Should throw exception with reason if source is not open.
+     */
+    void assertOpen();
+
+    AssertOpen ALWAYS_OPEN = AlwaysOpen.INSTANCE;
```
MishaDemianenko: why not just:
`    AssertOpen ALWAYS_OPEN = () ->
     {
        //always open
    };
` instead of enum that can be even additional referenced and pollute interface since can be references as `AssertOpen.AlwaysOpen.INSTANCE` without any visible in a glance reason?

community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
```diff
@@ -0,0 +1,46 @@
...
+ * Used to verify that some source of operation is still open when operation is performed.
+ * E.g. verify that transaction is still open when lazily reading property values from store.
+ */
+public interface AssertOpen
+{
+    /**
+     * Assert that source tied to instance is open.
+     * Should throw exception with reason if source is not open.
+     */
+    void assertOpen();
```
MishaDemianenko, line 32: if this thing should throw some exception would it make sense to throw some specific exception when underlying resource is closed (it can be runtime to avoid exception declaration anywhere), wdyt?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/StoreAbstractNodeCursor.java
```diff
@@ -187,27 +194,28 @@ private Lock shortLivedReadLock()
...
+        return nodeRelationshipCursor.get().init( nodeRecord.isDense(), nodeRecord.getNextRel(), nodeRecord.getId(),
+                direction, assertOpen );
     }
 
     @Override
     public Cursor<RelationshipItem> relationships( Direction direction, int... relTypes )
     {
         return nodeRelationshipCursor.get().init( nodeRecord.isDense(), nodeRecord.getNextRel(), nodeRecord.getId(),
-                direction, relTypes );
+                direction, assertOpen, relTypes );
```
MishaDemianenko, line 218: shouldn't it be the last param for the god of symmetry?)

community/kernel/src/main/java/org/neo4j/storageengine/api/StorageStatement.java
```diff
@@ -68,19 +69,21 @@
      * to place the cursor over the first item and then more calls to move the cursor through the selection.
      *
      * @param nodeId id of node to get cursor for.
+     * @param assertOpen to check if source transaction is still open on each property value fetch
```
MishaDemianenko: why do we need to say that assertOpen checking source transaction? shouldn't comment have a more generic expectation from what assert open doing? (this comment related to all new assertOpen params introduced)

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
```diff
@@ -147,7 +147,7 @@ public StateHandlingStatementOperations(
     @Override
     public Cursor<NodeItem> nodeCursor( KernelStatement statement, long nodeId )
     {
-        Cursor<NodeItem> cursor = statement.getStoreStatement().acquireSingleNodeCursor( nodeId );
+        Cursor<NodeItem> cursor = statement.getStoreStatement().acquireSingleNodeCursor( nodeId, statement );
```
MishaDemianenko, line 150: using statements `assertOpen` will do volatile read each time when we read something from cursor, did we check what effect this will have on performance?

MishaDemianenko (commented):
@burqen couple of comments

community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
tinwelint: That is probably what that sentence means.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/StoreAbstractNodeCursor.java
tinwelint: Because relTypes is vararg, no?

community/kernel/src/main/java/org/neo4j/storageengine/api/StorageStatement.java
tinwelint: May I say that some comments are simply too detailed/bike-shed:y. Would you agree?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
tinwelint: That would be interesting yeah. I'd expect a very small (unnoticable) impact since the volatile `open` flag was checked once per property request (one or more properties) previously, now it's once per property.

community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
MishaDemianenko: what? comment is about having some specific type of exception no? comment do not mention anything about that

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/StoreAbstractNodeCursor.java
MishaDemianenko: yes make sence

MishaDemianenko: yes make sense

community/kernel/src/main/java/org/neo4j/storageengine/api/StorageStatement.java
MishaDemianenko: i think anyone are free to leave comment if they think it can make PR better in one way or another

community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
tinwelint: I agree, this is a weird way of doing it

community/kernel/src/main/java/org/neo4j/kernel/api/AssertOpen.java
tinwelint: If we rely on specific exceptions thrown then absolutely that should be mentioned. @burqen do we?

burqen: `KernelStatement` will through two different variants of `RuntimeException`.
As that is an unchecked exception I would not put it in method signature, but maybe clarify it in javadoc. Sounds good?

From `KernelStatement`:
```
    @Override
    public void assertOpen()
    {
        if ( referenceCount == 0 )
        {
            throw new NotInTransactionException( "The statement has been closed." );
        }

        Status terminationReason = transaction.getReasonIfTerminated();
        if ( terminationReason != null )
        {
            throw new TransactionTerminatedException( terminationReason );
        }
    }
```

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
burqen: I have made no such investigation. Will trigger a jmh-run.
EDIT: ... once that service is up and running again.

chrisvest: What we can expect to see is that the cost of a _volatile_ load compared to a normal load, is usually negligible on x86_64 except in tight loops, and slightly more expensive on Power.

burqen: @chrisvest Would you go so far as to say that this is safe enough to merge without benchmarking it?

MishaDemianenko: I'm more interested in the influence of this change to set of optimisations that runtime can apply. Anyway JMH run will show if this change brings any changes.

MishaDemianenko: @burqen do you have any news about benchmarks run?

MishaDemianenko (approved):

PR #9017, opened 2017-03-13
Cleanup, use gen instead of generation in index module

burqen:
In code, be consistent with using 'gen' instead of 'generation'.
In comments, use 'generation'


MishaDemianenko:
@burqen to be quite honest I not quite sure I understand why we need to do this and use gen instead of generation because it makes code/docs more cryptic then before and require even more context then before.

burqen:
@MishaDemianenko That's exactly the kind of feedback I was looking for!
What do you think about doing it the other way instead? To use 'generation' everywhere?

MishaDemianenko:
@burqen i think that would be better since you do not need to guess what `gen` by itself means 👍

burqen:
@MishaDemianenko Checkout #9030 
Closing this PR

PR #9002, opened 2017-03-10
Support composite uniqueness constraints from Cypher

craigtaverner:
Includes support for constraint population and verification

changelog: Support composite uniqueness constraints from Cypher, including creation, dropping and verification

PR #9019, opened 2017-03-13
Composite unique index population

fickludd:
Restructuring of index provider tests suite to now properly test composite indexes.

The actual implementation of the composite variant of `DuplicateCheckingCollector` was already implemented as part of the cypher composite constraint PR (#9002), but is included here for completeness


ragadeeshu:
@fickludd needs a rebase

PR #9022, opened 2017-03-14
Update QueryLoggerIT queries log file check.

MishaDemianenko:


MishaDemianenko:
🔴  **N.B!**  this PR should be null merged since it's this test is already fixed in 3.1 and beyond

PR #9023, opened 2017-03-14
Tell user that consistency check is running

spacecowboy:
changelog: `neo4j-admin backup` now reports different error codes:
- code `1`: backup failed to run
- code `2`: backup succeeded but consistency check failed to run
- code `3`: backup succeeded but consistency check found inconsistencies

### DIFF
--- a/community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
+++ b/community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
@@ -32,6 +32,10 @@ import static java.lang.String.format;
 
 public class AdminTool
 {
+
+    static final int STATUS_SUCCESS = 0;
+    public static final int STATUS_ERROR = 1;
+
     public static void main( String[] args ) throws IOException
     {
         Path homeDir = Paths.get( System.getenv().getOrDefault( "NEO4J_HOME", "" ) );
@@ -159,7 +163,7 @@ public class AdminTool
 
     private void failure( String message, Exception e )
     {
-        failure( message, e, 1 );
+        failure( message, e, STATUS_ERROR );
     }
 
     private void failure( String message, Exception e, int code )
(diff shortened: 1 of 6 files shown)


community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
```diff
@@ -159,17 +159,22 @@ private void failure()
 
     private void failure( String message, Exception e )
     {
+        failure( message, e, 1 );
```
praveenag: It might be useful to extract `1, 2,3` as useful constants or enums to indicate what the code means.

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
```diff
@@ -236,6 +236,7 @@ public void execute( String[] args ) throws IncorrectUsage, CommandFailed
         {
             try
             {
+                outsideWorld.stdOutLine( "Doing consistency check..." );
```
praveenag, line 241: `Performing consistency check` feels like a better error message (I am not a native speaker though :/)

praveenag (changes requested):

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
spacecowboy: I think they are both fine. I did `Doing` because it matches what we say for the import above:

> Doing full backup...

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
spacecowboy: extracted as constants

praveenag (approved):

PR #9024, opened 2017-03-14
Only log changes in read replica topology rather than the whole thing

mneedham:


mneedham:
Unrelated test failures

PR #9028, opened 2017-03-14
GBPTree should not checkpoint during close if panic

burqen:
Checkpoint will increment stable and unstable generation of tree, making it impossible to recognise crash-pointers during the recovery following the database crash.


community/common/src/main/java/org/neo4j/kernel/Health.java
```diff
@@ -0,0 +1,56 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel;
+
+public interface Health
```
MishaDemianenko: Lets add javadocs for this shiny new interface

community/common/src/main/java/org/neo4j/kernel/Health.java
```diff
@@ -0,0 +1,56 @@
...
+{
+    /**
+     * Asserts that the database is in good health. If that is not the case then the cause of the
+     * unhealthy state is wrapped in an exception of the given type, i.e. the panic disguise.
+     *
+     * @param panicDisguise the cause of the unhealthy state wrapped in an exception of this type.
+     * @throws EXCEPTION exception type to wrap cause in.
+     */
+    default <EXCEPTION extends Throwable> void assertHealthy( Class<EXCEPTION> panicDisguise ) throws EXCEPTION
```
MishaDemianenko: Why all methods in this newly introduced interface are default?

community/common/src/main/java/org/neo4j/kernel/Health.java
```diff
@@ -0,0 +1,56 @@
...
+    default void healed()
+    {
+    }
+
+    default Throwable cause()
+    {
+        return null;
+    }
+
+    class Adapter implements Health
```
MishaDemianenko: What if instead of this simple Adapter class we will have Health implementor in it's own class file with descriptive name?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -757,9 +765,12 @@ private void close( Header.Writer headerWriter ) throws IOException
 
         try
         {
-            // Perform a checkpoint before closing. If no changes has happened since last checkpoint,
-            // no new checkpoint will be created.
-            checkpoint( IOLimiter.unlimited(), headerWriter );
+            if ( health.isHealthy() )
```
MishaDemianenko: Maybe lets add some log statement for case when we skip checkpointing?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -107,6 +114,10 @@
     private final FullStoreChangeStream fullStoreChangeStream;
 
     /**
+     * todo
```
MishaDemianenko: todo

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
```diff
@@ -278,7 +280,9 @@ public BatchInserterImpl( final File storeDir, final FileSystemAbstraction fileS
         indexStoreView = new NeoStoreIndexStoreView( LockService.NO_LOCK_SERVICE, neoStores );
 
         Dependencies deps = new Dependencies();
-        deps.satisfyDependencies( fileSystem, config, logService, indexStoreView, pageCache );
+        Monitors monitors = new Monitors();
+        Health databaseHealth = new Health.Adapter();
```
MishaDemianenko: do we really need to use dummy health here?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -161,6 +162,8 @@ private BatchingNeoStores( FileSystemAbstraction fileSystem, PageCache pageCache
         dependencies.satisfyDependency( logService );
         dependencies.satisfyDependency( IndexStoreView.EMPTY );
         dependencies.satisfyDependency( pageCache );
+        dependencies.satisfyDependency( new Monitors() );
+        dependencies.satisfyDependency( new Health.Adapter() );
```
MishaDemianenko: same here, do we need to use dummy health here?

MishaDemianenko (changes requested):
Couple of comments, good otherwise

community/common/src/main/java/org/neo4j/kernel/Health.java
burqen: To simplify `Adapter` implementation. Will remove `Adapter` and create more descriptive class as suggested 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: Ye, it would be good. It should to be logged from `LabelScanStore` level though as `GBPTree` is just some underlying implementation detail.

It is not super clear to me how this should be with relation to `LuceneLabelScanStore`.

But maybe something like:
- Add `close( didCheckpoint )` to `GBPTree.Monitor` and trigger during `GBPTree.close()`.
- Propagate monitor call to `LabelScanStore.Monitor` which register it as `shutdown( didCheckpoint )`. 
- `LoggingMonitor` implementation log some line about this.

Problem: What does `LabelScanStore.shutdown( didCheckpoint )` mean for `Lucene` implementation?

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
burqen: We need _some_ `Health` to make the `KernelExtension` load and we don't need full `DatabaseHealth` so might as well use most simple one. No?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
burqen: Same as above

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: I like the monitor approach. If `didCheckpoint` would instead be called e.g. `cleanShutdown` it would apply better to other label index implementations too.

burqen: 👍

burqen: @tinwelint , I talked to @MishaDemianenko about it. He suggested just injecting a `Log` instance and use that directly.

Please have a look at commit `2217c3b` to see what you think about it.

community/common/src/main/java/org/neo4j/kernel/AlwaysHealthy.java
```diff
@@ -0,0 +1,24 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel;
+
+public class AlwaysHealthy extends Health.Adapter
```
MishaDemianenko: Now we have `AlwaysHealthy` and Adapter and first one which is basically Adapter except it's public, what if we will have SilentHealth (current AlwaysHealthy thingy) and will remove Adapter since all implementations can extend that SilentHealth instead. WDYT?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -754,16 +772,25 @@ public void close( Consumer<PageCursor> headerWriter ) throws IOException
...
+                // Perform a checkpoint before closing. If no changes has happened since last checkpoint,
+                // no new checkpoint will be created.
+                anyWrites = checkpoint( IOLimiter.unlimited(), headerWriter );
+                didCheckpoint = true;
+            }
         }
         finally
         {
             pagedFile.close();
+            String checkpointMessage = didCheckpoint ?
```
MishaDemianenko: i would just say: `Skip index checkpoint since database is not healthy.` and did not log anything in any other case, but maybe that just me so if you think that strange or not enough, just ignore this comment.

community/kernel/src/test/java/org/neo4j/index/GBPTreeLogTest.java
```diff
@@ -0,0 +1,128 @@
...
+    private final TestDirectory directory = TestDirectory.testDirectory( getClass(), fs.get() );
+    private final PageCacheRule pageCacheRule = new PageCacheRule();
+    private final AssertableLogProvider logProvider = new AssertableLogProvider();
+
+    @Rule
+    public final RuleChain rules = outerRule( fs ).around( directory ).around( pageCacheRule ).around( logProvider );
+
+    private final Layout<MutableLong,MutableLong> layout = new SimpleLongLayout();
+
+    private void insert( GBPTree<MutableLong, MutableLong> index, long key ) throws IOException
```
MishaDemianenko: please move private helper methods bellow all the tests

MishaDemianenko (commented):
Couple of minor comments/questions

community/common/src/main/java/org/neo4j/kernel/AlwaysHealthy.java
burqen: Sounds good, I was not very happy with the solution I came up with but I wanted the flexibility of an Adapter and also the "more useful" class name.
`SilentHealth` seems like a good middle road 👍

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: I think will keep it since the problem we are addressing is when we do checkpoints even though we shouldn't.

burqen:
@MishaDemianenko Comments addressed and PR rebased

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: @burqen @MishaDemianenko I can't agree with that. May I say that Monitor is _always_ preferred over direct Log. A Monitor has way more usability, is customizable and is easier to test (you do want to test this don't you). We don't use direct logging anywhere else in index component, why here?

MishaDemianenko: @tinwelint can you accept current way of logging, or you would prefer monitor instead? I'm fine with any of those, sure monitor have much wider range of applicability but since here we just need to log i think using just log is fine (and we have assertable log to be able to test that)

MishaDemianenko (approved):

MishaDemianenko:
@burqen fine for me, lets wait for @tinwelint if he is fine with proposed changes.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: @MishaDemianenko It's just a very strange inconsistency in the index component, to bypass the usual pattern for a particular event. Also, it's much better to have the layer outside gb+tree, e.g. label scan store to log something more contextual, saying that "native label scan store didn't checkpoint blabla" instead of having a log statement saying that some gb+tree did something... what's that to the user? Weird I'd say. Think about logging as surface too.

Let's keep abstractions and try not to bypass for quick fixes. Please.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: See above. We can't really know what to call this thing here. `index`, meh, for different uses this is a different thing, e.g. in label scan store this is a `label index`, for a schema index this is a `schema index for label L and property P`, not just `index`. This has to be a monitor to be done properly.

burqen:
@MishaDemianenko @tinwelint Some new commits to this bad boy.
Whole responsibility for checkpointing `GBPTree` moves to `NativeLabelScanStore`. This is to avoid unsafe checkpoints in the case where recovery fails.

tinwelint:
@burqen With the recent developments the DatabaseHealth refactorings are a bit unnecessary, aren't they Although I'd guess it doesn't harm us. One thing you could do though is to reduce `Health` down to `isHealthy()` w/ perhaps additionally `assertHealthy()` because I'd like to have the database, i.e. for now the specific implementation `DatabaseHealth` to control when panic happens.

So simply reduce number of methods in `Health` interface.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -426,6 +443,12 @@ public void stop() throws IOException
     @Override
     public void shutdown() throws IOException
     {
+        boolean flush = started && health.isHealthy();
+        monitor.flushDuringShutdown( flush );
+        if ( flush )
+        {
+            index.checkpoint( IOLimiter.unlimited() );
```
MishaDemianenko, line 450: shouldn't we close index first (at least index writer) and then checkpoint?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -426,6 +443,12 @@ public void stop() throws IOException
     @Override
     public void shutdown() throws IOException
     {
+        boolean flush = started && health.isHealthy();
+        monitor.flushDuringShutdown( flush );
```
MishaDemianenko, line 447: why do we call this thing flush instead of checkpoint? wouldn't that be a bit confusing?

MishaDemianenko:
@burqen 2 comments left :)

burqen:
This PR is getting out of hand. Redo fix in https://github.com/neo4j/neo4j/pull/9111
Will cherry-pick the good stuff from this PR into new smaller PRs.

PR #9029, opened 2017-03-14
Close index resources properly.

MishaDemianenko:


MishaDemianenko:
🔴  **N.B!** PR should be null merged since 3.1 and above are fixed by separate PR.

burqen (changes requested):
There are quite a few calls to `index.query` in this test. Do we not need to close all of them?

MishaDemianenko:
@burqen change is exact copy of changes done to 3.1-3.x branches current solution works for them ..

burqen (approved):

PR #9032, opened 2017-03-14
Check for repeated properties in composite indexes and constraints

fickludd:
changelog: Creating composite indexes or constraints where the same property occurs in more than one position is no longer allowed


thobe (approved):
Simple and straightforward.

Should be rebased on a non-broken branch.

PR #9033, opened 2017-03-14
Optimized property lookup in index population

fickludd:
Fixes performance regression in modifying nodes with at least one label and many properties. This was caused by loading all properties on any change, which was done to find out what indexes that needed updating. With this PR, only properties that are relevant to at least one online index are loaded.


community/kernel/src/main/java/org/neo4j/kernel/api/index/NodeUpdates.java
```diff
@@ -125,218 +118,176 @@ private NodeUpdates( long nodeId, long[] labelsBefore, long[] labelsAfter )
...
+                hasPropsAfter( schema.getPropertyIds() );
+    }
+
+    private void loadProperties( PropertyLoader propertyLoader, PrimitiveIntSet additionalPropertiesToLoad )
+    {
+        PrimitiveIntIterator toLoad = additionalPropertiesToLoad.iterator();
+        while ( toLoad.hasNext() )
+        {
+            int propertyId = toLoad.next();
+            Object value = propertyLoader.loadProperty( nodeId, propertyId );
```
thobe: Ouch. Is there any way that we could load all properties in one call instead? Like how it's done in `PropertyContainerProxyHelper`.

community/kernel/src/main/java/org/neo4j/kernel/api/index/NodeUpdates.java
```diff
@@ -487,66 +503,32 @@ private static boolean propertyValueEqual( Object a, Object b )
...
+    }
 
-        private PropertyChanged( int propertyKeyId, Object before, Object after )
-        {
-            super( propertyKeyId, after, UpdateMode.CHANGED );
-            this.before = before;
-        }
+    public interface PropertyLoader
+    {
+        Object loadProperty( long nodeId, int propertyId );
```
thobe: I'm thinking this method could be made more specialized to its purpose:

`void loadProperties( long nodeId, PrimitiveIntSet keys, PropertyLoadSink sink );`

Where `NodeUpdates` would implement `PropertyLoadSink` which would be defined as:
```
interface PropertyLoadSink {
    void loadedProperty( int key, Object value);
}
```
and implemented as:
```
class NodeUpdates implements PropertyLoadSink {
...
    public void loadedProperty(int key, Object value) {
        put(key, unchanged(value));
    }
...
}
```

I'm also thinking that implementations of `PropertyLoader.loadProperties(...)` would be allowed to modify `keys` so that after the call it only contains the keys that have no value.

This would allow avoiding multiple calls to the store layer for loading properties.

thobe: With such a change to `PropertyLoader`, the `NodeUpdates.loadProperties(...)` method would change to:
```
void loadProperties( PropertyLoader propertyLoader, PrimitiveIntSet additionalPropertiesToLoad ) {
    propertyLoader.loadProperties( nodeId, additionalPropertiesToLoad, this );
    for ( PrimitiveIntIterator remaining = additionalPropertiesToLoad.iterator(); remaining.hasNext(); ) {
        knownProperties.put( remaining.next(), noValue );
    }
}
```

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -433,13 +434,25 @@ private void apply( IndexUpdates updates, IndexUpdateMode updateMode )
...
+                    try
                     {
-                        updaterMap.getUpdater( descriptor ).process( entry.get() );
+                        return storeView.getProperty( nodeId, propertyId ).value();
                     }
+                    catch ( EntityNotFoundException e )
+                    {
+                        // TODO: we should not use exceptions for this, because that is slow
+                        return null;
+                    }
```
thobe: This catch block should never be reached in this case, so it's not so bad.

What might be worse than the slowness is that if this happens it would be a symptom of something gone quite horribly wrong, so perhaps we shouldn't just silently ignore it like this...

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -433,13 +434,25 @@ private void apply( IndexUpdates updates, IndexUpdateMode updateMode )
...
+                    catch ( PropertyNotFoundException e )
+                    {
+                        // TODO: we should not use exceptions for this, because that is slow
+                        return null;
+                    }
```
thobe: You can (and should) avoid this catch block by using `storeView.getProperty( nodeId, propertyId ).value( null )` instead of `storeView.getProperty( nodeId, propertyId ).value()`.

But, as stated elsewhere, I think this should change to an interface that allows loading multiple properties at once, at which point I think `storeView` could implement that interface (the replacement for `NodeUpdate.PropertyLoader`) directly, allowing us to avoid allocating a lambda here.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
```diff
@@ -533,7 +539,18 @@ public boolean visit( NodeUpdates updates ) throws IndexPopulationFailedKernelEx
 
         private void add( NodeUpdates updates )
         {
-            forEachPopulation( population -> population.add( updates ) );
+            for ( IndexEntryUpdate indexUpdate : updates.forIndexes( populations.keySet(), NO_UNCHANGED_PROPERTIES ) )
```
thobe: What makes us know that there are `NO_UNCHANGED_PROPERTIES` here?
It's one of those things that would be nice to have in a comment, so that you don't have to go digging through the code for the reason.

I'm guessing it is because this is called from `StoreViewNodeStoreScan.process(...)` where we fully load all properties of a node into the `NodeUpdates` already, but it would be nice to have things like this in comments, so that someone reading the code or making changes to it only has to verify this assumption, and not completely figure out the whole reason on their own.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
```diff
@@ -311,18 +317,17 @@ public void flipAfterPopulation()
 
     private int[] propertyKeyIds()
     {
-        return populations.stream().flatMapToInt( this::propertyKeyIds ).distinct().toArray();
+        return populations.keySet().stream().flatMapToInt( this::propertyKeyIds ).distinct().toArray();
     }
```
thobe: This method is actually pretty inefficient, it is only ever used together with `labelIds()`, it would be better to just have a single for-loop build both arrays at once, the streams just add overhead.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
```diff
@@ -533,7 +539,18 @@ public boolean visit( NodeUpdates updates ) throws IndexPopulationFailedKernelEx
 
         private void add( NodeUpdates updates )
         {
-            forEachPopulation( population -> population.add( updates ) );
+            for ( IndexEntryUpdate indexUpdate : updates.forIndexes( populations.keySet(), NO_UNCHANGED_PROPERTIES ) )
+            {
+                IndexPopulation population = populations.get( indexUpdate.descriptor() );
```
thobe: This seems to be the only reason why `populations` changed from a List to Map, I wonder if we couldn't somehow allow passing the `IndexPopulation` objects (opaquely) through the `NodeUpdates.forIndexes(...)` method, so that we can avoid this lookup altogether.

It seems to me that we add quite a bit of overhead to the data structures with very little gain. Although passing the `IndexPopulation` objects through might add code complexity instead. Feels like one of those things that we will not know until we try it out.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/OnlineIndexUpdates.java
```diff
@@ -181,18 +173,19 @@ private NodeRecord loadNode( long nodeId )
...
+//            List<PropertyCommand> propertyCommands )
+//    {
+//        NodeRecord nodeRecord = (nodeCommand == null) ? loadNode( nodeId ) : nodeCommand.getAfter();
+//
+//        IteratingPropertyReceiver receiver = new IteratingPropertyReceiver();
+//        PrimitiveLongObjectMap<PropertyRecord> propertiesById =
+//                propertiesFromCommandsForNode( propertyCommands );
+//        propertyLoader.nodeLoadProperties( nodeRecord, propertiesById, receiver );
+//        return receiver;
+//    }
```
thobe: Just remove it. If it is not used it servers no purpose.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/StoreViewNodeStoreScan.java
```diff
@@ -107,10 +107,8 @@ public void process( NodeRecord node ) throws FAILURE
                     updates.added( propertyKeyId, value );
                 }
             }
-            if ( updates.hasUpdates() )
-            {
-                propertyUpdatesVisitor.visit( updates.build() );
-            }
+            // TODO: only visit if has label change or label + prop change?
+            propertyUpdatesVisitor.visit( updates.build() );
```
thobe: if `updates` has _any_ label/property changes at all? We already know that since the node is from the persistent store there are no _changes_. And determining if the changes are actually relevant or not is what the implementation of the visitor already does.

What we could do is a crude pre-filtering, where we keep a single boolean flag that we set to true in the `if ( propertyKeyIdFilter.test( propertyKeyId ) )` block, is that what you mean?

I can't see that hurting, so why not just do that instead of leaving a vague comment in the code that is mostly just going to cause confusion for future readers of this code.

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
```diff
@@ -443,47 +443,48 @@ private void repopulateAllIndexes() throws IOException, IndexEntryConflictExcept
...
+            populator.create();
+            populators.put( index.schema(), populator );
+            indexes.put( index.schema(), index );
         }
 
         Visitor<NodeUpdates, IOException> propertyUpdateVisitor = updates -> {
             // Do a lookup from which property has changed to a list of indexes worried about that property.
-            for ( int i = 0; i < descriptors.length; i++ )
+            for ( IndexEntryUpdate indexUpdate :
+                    updates.forIndexes( Iterables.asIterable( descriptors ), NO_UNCHANGED_PROPERTIES ) )
```
thobe: Again, it would be nice to note _why_ `NO_UNCHANGED_PROPERTIES` is appropriate here.

thobe (changes requested):
I think we can do slightly better.

I would also really like to see the benchmark results to know that the changes made really improves the things they intend to improve.

thobe:
This PR also needs rebasing on top of a non-broken base branch.

fickludd:
Thanks for the great review @thobe. I do have benchmarking data to prove this works, will share via other channel. In hindsight I realize that I made the PR directly after benchmark confirmation,  without making the final clean-up pass...

Will iterate this and return!

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
fickludd: That is true, but this is a place where any solution works because it's called once per node scan.

community/kernel/src/main/java/org/neo4j/kernel/api/index/PropertyLoader.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api.index;
+
+import org.neo4j.collection.primitive.PrimitiveIntSet;
+
+/**
+ * TODO: make doc. This should probably not be in API, unless directly used by index providers.
```
thobe: I think we should act on this `TODO` before merging.

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/LabelSchemaSupplier.java
```diff
@@ -0,0 +1,48 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api.schema_new;
+
+public interface LabelSchemaSupplier extends SchemaDescriptor.Supplier
+{
+    LabelSchemaDescriptor schema();
+
+    class Box<VALUE> implements LabelSchemaSupplier
```
thobe: `Box` does not appear to be used anywhere.

thobe (changes requested):
Looks good in general, but there are a few things that don't look like they were completed:

* There is a `TODO` comment that I think should be addressed.
* There is an unused class that I think should be removed.

thobe (approved):

PR #9034, opened 2017-03-14
NODE KEY constraint as syntactic sugar over uniqueness and existence

craigtaverner:
Support for the `NODE KEY` constraint. This PR allows `CREATE` and `DROP` over cypher, as well as listing via Core API. 

In this implementation `NODE KEY`s are syntax sugar for doing one uniqueness constraint, and one node property existence constraint per property in the `NODE KEY`. Because of this, extra logic is needed in the `DataIntegrityValidatingStatementOperations` to maintain this illusion outwards so that we don't accidentally delete some part of a NODE KEY, or expose parts to the user. Note also that this means that on upgrade any uniqueness constraints with matching existence constraints will be presented as `NODE KEY` only.

On the plus side the enforcement and internal management of constraints become more direct, and no extra code is needed in the TxState, SchemaStore or constraint enforcement. The orthogonal concepts of property uniqueness and existence remain separated in code. 

changelog: New `NODE KEY` constraint on one or many properties. A `NODE KEY` (eg. `CREATE CONSTRAINT ON (n:Person) ASSERT (n.email,n.name) IS NODE KEY`) implies both that nodes with the given label (`Person`) must have the properties (`email` and `name`), and that the values for these properties are unique among nodes with the label.


fickludd:
@thobe @MishaDemianenko Now I think it should be reasonable to review this PR. First 6 commits are Cypher, rest are kernel implementation + Cypher testing.

fickludd:
This PR is closed, because we have decided to follow the full stack implementation instead (#9046).

PR #9037, opened 2017-03-15
Show allocated memory and page hits/faults in `dbms.listQueries`

thobe:
Sample output:

```
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| queryId   | username | metaData | query                         | parameters | planner     | runtime       | indexes | startTime                       | elapsedTime    | protocol | clientAddress | requestUri | status    | resourceInformation                                                                          | activeLockCount | elapsedTimeMillis | cpuTimeMillis | waitTimeMillis | idleTimeMillis | allocatedBytes | pageHits | pageFaults |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| "query-1" | ""       | {}       | "MATCH (n) SET n.v = n.v + 1" | {}         | "idp"       | "interpreted" | []      | "2017-03-15T14:27:36.243+01:00" | "00:00:00.177" | <null>   | <null>        | <null>     | "waiting" | {waitTimeMillis -> 169, lockMode -> "EXCLUSIVE", resourceType -> "NODE", resourceIds -> [0]} | 1               | 177               | 12            | 165            | 0              | 617232         | 2        | 0          |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

This change set also adds the ability to include the memory and page hit/fault information in the query log. This is toggled through settings, and off by default in order not to change the format of the query log.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -473,7 +473,6 @@ public class GraphDatabaseSettings implements LoadableConfig
             "Log entries are written to the file _query.log_ located in the Logs directory. " +
             "For location of the Logs directory, see <<file-locations>>. " +
             "This feature is available in the Neo4j Enterprise Edition." )
-
     public static final Setting<Boolean> log_queries = setting("dbms.logs.query.enabled", BOOLEAN, FALSE );
 
     @Description("Path of the logs directory")
@@ -485,16 +484,16 @@ public class GraphDatabaseSettings implements LoadableConfig
             ( logs ) -> new File( logs, "query.log" ),
             PATH );
 
-    @Description( "Log parameters for executed queries that took longer than the configured threshold." )
+    @Description( "Log parameters for the executed queries being logged." )
     public static final Setting<Boolean> log_queries_parameter_logging_enabled = setting( "dbms.logs.query.parameter_logging_enabled", BOOLEAN, TRUE );
 
-    @Description( "Log detailed time information for executed queries that took longer than the configured threshold." )
+    @Description( "Log detailed time information for the executed queries being logged." )
     public static final Setting<Boolean> log_queries_detailed_time_logging_enabled = setting( "dbms.logs.query.time_logging_enabled", BOOLEAN, FALSE );
 
-    @Description( "Log allocated bytes for executed queries that took longer than the configured threshold." )
+    @Description( "Log allocated bytes for the executed queries being logged." )
     public static final Setting<Boolean> log_queries_allocation_logging_enabled = setting( "dbms.logs.query.allocation_logging_enabled", BOOLEAN, FALSE );
 
-    @Description( "Log page hits and page faults for executed queries that took longer than the configured threshold." )
+    @Description( "Log page hits and page faults for the executed queries being logged." )
     public static final Setting<Boolean> log_queries_page_detail_logging_enabled = setting( "dbms.logs.query.page_logging_enabled", BOOLEAN, FALSE );
 
     @Description("If the execution of query takes more time than this threshold, the query is logged - " +
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -488,6 +488,12 @@ public class GraphDatabaseSettings implements LoadableConfig
     @Description( "Log parameters for executed queries that took longer than the configured threshold." )
     public static final Setting<Boolean> log_queries_parameter_logging_enabled = setting( "dbms.logs.query.parameter_logging_enabled", BOOLEAN, TRUE );
 
+    @Description( "Log detailed time information for executed queries that took longer than the configured threshold." )
+    public static final Setting<Boolean> log_queries_detailed_time_logging_enabled = setting( "dbms.logs.query.time_logging_enabled", BOOLEAN, FALSE );
+
+    @Description( "Log allocated bytes for executed queries that took longer than the configured threshold." )
+    public static final Setting<Boolean> log_queries_allocation_logging_enabled = setting( "dbms.logs.query.allocation_logging_enabled", BOOLEAN, FALSE );
+
     @Description("If the execution of query takes more time than this threshold, the query is logged - " +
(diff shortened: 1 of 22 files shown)


community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -488,6 +488,12 @@
     @Description( "Log parameters for executed queries that took longer than the configured threshold." )
     public static final Setting<Boolean> log_queries_parameter_logging_enabled = setting( "dbms.logs.query.parameter_logging_enabled", BOOLEAN, TRUE );
 
+    @Description( "Log detailed time information for executed queries that took longer than the configured threshold." )
```
praveenag: I can't see these log settings in neo4j.conf. It would be useful to add them as commented out defaults.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -494,6 +494,9 @@
     @Description( "Log allocated bytes for executed queries that took longer than the configured threshold." )
     public static final Setting<Boolean> log_queries_allocation_logging_enabled = setting( "dbms.logs.query.allocation_logging_enabled", BOOLEAN, FALSE );
 
+    @Description( "Log page hits and page faults for executed queries that took longer than the configured threshold." )
+    public static final Setting<Boolean> log_queries_page_detail_logging_enabled = setting( "dbms.logs.query.page_logging_enabled", BOOLEAN, FALSE );
```
praveenag: I can't see these log settings in neo4j.conf. It would be useful to add them as commented out defaults.

praveenag (changes requested):

praveenag:
@thobe I'd probably refactor some of the methods into smaller classes of their own. I can do that once the requested changes are implemented and push it through for your review.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
thobe: @praveenag: Is there a file in which you do that manually these days? I thought we used to have an automated step that generated that file.

thobe:
@praveenag I'm happy to break things into smaller components. Which ones do you have in mind?

- [x] `QueryLogger` - break out formatting code.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
spacecowboy: @thobe we don't generate anything. See 

https://github.com/neo4j/neo4j/blob/3.2/packaging/standalone/standalone-enterprise/src/main/distribution/text/enterprise/conf/neo4j.conf

and

https://github.com/neo4j/neo4j/blob/3.2/packaging/standalone/standalone-community/src/main/distribution/text/community/conf/neo4j.conf

praveenag: Spoke @spacecowboy he mentioned that not everything is auto generated and we'll have to change the default config files here 
https://github.com/neo4j/neo4j/blob/3.2/packaging/standalone/standalone-community/src/main/distribution/text/community/conf/neo4j.conf

https://github.com/neo4j/neo4j/blob/3.2/packaging/standalone/standalone-enterprise/src/main/distribution/text/enterprise/conf/neo4j.conf

thobe: Looking at those files, I don't see `dbms.logs.query.parameter_logging_enabled` in there either. Surely these settings should be on equal footing with that setting?

(Also: I'm sad that there is no sad-face emoji in github I can add as a reaction to the fact that we don't generate the config files anymore - do we at least have tests that verify that the commented out settings in there are settings that actually exist?)

spacecowboy: @thobe not really.

We can't generate them because some settings in there are defined with different values than the actual defaults. With comments saying "Uncomment this line to enable XXX". And we don't want ALL settings in there.

thobe: We could solve that by generating the files from some annotation that specified what to put in the generated file - i.e. have different annotations for what to put in the generated file from what to put in the documentation.

Anyhow, we've added entries in the file now. (this only applies to enterprise)

praveenag (approved):

packaging/standalone/standalone-enterprise/src/main/distribution/text/enterprise/conf/neo4j.conf
```diff
@@ -137,6 +137,18 @@ dbms.connector.https.enabled=true
...
+
+# Uncomment this line to include detailed time information for executed queries:
+#dbms.logs.query.time_logging_enabled=true
+
+# Uncomment this line to include bytes allocated by executed queries:
+#dbms.logs.query.allocation_logging_enabled=true
+
+# Uncomment this line to include page hits and page faults information for executed queries:
+#dbms.logs.query.page_logging_enabled=true
+
```
thobe: @spacecowboy do these changes look like you would expect from this changeset?

spacecowboy: Hmm, reading the documentation vs this seems slightly off

* `dbms.logs.query.time_logging_enabled`: the docs mention `queries that took longer than the configured threshold` whereas the comment indicates that details regarding all queries will be logged

I think we should add this at the top of this changeset so it is clear what threshold we are talking about and to make sure people enable the global setting:

```
# Uncomment this line to enable query logging
#dbms.logs.query.enabled=true

# Log only queries exceeding this threshold
#dbms.logs.query.threshold=0s
```

spacecowboy (changes requested):

spacecowboy (approved):
lgtm

PR #8991, opened 2017-03-09
3.2 remove node property descriptor

ragadeeshu:

PR #9042, opened 2017-03-16
Remove relationship property descriptor and cleanup old constraints api

ragadeeshu:
Based on https://github.com/neo4j/neo4j/pull/8991


community/kernel/src/main/java/org/neo4j/kernel/api/constraints/UniquenessConstraint.java
```diff
@@ -21,12 +21,16 @@
...
+import org.neo4j.kernel.api.schema_new.constaints.UniquenessConstraintDescriptor;
 import org.neo4j.kernel.api.schema_new.index.NewIndexDescriptor;
 import org.neo4j.kernel.api.schema_new.index.NewIndexDescriptorFactory;
 
 /**
  * Description of uniqueness constraint over nodes given a label and a property key id.
+ *
+ * @deprecated use {@link UniquenessConstraintDescriptor} instead.
  */
+@Deprecated
```
fickludd, line 33: Aha, nice! This should have been done long ago.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/operations/SchemaReadOperations.java
```diff
@@ -89,8 +86,7 @@ PopulationProgress indexGetPopulationProgress( KernelStatement state, NewIndexDe
     String indexGetFailure( Statement state, NewIndexDescriptor descriptor ) throws IndexNotFoundKernelException;
 
     /**
-     * Get all constraints applicable to label and propertyKeys. There are only {@link NodePropertyConstraint}
-     * for the time being.
+     * Get all constraints applicable to label and propertyKeys.
```
fickludd, line 89: I guess that since the type is `SchemaDescriptor`, this could be used to fetch rel-type-property-existence-constraints as well.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureArgumentFormatter.java
```diff
@@ -107,25 +111,24 @@ else if ( arg instanceof LabelSchemaDescriptor )
...
         {
-            UniquenessConstraint constraint = (UniquenessConstraint) arg;
-            int labelId = constraint.label();
-            builder.append( format( "new UniquenessConstraint( SchemaDescriptorFactory.forLabel( %d, %s ) )", labelId,
-                    asString( constraint.descriptor().getPropertyIds() ) ) );
+            UniquenessConstraintDescriptor constraint = (UniquenessConstraintDescriptor) arg;
+            int labelId = constraint.schema().getLabelId();
+            builder.append( format( "ConstraintDescriptorFactory.uniqueForLabel( %d, %s )",
+                    labelId,
+                    asString( constraint.schema().getPropertyIds() ) ) );
```
fickludd, line 120: I guess this actually solves compositeness in `DbStructure`?

fickludd (approved):
Looks well and correct. Since the underlying PR is approved, I will merge this is one chunk if green.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureArgumentFormatter.java
ragadeeshu: I think there are a few other places as well that needs tweaking.

PR #9044, opened 2017-03-16
Lenient transaction log reader

tinwelint:
Reading transaction logs can now be more lenient and attempt to read through bad sections. This feature becomes very handy when debugging damaged transaction log files.

By default log reading is strict, but can be made lenient when merely dumping transaction logs to text and in similar tooling.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/InvalidLogEntryHandler.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/InvalidLogEntryHandler.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.transaction.log.entry;
+
+import org.neo4j.kernel.impl.transaction.log.LogPosition;
+
+/**
+ * Decides what happens to invalid log entries read by {@link LogEntryReader}.
+ */
+public interface InvalidLogEntryHandler
+{
+    /**
+     * Allows no invalid log entries.
+     */
+    public static final InvalidLogEntryHandler STRICT = new InvalidLogEntryHandler()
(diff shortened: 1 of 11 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/LogEntrySanity.java
```diff
@@ -0,0 +1,87 @@
...
+
+import static java.lang.Math.abs;
+import static java.lang.System.currentTimeMillis;
+
+/**
+ * Sanity checking for read {@link LogEntry log entries}.
+ */
+class LogEntrySanity
+{
+    private static final long UNREASONABLY_LONG_TIME = TimeUnit.DAYS.toMillis( 30 * 356 /*years*/ );
```
thobe: A year is approximately `365.25` days, not `356` days - you seem to have at least swapped the `6` and the `5` (and forgotten to account for leap days, but that could perhaps be written off).

(the value probably still represents an unreasonable long time, but the comment is wrong)

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/VersionAwareLogEntryReader.java
```diff
@@ -61,41 +65,70 @@ public LogEntry readLogEntry( SOURCE channel ) throws IOException
...
             while ( true )
             {
-                LogEntryVersion version = null;
-                LogEntryParser<LogEntry> entryReader;
+                channel.getCurrentPosition( positionMarker );
+
+                /*
+                 * if the read type is negative than it is actually the log entry version
+                 * so we need to read an extra byte which will contain the type
+                 */
```
thobe: This comment would be better closer to the logic that does this checking, rather than here where the variables are declared. i.e. move it back down to where it was, or even to the if-statement.

thobe (commented):
Just a few comments since I happened to look at these changes.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/InvalidLogEntryHandler.java
```diff
@@ -0,0 +1,60 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.transaction.log.entry;
+
+import org.neo4j.kernel.impl.transaction.log.LogPosition;
+
+/**
+ * Decides what happens to invalid log entries read by {@link LogEntryReader}.
+ */
+public interface InvalidLogEntryHandler
```
chrisvest: Could this be an abstract class instead?

tools/src/main/java/org/neo4j/tools/dump/LenientInvalidLogEntryHandler.java
```diff
@@ -0,0 +1,52 @@
...
+    public boolean handleInvalidEntry( Exception e, LogPosition position )
+    {
+        out.println( "Read broken entry error:" + e + " will go one byte ahead and try again" );
+        return true;
+    }
+
+    @Override
+    public void bytesSkipped( long bytesSkipped )
+    {
+        out.println( "... skipped " + bytesSkipped + " bytes of weird tx log data" );
```
chrisvest: Maybe call the data `indecipherable` instead of `weird`?

chrisvest (approved):
Looks good. Just a couple of comments.

PR #9046, opened 2017-03-16
NODE KEY constraint as full stack constraint combining features of uniqueness and existence constraints

craigtaverner:
This work builds on, but replaces parts of, https://github.com/neo4j/neo4j/pull/9034

Changelog: Composite uniqueness and NODE KEY

* The existing uniqueness constraint has been generalized to multiple properties
* A new NODE KEY constraint has been developed combining the features of a composite uniqueness constraint and property existence constraint over one or more properties

### DIFF
--- a/community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
+++ b/community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
@@ -37,8 +37,6 @@ import org.neo4j.kernel.impl.store.record.PropertyKeyTokenRecord;
 import org.neo4j.kernel.impl.store.record.RelationshipTypeTokenRecord;
 import org.neo4j.storageengine.api.schema.SchemaRule;
 
-import static org.neo4j.kernel.api.schema_new.constaints.ConstraintDescriptor.Type.UNIQUE;
-
 /**
  * Note that this class builds up an in-memory representation of the complete schema store by being used in
  * multiple phases.
@@ -156,7 +154,7 @@ public class SchemaRecordCheck implements RecordCheck<DynamicRecord, Consistency
         {
             checkSchema( rule, record, records, engine );
 
-            if ( rule.getConstraintDescriptor().type() == UNIQUE )
+            if ( rule.getConstraintDescriptor().type().enforcesUniqueness() )
             {
                 DynamicRecord previousObligation = indexObligations.put( rule.getOwnedIndex(), record.clone() );
                 if ( previousObligation != null )
@@ -201,7 +199,7 @@ public class SchemaRecordCheck implements RecordCheck<DynamicRecord, Consistency
         public void checkConstraintRule( ConstraintRule rule, DynamicRecord record,
                 RecordAccess records, CheckerEngine<DynamicRecord,ConsistencyReport.SchemaConsistencyReport> engine )
         {
-            if ( rule.getConstraintDescriptor().type() == UNIQUE )
+            if ( rule.getConstraintDescriptor().type().enforcesUniqueness() )
             {
                 DynamicRecord obligation = constraintObligations.get( rule.getId() );
                 if ( obligation == null )
--- a/community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
+++ b/community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
@@ -154,7 +154,7 @@ public class SchemaRecordCheck implements RecordCheck<DynamicRecord, Consistency
         {
             checkSchema( rule, record, records, engine );
 
-            if ( rule.getConstraintDescriptor().type().enforcesUniqueness() )
+            if ( rule.getConstraintDescriptor().enforcesUniqueness() )
             {
                 DynamicRecord previousObligation = indexObligations.put( rule.getOwnedIndex(), record.clone() );
                 if ( previousObligation != null )
@@ -199,7 +199,7 @@ public class SchemaRecordCheck implements RecordCheck<DynamicRecord, Consistency
         public void checkConstraintRule( ConstraintRule rule, DynamicRecord record,
                 RecordAccess records, CheckerEngine<DynamicRecord,ConsistencyReport.SchemaConsistencyReport> engine )
         {
-            if ( rule.getConstraintDescriptor().type().enforcesUniqueness() )
+            if ( rule.getConstraintDescriptor().enforcesUniqueness() )
             {
                 DynamicRecord obligation = constraintObligations.get( rule.getId() );
                 if ( obligation == null )
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/LabelSchemaDescriptor.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/LabelSchemaDescriptor.java
@@ -102,29 +102,4 @@ public class LabelSchemaDescriptor implements SchemaDescriptor, LabelSchemaSuppl
     {
         return this;
     }
-
-    public int compareTo( SchemaDescriptor o )
-    {
-        if ( o instanceof LabelSchemaDescriptor )
-        {
-            LabelSchemaDescriptor other = (LabelSchemaDescriptor) o;
-            if ( labelId == other.getLabelId() )
-            {
-                return SchemaUtil.comparePropertyKeyIds( propertyIds, other.getPropertyIds() );
-            }
-            else
-            {
-                return labelId - other.labelId;
-            }
-        }
-        else
-        {
-            return -1;
-        }
-    }
-
-    public interface Supplier extends SchemaDescriptor.Supplier
-    {
-        LabelSchemaDescriptor schema();
-    }
 }
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/LabelSchemaDescriptor.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/LabelSchemaDescriptor.java
@@ -102,4 +102,24 @@ public class LabelSchemaDescriptor implements SchemaDescriptor, LabelSchemaSuppl
     {
         return this;
     }
+
+    public int compareTo( SchemaDescriptor o )
+    {
+        if ( o instanceof LabelSchemaDescriptor )
+        {
+            LabelSchemaDescriptor other = (LabelSchemaDescriptor) o;
+            if ( labelId == other.getLabelId() )
+            {
+                return SchemaUtil.comparePropertyKeyIds( propertyIds, other.getPropertyIds() );
+            }
+            else
+            {
+                return labelId - other.labelId;
+            }
+        }
+        else
+        {
+            return -1;
+        }
+    }
 }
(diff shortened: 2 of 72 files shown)


community/consistency-check/src/main/java/org/neo4j/consistency/checking/SchemaRecordCheck.java
```diff
@@ -156,7 +154,7 @@ public void checkConstraintRule( ConstraintRule rule, DynamicRecord record,
         {
             checkSchema( rule, record, records, engine );
 
-            if ( rule.getConstraintDescriptor().type() == UNIQUE )
+            if ( rule.getConstraintDescriptor().type().enforcesUniqueness() )
```
thobe: I love the readability improvement here. We might actually be able to improve it even further, by having the `ConstraintDescriptor` have an `enforcesUniqueness()` method (and an `enforcesPropertyExistence()` method) that delegates to the state of the `type` that it holds.

In fact, if we make that change the `ConstraintDescriptor.Type` enum does not need a public `enforcesUniqueness()` or `enforcesPropertyExistence()` method, since all places that use these methods follows the pattern of having a `ConstraintDescriptor` and invoking the `type()` method of that in order to invoke the `enforcesUniqueness()` or `enforcesPropertyExistence()` method on the result. In fact the `ConstraintDescriptor.Type` wouldn't need the methods at all, since the `ConstraintDescriptor` can access the fields of its inner class directly.

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/LabelSchemaDescriptor.java
```diff
@@ -97,6 +97,27 @@ public String toString()
...
+                return SchemaUtil.comparePropertyKeyIds( propertyIds, other.getPropertyIds() );
+            }
+            else
+            {
+                return labelId - other.labelId;
+            }
+        }
+        else
+        {
+            return -1;
```
thobe: This is not a valid implementation of the `Comparable` interface.

If comparing a `LabelSchemaDescriptor` to a `RelationTypeSchemaDescriptor` (which are comparable according to the type hierarchy, since they are both `SchemaDescriptor`), they would be considered out-of-order (`-1`) regardless of which order they were compared in. That is wrong.

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/RelationTypeSchemaDescriptor.java
```diff
@@ -92,4 +92,25 @@ public int hashCode()
...
+                return SchemaUtil.comparePropertyKeyIds( propertyIds, other.getPropertyIds() );
+            }
+            else
+            {
+                return relTypeId - other.getRelTypeId();
+            }
+        }
+        else
+        {
+            return -1;
```
thobe: Invalid implementation of `Comparable`, cannot return `-1` as a default like this.

See comment on `LabelSchemaDescriptor`.

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/SchemaDescriptor.java
```diff
@@ -33,7 +33,7 @@
  * how this is done in eg. LabelSchemaDescriptor, and the SchemaProcessor and SchemaComputer interfaces need to be
  * extended with methods taking the new concrete type as argument.
  */
-public interface SchemaDescriptor
+public interface SchemaDescriptor extends Comparable<SchemaDescriptor>
```
thobe: Where do we actually sort `SchemaDescriptor` objects?

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/SchemaUtil.java
```diff
@@ -31,10 +31,15 @@ private SchemaUtil()
...
+
+    public static String niceProperties( TokenNameLookup tokenNameLookup, int[] propertyIds, String prefix )
+    {
         String[] properties = new String[propertyIds.length];
         for ( int i = 0; i < propertyIds.length; i++ )
         {
-            properties[i] = tokenNameLookup.propertyKeyGetName( propertyIds[i] );
+            properties[i] = prefix + tokenNameLookup.propertyKeyGetName( propertyIds[i] );
         }
         return String.join( ", ", properties );
```
thobe: This makes me cringe a bit. We allocate so many strings here, as well as an array that we don't really need.

This code could have been:
```
StringBuilder properties = new StringBuilder();
for ( int i = 0; i < propertyIds.length; i++ ) {
    if ( i > 0 ) { properties.append(", "); }
    properties.append( prefix ).append( tokenNameLookup.propertyKeyGetName( propertyIds[i] ) );
}
return properties.toString();
```

It gets even worse when you look at where this code is actually used, those places in the code build strings by a combination of concatenation and string formatting, where a simple direct use of a `StringBuilder` would have been just as readable, and would perform fewer allocations. Then that `StringBuilder` could be passed in to this method...

Not a biggie, just a default thing I react on.

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/SchemaUtil.java
```diff
@@ -59,4 +64,23 @@ public String propertyKeyGetName( int propertyKeyId )
...
+                {
+                    return these[i] - those[i];
+                }
+            }
+            return 0;
+        }
+        else
+        {
+            return these.length - those.length;
+        }
```
thobe: Is ordering by length first really the intention?

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/constaints/ConstraintDescriptor.java
```diff
@@ -32,8 +32,28 @@
 {
     public enum Type
     {
-        UNIQUE,
-        EXISTS
+        UNIQUE( true, false ),
+        EXISTS( false, true ),
+        UNIQUE_EXISTS( true, true );
```
thobe, line 37: Should we perhaps call this `KEY` instead of `UNIQUE_EXISTS`, since that is what we call it in the rest of the codebase...

community/kernel/src/main/java/org/neo4j/kernel/impl/store/record/SchemaRuleSerialization.java
```diff
@@ -50,7 +51,7 @@
     private static final byte GENERAL_INDEX = 31, UNIQUE_INDEX = 32;
 
     // Constraint type
-    private static final byte EXISTS_CONSTRAINT = 61, UNIQUE_CONSTRAINT = 62;
+    private static final byte EXISTS_CONSTRAINT = 61, UNIQUE_CONSTRAINT = 62, UNIQUE_EXISTS_CONSTRAINT = 63;
```
thobe, line 54: Why not call it `KEY_CONSTRAINT` instead of `UNIQUE_EXISTS_CONSTRAINT`?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/EnterpriseConstraintSemantics.java
```diff
@@ -42,21 +43,27 @@
...
 {
     @Override
-    protected ConstraintDescriptor readNonStandardConstraint( ConstraintRule rule )
+    protected ConstraintDescriptor readNonStandardConstraint( ConstraintRule rule, String errorMessage )
     {
-        if ( rule.getConstraintDescriptor().type() != EXISTS )
+        if ( !rule.getConstraintDescriptor().type().enforcesPropertyExistence() )
         {
             throw new IllegalStateException( "Unsupported constraint type: " + rule );
         }
```
thobe: Do we really need to check anything here? Surely the enterprise edition supports all constraints?

thobe (changes requested):
In general this looks good. The only thing I would really like to see changed is the implementation of `Comparable` in `SchemaDescriptor`.

There is one additional change I would like to see, in addition to what has been done in this PR, and that is to change how the `PropertyExistenceEnforcer` is implemented. Instead of containing lists of schema descriptors, I think it should be changed to have a map from labels with existence constraints to the properties that are mandatory for that label.

community/kernel/src/main/java/org/neo4j/kernel/api/schema_new/constaints/ConstraintDescriptor.java
chrisvest: I think `UNIQUE_EXISTS` is descriptive of the behaviour while "key" is an overloaded term, so I think this is fine as is.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/DataIntegrityValidatingStatementOperations.java
```diff
@@ -204,7 +221,7 @@ public void constraintDrop( KernelStatement state, ConstraintDescriptor descript
         schemaWriteDelegate.constraintDrop( state, descriptor );
     }
 
-    private void assertIndexDoesNotExist( KernelStatement state, OperationContext context,
+    private void assertUniqueIndexDoesNotExist( KernelStatement state, OperationContext context,
```
chrisvest: This method will throw regardless of whether the existing index is used for uniqueness constraints or not, won't it?

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/TransactionToRecordStateVisitor.java
```diff
@@ -202,23 +204,40 @@ public void visitAddedConstraint( ConstraintDescriptor constraint ) throws Creat
...
+    {
+        IndexRule indexRule = schemaStorage.indexGetForSchema( uniqueConstraint.ownedIndexDescriptor() );
+        recordState.createSchemaRule( constraintSemantics.createUniquenessConstraintRule(
+                constraintId, uniqueConstraint, indexRule.getId() ) );
+        recordState.setConstraintIndexOwner( indexRule, constraintId );
+    }
+
+    private void visitAddedNodeKeyConstraint(NodeKeyConstraintDescriptor uniqueConstraint, long constraintId)
+            throws CreateConstraintFailureException
+    {
```
chrisvest: These two method declarations looks like they could use a reformat.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/record/SchemaRuleDeserializer2_0to3_1.java
```diff
@@ -138,6 +138,13 @@ public static ConstraintRule readUniquenessConstraintRule( long id, int labelId,
                 readOwnedIndexRule( buffer ) );
     }
 
+    public static ConstraintRule readNodeKeyConstraintRule( long id, int labelId, ByteBuffer buffer )
+    {
+        return new ConstraintRule( id,
+                ConstraintDescriptorFactory.nodeKeyForLabel( labelId, readConstraintPropertyKeys( buffer ) ),
+                readOwnedIndexRule( buffer ) );
+    }
```
chrisvest: Where is this used? This class is about deserialising legacy schema rule records, I think.

Also, I don't know why this isn't already the case, but the `read*` methods should assign their result to a local variable, before passing the values into any other methods. Otherwise, a refactoring that changes the parameter order will end up (almost certainly unintentionally) implying a format change.

chrisvest (approved):
There's a couple of things I don't think are quite right, but in general I like this.

thobe (approved):
Looks good in general, but as @chrisvest pointed out, there are stylistic problems that are so bad that even the build fails.

PR #9056, opened 2017-03-19
Overwriting property with same value no longer generates write command

tinwelint:
It seems obvious, but setting a node/relationship property to the same value
as it already is doesn't require any change to that node/relationship.
Even so this was the case previously. A command would be generated, property
record would be updates on apply, indexing changes would be applied...
all for nothing.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
@@ -1034,7 +1034,12 @@ public class StateHandlingStatementOperations implements
         Property existingProperty = existingPropertyValue == null ?
                 Property.noGraphProperty( property.propertyKeyId() ) :
                 Property.property( property.propertyKeyId(), existingPropertyValue );
-        state.txState().graphDoReplaceProperty( existingProperty, property );
+
+        if ( !property.equals( existingProperty ) )
+        {
+            state.txState().graphDoReplaceProperty( existingProperty, property );
+        }
+
         return existingProperty;
     }
 
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
@@ -964,10 +964,12 @@ public class StateHandlingStatementOperations implements
             NodeItem node = cursor.get();
             Property existingProperty = readProperty( property.propertyKeyId(), node, EntityType.NODE );
 
+            // TODO Explicitly auto-index properties, even if setting the same value as the existing one.
+            // This because there are established ways of rebuilding auto-indexes which involves this function.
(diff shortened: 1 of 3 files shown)


tinwelint:
Discussing whether or not legacy auto indexes should keep getting notified even on same value updates, due to there being established ways of reindexing auto indexes by re-setting their property values.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
```diff
@@ -960,27 +962,44 @@ public Property nodeSetProperty( KernelStatement state, long nodeId, DefinedProp
         try ( Cursor<NodeItem> cursor = nodeCursorById( state, nodeId ) )
         {
             NodeItem node = cursor.get();
-            Property existingProperty;
-            try ( Cursor<PropertyItem> properties = node.property( property.propertyKeyId() ) )
+            Property existingProperty = readProperty( property.propertyKeyId(), node, EntityType.NODE );
+
+            // TODO Explicitly auto-index properties, even if setting the same value as the existing one.
```
MishaDemianenko, line 967: Should it really be todo thing here? would we ever change it (except removing it when removing legacy indexing)?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
```diff
@@ -960,27 +962,44 @@ public Property nodeSetProperty( KernelStatement state, long nodeId, DefinedProp
...
+            // This because there are established ways of rebuilding auto-indexes which involves this function.
+            autoIndexProperty( nodeId, property, ops, existingProperty, autoIndexing.nodes() );
+
+            if ( !property.equals( existingProperty ) )
```
MishaDemianenko, line 971: We also have `graphSetProperty` which is not that widely used, as we know, but shouldn't that be updated as well with similar check?

MishaDemianenko (commented):
Couple of minor comments, great otherwise!

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
tinwelint: What I'd like to see is to build tooling for rebuilding auto indexes and as part of that deal with this TODO. The current means of rebuilding those is a hack and bloats the transaction log.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
tinwelint: That's true, I'll fix that

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
MishaDemianenko: Ok, lets create a card for that then?

MishaDemianenko (approved):

PR #9059, opened 2017-03-20
DBstructure support for composite indexes

ragadeeshu:


community/kernel/src/main/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureCollector.java
```diff
@@ -73,41 +73,42 @@ public DbStructureLookup lookup()
...
+            public Iterator<Pair<String,String[]>> knownUniqueIndices()
             {
                 return uniqueIndices.iterator();
             }
 
             @Override
-            public Iterator<Pair<String, String>> knownUniqueConstraints()
+            public Iterator<Pair<String,String[]>> knownUniqueConstraints()
             {
                 //TODO: Add support for composite indexes
```
burqen: TODO resolved?

community/kernel/src/main/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureCollector.java
```diff
@@ -353,12 +352,12 @@ public boolean hasNext()
...
 
                 @Override
-                public Pair<String, String> next()
+                public Pair<String,String[]> next()
                 {
                     //TODO: Add support for composite indexes
                     LabelSchemaDescriptor next = iterator.next();
                     String label = labels.byIdOrFail( next.getLabelId() );
-                    String propertyKey = propertyKeys.byIdOrFail( next.getPropertyIds()[0] );
+                    String[] propertyKey = propertyKeys.byIdOrFail( next.getPropertyIds() );
```
burqen: May I suggest updating variable name to `propertyKeyNames` (like what you use in `knownNodePropertyExistenceConstraints`)

community/kernel/src/test/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureArgumentFormatterTest.java
```diff
@@ -72,8 +72,16 @@ public void shouldFormatIndexDescriptors()
...
+        assertEquals( "ConstraintDescriptorFactory.uniqueForLabel( 23, 42 )",
+                formatArgument(
+                        ConstraintDescriptorFactory.uniqueForLabel( 23, 42 ) ) );
+    }
+
+    @Test
+    public void shouldFormatCompositeUniquenessConstraints()
+    {
+        assertEquals( "ConstraintDescriptorFactory.uniqueForLabel( 23, 42, 43 )",
+                formatArgument( ConstraintDescriptorFactory.uniqueForLabel( 23, 42, 43 ) ) );
```
burqen, line 84: Don't we still want to test `uniqueForSchema`?

burqen (changes requested):
Got some small comments, look good otherwise

community/kernel/src/test/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureArgumentFormatterTest.java
ragadeeshu: No. You have to remember that `ConstraintDescriptorFactory.uniqueForLabel( 23, 42, 43 ) )` is computed to `ConstraintDescriptor` before it is sent down to the method, so whenever we create it via a schema or with the labels directly does not matter. Thus I choose to do it this way, since it matches the result better :)

burqen (approved):

community/kernel/src/test/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureArgumentFormatterTest.java
craigtaverner: Aren't we missing a test for `nodeKeyForLabel` here?

ragadeeshu: That did not exist when this PR was created.

PR #9060, opened 2017-03-20
Added new role `Editor`

OliviaYtterbrink:
Added new role `Editor`, reverted `Publisher` role.
Removed setting to control if `Publisher` is allowed to create new tokens or not, instead `Publisher` is always allowed to create new tokens.
Added new role `Editor` that is allowed to read and write data but not allowed to create new tokens.
Procedures flagged with `Mode=WRITE` are always allowed to create new tokens, given that the user executing them is allowed.

changelog: Added new role `Editor`

* Allowed to read and write data but not create new tokens

Changed  role `Publisher`

* Allowed to read and write data and create new tokens


OliviaYtterbrink:
Replaced by https://github.com/neo4j/neo4j/pull/9081

PR #9061, opened 2017-03-20
Composite support for indexProcedures

ragadeeshu:


community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/IndexSpecifier.java
```diff
@@ -58,21 +58,31 @@ public String toString()
...
         }
-        return Pair.of(
-                either( matcher.group( "simpleLabel" ), matcher.group( "complexLabel" ) ),
-                either( matcher.group( "simpleProperty" ), matcher.group( "complexProperty" ) )
-        );
+        int group = 0;
+        String label = either( matcher.group( ++group ), matcher.group( ++group ) );
+        ArrayList<String> properties = new ArrayList<>();
+
+        while ( group + 2 <= matcher.groupCount() )
```
ragadeeshu: There is probably a better way to do this.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/IndexSpecifier.java
```diff
@@ -82,16 +92,16 @@ private String either( String first, String second )
 
     private static String or( String first, String second )
     {
-        return "(:?" + first + "|" + second + ")";
+        return "(?:" + first + "|" + second + ")";
```
chrisvest, line 94: Subtle…

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/IndexSpecifier.java
```diff
@@ -82,16 +92,16 @@ private String either( String first, String second )
...
     {
-        return format( "(?<%s>[A-Za-z0-9_]+)", name );
+        return "([A-Za-z0-9_]+)";
     }
 
-    private static String complexIdentifier( String name )
+    private static String qoutedIdentifier()
     {
-        return format( "(?:`(?<%s>[^`]+)`)", name );
+        return "(?:`([^`]+)`)";
```
chrisvest: Quoted identifier creates a non-capturing group, but the identifier creates a capturing group…

I _suspect_ that more tests can be written for this.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/IndexSpecifier.java
```diff
@@ -58,21 +58,31 @@ public String toString()
...
-    private Pair<String, String> parse()
+    private Pair<String,String[]> parse()
     {
-        Pattern pattern = Pattern.compile( "" +
-                ":" + or( simpleIdentifier( "simpleLabel" ), complexIdentifier( "complexLabel" ) ) +
-                "\\(" + or( simpleIdentifier( "simpleProperty" ), complexIdentifier( "complexProperty" ) ) + "\\)"
+        Pattern pattern = Pattern.compile(
+                ":" + or( identifier(), qoutedIdentifier() ) + // Match the label
+                "\\(" + or( identifier(), qoutedIdentifier() ) + // Match the first property
+                "(?:\\,\\s" + or( identifier(), qoutedIdentifier() ) + ")*\\)" // Match following properties
```
chrisvest: You probably a want a `*` after the `\\s` for a more flexible syntax. Looking at the legacy open Cypher specification, space is also allowed between the `:` and the label identifier when creating indexes, though not when doing index lookups via the `START` clause.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/IndexSpecifier.java
chrisvest: Oh, I see the capturing group is inside the back-ticks.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/IndexSpecifier.java
```diff
@@ -58,21 +58,31 @@ public String toString()
...
+
+        while ( group + 2 <= matcher.groupCount() )
+        {
+            String property = either( matcher.group( ++group ), matcher.group( ++group ) );
+            if ( property != null )
+            {
+                properties.add( property );
+            }
+        }
+        return Pair.of( label, properties.toArray( new String[0] ) );
```
chrisvest: This code is not easy to follow. It is tightly coupled with how the capturing and non-capturing groups are organised in the regular expression. Some comments would be nice.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/IndexSpecifier.java
```diff
@@ -58,21 +59,28 @@ public String toString()
...
+    private Pair<String,String[]> parse()
     {
-        Pattern pattern = Pattern.compile( "" +
-                ":" + or( simpleIdentifier( "simpleLabel" ), complexIdentifier( "complexLabel" ) ) +
-                "\\(" + or( simpleIdentifier( "simpleProperty" ), complexIdentifier( "complexProperty" ) ) + "\\)"
+        // Note that this now matches all properties in a single group, in order to split them later.
+        Pattern pattern = Pattern.compile(
+                ":" + or( identifier(true), qoutedIdentifier(true) ) + // Match the label
+                "\\((" + or( identifier(false), qoutedIdentifier(false) ) + // Match the first property
+                "(?:\\,\\s" + or( identifier(false), qoutedIdentifier(false) ) + ")*)\\)" // Match following properties
```
chrisvest: You probably a want a `*` after the `\\s` for a more flexible syntax. Looking at the legacy open Cypher specification, space is also allowed between the `:` and the label identifier when creating indexes, though not when doing index lookups via the `START` clause.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/IndexSpecifier.java
```diff
@@ -58,21 +59,28 @@ public String toString()
...
             throw new IllegalArgumentException( "Cannot parse index specification " + specification );
         }
-        return Pair.of(
-                either( matcher.group( "simpleLabel" ), matcher.group( "complexLabel" ) ),
-                either( matcher.group( "simpleProperty" ), matcher.group( "complexProperty" ) )
-        );
+        String label = either( matcher.group( 1 ), matcher.group( 2 ) );
+        String propertyString = matcher.group( 3 );
+        //Split string on commas, but ignore commas in quotes
+        String[] properties = propertyString.split(",\\s(?=(?:`[^`]*`)|[^`]*$)", -1);
```
chrisvest: Likewise here you probably want a `*` after the `\\s`.
Also note that according to open Cypher, backticks can be escaped with double backticks.
I also don't see a test that involves a comma in a quoted identifier.
And why the `-1` instead of `0` to `String.split`?

chrisvest (commented):
I have questions about the regex parsing of index specifiers.

PR #9062, opened 2017-03-20
Full pruning var expand

fickludd:
New Cypher operator for improved handling of longer var-length expands with distinct start node / end node combinations. This new operator is called FullPruningVarExpand, and is planned for var-lengths above `4..5`. Above this length the FullPruningVarExpand gives improved execution times in most cases, with speed-ups of an order of magnitude or more on even longer var-lengths. 

changelog: Improved cypher execution speed on variable length queries where only the distinct pairs of start and end node are of interest. One query to benefit would be `MATCH (a)-[*4..5]->(b) RETURN DISTINCT a, b`.

### DIFF
--- a/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/pipes/FullPruningVarLengthExpandPipe.scala
+++ b/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/pipes/FullPruningVarLengthExpandPipe.scala
@@ -388,8 +388,6 @@ case class FullPruningVarLengthExpandPipe(source: Pipe,
   }
 
   override protected def internalCreateResults(input: Iterator[ExecutionContext], state: QueryState): Iterator[ExecutionContext] = {
-    state.decorator.registerParentPipe(this)
-
     new FullyPruningIterator(input, state)
   }
 }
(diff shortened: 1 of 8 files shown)


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/pipes/FullPruningVarLengthExpandPipe.scala
```diff
@@ -0,0 +1,395 @@
...
+                                          max: Int,
+                                          filteringStep: VarLengthPredicate = VarLengthPredicate.NONE)
+                                         (val id: Id = new Id)
+                                         (implicit pipeMonitor: PipeMonitor) extends PipeWithSource(source, pipeMonitor) with Pipe {
+  self =>
+
+  assert(min <= max)
+
+  /*
+  This algorithm has been implemented using a state machine. This Cypher statement shows the static connections between the states.
```
systay: Really?

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/pipes/FullPruningVarLengthExpandPipe.scala
```diff
@@ -0,0 +1,395 @@
...
+    var isEmitted = false
+
+    /**
+      * Computes the minimum outgoing full expanded depth.
+      *
+      * @param incomingRelId id of the incoming relationship, which is omitted from this computation
+      * @return the full expand depth
+      */
+    def minOutgoingDepth(incomingRelId: Long): Int = {
+      var max = Constants.VERY_BIG_VALUE
```
systay: Maybe call this variable `min` instead?

systay: How about using `Integer.MAX_VALUE` instead of this arbitrarily chosen `VERY_BIG_VALUE`?

This is the only place `VERY_BIG_VALUE` is used, so if that is changed, we no longer need to declare that constant - making the code easier to read (because a reader no longer needs to try to understand why the value is what it is).

PR #9063, opened 2017-03-21
More scalable index logging

tinwelint:
Logging for indexing has been a bit verbose, but isn't a problem for scenarios where there are few indexes. Excessive logging includes:

- index state (ONLINE/POPULATING/FAILED) during init
- index state during start
- sampling recovery at startup
- sampling at runtime
- listing of index files during startup and log rotation

In scenarios with thousands of indexes the above logging quickly fills most of the `debug.log`, considering that the biggest one, the file listing, also gets logged on log file rotation. The excessive logging would quickly approach the critical log file rotation threshold to the point where the logging done while doing a log rotation would immediately cause log rotation itself.

Some of the logging is left as it was, although instead as DEBUG and some more scalable INFO logging is done to replace it for most users.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
@@ -738,21 +738,6 @@ public class IndexingService extends LifecycleAdapter
         }
     }
 
-    public void flushAll()
-    {
-        for ( IndexProxy index : indexMapRef.getAllIndexProxies() )
-        {
-            try
-            {
-                index.flush();
-            }
-            catch ( IOException e )
-            {
-                throw new UnderlyingStorageException( "Unable to force " + index, e );
-            }
-        }
-    }
-
     private void closeAllIndexes()
     {
         Iterable<IndexProxy> indexesToStop = indexMapRef.clear();
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
@@ -199,7 +199,7 @@ public class IndexingService extends LifecycleAdapter
             SchemaIndexProvider.Descriptor providerDescriptor = indexRule.getProviderDescriptor();
             SchemaIndexProvider provider = providerMap.apply( providerDescriptor );
             InternalIndexState initialState = provider.getInitialState( indexId );
-            log.info( indexStateInfo( "init", indexId, initialState, descriptor ) );
+            log.debug( indexStateInfo( "init", indexId, initialState, descriptor ) );
             boolean constraint = indexRule.isConstraintIndex();
 
             switch ( initialState )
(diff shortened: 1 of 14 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -233,10 +233,44 @@ public void init()
...
             indexMap.putIndexProxy( indexId, indexProxy );
         }
+        logIndexStateSummary( "init" );
 
         indexMapRef.setIndexMap( indexMap );
     }
 
+    private void logIndexStateSummary( String method )
+    {
+        int[] counts = new int[InternalIndexState.values().length];
```
MishaDemianenko, line 243: If you will create map of states to -> indexes to indexes we can simplify following calculations. For example
if we will do something like:
```
 Map<InternalIndexState,List<IndexRule>> stateListMap =
                StreamSupport.stream( indexRules.spliterator(), false ).collect( Collectors.groupingBy( rule ->
                {
                    SchemaIndexProvider provider = providerMap.apply( rule.getProviderDescriptor() );
                    return provider.getInitialState( rule.getId() );
                } ) );
```
then we have all indexes already grouped by type and should be super simple to print non popular one

community/kernel/src/main/java/org/neo4j/kernel/internal/KernelDiagnostics.java
```diff
@@ -93,15 +95,25 @@ private long logStoreFiles( Logger logger, String prefix, File dir )
...
             // Sort by name
             List<File> fileList = Arrays.asList( files );
-            Collections.sort( fileList, new Comparator<File>()
+            if ( fileList.size() > MAX_DIRECTORY_LISTING_THRESHOLD )
             {
-                @Override
-                public int compare( File o1, File o2 )
+                // If this directory contains lots and lots of files then sort on total size
+                // so that the biggest and to some extent most important ones gets printed below.
+                Collections.sort( fileList, ( o1, o2 ) -> Long.compare( totalSize( o1 ), totalSize( o2 ) ) );
```
MishaDemianenko: ``` fileList.sort( Comparator.comparingLong( this::totalSize ) );```

community/kernel/src/main/java/org/neo4j/kernel/internal/KernelDiagnostics.java
```diff
@@ -93,15 +95,25 @@ private long logStoreFiles( Logger logger, String prefix, File dir )
 
             // Sort by name
             List<File> fileList = Arrays.asList( files );
-            Collections.sort( fileList, new Comparator<File>()
+            if ( fileList.size() > MAX_DIRECTORY_LISTING_THRESHOLD )
```
MishaDemianenko: i think we need to mention in the logs what the order is to avoid our own confusion during investigation:
Like:
```Bla ...bla.. store files in accordance to their store size``` or similar

File: community/kernel/src/main/java/org/neo4j/kernel/internal/KernelDiagnostics.java [hunk 4]
```diff
@@ -122,10 +134,37 @@ public int compare( File o1, File o2 )
                 logger.log( fileInformation );
 
                 total += size;
+                if ( ++count == MAX_DIRECTORY_LISTING_THRESHOLD )
```
MishaDemianenko: i do not think we should do this, yes it will remove some logging but also can consume so important information for future investigation

MishaDemianenko (changes requested):
Couple of comments otherwise good!

MishaDemianenko (approved):

PR #9064, opened 2017-03-21
Always use latest available commit for lucene index snapshoting

MishaDemianenko:
Always use latest available lucene index commit for snapshot operations
to avoid cases when older snapshot is reused in backup/store copying
that can cause index corruption since only transaction from the backup start will be replayed
on top of commit, while older snapshot can represent index state way before that point.

Example scenario:
1. Store copy starts on master.
2. Snapshot(1) of indexes taken.
3. Some transaction closed on master.
4. Another store copy starts from master
5. Snapshot(2) of indexes taken.
Snapshot(2) should always contain latest index snapshot that was taken
when we start doing store copy(4).


community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/MultipleBackupDeletionPolicy.java
```diff
@@ -19,19 +19,14 @@
...
 
-import java.io.IOException;
-
 import org.apache.lucene.index.IndexCommit;
 import org.apache.lucene.index.KeepOnlyLastCommitDeletionPolicy;
 import org.apache.lucene.index.SnapshotDeletionPolicy;
 
+import java.io.IOException;
+
 public class MultipleBackupDeletionPolicy extends SnapshotDeletionPolicy
```
tinwelint: This whole class becomes irrelevant now, doesn't it?

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/MultipleBackupDeletionPolicy.java
```diff
@@ -19,19 +19,14 @@
...
 {
-    static final String SNAPSHOT_ID = "backup";
-
-    private IndexCommit snapshot;
-    private int snapshotUsers;
-
     public MultipleBackupDeletionPolicy()
     {
         super( new KeepOnlyLastCommitDeletionPolicy() );
```
tinwelint: Can we be sure that the initial snapshot keeps the files it lists, even if there are more commits? The class name looks like it will not be the case. I don't think your tests explicitly tests that because file deletion perhaps may not happen with the given load that test has.

community/neo4j/src/test/java/org/neo4j/index/backup/IndexBackupIT.java
```diff
@@ -0,0 +1,146 @@
...
+    {
+        Label label = Label.label( "testLabel" );
+        prepareDatabase( label );
+
+        CheckPointer checkPointer = resolveDependency( CheckPointer.class );
+        IndexingService indexingService = resolveDependency( IndexingService.class );
+
+        forceCheckpoint( checkPointer );
+        ResourceIterator<File> firstCheckpointSnapshot = indexingService.snapshotStoreFiles();
+        generateData( label );
```
tinwelint: How can we enhance this test so that we know that the changes result in lots of Lucene file changes, at least some added, some changed and some deleted?

tinwelint (changes requested):

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/MultipleBackupDeletionPolicy.java
MishaDemianenko: yep cleaned

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/MultipleBackupDeletionPolicy.java
MishaDemianenko: yes, added additional snapshotting to test

community/neo4j/src/test/java/org/neo4j/index/backup/IndexBackupIT.java
MishaDemianenko: test updated and enhanced!

tinwelint (approved):

PR #9065, opened 2017-03-21
Ability to call GBPTree#close() multiple times gracefully

tinwelint:


community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -753,17 +759,31 @@ public void close( Consumer<PageCursor> headerWriter ) throws IOException
 
     private void close( Header.Writer headerWriter ) throws IOException
     {
-        writer.close();
-
+        writerCheckpointMutex.lock();
```
burqen, line 762: So `writer.close()` was not protected from concurrent writes before?
Lets add a test for this!

burqen (changes requested):
Looks good, one small comment

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: It is and there was a test for it already :)

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -289,6 +289,12 @@ default void recoveryCompleted( long numberOfPagesVisited, long numberOfCleanedC
     private final Monitor monitor;
 
     /**
+     * Whether or not this tree has been closed. Accessed and changed solely in
+     * {@link #close()} and {@link #close(Consumer)} to be able to close tree multiple times gracefully.
+     */
+    private volatile boolean closed;
```
davidegrohmann: Is the volatile really necessary since you access `close` under a lock?

tinwelint: Best-case, the lock is an `unsafe.compareAndSwapInt` which has a javadoc mentioning that it has memory semantics of a volatile read and write. So I'd say it's not necessary to have it volatile. It's a minor thing and only checked in close, but can do this change anyway.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: Tests added. Content with those?

burqen: EDIT: NVM, forgot to pull latest commits. It's fine!

I can't get test to fail locally by removing the lock from close. Seems like test is not testing what it should.
Although, this might not be relevant. The PR suggest we should not checkpoint at all during close https://github.com/neo4j/neo4j/pull/9028

burqen (approved):

PR #9068, opened 2017-03-21
Cluster has ability to recover from half-applied constraint

tinwelint:
Creating a constraint involves two transactions:
- one that creates the backing constraint index, where index population starts once committed and applied
- one that creates and activates the constraint after successful index population

There was a problem in distributed environments where this could happen:
- constraint would begin being created on a cluster instance A, i.e. the first transaction creating the constraint index
- another cluster instance B would pull and apply this transaction
- B would crash
- A would complete the constraint creation by commiting the second transaction activating the constraint
- B would be started again, where recovery would treat this half-baked constraint as an unsuccessful constraint creation and mark it as failed.
  (single mode differs here where the constraint index would even be deleted)
- when B would continue to pull transactions from A and eventually get to the transaction activating the constraint, B would not be able to apply it because the backing index had been marked as failed during recovery

Fix proposed here is to not try and be helpful and delete these "orphaned" constraint indexes during startup because even if that only happened in single db mode then an instance may switch back and forth between clustered and single, making this hard to look consistent to user.

The other part of the fix is to remove the optimization to not wait and immediately mark such constraint indexes as failed during recovery, now that they will not be deleted.

Additionally instead of cleaning up orphaned constraint indexes, such indexes can be reused for another constraint. So a retry of such a transaction after a crash would work, just as it would before this change.
Having the ability to reuse orphaned constraint indexes can have multiple concurrent transactions wanting to reuse it to fight over it. One of them will win and the others will fail, this is concurrently safe because of the schema write lock when creating the second transaction that activates the constraint.
changelog: [3.1] Cluster has ability to recover from half-applied constraint


thobe:
>  this is concurrently safe because of the schema write lock when creating the second transaction that activates the constraint.

[`ConstraintIndexCreator` (which performs the inner transaction) does not actually acquire a lock while creating the index](https://github.com/neo4j/neo4j/blob/3.2/community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/ConstraintIndexCreator.java#L71) - it relies on the outer transaction holding that lock already.

(I might be misreading what you mean with that sentence though)

thobe:
There are some test failures. The `NullPointerException` ones look concerning, but the `ConstraintRecoveryIT` might just be a case of the semantics having changed, so the test cases needs to be updated.

thobe (changes requested):
I would like to see a test case where the database crashes during index population, then when it is resumed a duplicate node is found - making the index non-unique.

What state does the index enter then?

tinwelint:
`ConstraintIndexCreator#createUniquenessConstraintIndex` is called when having the schema write lock, so `createConstraintIndex` is called when having schema write lock. It's temporarily released and then acquired after population again, and so the second transaction (the parent transaction) will have the schema write lock when exiting the method and the executing thread will move on to telling tx state about the constraint to create (which is activated when that tx is applied).

At least, this is my understanding.

tinwelint:
I'll take a look at the test failures and a good idea about your test suggestion.

tinwelint:
@thobe while adding that testing an interesting thing arose where the index creation gets recovered and its index built. Furthermore there are transactions violating the (future to-be-created constraint). When the transaction which creates and activates the constraint gets applied it won't notice that the index isn't unique, because it didn't have to previously.

I believe this is mostly a theoretical issue because there's no reason that there would be a violating transaction before the constraint creation, because that would mean that the master serving those updates to be applied would also have a violated constraint online. WDYT about it?

thobe:
@tinwelint yes, it sounds like a theoretical case. I don't think it can happen in reality for the reasons you state.

tinwelint:
@thobe so are you OK with the changes in this PR? Noticed your `requested changes` status still being there...

thobe:
@tinwelint I'm ok with that analysis. I'm reading through the complete changes right now.

thobe:
I still don't see the test case I've been asking for:

- Start with a database that contains duplicates.
- Create a constraint.
- While the index is populating, but before the duplicates have been found *crash*.
- Handle the duplicates during recovery.

@tinwelint can you point out where that is tested?

tinwelint:
@thobe what would you consider happening in your last point `Handle the duplicates during recovery.`?

Index population nowadays doesn't verify uniqueness constraint, that is done when creating a constrain, in between index population and actually creating the constraint rule. So per definition index population cannot find duplicates.

The database will recover this index like any other index because it won't consider uniqueness because there's no constraint. The HA scenario (at crashed slave) would be that the constraint would be attempted to be reapplied, but that's a fictional scenario since the constraint will not have been created because the index wasn't unique. Is it this scenario you're thinking about? If so then testing shows that the uniqueness won't be checked on reapply, which hasn't changed from before this change.

If you have something else in mind then please provide details about such testing. In the meantime I'll add some more test cases nonetheless.

tinwelint:
Also I'm still unsure about the general solution to this problem. What's your take on leaving constraint indexes in the db like this?

thobe:
My general take on leaving the indexes around is that it seems useful to do so *iff* we are able to use those indexes. Primarily then if we, when trying to re-create the constraint, can use the index from a previously failed attempt to create a constraint to back a constraint in a second attempt.

That should answer the question you raised before as well, @tinwelint, about what should happen in the point about `Handle the duplicates during recovery`: If there are duplicates in the index, that are not discovered before recovery, it should not be possible to create a constraint, and perhaps the desired behaviour should then be to remove the index (since the index was created to back the constraint) - although I don't really care if that (the detection of duplicates and the reaction to that condition) happens as part of the index building that follows recovery or as part of the re-attempt to create the constraint on top of that index.

The reason I said that the duplicate should be handled during recovery is because the way it was designed when we first implemented these indexes, it was the actual index population phase that handled duplicate detection, not a separate explicit phase that took place after the index was fully populated. If this has changed, I was not aware of that change. So technically you are correct, the recovery phase would not detect those duplicates, but it would need to initialize the population of the recovered index such that it does duplicate detection (unless the logic for duplicate detection has been changed), because last I looked, those indexes were initialized the same as any other index during recovery, since the assumption was that they would be deleted before the index populator was started anyhow.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/constraints/HalfAppliedConstraintRecoveryIT.java
```diff
@@ -0,0 +1,271 @@
...
+        }
+    }
+
+    private GraphDatabaseAPI newDb()
+    {
+        return (GraphDatabaseAPI) new TestGraphDatabaseFactory().setFileSystem( fs ).setMonitors( monitors )
+                .newImpermanentDatabase();
+    }
+
+
```
thobe: Checkstyle is barfing on these double blank lines.

tinwelint:
@thobe OK then I see what you mean and the last test I added in `HalfAppliedConstraintRecoveryIT` checks that exactly that works. I do think we're good here then... pending the checkstyle of course.

thobe (approved):
Ok, cool, looks good then.

PR #9072, opened 2017-03-22
Faster composite lucene population

fickludd:
Following the composite index work, a performance regression was seen in index population. This PR halves the size of that regression by implementing faster versions of 
 * how strings are encoded for value sampling
 * how thread local lucene `Field` objects are tracked and reused 

The slowdown is improved from 1.23x to 1.11x for small_string values, and from 1.18x to 1.05x for long values.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
@@ -47,11 +47,7 @@ import org.apache.lucene.util.NumericUtils;
 import org.apache.lucene.util.StringHelper;
 
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.HashMap;
 import java.util.Iterator;
-import java.util.Map;
-import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


MishaDemianenko:
@fickludd can you please add here at least your benchmark result - no reason to go to slack to look into it, not everybody have access to that channel, it will be hard to find that in future. Thx :)

fickludd:
@MishaDemianenko added that to description

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/LuceneDocumentStructure.java
```diff
@@ -258,7 +260,7 @@ protected AcceptStatus accept( BytesRef term ) throws IOException
         private final Field idField;
         private final Field idValueField;
 
-        private final Map<String,Field> valueFields = new HashMap<>();
+        private Field[] reusableValueFields = new Field[0];
```
ragadeeshu, line 263: If this works - why on earth didn't we do it from the beginning?

fickludd: Objection your honor! Premature optimization.

(I at least had no idea of how hot that spot was after just having dipped into Lucene for the first time. Now we know)

PR #9076, opened 2017-03-22
throw exception if policy does not exist

martinfurmanski:
The plugins can now throw procedure exceptions and the server policies
plugin will throw a ProcedureCallFailed if the supplied policy name
does not exist.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/plugins/server_policies/Policies.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/plugins/server_policies/Policies.java
@@ -23,6 +23,8 @@ import java.util.HashMap;
 import java.util.Map;
 
 import org.neo4j.causalclustering.load_balancing.filters.IdentityFilter;
+import org.neo4j.kernel.api.exceptions.ProcedureException;
+import org.neo4j.kernel.api.exceptions.Status;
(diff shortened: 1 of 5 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/load_balancing/plugins/server_policies/Policies.java
```diff
@@ -31,7 +33,7 @@
 {
     public static final String POLICY_KEY = "load_balancing.policy"; // TODO: move somewhere (driver support package?)
```
zhenlineo, line 34: Can we change this key to something else?
One plan is to pass routing context via bolt url, in we go down that road, then it is really a bad idea to have a `.` in the parameters as it requires a skip in url.

I would vote for `policy` or `load_balancing_policy`.

zhenlineo (changes requested):

digitalstain:
Just a note that this also allows for overriding the default default policy, which is a change in behaviour in addition to the exception being thrown.
Looks good to me.

martinfurmanski:
@digitalstain That is not accurate I think! What you describe is captured by shouldAllowOverridingDefaultPolicy and that has not changed. There is a change in behaviour here however, expressed in the change of other tests!

digitalstain:
@martinfurmanski Correct, my bad. Still looks good to me.

PR #9077, opened 2017-03-22
Add new store version, composite indexes

ragadeeshu:
We don't want people to open new stores with the new indexes with older Neo4j versions.
Also added a test that makes sure that no actual migration takes place.


tinwelint (changes requested):
At first glance it looks like there were a lot of high limit format files copied and I don't see the apparent need for doing so. I may have missed something, but marking as request changes and ask for a description of why it's necessary.

tinwelint (approved):

tinwelint:
For the record: It was decided to keep these copies since it goes in line with previous migrations. We can de-copy these as part of other work if needed.

ragadeeshu:
@tinwelint rebased

tinwelint:
@ragadeeshu compilation error

PR #9078, opened 2017-03-22
Add additional logging for store copy process on master

MishaDemianenko:
Introduce additional logging for store copy from master.
Track start/stop of store copy, files copy, transaction streaming start and finish.
Include thread name into the logs to be able to track parallel full backups,
and full store copies.


chrisvest:
There's at least a failing fragile, over-specifying test that needs to be fixed.

chrisvest (approved):
Looks good; just need that test to be fixed.

MishaDemianenko:
@chrisvest test updated

PR #9086, opened 2017-03-23
Bind to the specified host and not 0.0.0.0

### LINKED ISSUE #9071: causal_clustering.discovery_listen_address doesn't set the IP address
- Neo4j version: 3.1.1
- Operating system: Red Hat 7.3

We generally like to make sure all listening ports are bound to a specific IP for security reasons. This works for almost all neo4j configured addresses except for the discovery port. First I tried setting the connector default, "dbms.connectors.default_listen_address=172.17.0.3", but that did not set the address for the discovery. Nor does setting "causal_clustering.discovery_listen_address=172.17.0.3:5000"

Instead, the discovery port is always bound to ":::5000", meaning you can connect using any IP on the box.

Steps to reproduce:
1) In the neo4j.conf file set
causal_clustering.discovery_listen_address=172.17.0.3:5000
(For your own IP.)
2) bring up Neo4j.

- Expected behavior
netstat -plnt
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name 
...   
tcp6       0      0 **172.17.0.3:5000**         :::*                    LISTEN      2281/java         
...

- Actual behavior
netstat -plnt
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name   
... 
tcp6       0      0 **:::5000**                 :::*                    LISTEN      2281/java           
...

The only debug output IO could find for the parameter shows it is set properly:
/var/log/neo4j/debug.log:2017-03-21 22:15:38.275+0000 INFO  [o.n.k.i.DiagnosticsManager] causal_clustering.discovery_listen_address=172.17.0.3:5000

mneedham:
Have manually checked this while binding to localhost:

Fixes #9071

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastCoreTopologyService.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastCoreTopologyService.java
@@ -152,6 +152,7 @@ class HazelcastCoreTopologyService extends LifecycleAdapter implements CoreTopol
         ListenSocketAddress hazelcastAddress = config.get( discovery_listen_address );
         InterfacesConfig interfaces = new InterfacesConfig();
         interfaces.addInterface( hazelcastAddress.getHostname() );
+        interfaces.setEnabled( true );
 
         NetworkConfig networkConfig = new NetworkConfig();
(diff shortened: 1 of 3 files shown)


martinfurmanski:
@mneedham I think you should test with something other than the localhost.

PR #9087, opened 2017-03-23
Cache the PropertyExistanceEnforcer

thobe:
Only create new instances of PropertyExistanceEnforcer when needed after the schema has changed.

In addition to that we now only apply the PropertyExistanceEnforcer if there are data changes.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
@@ -25,12 +25,13 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 
 import org.neo4j.helpers.collection.Iterators;
 import org.neo4j.kernel.api.schema_new.LabelSchemaDescriptor;
 import org.neo4j.kernel.api.schema_new.SchemaDescriptor;
 import org.neo4j.kernel.api.schema_new.SchemaDescriptorPredicates;
-import org.neo4j.kernel.api.schema_new.constaints.ConstraintBoundary;
 import org.neo4j.kernel.api.schema_new.constaints.ConstraintDescriptor;
 import org.neo4j.kernel.api.schema_new.index.NewIndexDescriptor;
 import org.neo4j.kernel.impl.constraints.ConstraintSemantics;
@@ -55,6 +56,8 @@ public class SchemaCache
 
     private final Map<SchemaDescriptor, NewIndexDescriptor> indexDescriptors = new HashMap<>();
     private final ConstraintSemantics constraintSemantics;
+
+    private final Map<Class<?>,Object> dependantState = new ConcurrentHashMap<>();
 
     public SchemaCache( ConstraintSemantics constraintSemantics, Iterable<SchemaRule> initialRules )
     {
@@ -128,8 +131,14 @@ public class SchemaCache
         return Iterators.filter( SchemaDescriptor.equalTo( descriptor ), constraints.iterator() );
     }
 
+    public <P, T> T getOrCreateDependantState( Class<T> type, Function<P,T> factory, P parameter )
+    {
+        return type.cast( dependantState.computeIfAbsent( type, key -> factory.apply( parameter ) ) );
+    }
+
     public void addSchemaRule( SchemaRule rule )
     {
+        dependantState.clear();
         if ( rule instanceof ConstraintRule )
         {
             ConstraintRule constraintRule = (ConstraintRule) rule;
@@ -163,6 +172,7 @@ public class SchemaCache
 
     public void removeSchemaRule( long id )
     {
+        dependantState.clear();
         if ( constraintRuleById.containsKey( id ) )
         {
             ConstraintRule rule = constraintRuleById.remove( id );
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
@@ -322,7 +322,7 @@ public class RecordStorageEngine implements StorageEngine, Lifecycle
                     txStateVisitor, storeLayer, storageStatement, txState, countsRecordState );
             try ( TxStateVisitor visitor = txStateVisitor )
             {
-                txState.accept( txStateVisitor );
+                txState.accept( visitor );
             }
 
             // Convert record state into commands
--- a/enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
+++ b/enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
@@ -50,6 +50,7 @@ import org.neo4j.storageengine.api.txstate.TxStateVisitor;
 
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
+import static org.neo4j.collection.primitive.PrimitiveSortedArraySet.mergeSortedSet;
 import static org.neo4j.kernel.api.exceptions.schema.ConstraintValidationException.Phase.VALIDATION;
 
 class PropertyExistenceEnforcer
@@ -84,7 +85,7 @@ class PropertyExistenceEnforcer
         int[] current = map.get( key );
         if ( current != null )
         {
-            values = merge( current, values );
+            values = mergeSortedSet( current, values );
         }
         map.put( key, values );
     }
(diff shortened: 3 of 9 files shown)


thobe:
I believe it is safe to not do any specific locking in `SchemaCache`, since - as far as I know - changes made there are made by transactions holding an _exclusive schema lock_ - these are the cases where we clear the `dependantState`, and reads of the `dependantState` are made by transactions holding a _shared schema lock_.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
```diff
@@ -322,7 +322,7 @@ public void createCommands(
                     txStateVisitor, storeLayer, storageStatement, txState, countsRecordState );
             try ( TxStateVisitor visitor = txStateVisitor )
             {
-                txState.accept( txStateVisitor );
```
fickludd, line 325: This is just for clarity right, the behaviour is the same?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
```diff
@@ -42,191 +49,406 @@
...
+            return storageStatement == null ? storageStatement = storeLayer.newStatement() : storageStatement;
+        }
     }
 
-    private void validateNodeProperty( long nodeId, int propertyKey, LabelSchemaDescriptor descriptor )
-            throws NodePropertyExistenceException
+    private final List<LabelSchemaDescriptor> nodeConstraints;
+    private final List<RelationTypeSchemaDescriptor> relationshipConstraints;
+    private final PrimitiveIntObjectMap<int[]> mandatoryNodePropertiesByLabel = Primitive.intObjectMap();
+    private final PrimitiveIntObjectMap<int[]> mandaRelationshipPropertiesByType = Primitive.intObjectMap();
```
fickludd: mandaRelationshipPropertiesByType -> mandatoryRelationshipPropertiesByType? mandatoryRelPropertiesByType?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
```diff
@@ -42,191 +49,406 @@
...
-                PrimitiveIntSet labelIds = node.get().labels();
+                storageStatement.close();
+            }
+        }
 
-                propertyKeyIds.clear();
-                try ( Cursor<PropertyItem> properties = properties( node.get() ) )
+        private void validateNode( long nodeId ) throws NodePropertyExistenceException
+        {
+            if ( mandatoryNodePropertiesByLabel.isEmpty() )
```
fickludd: From a readability perspective, I would prefer to have the actual `PropertyExistenceEnforcer` before the inner `Decorator` class, because then the source structure follows the execution order.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
```diff
@@ -42,191 +49,406 @@
...
         }
     }
 
-    private Cursor<NodeItem> nodeCursor( long id )
+    private void validateNodeProperties( long id, PrimitiveIntSet labelIds, PrimitiveIntSet propertyKeyIds )
+            throws NodePropertyExistenceException
     {
-        Cursor<NodeItem> cursor = storeStatement().acquireSingleNodeCursor( id );
-        return txState.augmentSingleNodeCursor( cursor, id );
+        if ( labelIds.size() > mandatoryNodePropertiesByLabel.size() )
```
fickludd: pulling join tricks at a micro level? awesome.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
```diff
@@ -42,191 +49,406 @@
...
-    private StorageStatement storeStatement()
+    private void failNode( long id, int label, int propertyKey )
+            throws NodePropertyExistenceException
     {
-        return storageStatement == null ? storageStatement = storeLayer.newStatement() : storageStatement;
+        for ( LabelSchemaDescriptor constraint : nodeConstraints )
+        {
+            if ( constraint.getLabelId() == label && contains( constraint.getPropertyIds(), propertyKey ) )
+            {
+                throw new NodePropertyExistenceException( constraint, VALIDATION, id );
```
fickludd: Interestingly, because we now have composite existence constraints in form of Node Keys apart from the regular sort, we could actually be violating several constraints here. Unsure whether it's worth the bother to report that though...

enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcerMergeTest.java
```diff
@@ -0,0 +1,148 @@
...
+                lhs( 1, 2, 3 ).rhs( 1, 2, 3 ).expectLhs(),
+                lhs( 1, 2, 3 ).rhs( 1 ).expectLhs(),
+                lhs( 1, 2, 3 ).rhs( 2 ).expectLhs(),
+                lhs( 1, 2, 3 ).rhs( 3 ).expectLhs(),
+                lhs( 1 ).rhs( 1, 2, 3 ).expectRhs(),
+                lhs( 2 ).rhs( 1, 2, 3 ).expectRhs(),
+                lhs( 3 ).rhs( 1, 2, 3 ).expectRhs(),
+                lhs( 1, 2, 3 ).rhs( 4, 5, 6 ).expect( 1, 2, 3, 4, 5, 6 ),
+                lhs( 1, 3, 5 ).rhs( 2, 4, 6 ).expect( 1, 2, 3, 4, 5, 6 ),
+                lhs( 1, 2, 3, 5 ).rhs( 2, 4, 6 ).expect( 1, 2, 3, 4, 5, 6 )
```
fickludd: Nice tests.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
```diff
@@ -42,191 +49,406 @@
...
+    /**
+     * Merges two sets of integers represented as sorted arrays.
+     *
+     * @param lhs
+     *         a set of integers, represented as a sorted array.
+     * @param rhs
+     *         a set of integers, represented as a sorted array.
+     * @return a set of integers, represented as a sorted array.
+     */
+    static int[] merge( int[] lhs, int[] rhs )
```
fickludd: Since this method has nothing to do with property existence per se, how about lifting this method to some util class? also `contains`. New class `org.neo4j.collection.primitive.PrimitiveSortedArrays`?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
```diff
@@ -128,8 +131,14 @@ public boolean hasIndexRule( SchemaDescriptor descriptor )
         return Iterators.filter( SchemaDescriptor.equalTo( descriptor ), constraints.iterator() );
     }
 
+    public <P, T> T getOrCreateDependantState( Class<T> type, Function<P,T> factory, P parameter )
+    {
+        return type.cast( dependantState.computeIfAbsent( type, key -> factory.apply( parameter ) ) );
+    }
```
fickludd, line 137: This is a good trick.

fickludd (commented):
I think the functional behaviour is very nice, but have some minor comments on the implementation.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
thobe: yes.

Also, if we in the future add something to the line above that makes `visitor` not be the same as `txStateVisitor`, this ensures that we keep doing the right thing.

It also looks better to not have an un-used variable in the try-with-resource block.

I first tried making changes around here, but then reverted those, but decided this change was good enough to keep.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
thobe: i.e. to move the definition of the `Decorator` class to the bottom of the `PropertyExistenceEnforcer` class (below all the methods)?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
thobe: Yes, I thought about that too, and thought that we could implement it in a way that we report _all_ violations. That seemed like a different change though, so I didn't do it at this stage.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
thobe: Great idea.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
thobe: "trick"? What is it that is a trick?

fickludd (approved):

PR #9100, opened 2017-03-27
3.2 Unused indexes regression

fickludd:
This PR addresses the regression in `CreateDeleteNodePropertiesWithUnusedIndexes`, which is essentially a benchmark that sets and removes properties in the presence of a large amount of indexes that should not be updated. 

The implementation is made more efficient by introducing a new map in `SchemaCache` for looking up indexes by propertyId, and two new maps in `IndexMap` for index lookup by propertyId and labelId. In addition the `IndexingService` - `OnlineIndexUpdates` interplay has been modified to make one less index lookup per node.

Some new utility methods have also appeared for set operations on primitive long arrays, which was needed for label manipulations. 

This PR complete removes the slowdown on `CreateDeleteNodePropertiesWithUnusedIndexes` of 9-12x that was observed with 10000 unused indexes. Speed with PR is 1.00-1.02x. Performance regressions in `CreateUniqueNodeProperties.setProperty` are also reduced substantially, although these regressions were way less dramatic.


community/kernel/src/main/java/org/neo4j/helpers/SortedLongArrayUtil.java
```diff
@@ -0,0 +1,251 @@
...
+{
+    private static final long[] EMPTY = new long[]{};
+
+    /**
+     * Compute union of two sorted long arrays.
+     * @param left a sorted array set
+     * @param right another sorted array set
+     * @return the union, which is NOT sorted
+     */
+    public static long[] union( long[] left, long[] right )
```
thobe: This looks very similar to what I recently introduced in [`PrimitiveSortedArraySet`](https://github.com/neo4j/neo4j/blob/3.2/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveSortedArraySet.java#L35), how about you add this method there instead? Possibly implementing the same algorithm as the method that is already there (but with `long[]` in the signature instead of `int[]`) so that the method is composable.
This implementation not returning a sorted set means that it has to be sorted externally in order to use it as input in a later iteration, and I don't think that returning sorted output adds to the complexity of the implementation. In fact if you compare the two implementations, my sorted version has _lower_ time complexity (although slightly higher memory consumption in the worst case).

At the very least, I think this class belongs better in the `neo4j-primitives` component, since afaik the `helpers` package is being phased out.

community/kernel/src/main/java/org/neo4j/helpers/SortedLongArrayUtil.java
```diff
@@ -0,0 +1,251 @@
...
+        return union;
+    }
+
+    /**
+     * Compute intersect of two sorted long array sets.
+     * @param left a sorted array set
+     * @param right another sorted array set
+     * @return the union, which is NOT sorted
+     */
+    public static long[] intersect( long[] left, long[] right )
```
thobe: Again, I don't see the point of not returning a sorted array - it would make it more useable if it took two "sets" as input and returned a "set" as output.

This method should of course be moved with its buddy above...

community/kernel/src/main/java/org/neo4j/helpers/SortedLongArrayUtil.java
```diff
@@ -0,0 +1,251 @@
...
+        return intersect;
+    }
+
+    /**
+     * Compute the symmetric difference (set XOR basically) of two sorted long array sets.
+     * @param left a sorted array set
+     * @param right another sorted array set
+     * @return the union, which is NOT sorted
+     */
+    public static long[] symmetricDifference( long[] left, long[] right )
```
thobe: Same deal as above...

community/kernel/src/main/java/org/neo4j/helpers/SortedLongArrayUtil.java
```diff
@@ -0,0 +1,251 @@
...
+        return difference;
+    }
+
+    /**
+     * Compute the number of unique values in two sorted long array sets
+     * @param left a sorted array set
+     * @param right another sorted array set
+     * @return int pair packed into long
+     */
+    public static long countUnique( long[] left, long[] right )
```
thobe: This method is not used outside of the implementation of the methods above (and in tests), since it has a weird return type, I vote that it be made a private implementation detail rather than part of the API (or package-private in order to allow testing).

If the above methods no longer depend on this method when they return sorted output instead (which might become the case), I vote that we remove this method.

community/kernel/src/main/java/org/neo4j/helpers/SortedLongArrayUtil.java
```diff
@@ -0,0 +1,251 @@
...
+            }
+        }
+        uniqueInLeft += left.length - l;
+        uniqueInRight += right.length - r;
+        return intPair( uniqueInLeft, uniqueInRight );
+    }
+
+    public static long intPair( int left, int right )
+    {
+        return (long)left << (Integer.BYTES * 8) | right;
```
thobe: I'd really like some parenthesis here to ensure that:
1. The conversion from `int` to `long` happens before the shift.
2. The shift happens before the or.

Also, this method is implementation detail and should not be public.

community/kernel/src/main/java/org/neo4j/helpers/SortedLongArrayUtil.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    public static int left( long pair )
+    {
+        return (int)(pair >> (Integer.BYTES * 8));
```
thobe: Same deal as above.

community/kernel/src/main/java/org/neo4j/helpers/SortedLongArrayUtil.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    public static int right( long pair )
+    {
+        return (int)pair;
```
thobe: This method is wrong since the conversion from `long` to int is sign-preserving.

A correct implementation would be `return (int)(pair & 0xFFFF_FFFFL);`.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexMap.java
```diff
@@ -37,17 +44,34 @@
...
+    private final PrimitiveIntObjectMap<Set<LabelSchemaDescriptor>> descriptorsByProperty;
 
     public IndexMap()
     {
         this( new HashMap<>(), new HashMap<>(), new HashMap<>() );
     }
 
-    private IndexMap( Map<Long, IndexProxy> indexesById, Map<LabelSchemaDescriptor,IndexProxy> indexesByDescriptor, Map<LabelSchemaDescriptor,Long> indexIdsByDescriptor )
+    public IndexMap(
+            Map<Long, IndexProxy> indexesById )
```
thobe: this could fit on a single line...

thobe (commented):
An initial round of comments after giving this a quick look-over. I'd like to analyse these changes a bit deeper.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -460,26 +459,62 @@ else if ( state == State.RUNNING || state == State.STARTING )
         }
     }
 
-    private void apply( IndexUpdates updates, IndexUpdateMode updateMode )
+    private void apply( Iterable<IndexEntryUpdate<LabelSchemaDescriptor>> updates, IndexUpdateMode updateMode )
```
MishaDemianenko, line 463: WDYT about apply that look like that:
```
    private void apply( Iterable<IndexEntryUpdate<LabelSchemaDescriptor>> updates, IndexUpdateMode updateMode )
            throws IOException, IndexEntryConflictException
    {
        Map<LabelSchemaDescriptor, IndexUpdater> indexUpdaters = new HashMap<>();
        for ( IndexEntryUpdate<LabelSchemaDescriptor> indexUpdate : updates )
        {
            IndexUpdater indexUpdater =
                    indexUpdaters.computeIfAbsent( indexUpdate.indexKey(), getMappingFunction( updateMode ) );
            indexUpdater.process( indexUpdate );
        }
        IOUtils.closeAll( indexUpdaters.values() );
    }

    private Function<LabelSchemaDescriptor,IndexUpdater> getMappingFunction( IndexUpdateMode updateMode )
    {
        return labelSchemaDescriptor ->
        {
            try
            {
                return indexMapRef.getIndexProxy( labelSchemaDescriptor ).newUpdater( updateMode );
            }
            catch ( IndexNotFoundKernelException infe )
            {
                throw new RuntimeException( infe );
            }
        };
    }
```
that saves a bit of cycles on iteration over updates

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexMap.java
```diff
@@ -95,32 +130,204 @@ public void foreachIndexProxy( BiConsumer<Long, IndexProxy> consumer )
...
+     *
+     * @param unchangedLabels set of labels that are unchanged
+     * @param properties set of properties that have changed
+     * @return set of LabelSchemaDescriptors describing the potentially affected indexes
+     */
+    private Set<LabelSchemaDescriptor> getDescriptorsByProperties(
+            long[] unchangedLabels,
+            PrimitiveIntCollection properties )
+    {
+        int nIndexesForLabels = countIndexesByLabels( unchangedLabels );
```
MishaDemianenko, line 264: while we calculate counts over labels and properties we already know all the descriptors and we can save them for later use since we maybe will need to extract them; to avoid double work

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -460,26 +459,62 @@ else if ( state == State.RUNNING || state == State.STARTING )
...
         }
     }
 
+    public void convertToIndexUpdatesAndApply( Iterable<NodeUpdates> updates, IndexUpdateMode updateMode )
+            throws IOException, IndexEntryConflictException
+    {
+        final List<IndexEntryUpdate<LabelSchemaDescriptor>> indexUpdates = new ArrayList<>();
+        for ( NodeUpdates nodeUpdate : updates )
+        {
+            for ( IndexEntryUpdate<LabelSchemaDescriptor> indexUpdate : convertToIndexUpdates( nodeUpdate ) )
```
MishaDemianenko: we can create lazy iterator here instead of collecting updates into array list and pass that iterator into apply, so we can avoid one top additional for loop

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/OnlineIndexUpdates.java
```diff
@@ -113,8 +115,10 @@ private void gatherUpdatesFor( long nodeId, NodeCommand nodeCommand, List<Proper
                 gatherUpdatesFromCommandsForNode( nodeId, nodeCommand, propertyCommands );
 
         NodeUpdates nodeUpdates = nodePropertyUpdate.build();
-        updateService.loadAdditionalProperties( nodeUpdates );
-        updates.add( nodeUpdates );
+        for ( IndexEntryUpdate<LabelSchemaDescriptor> update :  updateService.convertToIndexUpdates( nodeUpdates ) )
```
MishaDemianenko: here we fetch updates from iterable and put it into array list and then we will take iterator out of that updates to iterate elsewhere throw them, shouldn't we avoid intermediate extraction step here then?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexMap.java
```diff
@@ -95,32 +130,204 @@ public void foreachIndexProxy( BiConsumer<Long, IndexProxy> consumer )
...
     }
 
+    /**
+     * Get all indexes that would be affected by changes in the input labels and/or properties
+     * @param changedLabels set of labels that have changed
+     * @param unchangedLabels set of labels that are unchanged
+     * @param properties set of properties
+     * @return set of LabelSchemaDescriptors describing the potentially affected indexes
+     */
+    public Set<LabelSchemaDescriptor> getRelatedIndexes(
```
MishaDemianenko, line 143: I think we need additional javadoc describing how we will evaluate affected indexes

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexMap.java
```diff
@@ -95,32 +130,204 @@ public void foreachIndexProxy( BiConsumer<Long, IndexProxy> consumer )
...
+        Map<LabelSchemaDescriptor, Long> map = new HashMap<>();
+        for ( Map.Entry<Long,IndexProxy> entry : indexesById.entrySet() )
+        {
+            map.put( entry.getValue().schema(), entry.getKey() );
+        }
+        return map;
+    }
+
+    /**
+     * Get descriptors affected by changed properties. Implementation is smart about whether doing
```
MishaDemianenko: would be nice to have a description what `smart` actually mean and how those lookups will be done

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/IndexTxStateUpdater.java
```diff
@@ -34,24 +34,25 @@
...
 
 public class IndexTxStateUpdater
 {
-    private final SchemaReadOperations schemaReadOps;
+    private final StoreReadLayer storeReadLayer;
     private final EntityReadOperations readOps;
     private final NodeSchemaMatcher nodeIndexMatcher;
 
-    public IndexTxStateUpdater( SchemaReadOperations schemaReadOps, EntityReadOperations readOps )
+    // We can use the StoreReadLayer directly instead of the SchemaReadOps, because we know that in transactions
```
MishaDemianenko, line 51: How do we can guarantee this? shouldn't we add some assertion to double check this?

community/kernel/src/main/java/org/neo4j/storageengine/api/StoreReadLayer.java
```diff
@@ -78,6 +78,11 @@
     Iterator<NewIndexDescriptor> uniquenessIndexesGetAll();
 
     /**
+     * Returns all indexes (including unique) related to a property.
+     */
+    Iterator<NewIndexDescriptor> indexesAndUniqueIndexesRelatedToProperty( int propertyId );
```
MishaDemianenko, line 83: just to much existing naming schema can we rename it to `indexesGetForProperty` or `indexesGetAllForProperty`?

community/kernel/src/test/java/org/neo4j/test/assertion/Assert.java
```diff
@@ -41,6 +41,23 @@ private Assert()
     {
     }
 
+    public static <E extends Exception> void assertException( ThrowingAction<E> f, Class typeOfException ) throws
```
MishaDemianenko, line 44: isn't it copy paste mostly of another assertException method? why not reused that?

MishaDemianenko, line 44: Also if this method is introduced why not to replace all usages that have dummy expected message passed like in IndexRuleTest: `assertException( indexRule::getOwningConstraint, IllegalStateException.class, "" );` ?

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveSortedArraySet.java
```diff
@@ -21,26 +21,43 @@
 
 import java.util.Arrays;
 
+/**
+ * Specialized methods for operations on sets represented as sorted primitive arrays.
+ *
+ * This class does not contain a complete set of operations for all primitives, but only
+ * the ones that were needed. Feel free to add specializations on demand, but remember to test.
+ */
 public class PrimitiveSortedArraySet
```
MishaDemianenko: as we talked yesterday, maybe we can find a better name for this handy helper class

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveSortedArraySet.java
```diff
@@ -21,26 +21,43 @@
...
 {
+    private static final long[] NO_LONGS = new long[]{};
```
MishaDemianenko: PrimitiveLongCollections.EMPTY_LONG_ARRAY

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveSortedArraySet.java
```diff
@@ -21,26 +21,43 @@
...
+    private static final int INT_BITS = Integer.BYTES * 8;
```
MishaDemianenko: Integer.SIZE ?

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveSortedArraySet.java
```diff
@@ -21,26 +21,43 @@
 
 import java.util.Arrays;
 
+/**
+ * Specialized methods for operations on sets represented as sorted primitive arrays.
+ *
+ * This class does not contain a complete set of operations for all primitives, but only
```
MishaDemianenko: this part of javadoc do not look like really usefull, imho

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveSortedArraySet.java
MishaDemianenko: would be nice to know if results returned from methods in this class also returns arrays that are actually sets

MishaDemianenko (changes requested):
Couple of comments, not sure about smart way of loading affected indexes would be nice to have some more javadocs in that area

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/OnlineIndexUpdates.java
fickludd: In this case we need to materialize here, as the when the `iterator()` is called, the store might have changed so the ´convertToIndexUpdates´ might not yield the same result.

community/kernel/src/main/java/org/neo4j/storageengine/api/StoreReadLayer.java
fickludd: This will be done by the PR that is merge last of this and #9119. The reason this method was named like this is that for similar methods in the interface unique and general indexes are separated. That is changed in #9119.

community/kernel/src/test/java/org/neo4j/test/assertion/Assert.java
fickludd: Yes, but that method is annoying if the exception message is `null`. If you do not care about the message, why should a called have to decide whether to but "" or null as the message argument.

Fixed the implementation to reuse the previous one though.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/IndexTxStateUpdater.java
fickludd: Good point. Added assertions in all method calls.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexMap.java
fickludd: Yeah, but then we need to first collect the subparts into a list for both the label and property paths, before turning the smaller one into a set and return. This means some extra construction work. 

And if we do not want to construct objects when there are no affected indexes, we need ´null´-checks which makes stuff harder to read.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexMap.java
fickludd: I improved the javadoc to explain the guarantees on the method, but since the implementation is quite short the details can be left to the code.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
fickludd: Yeah, I acted on an old TODO here, but it didn't turn out very well. Reverted to old behavior using `IndexUpdaterMap`, which achieves roughly what you propose here.

community/kernel/src/test/java/org/neo4j/test/assertion/Assert.java
fickludd: Aye, did that.

fickludd:
@MishaDemianenko @thobe Addresed both of your comments. I think the utility class is getting really nice, thanks for catching all the details in there.

Also am done with additional benchmarking, and it all looks good.

MishaDemianenko:
@fickludd looks like test failures are related

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveArrays.java
```diff
@@ -0,0 +1,297 @@
...
+
+/**
+ * Specialized methods for operations on primitive arrays.
+ *
+ * For set operations (union, intersect, symmetricDifference), input and output arrays
+ * are arrays containing unique values in sorted ascending order.
+ */
+public class PrimitiveArrays
+{
+    private static final long[] EMPTY_LONG_ARRAY = new long[]{};
```
MishaDemianenko: we already have empty long array defined in PrimitiveLongCollections.EMPTY_LONG_ARRAY can you reuse that instead?

MishaDemianenko (changes requested):
Have only one small minor thing, also as mentioned lest failures looks related; otherwise no more comments! looks good thx!

MishaDemianenko (approved):

PR #9103, opened 2017-03-28
Allow skip bad entries logging during import

MishaDemianenko:
Introduce new import tool option: `skip-bad-entries-logging` that will allow to skip logging of all bad entries.
Default value of new option: false

### DIFF
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -42,6 +42,7 @@ import org.neo4j.helpers.Strings;
 import org.neo4j.helpers.collection.IterableWrapper;
 import org.neo4j.helpers.collection.Iterables;
 import org.neo4j.helpers.collection.MapUtil;
+import org.neo4j.io.IOUtils;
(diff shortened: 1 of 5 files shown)


community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -177,25 +178,30 @@
                 "Number of bad entries before the import is considered failed. This tolerance threshold is "
                         + "about relationships refering to missing nodes. Format errors in input data are "
                         + "still treated as errors" ),
+        SKIP_BAD_ENTRIES_LOGGING ("skip-bad-entries-logging", Boolean.FALSE,
+                "<true/false>", "Whether or not log bad entries detected during import."),
```
tinwelint: Whether or not `to skip` log`ging` bad entries detected during import

tinwelint (approved):
Great that you also deferred the message string creation, which could be expensive.

tinwelint (changes requested):
Oh, the little change in the option description

MishaDemianenko:
@tinwelint updated :)

tinwelint (approved):

PR #9106, opened 2017-03-28
Update MERGE locking

systay:
Grabs merge locks to prevent other users to concurrently create the same nodes.

Before this change, this was only done when a uniqueness constraint was defined on the label/property combination. Now, these locks are independent of constraints.

changelog: Changes how locking is done when using MERGE, making sure we have concurrency protection even when no unique constraints are defined.


systay:
Will re-open when done

PR #9107, opened 2017-03-28
`neo4j-import`: Added ability to configure import input data buffer size

tinwelint:
cl `neo4j-import`: Ability to configure import input data buffer size using `--read-buffer-size`

### DIFF
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -210,7 +210,12 @@ public class ImportTool
                 "Page size in bytes, or e.g. 4M or 8k" ),
         LEGACY_STYLE_QUOTING( "legacy-style-quoting", Configuration.DEFAULT_LEGACY_STYLE_QUOTING,
                 "<true/false>",
-                "Whether or not backslash-escaped quote e.g. \\\" is interpreted as inner quote." );
+                "Whether or not backslash-escaped quote e.g. \\\" is interpreted as inner quote." ),
+        READ_BUFFER_SIZE( "read-buffer-size", org.neo4j.csv.reader.Configuration.DEFAULT.bufferSize(),
+                "<bytes, e.g. 10k, 4M>",
+                "Size of the buffer for reading input data. It has to be large enough to hold at least the " +
+                "biggest single value in the data. Actual memory used may be bigger than specified, " +
+                "given some amount of internal overhead" );
 
         private final String key;
         private final Object defaultValue;
@@ -723,6 +728,9 @@ public class ImportTool
         final Boolean multiLineFields = args.getBoolean( Options.MULTILINE_FIELDS.key(), null );
         final Boolean emptyStringsAsNull = args.getBoolean( Options.IGNORE_EMPTY_STRINGS.key(), null );
         final Boolean legacyStyleQuoting = args.getBoolean( Options.LEGACY_STYLE_QUOTING.key(), null );
+        final Number bufferSize = args.has( Options.READ_BUFFER_SIZE.key() )
+                ? parseLongWithUnit( args.get( Options.READ_BUFFER_SIZE.key(), null ) )
+                : null;
         return new Configuration.Default()
         {
             @Override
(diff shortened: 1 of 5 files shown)


community/import-tool/src/test/java/org/neo4j/tooling/ImportToolTest.java
```diff
@@ -1627,6 +1629,41 @@ public void shouldDisableLegacyStyleQuotingIfToldTo() throws Exception
...
+        catch ( IllegalStateException e )
+        {
+            // THEN good
+            assertThat( e.getMessage(), containsString( "input data" ) );
+        }
+    }
+
+    private static String longName( int characters )
+    {
+        StringBuilder builder = new StringBuilder();
```
MishaDemianenko: `return StringUtils.repeat( 'l', characters )`

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -210,7 +210,12 @@
                 "Page size in bytes, or e.g. 4M or 8k" ),
         LEGACY_STYLE_QUOTING( "legacy-style-quoting", Configuration.DEFAULT_LEGACY_STYLE_QUOTING,
                 "<true/false>",
-                "Whether or not backslash-escaped quote e.g. \\\" is interpreted as inner quote." );
+                "Whether or not backslash-escaped quote e.g. \\\" is interpreted as inner quote." ),
+        READ_BUFFER_SIZE( "read-buffer-size", org.neo4j.csv.reader.Configuration.DEFAULT.bufferSize(),
+                "<bytes, e.g. 10k, 4M>",
+                "Size of the buffer for reading input data. It has to be large enough to hold at least the " +
+                "biggest single value in the data. Actual memory used may be bigger than specified, " +
```
MishaDemianenko: Do you really think we should include `Actual memory used may be bigger than specified, given some amount of internal overhead` in option description here? isn't it the case for almost any kind of structures? Also it does not provide any hints or highlights how end user should react on that fact: should buffer be always 2k bigger or 5k bigger then expected bigger value?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/TicketedProcessing.java
```diff
@@ -153,15 +153,24 @@ public void submit( FROM job )
     {
         return future( () ->
         {
-            while ( input.hasNext() )
+            try
```
MishaDemianenko, line 156: Is this change just a nice find or it has something to do with data buffer size problem?

MishaDemianenko (changes requested):
Some comments

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/TicketedProcessing.java
tinwelint: This was an existing problem that was exposed by doing the other change, so had to be fixed in order for this PR to go green

community/import-tool/src/test/java/org/neo4j/tooling/ImportToolTest.java
tinwelint: Ah... that

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
tinwelint: I think it's there because I know that the overhead is 2x or perhaps even more with parallel parsing, so it's a non-neglectible overhead. Dunno how to communicate that best.

MishaDemianenko: hm....what if we will say something like that:
`Size of the shared buffer for reading input data. Biggest single value should be able to fit into buffer during reading.` Do you think it still provide enough details for end user to be able to understand what it actually does?

tinwelint: Yeah, I think that'd be enough

MishaDemianenko:
ping @tinwelint

tinwelint:
@MishaDemianenko rephrased the description a bit. It got to be something in between what we discussed, e.g. the buffer isn't exactly shared since there may be multiple when parsing stuff in parallel... things like that. I think it's good enough now.

All comments addressed.

MishaDemianenko (approved):

PR #9110, opened 2017-03-29
Resolves a deadlock scenario applying constraint

tinwelint:
The scenario, which takes place on database instance applying constraint
creation as an external transaction, looks like this:

- Transaction T1 creates the constraint index and population P starts
- Transaction T2 which activates the constraint starts applying
  and now has a read lock on the counts store
- Check point triggers, wants to rotate counts store and so acquires
  its write lock. It will have to block, but doing so will also blocks
  further read lock requests
- T2 moves on to activate the constraint. Doing so means first waiting for
  the index to come online
- P moves on to flip after population, something which includes initializing
  some sample data in counts store for this index. Will block on the counts
  store read lock, completing the deadlock

The solution in this commit breaks the deadlock by having T2 unlock its
read lock on the counts store before activating the constraint, i.e.
as soon as it notices that it is a schema transaction. Schema transactions
will not update the counts store anyway.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/CountsStoreTransactionApplier.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/CountsStoreTransactionApplier.java
@@ -30,6 +30,7 @@ public class CountsStoreTransactionApplier extends TransactionApplier.Adapter
 {
     private final TransactionApplicationMode mode;
     private final CountsTracker.Updater countsUpdater;
+    private boolean haveUpdates;
 
     public CountsStoreTransactionApplier( TransactionApplicationMode mode, CountsAccessor.Updater countsUpdater )
     {
@@ -56,6 +57,7 @@ public class CountsStoreTransactionApplier extends TransactionApplier.Adapter
     public boolean visitNodeCountsCommand( Command.NodeCountsCommand command )
     {
         assert countsUpdater != null || mode == TransactionApplicationMode.RECOVERY : "You must call begin first";
+        haveUpdates = true;
         if ( countsUpdater != null )
         {   // CountsUpdater is null if we're in recovery and the counts store already has had this transaction applied.
             countsUpdater.incrementNodeCount( command.labelId(), command.delta() );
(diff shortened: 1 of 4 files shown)


community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointerConstraintCreationDeadlockIT.java
```diff
@@ -0,0 +1,200 @@
...
+
+    @Rule
+    public final EphemeralFileSystemRule fs = new EphemeralFileSystemRule();
+    @Rule
+    public final OtherThreadRule<Void> t2 = new OtherThreadRule<>( "T2" );
+    @Rule
+    public final OtherThreadRule<Void> t3 = new OtherThreadRule<>( "T3" );
+
+    @Test
+    public void shouldNotDeadlock() throws Exception
```
MishaDemianenko: we need to add a timeout to this test just in case it will hang

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointerConstraintCreationDeadlockIT.java
```diff
@@ -0,0 +1,200 @@
...
+    {
```
MishaDemianenko: also we need to add a bit of assertions to the test, now we do not have any

community/kernel/src/main/java/org/neo4j/kernel/impl/api/CountsStoreTransactionApplier.java
```diff
@@ -68,4 +74,15 @@ public boolean visitRelationshipCountsCommand( Command.RelationshipCountsCommand
         }
         return false;
     }
+
+    @Override
+    public boolean visitSchemaRuleCommand( SchemaRuleCommand command ) throws IOException
```
MishaDemianenko, line 82: also as we just talk, would be nice to have guarding assertion that will fail in case when this assumption will brake

MishaDemianenko (commented):
couple of comments, otherwise we can give it a try

community/kernel/src/main/java/org/neo4j/kernel/impl/api/CountsStoreTransactionApplier.java
tinwelint: Yup, this has been added

MishaDemianenko:
ping @tinwelint

tinwelint:
@MishaDemianenko updated

MishaDemianenko (approved):

PR #9111, opened 2017-03-29
Never checkpoint GBPTree on close

burqen:
Checkpoint will increment stable and unstable generation of tree,
making it impossible to recognise crash-pointers during the recovery
following the database crash.

Responsibility for checkpointing is completely left to the system
so GBPTree will only checkpoint when rest of system is checkpointing.

A checkpoint is always done as last step of a clean shutdown, which means NativeLabelScanStore (with underlying GBPTree) will always be flushed.

First attempt at this prevented checkpoint on close only when system
saw panic, but that did not guard for some failures during recovery
that did not trigger panic.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
@@ -735,8 +735,8 @@ public class GBPTree<KEY,VALUE> implements Closeable
 
     /**
      * Closes this tree and its associated resources.
-     * NOTE: No {@link #checkpoint(IOLimiter) checkpoint} is performed. To make data persistent in store
-     * a checkpoint needs to be performed manually.
+     * <p>
+     * NOTE: No {@link #checkpoint(IOLimiter) checkpoint} is performed.
      *
      * @throws IOException on error closing resources.
      */
@@ -751,10 +751,10 @@ public class GBPTree<KEY,VALUE> implements Closeable
                 return;
             }
 
-            closed = true;
             // Force close on writer to not risk deadlock on pagedFile.close()
             writer.close();
             pagedFile.close();
+            closed = true;
         }
         finally
         {
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
@@ -290,7 +290,7 @@ public class GBPTree<KEY,VALUE> implements Closeable
 
     /**
      * Whether or not this tree has been closed. Accessed and changed solely in
-     * {@link #close()} and {@link #close(Consumer)} to be able to close tree multiple times gracefully.
+     * {@link #close()} to be able to close tree multiple times gracefully.
      */
     private boolean closed;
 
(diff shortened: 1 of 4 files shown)


community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -734,31 +734,15 @@ private void checkpoint( IOLimiter ioLimiter, Header.Writer headerWriter ) throw
     }
 
     /**
-     * Closes this tree and its associated resources. A {@link #checkpoint(IOLimiter)} is first performed
-     * as part of this call if there have been changes since last call to {@link #checkpoint(IOLimiter)}
-     * or since opening this tree.
+     * Closes this tree and its associated resources.
+     * NOTE: No {@link #checkpoint(IOLimiter) checkpoint} is performed. To make data persistent in store
+     * a checkpoint needs to be performed manually.
```
MishaDemianenko: i think `manually` is not quite a match here what if we will say that checkpoint is performed during system checkpoint instead?
Also please separate NOTE: part using javadoc formatting

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -767,19 +751,10 @@ private void close( Header.Writer headerWriter ) throws IOException
...
-                // Perform a checkpoint before closing. If no changes has happened since last checkpoint,
-                // no new checkpoint will be created.
-                checkpoint( IOLimiter.unlimited(), headerWriter );
-            }
-            finally
-            {
-                closed = true;
-                pagedFile.close();
-            }
+            closed = true;
```
MishaDemianenko: why do we marking index as closed before actually closing resource?

community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
```diff
@@ -28,10 +28,12 @@
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.OpenOption;
```
MishaDemianenko, line 31: just noticed that index tests create tons of test managed threads, why do we need to do that instead of using any sort of thread pool?
that will also allow you to use Callables and throw exceptions and not use throwing wrapper

MishaDemianenko (commented):
Couple of comments

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: That depends on where this tree is used. As a general statement it has to be done manually. In the context of being part of a neo4j database it's done by the integration points inside neo4j that manages these trees.

Comparing with documentation about recovery it's documented how interaction with the rest of the world is supposed to look like, not that "the system takes care of recovery". I'd like the javadocs in here to focus on how others should interact with it.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: Doesn't matter since `closed` is only read/changed inside this method, guarded by a lock.

MishaDemianenko: isn't that unexpected thing to have a flag that say that index is closed to be raised before it's actually is?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
MishaDemianenko: Why do you need to mention at all that `To make data persistent in store a checkpoint needs to be performed manually.` if javadoc already say that close does not perform it that sentence does not bring any additional value.

tinwelint: Sure, I believe we're splitting hairs here. I find it useful to have some amount of duplicate information in javadoc when referring to other methods to not require reader to jump around too much. I won't comment on this further.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: It makes no difference, is my point. I'll leave this up to @burqen

community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
tinwelint: That might be something to improve on at some point, agree

burqen: Fixed in #9117

MishaDemianenko (approved):

PR #9114, opened 2017-03-29
Handle LocalDatabase restart breaking bolt connections

mneedham:
This happens because TransactionIdStore doesn't get refreshed after a store copy.

* Introduce CANCELLED state so that polling process doesn't get restarted unless we explicitly want to restart it
* Pass around a supplier of TransactionIdStore and resolve it in TransactionIdTracker
* Bumped drivers to 1.2

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachineSPI.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachineSPI.java
@@ -22,6 +22,7 @@ package org.neo4j.bolt.v1.runtime;
 import java.time.Clock;
 import java.time.Duration;
 import java.util.Map;
+import java.util.function.Supplier;
 
 import org.neo4j.bolt.v1.runtime.TransactionStateMachine.BoltResultHandle;
 import org.neo4j.bolt.v1.runtime.spi.BoltResult;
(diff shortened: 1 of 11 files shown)


community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachineSPI.java
```diff
@@ -64,7 +65,7 @@
     TransactionStateMachineSPI( GraphDatabaseAPI db,
                                 ThreadToStatementContextBridge txBridge,
                                 QueryExecutionEngine queryExecutionEngine,
-                                TransactionIdStore transactionIdStoreSupplier,
+                                Supplier<TransactionIdStore> transactionIdStoreSupplier,
```
lutovich: Could we maybe remove this param and take `transactionIdStoreSupplier` out of `GraphDatabaseAPI db`?
This is not very important but will reduce amount of constructor arguments.

community/kernel/src/main/java/org/neo4j/kernel/api/txtracking/TransactionIdTracker.java
```diff
@@ -109,6 +110,6 @@ private boolean isReady( long oldestAcceptableTxId ) throws TransactionFailureEx
      */
     public long newestEncounteredTxId()
     {
-        return transactionIdStore.getLastClosedTransactionId();
+        return transactionIdStoreSupplier.get().getLastClosedTransactionId();
```
lutovich: Maybe extract `transactionIdStoreSupplier.get()` in a method with a comment saying why is it required.

lutovich (approved):
@mneedham changes look good to me. Made two minor comments. PR build failed with checkstyle violation: `TransactionIdTrackerTest.java:67: Line has trailing spaces`.

I'll look at failing driver tests now.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachineSPI.java
mneedham: Have changed this to get `TransactionIdStore` from `db.getDependencyResolver()...`

mneedham:
@lutovich I've merged your branch into this one and then squashed everything. checkstyle locally seems happy so let's see if the build server agrees

PR #9119, opened 2017-03-30
Merge index listing methods

fickludd:
This PR accomplished two things:

1) Merged index listing methods
In hot paths we need all indexes regardless of whether they're general or unique. It's therefor better to have one method return both types, and filter in the slow paths; than to concatenate in the hot paths as currently done.

2) Faster index lookup by label in schema cache
As some customers have a lot of indexes, we should be faster at this to not slow down label updates.

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/TransactionBoundPlanContext.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/TransactionBoundPlanContext.scala
@@ -43,7 +43,9 @@ class TransactionBoundPlanContext(tc: TransactionalContextWrapper, logger: Inter
   extends TransactionBoundTokenContext(tc.statement) with PlanContext with IndexDescriptorCompatibility {
 
   def indexesGetForLabel(labelId: Int): Iterator[IndexDescriptor] = {
-    tc.statement.readOperations().indexesGetForLabel(labelId).asScala.flatMap(getOnlineIndex)
+    tc.statement.readOperations().indexesGetForLabel(labelId).asScala
+      .filter(_.`type`() == KernelIndexDescriptor.Type.GENERAL)
+      .flatMap(getOnlineIndex)
   }
 
   def indexGet(labelName: String, propertyKeys: Seq[String]): Option[IndexDescriptor] = evalOrNone {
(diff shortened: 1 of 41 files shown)


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/TransactionBoundPlanContext.scala
```diff
@@ -54,23 +56,26 @@ class TransactionBoundPlanContext(tc: TransactionalContextWrapper, logger: Inter
...
 
   def uniqueIndexesGetForLabel(labelId: Int): Iterator[IndexDescriptor] = {
     // here we do not need to use getOnlineIndex method because uniqueness constraint creation is synchronous
-    tc.statement.readOperations().uniqueIndexesGetForLabel(labelId).asScala.map(kernelToCypher)
+    tc.statement.readOperations().indexesGetForLabel(labelId).asScala
+      .filter(_.`type`() == KernelIndexDescriptor.Type.UNIQUE)
+      .map(kernelToCypher)
   }
 
   def uniqueIndexGet(labelName: String, propertyKeys: Seq[String]): Option[IndexDescriptor] = evalOrNone {
```
ragadeeshu, line 73: Care to elaborate on how this differs from indexGet? Doesn't this need to filter any results?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
```diff
@@ -115,8 +115,7 @@
 import static org.neo4j.helpers.collection.Iterators.singleOrNull;
 import static org.neo4j.kernel.api.StatementConstants.NO_SUCH_NODE;
 import static org.neo4j.kernel.api.properties.DefinedProperty.NO_SUCH_PROPERTY;
-import static org.neo4j.kernel.api.schema_new.index.NewIndexDescriptor.Filter.GENERAL;
-import static org.neo4j.kernel.api.schema_new.index.NewIndexDescriptor.Filter.UNIQUE;
+import static org.neo4j.kernel.api.schema_new.index.NewIndexDescriptor.Filter.ANY;
```
ragadeeshu: All of the methods you now pass `ANY` to, do they still have calls where they are passed `UNIQUE`/`GENERAL`?

ragadeeshu (commented):
Some questions I would like answers to :)

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/TransactionBoundPlanContext.scala
fickludd: I didn't filter for `indexGet` or `uniqueIndexGet` because there can be only one index, and whether it is types as unique or not should not matter to Cypher. I think the plan is to remove the unique variants completely from the context and lets the index get versions return all indexes - just like this PR does to the Kernel interfaces.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
fickludd: Ooh, bonus. Nice find.

ragadeeshu:
@fickludd needs an update

PR #9123, opened 2017-03-30
Consistency check for incremental backup

MishaDemianenko:
Allow to perform consistency check of incremental backup for backup commands that use OnlineBackup API: check will be performed if user requested verification of created backup, by specifying `verification = true` in appropriate incremental backup call.

### DIFF
--- a/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackup.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackup.java
@@ -302,7 +302,7 @@ public class OnlineBackup
     public OnlineBackup incremental( String targetDirectory )
     {
         outcome = new BackupService().doIncrementalBackup( hostNameOrIp, port, new File( targetDirectory ),
-                timeoutMillis, defaultConfig() );
+                getConsistencyCheck( true ), timeoutMillis, defaultConfig() );
         return this;
     }
 
@@ -324,7 +324,7 @@ public class OnlineBackup
     public OnlineBackup incremental( String targetDirectory, boolean verification )
     {
         outcome = new BackupService().doIncrementalBackup( hostNameOrIp, port, new File( targetDirectory ),
-                timeoutMillis, defaultConfig() );
+                getConsistencyCheck( verification ), timeoutMillis, defaultConfig() );
         return this;
     }
 
--- a/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackup.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackup.java
@@ -302,7 +302,7 @@ public class OnlineBackup
     public OnlineBackup incremental( String targetDirectory )
     {
         outcome = new BackupService().doIncrementalBackup( hostNameOrIp, port, new File( targetDirectory ),
-                getConsistencyCheck( true ), timeoutMillis, defaultConfig() );
+                getConsistencyCheck( false ), timeoutMillis, defaultConfig() );
         return this;
     }
 
--- a/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
@@ -203,7 +203,7 @@ public class OnlineBackupCommand implements AdminCommand
             try
             {
                 backupService.doIncrementalBackup( address.getHost(), address.getPort(),
-                        destination, timeout, config );
+                        destination, ConsistencyCheck.NONE, timeout, config );
                 done = true;
             }
             catch ( Exception e )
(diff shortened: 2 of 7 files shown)


enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
```diff
@@ -203,7 +203,7 @@ public void execute( String[] args ) throws IncorrectUsage, CommandFailed
             try
             {
                 backupService.doIncrementalBackup( address.getHost(), address.getPort(),
-                        destination, timeout, config );
+                        destination, ConsistencyCheck.NONE, timeout, config );
```
tinwelint, line 206: Shouldn't we use the `check-consistency` argument for this one? Or rather introduce another one for incremental backup if we don't want to change behaviour? This is the main surface of doing a backup so it should be controllable from this surface IMHO.

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackup.java
```diff
@@ -302,7 +302,7 @@ public OnlineBackup full( String targetDirectory, boolean verification, Config t
     public OnlineBackup incremental( String targetDirectory )
     {
         outcome = new BackupService().doIncrementalBackup( hostNameOrIp, port, new File( targetDirectory ),
-                timeoutMillis, defaultConfig() );
+                getConsistencyCheck( true ), timeoutMillis, defaultConfig() );
```
tinwelint: This changes default behaviour, right? Should it?

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackup.java
```diff
@@ -324,7 +324,7 @@ public OnlineBackup incremental( String targetDirectory )
     public OnlineBackup incremental( String targetDirectory, boolean verification )
     {
         outcome = new BackupService().doIncrementalBackup( hostNameOrIp, port, new File( targetDirectory ),
-                timeoutMillis, defaultConfig() );
+                getConsistencyCheck( verification ), timeoutMillis, defaultConfig() );
```
tinwelint, line 327: So `verfication` wasn't quite used before? I'd guess it's fine to actually make use of it here, yeah.

tinwelint (commented):
Looks good and straight forward in general. Some questions above default behaviour changing as part of this PR.

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
MishaDemianenko: On line 238 you already have all of that right?

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackup.java
MishaDemianenko: Right i think it's fine to do that since it's not a patch fix release and by default it's kind of reasonable to have consistent backup, also if you really want to avoid check - if that taking to long you have a way to disable it, if you still using this deprecated api.

MishaDemianenko:
ping @tinwelint comments answered

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
spacecowboy: Yup, doing `ConsistencyCheck.NONE` here is absotely the correct thing to do.

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackup.java
spacecowboy: Field will be upset if you change the default behavior of the old `neo4j-backup`.

MishaDemianenko: @spacecowboy you think it's better to keep then without consistency checker?

MishaDemianenko: agree with @spacecowboy arguments, switch to not do consistency check by default

tinwelint:
@MishaDemianenko can you summarize here under which circumstances incremental backup will do consistency check? (because description is now out of date)

MishaDemianenko:
@tinwelint updated, hope it's more clear now 🌈

tinwelint:
And with that description you basically say that there's no way to trigger consistency check for incremental backup through our main surface (neo4j-admin)?

MishaDemianenko:
@tinwelint this PR does not change anything in relation to behaviour to neo4j-admin since it already have all the options to trigger consistency check for any kind of backups if requested, you can check that in `OnlineBackupCommand`, as it says in description now tis' changes only `OnlineBackup` API

tinwelint:
OK nice, gotcha. Thanks

PR #9124, opened 2017-03-30
Additional import tool error message test

MishaDemianenko:
@burqen Is my understanding is correct and now it's fine?


tinwelint:
@MishaDemianenko @burqen what to do with this PR?

PR #9133, opened 2017-04-03
Add possibility to register monitor for native label scan store

burqen:
And underlying GBPTree through Monitors.


chrisvest (approved):
Looks good! 👍

PR #9134, opened 2017-04-03
Add NODE KEY checks in consistency checker

fickludd:
As the node key constraint was implemented as a primary constraint we need to ensure that the consistency checker looks for node key constraints. Turns out this was already OK, but this PR adds tests.


fickludd:
The two failed tests seem to be flaky and unrelated.

chrisvest (approved):
Looks good! 👍

chrisvest:
@fickludd I've restarted the build. I've seen the `BoltCausalClusteringIT` fail other PRs, so if it fails again we can ignore it. The other one I haven't seen around so much.

PR #9136, opened 2017-04-03
testing active queries with cluster.

praveenag:


integrationtests/src/test/java/org/neo4j/causalclustering/scenarios/ListQueriesProcedureInClusterIT.java
```diff
@@ -0,0 +1,172 @@
...
+                ReadReplicaGraphDatabase replicaDb = cluster.findAnyReadReplica().database();
+                Node node = leaderDb.createNode();
+                tx.success();
+
+                threads.execute(parameter ->
+                {
+                    cluster.coreTx((leaderDb1, secondTransaction) -> {
+
+                        //lock node
+                        lock.accept(secondTransaction, node);
```
thobe: This seems like an unnecessary indirection, just call `secondTransaction.acquireWriteLock(node);` directly.

Then you don't need the `lock` `BiConsumer`.

integrationtests/src/test/java/org/neo4j/causalclustering/scenarios/ListQueriesProcedureInClusterIT.java
```diff
@@ -0,0 +1,172 @@
...
+        CountDownLatch listQueriesLatch = new CountDownLatch(1);
+
+        cluster.coreTx((CoreGraphDatabase leaderDb, Transaction tx) ->
+        {
+            try {
+
+                //Given
+                ReadReplicaGraphDatabase replicaDb = cluster.findAnyReadReplica().database();
+                Node node = leaderDb.createNode();
+                tx.success();
```
thobe: Since the transaction doesn't actually commit until `tx.close()` is invoked, the transaction has not committed at this point.

This means that the rest of this method relies on racy behaviour. It is quite possible that the execution of `CORE_QUERY` (and `REPLICA_QUERY`) will match no nodes at all, in which case the `CORE_QUERY` will not block.

The thread that locks the node will actually succeed, even though it shouldn't, this is for two reasons:
1. A transaction that creates a node does not lock it, since it should not be visible to any other transaction.
2. When locking a node, we don't actually check to see that it exists / is visible to that transaction.

I really don't think we should rely on such accidental behaviour.

integrationtests/src/test/java/org/neo4j/causalclustering/scenarios/ListQueriesProcedureInClusterIT.java
```diff
@@ -0,0 +1,172 @@
...
+    @Rule
+    public final ThreadingRule threads = new ThreadingRule();
+
+    @Before
+    public void setup() throws Exception {
+        cluster = clusterRule.startCluster();
+    }
+
+    @Test
+    public void shouldListQueriesOnCoreAndReplica() throws Exception {
```
thobe: I really would like it if this test was broken down into smaller method with descriptive names. There is too much code to make it easy to get an overview of what is actually being tested.

Breaking the method up will also make it cleaner to break the flow up in a simple sequence of transactions, avoiding the trouble of overlapping transactions that emerge from nesting the code as it is written now.

integrationtests/src/test/java/org/neo4j/causalclustering/scenarios/ListQueriesProcedureInClusterIT.java
```diff
@@ -0,0 +1,172 @@
...
+    private Optional<Map<String, Object>> getQueryListing(String query, GraphDatabaseFacade db) {
+        try (Result rows = db.execute("CALL dbms.listQueries()")) {
+            while (rows.hasNext()) {
+                Map<String, Object> row = rows.next();
+                if (query.equals(row.get("query"))) {
+                    return Optional.of(row);
+                }
+            }
+        }
+        return Optional.empty();
```
thobe: I wouldn't have written it like this, I would have simply let the method return `Map<String,Object>` instead of an `Optional` thereof, and then invoke `fail(...)` at the end if no matching query could be found. That way it would be easy to provide a good failure message (`NoSuchElementException("No value present")` - which is what `Optional.empty().get()` produces - is not a good failure message).

integrationtests/src/test/java/org/neo4j/causalclustering/scenarios/ListQueriesProcedureInClusterIT.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+
+    private static class Resource<T> implements AutoCloseable {
```
thobe: This class does not look like it is used anywhere.

thobe (changes requested):
The added test depends on racy timing to be "just right", this will lead to flaky tests.

praveenag:
@thobe I've made the requested changes how about now?

integrationtests/src/test/java/org/neo4j/causalclustering/scenarios/ListQueriesProcedureInClusterIT.java
```diff
@@ -0,0 +1,171 @@
...
+        CoreClusterMember leader = cluster.coreTx((CoreGraphDatabase leaderDb, Transaction tx) -> {
+            node[0] = leaderDb.createNode();
+            tx.success();
+            tx.close();
+        });
+
+        CoreGraphDatabase leaderDb = leader.database();
+        Result matchAllResult = leaderDb.execute("MATCH (n) RETURN n");
+        assertTrue(matchAllResult.hasNext());
+
```
thobe: Add `matchAllResult.close()` here to make sure that the transaction implied by the query gets closed properly.

thobe (approved):
Looks good enough.

integrationtests/src/test/java/org/neo4j/causalclustering/scenarios/ListQueriesProcedureInClusterIT.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.*;
+
+public class ListQueriesProcedureInClusterIT {
```
spacecowboy: The code style in this file is incorrect and it's failing the PR build.

spacecowboy (changes requested):

spacecowboy (approved):

PR #9138, opened 2017-04-04
Use last transaction log position for checkpoint that generated during upgrade

MishaDemianenko:
Use Last transaction log position for checkpoint and for last transaction metadata store fields.
LOG_POSITION field can provide logVersion that is bigger then any existing log
and that will generate checkpoint that have higher log version then subsequent checkpoint that will use
last committed transaction log version.

### DIFF
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigratorTest.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigratorTest.java
@@ -19,6 +19,7 @@
  */
 package org.neo4j.kernel.impl.storemigration.participant;
 
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigratorTest.java
```diff
@@ -67,12 +71,19 @@
     private final PageCacheRule pageCacheRule = new PageCacheRule();
     private final RandomRule random = new RandomRule();
     private final SchemaIndexProvider schemaIndexProvider = new InMemoryIndexProvider();
+    private PageCache pageCache;
```
burqen, line 74: Can we also update other tests to use this new global `PageCache`?

burqen (changes requested):
Small cleanup comment

MishaDemianenko:
Señor @burqen , done!

burqen (approved):

PR #9146, opened 2017-04-04
Fixes to legacy planner support for NODE KEY and query plan ascii-art

craigtaverner:
This PR fixes two bugs that involved overlapping work to the test infrastructure:

* The fact that legacy planners used uniqueness constraints for some error detection did not respect NODE KEY
* The fact that ascii-art rendering of query plans did not work from legacy planners

Most of the work is in improving testing infrastructure to better test compatibility differences between compilers.


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MergeNodeCompatibilityAcceptanceTest.scala
```diff
@@ -0,0 +1,221 @@
...
+class MergeNodeCompatibilityAcceptanceTest extends ExecutionEngineFunSuite with QueryStatisticsTestSupport
+  with NewPlannerTestSupport {
+
+  override protected def createGraphDatabase(config: Map[Setting[_], String] = databaseConfig()): GraphDatabaseCypherService = {
+    new GraphDatabaseCypherService(new TestEnterpriseGraphDatabaseFactory().newImpermanentDatabase(config.asJava))
+  }
+
+  Seq(UniquenessConstraintCreator, NodeKeyConstraintCreator).foreach { constraintCreator =>
+
+    test(s"$constraintCreator: should_match_on_merge_using_multiple_unique_indexes_if_only_found_single_node_for_both_indexes") {
```
pontusmelke, line 41: Why all the `_` in the test names?

craigtaverner: Copy-paste from 2.3 tests. There were no failing tests in the 3.2 code-base, and I had to go back far to find these. They could be edited to remove the spaces, although it could also be argued that keeping the `_` makes it easier to find the originals.

PR #9153, opened 2017-04-05
Aligns service availability log messages at startup

technige:
This also tweaks the `SocketAddress` representation to correctly display IPv6 addresses.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/helpers/SocketAddress.java
+++ b/community/kernel/src/main/java/org/neo4j/helpers/SocketAddress.java
@@ -58,7 +58,14 @@ public class SocketAddress
     @Override
     public String toString()
     {
-        return hostname + ":" + port;
+        if (hostname.contains(":"))
+        {
+            return "[" + hostname + "]:" + port;
+        }
+        else
+        {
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/helpers/SocketAddress.java
```diff
@@ -58,7 +58,14 @@ public InetSocketAddress socketAddress()
     @Override
     public String toString()
     {
-        return hostname + ":" + port;
+        if (hostname.contains(":"))
```
lutovich, line 61: Would this `#toString()` implementation return strange string when given host name already contains `[...]` like `[::1]`?
Maybe is it worth doing smth like:
```
InetAddress address = InetAddress.getByName( hostname );
if ( address instanceof Inet6Address )
{
    return "[" + address.getHostName() + "]:" + port;
}
else
{
    return address.getHostName() + ":" + port;
}
```

lutovich (commented):
@technige changes look good, one small question.

systay:
@technige Is this still valid? Should it be merged or closed?

systay:

PR #9155, opened 2017-04-05
Predicates on shortest path and optional match

pontusmelke:
When combining `OPTIONAL MATCH` and `shortestPath` with predicates we
forgot to keep track of the relationship name of the shortest path pattern,
causing the cost planner to choke on such queries.


pontusmelke:
whoops, should be 3.1

PR #9157, opened 2017-04-05
Only check uniqueness constraints on actual changes

thobe:
Skip checking if a property is re-assigned to the same value.
Skip checking if a label is added that already was assigned.

This avoids a deadlock scenario between concurrent MERGE queries, where the MERGE was followed by re-assignment of the same property. This is quite a common pattern, especially to bulk-re-assign all expected properties of the node.

Example query: `MERGE (user:User{id:$user.id}) SET user = $user` (where there is a uniqueness constraint on `(user:User) / user.id`

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
@@ -103,6 +103,10 @@ public class ConstraintEnforcingEntityOperations implements EntityOperations, Sc
             try ( Cursor<NodeItem> cursor = nodeCursorById( state, nodeId ) )
             {
                 NodeItem node = cursor.get();
+                if ( node.hasLabel( labelId ) )
+                {
+                    break; // no need to verify if the node already had the label
+                }
                 Object propertyValue = node.getProperty( propertyKeyId );
                 if ( propertyValue != null )
                 {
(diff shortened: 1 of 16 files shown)


thobe:
The code where the fix is implemented has changed greatly in `3.2`, so this needs quite a bit of care when forward-merging. @pontusmelke and me know what is needed, but it isn't hard to figure out.

community/kernel/src/test/java/org/neo4j/test/ConfigBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+    {
+        this.config = config;
+    }
+
+    public Map<Setting<?>,String> configuration()
+    {
+        return Collections.unmodifiableMap( config );
+    }
+
+    public ConfigBuilder and( Setting<?> key, String value )
```
burqen, line 89: May I suggest some javadoc here that points out that this will throw `NullPointerException` if called before `configure( ... )`.

community/kernel/src/test/java/org/neo4j/test/rule/concurrent/ThreadingRule.java
```diff
@@ -66,27 +70,76 @@ protected void after()
...
-        Predicates.await( threadTransfer, threadCondition, timeout, unit );
+        ConcurrentTransfer<Thread> transfer = new ConcurrentTransfer<>();
+        Future<TO> future = executor.submit( task( Barrier.NONE, function, function.toString(), parameter, transfer ) );
+        Predicates.await( transfer, threadCondition, timeout, unit );
         return future;
     }
 
     private static <FROM, TO, EX extends Exception> Callable<TO> task(
-            final Barrier barrier, final ThrowingFunction<FROM,TO,EX> function, final FROM parameter,
+            final Barrier barrier, final ThrowingFunction<FROM,TO,EX> function, final String name, final FROM parameter,
```
burqen: From what I can see, `barrier` is always `Barrier.NONE` and can be removed. I'll leave it up to you if you want to include that cleanup as part of this PR.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -120,27 +124,29 @@ public Property nodeSetProperty( KernelStatement state, long nodeId, DefinedProp
     {
         try ( Cursor<NodeItem> cursor = nodeCursorById( state, nodeId ) )
         {
-
             NodeItem node = cursor.get();
 
-            try ( Cursor<LabelItem> labels = node.labels() )
+            Object previousValue = node.getProperty( property.propertyKeyId() );
+            if (!property.valueEquals( previousValue ))
```
burqen: Code style

burqen (changes requested):
Changes look good. Just have some small comments.

community/kernel/src/test/java/org/neo4j/test/ConfigBuilder.java
thobe: Given that the `configure(...)` method is a static factory method, and the constructor is private, invoking this method will never lead to a `NullPointerException`.

burqen: Doh! Missed the fact that `and(...)` was _not_ static

burqen (approved):

PR #9162, opened 2017-04-06
Attempt to stabilize  BCCIT.shouldPickANewServerToWriteToOnLeaderSwitch

digitalstain:
It looks that the occasional failures of this test come from concurrent session
 creation and switching of leaders. This is not a problem in real deployments,
 but the fixed timeouts and specific requirements for leader election results
 along with the constant composition of the cluster violate some assumptions
 the test makes. This change ensures the leader switch that is needed completes
 before any further action is taken, and also makes sure that no additional
 elections are triggered during the session creations.


mneedham:
This looks good to me.

PR #9163, opened 2017-04-06
Bookmarks format extended.

jimwebber:
Bookmarks now allowed to be multiple like: tx99,tx56,tx101
Internal logic picks the highest of the supplied bookmarks.


lutovich:
Superseded by https://github.com/neo4j/neo4j/pull/9404

PR #9169, opened 2017-04-06
Better error message for existing index

pontusmelke:
If you first created a composite index and then tried to create a node
key constraint on the same properties we failed when trying to create the
proper error message. Instead we got the error message:

```
Single property schema requires one property but had 2
```
which was caused by calling `getProperty` when creating the error message.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/schema/SchemaKernelException.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/schema/SchemaKernelException.java
@@ -61,7 +61,7 @@ public abstract class SchemaKernelException extends KernelException
         if ( tokenNameLookup != null )
         {
             String propertyString = propertyIds.length == 1 ?
-                                    "property " + tokenNameLookup.propertyKeyGetName( descriptor.getPropertyId() ) :
+                                    "property '" + tokenNameLookup.propertyKeyGetName( propertyIds[0]) + "'" :
                                     "properties " + Arrays.stream( propertyIds )
                                             .mapToObj( (i) -> "'" + tokenNameLookup.propertyKeyGetName( i ) + "'")
                                             .collect( Collectors.joining( " and " ));
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/schema/SchemaKernelException.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/schema/SchemaKernelException.java
@@ -19,6 +19,9 @@
  */
 package org.neo4j.kernel.api.exceptions.schema;
 
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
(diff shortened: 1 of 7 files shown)


community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/schema/SchemaKernelException.java
```diff
@@ -50,20 +53,29 @@ public SchemaKernelException( Status statusCode, String message )
...
 
-    protected static String messageWithLabelAndPropertyName( TokenNameLookup tokenNameLookup, String formatString,
+    static String messageWithLabelAndPropertyName( TokenNameLookup tokenNameLookup, String formatString,
             LabelSchemaDescriptor descriptor )
     {
+        int[] propertyIds = descriptor.getPropertyIds();
+
         if ( tokenNameLookup != null )
         {
+            String propertyString = propertyIds.length == 1 ?
```
craigtaverner, line 63: I think this kind of logic should exist in one place. In an earlier version of this code we isolated the construction of a property description to the NodePropertyDescriptor which would internally know if it is single property or multi-property and format accordingly. Would that not be better here, so we don't repeat this logic in multiple places, and also don't have the conditional on the length of the property array?

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/schema/SchemaKernelException.java
```diff
@@ -50,20 +53,29 @@ public SchemaKernelException( Status statusCode, String message )
...
+                                            .collect( Collectors.joining( " and " ));
             return String.format( formatString,
-                    tokenNameLookup.labelGetName( descriptor.getLabelId() ),
-                    tokenNameLookup.propertyKeyGetName( descriptor.getPropertyId() ) );
+                    tokenNameLookup.labelGetName( descriptor.getLabelId() ), propertyString);
         }
         else
         {
+            String keyString = propertyIds.length == 1 ? "key[" + propertyIds[0] + "]" :
+                               "keys[" + Arrays.stream( propertyIds )
```
craigtaverner, line 74: We also seem to be solving this in multiple places. See for example https://github.com/neo4j/neo4j/blob/f8512e67e1533d675e2f045fdb027391ed4b731b/community/kernel/src/main/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureArgumentFormatter.java#L147

craigtaverner (approved):
Some suggestions that can be handled in later PRs

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/schema/SchemaKernelException.java
pontusmelke: Yes probably, for the error message to have the correct pluralization we still need to check the length of the array though.

PR #9172, opened 2017-04-07
Make query termination more deterministic

systay:


community/cypher/cypher/src/test/scala/org/neo4j/cypher/KillQueryTest.scala
```diff
@@ -0,0 +1,131 @@
...
+  This test creates 10 threads that run a Cypher query over and over again for 10 seconds.
+  Concurrently, another thread tries to terminate all running queries. This should not lead to weird behaviour - only
+  well known and expected exceptions should be produced.
+   */
+  val emptyMap = new util.HashMap[String, AnyRef]
+  val NODE_COUNT = 1000
+  val THREAD_COUNT = 10
+
+
+  test("apa") {
```
pontusmelke: a more descriptive name maybe :)

community/kernel/src/main/java/org/neo4j/kernel/impl/query/Neo4jTransactionalContext.java
```diff
@@ -56,6 +59,7 @@
     private InternalTransaction transaction;
     private Statement statement;
     private boolean isOpen = true;
+    private boolean terminated = false;
```
pontusmelke: AtomicBoolean?

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
```diff
@@ -95,6 +95,14 @@ public void streamResult( ThrowingConsumer<BoltResult, Exception> resultConsumer
         }
     }
 
+    /**
+     * Rollback and close transaction. Move back to {@link State#AUTO_COMMIT}.
+     * <p>
+     * <b>Warning:</b>This method should only be called by the bolt worker thread during it's regular message
+     * processing. It is wrong to call it from a different thread because kernel transactions are not thread-safe.
```
MishaDemianenko, line 102: Can we in some way force that to be a runtime check?

integrationtests/src/test/java/org/neo4j/bolt/SessionResetIT.java
```diff
@@ -0,0 +1,361 @@
...
+        }
+
+        assertNoActiveTransactions();
+    }
+
+    private Future<Void> runQueryInDifferentThreadAndResetSession( String query, boolean autoCommit ) throws Exception
+    {
+        AtomicReference<Session> sessionRef = new AtomicReference<>();
+
+        Future<Void> queryResult = runAsync( () ->
```
MishaDemianenko, line 240: Since this is CompletableFuture, we can avoid sleep and awaits here completely and just reset the session when query result future will be completed right?

MishaDemianenko (commented):
Couple of small comments otherwise looks fine!

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
lutovich: We can propagate bolt worker thread id from `RunnableBoltWorker` to `BoltStateMachine` and from there to this `TransactionStateMachine`. Then we can probably add runtime checks like this in all methods:

```java
if(Thread.currentThread() != boltWorkerThread)
    throw new IllegalStateException();
```

Do you think it is worth doing? Maybe as a separate PR?

MishaDemianenko: I think by doing that we will guard us against misuse since the comment is something that is not enforced; for sure that can be separate PR.

integrationtests/src/test/java/org/neo4j/bolt/SessionResetIT.java
lutovich: Future here executes query and we want to interrupt query execution with `#reset()` so can't wait for completion. Waiting for result happens only when asserting on success/failure. Subsequent `await` wants to see that query execution is in progress.

MishaDemianenko: ok

benbc:
@systay @MishaDemianenko Is there not a fix for this that could go into a 3.1 point release?

MishaDemianenko (approved):

MishaDemianenko:
@benbc it can go into 3.1 but in the same time neither me no @lutovich have seen those problems in 3.1 logs so we would like to avoid touching 3.1 for now

benbc:
@MishaDemianenko The original problem that I reported which spawned this work was in 3.1.

MishaDemianenko:
@benbc yes I know - benchmarks store; but since that is not really customer reported issue we would like to avoid touching 3.1 since we do not feel comfortable to introduce that change there so far.

PR #9173, opened 2017-04-07
Only use online indexes in planning

pontusmelke:
In planning we can only use indexes that are fully online. Failure
to do so will lead to planning errors saying that "indexes are still populating".

### DIFF
--- a/community/cypher/cypher/src/test/java/org/neo4j/cypher/CreateIndexStressIT.java
+++ b/community/cypher/cypher/src/test/java/org/neo4j/cypher/CreateIndexStressIT.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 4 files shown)


community/cypher/cypher/src/test/java/org/neo4j/cypher/CreateIndexStressIT.java
```diff
@@ -0,0 +1,96 @@
...
+    private final ExecutorService executorService = Executors.newFixedThreadPool( 10 );
+
+    @Test
+    public void shouldHandleConcurrentIndexCreationAndUsage() throws InterruptedException
+    {
+        // Given
+        HashMap<String,Object> params = new HashMap<>();
+        params.put( "param", NUM_PROPS );
+        db.execute( "FOREACH(x in range(0,$param) | CREATE (:A {prop:x})) ", params );
+        db.execute( "CREATE INDEX ON :A(prop) " );
```
craigtaverner, line 62: The fix is all about unique indexes, but the test is for non-unique indexes. Does this test really test the changes made?

craigtaverner (changes requested):

community/cypher/cypher/src/test/java/org/neo4j/cypher/CreateIndexStressIT.java
pontusmelke: yes when we do planning we first check if there is an index and then check if there is a unique index,  https://github.com/neo4j/neo4j/blob/3.2/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/planner/logical/steps/indexScanLeafPlanner.scala#L105. Since we now have changed the code not to use the `unique` variant but always use the normal index and filter lazily we need to explicitly check that the index is online before trying to use it.

To better understand this change it is better first look at https://github.com/neo4j/neo4j/pull/9119 where the index-handling operations in planning were merged but the old code was assuming that unique index construction is synchronous, this is no longer true since normal and unique now uses the same methods only with different filters.

You can also have a look at the cypher robustness tests to see that there are currently failures.

craigtaverner (approved):

PR #9182, opened 2017-04-10
Don't fail silently on native label scan store rebuild

burqen:
Before, if failed to complete rebuild of native label scan store db startup
would crash. But on next startup it would act as if nothing is wrong and
no new rebuild would be triggered, effectively ending up with a completely
inconsistent label scan store.

Now we mark native label scan store as dirty before starting rebuild from full
node scan and mark it as clean once it finishes. If native label scan store is
dirty during startup a complete rebuild will be triggered.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
@@ -19,6 +19,8 @@
  */
 package org.neo4j.kernel.impl.index.labelscan;
 
+import org.apache.commons.lang3.mutable.MutableBoolean;
+
 import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
@@ -81,6 +83,16 @@ public class NativeLabelScanStore implements LabelScanStore
      * Name of the file used for the native label scan store.
      */
     public static final String FILE_NAME = DEFAULT_NAME + ".labelscanstore.db";
+
+    /**
+     * Written in header to indicate native label scan store is clean
+     */
+    private static final byte CLEAN = (byte) 0x00;
+
+    /**
+     * Written in header to indicate native label scan store is dirty
+     */
+    private static final byte DIRTY = (byte) 0x01;
 
     /**
      * Whether or not this label scan store is read-only.
@@ -152,17 +164,12 @@ public class NativeLabelScanStore implements LabelScanStore
     /**
      * Write dirty bit to header
      */
-    private final Consumer<PageCursor> writeDirty;
+    private static final Consumer<PageCursor> writeDirty = pageCursor -> pageCursor.putByte( DIRTY );
 
     /**
      * Remove dirty bit from header
      */
-    private final Consumer<PageCursor> writeClean;
-
-    /**
-     * Read dirty bit from header
-     */
-    private final GetDirtyReader getDirty;
+    private static final Consumer<PageCursor> writeClean = pageCursor -> pageCursor.putByte( CLEAN );
 
     public NativeLabelScanStore( PageCache pageCache, File storeDir, FullStoreChangeStream fullStoreChangeStream,
             boolean readOnly, Monitors monitors )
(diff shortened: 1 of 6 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -290,14 +314,22 @@ public void init() throws IOException
...
+        }
+
+        if ( isDirty )
+        {
             monitor.notValidIndex();
             dropStrict();
             instantiateTree();
+        }
+        if ( isDirty || !storeExists )
+        {
```
chrisvest: The `|| !storeExists` part is covered by the `needsRebuild = !storeExists;` further up, is it not?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -342,6 +378,50 @@ public void recoveryCompleted( long numberOfPagesVisited, long numberOfCleanedCr
...
+        private boolean isDirty;
+        private boolean read;
+
+        @Override
+        public void read( PageCursor from, int length )
+        {
+            byte[] header = new byte[HEADER_LENGTH];
+            from.getBytes( header );
+            read = true;
+            isDirty = header[0] == DIRTY;
```
chrisvest: Why not just `from.getByte()` and skip the array construction and `HEADER_LENGTH`?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -342,6 +378,50 @@ public void recoveryCompleted( long numberOfPagesVisited, long numberOfCleanedCr
         };
     }
 
+    private static final int HEADER_LENGTH = 1;
+    private static final byte CLEAN = (byte) 0x00;
+    private static final byte DIRTY = (byte) 0x01;
```
chrisvest: `static` fields are usually kept together at the beginning of a class, no?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -321,9 +353,13 @@ private FileHandle storeFileHandle() throws IOException
...
 
-    private void instantiateTree() throws IOException
+    /**
+     * @return true instantiated tree needs to be rebuilt
+     */
+    private boolean instantiateTree() throws IOException
     {
-        index = new GBPTree<>( pageCache, storeFile, new LabelScanLayout(), pageSize, treeMonitor(), NO_HEADER );
+        index = new GBPTree<>( pageCache, storeFile, new LabelScanLayout(), pageSize, treeMonitor(), getDirty );
+        return getDirty.isDirty();
```
chrisvest: Does `getDirty` really need to be a field, of could you just create it here? I don't see you accessing it anywhere else.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -342,6 +378,50 @@ public void recoveryCompleted( long numberOfPagesVisited, long numberOfCleanedCr
...
+    }
+
+    private class CleanWriter implements Consumer<PageCursor>
+    {
+        @Override
+        public void accept( PageCursor pageCursor )
+        {
+            pageCursor.putByte( CLEAN );
+        }
+    }
```
chrisvest: These could also be lambdas, and their fields could be `static final`.

chrisvest (commented):
I had a few commits. Also, this should probably also be reviewed by someone with more familiarity with the code.

burqen:
@chrisvest Comments addressed 😄

burqen:
@tinwelint Fresh rebase.

tinwelint:
This will be merged as part of merging https://github.com/neo4j/neo4j/pull/9211

burqen:
Superseded by #9211

Issue #9130, opened 2017-04-01
"Client triggered an unexpected error" when loading empty CSV file


also:
Running `LOAD CSV WITH HEADERS` with an empty CSV file causes an unexpected, unrecoverable error.

**Neo4j Version:** 3.1.3  
**Operating System:** macOS 10.12.3
**API:** Cypher

### Steps to reproduce

Run this in `cypher-shell`:

```
// assert correct line count
LOAD CSV WITH HEADERS FROM "file:///path/to/empty.csv" AS line
RETURN count(*);
```

`empty.csv` is an empty (0 byte) file.

### Expected behavior
An error like "Invalid CSV" is returned. The error is recoverable in the client session, and the server does not log an exception.

### Actual behavior

The cypher-shell session shows this error, and is no longer usable:

```
next on empty iterator
Exception in thread "Thread-0" org.neo4j.driver.v1.exceptions.ClientException: Cannot run more statements in the current session as an unrecoverable error has happened. Please close the current session and re-run your statement in a new session.
        at org.neo4j.driver.internal.NetworkSession.ensureNoUnrecoverableError(NetworkSession.java:274)
        at org.neo4j.driver.internal.NetworkSession.reset(NetworkSession.java:132)
        at org.neo4j.shell.state.BoltStateHandler.reset(BoltStateHandler.java:169)
        at org.neo4j.shell.CypherShell.reset(CypherShell.java:183)
        at org.neo4j.shell.CypherShell$1.run(CypherShell.java:190)
```

This is logged in the server's `debug.log`:
```
2017-04-01 18:28:17.784+0000 ERROR [o.n.b.v.r.ErrorReporter] Client triggered an unexpected error [UnknownError]: next on empty iterator. See debug.log for more details, reference a937cd46-55dd-44c6-b91a-ab0745c8cb90.
2017-04-01 18:28:17.784+0000 ERROR [o.n.b.v.r.ErrorReporter] Client triggered an unexpected error [UnknownError]: next on empty iterator, reference a937cd46-55dd-44c6-b91a-ab0745c8cb90. next on empty iterator
java.util.NoSuchElementException: next on empty iterator
	at scala.collection.Iterator$$anon$2.next(Iterator.scala:39)
	at scala.collection.Iterator$$anon$2.next(Iterator.scala:37)
	at org.neo4j.cypher.internal.compiler.v3_1.spi.CSVResources$$anon$1.next(CSVResources.scala:101)
[...]

PR #9183, opened 2017-04-10
LOAD CSV WITH HEADERS from empty file

pontusmelke:
When calling `LOAD CSV WITH HEADERS FROM $url` where `$url` is an empty
file we should not fail with a
`java.util.NoSuchElementException: next on empty iterator` but handle
it gracefully.

Note that the failing docs build is expected and is true for all 2.3 builds

changelog: Fixed an unnecessary error when loading an empty file using LOAD CSV WITH HEADERS


pontusmelke:
fixes #9130

PR #9184, opened 2017-04-10
Do not leak cursors in the interpreted runtime

davidegrohmann:
This commit does _not_ fix the leaking in the compiled runtime


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/pipes/CachingExpandInto.scala
```diff
@@ -84,26 +92,35 @@ trait CachingExpandInto {
...
-      override def fetchNextOrNull(): Relationship = {
-        while (relationships.hasNext) {
-          val rel = relationships.next()
-          val other = rel.getOtherNode(start)
-          if (end == other) {
-            connectedRelationships.append(rel)
-            return rel
+    new Iterator[Relationship] with AutoCloseable
+    {
+      private val inner = new PrefetchingIterator[Relationship] {
```
pontusmelke, line 100: I know this is unrelated to your change, but should we depend on `PrefetchingIterator` in the cypher modules?

davidegrohmann: I don't know. Perhaps not? :)

thobe: Anther note on the actual implementation here, why delegate to an inner `PrefetchingIterator`? `PrefetchingIterator` is an `Iterator`, so could this not simply be `new PrefetchingIterator[Relationship] with AutoCloseable`. That would mean you would only need to implement `fetchNextOrNull()` and `close()`, both with the same implementation as they have in the code here, the delegating `hasNext()` and `next()` implementations would not be needed.

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/pipes/ExpandAllPipe.scala
```diff
@@ -34,16 +34,18 @@ case class ExpandAllPipe(source: Pipe,
                         (implicit pipeMonitor: PipeMonitor)
   extends PipeWithSource(source, pipeMonitor) {
 
+  private val relationships: ThreadLocal[Iterator[Relationship] with AutoCloseable] =
+    new ThreadLocal[Iterator[Relationship] with AutoCloseable]
```
thobe, line 38: This out-of-bounds state storage irks me. Could we not just close the iterator when we are done with it? i.e. make the caller responsible for closing the iterator it receives from `internalCreateResults` (and have that method return `Iterator[ExecutioContext] with AutoCloseable`).

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/pipes/ExpandIntoPipe.scala
```diff
@@ -45,21 +45,32 @@ case class ExpandIntoPipe(source: Pipe,
   self =>
   private final val CACHE_SIZE = 100000
 
+  private val _relationships : ThreadLocal[Iterator[Relationship] with AutoCloseable] =
+    new ThreadLocal[Iterator[Relationship] with AutoCloseable]
```
thobe, line 49: More `ThreadLocal` = more yuck. Again, it seems to me like having `internalCreateResults` return `Iterator[ExecutionContext] with AutoCloseable` is a better approach.

thobe (changes requested):
I didn't get very far before concluding that I disagree with the general pattern.

Storing resources that need to be managed in `ThreadLocal` seems like a recipe for future issues.

I'd much rather prefer if we explicitly manage the resources that need managing. In this case it would seem that the pipes should not be made stateful, but rather the resources they produce should be made to delegate the management of the resources they consume. I.e. the pipes should return `Iterator[ExecutionContext] extends AutoCloseable`, where the `close()` method delegates to the resources it depends on, rather than storing the resources in a `ThreadLocal`.

davidegrohmann:
@thobe There is a huge problem with the approach you are suggesting, that is to make it to work you basically need to reimplement all the scala collections api by hand, e.g., `map`, `flatmap`, `filter`, etc; indeed they cannot return anything else but `Iterator`s.

davidegrohmann:
@thobe Morever you also need to consider that you might have many nested iterators composing the `Iterator[ExecutionContext]`, which means you need to close them while processing and not at the end of the main stream.

thobe:
@davidegrohmann not being able to rely on scala collection methods is probably true, that's a bit of an annoyance, but if it leads to a better solution I think it might be worth the investment.

Multiple nested iterators being used to compose a single higher level iterator is pretty natural, and we would have to explicitly deal with the resources and close those objects when we are done with them in the higher level iterator. Tying the lifecycle management to the iterators does not mean that all resource management must be done by a `close()` method.

davidegrohmann:
@thobe Not being able to rely on scala collection methods means that we need to change/rewrite all the pipes we have. Which is something I am not really willing to do.

thobe:
Well, you actually _did_ rewrite a substantial portion of the pipes we have already...

To find out how "bad" it would be to write our own iterator type with our own transformation methods, I took a stab at it. It turns out it is not so bad. In most pipes the change ends up being only a change in method signatures, replacing `Iterator` with our new type, that is trivial to do. In a few other cases the change is a bit more involved, but never big enough to change the structure of the code, and frequently the resulting code ends up being _simpler_ because we can convert to our iterator type directly from the underlying representation types that we use rather than having to convert to other scala collection types in between.

In terms of operators that we use on `scala.collection.Iterator` we would need to implement these:
* `flatMap`
* `map`
* `filter`
* `++` (combining two iterators - this is used by `UnionPipe`)
* `drop` (used by `SKIP`)
* `collect` (used by `VarLengthExpand`)

What I also found was that not only did this make the resource management clear, straightforward, and explicit, but it also simplified the code in other ways. In fact, I found that completely decoupling our iterator type from the scala collection types made things even simple, since that allowed having more specific parameter types on the methods mentioned above.

As a final remark, I'll note that all the changes I made while outlining this change were confined to the `cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2` package, which means that it doesn't cause problems with Cypher version compatibilities.

To be fair, it did take me three hours to make the textual changes required to outline these changes, so it isn't no work at all, but I really do think it will lead to cleaner code and fewer hazards than the `ThreadLocal` approach.

My outline of the changes to using resource managing iterators throughout the Pipes: https://github.com/thobe/neo4j/commit/d171db106fbd43f04d5d615ee8e1b54dc0df285e

thobe:
One thing that struck me while investigating this is that even if we do patch this hazard in the 3.2 runtime, the compatibility versions of old Cypher releases will still have the same problem.

This further drives me towards thinking that the line between what is part of the Cypher version and what is part of the Neo4j version should be moved. I believe the entire runtime should be the same for all versions of Cypher within one version of Neo4j. The Cypher-version should do all steps up to and including planning, and that is where we should draw the SPI-line instead and have Neo4j be able to create runtime representations for logical plans from multiple versions of Cypher. I also think that would simplify a lot of the indirection that have been introduced when trying to decouple the Cypher runtime from the kernel version.

PR #9185, opened 2017-04-10
Invalid handling of composite index

pontusmelke:
There were several instances where we assumed a single property when creating error messages and exceptions. There was also a more severe error where multiple attempts to create a composite index, e.g.
```
CREATE INDEX ON :Person(firstname, lastname)
CREATE INDEX ON :Person(firstname, lastname)
```
ended up in an error instead with a non-sensical error message instead of being idempotent.

### DIFF
--- a/community/kernel/src/test/java/org/neo4j/kernel/api/exceptions/index/IndexPopulationFailedKernelExceptionTest.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/api/exceptions/index/IndexPopulationFailedKernelExceptionTest.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


community/kernel/src/test/java/org/neo4j/kernel/api/exceptions/index/IndexPopulationFailedKernelExceptionTest.java
```diff
@@ -0,0 +1,74 @@
...
+        when( lookup.propertyKeyGetName( 43 ) ).thenReturn( "BAR" );
+        when( lookup.propertyKeyGetName( 44 ) ).thenReturn( "BAZ" );
+
+        // When
+        IndexPopulationFailedKernelException index =
+                new IndexPopulationFailedKernelException( descriptor, "INDEX", "an act of pure evil occurred");
+
+        // Then
+        assertThat(index.getUserMessage( lookup ), equalTo(
+                "Failed to populate index for INDEX [labelId: 0, properties [42, 43, 44]], due to an act of pure evil occurred"));
```
fickludd, line 72: How about `properties` -> `propertyIds:`? Note the `:`

fickludd (approved):
Looks good. There's some small comment, fix or don't, then lets merge this guy.

craigtaverner (approved):

PR #9186, opened 2017-04-10
Cleanup cycling and configuration of page swappers

MishaDemianenko:
Build page cache factory only page cache is actually created.
Move configure to PageSwapper and remove ConfigurablePageSwapperFactory.


MishaDemianenko:
@chrisvest can you please take a look? WDYT?

community/io/src/main/java/org/neo4j/io/pagecache/PageSwapperFactory.java
```diff
@@ -33,17 +34,27 @@
...
+
+    /**
+     * Configure swapper with filesystem to use.
+     * @param fs file system to use in page swappers
      */
-    void setFileSystemAbstraction( FileSystemAbstraction fs );
+    default void configure( FileSystemAbstraction fs )
+    {
+        this.configure( fs, Configuration.EMPTY );
+    }
```
chrisvest: I think we can get away with only having one `configure` method; the one that takes both parameters.
I'd also like to call it `configureAndInitialise`, to highlight that it has important life-cycle semantics. Or alternatively, we can call it `open`, to be symmetric with the `close` method further down.

chrisvest (commented):
Though it is annoying to add a dependency on `neo4j-graphdb-api`, it does actually look like the best solution, since breaking up graphdb-api would be even more uncomfortable.

Just one comment about the `configure` method.

MishaDemianenko:
@chrisvest updated; i have even try to brake that dependency since it was strange in the beginning to have it, but braking and extracting configuration from kernel is quite thing to do :)

chrisvest (approved):

PR #9187, opened 2017-04-10
Taking Flight to the Land of Off-Heap, Act I: Page Meta-Data

chrisvest:
* The `MuninnPage` objects are gone.
* In their place is a new `PageList`.
* The `PageList` allocates and initialises a list of pages up, with IDs 0 to N.
* To work with a page, it must first be dereferenced, turning a _page ID_ (`int`) into a _page reference_ (`long`).
* All on-heap page data is immutable; all mutable page meta-data state is now off-heap.
* This reduces the memory usage for page meta-data from 64 bytes (including object headers) to 32 bytes.
* `MuninnPagedFile` now extends `PageList` via a copy-constructor, removing the `pageCache` indirection between it, and the set of memory pages.
* The `SequenceLock` is gone, replaced by an `OffHeapSequenceLock` which works on pointers to 8 byte lock-words.
* The new sequence lock also takes over the responsibility of managing the "dirty bit", or modified flag, as it is now called. This removes a memory write in maintaining page dirtiness.
* We can no longer store _references_ to page swappers in the pages, so we now indirect with an ID into a `SwapperSet`.
* The `SwapperSet` is now in charge of the life cycle of the page swappers, and their IDs, which can be reused.
* Of note, a swapper ID cannot be reused until all pages that uses it, have been evicted. This will only happen after the swapper has been closed.
* _Currently there is no vacuuming of swapper IDs, this needs to be added._
* The `MuninnPagedFile` translation tables are now `int[][]` arrays instead of `Object[][]`, to refer to _page IDs_.
* Since the translation table can now no longer contain references to `BinaryLatch` instances during page faulting the given entry, a new `LatchMap` is introduced to hold and manage latches for page faulting.
* The memory used for the page meta-data in the `PageList` is currently _NOT_ accounted against the memory allocated to the page cache. It is worth considering, though, as I think it simplifies capacity planning wrt. to memory consumption.


community/unsafe/src/main/java/org/neo4j/unsafe/impl/internal/dragons/MemoryManager.java
```diff
@@ -62,6 +66,22 @@ public MemoryManager( long expectedMaxMemory, long alignment )
      */
     public synchronized long allocateAligned( long bytes )
     {
+        if ( bytes > GRAB_SIZE )
+        {
+            // This is a huge allocation. Put it in its own slab and keep any existing slab at the head.
+            Slab nextSlab = slabs == null? null : slabs.next;
+            Slab allocationSlab = new Slab( nextSlab, bytes, alignment );
+            if ( !allocationSlab.canAllocate( bytes ) )
```
ragadeeshu: I think that this needs a comment. The first call works if stuff are perfectly aligned? And if the first call doesn't work you need `[1,alignment]` bytes more, and thus you add `alingment` to be safe?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
```diff
@@ -0,0 +1,354 @@
...
+ * <p>
+ * The OffHeapPageLock supports non-blocking optimistic concurrent read locks, non-blocking concurrent write locks,
+ * and non-blocking pessimistic flush and exclusive locks.
+ * <p>
+ * The optimistic read lock works through validation, so at the end of the critical section, the read lock has to be
+ * validated and, if the validation fails, the critical section has to be retried. The read-lock acquires a stamp
+ * at the start of the critical section, which is then validated at the end of the critical section. The stamp is
+ * invalidated if any write lock or exclusive lock was overlapping with the read lock.
+ * <p>
+ * The concurrent write locks works by assuming that writes are always non-conflicting, so no validation is required.
```
ragadeeshu, line 36: Why can we assume this?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
```diff
@@ -0,0 +1,354 @@
...
+ * The exclusive lock will also invalidate the optimistic read locks. The exclusive lock is try-lock only, and will
+ * never block. If a write or flush lock is currently held, the attempt to take the exclusive lock will fail, and
+ * the exclusive lock will likewise prevent write and flush locks from being taken.
+ * <p>
+ * Because all lock types are non-blocking, and because the lock-word itself is external to the implementation, this
+ * class does not need to maintain any state by itself. Thus, the class cannot be instantiated, and all methods are
+ * static.
+ * <p>
+ * Note that the lock-word is assumed to be 8 bytes, and should ideally be aligned to 8-byte boundaries, and ideally
+ * run on platforms that support 8-byte-wide atomic memory operations.
```
ragadeeshu, line 59: Any examples on platforms that don't support `8-byte-wide atomic memory operations`?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
```diff
@@ -0,0 +1,354 @@
...
+     * Validate a stamp from {@link #tryOptimisticReadLock(long)} or {@link #unlockExclusive(long)}, and return {@code true}
+     * if no write or exclusive lock overlapped with the critical section of the optimistic read lock represented by
+     * the stamp.
+     *
+     * @param stamp The stamp of the optimistic read lock.
+     * @return {@code true} if the optimistic read lock was valid, {@code false} otherwise.
+     */
+    public static boolean validateReadLock( long address, long stamp )
+    {
+        UnsafeUtil.loadFence();
```
ragadeeshu, line 153: Should this not be a storeFence, in order to make sure that the an invalidating store does not get reordered to happen after the folloing load?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
```diff
@@ -0,0 +1,354 @@
...
+     * invalidate any optimistic read lock that overlaps with them, and write locks will make any attempt at grabbing
+     * an exclusive lock fail. If an exclusive lock is currently held, then the attempt to take a write lock will fail.
+     * <p>
+     * Write locks must be paired with a corresponding {@link #unlockWrite(long)}.
+     *
+     * @return {@code true} if the write lock was taken, {@code false} otherwise.
+     */
+    public static boolean tryWriteLock( long address )
+    {
+        long s, n;
```
ragadeeshu: Just because the code resembles C doesn't mean that you have to adopt horrendous naming standards ;)

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
```diff
@@ -0,0 +1,354 @@
...
+            // bitwise-OR to reduce branching and allow more ILP
+            if ( unwritablyLocked | writeCountOverflow )
+            {
+                return failWriteLock( s, writeCountOverflow );
+            }
+
+            n = s + CNT_UNIT | MOD_MASK;
+            if ( compareAndSetState( address, s, n ) )
+            {
+                UnsafeUtil.storeFence();
```
ragadeeshu, line 195: And I suppose that this is in order to make sure that everyone sees that we took the lock, yes?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
```diff
@@ -0,0 +1,354 @@
...
+        {
+            s = getState( address );
+            if ( (s & CNT_MASK) == 0 )
+            {
+                throwUnmatchedUnlockWrite( s );
+            }
+            n = nextSeq( s ) - CNT_UNIT;
+        }
+        while ( !compareAndSetState( address, s, n ) );
+    }
```
ragadeeshu, line 233: No fence since it will be visible sooner or later, and there is no harm in an unlocked page looking like it is locked?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
```diff
@@ -0,0 +1,332 @@
...
+    {
+        long address = baseAddress - 8;
+        for ( long i = 0; i < pageCount; i++ )
+        {
+            UnsafeUtil.putLong( address += 8, 0 ); // lock word
+            UnsafeUtil.putLong( address += 8, 0 ); // pointer
+            UnsafeUtil.putLong( address += 8, PageCursor.UNBOUND_PAGE_ID ); // file page id
+            UnsafeUtil.putLong( address += 8, 0 ); // rest
+        }
+        UnsafeUtil.fullFence(); // Guarantee the visibility of the cleared memory
```
ragadeeshu, line 135: Would not a `storeFence` be enough here? Since the method only does store operations?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
```diff
@@ -0,0 +1,332 @@
...
+     * @param pageId The {@code pageId} to turn into a {@code pageRef}.
+     * @return A {@code pageRef} which is an opaque, internal and direct pointer to the meta-data of the given memory
+     * page.
+     */
+    public long deref( long pageId )
+    {
+        return baseAddress + pageId * META_DATA_BYTES_PER_PAGE;
+    }
+
+    private long offLock( long pageRef )
```
ragadeeshu: Curious method names

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    public int getSwapperId( long pageRef )
+    {
+        return UnsafeUtil.getInt( offSwapperId( pageRef ) );
+    }
+
+    private void setSwapperId( long pageRef, int swapperId )
+    {
+        UnsafeUtil.putInt( offSwapperId( pageRef ), swapperId );
```
ragadeeshu, line 336: No fence for this guy? (And his other `set...` buddies)

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
```diff
@@ -0,0 +1,332 @@
...
+        // Note: It is important that we assign the filePageId before we swap
+        // the page in. If the swapping fails, the page will be considered
+        // loaded for the purpose of eviction, and will eventually return to
+        // the freelist. However, because we don't assign the swapper until the
+        // swapping-in has succeeded, the page will not be considered bound to
+        // the file page, so any subsequent thread that finds the page in their
+        // translation table will re-do the page fault.
+        setFilePageId( pageRef, filePageId ); // Page now considered isLoaded()
+        PageObj page = new PageObj( cachePageSize, address( pageRef ) ); // todo avoid this allocation
+        long bytesRead = swapper.read( filePageId, page );
```
ragadeeshu: The purpose of this line is not immediately clear to me. Care to elaborate?

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/PageListTest.java
```diff
@@ -0,0 +1,1157 @@
...
+        assertFalse( pageList.decrementUsage( nextPageRef ) );
+    }
+
+    @Test
+    public void filePageIdIsUnboundByDefault() throws Exception
+    {
+        assertThat( pageList.getFilePageId( pageRef ), is( PageCursor.UNBOUND_PAGE_ID ) );
+    }
+
+    // xxx ---[ Page fault tests ]---
```
ragadeeshu, line 1302: For the interference tests, perhaps you should validate all fields of all other entries in the list? Or at least all fields of adjacent entries?

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/PageListTest.java
```diff
@@ -0,0 +1,1157 @@
...
+    @Before
+    public void setUp()
+    {
+        pageList = new PageList( pageIds.length, pageSize, mman, VictimPageReference.getVictimPage( pageSize ) );
+        pageRef = pageList.deref( pageId );
+        prevPageRef = pageList.deref( prevPageId );
+        nextPageRef = pageList.deref( nextPageId );
+    }
+
+    // xxx ---[ Sequence lock tests ]---
```
ragadeeshu: Some concurrent read-invalidation and contention for locks of same & different types would be interesting. Maybe a bit annoying to implement though.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
```diff
@@ -71,6 +70,22 @@
...
     }
 
+    /**
+     * This copy-constructor is useful for classes that want to extend the {@code PageList} class to inline its fields.
+     * All data and state will be shared between this and the given {@code PageList}. This means that changing the page
+     * list state through one has the same effect as changing it through the other – they are both effectively the same
+     * object.
+     * @param pageList The {@code PageList} instance whose state to copy.
+     */
+    PageList( PageList pageList )
```
ragadeeshu, line 115: Why do we need this, instead of passing the same object? I'm sure that the explanation is in the comment just above this, but I don't quite get it.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/SwapperSet.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.function.IntConsumer;
+
+import org.neo4j.collection.primitive.Primitive;
+import org.neo4j.collection.primitive.PrimitiveIntIterator;
+import org.neo4j.collection.primitive.PrimitiveIntSet;
+import org.neo4j.io.pagecache.PageSwapper;
+
+final class SwapperSet
+{
+    private volatile Allocation[] allocations = new Allocation[0];
```
ragadeeshu: You have three different locks in this class. Does this need to be volatile, or could you move it into one of the locks? And is this class such a hotspot that there is need to have three separate locks?

ragadeeshu (commented):
This is a first batch in order to not loose these comments as I move forward.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/SwapperSet.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.util.Arrays;
+import java.util.function.IntConsumer;
+
+import org.neo4j.collection.primitive.Primitive;
+import org.neo4j.collection.primitive.PrimitiveIntIterator;
+import org.neo4j.collection.primitive.PrimitiveIntSet;
+import org.neo4j.io.pagecache.PageSwapper;
+
+final class SwapperSet
```
ragadeeshu: A general comment about the purpose of this class would be nice :)

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/LatchMap.java
```diff
@@ -0,0 +1,104 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.io.pagecache.impl.muninn;
+
+import org.neo4j.concurrent.BinaryLatch;
+import org.neo4j.unsafe.impl.internal.dragons.FeatureToggles;
+import org.neo4j.unsafe.impl.internal.dragons.UnsafeUtil;
+
+final class LatchMap
```
ragadeeshu: This class could do with some explanation, I think, both general and method-specific.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
```diff
@@ -62,6 +62,20 @@
...
+    // xxx Two of these can go to the usage counter, and the rest to the swapper id, for up to 2048 swappers.
+    // xxx Do we even need the swapper id at this point? If we can already infer the file page id, the same should be
+    // xxx true for the swapper.
+    // xxx The trouble with this idea is that we won't be able to seamlessly turn the transaction tables into hash
+    // xxx tables when they get too big - at least not easily - since the index into the table no longer corresponds to
+    // xxx the file page id. We might still be able to get away with it, if with the page id we also keep a counter for
+    // xxx how many times the file page id loop around the translation table before arriving at the given entry.
+    // xxx That is, what the entry index should be multiplied by to get the file page id. To do this, we would, however,
+    // xxx have to either grab bits from the page id space, or make each entry more than 4 bytes. This will all depend
+    // xxx on where the cut-off point is. That is, what the max entry capacity for a translation table should be.
```
ragadeeshu, line 77: Intriguing

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
chrisvest: It's so `MuninnPagedFile` can extend `PageList` itself, and get all the relevant state carried over from the page list instance created in `MuninnPageCache`.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
chrisvest: The entities are protected by the transactional lock manager.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
chrisvest: You should not use a read lock to coordinate stores.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
chrisvest: Aww… Doug Lea does it in `StampedLock`. Why can't I? :)

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
chrisvest: No, this is to prevent plain stores within the critical section from reordering _into_ the compareAndSetState, which is theoretically possible, e.g. if it is implemented with LL/SC, or with a mutex. The plain store could then reorder with the store that is the taking of the lock, thus effectively escaping the critical section.

chrisvest: See http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/locks/StampedLock.java?r1=1.62&r2=1.63
And http://jsr166-concurrency.10961.n7.nabble.com/Does-StampedLock-need-a-releaseFence-in-theory-td13563.html

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/OffHeapPageLock.java
chrisvest: There is in principle no harm in coarsening a critical section. It's a _roach motel_; you can move in, but you can never move out.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
chrisvest: Wearing belt and suspenders here because I'm using plain writes, and this is happening before the `OffHeapPageLock` takes over the responsibility of protecting the pages and their meta-data.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
chrisvest: No, this state is guarded by the `OffHeapPageLock` in various forms.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
chrisvest: `swapper.read` is where we read the page from the file into memory.

PR #9188, opened 2017-04-10
Avoid multiple attempt to open same index on log summary in IndexingService

MishaDemianenko:
Collect indexes state to log summary during init and start of IndexService
to avoid multiple attempts and lucene failures to open same index multiple times on startup.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
@@ -23,6 +23,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -190,7 +191,7 @@ public class IndexingService extends LifecycleAdapter
     {
         IndexMap indexMap = indexMapRef.indexMapSnapshot();
 
-        Map<InternalIndexState, List<IndexLogRecord>> indexStates = new HashMap<>();
+        Map<InternalIndexState, List<IndexLogRecord>> indexStates = new EnumMap<>( InternalIndexState.class );
         for ( IndexRule indexRule : indexRules )
         {
             IndexProxy indexProxy;
@@ -251,13 +252,14 @@ public class IndexingService extends LifecycleAdapter
         IndexMap indexMap = indexMapRef.indexMapSnapshot();
 
         final Map<Long,RebuildingIndexDescriptor> rebuildingDescriptors = new HashMap<>();
-        Map<InternalIndexState, List<IndexLogRecord>> indexStates = new HashMap<>();
+        Map<InternalIndexState, List<IndexLogRecord>> indexStates = new EnumMap<>( InternalIndexState.class );
 
         // Find all indexes that are not already online, do not require rebuilding, and create them
         indexMap.foreachIndexProxy( ( indexId, proxy ) -> {
             InternalIndexState state = proxy.getState();
             IndexDescriptor descriptor = proxy.getDescriptor();
-            indexStates.computeIfAbsent( state, internalIndexState -> new ArrayList<>() ).add( new IndexLogRecord(indexId, descriptor) );
+            indexStates.computeIfAbsent( state, internalIndexState -> new ArrayList<>() )
+                    .add( new IndexLogRecord( indexId, descriptor ) );
             log.debug( indexStateInfo( "start", indexId, state, descriptor ) );
             switch ( state )
             {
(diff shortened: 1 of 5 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -190,6 +190,7 @@ public void init()
     {
         IndexMap indexMap = indexMapRef.indexMapSnapshot();
 
+        Map<InternalIndexState, List<IndexLogRecord>> indexStates = new HashMap<>();
```
chrisvest: Missed an opportunity to use an `EnumMap` :)

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -281,11 +251,13 @@ public void start() throws Exception
         IndexMap indexMap = indexMapRef.indexMapSnapshot();
 
         final Map<Long,RebuildingIndexDescriptor> rebuildingDescriptors = new HashMap<>();
+        Map<InternalIndexState, List<IndexLogRecord>> indexStates = new HashMap<>();
 
         // Find all indexes that are not already online, do not require rebuilding, and create them
         indexMap.foreachIndexProxy( ( indexId, proxy ) -> {
             InternalIndexState state = proxy.getState();
             IndexDescriptor descriptor = proxy.getDescriptor();
+            indexStates.computeIfAbsent( state, internalIndexState -> new ArrayList<>() ).add( new IndexLogRecord(indexId, descriptor) );
```
chrisvest: I think you're gonna have checkstyle problems on that IndexLogRecord constructor.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -805,4 +777,51 @@ public IndexSamplingController getSamplingController()
...
         return samplingController;
     }
+
+    private void logIndexStateSummary( String method,  Map<InternalIndexState, List<IndexLogRecord>> indexStates )
+    {
+        int mostPopularStateCount = Integer.MIN_VALUE;
+        InternalIndexState mostPopularState = null;
+        for ( Map.Entry<InternalIndexState,List<IndexLogRecord>> indexStateEntry : indexStates.entrySet() )
+        {
+            if (indexStateEntry.getValue().size() > mostPopularStateCount)
```
chrisvest: Looks like more checkstyle trouble.

enterprise/kernel/src/test/java/org/neo4j/test/TestEnterpriseGraphDatabaseFactory.java
```diff
@@ -44,9 +44,37 @@
...
-        return config -> {
-            config.put( "unsupported.dbms.ephemeral", "false" );
-            return new EnterpriseGraphDatabase( storeDir, config, state.databaseDependencies() );
+        return new GraphDatabaseBuilder.DatabaseCreator() {
+
+            @Override
+            public GraphDatabaseService newDatabase( Map<String,String> config )
+            {
+                config.put( "unsupported.dbms.ephemeral", "false" );
+                return new GraphDatabaseFacadeFactory( DatabaseInfo.ENTERPRISE, EnterpriseEditionModule::new )
```
chrisvest, line 53: It's not easy to see what's being changed here. Also, `TestGraphDatabaseFactory` already has a `TestGraphDatabaseFacadeFactory` class that can maybe extended instead? It creates a `TestDatabasePlatformModule` that already does special stuff for logging.

chrisvest (commented):
I'm concerned about the `TestEnterpriseGraphDatabaseFactory` changes, but the rest looks good.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
MishaDemianenko: oh noooooo, looooser! 🥇

enterprise/kernel/src/test/java/org/neo4j/test/TestEnterpriseGraphDatabaseFactory.java
MishaDemianenko: been there, tried that hehe:) That's so on a first view, it's impossible now to extract common subclass of PlatformModule because of file system problems you can't really pass that as contractor parameter since it will be used in super constructor and you will not gonna be able to customise that before usage.
It requires much bigger refactoring to do that, do not want to do that in maintenance brach

MishaDemianenko:
@chrisvest updated

chrisvest (approved):

PR #9189, opened 2017-04-11
Properly materialize maps in compiled runtime

pontusmelke:
We cannot assume that maps can be modified in place when we materialize them in the compiled runtime since the underlying map may not be mutable. Instead we should make a full copy.


zhenlineo:
@systay Pls ping me when this one gets in.

PR #9191, opened 2017-04-11
Flushes page cache sequentially during some import stages

tinwelint:
To get maximum sequential write I/O. Node and Relationship stages gets
a throughput improvement up to 2x or more, depending on hardware.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
@@ -22,7 +22,6 @@ package org.neo4j.unsafe.impl.batchimport;
 import java.io.File;
 import java.io.IOException;
 import java.util.Set;
-
 import org.neo4j.collection.primitive.PrimitiveLongIterator;
 import org.neo4j.helpers.Exceptions;
 import org.neo4j.helpers.Format;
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
```diff
@@ -303,8 +304,8 @@ private void importRelationships( NodeRelationshipCache nodeRelationshipCache,
 
         long nextRelationshipId = 0;
         Configuration relationshipConfig = withBatchSize( config,
-                neoStore.getRelationshipStore().getRecordsPerPage() );
-        Configuration nodeConfig = withBatchSize( config, neoStore.getNodeStore().getRecordsPerPage() );
+                neoStore.getRelationshipStore().getRecordsPerPage() * 100 );
+        Configuration nodeConfig = withBatchSize( config, neoStore.getNodeStore().getRecordsPerPage() * 100 );
```
chrisvest: A wild `* 100` appears, with no apparent explanation.

tinwelint: Didn't think that got it in. I'll split that out to another commit and clean that up

tinwelint:
The second commit went in here as to reduce number of PRs coming for the importer. It's fairly small.

chrisvest (approved):
Looks good to me! 👍

PR #9194, opened 2017-04-11
Don't startup single/HA if CC's cluster-state directory is present

mneedham:
Don't merge this yet, just want to see if all the tests pass


systay:
Re-open if it should still be around

PR #9197, opened 2017-04-12
Adds IndexStoreView to StoreMigration tool

tinwelint:
because it's required by label scan store dependencies.
Previously some tests failed with unsatisfied dependency on label index
since this view couldn't be found.


burqen:
@tinwelint What test was failing because of this?

tinwelint:
@burqen flaky tests at the time, I've rebased now

burqen:
This is wrong fix. Correct fix is already merged #9195

PR #9198, opened 2017-04-12
Flushes page cache sequentially during some import stages

tinwelint:
To get maximum sequential write I/O. Node and Relationship stages gets
a throughput improvement up to 2x or more, depending on hardware.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/PageCacheFlusher.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/PageCacheFlusher.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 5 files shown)


tinwelint:
Replaces #9191

tinwelint:
@chrisvest it's rebased now

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/PageCacheFlusher.java
```diff
@@ -0,0 +1,88 @@
...
+                error = e;
+                break;
+            }
+        }
+        done = true;
+        Thread localHalter = halter;
+        if ( localHalter != null )
+        {
+            LockSupport.unpark( localHalter );
+        }
```
chrisvest: You could use a `BinaryLatch` or a `CountDownLatch` instead of this halter business.

chrisvest (commented):
Looks good, but the `PageCacheFlusher` can be simplified.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/PageCacheFlusher.java
tinwelint: Good idea, I'm trying `BinaryLatch`

chrisvest:
I think the merge of this should wait for #9215.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/PageCacheFlusher.java
tinwelint: Updated now, have a look

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/PageCacheFlusher.java
```diff
@@ -35,9 +32,8 @@
 {
     private final PageCache pageCache;
     private volatile boolean halted;
-    private volatile boolean done;
+    private volatile BinaryLatch halt = new BinaryLatch();
```
chrisvest: This field can be `final`.

tinwelint: Amended to the last commit

chrisvest (approved):

tinwelint:
Some weirdness with branch locally, replaced by #9235

PR #9202, opened 2017-04-12
Notices bad health of DynamicTaskExecutor before being closed

tinwelint:
this to avoid a race where receiving a panic would mark panic and
then close. There was a window in between where poking that executor
wouldn't notice the panic, something that TicketedProcessing relied on
doing.


burqen:
@tinwelint Test failure looks relevant and it fails consistently locally.

tinwelint:
@burqen Test should be fixed now

burqen (approved):

PR #9203, opened 2017-04-12
Avoids NPE in SystemDiagnostics on null ClassLoader URLs

tinwelint:
This happened on an IBM JDK. Unexpected as it is there's no point failing
with NPE. Rather now this fact will be logged.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/info/SystemDiagnostics.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/info/SystemDiagnostics.java
@@ -50,6 +50,7 @@ import org.neo4j.kernel.impl.util.OsBeanUtil;
 import org.neo4j.logging.Logger;
 
 import static java.net.NetworkInterface.getNetworkInterfaces;
+
 import static org.neo4j.helpers.Format.bytes;
 
 enum SystemDiagnostics implements DiagnosticsProvider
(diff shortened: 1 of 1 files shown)


community/kernel/src/main/java/org/neo4j/kernel/info/SystemDiagnostics.java
```diff
@@ -160,9 +160,17 @@ void dump( Logger logger )
...
                     URLClassLoader urls = (URLClassLoader) loader;
-                    for ( URL url : urls.getURLs() )
-                        if ( "file".equalsIgnoreCase( url.getProtocol() ) )
-                            paths.put( url.toString(), pathValue( paths, "loader." + level, url.getPath() ) );
+                    URL[] classLoaderUrls = urls.getURLs();
+                    if ( classLoaderUrls != null )
+                    {
+                        for ( URL url : urls.getURLs() )
+                            if ( "file".equalsIgnoreCase( url.getProtocol() ) )
+                                paths.put( url.toString(), pathValue( paths, "loader." + level, url.getPath() ) );
```
chrisvest: The loop can use the `classLoaderUrls` variable, instead of constructing the array again.

Also, the formatting can be improved here. Otherwise, the forward merger will have to do it to make it build again.

chrisvest (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/info/SystemDiagnostics.java
tinwelint: My impression was that we generally change as little as possible in maint versions. I.e. chaning code style should not be done here. Am I wrong?

tinwelint: Although in this case the lines are changing _anyway_, so we can might as well do it. Perhaps that's why you saw this opportunity?

chrisvest: Yes, exactly.

chrisvest (approved):

PR #9204, opened 2017-04-12
3.1 ratelimit source

OliviaYtterbrink:
Rate limit authentication attempts per source instead of per user.
This solved an issue where it was possible to lock out a user from the server, just by attempting to log in with wrong credentials.

changelog: Limit authentication requests depending on their originating source

### DIFF
--- a/community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/integration/BoltConnectionIT.java
+++ b/community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/integration/BoltConnectionIT.java
@@ -55,6 +55,7 @@ public class BoltConnectionIT
 {
     private static final Map<String,Object> EMPTY_PARAMS = emptyMap();
     private static final String USER_AGENT = "BoltConnectionIT/0.0";
+    private static final String SOURCE = "127.0.0.1";
 
     @Rule
     public SessionRule env = new SessionRule();
@@ -63,7 +64,7 @@ public class BoltConnectionIT
     public void shouldCloseConnectionAckFailureBeforeInit() throws Throwable
     {
         // Given
-        BoltStateMachine machine = env.newMachine( "<test>" );
+        BoltStateMachine machine = env.newMachine( SOURCE, "<test>" );
 
         // when
         BoltResponseRecorder recorder = new BoltResponseRecorder();
@@ -77,7 +78,7 @@ public class BoltConnectionIT
     public void shouldCloseConnectionResetBeforeInit() throws Throwable
     {
         // Given
-        BoltStateMachine machine = env.newMachine( "<test>" );
+        BoltStateMachine machine = env.newMachine( SOURCE, "<test>" );
 
         // when
         BoltResponseRecorder recorder = new BoltResponseRecorder();
@@ -91,7 +92,7 @@ public class BoltConnectionIT
     public void shouldCloseConnectionOnRunBeforeInit() throws Throwable
     {
         // Given
-        BoltStateMachine machine = env.newMachine( "<test>" );
+        BoltStateMachine machine = env.newMachine( SOURCE, "<test>" );
 
         // when
         BoltResponseRecorder recorder = new BoltResponseRecorder();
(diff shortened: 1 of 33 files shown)


community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/integration/BoltConnectionIT.java
```diff
@@ -133,8 +134,8 @@ public void shouldCloseConnectionOnPullAllBeforeInit() throws Throwable
     public void shouldExecuteStatement() throws Throwable
     {
         // Given
-        BoltStateMachine machine = env.newMachine( "<test>" );
-        machine.init( USER_AGENT, emptyMap(), null );
+        BoltStateMachine machine = env.newMachine( SOURCE, "<test>" );
+        machine.init( USER_AGENT, EMPTY_MAP, null );
```
fickludd, line 138: Helper method for these two lines could save some boilerplate in this class.

community/security/src/main/java/org/neo4j/server/security/auth/RateLimitedAuthenticationStrategy.java
```diff
@@ -87,14 +87,27 @@ public AuthenticationResult authenticate( User user, String password)
...
+        if ( authMetaMap == null )
+        {
+            authMetaMap = new ConcurrentHashMap<>();
+            ConcurrentMap<String,AuthenticationMetadata> preExisting = authenticationData.putIfAbsent( username, authMetaMap );
+
+            if ( preExisting != null )
+            {
+                authMetaMap = preExisting;
+            }
+        }
```
fickludd, line 103: Instead of this dance, how about `ConcurrentMap.computeIfAbsent`? That would replace both `get` and `putIfAbsent` and guarantees concurrently nice behaviour.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ThreadedTransaction.java
```diff
@@ -102,22 +104,34 @@ public Throwable apply( S subject )
...
                         {
+                            if ( state == LatchState.NOT_STARTED )
+                            {
+                                latch.start();
+                                state = LatchState.STARTED;
+                            }
+                            if ( state == LatchState.STARTED )
+                            {
+                                latch.finish();
+                            }
```
fickludd, line 134: Eh, how is this related to rate limiting? Confused. And also I don't really understand when this code is supposed to kick in - the inner finally block ends with a `state = LatchState.FINISHED` so it should always be finished?

fickludd (commented):
Hey folks, I randomly saw this PR and decided to have a look. 

I must say that I'm kinda sceptical about _changing_ this behaviour in a _patch release_. How about config + retained default behaviour in `3.1`+`3.2` => full monkey in `3.3`?

Also did you think about the fact that this change now allows attempts per source + user, allows me to scale my brute-force attempt with my drone computer farm size? This reduces the strength of the rate limit by a few orders of magnitude.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/ThreadedTransaction.java
OliviaYtterbrink: You're correct in part, there's some redundant checks in this block, since it will always be in one of two states. But it's required that the latch be counted down since otherwise the tests will hang if there's an error in the outer try block.

community/security/src/main/java/org/neo4j/server/security/auth/RateLimitedAuthenticationStrategy.java
OliviaYtterbrink: Good suggestion

community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/integration/BoltConnectionIT.java
OliviaYtterbrink: +1

OliviaYtterbrink:
Since this is such a big change, we've decided to close this PR and have a look at if this should go in at all and if so then it should be configurable.

PR #9211, opened 2017-04-13
Native LSS retry rebuild | GBPTree clean crash internally | Flush LSS on import

burqen:
### Part 1 - Keep track of rebuild state for native label scan store
Before, if failed to complete rebuild of native label scan store db startup
would crash. But on next startup it would act as if nothing is wrong and
no new rebuild would be triggered, effectively ending up with a completely
inconsistent label scan store. Now we mark native label scan store as dirty before starting rebuild from full node scan and mark it as clean once it finishes. If native label scan store is
dirty during startup a complete rebuild will be triggered.
- Keep track of rebuild state as part of label scan store specific header in underlying GBPTree.

### Part 2 - Make cleaning of crashed pointers completely internal to GBPTree
Previously, prepare for checkpoint (bumping unstable generation to recognise crashed pointers) and finish checkpoint (trigger cleanup of crashed pointers) would have to be triggered explicitly from the outside. Now this is handled completely internally.
- Adding state 'clean' and 'dirty' to GBPTree. 'Clean' means tree was closed without any changes waiting to be checkpointed. 'Dirty' is all other scenarios, typically close without checkpoint or crash.
- If tree state on startup is 'dirty' the tree will automatically clean crashed pointers as part of construction.
- Native label scan store does not need to tell underlying GBPTree about recovery anymore as cleaning crashed pointers is handled internally.
- Bump format version because of added 'clean/dirty' flag in tree state.

### Part 3 - Explicitly flush label scan store during close of importer
Previously, import would finish but when starting db on imported store the (native) label scan store would appear be completely empty because the underlying GBPTree was not checkpointed.
- Explicitly flush label scan store as part of close of importer

Supersedes #9182 and #9215


burqen:
Test failure is a flaky causal clustering test

community/index/src/main/java/org/neo4j/index/internal/gbptree/Header.java
```diff
@@ -48,7 +48,9 @@
 
     static final Writer CARRY_OVER_PREVIOUS_HEADER = (from,length,to) ->
     {
-        from.copyTo( from.getOffset(), to, to.getOffset(), length );
+        int toOffset = to.getOffset();
+        from.copyTo( from.getOffset(), to, toOffset, length );
+        to.setOffset( toOffset + length );
```
tinwelint, line 53: Was there a quirk with `PageCursor` which made this change necessary?

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeStatePair.java
```diff
@@ -118,11 +118,18 @@ else if ( stateA.isValid() && stateB.isValid() )
...
+            // and include sanity check for stable generations such that there cannot be a state S compared
+            // to other state O where
             // S.unstableGeneration > O.unstableGeneration AND S.stableGeneration < O.stableGeneration
 
-            if ( stateA.stableGeneration() >= stateB.stableGeneration() &&
+            if ( stateA.stableGeneration() == stateB.stableGeneration() &&
+                    stateA.unstableGeneration() == stateB.unstableGeneration() &&
+                    stateA.isClean() != stateB.isClean() )
+            {
+                return Optional.of( stateA.isClean() ? stateA : stateB );
```
tinwelint, line 130: Should we throw if `stateA.isClean() == stateB.isClean()`?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -103,19 +101,16 @@
...
  * The writes that happened before the last checkpoint are durable and safe, but the writes after it are not.
- * The tree can however get back to a consistent state by:
- * <ol>
- * <li>Creator of this tree detects that recovery is required (i.e. non-clean shutdown) and if so must call
- * {@link #bumpUnstableGeneration()} ones, before any writes during recovery are made.</li>
- * <li>Replaying all the writes, exactly as they were made, since the last checkpoint all the way up
- * to the crash ({@code x}). Even including writes before the last checkpoint is OK, important is that
- * <strong>at least</strong> writes since last checkpoint are included.
- * </ol>
+ * The tree can however get back to a consistent state by replaying all the writes, exactly as they were made,
```
tinwelint: Good update to the javadocs. `exactly as they were made` sounds stricter than it actually is and somewhat vague at the same time. It could instead mention that all changes has to be replayed (which is well described in your changes below), but that the order isn't strictly that of the changes as they happened in the first place.

tinwelint (changes requested):
In general a very nice change to have GBPTree be a bit more easy to use.

community/index/src/main/java/org/neo4j/index/internal/gbptree/Header.java
burqen: Yes, `copyTo` will not actually move the `to` cursor, which means the header length is gonna be zero when we calculate it by calculating offset difference from before and after writing header.

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeStatePair.java
burqen: For other illegal states like that, for example `stateA.stable > stateB.stable` + `stateA.unstable < stateB.unstable`, we just return `Optional.empty`. Error handling is different depending on if we ask for newest or oldest tree state and is handled elsewhere.
This illegal state should be handled the same way.

burqen:
@tinwelint Comments addressed 🎉

tinwelint (approved):

PR #9215, opened 2017-04-18
Flush label scan store when closing importer

burqen:
After opening db on a freshly imported store the (native) label scan store would not see the imported labels because the label scan store was never flushed and so the underlying GBPTree was never checkpointed.

This PR fixes this problem by explicitly flushing the label scan store as part of `BatchingNeoStores.close()`.

Build on #9211 (only last commit is new for this PR)


burqen:
Changes from this PR included in #9211

PR #9216, opened 2017-04-18
Utilizes available memory and imports multiple types at a time

tinwelint:
Instead of only a single type at a time. This re-introduces caching for
keeping relationship heads for multiple types per dense node in
NodeRelationshipCache. This will get the best of both the previous strategy
of always importing all relationships in one round AND the current strategy
of importing one type per round (to reduce memory consumption).
Now the amount of available memory will decide how many rounds of relationship
imports are required.

This also removes the per-type-splitting on-disk caching of the input data
which was done when importing relationships of the first (and biggest) relationship type,
something which in most case will improve performance of the import in general
and avoid sections of relationship import where seemingly there were no progress,
due to only caching relationships.

There's now an additional setting for how much memory the importer can use
as a whole and is by default based on amount of free physical memory on the machine.

The defragmentation of relationship groups after relationship import also makes
use of available memory and can reduce number of rounds needed.

changelog: Utilizes available off-heap memory and imports relationships for multiple types at a time. Maximum amount of memory is by default to use 90% of free memory on the machine, but can be explicitly overridden using `--max-memory` option in `neo4j-import`.


community/consistency-check/src/test/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporterTest.java
```diff
@@ -124,6 +125,17 @@ public int maxNumberOfProcessors()
...
         }
+
+        @Override
+        public long maxMemoryUsage()
+        {
+            // This calculation is just to try and hit some sort of memory limit so that relationship import
+            // is split up into multiple rounds. Also to see that relationship group defragmentation works
+            // well when doing multiple rounds.
+            double ratio = (NODE_COUNT / 1_000D );
+            long mebi = 1024 * 1024;
```
MishaDemianenko: we already have ByteUnit that contains MebiByte definition

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -232,7 +234,13 @@
         READ_BUFFER_SIZE( "read-buffer-size", org.neo4j.csv.reader.Configuration.DEFAULT.bufferSize(),
                 "<bytes, e.g. 10k, 4M>",
                 "Size of each buffer for reading input data. It has to at least be large enough to hold the " +
-                "biggest single value in the input data." );
+                "biggest single value in the input data." ),
+        MAX_MEMORY( "max-memory", null,
+                "<max memory that importer can use>",
+                "(advanced) Maximum memory that importer can use for various data structures and caching " +
+                "to improve performance. By default set to " + DEFAULT_MAX_MEMORY_PERCENT +
```
MishaDemianenko: Description say that default value is `DEFAULT_MAX_MEMORY_PERCENT` in the same time default option value is `null` that is confusing

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -451,6 +461,20 @@ public static void main( String[] incomingArguments, boolean defaultSettingsSuit
         }
     }
 
+    private static Long parseMaxMemory( String maxMemoryString )
+    {
+        if ( maxMemoryString != null )
+        {
+            if ( maxMemoryString.endsWith( "%" ) )
```
MishaDemianenko: we at least need to trim user provided string before trying to figure out what is inside of it

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NodeRelationshipCache.java
```diff
@@ -626,4 +864,48 @@ private boolean chunkHasChange( long nodeId, byte chunkMask )
...
+                long currentSetOfRelationshipsMemoryUsage = 0;
+                Set<Object> typesToImportThisRound = new HashSet<>();
+                while ( peekableAllTypes.hasNext() )
+                {
+                    // Calculate worst-case scenario
+                    Map.Entry<Object,MutableLong> type = peekableAllTypes.peek();
+                    long relationshipCountForThisType = type.getValue().longValue();
+                    long maxDenseNodesForThisType = min( numberOfDenseNodes, relationshipCountForThisType * 2/*nodes/rel*/ );
+                    long memoryUsageForThisType = maxDenseNodesForThisType * NodeRelationshipCache.GROUP_ENTRY_SIZE;
+                    if ( currentSetOfRelationshipsMemoryUsage + memoryUsageForThisType > freeMemoryForDenseNodeCache &&
```
MishaDemianenko: would be nice to have clarifying parenthesis here

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
```diff
@@ -40,6 +43,7 @@
      */
     String BAD_FILE_NAME = "bad.log";
     long MAX_PAGE_CACHE_MEMORY = mebiBytes( 240 );
+    int DEFAULT_MAX_MEMORY_PERCENT = 90;
```
MishaDemianenko, line 46: How does this work together with page cache that will claim 50% by default?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
```diff
@@ -159,4 +175,25 @@ public int batchSize()
             }
         };
     }
+
+    public static long calculateMaxMemoryFromPercent( int percent )
```
MishaDemianenko: there is no need to declare it `public`

MishaDemianenko (changes requested):
Couple of comments

community/consistency-check/src/test/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporterTest.java
tinwelint: Yeah, slipped my mind somehow...

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
tinwelint: Yeah, although thing is that we can't quite have a default value there because this is also used to generate manual contents and so it would be specific to the machine that generated it. Lemme try to rephrase that a bit

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
tinwelint: Might be a good idea

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NodeRelationshipCache.java
tinwelint: Yeah or even a variable which would be able to explain what it is

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
tinwelint: It won't, in the import case... it will use max whatever the above line says

tinwelint:
@MishaDemianenko comments addressed

MishaDemianenko (approved):

PR #9217, opened 2017-04-18
Do not leak cursors

davidegrohmann:
This PR replaces #9184

### DIFF
--- a/community/cypher/cypher-compiler-3.2/src/test/scala/org/neo4j/cypher/internal/compiler/v3_2/spi/QueryContextAdaptation.scala
+++ b/community/cypher/cypher-compiler-3.2/src/test/scala/org/neo4j/cypher/internal/compiler/v3_2/spi/QueryContextAdaptation.scala
@@ -32,6 +32,8 @@ trait QueryContextAdaptation {
 
   override type EntityAccessor = this.type
 
+  override def resources: CloseableResource = ???
+
   override def indexScanByContains(index: IndexDescriptor, value: String): scala.Iterator[Node] = ???
 
   override def indexScanByEndsWith(index: IndexDescriptor, value: String): Iterator[Node] = ???
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/ResourceManager.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/ResourceManager.scala
@@ -23,6 +23,7 @@ import java.util
 import java.util.Collections.newSetFromMap
 
 import org.neo4j.cypher.internal.compiler.v3_2.spi.CloseableResource
+import org.neo4j.helpers.Exceptions
 
 class ResourceManager extends CloseableResource {
   private val resources: util.Set[AutoCloseable] = newSetFromMap(new util.IdentityHashMap[AutoCloseable, java.lang.Boolean]())
@@ -41,10 +42,14 @@ class ResourceManager extends CloseableResource {
 
   override def close(success: Boolean): Unit = {
     val iterator = resources.iterator()
-    while ( iterator.hasNext )
-    {
-      iterator.next().close()
+    var error: Throwable = null
+    while (iterator.hasNext) {
+      try iterator.next().close()
+      catch {
+        case t: Throwable => error = Exceptions.chain(error, t)
+      }
       iterator.remove()
     }
+    if (error != null) throw error
   }
 }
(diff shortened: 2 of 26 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/AggregationAcceptanceTest.scala
```diff
@@ -23,12 +23,6 @@ import org.neo4j.cypher.{ExecutionEngineFunSuite, NewPlannerTestSupport}
 
 class AggregationAcceptanceTest extends ExecutionEngineFunSuite with NewPlannerTestSupport {
 
-  test("apa") {
```
systay, line 26: Oops :)

systay:
This change feels much cleaner and easier to understand than the ThreadLocal variation in #9184. 👍 

There are no unit tests for the new ResourceManager, but other than that, LGTM

davidegrohmann:
@systay I have added the tests now

community/cypher/cypher-compiler-3.2/src/test/scala/org/neo4j/cypher/internal/compiler/v3_2/spi/QueryContextAdaptation.scala
```diff
@@ -32,6 +32,8 @@ trait QueryContextAdaptation {
 
   override type EntityAccessor = this.type
 
+  override def resources: CloseableResource = ???
```
pontusmelke, line 35: is this safe to leave unimplemented? If so maybe a comment to future readers :)

Oh I see we do a lot of that already, never mind then...

davidegrohmann: Basically almost all the methods in that class are unimplemented, moreover that class is used only in tests. I don't think there is any need to add any explanation.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/ResourceManager.scala
```diff
@@ -0,0 +1,50 @@
...
+    if (!resources.remove(resource)) {
+      throw new IllegalStateException(s"$resource is not in the resource set $resources")
+    }
+  }
+
+  override def close(success: Boolean): Unit = {
+    val iterator = resources.iterator()
+    while ( iterator.hasNext )
+    {
+      iterator.next().close()
```
thobe: Should we be handling potential exceptions from the `close()` method?

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/spi/BeansAPIRelationshipIteratorTest.scala
```diff
@@ -0,0 +1,55 @@
...
+    val iterator = new BeansAPIRelationshipIterator(relationships, mock[NodeManager], new ResourceManager)
+
+    iterator.hasNext
+    verify(relationships).hasNext
+    verify(relationships, times(0)).close()
+
+    iterator.close()
+    verify(relationships).close()
+    verifyNoMoreInteractions(relationships)
+  }
```
thobe, line 54: How about a test case to ensure that `close()` is idempotent?

thobe (commented):
Looks like a nice approach to solving the resource leak hazard.

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/spi/BeansAPIRelationshipIteratorTest.scala
davidegrohmann: Good point, I'll add one.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/ResourceManager.scala
davidegrohmann: Yep, you are right we should.

community/cypher/cypher-compiler-3.2/src/test/scala/org/neo4j/cypher/internal/compiler/v3_2/spi/QueryContextAdaptation.scala
pontusmelke: 👍

systay:
To be merged straight after release of Neo4j 3.2

PR #9220, opened 2017-04-19
Do not sort property ids from schema descriptors as part of PropertyE…

MishaDemianenko:
…xistenceEnforcer creation.

Avoid sorting of property ids as part of enforcer creation since that will change
property ids order in index descriptor, descriptors hashcode that as a result can cause
incorrect index updates, inability to find an index with correct descriptor etc.

### DIFF
--- a/enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
+++ b/enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
@@ -71,23 +71,31 @@ class PropertyExistenceEnforcer
         this.relationshipConstraints = rels;
         for ( LabelSchemaDescriptor constraint : nodes )
         {
-            update( mandatoryNodePropertiesByLabel, constraint.getLabelId(), constraint.getPropertyIds() );
+            update( mandatoryNodePropertiesByLabel, constraint.getLabelId(), copyIds( constraint.getPropertyIds() ) );
         }
         for ( RelationTypeSchemaDescriptor constraint : rels )
         {
-            update( mandatoryRelationshipPropertiesByType, constraint.getRelTypeId(), constraint.getPropertyIds() );
+            update( mandatoryRelationshipPropertiesByType, constraint.getRelTypeId(),
+                    copyIds( constraint.getPropertyIds() ) );
         }
     }
 
     private static void update( PrimitiveIntObjectMap<int[]> map, int key, int[] values )
     {
-        Arrays.sort( values );
         int[] current = map.get( key );
         if ( current != null )
         {
             values = union( current, values );
         }
         map.put( key, values );
+    }
+
+    private static int[] copyIds( int[] propertyIds )
+    {
+        int[] values = new int[propertyIds.length];
+        System.arraycopy( propertyIds, 0, values, 0, propertyIds.length );
+        Arrays.sort( values );
+        return values;
     }
 
     TxStateVisitor decorate( TxStateVisitor visitor, ReadableTransactionState txState, StoreReadLayer storeLayer )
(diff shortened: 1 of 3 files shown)


enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
```diff
@@ -90,6 +90,14 @@ private static void update( PrimitiveIntObjectMap<int[]> map, int key, int[] val
         map.put( key, values );
     }
 
+    private static int[] copyIds( int[] propertyIds )
```
fickludd: Maybe `copyAndSort`?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
```diff
@@ -71,17 +71,17 @@ private PropertyExistenceEnforcer( List<LabelSchemaDescriptor> nodes, List<Relat
...
         {
-            update( mandatoryRelationshipPropertiesByType, constraint.getRelTypeId(), constraint.getPropertyIds() );
+            update( mandatoryRelationshipPropertiesByType, constraint.getRelTypeId(),
+                    copyIds( constraint.getPropertyIds() ) );
         }
     }
 
     private static void update( PrimitiveIntObjectMap<int[]> map, int key, int[] values )
     {
-        Arrays.sort( values );
```
fickludd: As `values` now is assumed to be sorted, it would be nice to either
a) add an assert to check that it's sorted
or at least
b) rename it into `sortedValues` or similar

fickludd (changes requested):
Excellent find, this should indeed help solve things. Minor code cleanliness suggestions in review, otherwise great.

enterprise/neo4j-enterprise/src/test/java/org/neo4j/CompositeConstraintIT.java
```diff
@@ -0,0 +1,103 @@
...
+
+        Label label = Label.label( "label" );
+
+        try ( Transaction transaction = database.beginTx() )
+        {
+            Node node = database.createNode( label );
+            node.setProperty( "b", (short) 3 );
+            node.setProperty( "a", new double[]{0.6, 0.4, 0.2} );
+            transaction.success();
+        }
```
chrisvest, line 64: Maybe these properties should be set in different transactions to enforce the property key creation order?

chrisvest: Still puzzled as to why the other tests we tried to write didn't manage to capture then scenario. This looks deceptively similar to what we tried earlier.

MishaDemianenko: this initial transaction enforce IDs for properties, that should be opposite to order of properties in the defined constraint;
since I do not have it locally it's hard to say but let's check it ASAA

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
MishaDemianenko: renamed to `copyAndSortPropertyIds`

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/PropertyExistenceEnforcer.java
MishaDemianenko: renamed to `sortedValues`; since it's private static method I'm a bit hesitating that we need additional assertion, do you still think we need it?

MishaDemianenko:
@fickludd updated

fickludd (approved):

chrisvest (approved):

chrisvest:
The `LdapAuthIT.shouldTimeoutIfInvalidLdapServer` test failure is a known flaky.

PR #9222, opened 2017-04-19
Floating point comparisons by scaling up precision

systay:
When comparing a JVM float with a JVM double, indexes behaved
differently from the runtime comparison. The runtime comparison opted
for scaling down the precision to compare between two floats,
whereas the indexes scaled up precision.

This change makes sure to have the same semantics no matter if an index
exists or not.

changelog: [3.1]


systay:
This is already solved in Neo4j 3.2 and should be null-merged forward.

fickludd (approved):

PR #9223, opened 2017-04-19
Restore ability to use jline based console for shell

MishaDemianenko:
Restore required public constructors;
Add a test to detect jline based console creation problems in future.


MishaDemianenko:
@jexp this should fix the problem with console creation for the shell. Can you please take a look, thx

PR #9224, opened 2017-04-19
Give neo4j-admin commands the ability to prompt admins

chrisvest:
No tests are added, because nothing else in `RealOutsideWorld` has tests, and
also the implementations of these new methods are quite trivial.


spacecowboy (approved):
LGTM please merge at your discretion

Issue #9121, opened 2017-03-30
(Lucene) Database Panic - multiple Problems - kills Neo4j


jastax:
- Neo4j version: 3.1.3
- API: HTTP API

# UPDATE 
I can now reproduce it.

1) CALL apoc.index.addAllNodes("search_summary" [....]) 
and while the indexing is running
2) CALL apoc.index.nodes('search_summary" [...])

crashes neo4j

# Bug report

We are running into a database panic when our script removes nodes from an index and adds a new set of nodes after that. Neo4j stops working and needs to be restarted.

A shell script is executed to update the database - the relevant action takes place with the following call:

**
apoc.index.addAllNodes("index", ...)**
```
2017-03-30 09:05:21.602+0000 INFO  [o.n.k.i.t.l.c.CheckPointerImpl] Check Pointing triggered by scheduler for time threshold [1285818]:  Starting store flush...
2017-03-30 09:05:21.900+0000 INFO  [o.n.k.i.p.Procedures] Dropping existing index 'search_summary', with config: {keysForLabel:Document=title:teaser:search_summary, type=fulltext, to_lower_case=true, provider=lucene, labels=Docume
nt}
2017-03-30 09:05:21.922+0000 ERROR [o.n.k.i.DatabaseHealth] Database panic: Database has encountered some problem, please perform necessary action (tx recovery/restart) cannot close: prepareCommit was already called with no corresponding call to commit
java.lang.IllegalStateException: cannot close: prepareCommit was already called with no corresponding call to commit
        at org.apache.lucene.index.IndexWriter.shutdown(IndexWriter.java:1070)
        at org.apache.lucene.index.IndexWriter.close(IndexWriter.java:1125)
        at org.neo4j.index.impl.lucene.legacy.WritableIndexReference.disposeWriter(WritableIndexReference.java:69)
        at org.neo4j.index.impl.lucene.legacy.WritableIndexReference.dispose(WritableIndexReference.java:50)
```

From lucene's IndexWriter Documentation:

```
/**
   * Closes all open resources and releases the write lock.
   *
   * If IndexWriterConfig#commitOnClose is true,
   * this will attempt to gracefully shut down by writing any
   * changes, waiting for any running merges, committing, and closing.
   * In this case, note that:
   *
   *   <li>If you called prepareCommit but failed to call commit, this
   *       method will throw IllegalStateException and the IndexWriter
   *       will not be closed.</li>
   *   <li>If this method throws any other exception, the IndexWriter
[...]

PR #9227, opened 2017-04-20
Fixes race between starting transaction and shutdown

tinwelint:
This was designed to work already, but there was a slight gap in that
armor. DatabaseAvailability is the among last components to register in
the database lifecycle and therefore one of the first to be stopped.
Stopping DatabaseAvailability will raise the availability guard and then
await all active transactions to be finished (with some configurable
timeout). Starting transactions also checked the availability guard
state and registered themselves as active.

The problem was that the active state wasn't registered until after the
availability guard had been checked and so this little window between
those calls was racing with that same window in DatabaseAvailability
potentially allowing a transaction to start without being aware that
database was shutting down and shutdown process being unaware of an
active transaction running concurrently. This could result in
unspecified exceptions being thrown due to transaction running inside
components which were shutting down or even had been shut down.

This commit tightens this gap by letting transaction starting register
itself as active before checking availability guard. This way
transactions starting concurrently with shutdown will deterministically
succeed or fail, all depending on whether they made it in before the
availability guard was raised for shutdown.

This still leaves the max timeout that DatabaseAvailability waits before
going ahead with the shut down anyway. But that is by default a timeout
reasonable for most setups and it's configurable.

### DIFF
--- a/community/io/src/test/java/org/neo4j/test/rule/PageCacheAndDependenciesRule.java
+++ b/community/io/src/test/java/org/neo4j/test/rule/PageCacheAndDependenciesRule.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.test.rule;
+
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+import java.util.function.Supplier;
+
+import org.neo4j.io.fs.FileSystemAbstraction;
+import org.neo4j.io.pagecache.PageCache;
+import org.neo4j.test.rule.fs.EphemeralFileSystemRule;
+import org.neo4j.test.rule.fs.FileSystemRule;
+
+/**
+ * Very often when you want a {@link PageCacheRule} you also want {@link TestDirectory} and some {@link FileSystemRule}.
+ * This is tedious to write and apply in the correct order in every test doing this. This rule collects
+ * this threesome into one rule for convenience.
+ */
+public class PageCacheAndDependenciesRule implements TestRule
+{
+    private final RuleChain chain;
+    private final FileSystemRule<? extends FileSystemAbstraction> fs;
+    private final TestDirectory directory;
+    private final PageCacheRule pageCacheRule = new PageCacheRule();
+
+    public PageCacheAndDependenciesRule()
+    {
+        this( () -> new EphemeralFileSystemRule() );
+    }
+
+    /**
+     * @param fsSupplier as {@link Supplier} to make it clear that it is this class that owns the created
+     * {@link FileSystemRule} instance.
+     */
+    public PageCacheAndDependenciesRule( Supplier<FileSystemRule<? extends FileSystemAbstraction>> fsSupplier )
+    {
+        this.fs = fsSupplier.get();
+        this.directory = TestDirectory.testDirectory( fs );
+        this.chain = RuleChain.outerRule( fs ).around( directory ).around( pageCacheRule );
+    }
+
+    @Override
+    public Statement apply( Statement base, Description description )
+    {
+        return chain.apply( base, description );
+    }
+
+    public FileSystemRule<? extends FileSystemAbstraction> fileSystemRule()
+    {
+        return fs;
+    }
+
+    public FileSystemAbstraction fileSystem()
+    {
+        return fs.get();
+    }
+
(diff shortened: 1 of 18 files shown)


community/io/src/test/java/org/neo4j/test/rule/PageCacheAndDependenciesRule.java
```diff
@@ -0,0 +1,92 @@
...
+import org.neo4j.io.pagecache.PageCache;
+import org.neo4j.test.rule.fs.EphemeralFileSystemRule;
+import org.neo4j.test.rule.fs.FileSystemRule;
+
+/**
+ * Very often when you want a {@link PageCacheRule} you also want {@link TestDirectory} and some {@link FileSystemRule}.
+ * This is tedious to write and apply in the correct order in every test doing this. This rule collects
+ * this threesome into one rule for convenience.
+ */
+public class PageCacheAndDependenciesRule implements TestRule
```
chrisvest, line 39: 👏 👏 👏

community/kernel/src/main/java/org/neo4j/kernel/impl/api/Kernel.java
```diff
@@ -88,9 +96,21 @@ public KernelTransaction newTransaction( KernelTransaction.Type type, SecurityCo
...
+        try
+        {
+            availabilityGuard.await( transactionStartTimeout );
+        }
+        catch ( UnavailableException e )
+        {
+            transactionMonitor.transactionFinished( false, false );
+            throw new TransactionFailureException( Status.Transaction.TransactionStartFailed, e, e.getMessage() );
+        }
+        return transactions.newInstance( type, securityContext, timeout );
```
chrisvest, line 113: 👍 Good comment, too.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTest.java
```diff
@@ -72,6 +97,28 @@ public void shouldNotAllowCreationOfConstraintsWhenInHA() throws Exception
...
+                pageCacheRule.fileSystem(), pageCacheRule.pageCache(), dependencies );
+        dataSource.start();
+        KernelAPI kernel = dataSource.getKernel();
+
+        // WHEN
+        try ( KernelTransaction tx = kernel.newTransaction( explicit, AUTH_DISABLED ) )
+        {
+            // THEN
+            InOrder order = inOrder( transactionMonitor, availabilityGuard );
+            order.verify( transactionMonitor, times( 1 ) ).transactionStarted();
```
chrisvest, line 119: You forgot to verify the call to the availability guard.

chrisvest (changes requested):
Looks good, I just want that test to be fixed.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTest.java
tinwelint: Oh indeed

tinwelint: Now it's added

chrisvest (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/CoreAPIAvailabilityGuard.java
```diff
@@ -50,11 +50,16 @@ public void assertDatabaseAvailable()
...
-            throw new org.neo4j.graphdb.TransactionFailureException( e.getMessage() );
+            throw convertUnavailabilityException( e );
         }
     }
+
+    public RuntimeException convertUnavailabilityException( Exception e )
+    {
+        if ( guard.isShutdown() )
+        {
+            return new DatabaseShutdownException();
```
burqen, line 61: Should we not add initial exception as suppressed?

chrisvest:
@tinwelint There's a weird test failure.

tinwelint:
Looks related actually

tinwelint:
@chrisvest this has now been fixed... should go green

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
```diff
@@ -301,10 +300,6 @@ KernelTransactionHandle createHandle( KernelTransactionImplementation tx )
 
     private void assertRunning()
     {
-        if ( availabilityGuard.isShutdown() )
-        {
-            throw new DatabaseShutdownException();
-        }
```
chrisvest: I think `KernelTransactionsTest.exceptionWhenStartingNewTransactionOnShutdownInstance` is failing because of this change.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
```diff
@@ -301,10 +300,6 @@ KernelTransactionHandle createHandle( KernelTransactionImplementation tx )
 
     private void assertRunning()
     {
-        if ( availabilityGuard.isShutdown() )
```
MishaDemianenko: this check was done on this layer before to ensure that we hold newTransactionsLock when we doing so, now in cases when we block for various reasons on this lock for example and wake up, later on, we will allow the transaction to be started even db is already off do not think that is desired behaviour

MishaDemianenko (changes requested):

jastax:
Will this fix (some of) [9121](https://github.com/neo4j/neo4j/issues/9121)? 

I know you guys are incredibly busy and doing great work, but I need at least some kind of comment on the issue (as in: we have noticed, will look into it)

I'm running a system with a few thousand daily users and have to make a decision on how to go forward as the production database keeps getting killed every few weeks and won't recover without a full restart (and even that is not enough sometimes).

systay:
@tinwelint Any updates to this PR?

ragadeeshu:
@systay This will be updated shortly

ragadeeshu:
@MishaDemianenko updated :)

MishaDemianenko (approved):
@ragadeeshu do not have questions anymore, thx :)

PR #9231, opened 2017-04-20
GBPTree writer fail fast if hitting tree node with valid successor

burqen:
If writer reach a tree node with a valid successor it means that the
tree is not consistent in one out of three ways:
1. Tree state point to a root that is outdated (has a successor)
2. Child pointer in parent does not point to latest successor of child.
3. Sibling pointer does not point to latest successor of sibling.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
@@ -36,6 +36,7 @@ import java.util.function.Supplier;
 import org.neo4j.collection.primitive.Primitive;
 import org.neo4j.collection.primitive.PrimitiveLongSet;
 import org.neo4j.cursor.RawCursor;
+import org.neo4j.helpers.Exceptions;
 import org.neo4j.io.pagecache.CursorException;
 import org.neo4j.io.pagecache.IOLimiter;
 import org.neo4j.io.pagecache.PageCache;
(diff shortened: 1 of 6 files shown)


community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -987,8 +1002,15 @@ public void put( KEY key, VALUE value ) throws IOException
...
-            treeLogic.insert( cursor, structurePropagation, key, value, valueMerger,
-                    stableGeneration, unstableGeneration );
+            try
+            {
+                treeLogic.insert( cursor, structurePropagation, key, value, valueMerger,
+                        stableGeneration, unstableGeneration );
+            }
+            catch ( TreeInconsistencyException e )
+            {
+                throw appendTreeInformation( e );
```
MishaDemianenko, line 1015: why do we not close cursor in this case?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -1024,8 +1046,17 @@ private void setRoot( long rootPointer )
...
-                    stableGeneration, unstableGeneration );
+            VALUE result;
+            try
+            {
+                result = treeLogic.remove( cursor, structurePropagation, key, layout.newValue(),
+                        stableGeneration, unstableGeneration );
+            }
+            catch ( TreeInconsistencyException e )
+            {
+                throw appendTreeInformation( e );
```
MishaDemianenko, line 1060: and here?

community/index/src/main/java/org/neo4j/index/internal/gbptree/PointerChecking.java
```diff
@@ -26,6 +26,12 @@
  */
 class PointerChecking
 {
+    static final String WRITER_TRAVERSE_OLD_STATE_MESSAGE =
+                    "Writer traversed to a tree node that has a valid successor, " +
+                    "meaning writer will not act on latest version of the tree. This is not allowed. " +
+                    "This is most likely due to failure to checkpoint the tree before shutdown and/or tree state " +
```
MishaDemianenko: also as we discuss, message ask something that we do not want our users to do

MishaDemianenko (commented):
Couple of comments

burqen:

MishaDemianenko (approved):

PR #9232, opened 2017-04-20
Cleanup: Don't use unnecessary default interface implementations

burqen:
In GBPTree and NativeLabelScanStore


tinwelint:
I wonder why you consider Adapters being better than default methods in these cases. I quite like the defaults for this.

chrisvest:
@burqen I think a fix for the flaky test got merged.

tinwelint (approved):
I don't feel strongly about this, I'm OK with the change.

PR #9236, opened 2017-04-21
Defaults to 2G max-memory in importer if unable to detect free machine memory

tinwelint:


burqen (approved):

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -469,7 +470,13 @@ private static Long parseMaxMemory( String maxMemoryString )
...
             {
                 int percent = Integer.parseInt( maxMemoryString.substring( 0, maxMemoryString.length() - 1 ) );
-                return calculateMaxMemoryFromPercent( percent );
+                long result = calculateMaxMemoryFromPercent( percent );
+                if ( !canDetectFreeMemory() )
+                {
+                    System.err.println( "WARNING: amount of free memory couldn't be detected so defaults to " +
+                            bytes( result ) + ". For optimal performance instead explicitly specify amount of " +
+                            "memory that importer is allowed to use using " + Options.MAX_MEMORY.argument() );
+                }
```
burqen, line 479: We should return `result` here, right?

burqen (changes requested):

burqen (approved):

PR #9238, opened 2017-04-21
listQueries displays startime in UTC

praveenag:


thobe (approved):
Looks good.

spacecowboy (approved):

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
```diff
@@ -481,8 +481,8 @@ private QueryStatusResult queryStatusResult( ExecutingQuery q )
         private static String formatTime( final long startTime )
         {
             return OffsetDateTime
-                .ofInstant( Instant.ofEpochMilli( startTime ), ZoneId.systemDefault() )
-                .format( ISO_OFFSET_DATE_TIME );
+                    .ofInstant( Instant.ofEpochMilli( startTime ), ZoneId.of("GMT") )
```
pontusmelke: It doesn't really matter but shouldn't it be `UTC`?

spacecowboy: Hmm, good point

chrisvest: Also, how about lifting the `ZoneId` to a constant? Those instances are immutable and thread-safe.

chrisvest (commented):
Reviewed. I agree with Pontus, and had one additional comment.

PR #9240, opened 2017-04-21
id should return null not -1 

pontusmelke:
In the compiled runtime the `id` function should never return `-1` but instead return `null` whenever it is called with a `null` node or relationship.


systay:
LGTM. Just waiting for green build

fickludd (approved):

PR #9244, opened 2017-04-21
Delete the in-memory counts store code

chrisvest:
This counts store implementation was never completed, and one of its tests were
flaky. The existing counts store implementation has settled down in the mean
time, and now seems pretty stable, so there is no longer a reason to take this
one any further. It's dead code, so away it goes!


burqen (approved):
Good riddance! 🎉

PR #9246, opened 2017-04-21
Add private constructors to utility classes.

MishaDemianenko:
Unfortunately can't introduce checkstyle rule for automatic checks because of
settings classes which should have default or public constructor because of service
loading, while they are utility classes


chrisvest:
clirr is disagreeing.

PR #9247, opened 2017-04-21
OneStatementPerLine checkstyle check.

MishaDemianenko:


chrisvest:
This should probably be the first thing that goes into 3.3, so other PRs can be rebased and avoid breaking the build.

PR #9249, opened 2017-04-22
Parallel I/O in importer

tinwelint:
Previously focus has been on making all I/O single threaded and sequential in the importer. However with more modern hardware it's more and more beneficial to read, and sometimes even write, in parallel. Also work load have been redistributed so that writers do a little less, so basically only writes records to stores.

This series of commits opens up possibility for doing that and adding some, for the time internal, configuration options for enabling/disabling this parallelism.

Queueing and task execution has seen some simplifications and less overhead too so now it's only the first step in a stage of processing which will apply back pressure if the downstream step can't catch up, otherwise all the other step never waits for downstream steps to catch up.


tinwelint:
Will reissue for 3.2 instead

PR #9255, opened 2017-04-24
Change UNWIND to accept non-lists

Mats-SX:
- UNWIND in interpreted runtime now treats `null` as a single-element list


Mats-SX:
This is depending on discussion/decisions in https://github.com/opencypher/openCypher/pull/230 and https://github.com/opencypher/openCypher/issues/234

PR #9256, opened 2017-04-24
Use specific method to perform value comparison in OrderedPropertyValues to avoid wrappers creation.

MishaDemianenko:


MishaDemianenko:
Closing this one since related work will be done by cypher team

PR #9259, opened 2017-04-25
fix long recovery leading to oom in in-flight map

martinfurmanski:
A long recovery during startup could lead to the in-flight map
being filled up with enough entries to exhaust memory. This was
an unnoticed side-effect of a recent refactoring which was
previously protected against by holding the core state monitor
during the entire recovery step.

changelog: Fix OOM during startup.


digitalstain (approved):
Looks quite good to me

martinfurmanski:
@digitalstain I fixed some tests that relied upon the cache being enabled initially.

PR #9261, opened 2017-04-25
refuse_to_be_leader is controlled by multi_dc_license

digitalstain:
The CC setting refuse_to_be_leader is deemed to be a multi dc
 specific feature and as such must be controlled by the multi_dc_license
 switch. The implemented behaviour is that unless the license toggle is
 set to true, the follower pin will always default to false, regardless
 of the actual value.


martinfurmanski (commented):
It's a little bit confusing to call the same thing by different terms; skew, pin, refusal, in different constellations. Similar for license; toggle, multi-dc license.

Do you want a warning if there is a mismatch between license and refusal?

martinfurmanski (approved):

PR #8928, opened 2017-03-02
Flat cursors

davidegrohmann:
Have single "high level" cursors that deals with both transaction state and storage.

PR #9262, opened 2017-04-25
Cleanups around acquisition of cursors

davidegrohmann:
Builds on top of https://github.com/neo4j/neo4j/pull/8928


community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/StoreSingleNodeCursor.java
```diff
@@ -192,7 +192,7 @@ public boolean hasLabel( int labelId )
     @Override
     public boolean isDense()
     {
-        return nodeRecord.isDense();
+        return state != null && state.nodeIsAddedInThisTx( nodeId )  ? false : nodeRecord.isDense();
```
ragadeeshu, line 195: Is a record added in this transaction never dense? Or am I not reading this correctly?

davidegrohmann: Yes if it is not on disk it is not dense.

ragadeeshu (approved):
Looks good

PR #9265, opened 2017-04-26
Refactor node cursor

davidegrohmann:
Builds on https://github.com/neo4j/neo4j/pull/9262


community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/StoreNodeCursor.java
```diff
@@ -96,23 +103,34 @@ private boolean fetchNext()
             {
                 return true;
             }
+
+            if ( state != null && progression.mode() == FETCH && state.nodeIsAddedInThisTx( id ) )
```
ragadeeshu: This isn't a special case :)

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/Progression.java
```diff
@@ -0,0 +1,33 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api.store;
+
+public interface Progression
```
ragadeeshu, line 22: Should this be Nodeprogression

ragadeeshu (changes requested):

PR #9267, opened 2017-04-26
Only do prefix checking for the load balancing config

mneedham:


digitalstain (approved):
Looks good to me. Tried it, it works. I think the prefix validation idea is good and it's going to come in handy going forward.

PR #9269, opened 2017-04-26
Fix bug in reuse of linked page cursors

chrisvest:
Previously, when a linked cursor is closed it is returned to the cursor pool.
It could then be reused by another PagedFile.io() call, while still being considered linked by the parent cursor. Thus, when the parent cursor was closed, it would also close the previously-linked-now-reused cursor.

This bug is fixed by never returning linked cursors to the pool. They now stick around, forever being "linked" to their parent, and then reused as necessary by the parent cursor.


burqen (approved):
Looks good! 🎉

PR #9270, opened 2017-04-27
Accessing maps and lists with null

pontusmelke:
The following should all return `null` instead of failing with a `NullPointerException`

```
list[null]
list[null..5]
list[1..null]
map[null]
```


pontusmelke:
Replaced by https://github.com/neo4j/neo4j/pull/9272

PR #9271, opened 2017-04-27
Close open channels can be cleaned up on windows without any warnings.

MishaDemianenko:
Close tests resources. 
Update StoreLocker to close channel in case of failure.
Properly close lock channel in the case when the store was already locked in the batch inserter.
Try to stop lifecycle component that fails its start to clear all allocated resources. 
Close segments in tests.


MishaDemianenko:
@chrisvest updated

community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
```diff
@@ -1739,51 +1739,52 @@ public void pagesAddedWithNextWithPageIdMustBeAccessibleWithNoGrowSpecified() th
     public void writesOfDifferentUnitsMustHaveCorrectEndianess() throws Exception
     {
         configureStandardPageCache();
-        PagedFile pagedFile = pageCache.map( file( "a" ), 20 );
-
-        try ( PageCursor cursor = pagedFile.io( 0, PF_SHARED_WRITE_LOCK ) )
+        try ( PagedFile pagedFile = pageCache.map( file( "a" ), 20 ) )
```
chrisvest: These two try-with-resources could be combined.

community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
```diff
@@ -2067,81 +2068,87 @@ public void evictionMustFlushPagesToTheRightFiles() throws IOException
...
-        long pageId2 = 0;
-        boolean moreWorkToDo;
-        do
+        try ( PagedFile pagedFile1 = pageCache.map( file( "a" ), filePageSize ) )
         {
-            boolean cursorReady1;
-            boolean cursorReady2;
-
-            try ( PageCursor cursor = pagedFile1.io( pageId1, PF_SHARED_WRITE_LOCK ) )
+            try ( PagedFile pagedFile2 = pageCache.map( file2, filePageSize2 ) )
```
chrisvest: These two can also be combined.

community/dbms/src/main/java/org/neo4j/commandline/dbms/StoreLockChecker.java
```diff
@@ -58,10 +59,24 @@ static Closeable check( Path databaseDirectory ) throws CannotWriteException
...
+                try
+                {
+                    storeLocker.checkLock();
+                    return storeLocker;
+                }
+                catch ( StoreLockException le )
+                {
+                    try
+                    {
+                        storeLocker.close();
```
tinwelint, line 73: We've seen this before from previous attempts to "clean up" around this area, at least three times. We can't seem to do this because then the original lock gets released.

MishaDemianenko: we even recently updated StoreLocker tests suite and they claim that it's not the case but if you can't take a look on them and find where it can be the case that would be great, otherwise those test claims we still have a lock

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCacheTest.java
```diff
@@ -96,32 +97,25 @@ public void mustEvictCleanPageWithoutFlushing() throws Exception
...
-        cursorTracer.reportEvents();
-        assertNotNull( cursorTracer.observe( Fault.class ) );
-        assertEquals( 1, cursorTracer.faults() );
-        assertEquals( 1, tracer.faults() );
-
-        int clockArm = pageCache.evictPages( 1, 0, tracer.beginPageEvictions( 1 ) );
-        assertThat( clockArm, is( 1 ) );
-        assertNotNull( tracer.observe( Evict.class ) );
-    }
+            try ( PagedFile pagedFile = pageCache.map( file( "a" ), 8 ) )
```
chrisvest: As with the `PageCacheTest`, there are many places here where try-with-resources clauses can be combined.

community/dbms/src/main/java/org/neo4j/commandline/dbms/StoreLockChecker.java
chrisvest: Maybe it'd be a good idea to run a custom Windows build.

community/common/src/main/java/org/neo4j/kernel/lifecycle/LifeSupport.java
```diff
@@ -437,6 +437,14 @@ public void start()
...
                 {
                     currentStatus = changedStatus( instance, currentStatus, LifecycleStatus.STOPPED );
+                    try
+                    {
+                        instance.stop();
+                    }
+                    catch ( Throwable se )
+                    {
+                        e.addSuppressed( se );
+                    }
```
chrisvest, line 455: Should we do the same for `init`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/RemoteStore.java
```diff
@@ -103,30 +103,27 @@ private long getPullIndex( File storeDir ) throws IOException
         long lastTxId = BASE_TX_ID;
         try ( Lifespan ignored = new Lifespan( txStore ) )
         {
-            TransactionCursor cursor;
-            try
+            try ( TransactionCursor cursor = txStore.getTransactions( lastCleanTxId ) )
             {
```
chrisvest: I think these two could be combined as well.

chrisvest (commented):
I have a few comments, and TestBackup is failing for some reason. It mostly looks good, though.

MishaDemianenko:
@chrisvest updated, custom windows was there before rebase's it was an official way to confirm that warnings actually were gone; started another one.

community/kernel/src/main/java/org/neo4j/kernel/internal/locker/GlobalStoreLocker.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    protected void releaseLock() throws IOException
+    {
+        lockedFiles.remove( storeLockFile );
+        super.releaseLock();
+    }
```
chrisvest, line 83: Did you consider the design with a concurrent map from file to a reference count? I think it ought to be possible to do that fairly easily with the atomic update functions of the concurrent hash map.

community/kernel/src/test/java/org/neo4j/graphdb/LabelsAcceptanceTest.java
```diff
@@ -86,7 +87,9 @@
 public class LabelsAcceptanceTest
 {
     @Rule
-    public ImpermanentDatabaseRule dbRule = new ImpermanentDatabaseRule();
+    public final ImpermanentDatabaseRule dbRule = new ImpermanentDatabaseRule();
+    @Rule
+    public final TestDirectory testDirectory = TestDirectory.testDirectory();
```
chrisvest, line 92: RuleChain?

community/kernel/src/main/java/org/neo4j/kernel/internal/locker/GlobalStoreLocker.java
chrisvest: This is probably fine in practice, though.

chrisvest (approved):
Review done. Looks good. Just a few questions.

community/kernel/src/test/java/org/neo4j/graphdb/LabelsAcceptanceTest.java
MishaDemianenko: i usually follow rule of three to switching to rule chain :)

chrisvest: I use it when I think the order might matter.

community/kernel/src/main/java/org/neo4j/kernel/internal/locker/GlobalStoreLocker.java
MishaDemianenko: that was the first I did actually, but then switch to a simple set of opened lock files since I think it's a bit simpler and still do the thing

chrisvest: I agree after taking a second look.

community/kernel/src/test/java/org/neo4j/graphdb/LabelsAcceptanceTest.java
MishaDemianenko: that's cause and not even discussable since chain rule should solve exactly that, but it also nice to combine those rules into one that I was referring. 
I do not think that order of them are matter in this test, right?

PR #9272, opened 2017-04-27
Accessing maps and lists with null should return null

pontusmelke:
The following should all return `null` instead of failing with a `NullPointerException`

```
list[null]
list[null..5]
list[1..null]
map[null]
```
Note the docs failure is expected in 2.3

changelog: Fixed error with accessing maps and lists with null, so it now evaluates to null instead

### DIFF
--- a/community/cypher/cypher-compiler-2.3/src/main/scala/org/neo4j/cypher/internal/compiler/v2_3/commands/expressions/ContainerIndex.scala
+++ b/community/cypher/cypher-compiler-2.3/src/main/scala/org/neo4j/cypher/internal/compiler/v2_3/commands/expressions/ContainerIndex.scala
@@ -33,8 +33,9 @@ with CollectionSupport {
   def compute(value: Any, ctx: ExecutionContext)(implicit state: QueryState): Any = {
     value match {
       case IsMap(m) =>
-        val idx = CastSupport.castOrFail[String](index(ctx))
-        m(state.query).getOrElse(idx, null)
+        val item = index(ctx)
+        if (item == null) null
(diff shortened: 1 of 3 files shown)


community/cypher/cypher-compiler-2.3/src/main/scala/org/neo4j/cypher/internal/compiler/v2_3/commands/expressions/ContainerIndex.scala
```diff
@@ -33,18 +33,23 @@ with CollectionSupport {
   def compute(value: Any, ctx: ExecutionContext)(implicit state: QueryState): Any = {
     value match {
       case IsMap(m) =>
-        val idx = CastSupport.castOrFail[String](index(ctx))
-        m(state.query).getOrElse(idx, null)
+        val item = index(ctx)
+        if (item == null) null
+        else m(state.query).getOrElse(CastSupport.castOrFail[String](item), null)
```
Mats-SX: If we try `castOrFail()` first here, we could sometimes skip evaluating the map. Do you think that would matter?

Mats-SX (approved):
Looks good.

community/cypher/cypher-compiler-2.3/src/main/scala/org/neo4j/cypher/internal/compiler/v2_3/commands/expressions/ContainerIndex.scala
pontusmelke: I don't understand, it is `castOrFail` that gives the `NullPointerException`. I thought about adding the null check in `castOrFail` but we use it from a lot of places and not all places have the same `null` semantics.

PR #9273, opened 2017-04-27
Parallel I/O in importer

tinwelint:
Previously focus has been on making all I/O single threaded and sequential in the importer. However with more modern hardware it's more and more beneficial to read, and sometimes even write, in parallel. Also work load have been redistributed so that writers do a little less, so basically only writes records to stores.

This series of commits opens up possibility for doing that and adding some, for the time internal, configuration options for enabling/disabling this parallelism.

Queueing and task execution has seen some simplifications and less overhead too so now it's only the first step in a stage of processing which will apply back pressure if the downstream step can't catch up, otherwise all the other step never waits for downstream steps to catch up.


burqen:
@tinwelint should the config options be listed in the changelog?

tinwelint:
@burqen the thing is that those configuration options aren't actually available to configure. Something we'll have to think about.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/HighestId.java
```diff
@@ -0,0 +1,49 @@
...
+{
+    private final AtomicLong highestId = new AtomicLong();
+
+    public void offer( long candidate )
+    {
+        long currentHighest;
+        do
+        {
+            currentHighest = highestId.get();
+            if ( candidate < currentHighest )
```
burqen: Could use `<=` here

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/PropertyEncoderStep.java
```diff
@@ -19,81 +19,213 @@
...
-import static java.util.Arrays.copyOf;
+import static java.lang.Math.toIntExact;
 
 /**
- * Encodes property data into {@link PropertyBlock property blocks}, attaching them to each
+ * Encodes property data into {@link PropertyRecord property records}, attaching them to each
  * {@link Batch}. This step is designed to handle multiple threads doing the property encoding,
  * since property encoding is potentially the most costly step in this {@link Stage}.
+ * The delivered {@link PropertyRecord property records} all have local ids and so reassignment of those
+ * ids will have to be done later.
```
burqen: Is this true? I thought the reassigning of ids happened in the loop starting at line `126`. And I don't see any other later step doing the reassignment. What is it that I'm missing?

community/kernel/src/test/java/org/neo4j/test/rule/NeoStoresRule.java
```diff
@@ -72,17 +80,34 @@ public NeoStores open( RecordFormats format, String... config ) throws IOExcepti
...
         Config conf = Config.embeddedDefaults( stringMap( config ) );
         pageCache = getOrCreatePageCache( conf, efs );
-        return open( efs, pageCache, format, config );
+        return open( efs, pageCache, format, fs -> new DefaultIdGeneratorFactory( fs ), config );
     }
 
-    public NeoStores open( FileSystemAbstraction fs, PageCache pageCache, RecordFormats format, String... config )
-            throws IOException
+    public NeoStores open( FileSystemAbstraction fs, PageCache pageCache, RecordFormats format,
+            Function<FileSystemAbstraction,IdGeneratorFactory> idGeneratorFactory, String... config )
```
burqen: Class level javadoc reference this method. Please update signature there as well.

community/kernel/src/test/java/org/neo4j/test/rule/NeoStoresRule.java
```diff
@@ -113,4 +131,48 @@ private static PageCache getOrCreatePageCache( Config config, FileSystemAbstract
                 PageCursorTracerSupplier.NULL, log );
         return pageCacheFactory.getOrCreatePageCache();
     }
+
+    public class Builder
```
burqen, line 112: Right now we use both builder and multiple constructors to create `NeoStores`.
I would like it if we could:
1. use this pattern everywhere and remove the multiple `open(...)`
2. document the "default" behaviour of the builder if the fields are not set explicitly.

Did not look lite _too_ much work to move from `neoStoresRule.open()` -> `neoStoreRules.builder().build()` for the majority of the cases we have.

community/kernel/src/test/java/org/neo4j/test/rule/NeoStoresRule.java
```diff
@@ -72,17 +80,34 @@ public NeoStores open( RecordFormats format, String... config ) throws IOExcepti
...
     }
 
-    public NeoStores open( FileSystemAbstraction fs, PageCache pageCache, RecordFormats format, String... config )
-            throws IOException
+    public NeoStores open( FileSystemAbstraction fs, PageCache pageCache, RecordFormats format,
+            Function<FileSystemAbstraction,IdGeneratorFactory> idGeneratorFactory, String... config )
+                    throws IOException
     {
         assert neoStores == null : "Already opened";
+        if ( fs == null )
```
burqen: Can we move this "default" behaviour to the builder instead and also have it documented there. This ties together with comment about removing multiple constructor pattern and completely replace with builder.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/validation/IdValidator.java
```diff
@@ -88,4 +88,10 @@ public static void assertIdWithinCapacity( long id, long maxId )
             throw new IdCapacityExceededException( id, maxId );
         }
     }
+
+    public static boolean hasReservedIdInRange( long startIdInclusive, long endIdExclusive )
```
burqen, line 92: From what I understand, this means there will be a batch sized gap of ids around `INTEGER_MINUS_ONE` that is not used by importer. Right? But id generator will know this and fill this gap up during normal operations after import is done?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ReadRecordsStep.java
```diff
@@ -39,74 +37,62 @@
...
-    {
-        super.close();
-        cursor.close();
-    }
 
-    @Override
-    protected long position()
-    {
-        return count * recordSize;
+        sender.send( i == batchSize ? batch : Arrays.copyOf( batch, i ) );
```
burqen, line 96: The reason we do this is because we need the sent array to be completely full, right?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
```diff
@@ -43,7 +43,7 @@
      * database directory of the imported database, i.e. <into>/bad.log.
      */
     String BAD_FILE_NAME = "bad.log";
-    long MAX_PAGE_CACHE_MEMORY = mebiBytes( 240 );
+    long MAX_PAGE_CACHE_MEMORY = mebiBytes( 480 );
```
burqen, line 46: This number seem to be fairly low, even for systems with memory size only in the tens of GB. Why is it so low and not adaptive to the environment like the page cache normally is?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/NodeFirstRelationshipStage.java
```diff
@@ -41,13 +44,13 @@
...
+            NodeRelationshipCache cache )
     {
-        super( "Node --> Relationship" + topic, config );
-        add( new ReadNodeRecordsByCacheStep( control(), config, nodeStore, cache, nodeTypes ) );
+        super( "Node --> Relationship" + topic, config, ORDER_SEND_DOWNSTREAM );
+        add( new ReadNodeIdsByCacheStep( control(), config, cache, NodeType.NODE_TYPE_SPARSE ) );
+        add( new ReadRecordsStep<>( control(), config, true, nodeStore ) );
         add( new RecordProcessorStep<>( control(), "LINK", config,
-                new NodeFirstRelationshipProcessor( relationshipGroupStore, cache ), false ) );
+                new NodeFirstRelationshipProcessor( cache ), false ) );
```
burqen: `NodeFirstRelationshipProcessor` does not create group records for dense nodes any more. Update this in javadoc please.

(On that note. The javadocs for the different stages has helped me a lot. Good stuff!)

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/NodeFirstRelationshipProcessor.java
```diff
@@ -33,52 +31,28 @@
  * This step also creates {@link RelationshipGroupRecord group records} for the dense nodes as it encounters
  * dense nodes, where it gets all relationship group information from {@link NodeRelationshipCache}.
  */
-public class NodeFirstRelationshipProcessor implements RecordProcessor<NodeRecord>, GroupVisitor
+public class NodeFirstRelationshipProcessor implements RecordProcessor<NodeRecord>
```
burqen: Please update javadoc. This guy does nothing with dense nodes anymore.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/RelationshipGroupStage.java
```diff
@@ -0,0 +1,36 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.unsafe.impl.batchimport;
+
+import org.neo4j.kernel.impl.store.RecordStore;
+import org.neo4j.kernel.impl.store.record.RelationshipGroupRecord;
+import org.neo4j.unsafe.impl.batchimport.cache.NodeRelationshipCache;
+import org.neo4j.unsafe.impl.batchimport.staging.Stage;
+
+public class RelationshipGroupStage extends Stage
```
burqen: Can we have some awesome javadoc for this guy?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/Step.java
```diff
@@ -93,4 +93,6 @@
      */
     @Override
     void close() throws Exception;
+
+    long doneBatches();
```
burqen: This method is not called anywhere. Can we get rid of it?
And if I'm wrong, can we haz javadoc?

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/staging/ProcessorStepTest.java
```diff
@@ -92,51 +101,21 @@ public void shouldHaveTaskQueueSizeEqualToNumberOfProcessorsIfSpecificallySet()
...
         {
-            super( control, "test", Configuration.DEFAULT, maxProcessors );
+            super( control, "test", configuration, maxProcessors );
             this.latch = latch;
         }
 
         @Override
         protected void process( Void batch, BatchSender sender ) throws Throwable
         {
+            System.out.println( Thread.currentThread().getName() + " process" );
```
burqen: Oops.

burqen (changes requested):
Overall changes look good. I have some comments, mostly about readability and explanation about thing I did not fullt understand. 🎉

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ReadRecordsStep.java
```diff
@@ -39,74 +37,55 @@
...
-    private volatile long count;
+    protected final int batchSize;
 
     @SuppressWarnings( "unchecked" )
-    public ReadRecordsStep( StageControl control, Configuration config, RecordStore<RECORD> store,
-            RecordIdIterator ids )
+    public ReadRecordsStep( StageControl control, Configuration config, RecordStore<RECORD> store )
     {
-        super( control, config );
+        super( control, ">", config, 0 );
```
ragadeeshu: These names (">") are really non-descriptive.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
ragadeeshu: IIRC this is because the cache is only used for writes, and thus there is no benefit of keeping the data in memory.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/DynamicProcessorAssigner.java
```diff
@@ -65,23 +65,20 @@ public void start( StageExecution execution )
     @Override
     public void check( StageExecution execution )
     {
-        int permits = availableProcessors - countActiveProcessors( execution );
         if ( execution.stillExecuting() )
         {
+            int permits = availableProcessors - countActiveProcessors( execution );
             if ( permits > 0 )
```
ragadeeshu, line 71: Where is permits changed now?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
burqen: But we do read from store in quite a lot of our `Stages`, every time we use `ReadRecordsStep`.Would it not be beneficial to not have those reads result in I/O?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/DynamicProcessorAssigner.java
burqen: Seems like it's not changed, but recalculated every time we call `DynamicProcessorAssigner.check( StageExecution execution )`

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/PropertyEncoderStep.java
tinwelint: You're right, this was changed back and forth and ended up back in here :)

community/kernel/src/test/java/org/neo4j/test/rule/NeoStoresRule.java
tinwelint: Sure, there are so few places so this can be done I think

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/validation/IdValidator.java
tinwelint: No, these gaps will not be in the freelist. Doesn't matter much though as it's in the size of thousands of ids.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ReadRecordsStep.java
tinwelint: Yeah, and only the last batch will do this copy

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
tinwelint: What we generally see is that the smaller the page cache is the faster the importer goes, down to a certain point of course. The smaller it is the higher the chance the background flushing will hit a nice wave of sequential I/O. I don't think this change do change much, it was introduced when seeing that I/O went faster and so we wanted the page cache to have a little more wiggle room. Perhaps it can be reverted even.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/Step.java
tinwelint: Accidentally left it there

tinwelint:
@burqen comments addressed

burqen (approved):

PR #9284, opened 2017-04-29
Improvements for rel-type checking for traversing relationships

jexp:
* removed rel-type check for rel-records of dense nodes as each chain has a single rel-type
* added single-rel-type check instead of array iteration for single rel-types
* removed varargs for rel-type-id array from Kernel-API and below
* added dedicated method to ReadOperations if you only have a single rel-type, so no arrays are allocated on the way

I didn't change the tx-state handling for single rel-types, it's use of boxed values and non-primitive functions and collections should also be revisited.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/cursor/TxAbstractNodeCursor.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/cursor/TxAbstractNodeCursor.java
@@ -126,12 +126,20 @@ public abstract class TxAbstractNodeCursor
     }
 
     @Override
-    public Cursor<RelationshipItem> relationships( Direction direction, int... relTypes )
+    public Cursor<RelationshipItem> relationships( Direction direction, int[] relTypes )
     {
         return state.augmentNodeRelationshipCursor(
                 nodeIsAddedInThisTx ? Cursors.<RelationshipItem>empty() : cursor.get().relationships( direction,
                         relTypes ), nodeState,
                 direction, relTypes );
+    }
+
+    @Override
+    public Cursor<RelationshipItem> relationships(Direction direction, int relType)
+    {
+        return state.augmentNodeRelationshipCursor(
+                nodeIsAddedInThisTx ? Cursors.<RelationshipItem>empty() : cursor.get().relationships(direction, relType), nodeState,
+                direction, new int[]{relType} );
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


jexp:
@tinwelint would love to hear your opinion. This PR is mainly for discussion and benchmarking.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/cursor/TxAbstractNodeCursor.java
```diff
@@ -135,6 +135,14 @@ public long id()
     }
 
     @Override
+    public Cursor<RelationshipItem> relationships(Direction direction, int relType)
+    {
+        return state.augmentNodeRelationshipCursor(
+                nodeIsAddedInThisTx ? Cursors.<RelationshipItem>empty() : cursor.get().relationships(direction, relType), nodeState,
+                direction, new int[]{relType} );
```
tinwelint, line 142: An annoying new int[] for sure, even knowing that other creations have been avoided by avoiding varargs.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -163,10 +163,72 @@ public void delete()
...
+                    return result;
+                }
+                catch ( EntityNotFoundException e )
+                {
+                    statement.close();
+                    throw new NotFoundException( format( "Node %d not found", nodeId ), e );
+                }
+                catch ( Throwable e )
+                {
+                    statement.close();
```
tinwelint, line 227: I know it's a hacked thing, but a `finally` instead, right?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/StoreNodeRelationshipCursor.java
```diff
@@ -138,7 +163,7 @@ public boolean next()
                     }
 
                     // Type check
-                    if ( !checkType( relationshipRecord.getType() ) )
+                    if ( !isDense && !checkType( relationshipRecord.getType() ) )
```
tinwelint, line 166: Would be great with an execution path without even the condition for dense nodes.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -163,10 +163,72 @@ public void delete()
...
+        @Override
+        public boolean hasNext()
+        {
+            return false;
+        }
+
+        @Override
+        public Relationship next()
+        {
+            return null;
```
tinwelint, line 182: throw `NoSuchElementException` I suppose

tinwelint (commented):
All in all I think these changes highlight that there are nice gains to be had. I'd like the way forward to have goals such as reducing conditions/branches (mentioned in one of the comments).

systay:
Is this PR still headed for merging, or can we close it?

systay:

PR #9287, opened 2017-05-02
Make max auth max failed attempts configurable

lutovich:
This PR adds new internal configuration option `unsupported.dbms.security.auth_max_failed_attempts` to configure failed auth rate limiting. Previously it was fixed to 3 failed attempts before imposing a "cooldown". Ability to configure this parameter is especially valuable for tests that component behaviour when auth failed.

### DIFF
--- a/community/security/src/test/java/org/neo4j/server/security/auth/RateLimitedAuthenticationStrategyTest.java
+++ b/community/security/src/test/java/org/neo4j/server/security/auth/RateLimitedAuthenticationStrategyTest.java
@@ -21,6 +21,7 @@ package org.neo4j.server.security.auth;
 
 import org.junit.Test;
 
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 
 import org.neo4j.kernel.api.security.AuthenticationResult;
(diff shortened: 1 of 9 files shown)


community/security/src/test/java/org/neo4j/server/security/auth/RateLimitedAuthenticationStrategyTest.java
```diff
@@ -120,6 +122,31 @@ public void shouldSlowRequestRateOnMultipleFailedAttemptsWhereAttemptIsValid() t
...
+    private void testUnlimitedFailedAuthAttempts( int maxFailedAttempts )
+    {
+        FakeClock clock = getFakeClock();
+        AuthenticationStrategy authStrategy = new RateLimitedAuthenticationStrategy( clock, maxFailedAttempts );
+        User user = new User.Builder( "user", Credential.forPassword( "right" ) ).build();
+
+        int attempts = ThreadLocalRandom.current().nextInt( 5, 100 );
+        for ( int i = 0; i < attempts; i++ )
+        {
+            assertEquals( AuthenticationResult.TOO_MANY_ATTEMPTS, authStrategy.authenticate( user, "wrong" ) );
```
OliviaYtterbrink: I don't understand the idea behind this test/name. The unlimited part seems to me to indicate that we never should get `TOO_MANY_ATTMPTS` but the test is written to assert that it always says that from the first try...

lutovich: @OliviaYtterbrink yeah, I messed up with conditions in `RateLimitedAuthenticationStrategy` and this test was wrong. Should be fixed now.

OliviaYtterbrink (approved):

PR #9288, opened 2017-05-02
User that needs to change password should only be able to do that

OliviaYtterbrink:
Fixes the fact that a user that required password to be changed could run multiple different dbms procedures.


OliviaYtterbrink:
Replaced with https://github.com/neo4j/neo4j/pull/9296

PR #9296, opened 2017-05-03
When credentials have expired you should only be able to change password

OliviaYtterbrink:
changelog: [3.1] Fixed so a user whose credentials have expired aren't allowed to call procedures apart from `changePassword` and `showCurrentUser`


eebus (approved):
Looks good

PR #9304, opened 2017-05-04
Improved StoreLocker's handling of multiple calls to checkLock

klaren:
Prevent multiple channels to be opened on consecutive calls.


MishaDemianenko (commented):
Please rebase it on top of latest 3.3 and fix checkstyle error, otherwise looks good

PR #9307, opened 2017-05-05
Legacy code cleanup 2.0, 2.1, 2.2 

burqen:
We do not support upgrade directly from 2.0, 2.1 or 2.2 to 3.2.
This PR remove legacy code for those migrations.

co-author: @MishaDemianenko


community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/UpgradableDatabase.java
```diff
@@ -164,28 +157,12 @@ private Result checkCleanShutDownByCheckPoint( File storeDirectory )
         return new Result( Result.Outcome.storeNotCleanlyShutDown, null, null );
     }
 
-    private Result checkCleanShutDownByVersionTrailer( File storeDirectory, RecordFormats fromFormat )
```
ragadeeshu, line 166: Nice to see this die for real!

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
```diff
@@ -149,33 +140,29 @@
...
     }
 
     @Override
     public void migrate( File storeDir, File migrationDir, MigrationProgressMonitor.Section progressMonitor,
             String versionToMigrateFrom, String versionToMigrateTo ) throws IOException
     {
-        if ( versionToMigrateFrom.equals( StandardV2_0.STORE_VERSION ) ||
-             versionToMigrateFrom.equals( StandardV2_1.STORE_VERSION ) ||
-             versionToMigrateFrom.equals( StandardV2_2.STORE_VERSION ) )
+        if ( versionToMigrateFrom.equals( StandardV2_3.STORE_VERSION ) )
```
ragadeeshu, line 164: We didn't check for 2.3 before, was that a bug?

community/neo4j/src/test/java/upgrade/StoreUpgraderTest.java
```diff
@@ -146,9 +128,6 @@ public StoreUpgraderTest( String version )
     public static Collection<String> versions()
     {
         return Arrays.asList(
-                StandardV2_0.STORE_VERSION,
```
ragadeeshu, line 149: In a lot of tests the old formats are removed, but should newer formats be added as well?

ragadeeshu: If not, should a lot of these tests disappear completely?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -64,17 +66,38 @@ public NeoStoreFileListing( File storeDir, LabelScanStore labelScanStore, Indexi
...
         Resource schemaIndexSnapshots = gatherSchemaIndexFiles( files );
         Resource legacyIndexSnapshots = gatherLegacyIndexFiles( files );
 
+        placeMetaDataStoreLast( files );
+
         return resourceIterator( files.iterator(),
                 new MultiResource( asList( labelScanStoreSnapshot, schemaIndexSnapshots, legacyIndexSnapshots ) ) );
     }
 
+    private void placeMetaDataStoreLast( List<StoreFileMetadata> files )
```
ragadeeshu, line 82: It is a bit unclear to me why we are doing this :)

integrationtests/src/test/java/org/neo4j/storeupgrade/StoreUpgradeIntegrationTest.java
```diff
@@ -265,10 +265,10 @@ public void migratingOlderDataAndThanStartAClusterUsingTheNewerDataShouldWork()
         @Parameterized.Parameters( name = "{0}" )
         public static Collection<String[]> parameters()
         {
-            return Arrays.<String[]>asList(
+            return Arrays.asList(
```
ragadeeshu, line 270: Hmm, these look very old, why are they still here?

ragadeeshu (commented):
Some questions, mostly around old tests. Looks good otherwise, nice to get this out of the way :)

ragadeeshu:
@burqen reviewed

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
burqen: Yes I think so. But I'd have to bring in @chrisvest to confirm.

community/neo4j/src/test/java/upgrade/StoreUpgraderTest.java
burqen: Good point! I think there is value in keeping those tests and extend them to all relevant formats. Let't keep that outside of this PR though.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
burqen: It's used by `NeoStores.isStorePresent` to indicate that the store is present. Agreeably not very reliable.

integrationtests/src/test/java/org/neo4j/storeupgrade/StoreUpgradeIntegrationTest.java
burqen: You mean the `.tar`s with the stores? They are still there.

burqen:
@ragadeeshu Rebased and ready

integrationtests/src/test/java/org/neo4j/storeupgrade/StoreUpgradeIntegrationTest.java
ragadeeshu: Yes, but why? Didn't we just remove support for migrating them? So why are they still a part of this test? :)

burqen: Aah, this test makes sure that we fail if we try to migrate any of those stores.

ragadeeshu:
@burqen I think that this one needs another rebase.

ragadeeshu:
@burqen And even more rebase

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -389,7 +387,11 @@ private void dropStrict() throws IOException
             index.close();
             index = null;
         }
-        storeFileHandle().delete();
+        Optional<FileHandle> fileHandle = storeFileHandle();
+        if ( fileHandle.isPresent() )
```
tinwelint, line 391: Sounds like `dropStrict` should delete or throw, not be OK with file not being there.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StoreType.java
```diff
@@ -253,4 +259,14 @@ public static boolean shouldBeManagedByPageCache( String storeFileName )
         boolean isLabelScanStore = NativeLabelScanStore.FILE_NAME.equals( storeFileName );
         return isLabelScanStore || StoreType.typeOf( storeFileName ).map( StoreType::isRecordStore ).orElse( false );
     }
+
+    protected boolean isStoreFile( String fileName )
+    {
+        return matchStoreName( fileName, getStoreName() );
+    }
+
+    protected boolean matchStoreName( String fileName, String storeName )
```
tinwelint: Javadoc on what the arguments are would be welcome

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
tinwelint: Yup I think that's for another piece of investigation, not cleanup

tinwelint (changes requested):
Nice to see so much legacy code deleted.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
burqen: It will throw `NoSuchFileException` if file does not exist, this will be caught in `drop`.

burqen:
@ragadeeshu @tinwelint I think I've addressed all your comments now.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
tinwelint: Won't it just return silently if `!fileHandle.isPresent()`?

tinwelint (approved):

PR #9309, opened 2017-05-05
Asynchronous crash pointer cleanup

burqen:
### WHAT?
Do asynchronous cleanup of crash pointers in `GBPTree` if state of tree is dirty on startup.
Before this cleanup was blocking and done as part of `GBPTree` construction.

### WHY?
When `GBPTree` was only used for `NativeLabelScanStore` having a blocking cleanup procedure was fine because it was only performed on a single tree instance.
Moving forward the `GBPTree` will be used to implement native schema number index and thus there can be several tree instances in store. Up to multiple thousands has been seen in the wild. Having a blocking cleanup procedure would not scale.

### HOW?
- Register cleanup jobs to a `RecoveryCleanupWorkCollector` that will be triggered on `NeoStoreDataSource`s `life.start()`.
- Block checkpointing until cleanup is finished. Checkpoint would make it impossible to later recognise crash pointers that has not yet been cleaned up.
- Block close until cleanup is finished. Because closing the `GBPTree` would also mean closing the underlying `PagedFile` which is pinned by crash pointer cleaner.

Builds on #9231 
First new commit is `Shuffle recovery test load order ...`


community/neo4j/src/test/java/recovery/RecoveryCleanupIT.java
```diff
@@ -0,0 +1,194 @@
...
+public class RecoveryCleanupIT
+{
+    @Rule
+    public TestDirectory testDirectory = TestDirectory.testDirectory();
+
+    private TestGraphDatabaseFactory factory = new TestGraphDatabaseFactory();
+    private Barrier.Control recoveryCompleteBarrier;
+    private GraphDatabaseService db;
+    private File storeDir;
+    private ExecutorService executor = Executors.newFixedThreadPool( 2 );
```
MishaDemianenko, line 65: executors are not shutdown after the test

community/neo4j/src/test/java/recovery/RecoveryCleanupIT.java
```diff
@@ -0,0 +1,194 @@
...
+        LabelScanStore.Monitor recoveryBarrierMonitor = new RecoveryBarrierMonitor( recoveryCompleteBarrier );
+        monitors.addMonitorListener( recoveryBarrierMonitor );
+        factory.setMonitors( monitors );
+    }
+
+    @Test
+    public void recoveryCleanupShouldBlockCheckpoint() throws Throwable
+    {
+        // GIVEN
+        AtomicReference<Throwable> error = new AtomicReference<>();
```
MishaDemianenko, line 91: not quite sure why do we need AtomicReference here to rethrow exception since future.get will do that

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -236,13 +238,7 @@ default void startupState( boolean clean )
...
      * For this reason both writer and check-pointing acquires this lock.
      */
-    private final Lock writerCheckpointMutex = new ReentrantLock();
-
-    /**
-     * Currently an index only supports one concurrent writer and so this boolean will act as
-     * guard so that only one thread can have it at any given time.
-     */
-    private final AtomicBoolean writerTaken = new AtomicBoolean();
+    private final ReadWriteLock writerCheckpointMutex = new StampedLock().asReadWriteLock();
```
MishaDemianenko: Is there is any reason why you choose to use stamped lock here? (you do not want reentrancy or smth else?)

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -308,6 +304,21 @@ default void startupState( boolean clean )
...
+
+    /**
+     * Clean jobs part of recovery is posted here.
+     */
+    private final RecoveryCleanupWorkCollector recoveryCleanupWorkCollector;
+
+    /**
+     * If clean job fails, this throwable is set to the cause.
+     */
+    private volatile Throwable failedRecoveryCause;
```
MishaDemianenko: I think it should not be part of GBPTree, if we already have an abstraction that collects and execute recovery tasks then it should be responsible for providing its status, there is no reason to inject status variables inside of jobs

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -972,12 +1001,58 @@ public String toString()
...
+            }
+            catch ( TreeInconsistencyException e )
+            {
+                throw appendTreeInformation( e );
+            }
+            finally
+            {
+                if ( !success )
+                {
+                    closeCursor();
```
MishaDemianenko: why not just call close here?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GroupingRecoveryCleanupWorkCollector.java
```diff
@@ -0,0 +1,52 @@
...
+package org.neo4j.index.internal.gbptree;
+
+import java.util.LinkedList;
+import java.util.Queue;
+
+/**
+ * Collects recovery cleanup work to be performed and {@link #run() runs} them all as one job,
+ * each cleanup job sequentially one after the other.
+ */
+public class GroupingRecoveryCleanupWorkCollector implements RecoveryCleanupWorkCollector
```
MishaDemianenko: Can you please clarify how this should be used in relation to jobs that will be added from other threads and in relation to lifecycle restarts?
What is the use of started variable in that context?

MishaDemianenko (commented):
I have couple of comment-questions mostly, otherwise first review done

community/neo4j/src/test/java/recovery/RecoveryCleanupIT.java
burqen: It's because we can not have a throwing function inside the lambda.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: Because the thread that take lock in `cleanCrashedPointers` is not the same thread that will run the clean job and release the lock in the end.

burqen:

community/neo4j/src/test/java/recovery/RecoveryCleanupIT.java
MishaDemianenko:
```
Future<?> checkpointFuture = executor.submit( (Callable<Void>) () -> {
                checkpoint( db );
                return null;
            } );
```
for example will allow you to get future with checkpoint inside :) but this is also fine it can be like it is now

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
MishaDemianenko: now quite sure, I understand what you mean. Let me be more specific: does this lock needs to be StampedLock or it can be `ReentrantReadWriteLock` as well?

MishaDemianenko:
@burqen couple of comments to my previous comments and as we just talk - lifecycle of `GroupingRecoveryCleanupWorkCollector` still looks like a problem

burqen:
@MishaDemianenko How do you like this new lifecycle?

MishaDemianenko (approved):

MishaDemianenko:
@burqen will merge on green

burqen:
@MishaDemianenko Updated, please note that there are a few new commits addressing comments from @tinwelint (that he left in PR #9362 )

PR #9311, opened 2017-05-05
Create 3.3 Cypher modules

pontusmelke:
- The docs build failure will be resolved once https://github.com/neo4j/neo4j-documentation/pull/79 is in (confirmed locally). Will merge them at the same time, so for now just ignore that build.
- Note that I opted for leaving the dependency on 3.1 in order to handle `CREATE UNIQUE` requests, if we later choose to do this differently we can remove it then (in a later PR, please!)


community/pom.xml
```diff
@@ -131,6 +131,7 @@ the relevant Commercial Agreement.
                     <exclude>org.neo4j:neo4j-cypher-frontend-3.2</exclude>
                     <exclude>org.neo4j:neo4j-cypher-frontend-3.1</exclude>
                     <exclude>org.neo4j:neo4j-cypher-frontend-2.3</exclude>
+                    <exclude>org.neo4j:neo4j-cypher-compiler-3.2</exclude>
```
Mats-SX: Also exclude ir module here I think

PR #9317, opened 2017-05-08
Avoid accumulation of index updates during recovery

MishaDemianenko:
Do not collect index updates during recovery since that can cause OOM.
Use iterators instead to iterate over required updates.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexStoreView.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexStoreView.java
@@ -59,9 +59,9 @@ public interface IndexStoreView extends PropertyAccessor, PropertyLoader
      * and puts those updates into {@code target}.
      *
      * @param nodeId id of node to load.
-     * @param target {@link Collection} to add updates into.
+     * @return node updates container
      */
-    void nodeAsUpdates( long nodeId, Collection<NodeUpdates> target );
+    NodeUpdates nodeAsUpdates( long nodeId );
 
     DoubleLongRegister indexUpdatesAndSize( long indexId, DoubleLongRegister output );
 
@@ -129,8 +129,9 @@ public interface IndexStoreView extends PropertyAccessor, PropertyLoader
         }
 
         @Override
-        public void nodeAsUpdates( long nodeId, Collection<NodeUpdates> target )
+        public NodeUpdates nodeAsUpdates( long nodeId )
         {
+            return null;
         }
 
         @Override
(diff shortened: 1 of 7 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -545,28 +547,29 @@ private void applyRecoveredUpdates() throws IOException, IndexEntryConflictExcep
...
+                        Iterators.flatMap( nodeUpdates -> convertToIndexUpdates( nodeUpdates ).iterator(), updates );
+                while ( indexEntryUpdates.hasNext() )
+                {
+                    IndexEntryUpdate<LabelSchemaDescriptor> indexUpdate = indexEntryUpdates.next();
+                    monitor.applyingRecoveredUpdate( indexUpdate );
+                    IndexUpdater updater = updaterMap.getUpdater( indexUpdate.indexKey().schema() );
+                    if ( updater != null )
+                    {
+                        updater.process( indexUpdate );
+                    }
```
burqen: This code is also used in `apply( Iterable<IndexEntryUpdate<LabelSchemaDescriptor>> updates, IndexUpdateMode updateMode )`. Can we extract it?
```
IndexUpdater updater = updaterMap.getUpdater( indexUpdate.indexKey().schema() );
if ( updater != null )
{
    updater.process( indexUpdate );
}
```

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexStoreView.java
```diff
@@ -59,9 +59,9 @@
      * and puts those updates into {@code target}.
      *
      * @param nodeId id of node to load.
-     * @param target {@link Collection} to add updates into.
+     * @return node updates container
      */
-    void nodeAsUpdates( long nodeId, Collection<NodeUpdates> target );
+    NodeUpdates nodeAsUpdates( long nodeId );
```
burqen, line 64: Can we also update rest of javadoc to reflect this change?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -545,28 +547,29 @@ private void applyRecoveredUpdates() throws IOException, IndexEntryConflictExcep
                     updater.remove( recoveredNodeIds );
```
burqen, line 532: I don't understand why we need to do this. Seems like when we are in `IndexUpdateMode.RECOVERY` we will use `SwallowingIndexUpdater` which is just no-ops.

burqen (changes requested):
Behaviour changes look good! 🎉  
Got some small comments about readability.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
MishaDemianenko: no, not always; for example, we can use LuceneIndexUpdater. Also, this is not related to PR.

MishaDemianenko:
@burqen I believe I have cover all your comments

MishaDemianenko:
@burqen updated

burqen (approved):

PR #9318, opened 2017-05-08
Debug output now includes total size of store and mapped files

klaren:
The debug.log file now has a summary of the storage:

```
  INFO [o.n.k.i.DiagnosticsManager] Storage summary:
  INFO [o.n.k.i.DiagnosticsManager]   Total size of store: <total> kB
  INFO [o.n.k.i.DiagnosticsManager]   Total size of mapped files: <mapped> kB
```

Where `<total>` is equal the size of the storage directory and `<mapped>`
is the total size of files that are mapped through the page cache.


community/kernel/src/main/java/org/neo4j/kernel/impl/store/MappedFileNameFilter.java
```diff
@@ -0,0 +1,61 @@
...
+import java.io.FilenameFilter;
+
+/**
+ * Filter used to determine what files are map. Helps performance tuning of page cache memory configuration.
+ */
+public class MappedFileNameFilter implements FilenameFilter
+{
+    @Override
+    public boolean accept( File dir, String name )
+    {
```
chrisvest: You can use `StoreType.shouldBeManagedByPageCache()` instead of this.

dgordon5483:
Upon first inspection, this looks like it is displaying the entire store size, which is a little different than the requested feature in one major way: We would prefer this omits the transaction log files, and the indexes. Reporting indexes and transaction log sizes separate would be most excellent.

klaren:
@dgordon5483 The request was to help out with

> correctly configured the pagecache so as to fit entirely in memory or not

and this is exactly what is printed in the "size of mapped file"; the total size of the pagecache.

PR #9320, opened 2017-05-08
Avoid duplicated monitors for label scan store

MishaDemianenko:
Use label scan store tags to separate monitors for native and lucene label scan indexes.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
@@ -96,10 +96,9 @@ public class NativeLabelScanStore implements LabelScanStore
     private static final byte REBUILDING = (byte) 0x01;
 
     /**
-     * Native index monitor tag, to distinguish monitors for native index store
-     * from others.
+     * Native label index tag, to distinguish native label index from other label indexes
      */
-    public static final String NATIVE_INDEX_TAG = GraphDatabaseSettings.LabelIndex.NATIVE.name();
+    public static final String NATIVE_LABEL_INDEX_TAG = GraphDatabaseSettings.LabelIndex.NATIVE.name();
 
     /**
      * Whether or not this label scan store is read-only.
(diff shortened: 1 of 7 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -95,6 +96,12 @@
     private static final byte REBUILDING = (byte) 0x01;
 
     /**
+     * Native index monitor tag, to distinguish monitors for native index store
+     * from others.
+     */
+    public static final String NATIVE_INDEX_TAG = GraphDatabaseSettings.LabelIndex.NATIVE.name();
```
burqen: Can we be a little bit more explicit about what we call this tag to not confuse it with "native schema index"? My suggestion is something like: `NATIVE_LABEL_INDEX_TAG`.
Let this also be reflected in javadoc:
- "Native index monitor tag" -> "Native label index monitor tag"
- "monitors for native index store" -> "monitors for native label index"

community/neo4j/src/test/java/org/neo4j/index/LabelScanStoreLogging.java
```diff
@@ -0,0 +1,76 @@
...
+import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.kernel.api.labelscan.LabelScanWriter;
+import org.neo4j.kernel.api.labelscan.NodeLabelUpdate;
+import org.neo4j.kernel.impl.index.labelscan.NativeLabelScanStore;
+import org.neo4j.kernel.internal.GraphDatabaseAPI;
+import org.neo4j.logging.AssertableLogProvider;
+import org.neo4j.test.TestGraphDatabaseFactory;
+import org.neo4j.test.rule.TestDirectory;
+
+public class LabelScanStoreLogging
```
burqen: Do we not need the class name to include "Test" to make maven run it?

burqen (changes requested):
Changes look good for the most part, but I would like it if we could use the same static variable for tagging monitors as we do to give the providers name.
My suggestion is to move `NAME` from extension class (both Lucene and Native) to their respective `LabelScanStore` implementation and then always use that one.
Makes sense?

MishaDemianenko:
@burqen I was thinking about this PR being more example and showcase but if you think it's fine and it can be merged in - will brush it up!

MishaDemianenko:
@burqen updated!

MishaDemianenko:
@burqen updated just in case as well

burqen (approved):

PR #9321, opened 2017-05-08
Cleanup and clarify usage of empty iterable and empty set

MishaDemianenko:
Use default available emptySet, use default empty iterator,
clarify usage of empty iterator and empty resource iterator.


klaren (approved):
Looks good

PR #9322, opened 2017-05-09
Always close cursors in compiled runtime

pontusmelke:
Whenever we use `CursorRelationshipIterator` in the compiled runtime
we should always make sure to close it afterwards.

This should go in after https://github.com/neo4j/neo4j/pull/9217

### DIFF
--- a/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/codegen/GeneratedMethodStructure.scala
+++ b/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/codegen/GeneratedMethodStructure.scala
@@ -111,8 +111,16 @@ class GeneratedMethodStructure(val fields: Fields, val generator: CodeBlock, aux
   override def nextNode(targetVar: String, iterVar: String) =
     generator.assign(typeRef[Long], targetVar, invoke(generator.load(iterVar), nextLong))
 
-  override def createRelExtractor(relVar: String) =
+  override def setUpRelIteration(relVar: String, iterVar: String) = {
     generator.assign(typeRef[RelationshipDataExtractor], relExtractor(relVar), newRelationshipDataExtractor)
+
+    val field = generator.classGenerator().field(typeRef[RelationshipIterator], iterVar)
+    //Make sure cursor is always closed
+    _finalizers.append((_: Boolean) => (block) =>
+      using(block.ifStatement(Expression.notNull(Expression.get(block.self(), field)))) {inner =>
+        inner.expression(invoke(Expression.get(inner.self(), field), method[RelationshipIterator, Unit]("close")))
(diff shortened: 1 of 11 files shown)


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/codegen/GeneratedMethodStructure.scala
```diff
@@ -425,6 +425,12 @@ class GeneratedMethodStructure(val fields: Fields, val generator: CodeBlock, aux
 
   override def nodeGetRelationshipsWithDirection(iterVar: String, nodeVar: String, direction: SemanticDirection) = {
     val local = generator.declare(typeRef[RelationshipIterator], iterVar)
+    generator.assign(local, constant(null))
+    //Make sure cursor is always closed
+    _finalizers.append((_: Boolean) => (block) =>
+      using(block.ifStatement(Expression.notNull(block.load(local.name())))) {inner =>
+        inner.expression(invoke(inner.load(local.name()), method[RelationshipIterator, Unit]("close")))
+      })
```
davidegrohmann: This code is not gonna close the iterator when exhausted, is it?

pontusmelke: You are correct, nice catch! 

I fixed so that it is called each time it has exited loop, the solution is not super nice. But I am afraid making it nice will require some refactoring which I am not sure we should do at this point in time

davidegrohmann: Shouldn't the finalizers be called only when closing the result?

pontusmelke: They might not be in scope at that point, the change I made will close the iterator on the exit of every loop (as well as on other early exits such as exceptions etc but that was already there). So the generated code will look something like

```java
RelationshipIterator it;
it = null;
try
{
     it = this.getOrLoadReadOperations().nodeGetRelationships( v1, org.neo4j.graphdb.Direction.OUTGOING );
 }
 catch ( org.neo4j.kernel.api.exceptions.KernelException e )
{
    if ( it != null )
    {
          it.close();
     }
     throw new org.neo4j.cypher.internal.frontend.v3_2.CypherExecutionException( e.getUserMessage( new org.neo4j.kernel.api.StatementTokenNameLookup( this.ro ) ), e );
}
 while( it.hasNext() )
{
...
  if ( it != null )
  {
     it.close();
  }
}
```
It will also be closed on all exceptions etc within that loop but it is too much typing

PR #9327, opened 2017-05-10
Suggestion: Singleton annotation on procedures et al

pontusmelke:
Adds ability to annotate classes containing procedures, user-defined function, and user-defined aggregation functions with `@Singleton`. Whenever that annotation is present we only instantiate the surrounding class once.

Also allows singleton-classes to maintain non-static state.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/proc/ReflectiveProcedureCompiler.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/proc/ReflectiveProcedureCompiler.java
@@ -524,13 +524,11 @@ class ReflectiveProcedureCompiler
 
     private abstract static class ReflectiveBase
     {
-
         final List<FieldInjections.FieldSetter> fieldSetters;
         private final boolean singleton;
         protected Object[] args;
         protected Object cls;
         protected final MethodHandle constructor;
-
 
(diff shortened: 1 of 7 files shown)


systay:
Any progress on this PR, @pontusmelke?

systay:

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/ReflectiveProcedureCompiler.java
```diff
@@ -514,12 +524,18 @@ private QualifiedName extractName( Class<?> procDefinition, Method m, String val
 
     private abstract static class ReflectiveBase
     {
-
         final List<FieldInjections.FieldSetter> fieldSetters;
+        private final boolean singleton;
+        protected Object[] args;
+        protected Object cls;
```
thobe, line 530: This will not work!

It makes non-singleton procedures prone to race conditions, and thus not safe for invocations from multiple threads.

With these being fields they will be shared between different invocations of the same procedure, so you can get multiple concurrent invocations using another `cls` or `args` than their own, leading to very strange behaviour.

thobe (changes requested):
The arguments of a procedure cannot be a field on the procedure object.

And the instance field should not be used (at all) for non-singleton procedures.

PR #9330, opened 2017-05-11
Pooled concurrent merge scheduler for lucene indexes.

MishaDemianenko:
Lucene indexes merge scheduler that execute merges in a thread pool instead of starting separate thread for each
merge. Each writer have it's own scheduler but all of them use shared pool.
Current implementation is minimalistic in a number of things it knows about lucene internals to simplify
migrations overhead. It wraps up lucene internal merge tasks and execute them in a common thread pool.
In case if pool and queue exhausted merge will be performed in callers thread.

### DIFF
--- a/community/lucene-index/src/main/java/org/apache/lucene/index/PooledConcurrentMergeScheduler.java
+++ b/community/lucene-index/src/main/java/org/apache/lucene/index/PooledConcurrentMergeScheduler.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.apache.lucene.index;
+
+import java.io.IOException;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.LockSupport;
+
+import org.neo4j.helpers.NamedThreadFactory;
+import org.neo4j.unsafe.impl.internal.dragons.FeatureToggles;
+
+/**
+ * Lucene indexes merge scheduler that execute merges in a thread pool instead of starting separate thread for each
+ * merge. Each writer have it's own scheduler but all of them use shared pool.
+ *
+ * Current implementation is minimalistic in a number of things it knows about lucene internals to simplify
+ * migrations overhead. It wraps up lucene internal merge tasks and execute them in a common thread pool.
+ * In case if pool and queue exhausted merge will be performed in callers thread.
+ * Since we cant rely on lucene per writer merge threads we need to perform writer tasks counting ourselves to prevent
+ * cases while writer will be closed in the middle of merge and will wait for all writer related merges to complete
+ * before allowing close of writer scheduler.
+ */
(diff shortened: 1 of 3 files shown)


community/lucene-index/src/main/java/org/apache/lucene/index/PooledConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,157 @@
...
+ * before allowing close of writer scheduler.
+ */
+public class PooledConcurrentMergeScheduler extends ConcurrentMergeScheduler
+{
+    private static final int POOL_QUEUE_CAPACITY =
+            FeatureToggles.getInteger( PooledConcurrentMergeScheduler.class, "pool.queue.capacity", 100 );
+    private static final int POOL_MINIMUM_THREADS =
+            FeatureToggles.getInteger( PooledConcurrentMergeScheduler.class, "pool.minimum.threads", 4 );
+
+    private AtomicInteger writerTaskCounter = new AtomicInteger();
```
lutovich: Field can be final

community/lucene-index/src/main/java/org/apache/lucene/index/PooledConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,157 @@
...
+                MergeThread mergeThread = getMergeThread( writer, merge );
+                writerTaskCounter.incrementAndGet();
+                PooledConcurrentMergePool.mergeThreadsPool.submit( mergeTask( mergeThread ) );
+                success = true;
+            }
+            finally
+            {
+                if ( !success )
+                {
+                    writerTaskCounter.decrementAndGet();
```
lutovich: Maybe decrement should happen after `mergeFinish` so that `waitForAllTasks` observe finished merges?

community/lucene-index/src/main/java/org/apache/lucene/index/PooledConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,157 @@
...
+        //noop
+    }
+
+    @Override
+    void removeMergeThread()
+    {
+        // noop
+    }
+
+    AtomicInteger getWriterTaskCounter()
```
lutovich: Maybe expose just an `int` counter for tests and move `get()` inside here?

community/lucene-index/src/main/java/org/apache/lucene/index/PooledConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private Runnable mergeTask( MergeThread mergeThread )
+    {
+        return new MergeTask( mergeThread, writerTaskCounter );
+    }
+
+    private void waitForAllTasks()
+    {
+        while ( writerTaskCounter.get() > 0 )
```
lutovich: Would it make sense to use `Predicates.await()` here and limit max wait time somehow?

lutovich (commented):
I think this change generally looks good. Made couple comments.

community/lucene-index/src/main/java/org/apache/lucene/index/PooledConcurrentMergeScheduler.java
MishaDemianenko: yes will do that since that is more clear and expected and do not require any knowledge about mergeFinish content;
but in general, we do not need to do that since what we need to guarantee by the counter is that we do not have any ongoing merges going on and failing mergeFinish just do housekeeping with segments which should be safe.

MishaDemianenko:
@lutovich all the comments addressed, thx!

community/lucene-index/src/main/java/org/apache/lucene/index/PooledConcurrentMergeScheduler.java
```diff
@@ -0,0 +1,161 @@
...
+        {
+            MergePolicy.OneMerge merge = writer.getNextMerge();
+            if ( merge == null )
+            {
+                return;
+            }
+            boolean success = false;
+            try
+            {
+                MergeThread mergeThread = getMergeThread( writer, merge );
```
burqen, line 67: Will this not actually create a new `MergeThread` for every merge?

burqen (changes requested):
One comment 😄

community/lucene-index/src/main/java/org/apache/lucene/index/PooledConcurrentMergeScheduler.java
MishaDemianenko: `MergeThread` instance is created per each merge as before we only executing them inside of our thread pool

burqen: I thought the point was not to create new `Thread` objects all the time but to use an existing thread pool. However, `MergeThread extends Thread`.
So you will have a thread from the pool executing a runnable which is in fact actually a newly created thread. Seems a little bit weird?

burqen (approved):

PR #9336, opened 2017-05-13
Don't use pruning varexpand when relationship is needed

pontusmelke:
For queries where the relationship identifier is used in predicate
we can not use pruning varexpand. For example the query

```
MATCH (n)-[r:T*2]->() WHERE last(r).prop = 'foo' RETURN DISTINCT n
```
should not use pruning varexpand since it needs access to all the relationships.

changelog: Fixed error using pruning var-expand when relationship variable is referenced in predicate


fickludd (approved):

InverseFalcon:
Does this fix also work when we need to work with the path? This kind of query also errors out, we'd like to make sure it's covered too:

```
MATCH p=(n)-[*0..1]-()
 RETURN COLLECT(DISTINCT p)
```

PR #9339, opened 2017-05-14
Allow whitespace around transaction statements

cleishm:
Give transaction statements have the same flexibility wrt whitespace as
for regular cypher statements.


cleishm:
@jexp and I were trying to normalize dump/load commands for APOC and different shells and noticed that this wasn't consistent. Hopefully, this will help in future.

srbaker:
Hey, thanks for this. Can you re-submit this against 3.3, I'll happily merge it there. Thanks!

PR #9341, opened 2017-05-15
Lowers batch size of input-reading stages in import

tinwelint:
because the batches would be quite big for data that has a large
amount of property data and would risk out of memory. With this change
the batch size is around a couple of thosands, down from a couple of
tens of thousands (page size based)

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
@@ -374,7 +374,7 @@ public class ParallelBatchImporter implements BatchImporter
 
     private static Configuration configWithRecordsPerPageBasedBatchSize( Configuration source, RecordStore<?> store )
     {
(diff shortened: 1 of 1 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
```diff
@@ -374,7 +374,7 @@ private void importRelationships( NodeRelationshipCache nodeRelationshipCache,
 
     private static Configuration configWithRecordsPerPageBasedBatchSize( Configuration source, RecordStore<?> store )
     {
-        return Configuration.withBatchSize( source, store.getRecordsPerPage() * 100 );
+        return Configuration.withBatchSize( source, store.getRecordsPerPage() * 10 );
```
davidegrohmann, line 377: Why not making it configurable?

tinwelint: It's always painful introducing a config option for something that should just work. I don't think there's any benefit with the 100x multiplier... measuring as we speak

MishaDemianenko (approved):

PR #9350, opened 2017-05-16
Clears batch inserter cache on commit even if only reading

tinwelint:
otherwise the cache would grow indefinitely if there were no writes


MishaDemianenko:
@tinwelint let's suppress this one?

PR #9354, opened 2017-05-16
Handling START clauses in cost planner

pontusmelke:
We used to rely on the rule planner for most `START` queries, i.e. we would fail and fallback on the rule planner. Since we no longer have the rule planner around we need to support all legacy `START` queries in the cost planner. This is implemented as a rewriter that simply rewrites queries using `START` to an equivalent `MATCH` query.

changelog: Fixed error with `START` clauses in 3.2 caused by the removal of the rule planner.

### DIFF
--- a/community/cypher/frontend-3.2/src/main/scala/org/neo4j/cypher/internal/frontend/v3_2/ast/rewriters/startClauseRewriter.scala
+++ b/community/cypher/frontend-3.2/src/main/scala/org/neo4j/cypher/internal/frontend/v3_2/ast/rewriters/startClauseRewriter.scala
@@ -41,7 +41,7 @@ case object startClauseRewriter extends StatementRewriter {
           val invocation = FunctionInvocation(FunctionName(functions.Id.name)(pos), variable.copyId)(pos)
           In(invocation, ListLiteral(ids)(pos))(pos)
 
-        //START n=nodes(1,5,7)....
+        //START n=nodes({id})....
         case n@NodeByParameter(variable, parameter) =>
           val pos = n.position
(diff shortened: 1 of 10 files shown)


systay:
Looks wonderful. Are the two last commits really meant to have the same message?

community/cypher/frontend-3.2/src/main/scala/org/neo4j/cypher/internal/frontend/v3_2/ast/rewriters/startClauseRewriter.scala
```diff
@@ -0,0 +1,56 @@
...
+        case hint: LegacyIndexHint => hint
+      }
+      val nodes = items.collect {
+        case n: NodeStartItem => NodePattern(Some(n.variable.copyId), Seq.empty, None)(start.position)
+      }
+
+      val allPredicates = (newPredicates ++ where.map(_.expression)).toSet
+      val newWhere =
+        if (allPredicates.isEmpty) None
+        else if (allPredicates.size == 1) Some(Where(allPredicates.head)(start.position))
```
systay: You don't need to check this - `Ands.apply` already does it.

systay (approved):

community/cypher/frontend-3.2/src/main/scala/org/neo4j/cypher/internal/frontend/v3_2/ast/rewriters/startClauseRewriter.scala
pontusmelke: it does for the `commands.predicates.Ands` version not the `ast.Ands` which is what I am using here.

PR #9359, opened 2017-05-16
Force count store rebuild during store migration

MishaDemianenko:
Count store version was incremented but there was no migration added
and as result first DB after start will fail to open old count stores and
will rebuild it.
That way of doing count store migration is incorrect and produce
unexpected exceptions that confuse users.


chrisvest (approved):

chrisvest:
Weird test failure. Rerunning.

PR #9360, opened 2017-05-16
Added additional arguments to `neo4j-admin import` to match capabilities in `neo4j-import`

praveenag:
`neo4j-admin import` misses few functionalities that has been found useful in `neo4j-import`

We are porting the functionality either as is or with sensible defaults into `neo4j-admin` before removing `neo4j-import` in entirety.

This PR deals with:

- Use all available processors `--processors=(all available processors)`
- Always trim strings `--trim-strings=true`
- Import emptyQuotedStrings as empty Strings `--ignore-empty-strings=false`
- Buffer size is set to 4MB `--read-buffer-size=4194304`

### DIFF
--- a/community/csv/src/main/java/org/neo4j/csv/reader/Configuration.java
+++ b/community/csv/src/main/java/org/neo4j/csv/reader/Configuration.java
@@ -69,8 +69,8 @@ public interface Configuration
     {
         return DEFAULT_LEGACY_STYLE_QUOTING;
     }
-
     int KB = 1024, MB = KB * KB;
+    int DEFAULT_BUFFER_SIZE_4MB = 4 * MB;
 
     class Default implements Configuration
     {
@@ -83,7 +83,7 @@ public interface Configuration
         @Override
         public int bufferSize()
         {
-            return 4 * MB;
+            return DEFAULT_BUFFER_SIZE_4MB;
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


community/dbms/src/test/java/org/neo4j/commandline/dbms/CsvImporterTest.java
```diff
@@ -36,8 +38,14 @@
...
 import org.neo4j.test.rule.SuppressOutput;
 import org.neo4j.test.rule.TestDirectory;
+import org.neo4j.unsafe.impl.batchimport.Configuration;
+import org.neo4j.unsafe.impl.batchimport.input.csv.CsvInput;
 
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+
```
spacecowboy: This file has added imports but no changes in code. That means unused imports?

community/csv/src/main/java/org/neo4j/csv/reader/Configuration.java
```diff
@@ -69,9 +69,13 @@ default boolean legacyStyleQuoting()
     {
         return DEFAULT_LEGACY_STYLE_QUOTING;
     }
-
     int KB = 1024, MB = KB * KB;
 
+    static int defaultBufferSize4MB()
```
spacecowboy: Why a method? Why not a constant?

spacecowboy:
@praveenag do you intend to add commandline arguments to change the settings in this pr or in a second one?

Those arguments should probably target 3.3

praveenag:
@spacecowboy As far as this PR is concerned there aren't any commandline arguments to be added. The new ones will be on a separate PR but targeting 3.3.

PR #9362, opened 2017-05-17
GBPTree don't block close if need clean

burqen:
Close is not longer blocked by need cleaning. Instead, cleaner will just die silently if tree is closed. This is fine because the clean will just start again next time tree is opened.

Builds on #9309

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
@@ -998,7 +998,6 @@ public class GBPTree<KEY,VALUE> implements Closeable
         private final InternalTreeLogic<KEY,VALUE> treeLogic;
         private final StructurePropagation<KEY> structurePropagation;
         private PageCursor cursor;
-        private Lock sessionLock;
 
         // Writer can't live past a checkpoint because of the mutex with checkpoint,
         // therefore safe to locally cache these generation fields from the volatile generation in the tree
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
@@ -28,7 +28,7 @@ import java.nio.file.NoSuchFileException;
 import java.nio.file.StandardOpenOption;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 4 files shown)


community/common/src/main/java/org/neo4j/kernel/impl/util/JobSchedulerAdapter.java
```diff
@@ -0,0 +1,93 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.util;
+
+import java.util.Map;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+
+public class JobSchedulerAdapter implements JobScheduler
```
tinwelint: This could live in src/test, couldn't it? In the vein of cutting down on implementations active at runtime.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GroupingRecoveryCleanupWorkCollector.java
```diff
@@ -30,29 +34,44 @@
...
-    public synchronized void run()
+    public void start() throws Throwable
     {
-        assert !started : "Tried to start cleanup job more than once";
-        started = true;
         CleanupJob job;
         while ( (job = jobs.poll()) != null )
         {
-            job.run();
+            jobScheduler.schedule( recoveryCleanup, job );
```
tinwelint: I may be wrong, but it looks like we're scheduling all cleanup jobs individually. This also means that cleanup jobs could potentially run in parallel. What happens if there are multiple cleanup jobs, each using all available processors? There will be lots and lots of context switching in vain, won't there?

tinwelint:
Also there's a javadoc compilation issue in `GroupingRecoveryCleanupWorkCollector`

burqen:
@tinwelint Your comments a better suited for #9309 on which this PR builds.
Will address them there instead and rebase this branch.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GroupingRecoveryCleanupWorkCollector.java
burqen: Very good point and a nice catch! Addressing this in PR #9309

MishaDemianenko:
@burqen base was merged so you can rebase now

MishaDemianenko (approved):

PR #9365, opened 2017-05-17
ImportTools mentions regex capability in --nodes and --relationships

tinwelint:


burqen:
@tinwelint Changes look good, build does not.

PR #9372, opened 2017-05-17
3.3 import feature parity

praveenag:
These are new functionalities added to `neo4j-admin import` that has been found useful in `neo4j-import`

This PR deals with:

- `--delimiter`, `--array-delimiters`, `--quote`
- `--multiline-fields` to import csv data with line breaks
- `--max-memory` to define maximum memory that the importer can use
- Quoting respects RFC 4180 reported as an issue in https://github.com/neo4j/neo4j/issues/8472

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
@@ -40,6 +40,7 @@ import org.neo4j.helpers.collection.MapUtil;
 import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.impl.util.Validators;
 import org.neo4j.server.configuration.ConfigLoader;
+import org.neo4j.unsafe.impl.batchimport.input.csv.Configuration;
 
 import static org.neo4j.csv.reader.Configuration.DEFAULT;
 import static org.neo4j.unsafe.impl.batchimport.Configuration.DEFAULT_MAX_MEMORY_PERCENT;
@@ -49,7 +50,6 @@ public class ImportCommand implements AdminCommand
 {
     public static final String DEFAULT_REPORT_FILE_NAME = "import.report";
     private static final String[] allowedModes = {"database", "csv"};
-    private static final String[] allowedDelimiterCharacters = {",", "TAB", ";", "|", ":", "#"};
     private static final Arguments databaseArguments = new Arguments()
             .withArgument( new MandatoryNamedArg( "mode", "database", "Import a pre-3.0 installation." )
             {
@@ -109,10 +109,11 @@ public class ImportCommand implements AdminCommand
                     "Whether or not fields from input source can span multiple lines," +
                             " i.e. contain newline characters." ) )
             .withArgument( new OptionalNamedArg( "delimiter",
-                    allowedDelimiterCharacters,
-                    String.valueOf( COMMAS.delimiter() ), "Delimiter character between values in CSV data." ) )
+                    "delimiter-character",
+                    String.valueOf( COMMAS.delimiter() ),
+                    "Delimiter character between values in CSV data." ) )
             .withArgument( new OptionalNamedArg( "array-delimiter",
-                    allowedDelimiterCharacters,
+                    "array-delimiter-character",
                     String.valueOf( COMMAS.arrayDelimiter() ),
                     "Delimiter character between array elements within a value in CSV data." ) )
             .withArgument( new OptionalNamedArg( "max-memory",
@@ -170,10 +171,10 @@ public class ImportCommand implements AdminCommand
                     "Whether or not fields from input source can span multiple lines," +
                             " i.e. contain newline characters." ) )
             .withArgument( new OptionalNamedArg( "delimiter",
-                    allowedDelimiterCharacters,
+                    String.valueOf( COMMAS.delimiter() ),
                     String.valueOf( COMMAS.delimiter() ), "Delimiter character between values in CSV data." ) )
             .withArgument( new OptionalNamedArg( "array-delimiter",
-                    allowedDelimiterCharacters,
+                    String.valueOf( COMMAS.delimiter() ),
                     String.valueOf( COMMAS.arrayDelimiter() ),
                     "Delimiter character between array elements within a value in CSV data." ) )
             .withArgument( new OptionalNamedArg( "max-memory",
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
@@ -40,7 +40,6 @@ import org.neo4j.helpers.collection.MapUtil;
 import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.impl.util.Validators;
 import org.neo4j.server.configuration.ConfigLoader;
-import org.neo4j.unsafe.impl.batchimport.input.csv.Configuration;
 
 import static org.neo4j.csv.reader.Configuration.DEFAULT;
 import static org.neo4j.unsafe.impl.batchimport.Configuration.DEFAULT_MAX_MEMORY_PERCENT;
@@ -116,6 +115,12 @@ public class ImportCommand implements AdminCommand
                     "array-delimiter-character",
                     String.valueOf( COMMAS.arrayDelimiter() ),
                     "Delimiter character between array elements within a value in CSV data." ) )
+            .withArgument( new OptionalNamedArg( "quote",
+                    "quotation-character",
+                    String.valueOf( COMMAS.quotationCharacter() ),
+                    "Character to treat as quotation character for values in CSV data. "
+                            + "Quotes can be escaped as per RFC 4180 by doubling them, for example \"\" would be " +
+                            "interpreted as a literal \". You cannot escape using \\." ) )
             .withArgument( new OptionalNamedArg( "max-memory",
                     "max-memory-that-importer-can-use",
                     String.valueOf( DEFAULT_MAX_MEMORY_PERCENT ) + "%",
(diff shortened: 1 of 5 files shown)


community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
```diff
@@ -41,10 +41,15 @@
...
 
+import static org.neo4j.csv.reader.Configuration.DEFAULT;
+import static org.neo4j.unsafe.impl.batchimport.Configuration.DEFAULT_MAX_MEMORY_PERCENT;
+import static org.neo4j.unsafe.impl.batchimport.input.csv.Configuration.COMMAS;
+
 public class ImportCommand implements AdminCommand
 {
     public static final String DEFAULT_REPORT_FILE_NAME = "import.report";
     private static final String[] allowedModes = {"database", "csv"};
+    private static final String[] allowedDelimiterCharacters = {",", "TAB", ";", "|", ":", "#"};
```
spacecowboy: Why do we all of the sudden have an allowed set of delimiter characters? What prompted this design change?

community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
```diff
@@ -98,7 +103,28 @@ public String usage()
...
+                    String.valueOf( COMMAS.delimiter() ), "Delimiter character between values in CSV data." ) )
+            .withArgument( new OptionalNamedArg( "array-delimiter",
+                    allowedDelimiterCharacters,
+                    String.valueOf( COMMAS.arrayDelimiter() ),
+                    "Delimiter character between array elements within a value in CSV data." ) )
+            .withArgument( new OptionalNamedArg( "max-memory",
+                    "max-memory-that-importer-can-use",
+                    String.valueOf( DEFAULT_MAX_MEMORY_PERCENT ),
+                    "Maximum memory that importer can use for various data structures and caching " +
+                            "to improve performance. If left as unspecified (null) it is set to " +
```
spacecowboy: This text should be updated for `neo4j-admin`. I am thinking in particular about 

> If left as unspecified (null) it is set to ...

This entire sentence can be removed. The arguments handler automatically adds a suffix to the description with the default value. Null is also a value which is never relevant.

community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
```diff
@@ -140,8 +166,27 @@ public String usage()
...
```
spacecowboy: Same comment as above

community/dbms/src/test/java/org/neo4j/commandline/dbms/ImportCommandTest.java
```diff
@@ -198,67 +198,85 @@ public void shouldPrintNiceHelp() throws Throwable
...
+                            "      Delimiter character between values in CSV data. [default:,]\n" +
+                            "  --array-delimiter=<,|TAB|;|||:|#>\n" +
+                            "      Delimiter character between array elements within a value in CSV data.\n" +
+                            "      [default:;]\n" +
+                            "  --max-memory=<max-memory-that-importer-can-use>\n" +
+                            "      Maximum memory that importer can use for various data structures and\n" +
+                            "      caching to improve performance. If left as unspecified (null) it is set to\n" +
+                            "      90% of (free memory on machine - max JVM memory). Values can be plain\n" +
+                            "      numbers, like 10000000 or e.g. 20G for 20 gigabyte, or even e.g. 70%.\n" +
+                            "      [default:90]\n",
```
spacecowboy: Here you can see what I mean about the sentence saying

> If left unspecified (null) ....

should be removed. The default is already listed.

An additional point: the default should be listed as `90%` to match the text and increase clarity. Now it looks like the default is `90 bytes`.

spacecowboy (changes requested):

community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
praveenag: I thought it'd be best to have a strict set of delimiters because there are times other special characters are used in a different context. Ex: If you specify `\` as delimiter it would conflict because we are turning on` --legacy-style-quoting=true` by default.

Having said this, I have just now verified that the allowedDelimiterCharacters is of no use because there are no validations done on verifying the delimiter character is one of the allowedCharacters.

community/dbms/src/test/java/org/neo4j/commandline/dbms/ImportCommandTest.java
praveenag: Would it work if it said `[default:90% of (free-memory-on-machine - max-JVM-memory)]` or just `[default:90%]`?

community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
spacecowboy: But this is not new; `\` would always have conflicted. But by restricting to a specific set of characters you are setting yourself up to get a feature request to add X to the list of allowed characters, or complaints that "This used to work in import-tool". I just think it's not worth that hassle without good reason.

community/dbms/src/test/java/org/neo4j/commandline/dbms/ImportCommandTest.java
spacecowboy: you don't control what the default says, the default value is listed as is. So just make sure it's `90%` and listed as such. What the percentage means is covered in the description.

community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
tinwelint: Mmh, agreed. Any set of delimiters is probably too strict and doesn't simplify the surface for the user.

community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
tinwelint: The default value is null because it will be calculated per machine/environment. That's why that comment is in the description.

community/dbms/src/test/java/org/neo4j/commandline/dbms/ImportCommandTest.java
tinwelint: Generally the theme of copying these descriptions into three locations isn't great is it?

community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
tinwelint: Also the neo4j-import supports `\number`, which is a very nice "whatever character you want" thing.

praveenag: I compiled the list of characters after speaking with @sarmbruster .  I thought restricting the characters would mean less no of support requests when there are errors but maybe I'm looking at this the wrong way. What is your preference @sarmbruster ?

community/dbms/src/test/java/org/neo4j/commandline/dbms/ImportCommandTest.java
spacecowboy: no it isn't. problem is that the import tool is actually two commands in one. in an ideal world we  would have the two separate commands `neo4j-admin import-csv` and `neo4j-admin import-db`

praveenag: @tinwelint Yes. It is on my list of things to refactor.

community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
sarmbruster: the list above is the result of a discussion I've had with @praveenag . She thought allowing everything can cause issues if the delimiter interferes with array delimiter or quoting character. My take was to at least have a choice of the most common ones - which is that list.
Any CSV file used by batch importer needs to be specifically built for that purpose due to the conventions being used (think header column == prop name, type encoded in header columns, etc.). Because of that limiting down the allowed delimiters sounds like a good compromise.

praveenag:
@spacecowboy and @sarmbruster I have made the changes requested now. Is it ok to merge now?

sarmbruster:
OK with me.

community/dbms/src/test/java/org/neo4j/commandline/dbms/ImportCommandTest.java
```diff
@@ -198,67 +198,83 @@ public void shouldPrintNiceHelp() throws Throwable
...
+                            "      contain newline characters. [default:false]\n" +
+                            "  --delimiter=<,>\n" +
+                            "      Delimiter character between values in CSV data. [default:,]\n" +
+                            "  --array-delimiter=<,>\n" +
+                            "      Delimiter character between array elements within a value in CSV data.\n" +
+                            "      [default:;]\n" +
+                            "  --max-memory=<max-memory-that-importer-can-use>\n" +
+                            "      Maximum memory that neo4j-admin can use for various data structures and\n" +
+                            "      caching to improve performance. Values can be plain numbers, like 10000000\n" +
+                            "      or e.g. 20G for 20 gigabyte, or even e.g. 70%. [default:90%]\n",
```
spacecowboy: This test will fail on Windows since you are using `\n` lineendings. Note that the previous version used `%n` together with `String.format` to get a platform-dependant lineending.

spacecowboy (changes requested):

community/dbms/src/test/java/org/neo4j/commandline/dbms/ImportCommandTest.java
praveenag: Oh I see the problem I was having problems using some of the text using `%` in the help text. I will try and fix it

spacecowboy (approved):
lgtm now

PR #9381, opened 2017-05-18
Add property validation on label add.

ragadeeshu:
Validates properties that are to be indexed because of an added label.


tinwelint (approved):
Nice catch

PR #8937, opened 2017-03-02
Filter cost model update

systay:
Before this change, the cost model mistakenly did not account for store access in the cost calculation of Selection (a.k.a Filter).

changelog: Update cost for filtering that uses property and label access.

PR #9383, opened 2017-05-19
Reverted fix to Selection cost

pontusmelke:
A bug fix in the way we calculate the cost-per-row of `Selection`,
https://github.com/neo4j/neo4j/pull/8937, lead to regressions. This
reverts those changes and we need to investigate more about why the fix
introduced regressions.

changelog: [3.1] Reverted a planner cost model calculation fix that was taking into account multiple predicates in a way that caused bad plans and worse performance for some queries


pontusmelke:
Waiting for confirmation that this fix actually resolves the regression.

PR #9384, opened 2017-05-19
Remove NEW_THREAD scheduling strategy from Noe4jJobScheduler

davidegrohmann:
Since the best way to create a new thread by using the scheduler is to
use the thread factory, this changes will make more straightforward
how to create new thread using the scheduler.  And also note that the
NEW_THREAD strategy was unsupported in several schedule methods.


chrisvest (approved):
Very nice!

PR #9386, opened 2017-05-19
Don't cache unchanged records in DirectRecordAccess

tinwelint:
which is used in BatchInserter. Previously all loaded records were
cached in the map. That map was only cleared every N number of changes
so if there were no changes that map would just keep growing.

This change will only let changed records be put in the map. This will
remove the issue of growing memory on reads.


tinwelint:
@MishaDemianenko I'd think this is a better solution for what #9350 tries to do

MishaDemianenko (approved):
Agree this looks much better, approved!

PR #9387, opened 2017-05-19
Page cache hit ratio added to metrics and profiler

klaren:
There is now `neo4j.page_cache.hit_ratio` metric that reports global hit ratio.
Cypher `PROFILE` query's responds with this metric with hit ratio of that particular query.
Hit ratio is sent with the Bolt metadata for driver visibility.


community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/CypherAdapterStreamTest.java
```diff
@@ -151,10 +162,31 @@ public void shouldIncludeProfileIfPresent() throws Throwable
...
+                "operatorType", "Join",
+                "children", Arrays.asList( expectedChild ),
+                "rows", 1L,
+                "dbHits", 2L,
+                "pageCacheHits", 4L,
+                "pageCacheMisses", 3L,
+                "pageCacheHitRatio", 4.0 / 7
+        );
+
+        assertThat( meta.get( "profile" ), equalTo( expectedProfile ));
```
MishaDemianenko: since one of the values in a map a double if you want to compare doubles between each other usually you need some kind of comparison precision like you did in `ExecutionPlanConverterTest.profileStatisticConversion`

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/DefaultPageCursorTracer.java
```diff
@@ -159,6 +159,12 @@ public long flushes()
     }
 
     @Override
+    public double hitRatio()
+    {
+        return ((double) hits()) / (hits() + faults());
```
MishaDemianenko: I would extract hits to local variable here even if it's always same value that will be return from those calls

MishaDemianenko (commented):
Couple of small comments, as additional comment I do not think all of the places that require modification was updated please ping me when ready lets take a walk together around it

MishaDemianenko:
@klaren unfortunately yet another rebase required here

community/common/src/main/java/org/neo4j/helpers/MathUtil.java
```diff
@@ -20,11 +20,35 @@
...
 
+    private MathUtil()
+    {
+        throw new AssertionError();
+    }
+
+    /**
+     * Calculates the portion of the first value to all values passed
+     * @param n The values in the set
+     * @return the ratio of n[1] to the sum all n, 0 if result is {@link Double#NaN}
```
MishaDemianenko: small thing here: the ratio of n[0] to the sum of all.

MishaDemianenko (commented):
One small javadoc comment otherwise it's fine except we need another rebase because of the conflict and it looks like current base version was a bit broken as well

MishaDemianenko (approved):

PR #9388, opened 2017-05-19
Replace multiple locks in GBPTree with single simple lock

burqen:


community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTreeLock.java
```diff
@@ -0,0 +1,119 @@
...
+        do
+        {
+            currentState = state;
+            if ( !canUnlock( currentState, targetLockBit) )
+            {
+                throw new IllegalStateException( "Can not unlock lock that is already locked" );
+            }
+            newState = currentState & ~targetLockBit;
+        }
+        while ( !UnsafeUtil.compareAndSwapLong( this, stateOffset, currentState, newState ) );
```
tinwelint, line 102: I'd guess there _could_ be an `unpark` somewhere, but since this isn't a highly contented lock at all perhaps there's no need to bother, is there?

community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeLockTest.java
```diff
@@ -0,0 +1,189 @@
...
+        shouldWait( future );
+        runUnlock.run();
+        future.get();
+    }
+
+    private void shouldWait( Future<?> future )throws InterruptedException, ExecutionException
+    {
+        try
+        {
+            future.get( 200, TimeUnit.MILLISECONDS );
```
tinwelint, line 253: I'd prefer `OtherThreadRule` instead. With it you can wait for the code to get into wait or block in specific part of code. Regardless of how long this timeout is there's a chance of flakyness and best case it waits 200ms which is unnecessary and accumulates to a non-neglectible time if there are multiple tests.

tinwelint (changes requested):

burqen:
... and also need testing for two threads racing for same lock / overlapping lock.

community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeLockTest.java
tinwelint: Hmm, that rule exists in neo4j-kernel. Could be a separate PR for moving it to something like neo4j-common and making use of it.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTreeLock.java
burqen: I agree that we don't need to bother with that, since the cost of making the code more complex is higher.

tinwelint (approved):

PR #9391, opened 2017-05-19
More debug info for failed transaction apply

chrisvest:
We're not in a good spot if we fail to apply a transaction. Such failures would
happen after a transaction has already been committed, or during recovery. Both
are awful places for exceptions to pop out.

This commit captures exceptions that occur during apply, and wraps them in
another exception that adds information about the transaction, e.g. log
position, which might be helpful for debugging the cause of the failure.

### DIFF
--- a/community/common/src/main/java/org/neo4j/kernel/lifecycle/LifecycleException.java
+++ b/community/common/src/main/java/org/neo4j/kernel/lifecycle/LifecycleException.java
@@ -27,42 +27,63 @@ public class LifecycleException
     extends RuntimeException
 {
 
-    private static String humanReadableMessage( Object instance, LifecycleStatus from, LifecycleStatus to )
+    private static String humanReadableMessage(
+            Object instance, LifecycleStatus from, LifecycleStatus to, Throwable cause )
     {
-        switch(to)
+        String instanceStr = String.valueOf( instance );
+        StringBuilder message = new StringBuilder();
+        switch ( to )
         {
             case STOPPED:
-                if(from == LifecycleStatus.NONE)
+                if ( from == LifecycleStatus.NONE )
                 {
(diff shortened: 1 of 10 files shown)


community/common/src/main/java/org/neo4j/kernel/lifecycle/LifecycleException.java
```diff
@@ -27,42 +27,63 @@
...
-                instance.toString(), from.name(), to.name() );
+        return message.toString();
+    }
+
+    private static Throwable rootCause( Throwable cause )
+    {
+        int i = 0; // Guard against infinite self-cause exception-loops.
+        while ( cause.getCause() != null && i++ < 100 )
+        {
+            cause = cause.getCause();
```
MishaDemianenko, line 85: Maybe it makes sense to include into error message fact that causes were cycled?

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/TransactionApplyKernelException.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class TransactionApplyKernelException extends KernelException
+{
+    public TransactionApplyKernelException( Throwable cause, String message,
+                                               Object... parameters )
+    {
+        super( Status.General.UnknownError, cause, message, parameters );
+    }
+
+    public TransactionApplyKernelException( String message, Object... parameters )
```
MishaDemianenko: maybe we can avoid this constructor for now since it's not used

MishaDemianenko (commented):
Two minor comments, otherwise looks good

chrisvest:
@MishaDemianenko I've addressed you comments.

MishaDemianenko (approved):

PR #9393, opened 2017-05-20
fixes #9390 apply nodefilter only on endNode

### LINKED ISSUE #9390: GraphAlgoFactory.shortestPath does not honor PathExpanderBuilder.addNodeFilter
When `GraphAlgoFactory.shortestPath()` is used with a pathexpander acquired from `PathExpanderBuilder` using a node filter, that node filter is *not* applied when left and right traversal branch collide. So the "middle" nodes do not get checked.

sarmbruster:
(Added by @burqen)
Previously node filter was applied to node when expanding _from_ it with expansion only happening if node was accepted by the filter. This meant that the node on which collision happened in bidirectional traversal would never be checked against the filter.

This PR changes this to apply filter on the node we are expanding _to_ meaning we will never reach nodes that does not pass the filter.

### DIFF
--- a/community/graph-algo/src/test/java/org/neo4j/graphalgo/impl/path/TestShortestPath.java
+++ b/community/graph-algo/src/test/java/org/neo4j/graphalgo/impl/path/TestShortestPath.java
@@ -23,12 +23,27 @@ import common.Neo4jAlgoTestCase;
 import org.apache.commons.lang3.mutable.MutableInt;
 import org.junit.Test;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
 import java.util.function.Predicate;
 
 import org.neo4j.graphalgo.GraphAlgoFactory;
 import org.neo4j.graphalgo.PathFinder;
-import org.neo4j.graphdb.*;
+import org.neo4j.graphdb.Direction;
+import org.neo4j.graphdb.Label;
+import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.Path;
+import org.neo4j.graphdb.PathExpander;
+import org.neo4j.graphdb.PathExpanderBuilder;
+import org.neo4j.graphdb.PathExpanders;
+import org.neo4j.graphdb.Relationship;
+import org.neo4j.graphdb.RelationshipType;
+import org.neo4j.graphdb.ResourceIterator;
 import org.neo4j.graphdb.impl.StandardExpander;
 import org.neo4j.graphdb.traversal.BranchState;
 import org.neo4j.helpers.collection.Iterables;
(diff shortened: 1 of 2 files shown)


community/graph-algo/src/test/java/org/neo4j/graphalgo/impl/path/TestShortestPath.java
```diff
@@ -23,35 +23,20 @@
 import org.apache.commons.lang3.mutable.MutableInt;
 import org.junit.Test;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
tinwelint: No star imports, please

tinwelint:
I may be wrong, but this doesn't strike me as a correct fix, as it looks like it's changing other behaviour too. Is this a bi-directional traversal issue? If so perhaps the fix should be in the bi-directional part of the code.

Your commit doesn't explain the reasoning behind the change.

community/graph-algo/src/test/java/org/neo4j/graphalgo/impl/path/TestShortestPath.java
```diff
@@ -244,6 +229,34 @@ public void withFilters() throws Exception
     }
 
     @Test
+    public void filtersTouchesAllIntermediateNodes() throws Exception
```
burqen, line 257: This code is breaking some code style rules. Are you using the Neo4j formatting? If you do, try to auto-format this part. If you are not, I'll be happy to help you out.

(This is not a problem for 3.1 because checkstyle rules are not as mature there but it will be when forward merging so might as well fix it before)

burqen (changes requested):

community/graph-algo/src/test/java/org/neo4j/graphalgo/impl/path/TestShortestPath.java
sarmbruster: where can I find the current code style data?

systay:
@sarmbruster @tinwelint Any progress on this PR? Is it being worked on, or should we close it for now?

sarmbruster:
I've addressed the code style changes requested above.

sarmbruster:
The PR basically solves the issue referred to (#9390). When doing shortestPath (or other) algos, the nodefilter is not applied on the node joing left and right traversal. This is fixed here.

systay:
Do you have an answer to @tinwelint's comments above? (https://github.com/neo4j/neo4j/pull/9393#issuecomment-303711409)

burqen:
@tinwelint I added an explanation of the PR. What do you think about this change?

tinwelint:
I read the changes wrong previously, I do think the fix is correct!

burqen (approved):

burqen:
Talked with @tinwelint and we both agree that this is the correct thing to do.

PR #9396, opened 2017-05-22
Makes Hazelcast timeouts depend on leader_election_timeout

digitalstain:
Hazelcast specific timeouts were hardwired to 10s. This meant that
 GC pauses larger than that would result in timeout exceptions from
 HC, which, while mostly benign, could occasionally result in leader
 switches even if the leader_election_timeout could be set higher
 to compensate. Having HC timeouts bound to election timeout makes
 sense also if one thinks of the election timeout as an indication
 of the typical expected GC pause. Since HC runs on the same VM, it
 should also be configured in a similar way.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastCoreTopologyService.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastCoreTopologyService.java
@@ -19,6 +19,12 @@
  */
 package org.neo4j.causalclustering.discovery;
 
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
 import com.hazelcast.config.InterfacesConfig;
 import com.hazelcast.config.JoinConfig;
 import com.hazelcast.config.MemberAttributeConfig;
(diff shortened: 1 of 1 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastCoreTopologyService.java
```diff
@@ -165,10 +165,14 @@ private HazelcastInstance createHazelcastInstance()
...
+        Long electionTimeoutMillis = config.get( CausalClusteringSettings.leader_election_timeout );
+        Long baseHazelcastTimeoutMillis = ( 3 * electionTimeoutMillis ) / 2;
+
         com.hazelcast.config.Config c = new com.hazelcast.config.Config();
-        c.setProperty( OPERATION_CALL_TIMEOUT_MILLIS.getName(), String.valueOf( 10_000 ) );
-        c.setProperty( MERGE_NEXT_RUN_DELAY_SECONDS.getName(), "10" );
-        c.setProperty( MERGE_FIRST_RUN_DELAY_SECONDS.getName(), "10" );
+        c.setProperty( OPERATION_CALL_TIMEOUT_MILLIS.getName(), String.valueOf( baseHazelcastTimeoutMillis ) );
+        c.setProperty( MERGE_NEXT_RUN_DELAY_SECONDS.getName(), String.valueOf( baseHazelcastTimeoutMillis / 1000 ) );
+        c.setProperty( MERGE_FIRST_RUN_DELAY_SECONDS.getName(), String.valueOf( baseHazelcastTimeoutMillis / 1000 ) );
```
martinfurmanski: Would be good to round these two upwards to the nearest second.

martinfurmanski (changes requested):

PR #9397, opened 2017-05-22
Fix safe zone, assert KernelStatement is still open when reading properties

burqen:
Continuation of #9016

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/storageengine/api/StorageStatement.java
+++ b/community/kernel/src/main/java/org/neo4j/storageengine/api/StorageStatement.java
@@ -22,6 +22,7 @@ package org.neo4j.storageengine.api;
 import java.util.function.IntPredicate;
 
 import org.neo4j.cursor.Cursor;
+import org.neo4j.kernel.api.AssertOpen;
 import org.neo4j.kernel.api.exceptions.index.IndexNotFoundKernelException;
 import org.neo4j.kernel.api.schema.index.IndexDescriptor;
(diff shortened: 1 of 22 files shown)


community/kernel/src/main/java/org/neo4j/storageengine/api/StorageStatement.java
```diff
@@ -112,9 +113,10 @@
      -     */
     Cursor<RelationshipItem> relationshipsGetAllCursor();
 
-    Cursor<PropertyItem> acquirePropertyCursor( long propertyId, Lock shortLivedReadLock );
+    Cursor<PropertyItem> acquirePropertyCursor( long propertyId, Lock shortLivedReadLock, AssertOpen assertOpen );
```
MishaDemianenko, line 116: Original PR introduced `assertOpen` checks for `acquireSingleNodeCursor `, `acquireSingleRelationshipCursor`, etc why we do not need those anymore here? (and related stack)

MishaDemianenko (commented):
I have one comment/question

community/kernel/src/main/java/org/neo4j/storageengine/api/StorageStatement.java
burqen: In 3.2 you can no longer get a property cursor directly from a node cursor. Instead you explicitly grab a property cursor by providing node or relationship object and use `nextPropertyId` to initialise a new cursor in `StorageLayer`.

MishaDemianenko (approved):

PR #9404, opened 2017-05-22
Support for multiple bookmarks in Bolt server

lutovich:
Previously Bolt server could only accept and process a single bookmark supplied with parameters for `RUN(BEGIN)` command. However there are use-cases where clients want to supply multiple bookmarks produced by separate workers. Ability to supply multiple bookmarks is needed for such cases. Clients can't compare bookmarks because they are just opaque strings.

This commit makes it possible to supply multiple bookmarks in a new "bookmarks" metadata parameters field. Server will try to parse multiple bookmarks and will fallback to parsing a single bookmark denoted by "bookmark" key for backwards compatibility reasons. Latest bookmark will be selected and server will wait for it.

Related Java driver PR: https://github.com/neo4j/neo4j-java-driver/pull/376

### DIFF
--- a/community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/bookmarking/BookmarkTest.java
+++ b/community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/bookmarking/BookmarkTest.java
@@ -21,39 +21,83 @@ package org.neo4j.bolt.v1.runtime.bookmarking;
 
 import org.junit.Test;
 
+import java.util.HashMap;
+import java.util.Map;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.emptyMap;
+import static java.util.Collections.singletonMap;
+import static org.hamcrest.Matchers.instanceOf;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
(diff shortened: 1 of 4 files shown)


community/bolt/src/test/java/org/neo4j/bolt/v1/runtime/bookmarking/BookmarkTest.java
```diff
@@ -89,20 +171,206 @@ public void shouldFailWhenNoNumberFollowsThePrefix() throws Exception
...
+        Map<String,Object> params = params( null,
+                asList( "neo4j:bookmark:v1:tx3", "neo4j:bookmark:v1:tx5", null, "neo4j:bookmark:v1:tx17" ) );
+
+        Bookmark bookmark = Bookmark.fromParamsOrNull( params );
+
+        assertEquals( 17, bookmark.txId() );
+    }
+
+    @Test
+    public void shouldReturnNull() throws Exception
```
zhenlineo: ??
`shouldIgnoreNullSingleBookmarkWhenMultipleBookmarksIsNotNull`?

zhenlineo: ??
I think this test case is already covered.
Suggest to delete duplicate

zhenlineo (approved):
LGTM
Maybe edit the title for a better changelog message?

PR #9410, opened 2017-05-23
Support logging in with samaccountname

OliviaYtterbrink:
changelog: Enable support for logging in with sAMAccountName (`dbms.security.ldap.authentication.use_samaccountname=true`) instead of distinguished name to ldap server , requires setting `dbms.security.ldap.authorization.system_username` and `dbms.security.ldap.authorization.system_password`


systay:
Is this PR active or should we close it?

OliviaYtterbrink:
This is active and ready for review

enterprise/security/src/test/resources/ad_test_data.ldif
```diff
@@ -0,0 +1,79 @@
...
+ou: users
+
+# Entry 10: cn=neo4j,ou=users,dc=example,dc=com
+dn: cn=n neo4j,ou=users,dc=example,dc=com
+memberOf: cn=admin,ou=groups,dc=example,dc=com
+objectClass: top
+objectClass: inetOrgPerson
+objectclass: simulatedMicrosoftSecurityPrincipal
+samAccountName: neo4j
+cn: neo4j
```
fickludd: Should this be `cn: n neo4j`?

I actually find the cn `n neo4j` really weird. Why did you take that?

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/LdapRealm.java
```diff
@@ -517,6 +526,64 @@ private String parseLdapServerUrl( String rawLdapServer )
...
+            String filter = "sAMAccountName={0}";
+            NamingEnumeration<SearchResult> search = ctx.search( userSearchBase, filter, searchArguments, searchCtls );
+            if ( search.hasMore() )
+            {
+                final SearchResult next = search.next();
+                String loginUser = next.getNameInNamespace();
+                if ( search.hasMore() )
+                {
+                    securityLog.error( "More than one user matching: " + principal );
+                    throw new RuntimeException( "More than one user matching: " + principal );
```
fickludd: Maybe we should throw something nicer than raw `RuntimeException`?

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/LdapRealm.java
```diff
@@ -517,6 +526,64 @@ private String parseLdapServerUrl( String rawLdapServer )
...
+                        securityLog.warn( "Error in authentication for user " + loginUser, ex );
+                        // We have to rethrow the exception, to indicate invalid login
+                        throw ex;
+                    }
+                }
+            }
+            else
+            {
+                securityLog.warn( "No user matching: " + principal );
+                throw new RuntimeException( "No user matching: " + principal );
```
fickludd: Again, can we do better than `RuntimeException`?

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/LdapRealm.java
```diff
@@ -143,38 +144,45 @@ protected AuthenticationInfo queryForAuthenticationInfo( AuthenticationToken tok
...
-                AuthenticationInfo info =
-                        useStartTls ? queryForAuthenticationInfoUsingStartTls( token, ldapContextFactory ) :
-                        super.queryForAuthenticationInfo( token, ldapContextFactory );
-                securityLog.debug( withRealm( "Authenticated user '%s' against %s",
-                        token.getPrincipal(), serverString ) );
-                return info;
+                return queryForAuthenticationInfoSAM( token, ldapContextFactory );
             }
-            catch ( Exception e )
+            else
```
fickludd, line 152: The diff in this method looks big, but I interpret it as first checking if we can do the SAMAccountName login, and otherwise doing the same as before?

fickludd (changes requested):
I think this looks good from what LDAP foo I have left. Would just like better exceptions than RuntimeException. Nicely done!

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/LdapRealm.java
OliviaYtterbrink: Yes, the only change in this method was the check, everything else is indentation change

enterprise/security/src/test/resources/ad_test_data.ldif
OliviaYtterbrink: it should be, though I will change it to `n.neo4j`, my reason is that usually a cn is of the format name.sirname

fickludd (approved):

PR #9414, opened 2017-05-24
Values module

fickludd:
This PR introduces a new library/module for handling property values. This work is based on the `*Property` class hierarchy, which it should replace.

The scope of the implemented values is all types that are stored, and supports 

 * *equality*, including `'a' == "a"`, `(byte)1 == 1L`, `1 == 1.0f` and `[1, 2] == [1.0f, 2.0f]`
 * *hashCode*, consistent with equality
 * *comparison*, consistent with equality and eg. `compare( 'a', 2.0 )`

Also included is lazy versions of String and all arrays, as ported from the `Lazy*Property` classes. What is not yet included is support for non-property values that are needed by Cypher: `Node`, `Relationship` and `Path`. That extension will be added separately.

The suggested working order is to get review and feedback on design and architecture of the library in this PR, while we will continue on another PR where this library is used to replace the `*Property` classes so we can benchmark the effects. I recommend review by check-out and exploration. Commits are coherent but partly overlapping as the structure evolved, so commit-by-commit review makes sense for understanding, but not for commenting.


tinwelint:
I like the idea and the separation of it

fickludd:
@tinwelint nice!

@thobe has suggested ways to merge the Lazy and Direct subtrees. That could have several advantages: 
 * slightly fewer classes
 * maybe no longer need separate interfaces in ValueGroup, but can share the capability grouping with the unified implementation tree. 

Also considering to restructure NoValue to sit beside Value, so we can have typesafe knowledge of whether a value is guaranteed to be returned.

MishaDemianenko:
General comments:
* it's a good idea to start extracting code to smaller modules
* IMHO it would be nice to have separate packages for direct and lazy values
* do not really like V notation since that hurting readability
* would like if we will extract interfaces from ValueGroup do not see a reason why they need to be inside one class
* it looks like ValueImpl class can be removed at all
* do not like the reuse of `Object value` for producer and value itself in Lazy values
* looks like value comparison is insane cast machinery :)
* would be nice t see more javadocs in final variant

fickludd:
@MishaDemianenko good points.

* separating direct and lazy could be good. I tried something else now to solve some other points.
* V-notation is gone. I used it to avoid confusion between ValueGroup.VNumber and java.lang.Number during dev, and forgot to change back.
* interfaces in ValueGroup are gone
* ValueImpl is gone
* the Object value thing in LazyValue is a bit yucky, but it reduces memory footprint a little so I decided to keep it. It's contained in a place with little other complexity, so I think it's ok.
* cast-machinery fest in `ValueComparison` is a thing! Having a hard time fixing that without `compareTo( byte other )` methods for all the primitives on `Value`. Dunno, maybe someone can improve on this in some later PR.
 * javadocs will come!

community/values/src/main/java/org/neo4j/values/LazyValues.java
```diff
@@ -0,0 +1,76 @@
...
+                    maybeValue = ((ValueLoader<?>) maybeValue).load();
+                    value.registerValue( (T) maybeValue );
+                }
+            }
+        }
+        //noinspection unchecked
+        return (T) maybeValue;
+    }
+
+    static boolean valueIsLoaded( Object maybeValue )
```
thobe, line 48: This method does not appear to be used.

community/values/src/main/java/org/neo4j/values/LazyValues.java
```diff
@@ -0,0 +1,76 @@
...
+//                }
+//            }
+//        }
+//        return value;
+//    }
+
+    /**
+     * Get or loads the value.
+     */
+    static <T> T getOrLoad( LazyValue<T> value )
```
thobe: Nice and simple, I like it.

community/values/src/main/java/org/neo4j/values/BooleanArray.java
```diff
@@ -0,0 +1,185 @@
...
+{
+    abstract boolean[] value();
+
+    @Override
+    public int length()
+    {
+        return value().length;
+    }
+
+    public boolean booleanValue( int offset )
```
thobe, line 36: Is this method really used? since it doesn't `@Override` anything, I suspect that it is neither used, nor needed, and in that case I think it should be removed.

community/values/src/main/java/org/neo4j/values/BooleanArray.java
```diff
@@ -0,0 +1,185 @@
...
+    @Override
+    public Object asPublic()
+    {
+        return value().clone();
+    }
+
+    @Override
+    public String toString()
+    {
+        return format( "BooleanArray(%s)", Arrays.toString( value() ) );
```
thobe: Will look like this: `BooleanArray([true, true, false])` isn't that double bracketing a bit much? I'm thinking it could be enough to have this be `BooleanArray[true, true, false]`.

Though I don't think this matters much, since it is not something that will reach the user.

fickludd:
On the Object field magic with the lazy values:

According to [jol](http://openjdk.java.net/projects/code-tools/jol/) this reduces memory footprint. On HotSpot 64-bit with compressed refs and 8-byte align, the Object field magic gives a 16 byte object, while Callable + ValueRef field gives 24 byte object. I think that is the reason it was introduces in the Property classes, and maybe that's a good reason to keep doing that.

community/values/src/main/java/org/neo4j/values/FloatingPointArray.java
```diff
@@ -0,0 +1,77 @@
...
+//    public final int hashCode()
+//    {
+//        int result = 1;
+//        for ( int i = 0, len = length(); i < len; i++ )
+//        {
+//            int elementHash = NumberValues.hash( longValue( i ) );
+//            result = 31 * result + elementHash;
+//        }
+//        return result;
+//    }
```
thobe: What is the point of all this commented out code?

community/values/src/main/java/org/neo4j/values/IntegralArray.java
```diff
@@ -0,0 +1,77 @@
...
+//    public final int hashCode()
+//    {
+//        int result = 1;
+//        for ( int i = 0, len = length(); i < len; i++ )
+//        {
+//            int elementHash = NumberValues.hash( longValue( i ) );
+//            result = 31 * result + elementHash;
+//        }
+//        return result;
+//    }
```
thobe: Why all this commented out code?

community/values/src/main/java/org/neo4j/values/LazyValues.java
thobe: ... so I think we should remove the "alternative solution" comment above.

community/values/src/main/java/org/neo4j/values/Values.java
```diff
@@ -0,0 +1,354 @@
...
+        T load() throws ValueLoadException;
+    }
+
+    class ValueLoadException extends RuntimeException
+    {
+    }
+
+    // DIRECT FACTORY METHODS
+
+    public static Value NO_VALUE = NoValue.NO_VALUE;
```
thobe: should this field not be final?

community/values/src/main/java/org/neo4j/values/NoValue.java
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.values;
+
+final class NoValue extends Value
+{
+    @SuppressWarnings( "WeakerAccess" )
+    public static NoValue NO_VALUE = new NoValue();
```
thobe: Should this field not be final? Why does it need to be public?

community/values/src/main/java/org/neo4j/values/NoValue.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private NoValue()
+    {
+    }
+
+    @Override
+    public boolean equals( Object other )
+    {
+        return false;
```
thobe: replace with `return this == other;`

community/values/src/main/java/org/neo4j/values/NoValue.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return 123456789;
```
thobe: why this hash code?
in order to avoid other people wondering that very same thing in the future, replace it with `System.identityHashCode( this )` (which is what the default `Object.hashCode()` implementation returns)

community/values/src/main/java/org/neo4j/values/NoValue.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public boolean equals( Value other )
+    {
+        return false;
```
thobe: replace with `return this == other;`

community/values/src/main/java/org/neo4j/values/StoreableValue.java
```diff
@@ -0,0 +1,24 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.values;
+
+abstract class StoreableValue extends Value
```
thobe: What purpose does this class serve?

community/values/src/main/java/org/neo4j/values/ValueTuple.java
```diff
@@ -0,0 +1,120 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.values;
+
+public class ValueTuple
```
thobe: What is the use case of this class?

community/values/src/main/java/org/neo4j/values/Values.java
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    public static Value shortArray( short[] value )
+    {
+        return new ShortArray.Direct( value );
+    }
+
+    // BOXED FACTORY METHODS
+
+    public static Value of( Object value )
```
thobe: how about making the if-tree in here smaller by checking for `value instanceof Number` (instead of all the individual number types) and delegate to the `numberValue` method (which should be moved to this _"BOXED FACTORY METHODS"_ section by the way)?

community/values/src/main/java/org/neo4j/values/Values.java
```diff
@@ -0,0 +1,354 @@
...
+        {
+            return byteValue( (Byte) value );
+        }
+        if ( value instanceof Character )
+        {
+            return charValue( (Character) value );
+        }
+        if ( value instanceof byte[] )
+        {
+            return byteArray( ((byte[]) value).clone() );
```
thobe, line 295: We should document the fact that this method makes defensive copies of arrays (and document for the specific methods that they _don't_ make defensive copies).

community/values/src/main/java/org/neo4j/values/ValueGroup.java
```diff
@@ -0,0 +1,45 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.values;
+
+public enum ValueGroup
```
thobe: How do we extend this to work for other value types as well? i.e. the types that we need in the Cypher runtime.

for example:
* `NodeReference`
* `Point`
* `ValueMap`

community/values/src/main/java/org/neo4j/values/Value.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.values;
+
+public abstract class Value
+{
+    @Override
+    public abstract boolean equals( Object other );
```
thobe, line 25: This method could probably be made final and implemented by delegation to the `equals( Value other )`-method. i.e. like this:

```
public final boolean equals( Object other ) {
    if ( this == other ) {
        return true;
    }
    return other instanceof Value && equals( (Value)other );
}
```

community/values/src/main/java/org/neo4j/values/Value.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public abstract boolean equals( boolean[] x );
+
+    public abstract boolean equals( char x );
+
+    public abstract boolean equals( String x );
+
+    public abstract boolean equals( char[] x );
+
+    public abstract boolean equals( String[] x );
```
thobe, line 54: These `equals`-methods are not going to have very interesting implementations for most sub-classes. In particular for "extension" value types - the ones we are going to implement in the Cypher runtime, i.e. `NodeReference`, `ValueMap`, `Point`, et.c.

How about giving them a default implementation that returns `false`?

Perhaps we should also consider "sealing" them, so that such "extension" value types _cannot_ override these methods. We could do that through a pattern like this:

```
public final boolean equals( String[] that ) {
    return eq( that );
}
boolean eq( String[] that ) {
    return false; // default implementation
}
```

thobe (changes requested):
Looks like it is headed in the right direction. I had a few comments.

thobe:
We talked about differentiating between defined values and `NoValue`, should we do that or are there good reasons why that is not done? I'd do that through a hierarchy like this:

```
    +-------------+
    | MaybeValue  |
    +-------------+
           ^
           |
     +-----+-----+
     |           |
+---------+  +-------+
| NoValue |  | Value |
+---------+  +-------+
```

community/values/src/main/java/org/neo4j/values/Values.java
```diff
@@ -291,6 +312,12 @@ public static Value of( Object value )
         {
             return NoValue.NO_VALUE;
         }
+        if ( value instanceof Value )
+        {
+            throw new UnsupportedOperationException(
+                    "Converting a Value to a Value using Values.of() is not supported." );
```
thobe: perhaps a matter of taste and style, but I would just put `Converting a Value to a Value.` in the message, because the fact that it is not supported is given by the exception type, and the fact that it is using `Values.of()` is given by the stacktrace.

community/values/src/main/java/org/neo4j/values/Value.java
fickludd: Yes I wanted that, but our checkstyle rules do not allow implementing hashCode (which we want to do in concrete classes) without equals. That why I did this more annoying solution. I would love to fix that but don't know how.

community/values/src/main/java/org/neo4j/values/LazyValues.java
fickludd: That because of a broken toString in the lazy values (which loaded the value). That is fixed now.

community/values/src/main/java/org/neo4j/values/BooleanArray.java
fickludd: This is present for symmetry with `IntergralArray`, `FloatingPointArray` and `TextArray`, and is used by `NumberValues.compareBooleanArrays()`. It does not override because `BooleanArray` happens to be the only representation of boolean array values, while the other 3 types have multiple representations.

community/values/src/main/java/org/neo4j/values/Value.java
lassewesth: @MishaDemianenko can show you how to disable checkstyle, for great justice!

community/values/src/main/java/org/neo4j/values/Value.java
fickludd: we have solved this by introducing a new class VirtualValue, which is intended for extension. This also solves comparison of with new values using the new ValueGroup VIRTUAL.

community/values/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+  <parent>
+    <groupId>org.neo4j</groupId>
+    <artifactId>parent</artifactId>
+    <version>3.3.0-SNAPSHOT</version>
+    <relativePath>../..</relativePath>
+  </parent>
+
+  <properties>
+    <short-name>values</short-name>
+    <bundle.namespace>org.neo4j.security</bundle.namespace>
```
pontusmelke: Not sure what this does but pretty sure that it is a copy-paste error?

pontusmelke (changes requested):

pontusmelke (approved):

community/values/src/main/java/org/neo4j/values/Value.java
thobe: In this case I think it would be a good idea to re-configure checkstyle to not do that particular check for this module.

PR #9415, opened 2017-05-24
Handle Single instance store with no transactions

mneedham:


dgordon5483:
@mneedham This issue was actually reported against 3.1.x. Can that be the base branch and then forward merge from there?

mneedham:
Gonna do this on 3.1 instead --> https://github.com/neo4j/neo4j/pull/9424

PR #9419, opened 2017-05-24
Checkpoint now correctly assert successful cleaning

burqen:
Only last commit is new.
Builds on #9388


community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -771,10 +762,10 @@ private void checkpoint( IOLimiter ioLimiter, Header.Writer headerWriter ) throw
 
         // Block writers, or if there's a current writer then wait for it to complete and then block
         // From this point and till the lock is released we know that the tree won't change.
-        long stamp = cleanerLock.writeLock();
-        writerLock.lock();
+        lock.writerAndCleanerLock();
         try
         {
+            assertRecoveryCleanSuccessful();
```
MishaDemianenko, line 768: it should be here because we need to do it under lock right?

MishaDemianenko: on another side, it looks like `cleaning != null` check inside check is redundant

MishaDemianenko (commented):
One small question for my understanding and small comment

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: Yes, it should be there because we need to check it after we take the lock.

We need to check for `null` because checkpoint is used in constructor when bumping generation before cleaner is initialised.

burqen:

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
MishaDemianenko: sorry, Anton, I can't find where we do checkpoint during construction. Would be great if you will guide me through there

MishaDemianenko (approved):

PR #9420, opened 2017-05-24
print environment variables usage as part of help

praveenag:
```
./bin/neo4j-admin help dump
usage: neo4j-admin dump [--database=<name>] --to=<destination-path>

environment variables:
    HEAP_SIZE     Set size of JVM heap during command execution.
                  Takes a number and a unit, for example 512m.

Dump a database into a single-file archive. The archive can be used by the load
command. <destination-path> can be a file or directory (in which case a file
called <database>.dump will be created). It is not possible to dump a database
that is mounted in a running Neo4j server.

options:
  --database=<name>         Name of database. [default:graph.db]
  --to=<destination-path>   Destination (file or folder) of database dump.
```

### DIFF
--- a/community/command-line/src/test/java/org/neo4j/commandline/admin/UsageTest.java
+++ b/community/command-line/src/test/java/org/neo4j/commandline/admin/UsageTest.java
@@ -19,15 +19,15 @@
  */
 package org.neo4j.commandline.admin;
 
+import java.util.Collections;
+import java.util.function.Consumer;
+import javax.annotation.Nonnull;
+
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.InOrder;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
-
-import java.util.Collections;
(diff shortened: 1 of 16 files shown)


community/command-line/src/test/java/org/neo4j/commandline/admin/UsageTest.java
```diff
@@ -50,17 +50,22 @@ public void setUp()
     public void shouldPrintUsageForACommand() throws Exception
     {
         // given
-        AdminCommand.Provider commandProvier = mockCommand( "bam", "A summary", AdminCommandSection.general() );
-        AdminCommand.Provider[] commands = new AdminCommand.Provider[]{commandProvier};
+        AdminCommand.Provider commandProvider = mockCommand( "bam", "A summary", AdminCommandSection.general() );
```
spacecowboy, line 53: good catch ;)

spacecowboy (approved):
A nice change.

PR #9432, opened 2017-05-27
IGNORE THIS PR: Half of importer rewritten for less GC and more performance

tinwelint:


systay:
Still interesting to keep around, or safe to close now?

systay:

PR #9437, opened 2017-05-29
Removed invalid and flaky test

pontusmelke:
We have a test that asserts that you should be able to create nodes in
user function by spawning threads and creating transaction. This is
unsupported behavior and shouldn't be tested. The test is also
occasionally flaky which gives us even more reasons to remove.


OliviaYtterbrink:
Replaced by https://github.com/neo4j/neo4j/pull/9447

PR #9444, opened 2017-05-30
Clean ups, as found by Intellij

chrisvest:


tinwelint (approved):
The mindless mechanical changes quickly turns the review into mindlessly mechanical too. Couldn't spot anything weird in there.

PR #9448, opened 2017-05-30
NativeSchemaNumberIndexPopulator

burqen:
First PR towards native schema number index, the populator. Implemented using GBPTree. Online IndexAccessor and SchemaIndexProvider will come later.

### Layout
Legend:
- Unique layout - All property values are unique
- Non-unique layout - Multiple entities can have the exact same property value
- `compareValue` - double representation of property value
- `rawValueBits` - actual bit representation of value
- `type` - value type (long, float, double) define how actualValue should be interpreted.
- `entityId` - id of entity (node) that has this property value.
With syntax `<indexKey : indexValue>` where `indexKey` is key in index guiding search and `indexValue` is value mapped by `indexKey`, they can both have multiple parts.
Unique layout: `<compareValue : type, rawValueBits, entityId>`
Non-unique layout: `<compareValue, entityId : type, rawValueBits>`. By storing entityId as part of key we make the keys unique, even if the indexed property value is not unique. This is necessary because the `GBPTree` does not support non-unique keys.

By storing `type` and `rawValueBits` in the index we can avoid an additional entity property lookup for filtering higher up in the stack.

### Composite index?
Right now, composite index is not supported.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/GBPTreeUtil.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/GBPTreeUtil.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.index;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.NoSuchFileException;
+
+import org.neo4j.index.internal.gbptree.GBPTree;
+import org.neo4j.io.fs.FileHandle;
+import org.neo4j.io.pagecache.PageCache;
+
+/**
+ * Utilities for common operations around a {@link GBPTree}.
+ */
+public class GBPTreeUtil
+{
+    /**
+     * Deletes store file backing a {@link GBPTree}.
+     *
+     * @param pageCache {@link PageCache} which manages the file.
+     * @param storeFile the {@link File} to delete.
+     * @throws NoSuchFileException if the {@code storeFile} doesn't exist according to the {@code pageCache}.
+     * @throws IOException on failure to delete existing {@code storeFile}.
+     */
+    public static void delete( PageCache pageCache, File storeFile ) throws IOException
+    {
+        FileHandle fileHandle = storeFileHandle( pageCache, storeFile );
+        fileHandle.delete();
+    }
+
+    /**
+     * Deletes store file backing a {@link GBPTree}, if it exists according to the {@code pageCache}.
+     *
+     * @param pageCache {@link PageCache} which manages the file.
+     * @param storeFile the {@link File} to delete.
+     * @throws IOException on failure to delete existing {@code storeFile}.
+     */
+    public static void deleteIfPresent( PageCache pageCache, File storeFile ) throws IOException
+    {
+        try
+        {
+            delete( pageCache, storeFile );
+        }
+        catch ( NoSuchFileException e )
+        {
+            // File does not exist, we don't need to delete
+        }
+    }
+
+    /**
+     * Checks whether or not {@code storeFile} exists according to {@code pageCache}.
+     *
+     * @param pageCache {@link PageCache} which manages the file.
+     * @param storeFile the {@link File} to check for existence.
+     * @return {@code true} if {@code storeFile} exists according to {@code pageCache}, otherwise {@code false}.
+     */
+    public static boolean storeFileExists( PageCache pageCache, File storeFile )
+    {
+        try
+        {
+            storeFileHandle( pageCache, storeFile );
+            return true;
+        }
+        catch ( IOException e )
+        {
+            return false;
+        }
+    }
+
+    private static FileHandle storeFileHandle( PageCache pageCache, File storeFile ) throws IOException
+    {
+        return pageCache.getCachedFileSystem().streamFilesRecursive( storeFile ).findFirst().get();
+    }
+}
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FailureHeaderWriter.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FailureHeaderWriter.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.index.schema;
+
+import java.util.Arrays;
+import java.util.function.Consumer;
+
+import org.neo4j.index.internal.gbptree.GBPTree;
+import org.neo4j.io.pagecache.PageCursor;
+
+/**
+ * Writes a failure message to a header in a {@link GBPTree}.
+ */
+class FailureHeaderWriter implements Consumer<PageCursor>
+{
+    private final byte[] failureBytes;
+
+    FailureHeaderWriter( byte[] failureBytes )
+    {
+        this.failureBytes = failureBytes;
+    }
+
+    @Override
+    public void accept( PageCursor cursor )
+    {
+        byte[] bytesToWrite = failureBytes;
+        cursor.putByte( NativeSchemaIndexPopulator.BYTE_FAILED );
+        int availableSpace = cursor.getCurrentPageSize() - cursor.getOffset();
+        if ( bytesToWrite.length + 2 > availableSpace )
+        {
+            bytesToWrite = Arrays.copyOf( bytesToWrite, availableSpace - 2 );
+        }
+        cursor.putShort( (short) bytesToWrite.length );
+        cursor.putBytes( bytesToWrite );
+    }
+}
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSampler.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSampler.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.index.schema;
+
+import java.io.IOException;
+
+import org.neo4j.cursor.RawCursor;
+import org.neo4j.index.internal.gbptree.GBPTree;
+import org.neo4j.index.internal.gbptree.Hit;
+import org.neo4j.index.internal.gbptree.Layout;
+import org.neo4j.kernel.impl.api.index.sampling.DefaultNonUniqueIndexSampler;
+import org.neo4j.kernel.impl.api.index.sampling.IndexSamplingConfig;
+import org.neo4j.kernel.impl.api.index.sampling.NonUniqueIndexSampler;
+import org.neo4j.storageengine.api.schema.IndexSample;
+
+/**
+ * {@link NonUniqueIndexSampler} which performs a full scans of a {@link GBPTree} in {@link #result()}.
+ *
+ * @param <KEY> type of keys in tree.
+ * @param <VALUE> type of values in tree.
+ */
+class FullScanNonUniqueIndexSampler<KEY extends SchemaNumberKey, VALUE extends SchemaNumberValue>
+        extends NonUniqueIndexSampler.Adapter
+{
+    private final GBPTree<KEY,VALUE> gbpTree;
+    private final Layout<KEY,VALUE> layout;
+    private final IndexSamplingConfig samplingConfig;
+
+    FullScanNonUniqueIndexSampler( GBPTree<KEY,VALUE> gbpTree, Layout<KEY,VALUE> layout,
+            IndexSamplingConfig samplingConfig )
+    {
+        this.gbpTree = gbpTree;
+        this.layout = layout;
+        this.samplingConfig = samplingConfig;
+    }
+
+    @Override
+    public IndexSample result()
+    {
+        KEY lowest = layout.newKey();
+        lowest.initAsLowest();
+        KEY highest = layout.newKey();
+        highest.initAsHighest();
+        try ( RawCursor<Hit<KEY,VALUE>,IOException> seek = gbpTree.seek( lowest, highest ) )
+        {
+            NonUniqueIndexSampler sampler = new DefaultNonUniqueIndexSampler( samplingConfig.sampleSizeLimit() );
+            while ( seek.next() )
+            {
+                Hit<KEY,VALUE> hit = seek.get();
+                sampler.include( hit.key().propertiesAsString() );
+            }
+            return sampler.result();
+        }
+        catch ( IOException e )
+        {
+            throw new RuntimeException( e );
+        }
+    }
+
+    @Override
+    public IndexSample result( int numDocs )
+    {
+        throw new UnsupportedOperationException();
+    }
+}
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
@@ -65,8 +65,9 @@ public abstract class NativeSchemaIndexPopulator<KEY extends SchemaNumberKey, VA
 
     private Writer<KEY,VALUE> singleWriter;
     private byte[] failureBytes;
+    private boolean dropped;
 
-    GBPTree<KEY,VALUE> gbpTree;
+    GBPTree<KEY,VALUE> tree;
 
     NativeSchemaIndexPopulator( PageCache pageCache, File storeFile, Layout<KEY,VALUE> layout,
             RecoveryCleanupWorkCollector recoveryCleanupWorkCollector )
@@ -81,26 +82,38 @@ public abstract class NativeSchemaIndexPopulator<KEY extends SchemaNumberKey, VA
     }
 
     @Override
-    public void create() throws IOException
+    public synchronized void create() throws IOException
     {
         GBPTreeUtil.deleteIfPresent( pageCache, storeFile );
-        gbpTree = new GBPTree<>( pageCache, storeFile, layout, 0, NO_MONITOR, NO_HEADER,
-                recoveryCleanupWorkCollector );
+        instantiateTree();
         instantiateWriter();
     }
 
-    protected void instantiateWriter() throws IOException
+    private void instantiateTree() throws IOException
+    {
+        tree = new GBPTree<>( pageCache, storeFile, layout, 0, NO_MONITOR, NO_HEADER,
+                recoveryCleanupWorkCollector );
+    }
+
+    void instantiateWriter() throws IOException
     {
         assert singleWriter == null;
-        singleWriter = gbpTree.writer();
+        singleWriter = tree.writer();
     }
 
     @Override
-    public void drop() throws IOException
+    public synchronized void drop() throws IOException
     {
-        closeWriter();
-        gbpTree = closeIfPresent( gbpTree );
-        GBPTreeUtil.deleteIfPresent( pageCache, storeFile );
+        try
+        {
+            closeWriter();
+            closeTree();
+            GBPTreeUtil.deleteIfPresent( pageCache, storeFile );
+        }
+        finally
+        {
+            dropped = true;
+        }
     }
 
     @Override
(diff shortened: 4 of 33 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/UniqueSchemaNumberKey.java
```diff
@@ -0,0 +1,70 @@
...
+    {
+        assertValidSingleNumberPropertyValue( values );
+        value = ((Number) values[0]).doubleValue();
+        isHighest = false;
+    }
+
+    @Override
+    public String propertiesAsString()
+    {
+        throw new UnsupportedOperationException( "Implement me" );
```
tinwelint, line 48: Perhaps `propertiesAsString` shouldn't live in `SchemaNumberKey` interface, but rather in `NonUniqueSchemaNumberKey`? Although we don't know which other, when there are composite keys, will wanna make use of it.

burqen: Let's keep it where it is, at least for this PR and then move it if we find the need later. Those classes will likely see some refactoring.

tinwelint: Yup, they have indeed already seen a lot of refactorings after this PR

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/SchemaNumberKey.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.index.schema;
+
+/**
+ * Key in a native schema index for numbers.
+ */
+interface SchemaNumberKey
```
tinwelint, line 25: Note: The unique and non-unique versions have been merged into a single one in future work

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NonUniqueSchemaNumberIndexLayout.java
```diff
@@ -0,0 +1,117 @@
...
+ */
+package org.neo4j.kernel.impl.index.schema;
+
+import org.neo4j.index.internal.gbptree.Layout;
+import org.neo4j.io.pagecache.PageCursor;
+
+/**
+ * {@link Layout} for numbers where numbers doesn't need to be unique.
+ */
+public class NonUniqueSchemaNumberIndexLayout implements Layout<NonUniqueSchemaNumberKey,NonUniqueSchemaNumberValue>
```
tinwelint, line 28: Note: The unique and non-unique versions have been merged into a single one in future work, only `compare` will differ.

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexPopulator.java
```diff
@@ -104,18 +104,15 @@ void add( IndexEntryUpdate<?> update )
...
+     * false}.
      *
-     * @param failure the description of the failure.
-     * @throws IOException if marking failed.
+     * @param populationCompletedSuccessfully true if index should be marked as {@link InternalIndexState#ONLINE},
+     * otherwise false.
+     * @param failureMessage Description of population failure to store if populationCompletedSuccessfully is false.
      */
-    void markAsFailed( String failure ) throws IOException;
+    void close( boolean populationCompletedSuccessfully, String failureMessage ) throws IOException;
```
MishaDemianenko: I think method responsibilities currently are overloaded: it close index in case of successful and failed population, and sets error message - looks too much.
`populator.close( true, null )` - reads and looks weird.
Would propose to change it to 'complete() or close()' and `fail(mesasge)/failure(message)`

community/kernel/src/main/java/org/neo4j/kernel/impl/index/GBPTreeUtil.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Deletes store file backing a {@link GBPTree}, if it exists according to the {@code pageCache}.
+     *
+     * @param pageCache {@link PageCache} which manages the file.
+     * @param storeFile the {@link File} to delete.
+     * @throws IOException on failure to delete existing {@code storeFile}.
+     */
+    public static void deleteIfPresent( PageCache pageCache, File storeFile ) throws IOException
```
MishaDemianenko, line 56: looks strange that this method still throws IOException? if its fine - why to have both `delete` and `deleteIfPresent` look a bit confusing now

community/kernel/src/main/java/org/neo4j/kernel/impl/index/GBPTreeUtil.java
```diff
@@ -0,0 +1,92 @@
...
+            storeFileHandle( pageCache, storeFile );
+            return true;
+        }
+        catch ( IOException e )
+        {
+            return false;
+        }
+    }
+
+    private static FileHandle storeFileHandle( PageCache pageCache, File storeFile ) throws IOException
```
MishaDemianenko, line 88: this method should return `Optional<FileHandle>` instead, and then we will be able to safely check if it presents instead of capture exceptions.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FailureHeaderWriter.java
```diff
@@ -0,0 +1,53 @@
...
+        this.failureBytes = failureBytes;
+    }
+
+    @Override
+    public void accept( PageCursor cursor )
+    {
+        byte[] bytesToWrite = failureBytes;
+        cursor.putByte( NativeSchemaIndexPopulator.BYTE_FAILED );
+        int availableSpace = cursor.getCurrentPageSize() - cursor.getOffset();
+        if ( bytesToWrite.length + 2 > availableSpace )
```
MishaDemianenko, line 46: magic deuce

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSampler.java
```diff
@@ -0,0 +1,82 @@
...
+            while ( seek.next() )
+            {
+                Hit<KEY,VALUE> hit = seek.get();
+                sampler.include( hit.key().propertiesAsString() );
+            }
+            return sampler.result();
+        }
+        catch ( IOException e )
+        {
+            throw new RuntimeException( e );
```
MishaDemianenko, line 73: would propose to throw some kind of sampling exception instead

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSampler.java
```diff
@@ -0,0 +1,82 @@
...
+        lowest.initAsLowest();
+        KEY highest = layout.newKey();
+        highest.initAsHighest();
+        try ( RawCursor<Hit<KEY,VALUE>,IOException> seek = gbpTree.seek( lowest, highest ) )
+        {
+            NonUniqueIndexSampler sampler = new DefaultNonUniqueIndexSampler( samplingConfig.sampleSizeLimit() );
+            while ( seek.next() )
+            {
+                Hit<KEY,VALUE> hit = seek.get();
+                sampler.include( hit.key().propertiesAsString() );
```
MishaDemianenko, line 67: it looks like we potentially can avoid conversion to string here, in the same time it's not that important maybe

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
```diff
@@ -0,0 +1,260 @@
...
+        this.storeFile = storeFile;
+        this.layout = layout;
+        this.treeKey = layout.newKey();
+        this.treeValue = layout.newValue();
+        this.recoveryCleanupWorkCollector = recoveryCleanupWorkCollector;
+        this.conflictDetectingValueMerger = new ConflictDetectingValueMerger<>();
+    }
+
+    @Override
+    public synchronized void create() throws IOException
```
MishaDemianenko, line 85: why do we have `synchronized` methods here?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
```diff
@@ -0,0 +1,260 @@
...
+    public void add( IndexEntryUpdate<?> update ) throws IndexEntryConflictException, IOException
+    {
+        treeKey.from( update.getEntityId(), update.values() );
+        treeValue.from( update.getEntityId(), update.values() );
+        singleWriter.merge( treeKey, treeValue, conflictDetectingValueMerger );
+        if ( conflictDetectingValueMerger.wasConflict() )
+        {
+            long existingNodeId = conflictDetectingValueMerger.existingNodeId();
+            long addedNodeId = conflictDetectingValueMerger.addedNodeId();
+            // TODO: not sure about the OrderedPropertyValues#ofUndefined bit
```
MishaDemianenko, line 138: TODO

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
```diff
@@ -0,0 +1,260 @@
...
+        if ( tree == null )
+        {
+            throw new IllegalStateException( "Populator has already been closed." );
+        }
+    }
+
+    private void markTreeAsFailed( String failureMessage ) throws IOException
+    {
+        byte[] failureBytes =
+                failureMessage != null ? failureMessage.getBytes( StandardCharsets.UTF_8 ) : "".getBytes();
```
MishaDemianenko: `ArrayUtils.EMPTY_BYTE_ARRAY` or similar should be used instead

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NonUniqueNativeSchemaIndexPopulator.java
```diff
@@ -0,0 +1,109 @@
...
+        this.samplingConfig = samplingConfig;
+    }
+
+    @Override
+    public void includeSample( IndexEntryUpdate update )
+    {
+        if ( updateSampling )
+        {
+            checkSampler();
+            sampler.include( SamplingUtil.encodedStringValuesForSampling( update.values() ) );
```
MishaDemianenko, line 57: `StringUtils.join( update.values(), "\u001F" )` can help here instead of SamplingUtil so far

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NonUniqueNativeSchemaIndexPopulator.java
```diff
@@ -0,0 +1,109 @@
...
+        checkSampler();
+
+        // Close the writer before scanning
+        try
+        {
+            closeWriter();
+        }
+        catch ( IOException e )
+        {
+            throw new RuntimeException( e );
```
MishaDemianenko, line 81: runtime exceptions while trying to open/close writers

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NonUniqueNativeSchemaIndexPopulator.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        try
+        {
+            return sampler.result();
+        }
+        finally
+        {
+            // Start the writer again (TODO may be unnecessary)
```
MishaDemianenko, line 90: TODO

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/UniqueSchemaNumberIndexLayout.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    public int minorVersion()
+    {
+        return 1;
+    }
+
+    @Override
+    public int compare( UniqueSchemaNumberKey o1, UniqueSchemaNumberKey o2 )
```
MishaDemianenko, line 111: This comparison should be on pair with ordering used in cypher right?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/SchemaNumberValueConversion.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.neo4j.kernel.impl.index.schema.SchemaNumberValue.LONG;
+
+/**
+ * Utilities for converting number values to and from different representations.
+ */
+class SchemaNumberValueConversion
+{
+    static void assertValidSingleNumberPropertyValue( Object[] values )
+    {
+        // TODO: support multiple values, right?
```
MishaDemianenko, line 33: todo, maybe exception for now here?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/UniqueSchemaNumberValue.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Adds entity id because its unique key counterpart doesn't have it. The gain of having entity id in value
+ * is that keys in internal tree nodes becomes smaller so that internal tree nodes can contain more keys.
+ */
+class UniqueSchemaNumberValue extends SchemaNumberValue
+{
+    static final int SIZE =
+            Byte.SIZE + /* type */
+            Long.SIZE + /* value bits */
+            Long.SIZE;  /* entity id TODO 7 bytes could be enough. Also combine with the type byte thing*/
```
MishaDemianenko, line 36: todo

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSamplerTest.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private List<Number> generateNumberValues()
+    {
+        List<Number> result = new ArrayList<>();
+        for ( IndexEntryUpdate<?> update : NativeSchemaIndexPopulatorTest.someDuplicateIndexEntryUpdates() )
+        {
+            result.add( (Number) update.values()[0] );
+        }
+        // TODO: perhaps some more values?
```
MishaDemianenko, line 108: todo's in this class

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSamplerTest.java
```diff
@@ -0,0 +1,140 @@
...
+        // THEN
+        assertEquals( values.size(), sample.sampleSize() );
+        assertEquals( countUniqueValues( values ), sample.uniqueValues() );
+        assertEquals( values.size(), sample.indexSize() );
+    }
+
+    static int countUniqueValues( List<Number> values )
+    {
+        int count = 0;
+        Set<Double> seenValues = new HashSet<>();
```
MishaDemianenko, line 90: `values.stream().map( Number::doubleValue ).collect( Collectors.toSet() ).size()` if you really need doubleValues
otherwise just add all values to set and get size of it

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSamplerTest.java
```diff
@@ -0,0 +1,140 @@
...
+                count++;
+            }
+        }
+        return count;
+    }
+
+    private List<Number> generateNumberValues()
+    {
+        List<Number> result = new ArrayList<>();
+        for ( IndexEntryUpdate<?> update : NativeSchemaIndexPopulatorTest.someDuplicateIndexEntryUpdates() )
```
MishaDemianenko, line 104: test should not use another test to generate data

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulatorTest.java
```diff
@@ -0,0 +1,900 @@
...
+                {
+                    assertEquals( failureMessage, headerReader.failureMessage );
+                }
+            }
+        }
+    }
+
+    private String longString( int length )
+    {
+        String alphabet = "123xyz";
```
MishaDemianenko, line 684: use `RandomStringUtils` instead?

MishaDemianenko (commented):
Looks good was surprised how good it looks in general, thx guys!
Still, have some comments, mention observed TODOs mostly as a reminder in case you forgot them.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/GBPTreeUtil.java
tinwelint: It could still fail on actually deleting the file. It's good to have it strict in some cases, others not (existence-wise).

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexPopulator.java
tinwelint: That's how it was, and we thought it was confusing. Previously `markAsFailed` was always followed by `close( false )` and `close( true )` was run on its own. It was annoying to implement. Not proud of how it looks but according to me it feels better now.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/GBPTreeUtil.java
tinwelint: Not quite since the page cache method will throw `NoSuchFileException` if passing in a _file_ `File` which doesn't exist anyway.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FailureHeaderWriter.java
tinwelint: it's the _short_ length. Could be made more clear yeah. This have now been fixed in future work instead.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSampler.java
tinwelint: This is exactly what lucene index sampler does, we wanted to fit in with the current sampling classes available to reduce number of things invented here.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSampler.java
tinwelint: This is what lucene does too and I don't think there are sampling exceptions around. With that said, it could be `UncheckedIOException` instead. Failures thrown here are only propagated and potentially logged since they are run by `JobScheduler` so there's typically no user calling this method expected to catch specific sampling exception and handle it.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
tinwelint: We got into scenarios around fields and volatility since the populator is managed by two threads, life cycle wise. It was a simplistic approach to synchronize all lifecycle methods, methods which are called once.

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSamplerTest.java
tinwelint: This has been improved in future work

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/UniqueSchemaNumberIndexLayout.java
tinwelint: You mean lucene number index ordering? From what we could tell by a brief look Lucene doesn't sort the results and so this ordering is what all numbers coerced into double is. Since Lucene schema index doesn't have any specific ordering then this ordering cannot possibly break anything.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/UniqueSchemaNumberValue.java
tinwelint: There have been refactorings in this area in future work, let's keep this TODO until later.

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSamplerTest.java
tinwelint: Optional niceyfication

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexPopulator.java
tinwelint: @MishaDemianenko although there are other designs that we could do instead:

- `closeAsSuccessful()` / `closeAsFailed( String failure )`
- `markAsFailed( String failure )` / `close()` <-- no success arg

are any of those better?

tinwelint: Or should this refactoring simply be reverted?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
tinwelint: This TODO has been deleted/fixed in future work already

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
tinwelint: Has now been fixed in future work

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NonUniqueNativeSchemaIndexPopulator.java
tinwelint: Took advantage of this in `SamplingUtil` in future work

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NonUniqueNativeSchemaIndexPopulator.java
tinwelint: It's still a TODO although with better description now. This is what it now says:

```
// Start the writer again (TODO may be unnecessary since this happens after all population is completed,
// however would be weird to have this sampling method close the writer and tree and keep them closed.
// Time will tell whether or not this could be avoided.
```

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/SchemaNumberValueConversion.java
tinwelint: Yes, that's what the line(s) below does. It's a TODO because we know that we need to support composite keys soon and simply a reminder to do so here.

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSamplerTest.java
tinwelint: Removed TODOs since they were not relevant anymore, e.g.

`// TODO: shouldIncludeHighestAndLowestPossibleNumberValues`

the test already tested that (the sample data includes those values).

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulatorTest.java
tinwelint: Have changed to do so in future work.

tinwelint:
@MishaDemianenko Thanks for good feedback. I believe all comments have been addressed, except the `markAsFailed`/`close` conundrum. I'm OK with reverting that commit and doing that work separately instead if you think that's better.

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexPopulator.java
MishaDemianenko: Yeah, I think would be better for this refactoring to go separately.
`markAsFailed( String failure ) / close()` I think is good, or even shorter version `fail( String failure ) / close()`

community/kernel/src/main/java/org/neo4j/kernel/impl/index/GBPTreeUtil.java
MishaDemianenko: ah ok

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FullScanNonUniqueIndexSampler.java
MishaDemianenko: ok, let's make it UncheckedIOException then

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/UniqueSchemaNumberIndexLayout.java
MishaDemianenko: We have `PropertyValueComparison.SyperType.Number` for a range of numbers from negative infinity to Double.NaN as max value and also `NumberPropertyValueComparator` that are used for number comparison. And new index comparison should match to that I believe, especially special cases.

MishaDemianenko:
@tinwelint agree with everything mostly, added an additional note about UniqueSchemaNumberIndexLayout i think those comparisons should match

MishaDemianenko (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
```diff
@@ -0,0 +1,273 @@
...
+            {
+                throw new IllegalStateException( "Index updater has been closed." );
+            }
+            add( update );
+        }
+
+        @Override
+        public void remove( PrimitiveLongSet nodeIds ) throws IOException
+        {
+            throw new UnsupportedOperationException( "Implement me" );
```
chrisvest, line 264: This will be implemented in future work?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/SchemaNumberValueConversion.java
```diff
@@ -0,0 +1,63 @@
...
+            throw new IllegalArgumentException( "Tried to create composite key with non-composite schema key layout" );
+        }
+        if ( values.length < 1 )
+        {
+            throw new IllegalArgumentException( "Tried to create key without value" );
+        }
+        if ( !(values[0] instanceof Number) )
+        {
+            throw new IllegalArgumentException(
+                    "Key layout does only support numbers, tried to create key from " + values[0] );
```
chrisvest, line 45: I'd add the class of the value (if non-null value) to the exception message.

chrisvest (approved):
A few comments.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexPopulator.java
tinwelint: It will most likely not be needed. That very method exists due to recovery issues in Lucene. The `GBPTree` recovers itself together with normal recovery w/o any additional logic. This will be noted in this method as comments or javadoc.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/SchemaNumberValueConversion.java
tinwelint: I'll add that in the next PR, where all this have already been changed.

PR #9449, opened 2017-05-30
Casual Clustering id reuse

klaren:


klaren:
Can you guys please take a look at this @digitalstain @martinfurmanski ?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/IdReusabilityCondition.java
```diff
@@ -0,0 +1,113 @@
...
+        private final long idHighestWhenBecameLeader;
+
+        private volatile boolean oldTransactionsApplied;
+
+        LeaderIdReusabilityCondition( TransactionIdStore transactionIdStore )
+        {
+            this.transactionIdStore = transactionIdStore;
+
+            // Get highest transaction id seen
+            this.idHighestWhenBecameLeader = transactionIdStore.getLastCommittedTransactionId();
```
martinfurmanski: At this point there could be another member thinking it is the leader as well which is happily freeing IDs > transactionIdStore.getLastCommittedTransactionId() of this local instance. Leadership is eventually consistent state whilst lastCommittedTransactionId is a projection of the asynchronous transaction application process (ordered after but asynchronous to the commit path). The best and easiest way to get correct behaviour here is to order the transaction and resulting free IDs in the consensus log relative something else also in the consensus log.

klaren: Since the transactions are in the consensus log and the new leader uses them to prove it has the longest log, the highest seen transactions IDs should correlate with a leader switch. If there is any contention about the leadership the transactions belonging to older terms will not be accepted since only leaders are allowed to append new transactions. 

I don't see how

> thinking it is the leader

is enough to get new transactions committed in the consensus log or how you will accept new transaction from another leader during this uncertain state.

chrisvest: @martinfurmanski Are you still seeing a problem here, or is it good?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/ReplicatedIdGenerator.java
```diff
@@ -34,27 +39,38 @@
...
+        idContainer.close( highId );
     }
 
     @Override
     public void freeId( long id )
     {
+        if ( freeIdCondition.getAsBoolean() )
+        {
+            idContainer.freeId( id );
+        }
```
chrisvest: Do we have a feel for the probability of ending up here, and the condition is false?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdGeneratorImpl.java
```diff
@@ -64,31 +60,15 @@
...
-
-    private final File file;
-    private final FileSystemAbstraction fs;
-    private StoreChannel fileChannel;
-
     private final long max;
-    private final boolean aggressiveReuse;
-    private FreeIdKeeper keeper;
+    private final IdContainer idContainer;
+    private long highId = INTEGER_MINUS_ONE;
```
chrisvest: No need to initialise it to this strange constant because the constructor will overwrite it.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/IdReusabilityCondition.java
MishaDemianenko: we will have a talk with guys next week when Chris will return from vacation, maybe we will need to get highest transaction id from raft log instead

chrisvest (commented):
Had one more question.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/ReplicatedIdGenerator.java
MishaDemianenko: not sure, I fully understand your question. But in case I am: for period when new master catching up with transaction which already in the raft log but not applied yet

MishaDemianenko:
@martinfurmanski @digitalstain guys, can you please take a look into updated version that based on command ids instead, thx!

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/ClusterIdReuseIT.java
```diff
@@ -0,0 +1,178 @@
...
+    {
+        CoreClusterMember newLeader = cluster.awaitLeader();
+        IdController idController = newLeader.database().getDependencyResolver().resolveDependency( IdController.class );
+        idController.maintenance();
+        return idController;
+    }
+
+    private void removeTwoNodes( Cluster cluster, MutableLong first, MutableLong second ) throws Exception
+    {
+        cluster.coreTx( ( db, tx ) ->
```
martinfurmanski: You can extract the member on which this transaction was performed for use later on in the test. This is useful to protect against flaky behaviour. If you try to rely on leadership being stable, then you can and will experience flakiness. This comment applies to any place where you assume that the roles will not change.

An ill-timed leader-switch could also make it so that IDs are actually not reused by anyone, so asserting on that they are will be flaky by itself then. Imagine for example the leader stepping down due to GC right after the transaction got committed in consensus-land. It will eventually be applied but there will be no leader or a new (or the same) leader with a higher lastLogIndexBeforeWeBecameLeader.

A robust assertion that you could make is that IDs are reused on at most one member, but perhaps that is not interesting enough.

A way to make it more robust would be to run some sequence many times, e.g. many delete transactions, and assert that at least some percentage was in fact reused and on the expected members.

martinfurmanski (commented):
I have mostly looked at the causal clustering parts.

MishaDemianenko:
@chrisvest @martinfurmanski @digitalstain guys you are welcome to take another round of review!

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/EditionModule.java
```diff
@@ -68,6 +74,9 @@
  */
 public abstract class EditionModule
 {
+    protected static final boolean safeIdBuffering = FeatureToggles.flag(
+            EditionModule.class, "safeIdBuffering", true );
```
chrisvest: Do we still need the feature toggle?

chrisvest: Hmm… maybe still good to have in case there's a field fire.

chrisvest: In that case maybe add a comment that this toggle was placed in RecordStorageEngine in versions prior to 3.3.

chrisvest:
Not related to these changes, but one thing that's curious is that relationship group records are not aggressively reused in Enterprise Edition, according to `EnterpriseEditionSettings.idTypesToReuse`, and `CommunityIdTypeConfigurationProvider`.

MishaDemianenko:
@digitalstain @martinfurmanski do you guys have anything to add or we good to go?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/IdReusabilityCondition.java
```diff
@@ -0,0 +1,103 @@
...
+                return true;
+            }
+
+            boolean hasAppliedOldTransactions = commandIndexTracker.getAppliedCommandIndex() > commandIdWhenBecameLeader;
+            if ( hasAppliedOldTransactions )
+            {
+                oldTransactionsApplied = true;
+            }
+
+            return hasAppliedOldTransactions;
```
chrisvest: This code can be simplified. Would be nice with a comment describing the deeper meaning behind this logic.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdContainer.java
```diff
@@ -0,0 +1,283 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+import org.neo4j.io.fs.FileSystemAbstraction;
+import org.neo4j.io.fs.StoreChannel;
+import org.neo4j.kernel.impl.store.InvalidIdGeneratorException;
+import org.neo4j.kernel.impl.store.UnderlyingStorageException;
+
+
+public class IdContainer
```
chrisvest: Would be nice with some javadocs on this. For instance, it doesn't look thread-safe, so that might be a good thing to note.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/ReplicatedIdGenerator.java
```diff
@@ -35,26 +39,31 @@
...
     @Override
     public void close()
     {
+        idContainer.close( highId );
     }
 
     @Override
     public void freeId( long id )
     {
+        idContainer.freeId( id );
```
chrisvest: `IdGeneratorImpl` protects its `IdContainer` by making its methods `synchronized`. This class does not do that, so does this not expose the `IdContainer` to concurrency that it is not prepared to handle?

chrisvest (commented):
I'm not that familiar with the causal clustering stuff. I had a few comments.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/ReplicatedIdGenerator.java
klaren: Nice catch! BufferedIdGenerator runs in a separate thread when freeing IDs so this is definitely a problem.

klaren:
@chrisvest Done with changes. Thanks for the feedback!

PR #9460, opened 2017-06-01
3.3 graph values applied

fickludd:
Apply the new Values module to the kernel.

With this work we remove the *Property classes and replace them with their Value equivalents. Also places where property values were typed as Object have been changed to Value, with exceptions in some surfaces and tools close to the surface, and tests where that would just introduce a lot of boilerplate and reduce readability. 

Some places could be improved, like `PropertyStore.encodeValue`, but we felt too uncertain on this part of the kernel to make bigger changes.

[Core API benchmarks](http://jmh-benchmarks.neohq.net/?tab=compareVersions&benchmarkGroup=Core%2520API&neo4jVersion=3.3.0&neo4jVersion=fickludd%25203.3-graph-values-applied&date=10-5-2017) show identical performance as far as I can tell.


fickludd:
The failing test (`BoltFailuresIT.throwsWhenInitMessageProcessingFailsToComplete`) looks flaky, it has bad history and passes locally. We will rerun it when we address PR review.

community/consistency-check/src/main/java/org/neo4j/consistency/checking/PropertyRecordCheck.java
```diff
@@ -82,7 +82,7 @@ public static void checkDataBlock( PropertyBlock block,
             default:
                 try
                 {
-                    type.getValue( block, null );
+                    type.getValueNow( block, null );
```
MishaDemianenko: I would prefer if we would call those methods `getValue` and `getLazyValue` or even just `value` and `lazyValue` instead

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v2_3/TransactionBoundQueryContext.scala
```diff
@@ -307,25 +308,25 @@ final class TransactionBoundQueryContext(tc: TransactionalContextWrapper)
     }
 
     def getProperty(id: Long, propertyKeyId: Int): Any = try {
-      tc.statement.readOperations().nodeGetProperty(id, propertyKeyId)
+      tc.statement.readOperations().nodeGetProperty(id, propertyKeyId).asLegacyObject()
```
MishaDemianenko: why do we call a method that returns value `asLegacyObject` and method that copy value `asPublic` shouldn't we make it more logical and simply call them `value` and `copyValue`?

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
```diff
@@ -150,8 +154,55 @@ private String propertyString( TokenNameLookup tokenNameLookup, int[] propertyId
             sb.append( '`' );
             sb.append( tokenNameLookup.propertyKeyGetName( propertyIds[i] ) );
             sb.append( "` = " );
-            sb.append( OrderedPropertyValues.quote( propertyValues.valueAt( i ) ) );
+            sb.append( quote( propertyValues.valueAt( i ).asPublic() ) );
         }
         return sb.toString();
     }
+
+    private static String quote( Object propertyValue )
```
MishaDemianenko: I think this will make more sense if this is moved to Values for example or be splitter to each value as part of its `toString()` so we do not need quote method here at all, wdyt?

community/values/src/main/java/org/neo4j/values/ValueTuple.java
```diff
@@ -31,6 +33,18 @@ public static ValueTuple of( Value... values )
...
+        for ( int i = 0; i < values.length; i++ )
+        {
+            values[i] = Values.of( objects[i] );
+        }
+        return new ValueTuple( values );
+    }
+
     private final Value[] values;
 
     private ValueTuple( Value[] values )
```
MishaDemianenko, line 50: it looks like `getOnlyValue()` here is more like `getFirstValue` right?

community/kernel/src/main/java/org/neo4j/kernel/api/index/ArrayEncoder.java
```diff
@@ -33,6 +36,7 @@ private ArrayEncoder()
         throw new AssertionError( "Not for instantiation!" );
     }
 
+    @Deprecated
```
MishaDemianenko: why not remove this method completely instead? it's used only in tests?

community/kernel/src/main/java/org/neo4j/kernel/api/properties/PropertyKeyValue.java
```diff
@@ -0,0 +1,98 @@
...
+    {
+        return value.equals( other );
+    }
+
+    @Override
+    public Value value() throws PropertyNotFoundException
+    {
+        if ( value == Values.NO_VALUE )
+        {
+            // TODO
```
MishaDemianenko: TODO

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/PropertyContainerStateImpl.java
```diff
@@ -222,18 +211,33 @@ public boolean hasChanges()
...
     public boolean isPropertyRemoved( int propertyKeyId )
     {
         return removedProperties != null && removedProperties.containsKey( propertyKeyId );
     }
+
+    private Iterator<StorageProperty> toPropertyIterator( VersionedHashMap<Integer,Value> propertyMap )
+    {
+        return propertyMap == null ? Collections.emptyIterator() :
+                Iterators.map(
+                    entry -> new PropertyKeyValue( entry.getKey(), entry.getValue() ),
```
MishaDemianenko, line 233: this increase amount of things that we create for each iteration for each node/property state isn't it?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyStore.java
```diff
@@ -217,17 +218,19 @@ public static void allocateArrayRecords( Collection<DynamicRecord> target, Objec
...
     {
         encodeValue( block, keyId, value, stringStore, arrayStore );
     }
 
-    public static void encodeValue( PropertyBlock block, int keyId, Object value,
+    public static void encodeValue( PropertyBlock block, int keyId, Value value,
             DynamicRecordAllocator stringAllocator, DynamicRecordAllocator arrayAllocator )
     {
-        if ( value instanceof String )
+        // TODO: use ValueWriter
```
MishaDemianenko, line 229: TODO

community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyType.java
```diff
@@ -215,19 +192,16 @@ public String getValue( PropertyBlock block, PropertyStore store )
     ARRAY( 10 )
     {
         @Override
-        public DefinedProperty readProperty( int propertyKeyId, final PropertyBlock block, final Supplier<PropertyStore> store )
+        public Value getValue( final PropertyBlock block, final PropertyStore store )
         {
-            return Property.lazyArrayProperty(propertyKeyId, () -> getValue( block, store.get() ) );
+            // TODO: make me lazy again?
```
MishaDemianenko: yep that would be nice :)

community/kernel/src/main/java/org/neo4j/storageengine/api/StorageProperty.java
```diff
@@ -37,26 +38,18 @@
...
      */
-    boolean valueEquals( Object other );
+    boolean valueEquals( Value other );
 
     /**
      * @return the property value.
      * @throws PropertyNotFoundException if this property instance represented a non-existent property.
      */
-    Object value() throws PropertyNotFoundException;
+    Value value() throws PropertyNotFoundException;
```
MishaDemianenko: why do we need to keep this value? it looks like it used only in tests, would propose to have a new `value()` that does not throw an exception and do what `valueForced` do now and remove `valueForced` completely

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/ValueEncoding.java
```diff
@@ -114,28 +116,28 @@ String key()
...
-                    return stringField( name, value.toString() );
+                    return stringField( name, value.asLegacyObject().toString() );
                 }
 
                 @Override
-                void setFieldValue( Object value, Field field )
+                void setFieldValue( Value value, Field field )
                 {
-                    field.setStringValue( value.toString() );
+                    field.setStringValue( value.asLegacyObject().toString() );
```
MishaDemianenko: do we really need to go through legacy object in it's usages here?

community/values/src/main/java/org/neo4j/values/Values.java
```diff
@@ -37,11 +40,16 @@
...
+    public static final Value MAX_NUMBER = Values.doubleValue( Double.NaN );
+    public static final Value MIN_STRING = Values.stringValue( "" );
+    public static final Value MAX_STRING = Values.booleanValue( false );
+
     private Values()
     {
     }
 
-    interface ValueLoader<T>
+    public interface ValueLoader<T>
```
MishaDemianenko: would be really nice to this class and `ValueLoadException` in their own classes isntead

MishaDemianenko (changes requested):
A couple of comments, otherwise impressive amount of work!
Also, it looks like we will create a bit of more Value objects then we did with property objects before, but if benchmarks are fine with that i do not see any problems with that

MishaDemianenko:
@burqen would be nice if you will take a look into proposed changes in regards to indexing and how type of value is checked now, just to keep you in loop or change something now if you think that would be useful

community/values/src/main/java/org/neo4j/values/ValueTuple.java
fickludd: Not really... I added `getOnlyValue` only to make life easier in the parts of the code base that still do not support composite indexes. By calling that method we are assuming the tuple only contains one value, which is checked by `assert values.length == 1`. Calling the method `getFirstValue` would communicate that it should work even for cases when there's more than 1 value in the tuple, which it shouldn't. In these cases we should surely consider all the values!

Or have I misunderstood you?

community/kernel/src/main/java/org/neo4j/kernel/api/properties/PropertyKeyValue.java
fickludd: Can't remember what I wanted to do. Removed the whole method instead :).

community/kernel/src/main/java/org/neo4j/kernel/api/index/ArrayEncoder.java
fickludd: You're right. Did that.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/PropertyContainerStateImpl.java
fickludd: Yes it does. Honestly the whole `PropertyKeyValue` is a little subpar considering the hotness of that path. However, we are about to go over cursors anyway, so I didn't want to over work this now. Introduced `PropertyKeyValue` only to make the current cursors and iterators work as before.

I suggest we put reducing object allocation during property iteration into a separate card and revisit later in 3.3 when we have some cursors in place. Then it might already be fixed.

community/kernel/src/main/java/org/neo4j/storageengine/api/StorageProperty.java
fickludd: Agreed, fixed this.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyStore.java
fickludd: Would like kernel team support on this!

community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyType.java
fickludd: Would like kernel team support on this!

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
fickludd: Fixed

community/consistency-check/src/main/java/org/neo4j/consistency/checking/PropertyRecordCheck.java
fickludd: yup!

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v2_3/TransactionBoundQueryContext.scala
fickludd: How about `asLegacyObject -> asObject` and `asPublic -> asObjectCopy`?

MishaDemianenko: sounds good!

community/values/src/main/java/org/neo4j/values/ValueTuple.java
MishaDemianenko: yeah, that's true that we have code paths that do not support composite paths so far, totally understand; what I would think that this method is even temporary till we will actually add support for composite indexes and will go away after that. I only do not like `Only` part of name since it gives impression that there is value and something else and does not really highlight that it's only for a first, single value

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/PropertyContainerStateImpl.java
MishaDemianenko: agree

MishaDemianenko:
Ok, i have created 2 cards that related to help on our wall and we wil take a look on that and it can go separatelly to now block the work, approving

MishaDemianenko (approved):

community/values/src/main/java/org/neo4j/values/ValueTuple.java
fickludd: `assertOneValueAndGetIt` !?

PR #9462, opened 2017-06-01
Handle multiple hints better

OliviaYtterbrink:
A cartesian product is planned as an alternative to the expands.
When using multiple hints on the same single hop pattern relationship, the planner will produce joins.

### DIFF
--- a/community/cypher/cypher-compiler-3.3/src/main/scala/org/neo4j/cypher/internal/compiler/v3_3/planner/logical/idp/SingleComponentPlanner.scala
+++ b/community/cypher/cypher-compiler-3.3/src/main/scala/org/neo4j/cypher/internal/compiler/v3_3/planner/logical/idp/SingleComponentPlanner.scala
@@ -21,7 +21,9 @@ package org.neo4j.cypher.internal.compiler.v3_3.planner.logical.idp
 
 
 import org.neo4j.cypher.internal.compiler.v3_3.helpers.IteratorSupport._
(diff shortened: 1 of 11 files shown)


community/cypher/cypher-compiler-3.3/src/main/scala/org/neo4j/cypher/internal/compiler/v3_3/planner/logical/idp/SingleComponentPlanner.scala
```diff
@@ -110,17 +122,29 @@ case class SingleComponentPlanner(monitor: IDPQueryGraphSolverMonitor,
...
         val rightExpand = planSinglePatternSide(qg, pattern, plan, end)
-        leftExpand.toSet ++ rightExpand.toSet
+
+        val startJoinNodes = Set(start)
+        val endJoinNodes = Set(end)
+        val maybeStartPlan = leaves.find(_.availableSymbols == startJoinNodes)
+        val maybeEndPlan = leaves.find(_.availableSymbols == endJoinNodes)
+        val extraPlans: Iterable[LogicalPlan] = (maybeStartPlan, maybeEndPlan) match {
+          case (Some(startPlan), Some(endPlan)) =>
+            val cartesianProduct = planSinglePatternSide(qg, pattern, context.logicalPlanProducer.planCartesianProduct(startPlan, endPlan), start)
```
systay: I think we need either more descriptive method names, variable names, or more comments. there is a lot going on here and the why is not at all clear

systay (changes requested):

systay (approved):

PR #9463, opened 2017-06-01
Value Module Remainders clean up

SaschaPeukert:


pontusmelke:
I think we should rename the root package as well, `org.neo4j.values` -> `org.neo4j.internal.values`

fickludd:
We have changed the package name here, but in hindsight I think it's better to wait with that change until https://github.com/neo4j/neo4j/pull/9460 is in. Otherwise we will get a bothersome rebase.

craigtaverner:
Will be included in a future PR

PR #9464, opened 2017-06-01
Unified SSL support and secure cluster communications

martinfurmanski:
The SSL support of the neo4j platform has been extracted into a new SSL
module which allows configuring and defining SSL policies under distinct
policy names. These policies can then be utilized and shared by
different internal components that require SSL.

There is a legacy policy which derives from the old configuration
utilized by bolt and web servers which will be exposed by the new
module under the policy name of legacy. This legacy policy is however
exposed as an SslPolicy in the same way as policies defined in the
new system. The intention is to eventually deprecate the legacy
settings and their derived policy.

The purpose of this change is to unify and expand the configuration
surface and implementation code and allow for easy consumption of
SSL policies by various components existing today and in the future.

The new SSL policies allow plenty of flexibility, for example in
the allowed versions of SSL/TLS and ciphers. It also supports
setting up mutual authentication based on a PKI scheme.

The secure cluster communication builds on the SSL policy system
and works to secure Raft and TX channels between cluster members.


community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyConfig.java
```diff
@@ -0,0 +1,112 @@
...
+    public SslPolicyConfig( String policyName )
+    {
+        GroupSettingSupport group = new GroupSettingSupport( SslPolicyConfig.class, policyName );
+
+        this.base_directory = group.scope( pathSetting( "base_directory", NO_DEFAULT ) );
+        this.allow_key_generation = group.scope( setting( "allow_key_generation", BOOLEAN, FALSE ) );
+        this.trust_all = group.scope( setting( "trust_all", BOOLEAN, FALSE ) );
+
+        this.private_key = group.scope( derivedDefault( "private_key", base_directory, "private.key" ) );
+        this.public_certificate = group.scope( derivedDefault( "public_certificate", base_directory, "public.crt" ) );
```
zhenlineo, line 97: Shall we still keep `public.cert`? As `cert` is 1) commonly used, 2) the same with the old ending.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyConfig.java
```diff
@@ -0,0 +1,112 @@
...
+        this.trusted_dir = group.scope( derivedDefault( "trusted_dir", base_directory, "trusted" ) );
+        this.revoked_dir = group.scope( derivedDefault( "revoked_dir", base_directory, "revoked" ) );
+
+        this.private_key_password = group.scope( setting( "private_key_password", STRING, NO_DEFAULT ) );
+        this.client_auth = group.scope( setting( "client_auth", options( ClientAuth.class, true ), ClientAuth.REQUIRE.name() ) );
```
zhenlineo, line 102: **Case 1:**
```
Trust_all = true
Client_auth = true
```
Clients trust all servers, servers (verify but) trust all clients
 
**Case 2:**
``` 
Trust_all=true
Client_auth = false
```
Clients trust all servers, servers (do not verify, a.k.a.) trust all clients
 
**Case 3:**
 ```
Trust_all=false
Client_auth = true
```
Clients verify servers, servers verify clients
 
**Case 4:**
``` 
Trust_all=false
Client_auth = false
```
Clients verify servers, servers (do not verify, a.k.a.) trust all clients

Am I got this correct?

File: community/kernel/src/main/java/org/neo4j/kernel/impl/factory/CommunityEditionModule.java [hunk 3]
```diff
@@ -99,6 +100,8 @@ public CommunityEditionModule( PlatformModule platformModule )
 
         GraphDatabaseFacade graphDatabaseFacade = platformModule.graphDatabaseFacade;
 
+        dependencies.satisfyDependency( SslPolicyLoader.create( config, logging.getInternalLogProvider() ) ); // for bolt and web server
```
zhenlineo, line 103: Why this is for bolt and web server if SslPolicyLoader rejects Legacy ssl settings? Can we have some tests to verify the old behaviour works as before?

community/ssl/src/test/java/org/neo4j/ssl/SslResourceBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+ * This builder has a finite set of pre-generated resource
+ * keys and certificates which can be utilized in tests.
+ *
+ * The script gen-root.sh is used to generate the root
+ * resources and gey-key.sh is used to generate the
+ * instance resources.
+ */
+public class SslResourceBuilder
+{
+    private static final String ROOT_CERTIFICATE_NAME = "root.crt";
```
zhenlineo, line 47: Can we create these keys and certs in this class, rather than commit the key and cert files into the source code?

The reason for asking for this is 1) It is hard to understand what is in the committed key and cert files. 2) If any tests failed due to errors in pre-created cert/key such as cert/key expiration, then it is hard to debug tests.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyLoader.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    private void load( Config config, Log log )
+    {
+        Function<ConfigValues,Stream<String>> enumeration = GroupSettingSupport.enumerate( SslPolicyConfig.class );
+        Set<String> policyNames = config.view( enumeration ).collect( toSet() );
+
+        for ( String policyName : policyNames )
+        {
+            if ( policyName.equals( LEGACY_POLICY_NAME ) )
```
zhenlineo, line 168: Can we have a test which covers this branch?

zhenlineo (changes requested):
The change looks good. 

But a few comments:
1) A few TODOs are left in the code. Shall we consider fixing them before merging to production?
2) Can we have more tests regarding this big change? Can we have some tests to verify that the legacy ssl policy is still work as expected to ensure nothing old was broken?
3) I have no idea about the any changes in NOTICE and LICENSE, pls verify changes in these files with others if you are not sure about it.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyConfig.java
martinfurmanski: 1. I think it is fine to merge this with TODOs in place. It is functionally complete for the functionality that is in. The TODOs are clean cut-off points.

2. This is mainly an exercise in integration and refactoring so most tests stay in place, for example for TLS support of Bolt and webserver. They work the same under the legacy policy. The new functionality with regards to secure communication does have tests and ITs.

3. Me neither. That is what the licensing tool did so I will have to assume it is as it should be. I will try to double-check with someone else though.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyConfig.java
martinfurmanski: Yes, I think you got this correct.

3/4 are the most important ones
3 being "mutual auth" functionality
4 could be interesting for https, possibly bolt

Exactly what 1 means I don't know, but it's not too important either.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/CommunityEditionModule.java
martinfurmanski: The SslPolicyLoader doesn't reject the legacy settings, it just incorporates the old settings into the new framework which then bolt/web uses and it should yield the same result. This already had tests in place.

community/ssl/src/test/java/org/neo4j/ssl/SslResourceBuilder.java
martinfurmanski: I found this way easier than trying to write programmatic support for everything I require. I agree it would be an improvement long term.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyLoader.java
martinfurmanski: I get what you mean now. This is not a rejection of the SSL policy itself (which derives from old settings), but rather just filters out so that one cannot create an SSL policy using the new settings. I will add a test to clarify that.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyConfig.java
martinfurmanski: That was a response to the general comments!

As for the file extension there really aren't any standards. I will think about it a bit!

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyLoader.java
zhenlineo: I see. It is to protect from using `dbms.ssl.policy.legacy.*`

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/CommunityEditionModule.java
zhenlineo: solved

zhenlineo (approved):

PR #9465, opened 2017-06-01
Reduce the probability of getting false positive deadlocks from the Forseti lock manager used by default in Enterprise Edition

chrisvest:
Note that false positives are an inherent risk with the Dreadlocks algorithm.
With this change, we still use Dreadlocks for the initial deadlock test, but if
one is found we follow up with a more thorough search of the waiter/lock-holder
dependency graph. Unfortunately, this search is _also_ probabilistic and may
find false positives. However, we only act upon the deadlock result if we have
made at least 10 attempts at grabbing the lock.

The two different deadlock tests, combined requiring 20 consequitive positive
deadlock results, means that the probability of a false positive making it all
the way through, is now considerably lower than it was before.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/util/collection/SimpleBitSet.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/util/collection/SimpleBitSet.java
@@ -44,6 +44,7 @@ import org.neo4j.collection.primitive.PrimitiveIntIterator;
  */
 public class SimpleBitSet extends StampedLock implements PrimitiveIntIterable
 {
+    private long lastCheckPointKey;
     private long[] data;
 
     public SimpleBitSet( int size )
@@ -108,11 +109,37 @@ public class SimpleBitSet extends StampedLock implements PrimitiveIntIterable
         unlockWrite( stamp );
     }
 
-    public void clear()
+    public long checkPointAndPut( long checkPoint, int key )
     {
-        long stamp = writeLock();
-        Arrays.fill( data, 0 );
-        unlockWrite( stamp );
+        // We only need to clear the bit set if it was modified since the last check point
+        if ( !validate( checkPoint ) || key != lastCheckPointKey )
+        {
+            long stamp = writeLock();
+            int idx = key >>> 6;
+            if ( idx < data.length )
+            {
+                Arrays.fill( data, 0 );
+            }
+            else
+            {
+                int len = data.length;
+                len = findNewLength( idx, len );
+                data = new long[len];
+            }
+            data[idx] = data[idx] | (1L<< (key & 63));
+            lastCheckPointKey = key;
+            checkPoint = tryConvertToOptimisticRead( stamp );
+        }
+        return checkPoint;
+    }
+
+    private int findNewLength( int idx, int len )
+    {
+        while ( len <= idx )
+        {
+            len *= 2;
+        }
+        return len;
     }
 
     public int size()
@@ -125,8 +152,7 @@ public class SimpleBitSet extends StampedLock implements PrimitiveIntIterable
 
     private void ensureCapacity( int arrayIndex )
     {
-        while(data.length <= arrayIndex)
-            data = Arrays.copyOf(data, data.length * 2);
+        data = Arrays.copyOf( data, findNewLength( arrayIndex, data.length ) );
     }
 
     //
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/util/collection/SimpleBitSet.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/util/collection/SimpleBitSet.java
@@ -20,6 +20,7 @@
 package org.neo4j.kernel.impl.util.collection;
 
 import java.util.Arrays;
+import java.util.concurrent.locks.StampedLock;
 
 import org.neo4j.collection.primitive.PrimitiveIntIterable;
 import org.neo4j.collection.primitive.PrimitiveIntIterator;
@@ -27,18 +28,21 @@ import org.neo4j.collection.primitive.PrimitiveIntIterator;
 /**
  * A basic bitset.
  *
- * Space usage: 20bytes + 1 bit per key.
+ * Represented using an automatically expanding long array, with one bit per key.
  *
  * Performance:
  *  * put, remove, contains, size: O(1)
  *  * clear: O(size/64)
  *
  * Concurrency semantics:
- *  * Concurrent writes may lead to data loss.
- *  * Concurrent reads is fine
- *  * Concurrent reads during write is allowed, but bulk put/remove is not atomic.
+ *  * Concurrent writes synchronise and is thread-safe
+ *  * Concurrent reads are thread-safe and will not observe torn writes, but may become
+ *    out of date as soon as the operation returns
+ *  * Concurrent reads during write is thread-safe
+ *  * Bulk operations appear atomic to concurrent readers
+ *  * Only caveat being that the iterator is not thread-safe
  */
-public class SimpleBitSet implements PrimitiveIntIterable
+public class SimpleBitSet extends StampedLock implements PrimitiveIntIterable
 {
     private long[] data;
 
@@ -48,48 +52,67 @@ public class SimpleBitSet implements PrimitiveIntIterable
         int capacity = 1;
         while (capacity < initialCapacity)
             capacity <<= 1;
+        long stamp = writeLock();
         data = new long[capacity];
+        unlockWrite( stamp );
     }
 
     public boolean contains(int key)
     {
         int idx = key >>> 6;
-        return data.length > idx && (data[idx] & ((1L << (key & 63)))) != 0;
+        boolean result;
+        long stamp;
+        do
+        {
+            stamp = tryOptimisticRead();
+            result = data.length > idx && (data[idx] & ((1L << (key & 63)))) != 0;
+        }
+        while ( !validate( stamp ) );
+        return result;
     }
 
     public void put( int key )
     {
+        long stamp = writeLock();
         int idx = key >>> 6;
         ensureCapacity(idx);
         data[idx] = data[idx] | (1L<< (key & 63));
+        unlockWrite( stamp );
     }
 
     public void put( SimpleBitSet other )
     {
+        long stamp = writeLock();
         ensureCapacity( other.data.length - 1 );
         for(int i=0;i<data.length && i<other.data.length;i++)
             data[i] = data[i] | other.data[i];
+        unlockWrite( stamp );
     }
 
     public void remove( int key )
     {
+        long stamp = writeLock();
         int idx = key >>> 6;
         if(data.length > idx)
         {
             data[idx] = data[idx] & ~(1L<< (key & 63));
         }
+        unlockWrite( stamp );
     }
 
     public void remove( SimpleBitSet other )
     {
+        long stamp = writeLock();
         for(int i=0;i<data.length;i++)
             data[i] = data[i] & ~other.data[i];
+        unlockWrite( stamp );
     }
 
     public void clear()
     {
-        for(int i=0;i<data.length;i++)
-            data[i] = 0L;
+        long stamp = writeLock();
+        Arrays.fill( data, 0 );
+        unlockWrite( stamp );
     }
 
     public int size()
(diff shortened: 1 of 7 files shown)


chrisvest:
The build for the first commit failed on exactly the test that I thought would be very unlikely, so in the second commit we're trying a little harder.

tinwelint (commented):
I _think_ this looks OK. I love the commentary, guiding the reader along nicely!

The only thing I'd be concerned about is the stamped lock in SimpleBitSet, couldn't that have an impact on performance?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
```diff
@@ -791,10 +796,75 @@ private void markAsWaitingFor( ForsetiLockManager.Lock lock, ResourceType type,
...
+                    throw new DeadlockDetectedException( message );
+                }
             }
         }
+        waitingForLock = null;
+    }
+
+    private boolean isDeadlockReal( ForsetiLockManager.Lock lock, int tries )
+    {
+        Set<ForsetiLockManager.Lock> waitedUpon = new HashSet<>();
```
jakewins: It took me a long time to grok what this set was tracking; maybe something like `seenLocks` or `locksAlreadyExplored` or something, rather than `waitedUpon`, that highlights that it's just making sure the algorithm isn't going back to re-examine locks it's already looked at (assuming I'm understanding this right?).

I can't see anything obviously wrong below, but it's also a bit tricky to follow.. maybe I'm becoming feeble minded from too much python programming ;)

FWIW, and independent of this PR: I think you very much were right that this algorithm overall optimizes for the wrong thing. The problem that we needed to solve in the first place was not making deadlock detection fast, it was making acquiring a lock that's free fast. 

I think the overhead of continuing trying to patch over the problems with Dreadlocks is much higher than the overhead of writing a new lock manager that maintains a regular RAG but has a good primitive for acquiring a lock that can be acquired without waiting.

chrisvest:
Waiting for benchmark results.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
```diff
@@ -770,10 +773,12 @@ private void clearWaitList()
         waitList.put( clientId );
     }
 
-    private void markAsWaitingFor( ForsetiLockManager.Lock lock, ResourceType type, long resourceId )
+    private void waitFor( ForsetiLockManager.Lock lock, ResourceType type, long resourceId, int tries )
     {
+        waitingForLock = lock;
         clearWaitList();
         lock.copyHolderWaitListsInto( waitList );
+        applyWaitStrategy( type, tries );
```
thobe, line 772: For some reason this refactoring makes this clearer than it was before - which is good - but it brings up a comment unrelated to the actual change you have made:

We are applying the wait strategy _before_ we check for deadlocks, which means that in the case of a deadlock we will spend extra time (waiting) before we detect the deadlock. That seems like a bad thing.

There is actually an aspect of this that relates to the change at hand: since the lock acquisition retry loop _surrounds_ the call to `waitFor(...)` and the implementation clears the `waitingForLock` field on exit, there is a possibility that the deadlock detection algorithm will see null in the `waitingForLock` field of a LockClient that is actually trying to wait for the lock, thus deferring the detection of the deadlock (in theory indefinitely, in reality probably not). If we instead assigned `waitingForLock` in the loop we would not have to clear it in between, but could just re-set it to the newly discovered lock in each iteration, and then clear it afterwards when we know we are not waiting for any lock.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
thobe: @jakewins that's what I've been saying all along 😉  - which is a poor way of saying that I agree.

Actually, I don't mind if we use the dreadlocks algorithm for initial deadlock detection or not, and then fall back to complete resource graph traversal as in this patch, if that speeds things up while not complicating the code much and without having a big cost overhead, I think it's a great idea. But I agree that the focus should be on making uncontended acquisition fast, whatever we do before going to sleep waiting for a lock is minimal compared to the cost overhead of actually waiting. I also, firmly, believe that we should never ever report a false deadlock to users.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
```diff
@@ -791,10 +796,75 @@ private void markAsWaitingFor( ForsetiLockManager.Lock lock, ResourceType type,
...
+                // to find a lock that has us among the owners.
+                // We only act upon the result of this method if the `tries` count is above some threshold. The reason
+                // is that the Lock.collectOwners, which is algorithm relies upon, is inherently racy, and so only
+                // reduces the probably of a false positive, but does not eliminate them.
+                if ( isDeadlockReal( lock, tries ) )
+                {
+                    // After checking several times, this really does look like a real deadlock.
+                    waitList.clear();
+                    waitingForLock = null;
+                    throw new DeadlockDetectedException( message );
```
thobe: Since we are now doing a full check to collect the clients and locks that wait for one another, we could actually produce a better error message than before - like the message produced by the community lock manager.

enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiFalseDeadlockTest.java
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    @Parameterized.Parameters( name = "{0}" )
+    public static Iterable<Object[]> parameters()
+    {
+        List<Fixture> fixtures = new ArrayList<>();
+
+        // During development I also had iteration counts 1 and 2 here, but they never found anything, so for actually
+        // running this test, I leave only iteration count 100 enabled.
+        int[] iterations = new int[]{100};
```
thobe: this singleton array is confusing - if not used, how about just inlining the value?

thobe (commented):
I think this is a step in the right direction.

It makes me a bit sad that this only lowers the probability of a false deadlock rather than eliminating it completely.

It would be great to have a comment describing the scenarios where we can still get false deadlocks.

chrisvest:
Benchmark says everything is within expected variance.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
jakewins: Agreed. This prompted a bunch of reading PG source code, and recognizing two interesting things: One; PG will block on a lock a solid second before it even tries detecting a deadlock; the fact that that does not appear to be an issue corroborates anecdotal experience that real-world deadlocks are really rare; it is likely not much of a performance penalty if detection is slow.

More interestingly though, the implementation contains a bunch of code for reordering wait lists on deadlock; some deadlocks can be removed by doing so:

Imagine two locks, `{l1, l2}`, and three clients `{A, B, C}`:

    (l1 sharedHolders=[A], waitList=[B(exclusive)])
    (l2 exclusive=[C], waitList=[A(exclusive)])

Ie: 
    A has a shared lock on l1 and waits to get a shared lock on l2
    B holds no locks, and is waiting to get an exclusive lock on l1
    C holds a shared lock on l1, blocking A

If C now decides it wants a shared lock on l1, it needs to wait behind
B; that will cause a deadlock. Forseti will simply abort one of them.

However, a better solution would be to reorder the wait list; if C is 
allowed to move ahead of B, it can be granted the shared lock on l1, and
the deadlock disappears.

enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiFalseDeadlockTest.java
```diff
@@ -0,0 +1,388 @@
...
+
+    /**
+     * This takes a fair bit of time, and we don't want to wait that long. But more importantly, false deadlocks are
+     * still only very unlikely; they are not impossible, though.
+     *
+     * So this test is technically flaky, but the probability is, I think, quite low for the 'mild' test, but it is too
+     * high for this aggressive test. So therefor I have marked it as @Ignored, but I still keep it here for the sake
+     * of this comment, and to allow others to run it and get a feel for the probabilities involved.
+     */
+    @Ignore
```
MishaDemianenko, line 270: instead of just ignoring this test what about creating separate stress test that will do more test runs with possibly more lock clients involved to cover as more cases as we can?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/SharedLock.java
```diff
@@ -214,6 +215,26 @@ public String describeWaitList()
     }
 
     @Override
+    public void collectOwners( Set<ForsetiClient> owners )
+    {
+        for ( AtomicReferenceArray<ForsetiClient> ownerArray : clientsHoldingThisLock )
```
MishaDemianenko, line 220: we can potentially go over and build that list of owners and that list of owners was never the case for this lock

community/kernel/src/main/java/org/neo4j/kernel/impl/util/collection/SimpleBitSet.java
```diff
@@ -20,25 +20,29 @@
...
- *  * Concurrent reads during write is allowed, but bulk put/remove is not atomic.
+ *  * Concurrent writes synchronise and is thread-safe
+ *  * Concurrent reads are thread-safe and will not observe torn writes, but may become
+ *    out of date as soon as the operation returns
+ *  * Concurrent reads during write is thread-safe
+ *  * Bulk operations appear atomic to concurrent readers
+ *  * Only caveat being that the iterator is not thread-safe
  */
-public class SimpleBitSet implements PrimitiveIntIterable
+public class SimpleBitSet extends StampedLock implements PrimitiveIntIterable
```
MishaDemianenko, line 45: simple bit set was updated with new guarantees but no new tests were added, hmm interesting ;)

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
thobe: @jakewins that is an interesting approach, eschew fairness in order to avoid deadlocks.

enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiFalseDeadlockTest.java
chrisvest: I think it takes about 30 minutes to run, though.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/SharedLock.java
chrisvest: Yes, and I think this is the reason that, as far as I have observed, deadlocks that involve shared locks are still the still the most probable false positive to see.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/collection/SimpleBitSet.java
chrisvest: I'll add some.

enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiFalseDeadlockTest.java
chrisvest: I'll inline it, then.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
chrisvest: With the incremental back-off and that we still go through several iterations before declaring that a deadlock is probably real, it doesn't make much difference.

I'll pull the `waitingForLock` clearing out to the end of the blocking methods, though.

chrisvest:
@thobe @MishaDemianenko I believe I've addressed you comments.

enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiFalseDeadlockTest.java
MishaDemianenko: But that's fine right for separate stress test build, as one we already have for backup, index population etc.

PR #9468, opened 2017-06-01
Remove START support from cost planner

systay:
changelog: Removes support for START from the cost planner

Instead of supporting it through the cost planner, Neo4j now will fall back to the
rule planner from earlier versions. This is because problems were found when running
queries with START through the cost planner.


.gitignore
```diff
@@ -2,6 +2,7 @@
 \#*
 target
 out
+.floo*
```
pontusmelke, line 5: what is this we are ignoring?

pontusmelke:
Could we add the following test back to `StartAcceptanceTest`

```
test("should return correct results on combined node and relationship index starts") {
    val node = createNode()
    val resultNode = createNode()
    val rel = relate(node, resultNode)
    relate(node, createNode())

    graph.inTx {
      graph.index.forNodes("nodes").add(node, "key", "A")
      graph.index.forRelationships("rels").add(rel, "key", "B")
    }

    val result = innerExecute("START n=node:nodes(key = 'A'), r=rel:rels(key = 'B') MATCH (n)-[r]->(b) RETURN b")
    result.toList should equal(List(Map("b" -> resultNode)))
  }
```

It may be a good idea to add more integration tests for the situations that are currently failing with the current ASTRewriter implementation, just to make sure that future us don't repeat these bugs.

.gitignore
systay: http://floobits.com

community/cypher/frontend-3.2/src/main/scala/org/neo4j/cypher/internal/frontend/v3_2/ast/Clause.scala
```diff
@@ -67,7 +68,9 @@ case class LoadCSV(withHeaders: Boolean, urlString: Expression, variable: Variab
 case class Start(items: Seq[StartItem], where: Option[Where])(val position: InputPosition) extends Clause {
   override def name = "START"
 
-  override def semanticCheck = items.semanticCheck chain where.semanticCheck
+  override def semanticCheck =
+    SemanticError("START is no longer supported. You can accomplish the same thing using MATCH and WHERE", position)
```
pontusmelke, line 72: Users will never see this error message right?

pontusmelke (approved):

community/cypher/frontend-3.2/src/main/scala/org/neo4j/cypher/internal/frontend/v3_2/ast/Clause.scala
systay: Correct.

PR #9470, opened 2017-06-01
Use transaction counters to wait for all transactions completion on shutdown

MishaDemianenko:
Use transaction statistics counters to wait for all transaction to
complete before NeoStoreDataSource will try to acquire transactional log
file lock to prevent deadlock on shutdown

<b>N.B!</b> should be null merged to 3.2 and 3.3.


chrisvest (approved):
👍

PR #9472, opened 2017-06-02
Handle arrays coming from parameters

pontusmelke:
Arrays were not properly handled when used as parameters, especially
there was a a failure when combining arrays and predicates when using
an index.


pontusmelke:
wrong branch

PR #9480, opened 2017-06-05
NativeSchemaNumberIndexAccessor and friends

tinwelint:
Implements `IndexAccessor` on top of `GBPTree` for number values. These changes build on the foundations of #9448 , but since all aspects of native schema number indexing is still under heavy development nearly all aspects have changed. Everything from naming, class hierarchy, tree layout and testing structures have been changed, for the better, implementing more features of the API. At the same time some simplifications regarding layout and class hierarchy have been made.

The commit history have been reduced some to get a gist of the reasoning behind some choices, but preferably the entire changeset should be viewed as one unit because there are changes which go back and forth within this PR, between the commits.

Notable highlights in this PR are:
- Implementation of `IndexAccessor` and its co-interfaces `IndexReader`/`IndexUpdater`
- Layout change where `NumberKey` stores actual value, not comparison value at all, i.e. less space and comparison have been implemented directly on actual values.
- Layout change where `NumberValue` is now empty since `NumberKey` containing all vital information, although kept in code to easily allow for quick changes in layout (which is very expected) w/o having to change the entire stack. In the end of this development cycle if `NumberValue` is still empty it will be removed completely along with tons of generics along with it.
- Much better testing structure and utilities for testing native index populator/accessor and its friends.


community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexReader.java
```diff
@@ -0,0 +1,165 @@
...
+import org.neo4j.kernel.api.schema.IndexQuery;
+import org.neo4j.kernel.api.schema.IndexQuery.ExactPredicate;
+import org.neo4j.kernel.api.schema.IndexQuery.NumberRangePredicate;
+import org.neo4j.kernel.configuration.Config;
+import org.neo4j.kernel.impl.api.index.sampling.IndexSamplingConfig;
+import org.neo4j.storageengine.api.schema.IndexReader;
+import org.neo4j.storageengine.api.schema.IndexSampler;
+
+class NativeSchemaNumberIndexReader<KEY extends NumberKey, VALUE extends NumberValue>
+        implements IndexReader
```
chrisvest: > Reader for an index. Must honor repeatable reads, which means that if a lookup is executed multiple times the
> same result set must be returned.

Does the GBPTree uphold this requirement of the index reader?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberKey.java
```diff
@@ -0,0 +1,200 @@
...
+
+import org.neo4j.index.internal.gbptree.GBPTree;
+
+/**
+ * Includes value and entity id (to be able to handle non-unique values).
+ * A value can be any {@link Number} and is represented as a {@code long} to store the raw bits and a type
+ * to say if it's a long, double or float.
+ *
+ * Distinction between double and float exists because coersions between each other and long may differ.
+ * TODO this should be figured out and potentially reduced to long, double types only.
```
chrisvest, line 33: Not a trivial thing. The important thing is that an index lookup/seek/scan always returns the same data as a store-scan + filter would.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberKey.java
```diff
@@ -0,0 +1,200 @@
...
+     * Extracts data from a {@link Number} into state of this {@link NumberKey} instance.
+     *
+     * @param value actual {@link Number} value.
+     */
+    private void extractValue( Number value )
+    {
+        if ( value instanceof Double )
+        {
+            type = TYPE_DOUBLE;
+            rawValueBits = Double.doubleToLongBits( (Double) value );
```
chrisvest, line 141: This collapses all NaN values to a single canonical NaN value, meaning all NaNs are the same. I'm not sure what we are supposed to promise in this situation.

chrisvest: Maybe this is something the CLG wants to know about as well, who knows.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexReader.java
tinwelint: No it does not and this has been discussed internally previously. There are at least two ways forward that I can see:
- collaborate with cypher, which is the first and foremost user relying on this, to potentially alleviate/workaround this requirement
- eagerize/cache the results in places where this is a requirement, which is what Lucene does. Althought may be difficult to pull off well.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberKey.java
tinwelint: Yup, this isn't necessary to do (merge float and double). The native number index has full number precision (no precision loss due to coersion) and so will always return exactly the same results as store-scan + filter do.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberKey.java
tinwelint: Yes, that is an interesting thing to evaluate.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FailureHeaderWriter.java
```diff
@@ -41,9 +46,9 @@
     public void accept( PageCursor cursor )
     {
         byte[] bytesToWrite = failureBytes;
-        cursor.putByte( NativeSchemaIndexPopulator.BYTE_FAILED );
+        cursor.putByte( NativeSchemaNumberIndexPopulator.BYTE_FAILED );
         int availableSpace = cursor.getCurrentPageSize() - cursor.getOffset();
-        if ( bytesToWrite.length + 2 > availableSpace )
+        if ( bytesToWrite.length + HEADER_LENGTH_FIELD_LENGTH > availableSpace )
         {
             bytesToWrite = Arrays.copyOf( bytesToWrite, availableSpace - 2 );
```
MishaDemianenko: you can't just fix all magic 2 :)

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeNonUniqueSchemaNumberIndexPopulator.java
```diff
@@ -87,14 +87,16 @@ public IndexSample sampleResult()
         }
         finally
         {
-            // Start the writer again (TODO may be unnecessary)
+            // Start the writer again (TODO may be unnecessary since this happens after all population is completed,
```
MishaDemianenko: I think that this comment is mostly useless to be honest

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexPopulator.java
```diff
@@ -225,49 +198,8 @@ private void markTreeAsOnline() throws IOException
         tree.checkpoint( IOLimiter.unlimited(), ( pc ) -> pc.putByte( BYTE_ONLINE ) );
```
MishaDemianenko, line 198: empty byte array thingy is still here: `"".getBytes()`

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexReader.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @Override
+    public PrimitiveLongIterator query( IndexQuery... predicates ) throws IndexNotApplicableKernelException
+    {
+        if ( predicates.length != 1 )
+        {
+            throw new UnsupportedOperationException();
+        }
+
+        ensureOpenSeekerClosed();
```
MishaDemianenko: what is the reason of this limitation?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NonUniqueNumberLayout.java
```diff
@@ -0,0 +1,41 @@
...
+import org.neo4j.index.internal.gbptree.Layout;
+
+public class NonUniqueNumberLayout extends NumberLayout
+{
+    private static final String IDENTIFIER_NAME = "NUNI";
+
+    @Override
+    public long identifier()
+    {
+        // todo Is Number.Value.SIZE a good checksum?
```
MishaDemianenko: todo

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberKey.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    String propertiesAsString()
+    {
+        return String.valueOf( toNumberValue() );
+    }
+
+    void initAsLowest()
+    {
+        rawValueBits = Double.doubleToLongBits( Double.NEGATIVE_INFINITY );
```
MishaDemianenko, line 75: we need to agree with cypher on highers/lowest values and we should not redefine them here, we just need to use some common place where it's defined for us and for cypher

MishaDemianenko: the same goes for comparison order

MishaDemianenko (commented):
Have only couple of comments, otherwise fine

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexReader.java
burqen: Because we initially thought we only needed to support one open reader at any time, like how label scan store used to work (addressed in #9489 ). Will remove this.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberKey.java
burqen: Lucene uses `Double.compare` so that behaviour should be similar at least.
As for highest/lowest, this is only a way for us to create a range that is unbound (left, right or both) and because we use `Double.compare` for comparison it should not be possible to store anything outside of this limit.
Do you still we need to define a common place for this?

burqen:
@MishaDemianenko @chrisvest Thanks for sweet feedback! Comments should be addressed now.

MishaDemianenko:
@burqen rebase required.
Also, have you checked comparison/equality with cypher guys for NaN's, comparison order?

otherwise no more questions

burqen:
@MishaDemianenko Rebase done 🎉 
- NaN will be discussed within open cypher - no decision yet, address in different PR
- Align comparison / equality with cypher ties closely together with using new Value interface (which was the cause of rebase) and will be addressed in different PR.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexReader.java
```diff
@@ -0,0 +1,175 @@
...
+        return true;
+    }
+
+    private PrimitiveLongIterator startSeekForInitializedRange()
+    {
+        try
+        {
+            RawCursor<Hit<KEY,VALUE>,IOException> seeker = tree.seek( treeKeyFrom, treeKeyTo );
+            openSeekers.add( seeker );
+            return new NumberHitIterator<>( seeker, openSeekers );
```
MishaDemianenko, line 155: Since this iterator needs to be closed now shouldn't it be resource iterator instead?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberHitIterator.java
```diff
@@ -0,0 +1,83 @@
...
+        this.seeker = seeker;
+        this.toRemoveFromWhenExhausted = toRemoveFromWhenExhausted;
+    }
+
+    @Override
+    protected boolean fetchNext()
+    {
+        try
+        {
+            if ( seeker.next() )
```
MishaDemianenko: this will throw an exception in case: iterator.fetchNext() will return false and then you will call fetchNext again right?

MishaDemianenko (changes requested):
@burqen still have couple of comments

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexReader.java
burqen: That change would need to travel all the way up the stack for it to take any effect, right?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberHitIterator.java
burqen: Actually, `SeekCursor` can handle multiple calls to `next` even if it returns false. However, we only know that it is a `RawCursor` so might be better to explicitly handle this case here.

burqen:
@MishaDemianenko Thanks for the feedback! 🎉  Comments should be addressed.

MishaDemianenko (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberKey.java
tinwelint: I believe this is already handled in #9543

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberKey.java
tinwelint: I believe this is already handled in #9543

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexReader.java
tinwelint: This will be handled in future PR

PR #9482, opened 2017-06-05
Add preliminary syntax for RETURN GRAPH

Mats-SX:
Also makes using these preliminary Cypher constructs generate a different type of error, in order for another implementation to separate them properly.


craigtaverner (approved):

Mats-SX:
@craigtaverner The failing test is running green locally, and marked in TC as flaky. I restarted the build, but I doubt these changes are the cause of the failure.

PR #9483, opened 2017-06-05
Optional lookup filter for number queries

tinwelint:
Until now index number query results have needed double-checking w/ store
comparing actual values, due to the way Lucene schema index indexes numbers.

iThe native schema index doesn't need this and so it would be unnecessary
to have this filtering when using the native schema index.
Now IndexReader has a new method #hasFullNumberPrecision() which decides
whether or not number index query results must be filtered outside the index.

### DIFF
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/inmemory/HashBasedIndex.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/inmemory/HashBasedIndex.java
@@ -291,7 +291,7 @@ class HashBasedIndex extends InMemoryIndexImplementation
     @Override
     public boolean hasFullNumberPrecision()
     {
-        return true;
+        return false;
     }
 
     private interface StringFilter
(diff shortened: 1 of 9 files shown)


chrisvest (approved):

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/inmemory/HashBasedIndex.java
```diff
@@ -288,6 +288,12 @@ public PrimitiveLongIterator query( IndexQuery... predicates )
         }
     }
 
+    @Override
+    public boolean hasFullNumberPrecision()
+    {
+        return true;
+    }
```
chrisvest: I think you have failing tests because this isn't true. The HashBasedIndex coerces all numbers to doubles, and thus looses precision.

chrisvest:
Pushed a fix for that. Now we'll see if the build approves.

PR #9484, opened 2017-06-05
Split global schema lock

MishaDemianenko:
Split global schema lock into independent label and relationship type locks
that will be used depending from type of index, constraint etc.
Completely remove SCHEMA resource from type of resources that can be locked.

Slaves in ha now will take labels and rel type locks on master exactly like
for nodes and relationship locks.

This commit performs split in kernel/ha code.
Cypher code not updated as part of this commit.


community/kernel/src/main/java/org/neo4j/kernel/impl/api/LockingStatementOperations.java
```diff
@@ -481,34 +459,33 @@ public Value graphRemoveProperty( KernelStatement state, int propertyKeyId )
     }
 
     @Override
-    public void acquireExclusive( KernelStatement state, ResourceType resourceType, long resourceId )
+    public void acquireExclusive( KernelStatement state, ResourceType resourceType, long... resourceId )
```
fickludd: `resourceIds`? Similar for other methods below in that case.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -994,28 +1000,28 @@ public void uniqueIndexDrop( IndexDescriptor descriptor ) throws DropIndexFailur
 
     // <Locking>
     @Override
-    public void acquireExclusive( ResourceType type, long id )
+    public void acquireExclusive( ResourceType type, long... id )
```
fickludd: could be `ids`?

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/ResourceTypes.java
```diff
@@ -37,9 +37,11 @@
     NODE( 0, LockWaitStrategies.INCREMENTAL_BACKOFF ),
     RELATIONSHIP( 1, LockWaitStrategies.INCREMENTAL_BACKOFF ),
     GRAPH_PROPS( 2, LockWaitStrategies.INCREMENTAL_BACKOFF ),
-    SCHEMA( 3, LockWaitStrategies.INCREMENTAL_BACKOFF ),
+    // SCHEMA resource type had typeId 3 - skip it to avoid resource types conflicts
```
fickludd, line 40: Good call!

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/community/CommunityLockClient.java
```diff
@@ -270,21 +270,24 @@ private boolean reEnter( PrimitiveLongObjectMap<LockResource> localLocks, long r
...
-            LockResource resource = localLocks.get( resourceId );
-            if ( resource.releaseReference() != 0 )
+            for ( long resourceId : resourceIds )
             {
-                return;
-            }
+                LockResource resource = localLocks.get( resourceId );
+                if ( resource.releaseReference() != 0 )
+                {
+                    return;
```
fickludd: Do we not want to try releasing any remaining resources after the first failure? I don't know, maybe this will terminate the transaction and release everything anyway, just noting this observation.

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/community/CommunityLockClient.java
```diff
@@ -293,20 +296,23 @@ public void releaseShared( ResourceType resourceType, long resourceId )
...
-            if ( resource.releaseReference() != 0 )
+            for ( long resourceId : resourceIds )
             {
-                return;
-            }
-            localLocks.remove( resourceId );
+                LockResource resource = localLocks.get( resourceId );
+                if ( resource.releaseReference() != 0 )
+                {
+                    return;
```
fickludd: See return comment above.

fickludd (approved):
The idea of splitting the schema lock is really good, let's hope that we can back that by some SOAK-run or something! A workload with regular reads and increased amounts of index/constraint changes should definitely benefit.

I think the cypher part of this look good, but it did not go through all the kernel parts. Some very minor naming comments, but nothing that has to be fixed.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -586,11 +586,6 @@
     public static final Setting<String> procedure_unrestricted =
             setting( "dbms.security.procedures.unrestricted", Settings.STRING, "" );
 
-    @Description( "Whether or not to release the exclusive schema lock is while building uniqueness constraints index" )
-    @Internal
-    public static final Setting<Boolean> release_schema_lock_while_building_constraint = setting(
-            "unsupported.dbms.schema.release_lock_while_building_constraint", BOOLEAN, FALSE );
```
chrisvest, line 592: I don't think we can technically remove this. But we can certainly deprecate and ignore it.

community/kernel/src/main/java/org/neo4j/kernel/api/ReadOperations.java
```diff
@@ -276,17 +276,21 @@ IndexDescriptor indexGetForSchema( LabelSchemaDescriptor descriptor )
...
-    void acquireShared(    ResourceType type, long id );
+    void acquireExclusive( ResourceType type, long... id );
+
+    void acquireShared( ResourceType type, long... id );
+
+    void releaseExclusive( ResourceType type, long... id );
 
-    void releaseExclusive( ResourceType type, long id );
-    void releaseShared(    ResourceType type, long id );
+    void releaseShared( ResourceType type, long... id );
```
chrisvest: We can pluralise `id`

community/kernel/src/main/java/org/neo4j/kernel/impl/api/LockingStatementOperations.java
```diff
@@ -98,18 +94,7 @@ public LockingStatementOperations(
...
-        //
-        // We would like this locking to be done naturally when ConstraintEnforcingEntityOperations calls
-        // SchemaReadOperations#constraintsGetForLabel, but the SchemaReadOperations object that
-        // ConstraintEnforcingEntityOperations has a reference to does not lock because of the way the cake is
-        // constructed.
-        //
-        // It would be cleaner if the schema and data cakes were separated so that the SchemaReadOperations object used
-        // by ConstraintEnforcingEntityOperations included the full cake, with locking included.
-        acquireSharedSchemaLock( state );
+        sharedLabelLock( state, labelId );
```
chrisvest, line 97: 👍

community/kernel/src/main/java/org/neo4j/kernel/impl/api/operations/LockOperations.java
```diff
@@ -24,9 +24,9 @@
...
 {
-    void acquireExclusive( KernelStatement state, ResourceType resourceType, long resourceId );
-    void acquireShared( KernelStatement state, ResourceType resourceType, long resourceId );
+    void acquireExclusive( KernelStatement state, ResourceType resourceType, long... resourceId );
+    void acquireShared( KernelStatement state, ResourceType resourceType, long... resourceId );
 
-    void releaseExclusive( KernelStatement statement, ResourceType type, long id );
-    void releaseShared( KernelStatement statement, ResourceType type, long id );
+    void releaseExclusive( KernelStatement statement, ResourceType type, long... id );
+    void releaseShared( KernelStatement statement, ResourceType type, long... id );
```
chrisvest: Another opportunity to pluralise.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
```diff
@@ -560,47 +566,52 @@ public void releaseShared( ResourceType resourceType, long resourceId )
...
+            ConcurrentMap<Long,ForsetiLockManager.Lock> resourceTypeLocks = lockMaps[resourceType.typeId()];
+            PrimitiveLongIntMap sharedLocks = sharedLockCounts[resourceType.typeId()];
+            for ( long resourceId : resourceIds )
             {
-                return;
-            }
+                if ( releaseLocalLock( resourceType, resourceId, exclusiveLocks ) )
+                {
+                    return;
+                }
```
chrisvest: This should be a `continue`, not a `return`.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
```diff
@@ -536,21 +536,27 @@ public boolean reEnterExclusive( ResourceType resourceType, long resourceId )
...
+            PrimitiveLongIntMap exclusiveLocks = exclusiveLockCounts[resourceType.typeId()];
+            ConcurrentMap<Long,ForsetiLockManager.Lock> resourceTypeLocks = lockMaps[resourceType.typeId()];
+            for ( long resourceId : resourceIds )
             {
-                return;
-            }
+                if ( releaseLocalLock( resourceType, resourceId, sharedLocks ) )
+                {
+                    return;
+                }
```
chrisvest: This should be a `continue`, not a `return`.

chrisvest (changes requested):
I think you made a mistake in adding support for releasing multiple locks to Forseti. There's also a few parameter names that could be changed.

chrisvest (approved):

chrisvest:
👌

MishaDemianenko:
ping @chrisvest

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobTest.java
```diff
@@ -640,13 +644,14 @@ private IndexPopulationJob newIndexPopulationJob( FailedIndexProxyFactory failur
         return job;
     }
 
-    private IndexDescriptor indexDescriptor( Label label, String propertyKey, boolean constraint ) throws TransactionFailureException
+    private IndexDescriptor indexDescriptor( Label label, String propertyKey, boolean constraint )
+            throws TransactionFailureException, IllegalTokenNameException, TooManyLabelsException
     {
-        try ( KernelTransaction tx = kernel.newTransaction( KernelTransaction.Type.implicit, AnonymousContext.read() );
+        try ( KernelTransaction tx = kernel.newTransaction( KernelTransaction.Type.implicit, AnonymousContext.AUTH_DISABLED );
```
chrisvest, line 650: Why the change from `AnonymousContext.read()` to `AUTH_DISABLED`?

enterprise/ha/src/main/java/org/neo4j/kernel/ha/MasterTransactionCommitProcess.java
```diff
@@ -19,73 +19,46 @@
...
  */
 public class MasterTransactionCommitProcess implements TransactionCommitProcess
 {
-    /**
-     * Detector of transactions coming in from slaves which should acquire the shared schema lock before
-     * being applied (and validated for application).
-     */
-    private static final Visitor<StorageCommand,IOException> REQUIRES_SHARED_SCHEMA_LOCK = command ->
-            command instanceof NodeCommand && mayResultInIndexUpdates( (NodeCommand) command ) ||
-            command instanceof PropertyCommand && mayResultInIndexUpdates( (PropertyCommand) command );
```
chrisvest, line 54: 🎉

enterprise/ha/src/main/java/org/neo4j/kernel/ha/lock/SlaveStatementLocks.java
```diff
@@ -0,0 +1,84 @@
...
+import org.neo4j.kernel.impl.locking.ActiveLock;
+import org.neo4j.kernel.impl.locking.LockTracer;
+import org.neo4j.kernel.impl.locking.Locks;
+import org.neo4j.kernel.impl.locking.StatementLocks;
+
+/**
+ * Slave specific statement locks that in addition to standard functionality provided by configured delegate
+ * will grab selected shared locks on master during prepareForCommit phase.
+ */
+public class SlaveStatementLocks implements StatementLocks
```
chrisvest: Some more explanation on why, both here and in the `SlaveLocksClient` would be useful. Reviewing this, I have no idea why this is necessary.

chrisvest (commented):
Not a thorough review, but I have some questions.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobTest.java
MishaDemianenko: Since now we do not allow creation of index descriptors for non-existent properties and labels we should be able to create them if they are not explicitly created by tests and to be able to do that we can't use `AnonymousContext.read()`. And since it does not really matter for this test `AUTH_DISABLED` use instead here now

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/ExecutionEngine.scala
```diff
@@ -192,6 +206,34 @@ class ExecutionEngine(val queryService: GraphDatabaseQueryService,
...
+    import scala.collection.JavaConverters._
+
+    def planLabels = {
+      plan._1.plannerInfo.indexes().asScala.collect { case item: SchemaIndexUsage => item.getLabelId.toLong }
+    }
+
+    version match {
+      case CypherVersion.v3_3 => planLabels
+      case CypherVersion.v3_2 => planLabels
+      case _ => tc.statement.readOperations().labelsGetAllTokens().asScala.map( t => t.id().toLong ).toSeq
```
lutovich: Maybe a comment here is needed to describe why we need to lock all labels for older versions?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/ExecutionEngine.scala
```diff
@@ -192,6 +206,34 @@ class ExecutionEngine(val queryService: GraphDatabaseQueryService,
...
+  private def extractPlanLabels(plan: (ExecutionPlan, Map[String, Any]), version: CypherVersion, tc:
+  TransactionalContextWrapper): Seq[Long] = {
+    import scala.collection.JavaConverters._
+
+    def planLabels = {
+      plan._1.plannerInfo.indexes().asScala.collect { case item: SchemaIndexUsage => item.getLabelId.toLong }
+    }
+
+    version match {
+      case CypherVersion.v3_3 => planLabels
```
lutovich: This match will need to be updated in 3.4, right? Could we compare versions and to turn code into something like:

```
if(version greaterThan CypherVersion.v3_1) planLabels
else lockAllLabels()
```

enterprise/ha/src/test/java/org/neo4j/kernel/ha/lock/trace/RecordingLockTracer.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import org.neo4j.kernel.impl.locking.LockTracer;
+import org.neo4j.kernel.impl.locking.LockWaitEvent;
+import org.neo4j.storageengine.api.lock.ResourceType;
+
+public class RecordingLockTracer implements LockTracer
+{
+    private List<LockRecord> requestedLocks = Collections.synchronizedList( new ArrayList<>() );
```
lutovich: Might be better to use COWAL so clients can update and iterate at the same time.

lutovich (commented):
@MishaDemianenko changes generally look good. Made couple small comments. It would also be great to have some ITs that show advantages of schema lock spilt in single instance and CC. Something like: writes can happen when constraint is being built, two constraints on different labels/types can be built at the same time, etc.

MishaDemianenko:
@lutovich should be fixed now! thx, great feedback!

lutovich (approved):
@MishaDemianenko changes look good 👍

PR #9489, opened 2017-06-07
NativeLabelScanReader supports multiple open iterators

tinwelint:
previously opening a new iterator closed the previous, because that's
how the API was interpreted to work. New use cases would benefit from
multiple concurrent iterators opened, so this commit adds that.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
@@ -135,13 +135,10 @@ class NativeLabelScanReader implements LabelScanReader
 
     private void ensureOpenCursorsClosed() throws IOException
     {
-        if ( !openCursors.isEmpty() )
+        for ( RawCursor<Hit<LabelScanKey,LabelScanValue>,IOException> cursor : openCursors )
         {
-            for ( RawCursor<Hit<LabelScanKey,LabelScanValue>,IOException> cursor : openCursors )
-            {
-                cursor.close();
-            }
-            openCursors.clear();
+            cursor.close();
         }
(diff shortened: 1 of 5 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
```diff
@@ -137,10 +135,13 @@ public PrimitiveLongIterator nodesWithAllLabels( int... labelIds )
 
     private void ensureOpenCursorsClosed() throws IOException
     {
-        RawCursor<Hit<LabelScanKey,LabelScanValue>,IOException> cursor;
-        while ( ( cursor = openCursors.poll() ) != null )
+        if ( !openCursors.isEmpty() )
```
tinwelint: Silly check

community/kernel/src/test/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIteratorTest.java
```diff
@@ -0,0 +1,65 @@
...
+        LabelScanValueIterator iterator = new LabelScanValueIterator( cursor, toRemoveFrom );
+        verify( cursor, times( 0 ) ).close();
+
+        // WHEN
+        exhaust( iterator );
+        verify( cursor, times( 1 ) ).close();
+
+        // retrying to get more items from the first one should not close it again
+        iterator.hasNext();
+        verify( cursor, times( 1 ) ).close();
```
tinwelint, line 56: Perhaps assert removed from the set?

tinwelint (commented):
Since this was driven by a feature in Cypher it would be great with a Cypher test or integration test showing that.

burqen (approved):

PR #9496, opened 2017-06-09
Use G1 as garbage collector for tests

MishaDemianenko:
Switch test collector to G1 to match default recommendations and options.


systay:
I'm curious - why?

MishaDemianenko:
@systay By default, we ship with G1 as GC for product and from that perspective it make sense to make default for our tests

burqen (approved):

burqen:
@MishaDemianeno Approved! 🎉

PR #9497, opened 2017-06-09
Run tests in multiple forked VMs

MishaDemianenko:
Introduce possibility to run tests in multiple forked VMs.
Be default we will create up to CPU count VMs and will reuse them for afterwords.
Forkability is regulated by forkTests profile that can be deactivated
by using skipTestsFork property.


lassewesth:
I, for one, am lovin' it!

PR #9500, opened 2017-06-09
Relax a couple of slow running page cache tests, speeding them up

chrisvest:


community/io/src/test/java/org/neo4j/io/pagecache/PageCacheSlowTest.java
```diff
@@ -628,10 +630,12 @@ private void verifyAdversarialPagedContent( PagedFile pagedFile ) throws IOExcep
     @Test
     public void mustNotRunOutOfSwapperAllocationSpace() throws Exception
     {
+        assumeTrue( "This test is file system agnostic, and too slow on a real file system",
+                fs instanceof EphemeralFileSystemAbstraction );
         configureStandardPageCache();
 
         File file = file( "a" );
-        int iterations = Short.MAX_VALUE * 20; // Integer.MAX_VALUE;
+        int iterations = Short.MAX_VALUE * 3; // Integer.MAX_VALUE;
```
MishaDemianenko: there is a comment here mentioning his Majesty `Integer.MAX_VALUE` but it looks like it's a wrong place to do that :)

MishaDemianenko (approved):
There is a small super minor thing with a comment, feel free to push that without another round of review :)

community/io/src/test/java/org/neo4j/io/pagecache/PageCacheSlowTest.java
chrisvest: Yeah, looks like debugging or refactoring left-overs.

PR #9501, opened 2017-06-11
Replace and deprecate START

pontusmelke:
Add built-in procedures `db.nodeManualIndexSeek`, `db.nodeManualIndexSearch`, `db.relationshipManualIndexSeek`, and `db.relationshipManualIndexSearch` replacing `START` for reading from legacy index. Also deprecate `START` and add a notification that tell the user how to rewrite the query to use a procedure or `MATCH`.

changelog:  Add procedures for reading data from manual indexes. Replaces `START n=node:index(...)` and `START r =relationship:index(...)`.


chrisvest:
They're called _manual indexes_ now: https://neo4j.com/docs/java-reference/3.2/#tutorials-java-embedded-index

pontusmelke:
@chrisvest We updated the name of the procedures. ~We also renamed the exception and the status code, however I am unsure if we can do this renaming in 3.3?~ We should rename the status code and the exception as well but I am not sure what is the appropriate deprecation cycle for doing so.

community/cypher/frontend-3.3/src/main/scala/org/neo4j/cypher/internal/frontend/v3_3/ast/Clause.scala
```diff
@@ -86,15 +86,46 @@ case class EmitGraph(graphName: Expression)(val position: InputPosition) extends
...
+    val newState = state.addNotification(DeprecatedStartNotification(position, query))
+    SemanticCheckResult(newState, Seq(SemanticError(
+      s"""START is deprecated, use: `$query` instead.
+       """.stripMargin, position)))
+  }
+
+  private def rewrittenQuery: String = {
+    val rewritten = items.map {
+      case NodeByIdentifiedIndex(variable, index, key, expression) =>
+        s"CALL db.nodeLegacyIndexSeek('$index', '$key', '${expression.asCanonicalStringVal}') YIELD node AS ${variable.asCanonicalStringVal}"
```
systay: This should talk about manual indexes and not legacy stuff, right? Other examples in the next couple of lines

community/kernel/src/main/java/org/neo4j/graphdb/impl/notification/NotificationCode.java
```diff
@@ -184,7 +184,12 @@
     START_UNAVAILABLE_FALLBACK(
             SeverityLevel.WARNING,
             Status.Statement.PlannerUnavailableWarning,
-            "START is unsupported for current CYPHER version, the query has been execute by an older CYPHER version"
+            "START is not supported for current CYPHER version, the query has been executed by an older CYPHER version"
```
systay, line 187: Is this still used or lying around for compatibility reasons? I thought this PR removed the fallback.

systay (changes requested):

community/kernel/src/main/java/org/neo4j/graphdb/impl/notification/NotificationCode.java
pontusmelke: It is still around, just changed the wording. So my thinking was that it is two different notifications, one is saying: _this query could not be run on this version of cypher_ and the other one is saying: _`START` is deprecated_.  It is really two different things, or is my thinking wrong?

systay: Makes sense. 👍

systay (approved):

PR #9505, opened 2017-06-12
Update NOTICE files to copyright 2017

Mats-SX:
Also adds a test that fails the build if we forget this again.

### DIFF
--- a/community/licensecheck-config/src/test/java/org/neo4j/licensecheck/LicenseYearTest.java
+++ b/community/licensecheck-config/src/test/java/org/neo4j/licensecheck/LicenseYearTest.java
@@ -42,19 +42,25 @@ public class LicenseYearTest
     @Test
     public void testNoticePrefixYearGPL() throws FileNotFoundException
     {
-        URL resource = getClass().getClassLoader().getResource( NOTICE_GPL_FILE );
-        File gplFile = new File( resource.getFile() );
+        if ( System.getProperty( "ignoreNoticeYear" ).isEmpty() )
+        {
+            URL resource = getClass().getClassLoader().getResource( NOTICE_GPL_FILE );
+            File gplFile = new File( resource.getFile() );
 
-        checkYearInFile(gplFile);
+            checkYearInFile( gplFile );
+        }
     }
 
     @Test
     public void testNoticePrefixYearAGPL() throws FileNotFoundException
     {
-        URL resource = getClass().getClassLoader().getResource( NOTICE_AGPL_FILE );
-        File gplFile = new File( resource.getFile() );
+        if ( System.getProperty( "ignoreNoticeYear" ).isEmpty() )
+        {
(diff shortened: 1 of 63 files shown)


martinfurmanski (approved):
Should be good for the next 7982.5 years.

martinfurmanski:
Suitable for older branches as well?

community/licensecheck-config/src/test/java/org/neo4j/licensecheck/LicenseYearTest.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Test
+    public void testNoticePrefixYearAGPL() throws FileNotFoundException
+    {
+        URL resource = getClass().getClassLoader().getResource( NOTICE_AGPL_FILE );
+        File gplFile = new File( resource.getFile() );
+
+        checkYearInFile( gplFile );
+    }
```
chrisvest: This test will fail if you next year check out a tag from this year and try to build it. I think we could counter that if we added a `-Dparameter` that turns the test off, and then have the test fail with a message that tells you to either update the license headers (could even tell you what maven command would do this) or ignore the test with the given parameter if one insists on not wanting to change any files.

martinfurmanski: Is there some "official build" parameter flag we could use? Which would be yes for packaging builds, but generally false anywhere else.

Mats-SX: I would think that any version that we build has to have this field updated. If not, what is the point of having the year there at all? This means that at new year's, we'll update this branch's and all subsequent branches' `NOTICE` files, and then this wouldn't be a problem. What do you think?

Of course, you could make the argument that in a future where `3.3` enters the archive, then we'd stop updating and then any build past a new year's would fail, but I think at that point _very_ few builds will be made. I'd not want to spend too much time catering for that use case (this stuff with build parameters is a bit foreign to me).

Mats-SX:
@chrisvest @martinfurmanski Thanks for the feedback!

The doc-tools failure is expected, as it also needs a `NOTICE` file update. I'm creating a PR to that repo in parallel to this: https://github.com/neo4j/neo4j-documentation/pull/91

This should be good to merge if that one is approved (and that my argument vis-a-vis the test is accepted).

community/licensecheck-config/src/test/java/org/neo4j/licensecheck/LicenseYearTest.java
chrisvest: I'm talking about tags, not branches that can obviously be updated if need be.

Mats-SX: Ah, I misunderstood that.

Mats-SX:
@chrisvest I added a maven option that allows ignoring this test.

PR #9506, opened 2017-06-12
Relicense Cypher frontend under Apache License 2.0

Mats-SX:
- Regenerate all source file license headers


fickludd (approved):
This looks ok!

PR #9507, opened 2017-06-12
Optional match in compiled runtime now check that nodes exists.

eebus:
changelog: Optional match in compiled runtime now checks that nodes exist before checking the label.

### DIFF
--- a/enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/ir/expressions/HasLabel.scala
+++ b/enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/ir/expressions/HasLabel.scala
@@ -34,12 +34,16 @@ case class HasLabel(nodeVariable: Variable, labelVariable: String, labelName: St
     structure.declarePredicate(localName)
 
     structure.incrementDbHits()
-    if (nodeVariable.nullable)
-      structure.nullableReference(nodeVariable.name, CodeGenType.primitiveNode,
-                                  structure.box(
-                                    structure.hasLabel(nodeVariable.name, labelVariable, localName), CodeGenType.primitiveBool))
+    if (nodeVariable.nullable) {
+      structure.ifNotStatement(structure.equalityExpression(structure.loadVariable(nodeVariable.name),
+        structure.constantExpression(Long.box(-1L)),
+        CodeGenType.primitiveInt)){ inner =>
+        inner.hasLabel(nodeVariable.name, labelVariable, localName)
+      }
+    }
     else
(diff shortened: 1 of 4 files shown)


enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/ir/expressions/HasLabel.scala
```diff
@@ -34,12 +34,16 @@ case class HasLabel(nodeVariable: Variable, labelVariable: String, labelName: St
     structure.declarePredicate(localName)
```
pontusmelke, line 34: I think this will get rid of most of the compilation errors 
`structure.declareAndInitialize(localName, CodeGenType.primitiveBool)`

enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/ir/expressions/HasLabel.scala
```diff
@@ -34,12 +34,16 @@ case class HasLabel(nodeVariable: Variable, labelVariable: String, labelName: St
...
 
     structure.incrementDbHits()
-    if (nodeVariable.nullable)
-      structure.nullableReference(nodeVariable.name, CodeGenType.primitiveNode,
-                                  structure.box(
-                                    structure.hasLabel(nodeVariable.name, labelVariable, localName), CodeGenType.primitiveBool))
+    if (nodeVariable.nullable) {
```
pontusmelke, line 37: I think it should be something like this
```
structure.incrementDbHits()
    if (nodeVariable.nullable) {
      structure.ifNotStatement(structure.equalityExpression(structure.loadVariable(nodeVariable.name),
                                                            structure.constantExpression(Long.box(-1L)),
                                                            CodeGenType.primitiveInt)) { inner =>
        inner.hasLabel(nodeVariable.name, labelVariable, localName)
      }
      structure.nullableReference(nodeVariable.name, CodeGenType.primitiveNode,
                                  structure.box(
                                    structure.loadVariable(localName), CodeGenType.primitiveBool))

    }
    else {
      structure.hasLabel(nodeVariable.name, labelVariable, localName)
      structure.loadVariable(localName)
    }
```

The important thing is that: if the node id is `-1` the end-result should not be `false` but `null`. A test doing something like `MATCH (n) OPTIONAL MATCH (o: DOESNT_EXIST) RETURN o:L` is going to fail in the current implementation I think, so adding that test case might be a good idea.

pontusmelke (changes requested):

eebus:
This PR is replaced with #9524

PR #9513, opened 2017-06-13
Consistency checker now check label scan store for missing entries

burqen:
Previously, it was only verified that everything in label scan store
was consistent with the real store. It was never checked that
everything in real store was actually in the label scan store.

Verification is done by iterating over all entries in label scan store
and verify content with the store. If some node in the store was not
present in the label index, this would not be reported because this
node would never show up in the iteration over the label scan store.

Now, we wrap the label scan store iterator with a 
`GapFreeAllEntriesLabelScanReader` that produce "node with no labels"
records for all the gaps in the label scan index. So for example if
label scan index contain node 1, 2 and 3 with labels [1,2], [1] and [2] 
respectively, written as: `{[1:1,2],[3:1],[5:2]}`. The gap filling wrapper
will produce a view over the label scan index that also contain node
2 and 4 with no labels: `{[1:1,2],[2:_],[3:1],[4:_],[5:2]}`. In this way
we make sure all node records in the store is verified agains the
label scan store. Note that `[2:_]` is consistent with node 2 not
having any labels or not being in use at all.

[cl] Consistency checker now verifies that all nodes with some label(s) are present in Label Index. Previously it only checked that everything in Label Index aligned with the store.

### DIFF
--- a/community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/GapFreeAllEntriesLabelScanReader.java
+++ b/community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/GapFreeAllEntriesLabelScanReader.java
@@ -66,10 +66,10 @@ class GapFreeAllEntriesLabelScanReader implements AllEntriesLabelScanReader
         private int currentId;
         private boolean first;
 
-        GapFillingIterator( Iterator<NodeLabelRange> iterator, long highId )
+        GapFillingIterator( Iterator<NodeLabelRange> nodeLableRangeIterator, long highId )
         {
             this.highId = highId;
-            this.source = iterator;
+            this.source = nodeLableRangeIterator;
             this.first = true;
         }
 
@@ -84,6 +84,7 @@ class GapFreeAllEntriesLabelScanReader implements AllEntriesLabelScanReader
             int cursor = 0;
             for ( ; cursor < batchSize; cursor++, currentId++ )
             {
+                // First or empty source
                 if ( first || (sourceNodeIds != null && sourceIndex >= sourceNodeIds.length) )
                 {
                     first = false;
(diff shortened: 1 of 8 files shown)


community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/GapFreeAllEntriesLabelScanReader.java
```diff
@@ -0,0 +1,162 @@
...
+        private static final long[] EMPTY_LONG_ARRAY = new long[0];
+        private final long highId;
+        private Iterator<NodeLabelRange> source;
+        private NodeLabelRange nextFromSource;
+        private long[] sourceNodeIds;
+        private int sourceIndex;
+        private int currentId;
+        private boolean first;
+
+        GapFillingIterator( Iterator<NodeLabelRange> iterator, long highId )
```
ragadeeshu: Could `iterator` perhaps be made more descriptive?

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/GapFreeAllEntriesLabelScanReader.java
```diff
@@ -0,0 +1,162 @@
...
+        {
+            long baseId = currentId;
+            int batchSize = BATCH_SIZE;
+            long[] nodes = new long[batchSize];
+            long[][] labels = new long[batchSize][];
+
+            int cursor = 0;
+            for ( ; cursor < batchSize; cursor++, currentId++ )
+            {
+                if ( first || (sourceNodeIds != null && sourceIndex >= sourceNodeIds.length) )
```
ragadeeshu: This handles the case when we get the first, and when it is empty? A small comment would be nice :)

ragadeeshu (commented):
A few clarifications would be nice, looks good otherwise :+1:

ragadeeshu:
@burqen review done

burqen:
@ragadeeshu Thank you! Comments addressed 🎉

PR #9516, opened 2017-06-13
NativeLSS will redo population if did not complete first time

burqen:
Before if NativeLabelScanStore was initialised (init) for the first time
without being populated (start), then it would not retry population
job on next start.

Now it will note in GBPTree header that it needs to be rebuilt on next
start and only clean this header after population has completed.

[cl] Fix a problem where Label Index could be seen as online even if population did not complete, for example after migration, effectively creating an inconsistent Label Index.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
@@ -139,6 +139,9 @@ public class GBPTree<KEY,VALUE> implements Closeable
      */
     public interface Monitor
     {
+        /**
+         * Adapter for {@link Monitor}.
+         */
         class Adaptor implements Monitor
         {
             @Override
@@ -199,7 +202,7 @@ public class GBPTree<KEY,VALUE> implements Closeable
     /**
      * No-op header reader.
      */
-    static final Header.Reader NO_HEADER = (cursor,length) -> {};
+    static final Header.Reader NO_HEADER_READER = (cursor,length) -> {};
 
     /**
      * No-op header writer.
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
@@ -202,6 +202,11 @@ public class GBPTree<KEY,VALUE> implements Closeable
     static final Header.Reader NO_HEADER = (cursor,length) -> {};
 
     /**
+     * No-op header writer.
+     */
+    static final Consumer<PageCursor> NO_HEADER_WRITER = pc -> {};
+
+    /**
      * Paged file in a {@link PageCache} providing the means of storage.
      */
     private final PagedFile pagedFile;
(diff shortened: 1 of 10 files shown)


community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -202,6 +202,11 @@ public void startupState( boolean clean )
     static final Header.Reader NO_HEADER = (cursor,length) -> {};
 
     /**
+     * No-op header writer.
+     */
+    static final Consumer<PageCursor> NO_HEADER_WRITER = pc -> {};
```
ragadeeshu: Why does this not live in `Header`?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -523,7 +529,7 @@ private static void writerHeader( PagedFile pagedFile, Header.Writer headerWrite
         // Write/carry over header
         int headerOffset = cursor.getOffset();
         int headerDataOffset = headerOffset + Integer.BYTES; // will contain length of written header data (below)
-        if ( otherState.isValid() )
+        if ( otherState.isValid() || headerWriter != CARRY_OVER_PREVIOUS_HEADER )
```
ragadeeshu, line 535: Why do you check for `CARRY_OVER_PREVIOUS_HEADER` here? Could you not use it instead?

ragadeeshu (commented):
A few things I want some clarification on :)

ragadeeshu:
@burqen review done

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: Because `NO_HEADER` live in `GBPTree` as well, together with `NO_MONITOR`.
I think the convenience of having the common no-op implementations of the constructor parameters available directly in the class itself outweigh them being in the more low lever classes to which the more logically belong.

Especially for `NO_HEADER_WRITER` which is a generic `Consumer<PageCursor>` and thus has no obvious "aah, the no-op I'm looking for should obviously be [here]".
For `NO_HEADER`, that obvious place would be `Header` or `Header.Reader` as you pointed out but I rather have them living together.

Wow, wall of text, sorry about that.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: Because if we use `CARRY_OVER_PREVIOUS_HEADER` that means we copy the header from `otherState`. But if `otherState` is not valid then we will write corrupt data to our tree state, which would be sad. However, if the `headerWriter` is _not_ a `CARRY_OVER_PREVIOUS_HEADER` it is safe for us to just write.

burqen:
@ragadeeshu Comments addressed 🎉

community/kernel/src/test/java/org/neo4j/kernel/api/impl/labelscan/LabelScanStoreTest.java
```diff
@@ -633,5 +642,10 @@ public void init()
         {
             initCalled = true;
         }
+
+        public void reset()
+        {
+            initCalled = rebuildingCalled = rebuiltCalled = noIndexCalled = corruptedIndex = false;
```
MishaDemianenko: please use separate lines

MishaDemianenko (changes requested):
@burqen Only one thing with multiple assignments in a line otherwise looks good and ready to go!

burqen:
@MishaDemianenko Comment addressed 🎉

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -202,6 +202,11 @@ public void startupState( boolean clean )
     static final Header.Reader NO_HEADER = (cursor,length) -> {};
```
tinwelint: I think this should be renamed to `NO_HEADER_READER`. now that there's a `NO_HEADER_WRITER`

MishaDemianenko (approved):

PR #9520, opened 2017-06-14
Fix bug around pattern comprehensions

### LINKED ISSUE #9329: Neo4j Cypher node filtering by pattern comprehension
I have a following SDN 4 entities:

`Decision`, `Characteristic` and `Value`:

    @NodeEntity
    public class Value extends Votable {
    
    	private final static String SET_FOR = "SET_FOR";
    	private final static String SET_ON = "SET_ON";
    	private final static String CONTAINS = "CONTAINS";
    
    	@Relationship(type = SET_FOR, direction = Relationship.OUTGOING)
    	private Decision decision;
    
    	@Relationship(type = SET_ON, direction = Relationship.OUTGOING)
    	private Characteristic characteristic;
    
    	@Index(unique = false)
    	private Object value;
    
        ...
    }

I have created 3 `Decicion`(`Decicion1`, `Decision2`, `Decisison3`) nodes and 1 `Characteristic`(`Characteristic1`).
For `Decicion1`, `Decicion2` and `Characteristic1` I have created a `Double` `Value`, for example:

    Decision1 + Characteristic1 = Value(500d)
    Decision2 + Characteristic1 = Value(1000d)
    Decicion3 doesn't have any Value on Characteristic1

I need to create a query that will return all `Decision` which have `Value` in the defined range, for example `100 <= value <= 50000`

Based on the example above this query should return `Decision1` and `Decicion2` only.

Right now I have a following query:

    MATCH (parentD)-[:CONTAINS]->(childD:Decision)-[ru:CREATED_BY]->(u:User) 
        WHERE id(parentD) = {decisionId} 
            AND ALL(key IN keys({rangeFilters}) 
                WHERE size(
                    [(childD)<-[:SET_FOR]-(filterValue)-[:SET_ON]->(filterCharacteristic) WHERE id(filterCharacteristic) = toInt(key) AND ({rangeFilters}[key])[0] <= filterValue.value <= ({rangeFilters}[key])[1] | 1]
                ) > 0)  
    RETURN ru, u, childD AS decision SKIP 0 LIMIT 100 

where `rangeFilters` is a `Map<String, Double[]>` where key is a `Characteristic` `ID` and value `new Double[] { new Double(100.d), new Double(50000d) }`

But this query returns all 3 `Decision`, even the `Decision3` that don't have any values associated with `Characteristic1` .

Why the pattern comprehension doesn't work here as expected?

This is an example that highlights the issue: [http://console.neo4j.org/?id=6bv9y5](http://console.neo4j.org/?id=6bv9y5)

Why the only "d1" decision is considered there ?

systay:
In some situations, pattern comprehension patterns where not bound to
the already found nodes and thus produces wrong results.

Fixes #9329

### DIFF
--- a/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/planner/logical/QueryGraphSolver.scala
+++ b/community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/planner/logical/QueryGraphSolver.scala
@@ -50,7 +50,6 @@ trait PatternExpressionSolving {
 
   def planPatternComprehension(planArguments: Set[IdName], expr: PatternComprehension)
                               (implicit context: LogicalPlanningContext): (LogicalPlan, PatternComprehension) = {
-    val dependencies = expr.dependencies.map(IdName.fromVariable)
     val asQueryGraph = expr.asQueryGraph
     val qgArguments = planArguments intersect asQueryGraph.coveredIds
     val qg = asQueryGraph.withArgumentIds(qgArguments).addPredicates(expr.predicate.toIndexedSeq:_*)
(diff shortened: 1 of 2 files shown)


community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/planner/logical/QueryGraphSolver.scala
```diff
@@ -51,8 +51,9 @@ trait PatternExpressionSolving {
   def planPatternComprehension(planArguments: Set[IdName], expr: PatternComprehension)
                               (implicit context: LogicalPlanningContext): (LogicalPlan, PatternComprehension) = {
     val dependencies = expr.dependencies.map(IdName.fromVariable)
```
pontusmelke: dependencies no longer needed

pontusmelke (changes requested):

pontusmelke (approved):

PR #9523, opened 2017-06-14
`KernelIntegrationTest` retain db on restart

ragadeeshu:
Some extending tests already assume this to be the case.


ragadeeshu:
@MishaDemianenko Updated :)

MishaDemianenko:
@ragadeeshu everything looks fine except removed test, can you please explain why that gets removed?

ragadeeshu:
@MishaDemianenko Because it tests for old behavior, that was changed in https://github.com/neo4j/neo4j/pull/9068 This test should have been removed back then.

MishaDemianenko (approved):
Ok, thx. LGTM

PR #9524, opened 2017-06-14
Removed OptionalExpand from compiled runtime

eebus:
changelog: Fix issue with the compiled runtimes handling of optional expand when the incoming bound node is null. This change removes support for optional expand entirely from the compiled runtime.


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
```diff
@@ -97,6 +97,21 @@ class MatchAcceptanceTest extends ExecutionEngineFunSuite with QueryStatisticsTe
     }
   }
 
+  test("should handle optional match of none existant path.") {
```
craigtaverner: Minor spelling/grammar issue. Should be `non-existent path`.

craigtaverner (approved):

PR #9526, opened 2017-06-16
Consistency checker now warn if schema indexes are not online

burqen:


community/consistency-check/src/main/java/org/neo4j/consistency/checking/index/IndexAccessors.java
```diff
@@ -97,6 +108,6 @@ public void close() throws IOException
             accessor.close();
         }
         accessors.clear();
-        indexRules.clear();
+        onlineIndexRules.clear();
```
MishaDemianenko, line 111: if we already do clear here then we need to clear `notOnlineIndexRules` as well

community/consistency-check/src/main/java/org/neo4j/consistency/report/ConsistencyReport.java
```diff
@@ -152,6 +152,10 @@ void forCounts( CountsEntry countsEntry,
 
         @Documented( "The schema rule contained in the DynamicRecord chain is of an unrecognized Kind" )
         void unsupportedSchemaRuleKind( SchemaRule.Kind kind );
+
+        @Warning
+        @Documented( "The schema rule contained in the DynamicRecord chain a reference to a schema rule that is not online. It will be rebuilt on next startup." )
```
MishaDemianenko: The warning message is a bit wrong here -  at least _is_ word is missing, also it's not that correct since failed indexes will not gonna be rebuilt for example right?

community/consistency-check/src/test/java/org/neo4j/consistency/ConsistencyCheckServiceIntegrationTest.java
```diff
@@ -204,11 +209,77 @@ public void shouldAllowGraphCheckDisabled() throws IOException, ConsistencyCheck
...
+        assertThat( "Expected to see report about schema index not being online",
+                Files.readAllLines( reportFile.toPath() ).toString(), allOf(
+                        containsString( "schema rule" ),
+                        containsString( "not online" )
+                ) );
+    }
+
+    private File findFile( String targetFile, File directory )
+    {
+        File foundFile = null;
```
MishaDemianenko: maybe something like this instead?
```
        File file = new File( directory, targetFile );
        if (!file.exists())
        {
            fail( "Could not find file " + targetFile );
        }
        return file;
```

MishaDemianenko (changes requested):
A couple of code comments, also I do not think this should go to 3.2 since it will start to warn people in cases it did not warn before and I do not think we should do that in patch releases.
Otherwise good find!

burqen:
@MishaDemianenko Comments addressed and rebased and retarget for 3.3

MishaDemianenko (approved):

PR #9530, opened 2017-06-16
Fix discovery listen address ignored

martinfurmanski:
This was previously fixed by #9086 but it seems the
forward merge was partially botched.


mikernorton:
Thanks!

PR #9531, opened 2017-06-16
Update core pool size and max pool size evaluation for PooledConcurrentMergeScheduler.

MishaDemianenko:
Update core pool size to be equal to pool minimum thread,
change the way of evaluating max pool threads;
Use long adder instead of atomic to perform open tasks tracking.


burqen (approved):
Changes look good. Do we have any performance numbers for this change?

MishaDemianenko:
@burqen now test 1.5.x faster; link to the full benchmark run results is available on the card.

PR #9535, opened 2017-06-19
Fix broken backwards compatibility with array values

fickludd:
The recently introduced value module accidentally changed how property values of empty arrays were encoded in schema indexes. This PR restores the legacy index array encoding, so that migration works again.


fickludd:
Failing test should be unrelated.

burqen (approved):

PR #9542, opened 2017-06-21
Optimisations for the ParallelBatchImporter

chrisvest:
This seems to make the transfer of batches into ForkedProcessorSteps about 15-20 times faster. The impact of this is particularly clear on the running time of the `mustNotDetachProcessorsFromBatchChains` test.

changelog: Speed up data linking in importer by increasing parallelism, and allowing out-of-order batch transfer. The linking of data used to account for about half the time of an import, importing the raw unlinked data the other half). Now it typically accounts for far less than that.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/AbstractStep.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/AbstractStep.java
@@ -24,6 +24,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.LongAdder;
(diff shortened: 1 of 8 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/AbstractStep.java
```diff
@@ -61,10 +62,10 @@ public boolean test( long ticket )
 
     // Milliseconds awaiting downstream to process batches so that its queue size goes beyond the configured threshold
     // If this is big then it means that this step is faster than downstream.
-    protected final AtomicLong downstreamIdleTime = new AtomicLong();
+    protected final LongAdder downstreamIdleTime = new LongAdder();
```
tinwelint, line 65: Good find

tinwelint (approved):
Great to see new eyes on this code, you're doing wonders here

PR #9543, opened 2017-06-21
Align native index sorting with values module

burqen:
New RawBits class is responsible for comparison between values given represented
as raw bits and value type byte.

Another option would be to materialise raw bits and value type to NumberValue
objects but we want to avoid as much object creation as possible.


community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/RawBitsTest.java
```diff
@@ -0,0 +1,152 @@
...
+                new Object[]{"Unique",
+                        new UniqueNumberLayout()
+                },
+                new Object[]{"NonUnique",
+                        new NonUniqueNumberLayout()
+                }
+        );
+    }
+
+    final List<Object> objects = new ArrayList<>( Arrays.asList(
```
fickludd: ´new ArrayList<>( Arrays.asList(´... en list för mycket?

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/RawBitsTest.java
```diff
@@ -0,0 +1,152 @@
...
+import static org.junit.Assert.assertEquals;
+
+@RunWith( Parameterized.class )
+public class RawBitsTest
+{
+    @Parameterized.Parameter()
+    public String name;
+
+    @Parameterized.Parameter( 1 )
+    public NumberLayout layout;
```
fickludd, line 48: Nice to parametrize on the NumberLayout!

fickludd (approved):
Looks very good, just found a minor thing that you can fix if you want to.

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/RawBitsTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+        // when
+        values.sort( Values.COMPARATOR );
+        schemaNumberKeys.sort( layout );
+        List<Value> actual = schemaNumberKeys.stream()
+                .map( k -> RawBits.asNumberValue( k.rawValueBits, k.type ) )
+                .collect( Collectors.toList() );
+
+        // then
+        assertSameOrder( actual, values );
```
chrisvest: I think we want to test the comparison of all the values to all the values, so N² comparisons. Shuffling and sorting the elements will only give us N·log(N) comparisons. I'd also add in a handful of floats and doubles that only differ by one ulp, e.g. Math.PI vs Math.nextUp/nextDown(Math.PI), and similarly for floats. I'd also add some int/long values that floats/doubles can almost-but-not-quite represent accurately.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/SchemaNumberKey.java
```diff
@@ -0,0 +1,175 @@
...
+    void from( long entityId, Value... values )
+    {
+        extractRawBitsAndType( assertValidSingleNumber( values ) );
+        this.entityId = entityId;
+        entityIdIsSpecialTieBreaker = false;
+    }
+
+    private static NumberValue assertValidSingleNumber( Value... values )
+    {
+        // TODO: support multiple values, right?
```
chrisvest, line 69: Is this coming later?

chrisvest (changes requested):
The changes themselves are pretty straight forward, but I'd like to see more thorough testing. It would be annoying to get this wrong.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/SchemaNumberKey.java
tinwelint: It may do, if time permits. In any way a separate story from this PR.

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/RawBitsTest.java
tinwelint: Added some more testing, have a look if you think that's OK

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/RawBitsTest.java
```diff
@@ -0,0 +1,203 @@
...
+                List<SchemaNumberKey> withThisOneInWrongPlace = new ArrayList<>( withoutThisOne );
+                withThisOneInWrongPlace.add( i, numberKey );
+                withThisOneInWrongPlace.sort( layout );
+                List<Value> actual = asValues( withThisOneInWrongPlace );
+
+                // then
+                assertSameOrder( actual, values );
+            }
+        }
+    }
```
chrisvest, line 154: I was more thinking of a nested loop where you check that for all combinations `layout.compare` return the same as `Values.COMPARATOR.compare`.

tinwelint: Alright, added that too now.

chrisvest (approved):

PR #9544, opened 2017-06-21
Add a `WorkSync.applyAsync` method for asynchronous application of work

chrisvest:
The asynchronous work might be done in the submitting thread, or in any other
thread that interacts with the given `WorkSync`.

The submitting of asynchronous work will never block. All blocking is deferred
to the `AsyncApply.await` method. This method must be called before the work
can be expected to have been applied, and then only if it does not throw an
exception.

Additionaly, if the work ends up being applied in a different thread, then any
thrown exception will only propagate to that thread. The `AsyncApply`
insteances will still be marked as done, but there will be no way to tell if
the application was successful or exceptional.

If the work submitters require such information to be propagated, then they
must encode such propagation infrastructure into the work itself.

### DIFF
--- a/community/primitive-collections/src/test/java/org/neo4j/concurrent/WorkSyncTest.java
+++ b/community/primitive-collections/src/test/java/org/neo4j/concurrent/WorkSyncTest.java
@@ -19,6 +19,9 @@
  */
 package org.neo4j.concurrent;
 
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


community/primitive-collections/src/test/java/org/neo4j/concurrent/WorkSyncTest.java
```diff
@@ -257,4 +305,192 @@ public void mustNotApplyWorkInParallelUnderStress() throws Exception
...
+        {
+            assertThat( e.getCause(), sameInstance( boo ) );
+        }
+
+        assertThat( sum.sum(), is( 10L ) );
+        assertThat( count.sum(), is( 1L ) );
+    }
+
+    @Test
+    public void asyncWorkDoneInAnotherThreadThatThrowsMustRememberException() throws Exception // TODO!!!
```
tinwelint: Could you clarify how there's another thread in this test throwing an exception which that main test thread (calling `await`) gets to see?

chrisvest: Ah, the name is wrong. Also, that TODO needs to go.

PR #9545, opened 2017-06-21
Removed unnecessary tests

eebus:
Some tests caused flakey errors without checking any valuable functionality.


chrisvest:
@eebus A rebase should fix the Tools Compile build.

PR #9549, opened 2017-06-21
Fix outdated javadoc in GraphDatabaseService

MishaDemianenko:
Update javadoc to use new signature for newEmbeddedDatabase in GraphDatabaseFactory
that takes File now instead of string path.

Fixes neo4j/neo4j-documentation#90


klaren (approved):
Looks good!

PR #9554, opened 2017-06-23
Dynamic binding of bolt connector to available ports

MishaDemianenko:
Allow bolt connectors to perform dynamic binding to any free port
Add possibility to find out out which port was used to register any connector
to allow any further operations.

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/transport/NettyServer.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/transport/NettyServer.java
@@ -35,7 +35,7 @@ import java.util.concurrent.ThreadFactory;
 
 import org.neo4j.helpers.ListenSocketAddress;
 import org.neo4j.helpers.PortBindException;
-import org.neo4j.kernel.api.bolt.BoltConnectorRegister;
+import org.neo4j.kernel.api.bolt.BoltPortRegister;
 import org.neo4j.kernel.configuration.BoltConnector;
 import org.neo4j.kernel.lifecycle.LifecycleAdapter;
 
@@ -52,7 +52,7 @@ public class NettyServer extends LifecycleAdapter
 
     private final Map<BoltConnector, ProtocolInitializer> bootstrappersMap;
     private final ThreadFactory tf;
-    private BoltConnectorRegister connectionRegister;
+    private final BoltPortRegister connectionRegister;
     private EventLoopGroup bossGroup;
     private EventLoopGroup selectorGroup;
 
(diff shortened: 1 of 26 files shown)


MishaDemianenko:
@lutovich wdyt?

community/bolt/src/main/java/org/neo4j/bolt/transport/NettyServer.java
```diff
@@ -46,8 +50,9 @@
     private static final int NUM_SELECTOR_THREADS = Math.max( 1, Integer.getInteger(
             "org.neo4j.selectorThreads", Runtime.getRuntime().availableProcessors() * 2 ) );
 
-    private final Collection<ProtocolInitializer> bootstrappers;
+    private final Map<BoltConnector, ProtocolInitializer> bootstrappersMap;
     private final ThreadFactory tf;
+    private BoltConnectorRegister connectionRegister;
```
lutovich: Maybe make field `final`?

community/kernel/src/main/java/org/neo4j/kernel/api/bolt/BoltConnectorRegister.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.neo4j.helpers.HostnamePort;
+
+/**
+ * Connector tracker that keeps information about local address that any configured connector get during bootstrapping.
+ */
+public class BoltConnectorRegister
+{
+
+    private ConcurrentHashMap<String,HostnamePort> connectorsInfo = new ConcurrentHashMap<>();
```
lutovich: Maybe make field `final`?

community/kernel/src/main/java/org/neo4j/kernel/api/bolt/BoltConnectorRegister.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.net.InetSocketAddress;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.neo4j.helpers.HostnamePort;
+
+/**
+ * Connector tracker that keeps information about local address that any configured connector get during bootstrapping.
+ */
+public class BoltConnectorRegister
```
lutovich: What do you think about `BoltPortRegister` name?

lutovich (approved):
@MishaDemianenko changes look good! couple tiny comments.

MishaDemianenko:
@lutovich make all those things final and rename the class. Not sure if that will make sense in future since most probably the same component will track HTTP ports for corresponding connectors, but it's  good for now :)

PR #9560, opened 2017-06-25
Draft of Cypher-Kernel API

fickludd:
Initial draft of the new Cypher Kernel API.

The current draft is considered reasonably close to the final API, and should be incorporated in the neo4j repo to allow parallel runtime prototyping, and eventually integration into the product.

This draft has been tested by implementing parts of the read and cursor interface for reading Neo4j store files. That implementation (NeoLE) was successful, outperformed Neo4j in speed, and will be used to back cypher runtime prototyping.


tinwelint:
To avoid confusion, I'd like to clarify that I'd like to view this is a new design of the Kernel API... not an additional Cypher-Kernel API. Or perhaps this is a transitional term?

fickludd:
@tinwelint I don't know what to call it to be honest, and it's a bit annoying. I put Cypher in because Cypher is intended as the primary user, but you are right: it should probably become the _only_ Kernel API.

MishaDemianenko:
@fickludd this API should be a complete replacement for old-kernel-API right? and it also means that old should be removed correct?

fickludd:
@MishaDemianenko That is really a kernel team decision =). To me that sounds like a sane thing to do, but it might take some time before this API is implemented, all supported Cypher version use it, and the Core API reimplemented on top. Just saying that a decent name to separate this from `OperationsFacade` and friend would be useful.

But yes, this API should have all the powers of the current Kernel API, but be more explicit, so having both seems weird.

MishaDemianenko:
@fickludd if we already doing replacement for old API the end goal should be _replacement_ meaning we finish with only one kernel API and all modules (old cyphers) should be updated to use it

pontusmelke:
This looks awesome. My only minor objection is about naming, especially _edge_ vs _relationship_ and _search structure_ vs _manual index_. Having now worked on integrating the `value-module` with the cypher runtime I feel the inconsistency in naming is annoying me. I have no opinion on what names to use but we will end up with a lot of implementations like

```
Relationship getRelationship()
{
   EdgeScanCursor cursor ...
   ...
}
```
There has just been an effort to start to renaming _legacy indexes_ to _manual indexes_ in the code base, I feel that throwing in a third name in the mix is just making it unnecessarily complicated for us.

build/checkstyle.xml
```diff
@@ -85,7 +85,6 @@
       <property name="option" value="space"/>
     </module>
     <module name="AvoidStarImport"/>
-    <module name="MultipleVariableDeclarations"/>
```
chrisvest, line 88: Or just fix the one spot where you do this.

fickludd:
To interested parties:

I want to merge this PR as well as #9561, so that the runtime project can use NeoLE for prototyping. If anyone thinks that this is a completely horrible thing to do, shout out during the day. We can as always still change things, and I guess that runtime prototyping will drive minor API changes going forward.

MishaDemianenko:
@fickludd was waiting for you to come back to start looking into it, give me please some time to review it

community/kernel-api/src/main/java/org/neo4j/impl/kernel/api/SuspendableCursor.java
```diff
@@ -0,0 +1,34 @@
...
+package org.neo4j.impl.kernel.api;
+
+/**
+ * After invoking {@link #suspend()} this cursor is closed and needs to be re-initialized.
+ *
+ * A cursor can be (re-)initialized by passing a position to the {@linkplain #resume(CursorPosition) resume method}.
+ *
+ * @param <Position> the type of position used by this cursor.
+ */
+interface SuspendableCursor<Position extends CursorPosition<Position>> extends Cursor
```
MishaDemianenko, line 29: This suspendable cursor is looking like continuation and can be potentially interesting not only for this API, just a note

MishaDemianenko:
@fickludd look fine in general except naming part i do not think that introduing another naming strategies like `manual indexes` or `Egde....` is a good thing to do only on this API level. It will gonna generate confusions out of thin air and will add additional ommunication complexity.
If those things needs to be renamed the whole stack should be renamed.
cc @tinwelint

tinwelint:
@MishaDemianenko @fickludd agreed, these naming differences may do more bad than good. But we can rename this at some point later I'd guess (since it'd now been merged).

community/kernel-api/src/main/java/org/neo4j/impl/kernel/api/SuspendableCursor.java
fickludd: We introduced this while remembering the use case of someone wanting to scan nodes over multiple transactions.

fickludd:
@MishaDemianenko @tinwelint I have not forgotten about the naming problem, I have just been pre-occupied and undecisive. Two thoughts:

`manual indexes` I am under the impression that there is a PM decision to use this name instead. We (@pontusmelke) have already changed this all over cypher. I cannot back that up though, so maybe that should be verified.

`Edge`. This one has me completely dead-locked. In general I completely agree that there only should be one name for one thing, and if the surface says `relationship` (which is still does AFAIK), that's the name we use. On the other hand, replacing Edge with Relationship in the API makes the code _annoyingly_ much harder to read and is pretty soul-crushing. I'm never the less preparing a PR for that change so you can see what it looks like.

PR #9564, opened 2017-06-26
Speed up data linking in importer by increasing parallelism, and allowing out-of-order batch transfer

chrisvest:
Between certain steps, batches are required to be handed over in-order. Previously, this was ensured by letting steps wait for their turn to deliver their batches. This prevented a lot of parallelism, and hence `parallelRecordReadsWhenWriting` was `false` by default, meaning only a single thread would process those steps to avoid the turn waiting.

Now, a `WorkSync` is inserted in between the steps, and is used to sort the batches before their hand-over. This means the delivering step can now deliver batches out-of-order, which in turn makes it profitable to run it in parallel.

### DIFF
--- a/community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
+++ b/community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
@@ -44,11 +44,14 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.CopyOption;
 import java.nio.file.DirectoryStream;
+import java.nio.file.FileStore;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.OpenOption;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
+import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.BasicFileAttributes;
 
 import static java.nio.file.StandardOpenOption.APPEND;
@@ -68,7 +71,7 @@ public class FileUtils
 
     public static void deleteRecursively( File directory ) throws IOException
     {
-        if ( ! directory.exists() )
+        if ( !directory.exists() )
         {
             return;
         }
@@ -136,12 +139,12 @@ public class FileUtils
         if ( !toMove.exists() )
         {
             throw new FileNotFoundException( "Source file[" + toMove.getAbsolutePath()
-                    + "] not found" );
+                                             + "] not found" );
         }
         if ( target.exists() )
         {
             throw new IOException( "Target file[" + target.getAbsolutePath()
-                    + "] already exists" );
+                                   + "] already exists" );
         }
 
         if ( toMove.renameTo( target ) )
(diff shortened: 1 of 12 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
```diff
@@ -19,6 +19,12 @@
  */
 package org.neo4j.unsafe.impl.batchimport;
 
+import org.apache.commons.lang3.SystemUtils;
+import org.apache.lucene.util.IOUtils;
```
MishaDemianenko: Please no lucene dependecy in here! 🦀

tinwelint:
Looks like some simple checkstyle issue.

A great idea to reduce queuing latency!

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
chrisvest: I was torn between this, and building my own version of the same functionality as a utility method somewhere in the io module. You favour the latter approach?

chrisvest:
@MishaDemianenko I believe I've addressed your comment.

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
```diff
@@ -340,6 +343,86 @@ public static File path( File root, String... path )
...
+            return true;
+        }
+
+        if ( SystemUtils.IS_OS_LINUX )
+        {
+            try
+            {
+                FileStore fileStore = Files.getFileStore( pathOnDevice );
+                String name = fileStore.name();
+                if ( name.equals( "tmpfs" ) || name.equals( "hugetlbfs" ) )
```
MishaDemianenko, line 372: i would flip constants and name here

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
```diff
@@ -340,6 +343,86 @@ public static File path( File root, String... path )
...
+                if ( name.startsWith( "/dev/nvme" ) )
+                {
+                    // This is probably an NVMe device. Anything on that protocol is most likely very fast.
+                    return true;
+                }
+
+                Path device = Paths.get( name ).toRealPath(); // Use toRealPath to resolve any symlinks.
+                Path deviceName = device.getName( device.getNameCount() - 1 );
+                Path sysblock = Paths.get( "/sys/block" );
+                Path rotational = sysblock.resolve( deviceName ).resolve( "queue" ).resolve( "rotational" );
```
MishaDemianenko: would be nice to extract reading the first byte from rotational into a separate method and reuse it inside of further else block

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
```diff
@@ -340,6 +343,86 @@ public static File path( File root, String... path )
...
+     * A high IO device is expected to have negligible seek time, if any, and be able to service multiple IO requests
+     * in parallel.
+     *
+     * @param pathOnDevice Any path, hypothetical or real, that once fully resolved, would exist on a storage device
+     * that either supports high IO, or not.
+     * @param defaultHunch The default hunch for whether the device supports high IO or not. This will be returned if
+     * we otherwise have no clue about the nature of the storage device.
+     * @return Our best-effort estimate for whether or not this device supports a high IO workload.
+     */
+    public static boolean highIODevice( Path pathOnDevice, boolean defaultHunch )
```
MishaDemianenko, line 357: some tests would be nice :)

MishaDemianenko, line 357: I think it would be nice to have some feature flag here or higher during import that will gonna allow us to force highIODevice for cases when we know if high but check say that its not

MishaDemianenko (commented):
sorry, still have a couple of comments related to new helper method, otherwise thx to not using lucene :)

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
chrisvest: Just ImportTool, or also StoreMigrator?

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
chrisvest: That's already how it is. Do you mean building up the path to `rotational`?

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
chrisvest: I'm not a fan of that aesthetic, and it's no change to behaviour since without a name we have nothing to go by.

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
chrisvest: I didn't write automated tests for it because of how environment specific it is. We can try to stage replica environments with e.g. JimFS, though.

chrisvest:
@MishaDemianenko Comments addressed.

MishaDemianenko (approved):
Thx @chrisvest nice! LGTM.

PR #9565, opened 2017-06-26
An assortment of javadoc updates

chrisvest:


jjaderberg (commented):
No way to use `@link` for `EnterpriseGraphDatabaseFactory`? I think static links like `<a href="EnterpriseGraphDatabaseFactory.html">` will get us into trouble.

chrisvest:
@jjaderberg Nope, there's no way to use `{@link …}` without introducing a dependency from community into enterprise, and we can't have that. It's ugly, but I think we'll have to suffer this one, and fix it if it breaks.

jjaderberg (approved):
Well _ready, fire, aim_ then :)

PR #9569, opened 2017-06-27
Wait for open calls to finish on index drop and close

MishaDemianenko:
Currently any concurrent open call to index will cause exception to be throw,
that will cause database panic.
To relax current state a bit and do minimal changes to a way how we handle
state in contract checking index we will wait now for all open calls to finish
before performing drop or close operations.


MishaDemianenko:
@tinwelint can you please take a look, wdyt?

tinwelint (approved):
I don't see a problem doing this.

PR #9570, opened 2017-06-27
Consistency checker now check label scan store for missing entries

tinwelint:
This is take two of #9513 , which got reverted previously.

Besides making consistency checker check the correct things it also clarifies, simplifies and unifies how NodeLabelRange works, which has a side-effect of making the actual fix simpler and working.


community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/reader/LuceneAllEntriesLabelScanReader.java
```diff
@@ -48,28 +49,23 @@ public LuceneAllEntriesLabelScanReader( BoundedIterable<Document> documents, Bit
     public Iterator<NodeLabelRange> iterator()
     {
         final Iterator<Document> iterator = documents.iterator();
-        return new Iterator<NodeLabelRange>()
+        return new IteratorWrapper<NodeLabelRange,Document>( iterator )
```
MishaDemianenko, line 52: we can use `Iterators.map` here instead

MishaDemianenko (approved):
One small comment that can be addressed in separate PR, otherwise looks good!

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/labelscan/reader/LuceneAllEntriesLabelScanReader.java
tinwelint: We could, but I favor `IteratorWrapper` since it's only one object and generally simpler to create.

PR #9572, opened 2017-06-27
Improved JMX disk usage information

klaren:
- Deprected `org.neo4j.jmx.StoreFile` Mbean since the information is incomplete and invalid.
- Added `org.neo4j.jmx.StoreSize` Mbean which includes full information about the current store layout, replacing `org.neo4j.jmx.StoreFile`.
- **Enterprise** Added `org.neo4j.management.CausalClustering` Mbean to provide information about core members. This includes:
  - `getRole()` returning the current role, can be `LEADER`, `FOLLOWER`, `CANDIDATE`. Note that this state is eventual consistent across the cluster.
  - `getRaftLogSize()` returning the total size of the replicated raft log.
  - `getReplicatedStateSize()` returning the size of all other sates, exluding the raft log.
- **Enterprise** Added `getBranchedStoreSize()` to the `org.neo4j.management.BranchedStore` Mbean.


community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
```diff
@@ -628,4 +630,77 @@ public MaybeWindowsMemoryMappedFileReleaseProblem( IOException e )
...
+                @Override
+                public FileVisitResult postVisitDirectory( Path dir, IOException exc )
+                {
+                    return FileVisitResult.CONTINUE; // Ignore exception
+                }
+            } );
+        }
+        catch ( IOException e )
+        {
+            throw new AssertionError( "Exceptions during file size calculations should never happen", e );
```
chrisvest: No, I don't think you can make this assertion. I would just let the IOException bubble up.

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
```diff
@@ -628,4 +630,77 @@ public MaybeWindowsMemoryMappedFileReleaseProblem( IOException e )
...
+
+    /**
+     * Calculates the size of a given directory or file.
+     *
+     * @param path to the file or directory.
+     * @return the size, in bytes, of the file or the total size of the content in the directory, including
+     * subdirectories.
+     */
+    public static long size( File path )
+    {
```
chrisvest: Do we really need both implementations?

community/jmx/src/main/java/org/neo4j/jmx/StoreSize.java
```diff
@@ -0,0 +1,60 @@
...
+    long getStringStoreSize();
+
+    @Description( "Disk space used to store array properties, in bytes." )
+    long getArrayStoreSize();
+
+    @Description( "Disk space used to store labels, in bytes" )
+    long getLabelStoreSize();
+
+    @Description( "Disk space used to store counters, in bytes" )
+    long getCountStoreSize();
```
chrisvest, line 50: I don't like that we are breaking it down to such detail – especially since we're baking it into an API here. What are we going to do with this if we change the number of files we have in our store, or inline one store file into another?

I would boil it down to store, logs, index/, schema/, and total.

chrisvest: I think it was a mistake that the old bean did this, so let's right our course now that we're deprecating the old bean anyway.

community/jmx/src/main/resources/META-INF/services/org.neo4j.kernel.KernelExtension
```diff
@@ -1 +0,0 @@
-org.neo4j.jmx.impl.JmxExtension
```
chrisvest: Should this have said `org.neo4j.jmx.impl.JmxKernelExtension`?

community/jmx/src/test/java/org/neo4j/jmx/impl/StoreSizeBeanTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+        // Start DataSourceManager
+        dataSourceManager.register( dataSource );
+        dataSourceManager.start();
+
+        // Create bean
+        KernelData kernelData = new DefaultKernelData( fs, mock( PageCache.class ), storeDir, Config.empty(), db );
+        ManagementData data = new ManagementData( new StoreSizeBean(), kernelData, ManagementSupport.load() );
+        storeSizeBean = (StoreSize) new StoreSizeBean().createMBean( data );
+    }
```
chrisvest, line 110: I would use a `DatabaseRule` to create a real database instance, such that the test can be written with as little reliance on internal APIs and implementation details as possible.

As it is now, this test is relying on many internal APIs and implementation details, and it will easily break if these details change in ways that would otherwise preserve behaviour. It could likewise mask breakage from refactorings in the product, because the environment in this test is so artificial.

chrisvest (changes requested):
There's a couple of changes I'd like to see.

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
klaren: Both were used at one point but not anymore. I'll remove the unused one.

community/io/src/main/java/org/neo4j/io/fs/FileUtils.java
klaren: `walkFileTree` is guaranteed to not throw IOException unless the visitor does. This will be removed anyway since the method is not used anymore.

community/jmx/src/main/java/org/neo4j/jmx/StoreSize.java
klaren: I totally agree with you. However, the purpose of this PR is to give more details of the files as it has been requested by users.

community/jmx/src/main/resources/META-INF/services/org.neo4j.kernel.KernelExtension
klaren: There is no service provider for `KernelExtension` so this looks like a relic of the past. While verifying this I found 9 more leftovers that I will include in a separate PR.

community/jmx/src/test/java/org/neo4j/jmx/impl/StoreSizeBeanTest.java
klaren: While it would be better to have a less artificial scenario it's hard to deceive a reproducible test case that would catch any meaningful errors. The current test basically verifies the grouping of the files and have been renamed to reflect this.

klaren:
@chrisvest ready for a new round of comments

chrisvest (approved):

PR #9573, opened 2017-06-27
Add additional procedures to access transaction metadata

MishaDemianenko:
Add additional standard procedure `dbms.getTXMetaData` to access transactional metadata.
Provide demo usage example as test.


MishaDemianenko:
@OliviaYtterbrink can you please take a look, thx in advance!

MishaDemianenko:
Suppressed by #9652

PR #9574, opened 2017-06-27
Cleanup confusing import tool config related options.

MishaDemianenko:
Before import tool had 2 options to provide config:
 - db-config (good old import tool option)
 - additional-config (new option that is aligned with admin tool)
It was already confusing to have both of them to do the same thing,
in addition to that `additional-config` was not used at all.

This PR align Import tool and admin tool by using new `additional-config`
option and by deprecating old `db-config` option.
For now configs from ищер options will be merged with preference to configs
from `additional-config` for conflicting options.


MishaDemianenko:
@spacecowboy @tinwelint can you guys take a look, thx in advance!

tinwelint (approved):
Nice. I wonder how we got into this previous state

spacecowboy (approved):
LGTM :+1:

PR #9576, opened 2017-06-28
introduce 'storable' package in values module

fickludd:
Improves class organization and aligns packages with the actual class hierachy


MishaDemianenko (approved):
Without any insights, do not have any objections :)

PR #9580, opened 2017-06-28
Convenient PageCache rule

tinwelint:
Managing PageCache, FileSystemAbstraction and TestDirectory. This conveniently gathered into one rule instead of creating all thread and doing the RuleChain thing. Now it'll be much more fun writing these tests.

This commit was previously part of #9227 , but got reverted in #9547 so I decided to pull this commit out as a separate PR to get it in.


chrisvest (approved):
👍

PR #9581, opened 2017-06-28
Limit and aggregation cannot be combined

pontusmelke:
In the compiled runtime aggregation and limit are not playing nice
with each other, disable all queries that combine limit and aggregation
to be run  on the compiled runtime for now.

changelog: Fix for queries combining limit and aggregation in the compiled runtime.


fickludd (approved):
Nice!

PR #9584, opened 2017-06-28
Put import tool logs inside logs directory

MishaDemianenko:
Update way how path of import tool logs is constructed.
Before logs directory was missed and log file endup inside what ever
directory was current for importer.
This PR fix that behaviour and by put log file inside directory
defined by `logs_directory` setting.

Cleanup import tool tests

### DIFF
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -76,6 +76,7 @@ import org.neo4j.unsafe.impl.batchimport.input.csv.IdType;
 import org.neo4j.unsafe.impl.batchimport.staging.ExecutionMonitors;
 
 import static java.nio.charset.Charset.defaultCharset;
+import static org.neo4j.graphdb.factory.GraphDatabaseSettings.logs_directory;
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.store_internal_log_path;
 import static org.neo4j.helpers.Exceptions.launderedException;
 import static org.neo4j.helpers.Format.bytes;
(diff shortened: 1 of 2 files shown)


MishaDemianenko:
@spacecowboy can you please take a look on this as well, thx!

spacecowboy:
@MishaDemianenko I'm not sure this is wanted. Why do you want the import logs to end up in the database logs directory?

The import tool is a standalone tool and its output should not be mixed with database logs. I'd say having the logs put in your current working directory is the expected behavior.

MishaDemianenko:
@spacecowboy I was annoyed by logs/debug.log file that will be produced by tests and be placed in a module root directory resulting in git start tracking them.

Why do you think it's useful to have those logs somewhere else? Isn't that confusing when logs are placed relative to current directory only for import?
since for example `BackupService::fullBackup` will gonna place logs in path relative to `logs_directory`

spacecowboy:
So the motivation is that when something goes wrong, you don't want to go hunting for a log file in some strange directory (`/var/log/neo4j`). I'd expect all information to be right in front of me, which is either printed to screen directly or a file in my current directory.

Frankly, I don't think we should write any logs by default in the import tool *at all*. There shouldn't even be a debug log written by default in the import tool since only developers care about that.

So I'm fine with the change since I consider the current situation to be wrong anyway.

spacecowboy (approved):

tinwelint (approved):
I've been annoyed by this as well. I partially agree with @spacecowboy that perhaps import shouldn't write a log at all, except that it may actually be nice for some reason to see in the log whether or not a db was created with the importer...

So for me it's also a yes

PR #9593, opened 2017-06-29
List properties and index in compiled runtime

pontusmelke:
The compiled runtime wasn't doing the conversion from Cypher lists
to arrays supported by the stores (primitive arrays and String arrays),
before trying to access the index. This lead to errors when trying to
access nodes with array properties from Cypher, e.g.

```
MATCH (n:L) WHERE n.prop = [1,2,3] RETURN n
```

failed whenever there was an index on `:L(prop)`.

changelog: Fix an error that occurred for queries trying to query for an indexed list property in the compiled runtime.

### DIFF
--- a/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/codegen/GeneratedMethodStructure.scala
+++ b/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/codegen/GeneratedMethodStructure.scala
@@ -1370,20 +1370,6 @@ class GeneratedMethodStructure(val fields: Fields, val generator: CodeBlock, aux
     }
   }
 
-  override def indexUniqueSeek(nodeVar: String, descriptorVar: String, value: Expression, codeGenType: CodeGenType) = {
-    val predicate = generator.declare(typeRef[IndexQuery.ExactPredicate], s"${nodeVar}Query")
-    val local = generator.declare(typeRef[Long], nodeVar)
-    val boxedValue = if (codeGenType.isPrimitive) Expression.box(value) else value
-    handleKernelExceptions(generator, fields.ro, _finalizers) { body =>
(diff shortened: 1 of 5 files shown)


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/codegen/GeneratedMethodStructure.scala
```diff
@@ -1357,7 +1357,9 @@ class GeneratedMethodStructure(val fields: Fields, val generator: CodeBlock, aux
   override def indexSeek(iterVar: String, descriptorVar: String, value: Expression, codeGenType: CodeGenType) = {
     val predicate = generator.declare(typeRef[IndexQuery], s"${iterVar}Query")
     val local = generator.declare(typeRef[PrimitiveLongIterator], iterVar)
-    val boxedValue = if (codeGenType.isPrimitive) Expression.box(value) else value
+    val boxedValue =
```
OliviaYtterbrink, line 1360: Shouldn't this be for indexUniqueSeek as well?

OliviaYtterbrink (commented):
I like this, just have one question.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_2/codegen/GeneratedMethodStructure.scala
pontusmelke: Confusingly we no longer use this method, when reading from the index it doesn't matter what kind of index it is. Or it does matter in the unique variant we take locks that are not necessary when reading from the index. We removed the unique variant, here https://github.com/neo4j/neo4j/pull/9521/files but we forgot to remove the method. I removed the unused method in a new commit.

PR #9598, opened 2017-06-29
HA IPv6 support

martinfurmanski:


enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/SwitchToMaster.java
```diff
@@ -123,17 +125,26 @@ public URI switchToMaster( LifeSupport haCommunicationLife, URI me )
...
         String hostname = config.get( HaSettings.ha_server ).getHost();
-        if ( hostname == null || hostname.contains( "0.0.0.0" ) )
+        InetSocketAddress masterSocketAddress = masterServer.getSocketAddress();
+
+        if ( hostname == null || isWildcard( hostname ) )
         {
-            String masterAddress = ServerUtil.getHostString( masterServer.getSocketAddress() );
-            hostname = masterAddress.contains( "0.0.0.0" ) ? me.getHost() : masterAddress;
+            InetAddress masterAddress = masterSocketAddress.getAddress();
+            hostname = masterAddress.isAnyLocalAddress() ? me.getHost() : ServerUtil.getHostString( masterSocketAddress );
```
spacecowboy, line 134: Does `isAnyLocalAddress` cover the `0.0.0.0` case? The test below suggests that it does, but just want to make sure

enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/SwitchToSlave.java
```diff
@@ -334,11 +335,23 @@ private boolean catchUpWithMaster( UpdatePuller updatePuller ) throws IllegalArg
...
+            host = "[" + host + "]";
+        }
+
         InstanceId serverId = config.get( ClusterSettings.server_id );
-        String host = hostString.contains( HighAvailabilityModeSwitcher.INADDR_ANY ) ? me.getHost() : hostString;
-        return URI.create( "ha://" + host + ":" + port + "?serverId=" + serverId );
+        return URI.create( "ha://" + host + ":" + serverSocketAddress.getPort() + "?serverId=" + serverId );
+    }
+
+    private static boolean isWildcard( String hostString )
```
spacecowboy: Duplicate function, please de-duplicate :)

enterprise/ha/src/test/java/org/neo4j/kernel/ha/HaIPv6ConfigurationTest.java
```diff
@@ -0,0 +1,62 @@
...
+    public void testClusterWithLocalhostAddresses() throws Throwable
+    {
+        GraphDatabaseService db = new HighlyAvailableGraphDatabaseFactory()
+                .newEmbeddedDatabaseBuilder( dir.makeGraphDbDir() )
+                .setConfig( ClusterSettings.cluster_server, "[::1]:5000" )
+                .setConfig( ClusterSettings.initial_hosts, "[::1]:5000" )
+                .setConfig( HaSettings.ha_server, "[::1]:6000" )
+                .setConfig( ClusterSettings.server_id, "1" )
+                .newGraphDatabase();
+
```
spacecowboy: I know you're just making sure there is no crash or such, but could you add something which actually uses the database? Create a node or something basic, just to really really verify that it works

enterprise/ha/src/test/java/org/neo4j/kernel/ha/HaIPv6ConfigurationTest.java
```diff
@@ -0,0 +1,62 @@
...
+    public void testClusterWithWildcardAddresses() throws Throwable
+    {
+        GraphDatabaseService db = new HighlyAvailableGraphDatabaseFactory()
+                .newEmbeddedDatabaseBuilder( dir.makeGraphDbDir() )
+                .setConfig( ClusterSettings.cluster_server, "[::]:5000" )
+                .setConfig( ClusterSettings.initial_hosts, "[::1]:5000" )
+                .setConfig( HaSettings.ha_server, "[::]:6000" )
+                .setConfig( ClusterSettings.server_id, "1" )
+                .newGraphDatabase();
+
```
spacecowboy: Same comment as above

spacecowboy (approved):
Looks good so far.

Have you actually tested the changes by standing up a cluster in an IPv6(-only?) environment?

enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/SwitchToSlave.java
martinfurmanski: It is actually subtly different. One checks a String straight out of config where I expect the [ ] format whereas the other one checks an address without the square brackets. I know it is crazy to have this code in this place this far down in the stack and duplicated, but I do not want to attempt a refactoring and cleanup.

enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/SwitchToMaster.java
martinfurmanski: I believe it does, yes.

enterprise/ha/src/test/java/org/neo4j/kernel/ha/HaIPv6ConfigurationTest.java
martinfurmanski: Ok! I actually had that in place for my own sanity but removed it prior to pushing. I will add it back.

enterprise/ha/src/test/java/org/neo4j/kernel/ha/HaIPv6ConfigurationTest.java
martinfurmanski: Same answer as above

martinfurmanski:
@spacecowboy Not an IPv6-only environment, but local testing with IPv6 only configuration on my MBP, i.e. [::] and [::1] as ha.host.coordination and ha.host.data. All basic functionality seems to be working. System test pending.

spacecowboy:
@martinfurmanski ok good. before merge I would like us to do a test using docker (or aws if you prefer that) to make sure we try it outside of a localhost environment

martinfurmanski:
@spacecowboy That would be amazing, thanks!

digitalstain (approved):

spacecowboy:
Running tests in a docker network, so far the database fails to bind to the ipv6 address. Will investigate why.

spacecowboy:
Some testing:
IPV6 with globally routed address works
IPV6 with link-local address fails

@martinfurmanski already has a plan to fix the link-local problem (URI mangles the interface specification https://en.wikipedia.org/wiki/IPv6_address#Link-local_addresses_and_zone_indices)

martinfurmanski:
Updated for link local address support in URIs.

spacecowboy (approved):
Tested and now it works with link-local address

PR #9605, opened 2017-06-30
Invokes afterRollback on tx event handlers with failing beforeCommit

### LINKED ISSUE #2660: afterRollback not always called when TX tolls back
When a transaction rolls back due to constraint violations, afterRollback() on TransactionEventHandler is called, as illustrated by this passing test:

```
@Test
public void afterRollbackShouldBeCalledOnHandlers() {
    GraphDatabaseService database = new TestGraphDatabaseFactory().newImpermanentDatabase();

    try (Transaction tx = database.beginTx()) {
        Node node1 = database.createNode();
        Node node2 = database.createNode();
        node1.createRelationshipTo(node2, DynamicRelationshipType.withName("TEST"));
        tx.success();
    }

    final AtomicBoolean called = new AtomicBoolean(false);

    database.registerTransactionEventHandler(new TransactionEventHandler.Adapter<Void>() {
        @Override
        public void afterRollback(TransactionData data, Void state) {
            called.set(true);
        }
    });

    try {
        try (Transaction tx = database.beginTx()) {
            database.getNodeById(0).delete();
            tx.success();
        }

        fail("This tx should not succeed");
    } catch (RuntimeException e) {
        //ok
    }

    assertTrue("Handler wasn't called!", called.get());
}
```

However, when the client code throws a RuntimeException in the middle of a transaction or in beforeCommit(), afterRollback() isn't called, although the transaction is rolled back. Illustrated by the following failing test:

```
@Test
public void afterRollbackShouldBeCalledOnHandlers2() {
    GraphDatabaseService database = new TestGraphDatabaseFactory().newImpermanentDatabase();

    final AtomicBoolean called = new AtomicBoolean(false);

    database.registerTransactionEventHandler(new TransactionEventHandler.Adapter<Void>() {
        @Override
        public Void beforeCommit(TransactionData data) throws Exception {
            throw new RuntimeException();
        }

        @Override
        public void afterRollback(TransactionData data, Void state) {
            called.set(true);
        }
    });

    try {
        try (Transaction tx = database.beginTx()) {
            database.createNode();
            tx.success();
        }

        fail("This tx should not succeed");
    } catch (RuntimeException e) {
        //ok
    }

    assertTrue("Handler wasn't called!", called.get());
}
```

Neo4j 2.1.2

tinwelint:
Previously only those TransactionEventHandlers with successful call to
beforeCommit would see a call to afterRollback. However such information
may be interesting to all event handlers, even failing ones.

The contractual difference here is that the state argument will be null
in such cases.

Fixes #2660

### DIFF
--- a/community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionEventHandler.java
+++ b/community/graphdb-api/src/main/java/org/neo4j/graphdb/event/TransactionEventHandler.java
@@ -113,8 +113,9 @@ public interface TransactionEventHandler<T>
      * can provide will require a new transaction to be opened.
      *
      * @param data the changes that were attempted to be committed in this transaction.
-     * @param state the object returned by
-     *            {@link #beforeCommit(TransactionData)}.
+     * @param state the object returned by {@link #beforeCommit(TransactionData)}.
+     * If this handler failed when executing {@link #beforeCommit(TransactionData)} this
+     * {@code state} will be {@code null}.
      */
     // TODO: should this method take a parameter describing WHY the tx failed?
     void afterRollback( TransactionData data, T state );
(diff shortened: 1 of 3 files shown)


tinwelint:
Replaces #9601

MishaDemianenko (approved):

PR #9608, opened 2017-06-30
Check validity of snapshot filesets based on `segments` files

MishaDemianenko:
Remove expectation of lucene label scan store to be included into set of
lucene indexes that we expect to have snapshots for.
Check snapshots validity based on segments file as an indicator what
index commit was used for particular index snapshot.

### DIFF
--- a/community/neo4j/src/test/java/org/neo4j/index/backup/IndexBackupIT.java
+++ b/community/neo4j/src/test/java/org/neo4j/index/backup/IndexBackupIT.java
@@ -20,6 +20,7 @@
 package org.neo4j.index.backup;
 
 import org.apache.commons.io.FilenameUtils;
+import org.apache.lucene.index.IndexFileNames;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


community/neo4j/src/test/java/org/neo4j/index/backup/IndexBackupIT.java
```diff
@@ -249,4 +253,9 @@ private DependencyResolver getDatabaseResolver()
     {
         return database.getDependencyResolver();
     }
+
+    private boolean segmentsFilePredicate( String fileName )
+    {
+        return fileName.contains( "segments_" );
```
klaren: Since we compare absolute paths here I think it's safer to use
```
return new File(fileName).getName().startsWith( org.apache.lucene.index.IndexFileNames.SEGMENTS );
```
and avoid false positives.

community/neo4j/src/test/java/org/neo4j/index/backup/IndexBackupIT.java
```diff
@@ -146,20 +145,25 @@ public void snapshotFilesDeletedWhenSnapshotReleased() throws IOException
     private void compareSnapshotFiles( Set<String> firstSnapshotFileNames, Set<String> secondSnapshotFileNames,
             FileSystemAbstraction fileSystem )
     {
-        assertThat(
-                format( "Should have at least %d modified index files. Snapshot files  are: %s", NUMBER_OF_INDEXES + 1,
-                        firstSnapshotFileNames ), firstSnapshotFileNames,
-                hasSize( greaterThanOrEqualTo( NUMBER_OF_INDEXES + 1 ) ) );
-        for ( String fileName : firstSnapshotFileNames )
+        Set<String> firstSnapshotIndexesSegments = firstSnapshotFileNames.stream()
+                        .filter( this::segmentsFilePredicate )
```
klaren: The filter might be moved to `getFileNames` instead. Not sure what other implications this has for the rest of the test though.

klaren (commented):
Two small observations.

MishaDemianenko:
@klaren thx, updated. Great suggestions!

klaren (approved):
Looks good now!

PR #9614, opened 2017-07-03
Parallel execution of server based tests

MishaDemianenko:
Allow server to be started on random http/https port.
Update tests to be able to resolve port on which server is running.
Register http and https connector ports in port register component.

### DIFF
--- a/community/server/src/main/java/org/neo4j/server/web/Jetty9WebServer.java
+++ b/community/server/src/main/java/org/neo4j/server/web/Jetty9WebServer.java
@@ -354,14 +354,14 @@ public class Jetty9WebServer implements WebServer
     @Override
     public InetSocketAddress getLocalHttpAddress()
     {
-        return new InetSocketAddress( serverConnector.getHost(), serverConnector.getLocalPort() );
+        return toSocketAddress( serverConnector );
     }
 
     @Override
     public InetSocketAddress getLocalHttpsAddress()
     {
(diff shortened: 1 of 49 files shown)


MishaDemianenko:
@lutovich can you please take a look? thx!
cc @lassewesth

lassewesth:
@MishaDemianenko Wow, thanks massive, well done, looks good!

community/server/src/main/java/org/neo4j/server/web/Jetty9WebServer.java
```diff
@@ -375,6 +351,20 @@ protected void startJetty() throws Exception
         }
     }
 
+    @Override
+    public InetSocketAddress getLocalHttpAddress()
+    {
+        return new InetSocketAddress( serverConnector.getHost(), serverConnector.getLocalPort() );
```
lutovich: Maybe extract a method like `InetSocketAddress toSocketAddress(ServerConnector)` to use here and in `#getLocalHttpsAddress()`?

lutovich (approved):
@MishaDemianenko looks great!

community/server/src/main/java/org/neo4j/server/web/Jetty9WebServer.java
MishaDemianenko: @lutovich absolutely, thx! updated!

PR #9616, opened 2017-07-04
Make CountsTracker rotation timeout robust against drift

chrisvest:
Hopefully this will also make the AbstractKeyValueStoreTest less flaky on very resource strained systems, but just to be sure I have also increased the timeout by a factor 10.


burqen (dismissed):
Changes look good, but please have a look at the build failure.

PR #9617, opened 2017-07-04
ListValue.equals(ArrayValue)

SaschaPeukert:
This includes tests and an implementation for an equality check between ListValues and ArrayValues if they contain the same values (with the same ordering).
Introduced SequenceValue as an interface that is implemented by ListValue and ArrayValue. Actual equality check is statically done in AnyValues.

### DIFF
--- a/community/values/src/main/java/org/neo4j/values/AnyValues.java
+++ b/community/values/src/main/java/org/neo4j/values/AnyValues.java
@@ -32,4 +32,23 @@ public final class AnyValues
      */
     public static final Comparator<AnyValue> COMPARATOR =
             new AnyValueComparator( Values.COMPARATOR, VirtualValueGroup::compareTo );
+
+    public static boolean equalityOfValuesInSequences( SequenceValue a, SequenceValue b )
+    {
+        if ( a.length() != b.length() )
+        {
+            return false;
+        }
+
+        for ( int i = 0; i < a.length(); i++ )
+        {
+            AnyValue myValue = a.value( i );
+            AnyValue otherValue = b.value( i );
+            if ( myValue.equals( otherValue ) == false )
+            {
+                return false;
+            }
+        }
+        return true;
+    }
 }
--- a/community/values/src/main/java/org/neo4j/values/AnyValues.java
+++ b/community/values/src/main/java/org/neo4j/values/AnyValues.java
@@ -32,23 +32,4 @@ public final class AnyValues
      */
     public static final Comparator<AnyValue> COMPARATOR =
             new AnyValueComparator( Values.COMPARATOR, VirtualValueGroup::compareTo );
-
-    public static boolean equalityOfValuesInSequences( SequenceValue a, SequenceValue b )
-    {
-        if ( a.length() != b.length() )
-        {
-            return false;
-        }
-
-        for ( int i = 0; i < a.length(); i++ )
-        {
-            AnyValue myValue = a.value( i );
-            AnyValue otherValue = b.value( i );
-            if ( myValue.equals( otherValue ) == false )
-            {
-                return false;
-            }
-        }
-        return true;
-    }
 }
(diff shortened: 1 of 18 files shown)


community/values/src/main/java/org/neo4j/values/AnyValues.java
```diff
@@ -32,4 +32,23 @@
...
+        if ( a.length() != b.length() )
+        {
+            return false;
+        }
+
+        for ( int i = 0; i < a.length(); i++ )
+        {
+            AnyValue myValue = a.value( i );
+            AnyValue otherValue = b.value( i );
+            if ( myValue.equals( otherValue ) == false )
```
MishaDemianenko: this line should be simplified to `!myValue.equals( otherValue )`

community/values/src/main/java/org/neo4j/values/AnyValues.java
```diff
@@ -32,4 +32,23 @@
      */
     public static final Comparator<AnyValue> COMPARATOR =
             new AnyValueComparator( Values.COMPARATOR, VirtualValueGroup::compareTo );
+
+    public static boolean equalityOfValuesInSequences( SequenceValue a, SequenceValue b )
```
MishaDemianenko: It looks like this method is responsible for equality check for all possible sequence and it's quite suites as the default implementation of aquals in SequenceValue instead.

community/values/src/main/java/org/neo4j/values/SequenceValue.java
```diff
@@ -0,0 +1,29 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.values;
+
+public interface SequenceValue
```
MishaDemianenko: Missing java doc.

community/values/src/main/java/org/neo4j/values/VirtualValue.java
```diff
@@ -33,6 +33,10 @@
     @Override
     public final boolean equals( Object other )
     {
+        if ( (other instanceof SequenceValue) && (this instanceof SequenceValue) )
```
MishaDemianenko: the whole goal of new values API was to remove casts as much as possible from all the equality methods, this change does not follow that paradigm (this place and all other updated equals methods) is there is any way we can avoid that?

community/values/src/main/java/org/neo4j/values/virtual/ListValue.java
```diff
@@ -118,11 +117,18 @@ public String toString()
         return sb.toString();
     }
 
-    public int size()
+    public int length()
```
MishaDemianenko: Missing`@Override` annotation here

MishaDemianenko (changes requested):
Couple of small comments and questions, also since I'm mostly on the side of this work would really like if @fickludd would take a look at proposed changes as soon as will have time

community/values/src/main/java/org/neo4j/values/VirtualValue.java
```diff
@@ -29,15 +29,28 @@
...
-        if ( (other instanceof SequenceValue) && (this instanceof SequenceValue) )
+        if ( other == null )
+        {
+            return false;
+        }
+
+        // suppose that we will only ever call equals with Objects of AnyValue but just in case
+        try
+        {
+            if ( ((AnyValue) other).isSequenceValue() && this.isSequenceValue() )
```
MishaDemianenko: doing
```
try
{
(AnyValue) other)
}
catch ( ClassCastException e )
{
return false
}
```
is much more slower and not that easy to read compared to using instanceOf operator, so would suggest here to use instanceOf AnyValue instead. 
 This also applies to ArrayValue `equals`

MishaDemianenko (commented):
One comment regarding casting and catching class cast otherwise looks good!

MishaDemianenko (approved):

PR #9619, opened 2017-07-05
Comparing NaN with any other number will return an unordered result

OliviaYtterbrink:


OliviaYtterbrink:
Replaced by https://github.com/neo4j/neo4j/pull/9661

PR #9627, opened 2017-07-06
Allow replicated id generator factory to create empty file when requested

MishaDemianenko:
Add possibility to create empty generator file from the corresponding factory
to allow id file recreation in case if a rebuild of id generators is required.


klaren (approved):
Nice to have a test for the new generator!
Well done.

PR #9629, opened 2017-07-06
Syntax and AST for Multiple Graphs Cypher

Mats-SX:
The syntax reflects the changes proposed in https://github.com/opencypher/openCypher/pull/241


Mats-SX:
Extracted two commits into #9691 to allow them to go in while dealing with the difficult projection clause changes.

PR #9634, opened 2017-07-06
Fixed NullPointerException in distance function

SaschaPeukert:
Whenever any value of a given point is null, the application of the distance function on that point should yield null from now on

### DIFF
--- a/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
+++ b/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
@@ -185,47 +185,21 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with NewPla
     result.toList should equal(List(Map("dist" -> null)))
   }
 
-  test("distance function should return null with null in any kind of point on rhs") {
+  test("distance function should return null if a point is null") {
     var result = executeWithAllPlanners(
-      """return distance(point({latitude:3,longitude:7}),point({latitude:null,
-        |longitude:3})) as dist;""".stripMargin)
+      "RETURN distance(point({latitude:3,longitude:7}),point({latitude:null, longitude:3})) as dist;")
     result.toList should equal(List(Map("dist" -> null)))
 
     result = executeWithAllPlanners(
-      """return distance(point({latitude:3,longitude:7}),point({latitude:3,
-        |longitude:null})) as dist;""".stripMargin)
+      "RETURN distance(point({latitude:3,longitude:null}),point({latitude:7, longitude:3})) as dist;")
     result.toList should equal(List(Map("dist" -> null)))
 
     result = executeWithAllPlanners(
-      """return distance(point({x:3,y:7}),point({x:null,
-        |y:3})) as dist;""".stripMargin)
+      "RETURN distance(point({x:3,y:7}),point({x:null, y:3})) as dist;")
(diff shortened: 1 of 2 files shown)


community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/commands/expressions/PointFunction.scala
```diff
@@ -19,18 +19,22 @@
...
+import org.neo4j.cypher.internal.compiler.v3_1.{ExecutionContext, _}
 import org.neo4j.cypher.internal.frontend.v3_1.CypherTypeException
 import org.neo4j.cypher.internal.frontend.v3_1.symbols._
 
 case class PointFunction(data: Expression) extends NullInNullOutExpression(data) {
 
   override def compute(value: Any, ctx: ExecutionContext)(implicit state: QueryState): Any = value match {
     case IsMap(mapCreator) =>
-      Points.fromMap(mapCreator(state.query))
+      if(mapCreator(state.query).exists(_._2==null)){
```
craigtaverner: Style preference includes spaces

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/commands/expressions/PointFunction.scala
```diff
@@ -19,18 +19,22 @@
...
+        null
+      } else {
+        Points.fromMap(mapCreator(state.query))
```
craigtaverner: mapCreator called twice, perhaps it would be better to have Points.fromMap return null when passed maps containing any null values?

community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -185,6 +185,50 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with NewPla
     result.toList should equal(List(Map("dist" -> null)))
   }
 
+  test("distance function should return null with null in any kind of point on rhs") {
+    var result = executeWithAllPlanners(
+      """return distance(point({latitude:3,longitude:7}),point({latitude:null,
```
craigtaverner: All tests are for distance(), but the implementation is actually on `point()`. It would be good to have a test on point function returning null. And then you also won't need as many tests.

community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -185,6 +185,50 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with NewPla
...
+        |longitude:3})) as dist;""".stripMargin)
+    result.toList should equal(List(Map("dist" -> null)))
+
+    result = executeWithAllPlanners(
+      """return distance(point({latitude:3,longitude:7}),point({latitude:3,
+        |longitude:null})) as dist;""".stripMargin)
```
craigtaverner: Choice of the column to wrap on looks a little odd, perhaps one line is better, with normal quotes and no stripMargin?

craigtaverner (changes requested):

craigtaverner (approved):

PR #9635, opened 2017-07-06
Use file for storeDir representation if GraphDatabaseFacade API.

MishaDemianenko:
Use file instead of string to represent store dir,
cleanup all the wrapping and converting around that


chrisvest (approved):
Looks good. Just be aware that it introduces a compilation failure in HAR, which will have to be dealt with when you merge.

MishaDemianenko:
@chrisvest thx, yes I saw that :)

PR #9640, opened 2017-07-07
Make configuration values mutable during runtime, Part 1

klaren:
This is a preparatory PR for making the settings mutable during runtime. 

- Settings and configuration API have been reduced to a smaller surface to simplify thread-safety.
- `Config` is now augmented instead of copied everywhere, thus everyone refers to the same object, to allow a single source of configuration values that can be polled or subscribed to at a later stage.
- `Setting` has a new property, `dynamic`, that indicates whether it can be changed at runtime or not. 

Three settings have been changed to use polling, namely:
- dbms.transaction.timeout
- dbms.logs.query.threshold
- dbms.logs.query.enabled

When changed it will only affect newly created transaction and queries. Even though this is the case, there is not yet any way of changing the actual values from the client side and this will come in part 2 of this feature. This PR is proposed to be merged now due to the massive changeset.

This PR breaks some tools that have to be fixed after this is merged, separate PRs for affected projects is already in place.


tinwelint (approved):
There's a lot of changes and whereas it feels hard to fully get into every little detail it looks proper and in the right direction

tinwelint:
@klaren a rebase would be in order too

community/neo4j/pom.xml
```diff
@@ -134,6 +134,11 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>annotations</artifactId>
+      <scope>test</scope>
```
MishaDemianenko: not sure why scope of this dependency is different compare with other findbugs annotations?

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
```diff
@@ -296,19 +293,11 @@ private String renameExistingBackup( final Path folder, final String oldName ) t
...
-
     private Config loadConfig( Optional<Path> additionalConfig ) throws CommandFailed
     {
-        //noinspection unchecked
         return withAdditionalConfig( additionalConfig,
-                ConfigLoader
-                        .loadConfigWithConnectorsDisabled(
-                                Optional.of( homeDir.toFile() ),
-                                Optional.of( configDir.resolve( "neo4j.conf" ).toFile() ) ) );
+                Config.fromFile( configDir.resolve( "neo4j.conf" ) ).withHome( homeDir )
```
MishaDemianenko: Config class have constant defined for default config file name

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupSettings.java
```diff
@@ -36,9 +36,9 @@
 public class OnlineBackupSettings implements LoadableConfig
 {
     @Description( "Enable support for running online backups" )
-    public static final Setting<Boolean> online_backup_enabled = setting( "dbms.backup.enabled", BOOLEAN, TRUE );
+    public static final Setting<Boolean> online_backup_enabled = setting( "dbms.backup.enabled", BOOLEAN, TRUE ).build();
```
MishaDemianenko: I do not think we need to add those 'build()' part everywhere settings are defined right?

enterprise/backup/src/main/java/org/neo4j/restore/RestoreDatabaseCli.java
```diff
@@ -55,10 +51,10 @@ public RestoreDatabaseCli( Path homeDir, Path configDir )
 
     private static Config loadNeo4jConfig( Path homeDir, Path configDir, String databaseName )
     {
-        Config config = ConfigLoader.loadConfigWithConnectorsDisabled( Optional.of( homeDir.toFile() ),
-                Optional.of( configDir.resolve( "neo4j.conf" ).toFile() ) );
-
-        return config.with( stringMap( DatabaseManagementSystemSettings.active_database.name(), databaseName ) );
+        return Config.fromFile( configDir.resolve( "neo4j.conf" ) )
```
MishaDemianenko: default config file name is defined in Config class already as a constant

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusterConfigurationValidator.java
```diff
@@ -38,41 +36,36 @@
...
+            validateBoltConnector( config );
+            validateLoadBalancing( config, log );
         }
 
-        validateInitialDiscoveryMembers( rawConfig::get );
-        validateBoltConnector( rawConfig );
-        validateLoadBalancing( rawConfig, log );
-
-        return rawConfig;
+        return Collections.emptyMap();
```
MishaDemianenko, line 50: not sure why we need to return empty map now at all

enterprise/causal-clustering/src/main/java/org/neo4j/commandline/dbms/UnbindFromClusterCommand.java
```diff
@@ -65,16 +60,9 @@ static Arguments arguments()
...
-                Optional.of( configDir.resolve( "neo4j.conf" ).toFile() ) );
-        Map<String,String> additionalConfig = new HashMap<>();
-        additionalConfig.put( DatabaseManagementSystemSettings.active_database.name(), databaseName );
-        return config.with( additionalConfig );
-    }
-
-    private static List<Class<?>> settings()
-    {
-        return Arrays.asList( GraphDatabaseSettings.class, DatabaseManagementSystemSettings.class );
+        return fromFile( configDir.resolve( "neo4j.conf" ) )
```
MishaDemianenko: Constant for default config file name already defined in Config class

enterprise/cluster/src/main/java/org/neo4j/configuration/HaConfigurationValidator.java
```diff
@@ -39,36 +38,30 @@
...
+            validateServerId( config );
+            validateInitialHosts( config );
         }
 
-        validateServerId( rawConfig::get );
-
-        validateInitialHosts( rawConfig::get );
-
-        return rawConfig;
+        return Collections.emptyMap();
```
MishaDemianenko, line 51: here empty map as result always as well

community/dbms/src/main/java/org/neo4j/commandline/dbms/DumpCommand.java
```diff
@@ -101,12 +99,11 @@ public void execute( String[] args ) throws IncorrectUsage, CommandFailed
 
     private Path toDatabaseDirectory( String databaseName )
     {
-        //noinspection unchecked
-        return ConfigLoader
-                .loadConfigWithConnectorsDisabled( Optional.of( homeDir.toFile() ),
-                        Optional.of( configDir.resolve( "neo4j.conf" ).toFile() ) )
-                .with( stringMap( DatabaseManagementSystemSettings.active_database.name(), databaseName ) )
-                .get( database_path ).toPath();
+        return Config.fromFile( configDir.resolve( "neo4j.conf" ) )
```
MishaDemianenko: please use already defined constant for default config file name

community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
```diff
@@ -287,9 +286,10 @@ public void execute( String[] args ) throws IncorrectUsage, CommandFailed
     private static Config loadNeo4jConfig( Path homeDir, Path configDir, String databaseName,
             Map<String,String> additionalConfig )
     {
-        Config config = ConfigLoader.loadConfigWithConnectorsDisabled( Optional.of( homeDir.toFile() ),
-                Optional.of( configDir.resolve( "neo4j.conf" ).toFile() ) );
-        additionalConfig.put( DatabaseManagementSystemSettings.active_database.name(), databaseName );
-        return config.with( additionalConfig );
+        return Config.fromFile( configDir.resolve( "neo4j.conf" ) )
```
MishaDemianenko: config name here as well

enterprise/kernel/src/main/java/org/neo4j/graphdb/factory/EnterpriseGraphDatabaseFactory.java
```diff
@@ -45,14 +45,15 @@
...
             public GraphDatabaseService newDatabase( Map<String,String> config )
             {
-                return newDatabase( Config.embeddedDefaults( config ) );
+                return newDatabase( Config.defaults( config ) );
             }
 
             @Override
             public GraphDatabaseService newDatabase( Config config )
             {
+                config.augment( "unsupported.dbms.ephemeral", FALSE );
```
MishaDemianenko: we can also update code to reference setting instead of string here?

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyLoader.java
```diff
@@ -165,8 +161,7 @@ private SslPolicy loadOrCreateLegacyPolicy()
 
     private void load( Config config, Log log )
     {
-        Function<ConfigValues,Stream<String>> enumeration = GroupSettingSupport.enumerate( SslPolicyConfig.class );
-        Set<String> policyNames = config.view( enumeration ).collect( toSet() );
+        Set<String> policyNames = config.identifiersFromPrefix( SslPolicyConfig.class.getAnnotation( Group.class ).value() );
```
MishaDemianenko: it looks like it would be more clear if we will pass only class that has Group annotation and will do the extraction of annotation value internally and not sure 'identifiersFromPrefix' is really describing what we doing here with group of settings

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -604,10 +793,25 @@ private void warnAboutDeprecations( Map<String,String> userSettings )
...
     {
         return httpConnectors( params )
                 .filter( c -> c.enabled.apply( params::get ) )
                 .collect( Collectors.toList() );
     }
+
+    @Override
+    public String toString()
+    {
+        List<String> keys = new ArrayList<>( params.keySet() );
```
MishaDemianenko: this can be done simpler using a sorted stream of param entries to produce output string

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -74,237 +73,467 @@
...
+    public static final String DEFAULT_CONFIG_FILE_NAME = "neo4j.conf";
+
     private final List<ConfigOptions> configOptions;
 
-    private final Map<String,String> params = new ConcurrentHashMap<>();
+    private final Map<String,String> params = new CopyOnWriteHashMap<>();
     private final ConfigurationMigrator migrator;
-    private final Optional<File> configFile;
     private final List<ConfigurationValidator> validators = new ArrayList<>();
+    private final Map<String,String> overriddenDefaults = new CopyOnWriteHashMap<>(); // TODO : has bo be reapplied every update
```
MishaDemianenko: todo

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -74,237 +73,467 @@
  */
 public class Config implements DiagnosticsProvider, Configuration
 {
+    public static final String DEFAULT_CONFIG_FILE_NAME = "neo4j.conf";
+
     private final List<ConfigOptions> configOptions;
 
-    private final Map<String,String> params = new ConcurrentHashMap<>();
+    private final Map<String,String> params = new CopyOnWriteHashMap<>();
```
MishaDemianenko: why do we use COW map instead of for example CHM?

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -74,237 +73,467 @@
...
-     * specified value.
+     * Extract all identifiers within a group, e.g. giving that prefix is {@code "dbms.ssl.policy"}), the
+     * setting {@code "dbms.ssl.policy.default.base_directory"} will return {@code "default"}.
      *
-     * @return a new modified config, leaves this config unchanged.
+     * @param prefix group prefix for the settings interested in
+     * @return A set of uniquely and sorted group keys.
      */
-    public Config withDefaults( Map<String,String> additionalDefaults )
+    public Set<String> identifiersFromPrefix( String prefix )
```
MishaDemianenko: if this is prefix base search why we can't simply use start with?

community/kernel/src/main/java/org/neo4j/kernel/configuration/IndividualSettingsValidator.java
```diff
@@ -102,6 +94,6 @@ public IndividualSettingsValidator( boolean warnOnUnknownSettings )
             }
         } );
 
-        return validConfig;
+        return Collections.emptyMap();
```
MishaDemianenko, line 97: empty map as result again

MishaDemianenko (changes requested):
Couple of small comments and questions otherwise looks good!

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
klaren: To avoid locks for read heavy workloads

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
klaren: Todo for part 2 of this PR, already gone on that branch

community/neo4j/pom.xml
klaren: Intellij automatically added that, removed!

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
klaren: All hard coded strings have been replaced now.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusterConfigurationValidator.java
klaren: Config validators are allowed to produce new values for the config. This behavior is questionable and should probably be removed in the future. We'll leave it for now since public API restricts refactoring.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyLoader.java
klaren: Agree, updated

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
klaren: We use the match to extract the identifier. Javadoc is updated to make this more clear.

klaren:
@MishaDemianenko @tinwelint thanks for comments! New commits available addressing them.

MishaDemianenko (approved):

PR #9647, opened 2017-07-10
Added some test for aggregation and optional match

SaschaPeukert:
The original problem seems to be fixed but there was no actual test.
Thus I added one for mixing optional and count(distinct n.prop)

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/PatternPredicates.feature
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/PatternPredicates.feature
@@ -304,6 +304,23 @@ Feature: PatternPredicates
       | (:A)  |
     And no side effects
 
+  Scenario: Using a optional match after aggregation and before an aggregation
+    And having executed:
+      """
+      CREATE (:Z{key:1})-[:IS_A]->(:A)
+      """
+    When executing query:
+      """
+      MATCH (a:A)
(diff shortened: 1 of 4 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/PatternPredicates.feature
```diff
@@ -304,6 +304,23 @@ Feature: PatternPredicates
       | (:A)  |
     And no side effects
 
+  Scenario: Using a optional match after aggregation and before an aggregation
```
alexaverbuch: I think this scenario should be moved to : https://github.com/SaschaPeukert/neo4j/blob/6af43b6ff2a740b37f71b9c75a1318809f30a825/enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/OptionalMatchAcceptance.feature
(1) there is no pattern predicate in the scenario, e.g., `WHERE <pattern>`
(2) because that is an Optional Match-specific feature file

alexaverbuch (changes requested):

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/PatternPredicates.feature
SaschaPeukert: I created a new AggregationAcceptance.feature file and added the test to that.

PR #9649, opened 2017-07-10
More helpful error message when procedure loading fails due to sandboxing

chrisvest:


OliviaYtterbrink (approved):
This looks like a good improvement

PR #9650, opened 2017-07-10
Fix broken label scan store for read only slaves in HA

burqen:
Previously you could not start a label scan store writer in on a
read only slave and thus you could not apply updates.

Now the read only mode only applies to the label scan store if
operation mode is single instance, just like how it works in lucene.


MishaDemianenko (approved):
I have updated labels a bit, otherwise good

burqen:
@MishaDemianenko Thanks for the additional labels 🎉

PR #9651, opened 2017-07-10
NoOp lock manager for read replicas

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/NoOpLockManager.java
```diff
@@ -0,0 +1,123 @@
...
+    {
+        @Override
+        public void acquireShared( LockTracer tracer, ResourceType resourceType, long... resourceIds ) throws AcquireLockTimeoutException
+        {
+        }
+
+        @Override
+        public void acquireExclusive( LockTracer tracer, ResourceType resourceType, long... resourceIds ) throws AcquireLockTimeoutException
+        {
+            throw new IllegalStateException();
```
chrisvest: This might be overly restrictive. A procedure could, for instance, be using exclusive locks to keep concurrent writers away from a data structure while it is reading it.

chrisvest: Although a shared lock would have the same effect in that example.

martinfurmanski: Yea it is interesting what to do here really, because replication happens behind the scenes regardless. None of these locks will have a distributed effect. Neither an exclusive nor a shared lock will be able to prevent upstream updates coming in. I am not really sure which semantics to aim for here. Would be better if all instances participated in concurrency control properly.

This is trying to catch the case early of someone doing a writing query on a read replica.

PR #9652, opened 2017-07-11
Add additional procedures to access transaction metadata

MishaDemianenko:
changelog: Add additional standard procedure dbms.getTXMetaData to access transactional metadata.

### DIFF
--- a/enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
+++ b/enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
@@ -53,7 +53,6 @@ import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Name;
 import org.neo4j.procedure.Procedure;
 
-import static java.lang.String.format;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static org.neo4j.function.ThrowingFunction.catchThrown;
@@ -66,8 +65,6 @@ import static org.neo4j.procedure.Mode.DBMS;
 @SuppressWarnings( "unused" )
 public class EnterpriseBuiltInDbmsProcedures
 {
-    private static final int HARD_CHAR_LIMIT = 2048;
-
     @Context
     public DependencyResolver resolver;
 
@@ -83,20 +80,20 @@ public class EnterpriseBuiltInDbmsProcedures
     public void setTXMetaData( @Name( value = "data" ) Map<String,Object> data )
     {
         securityContext.assertCredentialsNotExpired();
-        int totalCharSize = data.entrySet().stream()
-                .mapToInt( e -> e.getKey().length() + e.getValue().toString().length() )
-                .sum();
-
-        if ( totalCharSize >= HARD_CHAR_LIMIT )
-        {
-            throw new IllegalArgumentException(
-                    format( "Invalid transaction meta-data, expected the total number of chars for " +
-                            "keys and values to be less than %d, got %d", HARD_CHAR_LIMIT, totalCharSize ) );
-        }
-
         try ( Statement statement = getCurrentTx().acquireStatement() )
         {
             statement.queryRegistration().setMetaData( data );
+        }
+    }
+
+    @Description( "Provides attached transaction metadata." )
+    @Procedure( name = "dbms.getTXMetaData", mode = DBMS )
+    public Stream<MetadataResult> getTXMetaData()
+    {
+        securityContext.assertCredentialsNotExpired();
+        try ( Statement statement = getCurrentTx().acquireStatement() )
+        {
+            return Stream.of( statement.queryRegistration().getMetaData() ).map( MetadataResult::new );
         }
     }
 
@@ -525,6 +522,16 @@ public class EnterpriseBuiltInDbmsProcedures
         {
             this.username = username;
             this.connectionCount = connectionCount;
+        }
+    }
+
+    public static class MetadataResult
+    {
+        public final Map<String,Object> metadata;
+
+        MetadataResult( Map<String,Object> metadata )
+        {
+            this.metadata = metadata;
         }
     }
 }
--- a/enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
+++ b/enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
@@ -53,6 +53,7 @@ import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Name;
 import org.neo4j.procedure.Procedure;
 
+import static java.lang.String.format;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static org.neo4j.function.ThrowingFunction.catchThrown;
@@ -65,6 +66,8 @@ import static org.neo4j.procedure.Mode.DBMS;
 @SuppressWarnings( "unused" )
 public class EnterpriseBuiltInDbmsProcedures
 {
+    private static final int HARD_CHAR_LIMIT = 2048;
+
     @Context
     public DependencyResolver resolver;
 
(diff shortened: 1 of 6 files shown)


MishaDemianenko:
@OliviaYtterbrink can you please take a look! thx!

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
```diff
@@ -83,20 +80,20 @@
     public void setTXMetaData( @Name( value = "data" ) Map<String,Object> data )
     {
         securityContext.assertCredentialsNotExpired();
-        int totalCharSize = data.entrySet().stream()
-                .mapToInt( e -> e.getKey().length() + e.getValue().toString().length() )
-                .sum();
-
-        if ( totalCharSize >= HARD_CHAR_LIMIT )
```
craigtaverner: Why do we remove this limit? I am guessing it was there to prevent a kind of DoS, whereby injecting massive maps would slow the server down, since the maps are used in query logging and query listing.

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
```diff
@@ -527,4 +524,14 @@ public QueryTerminationResult( QueryId queryId, String username )
             this.connectionCount = connectionCount;
         }
     }
+
+    public static class MetadataResult
+    {
+        public final Map<String,Object> metadata;
```
craigtaverner, line 544: I'm not sure of the original use case for this feature, but I would have thought the result table might look nice as a two column table, with key and value pairs, instead of a single column, single row with a single map value. Just a suggestion.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/BuiltInProceduresInteractionTestBase.java
```diff
@@ -638,17 +640,38 @@ public void shouldHaveSetTXMetaDataProcedure() throws Throwable
     }
 
     @Test
-    public void shouldFailOnTooLargeTXMetaData() throws Throwable
+    public void readUpdatedMetadataValue() throws Exception
     {
-        ArrayList<String> list = new ArrayList<>();
-        for ( int i = 0; i < 200; i++ )
+        GraphDatabaseFacade localGraph = neo.getLocalGraph();
+        try ( Transaction transaction = localGraph.beginTx() )
```
craigtaverner: Why do these test bypass security, which is not following the convention of all other tests in this class?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/BuiltInProceduresInteractionTestBase.java
```diff
@@ -638,17 +640,38 @@ public void shouldHaveSetTXMetaDataProcedure() throws Throwable
...
         {
-            list.add( format( "key%d: '0123456789'", i ) );
+            String testValue = "testValue";
+            String testKey = "test";
+            localGraph.execute( "CALL dbms.setTXMetaData", map( "data", map( testKey, testValue ) ) );
```
craigtaverner: This syntax is surprising to me. Is it possible to call procedures without specifying their arguments, and then using a magic 'data' parameter to get the arguments into the procedure?

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/BuiltInProceduresInteractionTestBase.java
```diff
@@ -638,17 +640,38 @@ public void shouldHaveSetTXMetaDataProcedure() throws Throwable
...
+            Map<String,Object> metadata =
+                    (Map<String,Object>) localGraph.execute( "CALL dbms.getTXMetaData " ).next().get( "metadata" );
```
craigtaverner: The `get("metadata")` seems unnecessary. The procedure already says `getTXMetaData` in the name. I think it would be more intuitive to have the return stream be a stream of tuples, with key:value columns.

craigtaverner (changes requested):

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
MishaDemianenko: I removed it since most of the usages of metadata that I saw was by actually working with metadata map in TKI directly bypassing procedure to set values and that means that this check does not really solve it purpose completely, and instead add weirdness of failing depending on your toString() result size. But i'm fully trust you here so will restore this check

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/BuiltInProceduresInteractionTestBase.java
MishaDemianenko: i do not quite see how that can be helpful since that will require Q(n) time to find your key in a worst case

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/BuiltInProceduresInteractionTestBase.java
MishaDemianenko: i actually need those 2 procedures to be invoked as part of 1 transaction and i didn't found another easy way how to do that :)

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
MishaDemianenko: The original idea is that people will call this procedure and will get a map as a result so they can easily find a value that they set in metadata by doing a lookup, having results of this procedure as a list of key-value pairs will make it more complex and cumbersome to do from a user perspective.

MishaDemianenko:
@craigtaverner thx for comments! I have tried to answer/fix some of them, so can you please take another round! thx

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
craigtaverner: The Neo4j results objects are kind-of a streamed version of a map, so work better for memory and for display in tools that already work with results, like the browser, neo4j-shell and cypher-shell. Only slightly less convenient for a client developer just wanting to pull the entire map and do a single lookup, which is what you are optimizing for.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/BuiltInProceduresInteractionTestBase.java
craigtaverner: See the example for setTxMetaData in https://github.com/neo4j/neo4j/blob/3.3/enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerIT.java#L157-L164

That looks much cleaner and does not bypass security.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/BuiltInProceduresInteractionTestBase.java
```diff
@@ -638,17 +640,38 @@ public void shouldHaveSetTXMetaDataProcedure() throws Throwable
...
+            assertEquals( testValue, metadata.get( testKey ) );
+        }
+    }
+
+    @Test
+    public void readEmptyMetadataInOtherTransaction()
+    {
+        String testValue = "testValue";
+        String testKey = "test";
+        GraphDatabaseFacade localGraph = neo.getLocalGraph();
```
craigtaverner: You should not need to get the localGraph or create separate transactions here. Perhaps:

    assertEmpty( writeSubject, "CALL dbms.setTXMetaData( { "+testKey+":"+testValue+" } )" );
    assertSuccess( writeSubject, "CALL dbms.getTXMetaData()", r -> {...} );

craigtaverner (changes requested):

MishaDemianenko:
@craigtaverner thx a lot, updated now we should use correct transactions in tests and use appropriate methods

craigtaverner (approved):

PR #9653, opened 2017-07-11
Improve exceptions message for lifecycle graceful stop/shutdown attempts

MishaDemianenko:
Improve exception message to be less confusing and highlight miss leading
suppressed exception nature and origin.


MishaDemianenko:
@digitalstain can you please take a look if this message decreasing confusion level, thx!

digitalstain (approved):

PR #9655, opened 2017-07-12
3.2 dns discovery

phughk:


jimwebber:
The DnsResolutionResolver implementation is coupled directly to InetAddress. I think that DnsResolutionResolver will fail when you run it disconnected or if your DNS is slow.

I bring this up because I wrote the same anti-pattern in neo4j-server some years back and it failed weirdly when I was on a crappy 3G connection.

Suggest wrapping calls to InetAddress and friends in an abstraction we own and injecting a test implementation to drive our the correctness of your application logic.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/SharedDiscoveryServiceIT.java
```diff
@@ -89,23 +90,26 @@ public void shouldDiscoverCompleteTargetSetWithoutDeadlocks() throws Exception
...
+            Set<MemberId> expectedTargetSet ) throws ExecutionException, InterruptedException
     {
         Neo4jJobScheduler jobScheduler = new Neo4jJobScheduler();
         jobScheduler.init();
+        ResolutionResolver resolutionResolver = new NoOpResolutionResolver();
 
-        CoreTopologyService topologyService = disoveryServiceFactory.coreTopologyService( config(), null, member,
-                jobScheduler, logProvider, userLogProvider );
+        CoreTopologyService topologyService = disoveryServiceFactory
+                .coreTopologyService( config(), null, member, jobScheduler, logProvider, userLogProvider,
```
martinfurmanski: You have some reformatting all over the place which does not look standard to me. Please refrain from reformatting the entire files or find a reasonably aligned code style.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/DnsResolutionResolverTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    public void shouldResolveIPv6LocalhostDNSToIPs()
+    {
+        Collection<AdvertisedSocketAddress> resolve = resolver.resolve( new AdvertisedSocketAddress( "[::1]", 80 ) );
+        assertThat( resolve.size(), greaterThanOrEqualTo( 1 ) );
+    }
+
+    @Test
+    public void shouldExceptionAndGiveDefaultValue()
```
martinfurmanski: I don't really get this test. Don't you want the error to bubble up here really?

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/SharedDiscoveryServiceIT.java
```diff
@@ -53,12 +53,13 @@
 public class SharedDiscoveryServiceIT
 {
     private static final long TIMEOUT_MS = 15_000;
-    private static final long RUN_TIME_MS = 1000;;
+    private static final long RUN_TIME_MS = 1000;
+    ;
```
martinfurmanski: I think you can safely remove this extra semicolon

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
```diff
@@ -121,14 +127,37 @@
...
+    {
+        DNS,
+        LIST
+    }
+
+    @Description( "Configure the discovery type" )
+    public static final Setting<DiscoveryType> discovery_type =
+            setting( "causal_clustering.discovery_type", options( DiscoveryType.class ), DiscoveryType.LIST.name() );
+
+    public static ResolutionResolver chooseResolver( Config config, LogProvider logProvider,
```
martinfurmanski: Would it perhaps make sense to move this somewhere else outside the Settings-class? I think it might get a bit polluted here otherwise.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/SharedDiscoveryServiceIT.java
phughk: @martinfurmanski I have reversed a lot of the format changes, what remains are largely improvements or actual changes (with the odd import re-order). I will avoid formatting files like this in the future, thanks

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/DnsResolutionResolverTest.java
phughk: @martinfurmanski Sorry for the late reply - I can change it to being wrapped in a `RuntimeException(e)` and thrown back up if you like? That is probably the best solution anyway come to think of it.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
phughk: I have moved this to `ResolutionResolverFactory`

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/SharedDiscoveryServiceIT.java
martinfurmanski: @phughk We should have better code styles and guidelines internally :)

martinfurmanski (approved):

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/DnsResolutionResolverTest.java
phughk: Test has been removed, exception has been wrapped in a unchecked `UnknownHostException`

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
```diff
@@ -121,6 +122,16 @@
             setting( "causal_clustering.initial_discovery_members", list( ",", ADVERTISED_SOCKET_ADDRESS ),
                     NO_DEFAULT );
 
+    public enum DiscoveryType
+    {
+        DNS,
+        LIST
+    }
+
+    @Description( "Configure the discovery type" )
```
digitalstain: Perhaps this description can be a bit more verbose? Not greatly, just a bit - the discovery type of what, for example. Or a very small second sentence explaining possible values or what the default is and what it does.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/ResolutionResolver.java
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.causalclustering.discovery;
+
+import java.util.Collection;
+
+import org.neo4j.helpers.AdvertisedSocketAddress;
+
+public interface ResolutionResolver
```
digitalstain: The name of this interface is slightly too generic and the single method signature is not helpful in understanding the service it provides. Either a name change or a comment would be helpful.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
phughk: How about

> Discovery type used for cluster name resolution

PR #9656, opened 2017-07-12
Remove upgrade time and upgrade id from StoreId

MishaDemianenko:
Remove upgrade information from StoreId to allow multiple different
upgrades in a cluster.
Increment protocol version


MishaDemianenko:
ping @digitalstain

MishaDemianenko:
This idea will not gonna work, closing.

PR #9657, opened 2017-07-12
Remove Lucene label scan index

burqen:
And remove label scan store as kernel extension.

[cl] Completely remove Lucene label index which is replaced with native implementation since 3.2.


community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
```diff
@@ -241,17 +238,18 @@ public Result runFullConsistencyCheck( final File storeDir, Config config, Progr
         {
             IndexStoreView indexStoreView = new NeoStoreIndexStoreView( LockService.NO_LOCK_SERVICE, neoStores );
```
klaren: This is created by never used

community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
```diff
@@ -241,17 +238,18 @@ public Result runFullConsistencyCheck( final File storeDir, Config config, Progr
...
             Dependencies dependencies = new Dependencies();
```
klaren: This is created but dependencies are never satisfied

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -544,28 +543,11 @@
...
-         */
-        LUCENE,
-
-        /**
-         * Selects which ever label index is present in a store, or the default (NATIVE) if no label index present.
-         */
-        AUTO
-    }
-
+    // todo Can we simply remove this setting or do we need to depricate it?
```
klaren: You have to revert everything in this file and add `@Deprecated` to it since its public, yes, even the enum.
What you could do is change `options( enumNames( LabelIndex.class ), true )` to `optionsIgnoreCase( "native" )` which would make any other setting than `native` fail with a nice error.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/scan/FullLabelStream.java
```diff
@@ -37,22 +36,21 @@
  */
 public class FullLabelStream implements FullStoreChangeStream, Visitor<NodeLabelUpdate,IOException>
 {
-    private final Supplier<IndexStoreView> lazyIndexStoreView;
+    private final IndexStoreView indexStoreView;
```
klaren, line 39: Nice removal of lazy :+1:

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
```diff
@@ -207,10 +214,13 @@ public RecordStorageEngine(
             schemaCache = new SchemaCache( constraintSemantics, Collections.emptyList() );
             schemaStorage = new SchemaStorage( neoStores.getSchemaStore() );
 
-            labelScanStore = labelScanStoreProvider.getLabelScanStore();
+            NeoStoreIndexStoreView neoStoreIndexStoreView = new NeoStoreIndexStoreView( lockService, neoStores );
+            Boolean readOnly = config.get( GraphDatabaseSettings.read_only ) && operationalMode == OperationalMode.single;
```
klaren, line 218: Just curious, so read_only only applies to single instance?

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/SchemaIndexMigrator.java
```diff
@@ -40,7 +40,7 @@
 {
     private final FileSystemAbstraction fileSystem;
     private boolean deleteObsoleteIndexes;
-    private File labelIndexDirectory;
+    private File luceneLabelIndexDirectory;
```
klaren: Unused

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/DatabaseMigrator.java
```diff
@@ -87,8 +83,8 @@ public void migrate( File storeDir )
         StoreUpgrader storeUpgrader = new StoreUpgrader( upgradableDatabase, progressMonitor, config, fs, pageCache,
                 logProvider );
 
-        StoreMigrationParticipant schemaMigrator = schemaIndexProvider.storeMigrationParticipant( fs, pageCache,
-                labelScanStoreProvider );
+        StoreMigrationParticipant schemaMigrator = schemaIndexProvider.storeMigrationParticipant( fs, pageCache
```
klaren: Format

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -288,7 +262,7 @@ public CountsTracker getCountsStore()
     @Override
     public void close() throws IOException
     {
-        // Here as a safety mechanism when e.g. panicing.
+        // Here as a safety mechanism when e.g. panicking.
```
klaren, line 265: Nice catch!

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexCRUDIT.java
```diff
@@ -169,8 +168,8 @@ private Node createNode( Map<String, Object> properties, Label ... labels )
     @Before
     public void before() throws Exception
     {
-        when( mockedIndexProvider.storeMigrationParticipant( any( FileSystemAbstraction.class ), any( PageCache.class ),
-                any( LabelScanStoreProvider.class ) ) )
+        when( mockedIndexProvider.storeMigrationParticipant( any( FileSystemAbstraction.class ), any( PageCache.class )
+        ) )
```
klaren: Format

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexingServiceTest.java
```diff
@@ -1232,8 +1231,8 @@ private IndexingService newIndexingServiceWithMockedDependencies( IndexPopulator
                 any( IndexSamplingConfig.class ) ) )
                 .thenReturn( accessor );
         when( indexProvider.snapshotMetaFiles() ).thenReturn( Iterators.emptyResourceIterator() );
-        when( indexProvider.storeMigrationParticipant( any( FileSystemAbstraction.class ), any( PageCache.class ),
-                any( LabelScanStoreProvider.class ) ) )
+        when( indexProvider.storeMigrationParticipant( any( FileSystemAbstraction.class ), any( PageCache.class )
+        ) )
```
klaren: Format

community/kernel/src/test/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStoreTest.java
```diff
@@ -97,4 +97,10 @@ public void shouldStartPopulationAgainIfNotCompletedFirstTime() throws Exception
         assertTrue( monitor.rebuiltCalled );
         life.shutdown();
     }
+
+    @Test
+    public void shouldWorkIf() throws Exception
```
klaren: Oh, a cliffhanger. Can't wait to find out what happens t

klaren (commented):
A few comments on things that I spotted.
Always nice to put the code base on a diet! 
Good work :+1:

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
burqen: Yes, because in a cluster, all of the machines need to be `read_only` for it to be true and this is not verified in HA or causal clustering.
Problems will arise if a slave is configured to be `read_only` but master is not. The slave should thus still receive updates from master which violates the its `read_only` state.

Conclusion is that this setting only makes sense for single instance.

burqen:
@klaren Nice feedback 🎉  Comments are addressed and PR rebased.

community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
```diff
@@ -239,18 +228,14 @@ public Result runFullConsistencyCheck( final File storeDir, Config config, Progr
...
-                    indexStoreView, pageCache, new Monitors(), RecoveryCleanupWorkCollector.IMMEDIATE );
-            KernelContext kernelContext = new SimpleKernelContext( storeDir, UNKNOWN, dependencies );
-            KernelExtensions extensions = life.add( new KernelExtensions(
-                    kernelContext, (Iterable) load( KernelExtensionFactory.class ), dependencies, ignore() ) );
             life.start();
-            LabelScanStore labelScanStore = life.add( extensions.resolveDependency( LabelScanStoreProvider.class,
-                    new NamedLabelScanStoreSelectionStrategy( config ) ).getLabelScanStore() );
-            SchemaIndexProvider indexes = life.add( extensions.resolveDependency( SchemaIndexProvider.class,
-                    HighestSelectionStrategy.getInstance() ) );
+            SchemaIndexProvider indexes = new LuceneSchemaIndexProvider( fileSystem, DirectoryFactory.PERSISTENT,
```
MishaDemianenko: SchemaIndexProvider is technically a lifecycle and before it was part of life, but it's a dummy lifecycle indeed. Just something I have noticed in case if you had some hesitation before

enterprise/ha/src/test/java/org/neo4j/kernel/api/impl/labelscan/NativeLabelScanStoreHaIT.java
```diff
@@ -19,21 +19,177 @@
...
+    private enum Labels implements Label
     {
-        return new NativeLabelScanStoreExtension();
+        First,
+        Second
     }
 
-    @Override
-    protected String labelIndexSettingName()
+    @Rule
```
MishaDemianenko: since this test is updated now lets do some house cleaning work: fields on top, tests after setup, etc..

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -562,12 +567,14 @@
...
+     */
+    @Deprecated
     @Description( "Backend to use for label --> nodes index" )
     @Internal
-    public static final Setting<String> label_index = setting( "dbms.label_index",
-            options( enumNames( LabelIndex.class ), true ), LabelIndex.NATIVE.name() );
-
-    // Security settings
+    public static final Setting<String> label_index =
+            setting( "dbms.label_index", optionsIgnoreCase( NATIVE_LABEL_INDEX_TAG ), NATIVE_LABEL_INDEX_TAG );
```
MishaDemianenko: hm not quite sure why do we need to reference NATIVE_LABEL_INDEX_TAG here? we can use LabelIndex.NATIVE.name() here instead of exposing constant from impl package ` org.neo4j.kernel.impl.index.labelscan` right?

MishaDemianenko: also should we maybe also allow AUTO, still work and use native always now?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LuceneLabelScanDirectory.java
```diff
@@ -17,28 +17,14 @@
...
-        this.bitmap = bitmap;
-    }
-
-    public long bitmap()
-    {
-        return bitmap;
-    }
-
-    public boolean hasContent()
+public class LuceneLabelScanDirectory
```
MishaDemianenko: what do you think about removing this class completely and just use `getLuceneStoreDirectory` as private static method in migrator (the only place where it's used now) directly instead?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -99,7 +96,7 @@
     /**
      * Native label index tag, to distinguish native label index from other label indexes
      */
-    public static final String NATIVE_LABEL_INDEX_TAG = GraphDatabaseSettings.LabelIndex.NATIVE.name();
+    public static final String NATIVE_LABEL_INDEX_TAG = "native";
```
MishaDemianenko: why we need another constant here?

MishaDemianenko: also since we will have only one label index i'm not sure we still need it, since we do not need to tag monitors anymore

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/SchemaIndexMigrator.java
```diff
@@ -72,8 +70,8 @@ public void moveMigratedFiles( File migrationDir, File storeDir, String versionT
         if ( deleteObsoleteIndexes )
         {
             deleteIndexes( schemaIndexDirectory );
-            deleteIndexes( labelIndexDirectory );
         }
+        deleteIndexes( LuceneLabelScanDirectory.getLuceneStoreDirectory( storeDir ) );
```
MishaDemianenko: This will gonna delete old label scan store if that was used which is a nice thing. In the same time, we will not gonna create the new index as part of migration for those guys and people will be forced to wait during the startup, while they expecting everything is ready to fly. I know that proposed behaviour differs from how it was before at the same time that is something we had complaints in the recent past and as well it sounds like a correct thing to do. WDYT?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/DynamicIndexStoreView.java
```diff
@@ -69,13 +83,43 @@ public DynamicIndexStoreView( LabelScanStore labelScanStore, LockService locks,
     {
         if ( forceStoreScan || !USE_LABEL_INDEX_FOR_SCHEMA_INDEX_POPULATION || useAllNodeStoreScan( labelIds ) )
```
MishaDemianenko, line 84: since now we will always use native label index i think we should o use it always to determine affected nodes or at least change threshold since we do not have the same problem here as we had with lucene index that we need to reopen before quite often to observe latest updates

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/LabelScanViewIdIterator.java
```diff
@@ -34,20 +32,13 @@
  */
 class LabelScanViewIdIterator implements PrimitiveLongResourceIterator
```
MishaDemianenko, line 29: this class not can go completely since we do not need it anymore and it performs only delegation operations now, also javadoc is wrong now

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
```diff
@@ -1133,6 +1134,12 @@ NeoStores getNeoStores()
         return neoStores;
     }
 
+    // test-access
+    LabelScanStore getLabelScanStore()
```
MishaDemianenko: hm...but no tests are actually using it

community/kernel/src/test/java/org/neo4j/graphdb/NativeLabelScanStoreUpdateIT.java
```diff
@@ -209,7 +214,10 @@ public void shouldHandleLargeAmountsOfNodesAddedAndRemovedInSameTx() throws Exce
         }
     }
 
-    protected abstract GraphDatabaseService db();
+    protected GraphDatabaseService db()
```
MishaDemianenko: we do not need this method to be protected anymore, mostly we do not need it at all

community/kernel/src/test/java/org/neo4j/unsafe/batchinsert/internal/BatchInsertTest.java
```diff
@@ -1034,32 +1020,45 @@ public void shouldPopulateLabelScanStoreOnShutdown() throws Exception
...
+        assertLabelScanStoreContains( labelScanStore, 0, node1, node4, node5 );
+        assertLabelScanStoreContains( labelScanStore, 1, node2, node4 );
+        assertLabelScanStoreContains( labelScanStore, 2, node3, node5 );
+
+        labelScanStore.shutdown();
     }
 
-    @Test
-    public void shouldSkipStoreScanIfNoLabelsAdded() throws Exception
+    private LabelScanStore getLabelScanStore()
```
MishaDemianenko: lets move this private methods to the bottom

MishaDemianenko (changes requested):
Great things! 🕺
while still have some comments! 🐼

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
burqen: Legacy, forgot to change that back after I reverted removal of `LabelIndex` enum. Fixing!

Hmm, my initial thinking was to remove it because it "has no meaning" anymore. However you could argue that whoever is using this AUTO setting "just want stuff to work", which is fair. So I agree, allow AUTO seems to be the least intrusive thing to do.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
burqen: ~~To not depend on deprecated enum. Do you think it is better to wait with this change until the enum is actually removed?~~ Removing tag.

burqen: Good point! I'll remove it.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/SchemaIndexMigrator.java
burqen: I think deleting Lucene label scan store during migration is absolutely the correct thing to do, since we would need to delete it anyway on next startup.

The question is if we need to rebuild native label scan store as part of migration or not and I think that would be the best thing to do. Although I think it is outside the scope of this PR.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/LabelScanViewIdIterator.java
burqen: We need it to close the reader when we are done with iteration. Have any simpler ideas on how to do this?
Nice catch on javadoc. Fixing!

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
burqen: Legacy. Nice catch!

burqen:
@MishaDemianenko Lot's of nice feedback! Thanks 😄  Comments addressed.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/LabelScanViewIdIterator.java
MishaDemianenko: oh yes, correct, let's keep it them!

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/DynamicIndexStoreView.java
```diff
@@ -124,8 +124,7 @@ private boolean useAllNodeStoreScan( int[] labelIds )
     {
         try
         {
-            return ArrayUtils.isEmpty( labelIds ) || isEmptyLabelScanStore() ||
-                    isNumberOfLabeledNodesExceedThreshold( labelIds );
+            return ArrayUtils.isEmpty( labelIds ) || isEmptyLabelScanStore();
```
MishaDemianenko, line 127: we need to update java doc accordingly as well

burqen: Nice one! Addressed 👍

burqen:
@MishaDemianenko Previous test failure was a flaky test. Can be reproduced locally.

MishaDemianenko (approved):

jotomo:
I'm a bit out of the loop lately, but so far the kernel module had an implicit dependency (to avoid a cyclic dependency I guess) on the lucene module. Does this change imply that the kernel module can now be used _without_ the lucene module if only schema indexes are used? At this time it's probably only theoretically possible, since kernel pulls in the lucene-upgrade module ...

burqen:
@jotomo Nope. All schema indexes are still using Lucene. It is only the label index that has been moved off Lucene.

jotomo:
@burqen Alright, thanks for the info!

PR #9658, opened 2017-07-12
Log original failure when channel closed in Bolt

lutovich:
Drivers can close socket connections while having ongoing transactions. This results in transaction termination in the database. No errors can then be reported back to the client because connection has been terminated.

Previously Bolt server logged inability to write to a closed channel and lost the original failure message (`TransactionTerminatedException` in most cases).

This PR makes it always log the root cause. Also changed `AssertableLogProvider` to use COWAL instead of a synchronized list to allow concurrent iteration and modification.


zhenlineo:
LG

PR #9664, opened 2017-07-13
Improve bookmark waiting

martinfurmanski:
Bookmark waiting was previously based on polling and now instead
uses wait/notify primitives. This is implemented in the AQOOS
which now supports waiting for a closed transaction id.

The highestGapFreeNumber is now also volatile so that visibility
is ensured. This is not necessary for the new bookmark waiting, but
potentially for other uses.

The tests for the TransactionIdTracker were also improved to remove
cluttering boilerplate and make it more of a unit test rather than
a mocked up messy gray box of integration.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/txtracking/TransactionIdTracker.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/txtracking/TransactionIdTracker.java
@@ -20,7 +20,7 @@
 package org.neo4j.kernel.api.txtracking;
 
 import java.time.Duration;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Supplier;
 
 import org.neo4j.kernel.AvailabilityGuard;
@@ -28,7 +28,6 @@ import org.neo4j.kernel.api.exceptions.Status;
 import org.neo4j.kernel.api.exceptions.TransactionFailureException;
 import org.neo4j.kernel.impl.transaction.log.TransactionIdStore;
 
-import static org.neo4j.function.Predicates.tryAwaitEx;
 import static org.neo4j.kernel.impl.transaction.log.TransactionIdStore.BASE_TX_ID;
 
 /**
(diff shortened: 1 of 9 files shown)


community/kernel/src/main/java/org/neo4j/kernel/api/txtracking/TransactionIdTracker.java
```diff
@@ -78,23 +74,22 @@ public void awaitUpToDate( long oldestAcceptableTxId, Duration timeout ) throws
...
             throw new TransactionFailureException( Status.General.DatabaseUnavailable,
                     "Database had become unavailable while waiting for requested version %d.", oldestAcceptableTxId );
         }
-        return oldestAcceptableTxId <= transactionIdStore().getLastClosedTransactionId();
+
+        try
+        {
+            transactionIdStore().awaitClosedTransactionId( oldestAcceptableTxId, timeout.toMillis() );
+        }
+        catch ( InterruptedException | TimeoutException e )
```
lutovich, line 86: I think interruption status of the thread should be restored here, if it was actually interrupted

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ArrayQueueOutOfOrderSequence.java
```diff
@@ -96,6 +99,24 @@ public long highestEverSeen()
         return createResult( number, meta );
     }
 
+    @Override
+    public synchronized void await( long awaitedNumber, long timeoutMillis ) throws TimeoutException, InterruptedException
+    {
+        long endTime = System.currentTimeMillis() + timeoutMillis;
```
lutovich, line 105: I think value of `timeoutMillis` should be verified somewhere to avoid overflow of `endTime`

community/kernel/src/main/java/org/neo4j/kernel/api/txtracking/TransactionIdTracker.java
martinfurmanski: Right, updated!

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ArrayQueueOutOfOrderSequence.java
martinfurmanski: Although prudent, in practice someone would have to enter a timeout in excess of 290 million years for that to be a problem.

digitalstain (approved):

PR #9668, opened 2017-07-14
Remove old count store files during migration.

ragadeeshu:


MishaDemianenko (changes requested):
Would expect to see some tests

ragadeeshu:
@MishaDemianenko updated

MishaDemianenko (approved):

PR #9670, opened 2017-07-14
The importer can now process datasets that are so big that even the external-to-internal ID map cannot fit in memory

chrisvest:
- [x] Make the page cache backed NumberArray implementations safe to access concurrently by multiple threads.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/NeoStores.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/NeoStores.java
@@ -151,6 +151,11 @@ public class NeoStores implements AutoCloseable
         return storeDir;
     }
 
+    public PageCache getPageCache()
+    {
+        return pageCache;
+    }
+
     private File getStoreFile( String substoreName )
     {
         return new File( neoStoreFileName.getPath() + substoreName );
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/NeoStores.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/NeoStores.java
@@ -151,11 +151,6 @@ public class NeoStores implements AutoCloseable
         return storeDir;
     }
 
-    public PageCache getPageCache()
-    {
-        return pageCache;
-    }
-
     private File getStoreFile( String substoreName )
     {
         return new File( neoStoreFileName.getPath() + substoreName );
@@ -637,7 +632,7 @@ public class NeoStores implements AutoCloseable
             public void initialize( CountsAccessor.Updater updater )
             {
                 log.warn( "Missing counts store, rebuilding it." );
-                new CountsComputer( neoStores ).initialize( updater );
+                new CountsComputer( neoStores, pageCache ).initialize( updater );
                 log.warn( "Counts store rebuild completed." );
             }
 
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/RelationshipGroupDefragmenter.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/RelationshipGroupDefragmenter.java
@@ -44,6 +44,9 @@ import static org.neo4j.unsafe.impl.batchimport.staging.ExecutionSupervisors.sup
 public class RelationshipGroupDefragmenter
 {
 
+    private final Configuration config;
+    private final ExecutionMonitor executionMonitor;
+    private final Monitor monitor;
     private final NumberArrayFactory numberArrayFactory;
 
     public interface Monitor
@@ -58,15 +61,11 @@ public class RelationshipGroupDefragmenter
         default void defragmentingNodeRange( long fromNodeId, long toNodeId )
         {   // empty
         }
-
         Monitor EMPTY = new Monitor()
         {   // empty
         };
-    }
 
-    private final Configuration config;
-    private final ExecutionMonitor executionMonitor;
-    private final Monitor monitor;
+    }
 
     public RelationshipGroupDefragmenter( Configuration config, ExecutionMonitor executionMonitor,
                                           NumberArrayFactory numberArrayFactory )
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/RelationshipGroupDefragmenter.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/RelationshipGroupDefragmenter.java
@@ -23,12 +23,12 @@ import org.neo4j.kernel.impl.store.RecordStore;
 import org.neo4j.kernel.impl.store.record.Record;
 import org.neo4j.kernel.impl.store.record.RelationshipGroupRecord;
 import org.neo4j.unsafe.impl.batchimport.cache.ByteArray;
+import org.neo4j.unsafe.impl.batchimport.cache.NumberArrayFactory;
 import org.neo4j.unsafe.impl.batchimport.staging.ExecutionMonitor;
 import org.neo4j.unsafe.impl.batchimport.staging.Stage;
 import org.neo4j.unsafe.impl.batchimport.store.BatchingNeoStores;
 
 import static org.neo4j.unsafe.impl.batchimport.Configuration.withBatchSize;
-import static org.neo4j.unsafe.impl.batchimport.cache.NumberArrayFactory.AUTO;
 import static org.neo4j.unsafe.impl.batchimport.staging.ExecutionSupervisors.superviseExecution;
 
 /**
@@ -43,6 +43,9 @@ import static org.neo4j.unsafe.impl.batchimport.staging.ExecutionSupervisors.sup
  */
 public class RelationshipGroupDefragmenter
 {
+
+    private final NumberArrayFactory numberArrayFactory;
+
     public interface Monitor
     {
         /**
@@ -65,22 +68,25 @@ public class RelationshipGroupDefragmenter
     private final ExecutionMonitor executionMonitor;
     private final Monitor monitor;
 
-    public RelationshipGroupDefragmenter( Configuration config, ExecutionMonitor executionMonitor )
+    public RelationshipGroupDefragmenter( Configuration config, ExecutionMonitor executionMonitor,
+                                          NumberArrayFactory numberArrayFactory )
     {
-        this( config, executionMonitor, Monitor.EMPTY );
+        this( config, executionMonitor, Monitor.EMPTY, numberArrayFactory );
     }
 
-    public RelationshipGroupDefragmenter( Configuration config, ExecutionMonitor executionMonitor, Monitor monitor )
+    public RelationshipGroupDefragmenter( Configuration config, ExecutionMonitor executionMonitor, Monitor monitor,
+                                          NumberArrayFactory numberArrayFactory )
     {
         this.config = config;
         this.executionMonitor = executionMonitor;
         this.monitor = monitor;
+        this.numberArrayFactory = numberArrayFactory;
     }
 
     public void run( long memoryWeCanHoldForCertain, BatchingNeoStores neoStore, long highNodeId )
     {
         try ( RelationshipGroupCache groupCache =
-                new RelationshipGroupCache( AUTO, memoryWeCanHoldForCertain, highNodeId ) )
+                new RelationshipGroupCache( numberArrayFactory, memoryWeCanHoldForCertain, highNodeId ) )
         {
             // Read from the temporary relationship group store...
             RecordStore<RelationshipGroupRecord> fromStore = neoStore.getTemporaryRelationshipGroupStore();
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.unsafe.impl.batchimport.cache;
+
+import static java.lang.Long.min;
+
+/**
+ * Used as part of the fallback strategy for {@link Auto}. Tries to split up fixed-size arrays
+ * ({@link #newLongArray(long, long)} and {@link #newIntArray(long, int)} into smaller chunks where
+ * some can live on heap and some off heap.
+ */
+public class ChunkedNumberArrayFactory extends NumberArrayFactory.Adapter
+{
+    private final NumberArrayFactory delegate;
+
+    public ChunkedNumberArrayFactory()
+    {
+        this( OFF_HEAP, HEAP );
+    }
+
+    public ChunkedNumberArrayFactory( NumberArrayFactory... delegateList )
+    {
+        delegate = new Auto( delegateList );
+    }
+
+    @Override
+    public LongArray newLongArray( long length, long defaultValue, long base )
+    {
+        // Here we want to have the property of a dynamic array which makes some parts of the array
+        // live on heap, some off. At the same time we want a fixed size array. Therefore first create
+        // the array as a dynamic array and make it grow to the requested length.
+        LongArray array = newDynamicLongArray( fractionOf( length ), defaultValue );
+        array.at( length - 1 );
+        return array;
+    }
+
+    @Override
+    public IntArray newIntArray( long length, int defaultValue, long base )
+    {
+        // Here we want to have the property of a dynamic array which makes some parts of the array
+        // live on heap, some off. At the same time we want a fixed size array. Therefore first create
+        // the array as a dynamic array and make it grow to the requested length.
+        IntArray array = newDynamicIntArray( fractionOf( length ), defaultValue );
+        array.at( length - 1 );
+        return array;
+    }
+
+    @Override
+    public ByteArray newByteArray( long length, byte[] defaultValue, long base )
+    {
+        // Here we want to have the property of a dynamic array which makes some parts of the array
+        // live on heap, some off. At the same time we want a fixed size array. Therefore first create
+        // the array as a dynamic array and make it grow to the requested length.
+        ByteArray array = newDynamicByteArray( fractionOf( length ), defaultValue );
+        array.at( length - 1 );
+        return array;
+    }
+
+    private long fractionOf( long length )
+    {
+        int idealChunkCount = 10;
+        if ( length < idealChunkCount )
+        {
+            return length;
+        }
+        int maxArraySize = Integer.MAX_VALUE - Short.MAX_VALUE;
+        return min( length / idealChunkCount, maxArraySize );
+    }
+
+    @Override
+    public IntArray newDynamicIntArray( long chunkSize, int defaultValue )
+    {
+        return new DynamicIntArray( delegate, chunkSize, defaultValue );
+    }
+
+    @Override
+    public LongArray newDynamicLongArray( long chunkSize, long defaultValue )
+    {
+        return new DynamicLongArray( delegate, chunkSize, defaultValue );
+    }
+
+    @Override
+    public ByteArray newDynamicByteArray( long chunkSize, byte[] defaultValue )
+    {
+        return new DynamicByteArray( delegate, chunkSize, defaultValue );
+    }
+
+    @Override
+    public String toString()
+    {
+        return "CHUNKED_FIXED_SIZE";
+    }
+}
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
@@ -28,14 +28,18 @@ import static java.lang.Long.min;
  */
 public class ChunkedNumberArrayFactory extends NumberArrayFactory.Adapter
 {
+    static final int MAGIC_CHUNK_COUNT = 10;
+    // This is a safe bet on the maximum number of items the JVM can store in an array. It is commonly slightly less
+    // than Integer.MAX_VALUE
+    private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - Short.MAX_VALUE;
     private final NumberArrayFactory delegate;
 
-    public ChunkedNumberArrayFactory()
+    ChunkedNumberArrayFactory()
     {
         this( OFF_HEAP, HEAP );
     }
 
-    public ChunkedNumberArrayFactory( NumberArrayFactory... delegateList )
+    ChunkedNumberArrayFactory( NumberArrayFactory... delegateList )
     {
         delegate = new Auto( delegateList );
     }
@@ -43,45 +47,34 @@ public class ChunkedNumberArrayFactory extends NumberArrayFactory.Adapter
     @Override
     public LongArray newLongArray( long length, long defaultValue, long base )
     {
-        // Here we want to have the property of a dynamic array which makes some parts of the array
-        // live on heap, some off. At the same time we want a fixed size array. Therefore first create
-        // the array as a dynamic array and make it grow to the requested length.
-        LongArray array = newDynamicLongArray( fractionOf( length ), defaultValue );
-        array.at( length - 1 );
-        return array;
+        // Here we want to have the property of a dynamic array so that some parts of the array
+        // can live on heap, some off.
+        return newDynamicLongArray( fractionOf( length ), defaultValue );
     }
 
     @Override
     public IntArray newIntArray( long length, int defaultValue, long base )
     {
-        // Here we want to have the property of a dynamic array which makes some parts of the array
-        // live on heap, some off. At the same time we want a fixed size array. Therefore first create
-        // the array as a dynamic array and make it grow to the requested length.
-        IntArray array = newDynamicIntArray( fractionOf( length ), defaultValue );
-        array.at( length - 1 );
-        return array;
+        // Here we want to have the property of a dynamic array so that some parts of the array
+        // can live on heap, some off.
+        return newDynamicIntArray( fractionOf( length ), defaultValue );
     }
 
     @Override
     public ByteArray newByteArray( long length, byte[] defaultValue, long base )
     {
-        // Here we want to have the property of a dynamic array which makes some parts of the array
-        // live on heap, some off. At the same time we want a fixed size array. Therefore first create
-        // the array as a dynamic array and make it grow to the requested length.
-        ByteArray array = newDynamicByteArray( fractionOf( length ), defaultValue );
-        array.at( length - 1 );
-        return array;
+        // Here we want to have the property of a dynamic array so that some parts of the array
+        // can live on heap, some off.
+        return newDynamicByteArray( fractionOf( length ), defaultValue );
     }
 
     private long fractionOf( long length )
     {
-        int idealChunkCount = 10;
-        if ( length < idealChunkCount )
+        if ( length < MAGIC_CHUNK_COUNT )
         {
             return length;
         }
-        int maxArraySize = Integer.MAX_VALUE - Short.MAX_VALUE;
-        return min( length / idealChunkCount, maxArraySize );
+        return min( length / MAGIC_CHUNK_COUNT, MAX_ARRAY_SIZE );
     }
 
     @Override
@@ -105,6 +98,6 @@ public class ChunkedNumberArrayFactory extends NumberArrayFactory.Adapter
     @Override
     public String toString()
     {
-        return "CHUNKED_FIXED_SIZE";
+        return "ChunkedNumberArrayFactory with delegate " + delegate;
     }
 }
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArray.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArray.java
@@ -34,12 +34,10 @@ public interface NumberArray<N extends NumberArray<N>> extends MemoryStatsVisito
     /**
      * Swaps {@code numberOfEntries} items from {@code fromIndex} to {@code toIndex}, such that
      * {@code fromIndex} and {@code toIndex}, {@code fromIndex+1} and {@code toIndex} a.s.o swaps places.
-     *
-     * @param fromIndex where to start swapping from.
+     *  @param fromIndex where to start swapping from.
      * @param toIndex where to start swapping to.
-     * @param numberOfEntries number of entries to swap, starting from the given from/to indexes.
      */
-    void swap( long fromIndex, long toIndex, int numberOfEntries );
+    void swap( long fromIndex, long toIndex );
 
     /**
      * Sets all values to a default value.
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArray.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArray.java
@@ -32,8 +32,9 @@ public interface NumberArray<N extends NumberArray<N>> extends MemoryStatsVisito
     long length();
 
     /**
-     * Swaps {@code numberOfEntries} items from {@code fromIndex} to {@code toIndex}, such that
+     * Swaps items from {@code fromIndex} to {@code toIndex}, such that
      * {@code fromIndex} and {@code toIndex}, {@code fromIndex+1} and {@code toIndex} a.s.o swaps places.
+     * The number of items swapped is equal to the length of the default value of the array.
      *  @param fromIndex where to start swapping from.
      * @param toIndex where to start swapping to.
      */
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayFactory.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayFactory.java
@@ -19,12 +19,12 @@
  */
 package org.neo4j.unsafe.impl.batchimport.cache;
 
+import java.io.File;
 import java.util.Arrays;
 
 import org.neo4j.helpers.Exceptions;
+import org.neo4j.io.pagecache.PageCache;
 
-import static java.lang.Long.min;
-import static java.lang.Math.toIntExact;
 import static java.lang.String.format;
 import static org.neo4j.helpers.Exceptions.launderedException;
 import static org.neo4j.helpers.Format.bytes;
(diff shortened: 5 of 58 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/store/NeoStores.java
```diff
@@ -151,6 +151,11 @@ public File getStoreDir()
         return storeDir;
     }
 
+    public PageCache getPageCache()
```
MishaDemianenko: NeoStores should not be _and yet another page cache provider_ just pass that dependency where you need to use it

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
```diff
@@ -397,7 +397,7 @@ private int detectAndMarkCollisions( ProgressListener progress )
                     if ( dataIndexA > dataIndexB )
                     {
                         // Swap so that lower tracker index means lower data index. TODO Why do we do this?
```
MishaDemianenko, line 399: is this TODO still valid? looks like we know what we do here?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/Tracker.java
```diff
@@ -52,9 +52,8 @@
      *
```
MishaDemianenko, line 52: javadoc needs to be adapted, reference count parameter still

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public ChunkedNumberArrayFactory( NumberArrayFactory... delegateList )
+    {
+        delegate = new Auto( delegateList );
+    }
+
+    @Override
+    public LongArray newLongArray( long length, long defaultValue, long base )
+    {
+        // Here we want to have the property of a dynamic array which makes some parts of the array
```
MishaDemianenko: this comment should be just a javadoc to `newDynamicLongArray` instead of being copied 3 times

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
```diff
@@ -0,0 +1,110 @@
...
+        // live on heap, some off. At the same time we want a fixed size array. Therefore first create
+        // the array as a dynamic array and make it grow to the requested length.
+        LongArray array = newDynamicLongArray( fractionOf( length ), defaultValue );
+        array.at( length - 1 );
```
MishaDemianenko: rely on side effects of magic `at`?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
```diff
@@ -0,0 +1,110 @@
...
+        // live on heap, some off. At the same time we want a fixed size array. Therefore first create
+        // the array as a dynamic array and make it grow to the requested length.
+        ByteArray array = newDynamicByteArray( fractionOf( length ), defaultValue );
+        array.at( length - 1 );
+        return array;
+    }
+
+    private long fractionOf( long length )
+    {
+        int idealChunkCount = 10;
```
MishaDemianenko: magic 10? any reasons why it's 10?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
```diff
@@ -0,0 +1,110 @@
...
+    {
+        // Here we want to have the property of a dynamic array which makes some parts of the array
+        // live on heap, some off. At the same time we want a fixed size array. Therefore first create
+        // the array as a dynamic array and make it grow to the requested length.
+        ByteArray array = newDynamicByteArray( fractionOf( length ), defaultValue );
+        array.at( length - 1 );
+        return array;
+    }
+
+    private long fractionOf( long length )
```
MishaDemianenko: what if we will try to create a power of 2 chunks instead?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArray.java
```diff
@@ -34,12 +34,10 @@
     /**
      * Swaps {@code numberOfEntries} items from {@code fromIndex} to {@code toIndex}, such that
```
MishaDemianenko: javadoc should be updated as well

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public ByteArray newDynamicByteArray( long chunkSize, byte[] defaultValue )
+    {
+        return new DynamicByteArray( delegate, chunkSize, defaultValue );
+    }
+
+    @Override
+    public String toString()
+    {
+        return "CHUNKED_FIXED_SIZE";
```
MishaDemianenko: what if we will have lesst magic constant toString instead?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    private long fractionOf( long length )
+    {
+        int idealChunkCount = 10;
+        if ( length < idealChunkCount )
+        {
+            return length;
+        }
+        int maxArraySize = Integer.MAX_VALUE - Short.MAX_VALUE;
```
MishaDemianenko: why do we to substract short here?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayFactory.java
```diff
@@ -276,76 +276,19 @@ private Throwable error( long length, int itemSize, Throwable error )
...
-            return new DynamicByteArray( delegate, chunkSize, defaultValue );
-        }
-
-        @Override
-        public String toString()
-        {
-            return "CHUNKED_FIXED_SIZE";
-        }
-    };
+    NumberArrayFactory CHUNKED_FIXED_SIZE = new ChunkedNumberArrayFactory();
```
MishaDemianenko: fields should be moved on top

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayFactory.java
```diff
@@ -276,76 +276,19 @@ private Throwable error( long length, int itemSize, Throwable error )
...
 
     /**
      * {@link Auto} factory which uses JVM stats for gathering information about available memory.
      */
     NumberArrayFactory AUTO = new Auto( OFF_HEAP, HEAP, CHUNKED_FIXED_SIZE );
+
+    static NumberArrayFactory autoWithPageCacheFallback( PageCache pageCache, File dir )
```
MishaDemianenko: I think this method should still be called `auto` and if we want to highlight that it will have fall back mechanics that should be described in javadoc instead

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheByteArray.java
```diff
@@ -0,0 +1,355 @@
...
+import static org.neo4j.io.pagecache.PagedFile.PF_SHARED_READ_LOCK;
+import static org.neo4j.io.pagecache.PagedFile.PF_SHARED_WRITE_LOCK;
+
+public class PageCacheByteArray extends PageCacheNumberArray<ByteArray> implements ByteArray
+{
+    private final byte[] defaultValue;
+
+    public PageCacheByteArray( PagedFile pagedFile, long length, byte[] defaultValue, long base ) throws IOException
+    {
+        // '0' default value means we skip filling it out in super
```
MishaDemianenko: why not to introduce another constructor instead without passing default value?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheByteArray.java
```diff
@@ -0,0 +1,355 @@
...
+        super( pagedFile, defaultValue.length, length, 0, base );
```
MishaDemianenko: and why not to pass -1 directly here as a default value?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheByteArray.java
```diff
@@ -0,0 +1,355 @@
...
+        offset += offset( index );
+        try ( PageCursor cursor = pagedFile.io( pageId, PF_SHARED_READ_LOCK ) )
+        {
+            cursor.next();
+            long result;
+            do
+            {
+                long low4b = cursor.getInt( offset ) & 0xFFFFFFFFL;
+                long high2b = cursor.getShort( offset + Integer.BYTES );
+                result = low4b | (high2b << 32);
```
MishaDemianenko: 32 should be replaced by Integer.SIZE

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheByteArray.java
```diff
@@ -0,0 +1,355 @@
...
+        }
+    }
+
+    @Override
+    public void set( long index, byte[] value )
+    {
+        assert value.length == entrySize;
+        long pageId = pageId( index );
+        int offset = offset( index );
+        try ( PageCursor cursor = pagedFile.io( pageId, PF_SHARED_WRITE_LOCK | PF_NO_GROW ); )
```
MishaDemianenko: unnecessary semicolons here and in some other try blocks for page cursor creation

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheByteArray.java
```diff
@@ -0,0 +1,355 @@
...
+    @Override
+    public void set6ByteLong( long index, int offset, long value )
+    {
+        long pageId = pageId( index );
+        offset += offset( index );
+        try ( PageCursor cursor = pagedFile.io( pageId, PF_SHARED_WRITE_LOCK | PF_NO_GROW ); )
+        {
+            cursor.next();
+            cursor.putInt( offset, (int) value );
+            cursor.putShort( offset + Integer.BYTES, (short) (value >>> 32) );
```
MishaDemianenko: Integer.SIZE here as well

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
```diff
@@ -0,0 +1,178 @@
...
+    public void acceptMemoryStatsVisitor( MemoryStatsVisitor visitor )
+    {
+        visitor.offHeapUsage( length() * entrySize );
+    }
+
+    protected void checkBounds( PageCursor cursor )
+    {
+        if ( cursor.checkAndClearBoundsFlag() )
+        {
+            throw new IllegalStateException();
```
MishaDemianenko: this exception should provide a bit more details, at least pageId and offset that was used

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheByteArray.java
```diff
@@ -0,0 +1,355 @@
...
+
+import static org.neo4j.io.pagecache.PagedFile.PF_NO_GROW;
+import static org.neo4j.io.pagecache.PagedFile.PF_SHARED_READ_LOCK;
+import static org.neo4j.io.pagecache.PagedFile.PF_SHARED_WRITE_LOCK;
+
+public class PageCacheByteArray extends PageCacheNumberArray<ByteArray> implements ByteArray
+{
+    private final byte[] defaultValue;
+
+    public PageCacheByteArray( PagedFile pagedFile, long length, byte[] defaultValue, long base ) throws IOException
```
MishaDemianenko: this constructor can be package private

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCachedNumberArrayFactory.java
```diff
@@ -0,0 +1,87 @@
...
+
+import static java.nio.file.StandardOpenOption.CREATE;
+import static java.nio.file.StandardOpenOption.DELETE_ON_CLOSE;
+
+public class PageCachedNumberArrayFactory extends NumberArrayFactory.Adapter
+{
+    private final PageCache pageCache;
+    private final File storeDir;
+
+    public PageCachedNumberArrayFactory( PageCache pageCache, File storeDir )
```
MishaDemianenko: this can be package private

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCachedNumberArrayFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.IOException;
+import java.io.UncheckedIOException;
+
+import org.neo4j.io.pagecache.PageCache;
+import org.neo4j.io.pagecache.PagedFile;
+
+import static java.nio.file.StandardOpenOption.CREATE;
+import static java.nio.file.StandardOpenOption.DELETE_ON_CLOSE;
+
+public class PageCachedNumberArrayFactory extends NumberArrayFactory.Adapter
```
MishaDemianenko: a bit of javadocs would be nice

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheIntArray.java
```diff
@@ -0,0 +1,78 @@
...
+import org.neo4j.io.pagecache.PageCursor;
+import org.neo4j.io.pagecache.PagedFile;
+
+import static org.neo4j.io.pagecache.PagedFile.PF_NO_GROW;
+import static org.neo4j.io.pagecache.PagedFile.PF_SHARED_READ_LOCK;
+import static org.neo4j.io.pagecache.PagedFile.PF_SHARED_WRITE_LOCK;
+
+public class PageCacheIntArray extends PageCacheNumberArray<IntArray> implements IntArray
+{
+    public PageCacheIntArray( PagedFile pagedFile, long length, long defaultValue, long base ) throws IOException
```
MishaDemianenko: constructor can be package private

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheLongArray.java
```diff
@@ -0,0 +1,78 @@
...
+import org.neo4j.io.pagecache.PageCursor;
+import org.neo4j.io.pagecache.PagedFile;
+
+import static org.neo4j.io.pagecache.PagedFile.PF_NO_GROW;
+import static org.neo4j.io.pagecache.PagedFile.PF_SHARED_READ_LOCK;
+import static org.neo4j.io.pagecache.PagedFile.PF_SHARED_WRITE_LOCK;
+
+public class PageCacheLongArray extends PageCacheNumberArray<LongArray> implements LongArray
+{
+    public PageCacheLongArray( PagedFile pagedFile, long length, long defaultValue, long base ) throws IOException
```
MishaDemianenko: package private constructor here as well

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
```diff
@@ -0,0 +1,178 @@
...
+import java.io.UncheckedIOException;
+
+import org.neo4j.io.pagecache.PageCursor;
+import org.neo4j.io.pagecache.PagedFile;
+
+import static java.lang.Math.toIntExact;
+import static org.neo4j.io.pagecache.PagedFile.PF_NO_GROW;
+import static org.neo4j.io.pagecache.PagedFile.PF_SHARED_WRITE_LOCK;
+
+public abstract class PageCacheNumberArray<N extends NumberArray<N>> implements NumberArray<N>
```
MishaDemianenko: javadoc is missing

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
```diff
@@ -0,0 +1,178 @@
...
+{
+    protected final PagedFile pagedFile;
+    protected final int entriesPerPage;
+    protected final int entrySize;
+    private final long length;
+    private final long defaultValue;
+    private final long base;
+    private boolean closed;
+
+    protected PageCacheNumberArray( PagedFile pagedFile, int entrySize, long length,
```
MishaDemianenko: some of the protected items here can be package private

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
```diff
@@ -0,0 +1,178 @@
...
+        {
+            setDefaultValue( defaultValue );
+        }
+    }
+
+    private void setLength( PageCursor cursor, long length ) throws IOException
+    {
+        if ( !cursor.next( (length - 1) / entriesPerPage ) )
+        {
+            throw new IllegalStateException();
```
MishaDemianenko: Exception should contain some details like length and related size to help understand the problem when it will pop up

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    private long rebase( long index )
+    {
+        return index - base;
+    }
+
+    protected void setDefaultValue( long defaultValue ) throws IOException
+    {
+        if ( entrySize == Integer.BYTES )
```
MishaDemianenko: really strange to see this check here, this should be extracted to corresponding subclass instead

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
```diff
@@ -0,0 +1,178 @@
...
+        {
+            defaultValue |= defaultValue << 32;
```
MishaDemianenko: Integer.SIZE her as well

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    protected long pageId( long index )
+    {
+        return rebase( index ) / entriesPerPage;
+    }
+
+    protected int offset( long index )
+    {
+        return toIntExact( rebase( index ) % entriesPerPage ) * entrySize;
```
MishaDemianenko: whole expression should be wrapped into `toIntExact` not only `rebase( index ) % entriesPerPage`

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
```diff
@@ -0,0 +1,178 @@
...
+
+    protected void setDefaultValue( long defaultValue ) throws IOException
+    {
+        if ( entrySize == Integer.BYTES )
+        {
+            defaultValue |= defaultValue << 32;
+        }
+        try ( PageCursor writeCursor = pagedFile.io( 0, PF_SHARED_WRITE_LOCK | PF_NO_GROW ) )
+        {
+            writeCursor.next();
```
MishaDemianenko: would be nice to have bounds check for write cursors here as well

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -283,6 +283,11 @@ public CountsTracker getCountsStore()
         return neoStores.getCounts();
     }
 
+    public PageCache getPageCache()
```
MishaDemianenko: should not be your page cache provider, in a single usage place it's already available as a field

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/RelationshipGroupDefragmenter.java
```diff
@@ -43,6 +43,9 @@
  */
 public class RelationshipGroupDefragmenter
 {
+
+    private final NumberArrayFactory numberArrayFactory;
```
MishaDemianenko: should live together with other fields

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayPageCacheTestSupport.java
```diff
@@ -0,0 +1,63 @@
...
+import org.neo4j.test.rule.TestDirectory;
+
+public class NumberArrayPageCacheTestSupport
+{
+    public static Fixture prepareDirectoryAndPageCache( Class<?> testClass ) throws IOException
+    {
+        DefaultFileSystemAbstraction fileSystem = new DefaultFileSystemAbstraction();
+        TestDirectory testDirectory = TestDirectory.testDirectory( testClass, fileSystem );
+        File dir = testDirectory.prepareDirectoryForTest( "test" );
+        fileSystem.mkdirs( dir );
```
MishaDemianenko: you don't need to do it, testDirectory does that for you already

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/ByteArrayTest.java
```diff
@@ -20,34 +20,64 @@
...
 {
     private static final byte[] DEFAULT = new byte[25];
+    private static final int LENGTH = 1_000;
+    private static Fixture fixture;
 
     @Parameters
-    public static Collection<Supplier<ByteArray>> data()
+    public static Collection<Supplier<ByteArray>> data() throws IOException
     {
+        fixture = prepareDirectoryAndPageCache( ByteArrayTest.class );
```
MishaDemianenko, line 57: not sure why do you need fixture and manual interaction with directories and page cache, why not using class rules instead?

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/ByteArrayTest.java
```diff
@@ -20,34 +20,64 @@
...
+        PageCache pageCache = fixture.pageCache;
+        File dir = fixture.directory;
+        NumberArrayFactory autoWithPageCacheFallback = autoWithPageCacheFallback( pageCache, dir );
+        NumberArrayFactory pageCacheArrayFactory = new PageCachedNumberArrayFactory( pageCache, dir );
+        int chunkSize = LENGTH / 10;
```
MishaDemianenko: magic 10 again, hehe :)

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/IntArrayTest.java
```diff
@@ -62,29 +72,49 @@ public void shouldHandleSomeRandomSetAndGet() throws Exception
...
+        long fromValue = expected[fromIndex];
+        expected[fromIndex] = expected[toIndex];
+        expected[toIndex] = fromValue;
     }
 
     @Parameters
-    public static Collection<NumberArrayFactory> data()
+    public static Collection<NumberArrayFactory> data() throws IOException
+    {
+        fixture = prepareDirectoryAndPageCache( IntArrayTest.class );
```
MishaDemianenko, line 106: same wondering about fixture as before

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArrayConcurrencyTest.java
```diff
@@ -0,0 +1,88 @@
...
+    @Test
+    public void shouldHandleConcurrentAccessToDifferentData() throws Throwable
+    {
+        doRace( this::fileRangeRacer );
+    }
+
+    private void doRace( BiFunction<NumberArray,Integer,Runnable> contestantCreator ) throws Throwable
+    {
+        PageCache pageCache = pageCacheRule.getPageCache( fs );
+        PagedFile file = pageCache.map( dir.file( "file" ), pageCache.pageSize(), CREATE, DELETE_ON_CLOSE );
```
MishaDemianenko, line 69: this file needs to be closed

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheByteArrayConcurrencyTest.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        @Override
+        public void run()
+        {
+            for ( int o = 0; o < LAPS; o++ )
+            {
+                for ( int i = 0; i < COUNT; i++ )
+                {
+                    byte[] value = {1, 2, 3, 4};
```
MishaDemianenko, line 64: would be better if this value was random instead

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheIntArrayConcurrencyTest.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        @Override
+        public void run()
+        {
+            for ( int o = 0; o < LAPS; o++ )
+            {
+                for ( int i = 0; i < COUNT; i++ )
+                {
+                    array.set( i, i );
```
MishaDemianenko, line 64: better to have random value

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheLongArrayConcurrencyTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+        @Override
+        public void run()
+        {
+
+            for ( int o = 0; o < LAPS; o++ )
+            {
+                for ( long i = 0; i < COUNT; i++ )
+                {
+                    array.set( i, i );
```
MishaDemianenko, line 65: better to have random value

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheLongArrayTest.java
```diff
@@ -0,0 +1,97 @@
...
+    private final PageCacheRule pageCacheRule = new PageCacheRule();
+
+    @Rule
+    public final RuleChain ruleChain = RuleChain.outerRule( fs ).around( dir ).around( random ).around( pageCacheRule );
+
+    @Test
+    public void verifyPageCacheLongArray() throws Exception
+    {
+        PageCache pageCache = pageCacheRule.getPageCache( fs );
+        PagedFile file = pageCache.map( dir.file( "file" ), pageCache.pageSize(), CREATE, DELETE_ON_CLOSE );
```
MishaDemianenko, line 55: should be part of try with resources block

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NumberArrayFactory.java
```diff
@@ -276,76 +276,19 @@ private Throwable error( long length, int itemSize, Throwable error )
...
-
-        @Override
-        public String toString()
-        {
-            return "CHUNKED_FIXED_SIZE";
-        }
-    };
+    NumberArrayFactory CHUNKED_FIXED_SIZE = new ChunkedNumberArrayFactory();
 
     /**
```
MishaDemianenko: I think we need to change a way how we keep exception in case if we fail to create array using all of the candidates, better keep each exception instead of overwriting it with the last one to see the actual reason instead of guessing it in case if all candidates will fail

MishaDemianenko (changes requested):
Nice idea in general. Do we have any real run results that were backed by newly introduced arrays types? 
Still have couple of other comments here and there

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
ragadeeshu: Not that I know of. @tinwelint or @SRaghuram might know.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
ragadeeshu: The maximum length of an array varies from JVM to JVM. It is usually a small number (e.g. 5) less than Inteter.MAX_VALUE. We subtract a short here in order to be on the safe side.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
ragadeeshu: Actually, I'm having a hard time finding what this is supposed to do.

ragadeeshu: I think that this is some leftovers. Passing -1 seems to do nothing. If we actually passed the desired length here, it would do something.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
ragadeeshu: I have no idea how that would behave, other than that we would end up with a larger amount of chunks.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheByteArray.java
ragadeeshu: Because we need to assign the defaultValue field in the bytearray class before the call to setdefaultvalue, and we cannot do assignments before the call to super.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
MishaDemianenko: If that is the reason please create constant for that once and reference some of the JDK classes where MAX_ARRAY_SIZE constant is defined for example ArrayList

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
ragadeeshu: Again, maybe @tinwelint knows.

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/ByteArrayTest.java
ragadeeshu: This is needed due to some unfortunate interaction between class rules, parameterized tests and testdirectories, IIRC.

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheByteArrayConcurrencyTest.java
ragadeeshu: The `WholeFileRacer` has non-random data for a reason, it is fine for it to read another threads data, and thus we cannot put random data here. This test is to make sure that we do not read junk or crash when two threads access the same data :)

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheIntArrayConcurrencyTest.java
ragadeeshu: The `WholeFileRacer` has non-random data for a reason, it is fine for it to read another threads data, and thus we cannot put random data here. This test is to make sure that we do not read junk or crash when two threads access the same data :)

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheLongArrayConcurrencyTest.java
ragadeeshu: The `WholeFileRacer` has non-random data for a reason, it is fine for it to read another threads data, and thus we cannot put random data here. This test is to make sure that we do not read junk or crash when two threads access the same data :)

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArrayConcurrencyTest.java
ragadeeshu: The numberArray will close this file upon close. But I guess it won't hurt :)

ragadeeshu:
@MishaDemianenko Updated. No real huge run yet, but there is plans for one.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/ChunkedNumberArrayFactory.java
SRaghuram: Yup, I had the same Q to Mattias, and he said - no specific reason, just looked reasonable.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
```diff
@@ -0,0 +1,188 @@
...
+            int pageSize = pagedFile.pageSize();
+            fillPageWithDefaultValue( writeCursor, defaultValue, pageSize );
+            if ( pageId( length - 1 ) > 0 )
+            {
+                try ( PageCursor cursor = pagedFile.io( 1, PF_NO_GROW | PF_SHARED_WRITE_LOCK ) )
+                {
+                    while ( cursor.next() )
+                    {
+                        writeCursor.copyTo( 0, cursor, 0, pageSize );
```
MishaDemianenko, line 110: we missing bounds check here still

MishaDemianenko (changes requested):
One small comment otherwise looks good, would be really nice to have some results to see if this actually helps. Can we start some test import on one of our test machines with some native memory limitations to force usage of the new strategy?

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheLongArrayTest.java
chrisvest: No, the `PageCacheLongArray` will close the file when it is closed.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/PageCacheNumberArray.java
chrisvest: This has been added.

chrisvest:
@MishaDemianenko I think all comments were already addressed.

MishaDemianenko (approved):
That was the last open thing, approved! Still would be nice to get some numbers from runs.

chrisvest:
@jexp confirmed that it can import with larger-than-memory caches now.

PR #9680, opened 2017-07-18
Made sure that cardinality estimation doesn't reach Infinity or NaN

SaschaPeukert:
Due to numerical overflows it is possible to get cardinality estimation of infinity or NaN, and this leads to bad plans. It is more meaningful to use Double.MaxValue for unknown very large numbers, and allow multiplication by zero to always get zero.

changelog: Fixed Cypher planning problems due to numerical overflow when querying a large graph with a very long query


community/kernel/src/main/java/org/neo4j/kernel/impl/util/dbstructure/InvocationTracer.java
```diff
@@ -131,7 +131,7 @@ private void formatPreamble( Appendable builder ) throws IOException
         }
         formatln( builder );
         formatln( builder, "//" );
-        formatln( builder, "// GENERATED FILE. DO NOT EDIT. " );
+        formatln( builder, "// GENERATED FILE. DO NOT EDIT." );
```
alexaverbuch, line 134: @craigtaverner @SaschaPeukert I suspect this should not be committed?

community/kernel/src/test/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureLargeOptionalMatchStructure.java
```diff
@@ -0,0 +1,111 @@
+/*
```
alexaverbuch, line 1: @craigtaverner @SaschaPeukert I suspect this should not be committed?

alexaverbuch (changes requested):
I think you have committed unintended changes

community/kernel/src/main/java/org/neo4j/kernel/impl/util/dbstructure/InvocationTracer.java
craigtaverner: That was deliberately included because it turns out that the generated code fails the checkstyle. Even though we manually edited the generated file and fixed this manually, we thought it correct to include this fix in the generator so that the next person to use this capability would not get burned as well. While we could have made a separate PR for this, a whole PR for removal of one space character seems silly, and this problem was related to our work, so this seemed best.

community/kernel/src/test/java/org/neo4j/kernel/impl/util/dbstructure/DbStructureLargeOptionalMatchStructure.java
craigtaverner: This is part of the test. This file represents the relevant subset of the real statistics of a real database that was originally used to reproduce the problem. The fact that the generated statistics structure is saved to kernel, while the test is in cypher is a bit odd, but that is the pattern used by all previous similar tests, so we kept it for consistency.

alexaverbuch (approved):

PR #9682, opened 2017-07-18
Allow checkpoint finder to find transaction after checkpoint in large files

MishaDemianenko:
Update log positions comparison rules to be ready for cases when bytes
offsets or version differences are bigger than an integer type capacity.
This change will allow to properly compare records on any distant positions
from each other and for example, detect transactions records that come after
the checkpoint records in large tx log files.


chrisvest (approved):

chrisvest:
@MishaDemianenko Good stuff.

PR #9687, opened 2017-07-20
Make sure to LIMIT before EAGER

systay:


community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/planner/logical/plans/rewriter/cleanUpEager.scala
```diff
@@ -37,6 +37,10 @@ case object cleanUpEager extends Rewriter {
     // E LCSV => LCSV E
     case eager@Eager(loadCSV@LoadCSV(source, _, _, _, _, _)) =>
       loadCSV.copy(source = eager.copy(inner = source)(eager.solved))(eager.solved)
+
+    // E LIMIT => LIMIT E
+    case eager@Eager(limit@Limit(source, _, _)) =>
```
Mats-SX: I think this should be flipped -- `Limit` before `Eager` would reduce the size of the eagerised set of records.

Mats-SX (approved):

PR #9688, opened 2017-07-20
Open legacy index in writable mode for non single instance read only environments

MishaDemianenko:
Open legacy indexes in a writable mode for all read only environments
except single instance. We need to allow indexes to be writable since
they need to be updated during catch up updates.


community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
```diff
@@ -113,21 +115,23 @@ public String toString()
     /**
      * Constructs this data source.
      */
-    public LuceneDataSource( File storeDir, Config config, IndexConfigStore indexStore, FileSystemAbstraction fileSystemAbstraction )
+    public LuceneDataSource( File storeDir, Config config, IndexConfigStore indexStore, FileSystemAbstraction
+            fileSystemAbstraction, OperationalMode operationalMode )
```
burqen: Formatting. Let's keep parameter declaration and name on same line.

burqen (approved):
Changes look good 🎉 
Just a little bit of bikeshedding.

PR #9692, opened 2017-07-20
Prevent database healing in case of critical errors

MishaDemianenko:
Do not allow to heal database health in case of critical exception was
observed.
So far only OutOfMemoryError considered as critical exception.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
@@ -20,6 +20,7 @@
 package org.neo4j.kernel.internal;
 
 import org.neo4j.graphdb.event.ErrorState;
+import org.neo4j.helpers.Exceptions;
 import org.neo4j.kernel.impl.core.DatabasePanicEventGenerator;
 import org.neo4j.logging.Log;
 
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
```diff
@@ -88,21 +89,35 @@ public void panic( Throwable cause )
...
+            healthy = true;
+            causeOfPanic = null;
+            log.info( "Database health set to OK" );
+            return true;
+        }
+    }
+
+    private boolean hasCriticalFailure()
+    {
+        return !isHealthy() && Exceptions.contains( causeOfPanic, CRITICAL_EXCEPTIONS );
```
burqen, line 120: `hasCriticalFailure` will only report failure to be critical if first `cause` when `panic` was called was critical. Else, if some other `cause` that IS critical is later reported (before `healed`), then this will be ignored.

community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
```diff
@@ -29,9 +30,9 @@
 {
     private static final String panicMessage = "Database has encountered some problem, "
             + "please perform necessary action (tx recovery/restart)";
+    private static final Class<?>[] CRITICAL_EXCEPTIONS = new Class[]{OutOfMemoryError.class};
```
burqen, line 33: Does it make sense to use `VirtualMachineError.class` here?

burqen (changes requested):
Changes look good 🎉  
Got two suggestions for possible improvements.

community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
MishaDemianenko: not really since for example, `StackOverflowError` is not that critical for us

community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
MishaDemianenko: We talked and agree that for now, we will keep things as it is - we will only inspect root cause of original panic

MishaDemianenko:
@burqen looks good otherwise right and can be merged in?

burqen (approved):

burqen:
@MishaDemianenko Yes, PR approved 🎉

PR #9708, opened 2017-07-25
Added nullability support to Registered ExpandAll

alexaverbuch:


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/ExpandAllRegisterPipe.scala
```diff
@@ -55,10 +58,16 @@ case class ExpandAllRegisterPipe(source: Pipe,
...
-          outputRow.setLongAt(toOffset, otherSide)
+          outputRow.setLongAt(relSlot.offset, relId)
+          outputRow.setLongAt(toSlot.offset, otherSide)
           outputRow
         })
     }
   }
+
+  private def resultForNull(slot: Slot) =
+    if (slot.nullable)
```
systay: IMHO - There is no need to check the nullability in the tight loop like this. We want to do as little work as possible in `Pipe`s, which means trusting other components to do their thing correctly.

Same comment applies to the ExpandInto logic

craigtaverner (approved):

PR #9710, opened 2017-07-25
Use standard TestDirectory as test folder provider for EmbeddedDatabaseRule

MishaDemianenko:
Use standard TestDirectory rule to provide test directory required for
embedded database, instead of yet another custom implementation of TemporaryFolder rule.
Remove redundant class constructor argument.
Cleanup.


burqen (approved):
PR approved 👍  
However it needs rebase and docs_compile need an update when this is merged.

PR #9711, opened 2017-07-25
Added NodeIndexScanRegisterPipeline

eebus:


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/NodeIndexScanRegisterPipe.scala
```diff
@@ -0,0 +1,53 @@
...
+                                    (val id: Id = new Id)
+  extends Pipe {
+
+
+  private val offset = pipelineInformation.getLongOffsetFor(ident)
+
+  private val descriptor = IndexDescriptor(label.nameId.id, propertyKey.nameId.id)
+
+  protected def internalCreateResults(state: QueryState): Iterator[ExecutionContext] = {
+    val baseContext = state.createOrGetInitialContext()
```
systay: This method is not the register version you are looking for. You should instead create the ExecutionContext just like you are doing, but instead of using `copyFrom`, you should use `copyArgumentStateTo`.  You can see how it is done [here](https://github.com/neo4j/neo4j/blob/3.3/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/AllNodesScanRegisterPipe.scala#L35).

systay (changes requested):

systay (approved):

PR #9714, opened 2017-07-26
Cleanup of ExecutionContext and its use

systay:


craigtaverner:
Personally I think the tuple and map versions read better than the newWith1 and sequence versions. Is there a functional reason for moving to the latter? Performance?

systay:
@craigtaverner Yes. Performance is the key. The tuple method was creating a Tuple object just to pass it into the ExecutionContext

alexaverbuch:
@systay out of curiosity, did you run any benchmarks to measure the impact of these changes?

PR #9715, opened 2017-07-26
Moved TopologyService up into classes that use StoreCopyClient

phughk:
This change is related to improvements in the backup client.
We would like to re-use the `StoreCopyClient` in different backup scenarios. In order to reuse this class we cannot rely on MemberId for being the default way of connecting to existing databases.
This change moves the `TopologyService` out of `StoreCopyClient` and into all the existing classes dependant on it, so that `StoreCopyClient` isn't reliant on existing within a neo4j cluster.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Optional;
 
 import org.neo4j.causalclustering.core.CausalClusteringSettings;
+import org.neo4j.causalclustering.core.state.snapshot.TopologyLookupException;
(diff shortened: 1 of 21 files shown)


digitalstain (approved):

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
```diff
@@ -93,9 +94,9 @@ public ReadReplicaTopology readReplicas()
     }
 
     @Override
-    public Optional<AdvertisedSocketAddress> findCatchupAddress( MemberId memberId )
+    public AdvertisedSocketAddress findCatchupAddress( MemberId memberId )
     {
-        return Optional.ofNullable( catchupAddressMap.get( memberId ) );
+        return Optional.ofNullable( catchupAddressMap.get( memberId ) ).orElseThrow( () -> new TopologyLookupException( memberId ) );
```
phughk, line 99: Remove Optional - no
Expceiton -> RuntimeExecption - yes

PR #9716, opened 2017-07-26
Periodic cleanup of kernel unused classes, methods, wrong javadocs and comments

MishaDemianenko:


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/tracing/TransactionEvent.java
```diff
@@ -81,7 +81,7 @@ public void setReadOnly( boolean wasReadOnly )
 
     /**
      * Set the type of the transaction, as given by
-     * {@link org.neo4j.kernel.impl.api.KernelTransactionImplementation.TransactionType}.
+     * {@link org.neo4j.kernel.impl.api.KernelTransactionImplementation.TransactionWriteState}.
      */
     void setTransactionType( String transactionTypeName );
```
burqen: `TransactionEvent` has no implementations that is not `no_op`. Is this part of public API? If so, why does it live in `impl`? If not, can we just delete it completely?

burqen (changes requested):
Changes look good! Nice to have a little bit cleaner house 👍  Just one comment.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/tracing/TransactionEvent.java
burqen: Nevermind, my bad. It _has_ a real implementation.

burqen: Although, may I suggest changing method and parameter name to reflect the change from "transactionType" to "transactionWriteState"?

MishaDemianenko: @burqen It's not a public API, and it looks reasonable so why not. Updated.

burqen (approved):

PR #9718, opened 2017-07-26
Remove channels on core topology change

mneedham:
At the moment if a server goes away and comes back
with the same hostname but different underlying IP
address we won't be able to do any Raft communication
with it anymore because we cache the channel in SenderService.

This commit adds ChannelExpiryService which translates
core topology change events -> channel expired events that
the SenderService can react to

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastCoreTopologyService.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastCoreTopologyService.java
@@ -19,6 +19,12 @@
  */
 package org.neo4j.causalclustering.discovery;
 
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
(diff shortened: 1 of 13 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastCoreTopologyService.java
```diff
@@ -110,7 +111,7 @@
     public void addCoreTopologyListener( Listener listener )
     {
         listenerService.addCoreTopologyListener( listener );
-        listener.onCoreTopologyChange( coreTopology );
+        listener.onCoreTopologyChange( coreTopology, coreTopology);
```
phughk, line 114: pedantic space before end of bracket

phughk (approved):

mneedham:
We don't need this as @martinfurmanski recently fixed it with this PR - https://github.com/neo4j/neo4j/pull/9679.

Will close.

PR #9719, opened 2017-07-26
Relax bytes expectation assumptions of ListQueriesProcedureTest

MishaDemianenko:
Remove assumption that second run will allocate fewer bytes then first.
It can be the case that there will be no re-planning for both cases and the second run will take a few bytes more, so relax that assumption.


chrisvest (approved):
Looks good.

PR #9722, opened 2017-07-27
Uniqueness check with different duplicate check strategies.

MishaDemianenko:
Introduce duplicate checking strategies:
use simple map strategy for cases when expected number of elements is
small enough, use dynamic bucket based array strategy in case if expected
number of elements is quite high.
Proposed changes speedups things significantly and for quite big numbers of
duplicates allow create unique indexes in seconds instead of minutes.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
@@ -38,7 +38,7 @@ public class IndexEntryConflictException extends Exception
     private final long addedNodeId;
     private final long existingNodeId;
 
-    public IndexEntryConflictException( long existingNodeId, long addedNodeId, Object propertyValue )
+    public IndexEntryConflictException( long existingNodeId, long addedNodeId, Object... propertyValue )
     {
         this( existingNodeId, addedNodeId, OrderedPropertyValues.ofUndefined( propertyValue ) );
     }
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
@@ -38,9 +38,9 @@ public class IndexEntryConflictException extends Exception
     private final long addedNodeId;
     private final long existingNodeId;
 
-    public IndexEntryConflictException( long existingNodeId, long addedNodeId, Object... propertyValue )
+    public IndexEntryConflictException( long existingNodeId, long addedNodeId, Object... propertyValues )
     {
-        this( existingNodeId, addedNodeId, OrderedPropertyValues.ofUndefined( propertyValue ) );
+        this( existingNodeId, addedNodeId, OrderedPropertyValues.ofUndefined( propertyValues ) );
     }
 
     public IndexEntryConflictException( long existingNodeId, long addedNodeId, OrderedPropertyValues propertyValues )
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/verification/DuplicateCheckStrategy.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/verification/DuplicateCheckStrategy.java
@@ -140,9 +140,9 @@ abstract class DuplicateCheckStrategy
             public int hashCode()
             {
                 int result = 0;
-                for ( Property property : properties )
+                for ( Object value : values )
                 {
-                    result = 31 * (result + property.hashCode());
+                    result = 31 * (result + value.hashCode());
                 }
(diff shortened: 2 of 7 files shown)


MishaDemianenko:
<b>N.B.: should be null merged to 3.3 because of values refactoring, some changes will be required.</b>

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/index/IndexEntryConflictException.java
```diff
@@ -38,7 +38,7 @@
     private final long addedNodeId;
     private final long existingNodeId;
 
-    public IndexEntryConflictException( long existingNodeId, long addedNodeId, Object propertyValue )
+    public IndexEntryConflictException( long existingNodeId, long addedNodeId, Object... propertyValue )
```
burqen: Could use `propertyValues` to reflect var args in parameter name.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/verification/DuplicateCheckStrategy.java
```diff
@@ -0,0 +1,274 @@
...
+                {
+                    return true;
+                }
+                if ( o == null || getClass() != o.getClass() )
+                {
+                    return false;
+                }
+
+                PropertyValues that = (PropertyValues) o;
+                return propertyValuesEqual( properties, that.values );
```
burqen, line 136: `equals` and `hashCode` does not align. Both property key and value is used to calculate `hashCode` but equals only check property value, which means two `PropertyValues` object can be equal but have different hash code. For the use case that we have this is fine, but confusing. 

Suggestion: Change the hash code function to only check values array instead to make it less confusing.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/verification/DuplicateCheckStrategy.java
```diff
@@ -0,0 +1,274 @@
...
+     */
+    static class BucketsDuplicateCheckStrategy extends DuplicateCheckStrategy
+    {
+        private static final int BASE_ENTRY_SIZE = 1000;
+        private static final int DEFAULT_BUCKETS = 10;
+        static final int BUCKET_STRATEGY_ENTRIES_THRESHOLD = BASE_ENTRY_SIZE * DEFAULT_BUCKETS;
+
+        private static final int MAX_NUMBER_OF_BUCKETS = 100;
+        private final int numberOfBuckets;
+        private EntrySet[] actualValues;
```
burqen: I was a bit confused about what was the "buckets". Can we improve naming to make this more clear?

Something like (but probably not):
`actualValues` -> `buckets`
`EntrySet` -> `BucketNode`
`entrySetSize` -> `bucketNodeSize`

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/verification/DuplicateCheckStrategy.java
```diff
@@ -0,0 +1,274 @@
...
+            EntrySet current = actualValues[bucket];
+            if ( current == null )
+            {
+                current = new EntrySet( entrySetSize );
+                actualValues[bucket] = current;
+            }
+            return current;
+        }
+
+        private static class EntrySet
```
burqen: A small javadoc about this guy being a linked list where each entry contain `nodeId`:`propertyValue` pairs would be nice.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/verification/PartitionedUniquenessVerifier.java
```diff
@@ -133,18 +131,55 @@ private Terms termsForField( String fieldName ) throws IOException
...
-             * Here {@link DuplicateCheckingCollector#reset()} is deliberately not called to preserve accumulated
-             * state (knowledge about duplicates) across all {@link IndexSearcher#search(Query, Collector)} calls.
-             */
-            SimpleCollector collector = DuplicateCheckingCollector.forProperties( accessor, propertyKeyIds );
             for ( PartitionSearcher searcher : searchers )
             {
+                /*
+                 * Here {@link DuplicateCheckingCollector#init}} is deliberately not called to preserve accumulated
+                 * state (knowledge about duplicates) across all {@link IndexSearcher#search(Query, Collector)} calls.
+                 */
```
burqen, line 182: Nice clarification!

burqen (changes requested):
The map approach has O(1) in time complexity for `checkForDuplicate`.
Bucket approach has O(n).
This makes a little bit worried. Do we have any benchmarks that verifies that this is not a significant regression?

MishaDemianenko:
@burqen updated

burqen (approved):

PR #9725, opened 2017-07-27
Unify numeric types conversion utilities. Introduce Numbers.

MishaDemianenko:
Remove duplicated conversions across code base that perform
safe conversion from one numeric type to another. Unify them and cleanup.

### DIFF
--- a/community/common/src/main/java/org/neo4j/helpers/Numbers.java
+++ b/community/common/src/main/java/org/neo4j/helpers/Numbers.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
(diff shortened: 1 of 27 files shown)


community/common/src/main/java/org/neo4j/helpers/Numbers.java
```diff
@@ -0,0 +1,74 @@
...
+    public static short safeCastIntToUnsignedShort( int value )
+    {
+        if ( (value & ~0xFFFF) != 0 )
+        {
+            throw new ArithmeticException( getOverflowMessage( value, "unsigned short" ) );
+        }
+        return (short) value;
+    }
+
+    public static int safeCastLongToInt( long value )
```
tinwelint, line 33: Isn't this `Math.toIntExact` btw?

MishaDemianenko: Almost, except error message that aligns with other safe casts error messages

community/common/src/test/java/org/neo4j/helpers/NumbersTest.java
```diff
@@ -0,0 +1,120 @@
...
+        assertEquals(Short.MAX_VALUE, safeCastLongToShort( Short.MAX_VALUE ));
+        assertEquals(Short.MIN_VALUE, safeCastLongToShort( Short.MIN_VALUE ));
+    }
+
+    @Test
+    public void castIntToUnsighedShort()
+    {
+        assertEquals(1, safeCastIntToUnsignedShort( 1 ));
+        assertEquals(10, safeCastIntToUnsignedShort( 10 ));
+        assertEquals(Short.MAX_VALUE, safeCastIntToUnsignedShort( Short.MAX_VALUE ));
```
burqen: We are not testing the largest value for an unsigned short here.

burqen (changes requested):
Found a missing test case, otherwise this looks fine 🎉 
Nice cleanup!

MishaDemianenko:
@burqen updated

burqen (approved):

PR #9733, opened 2017-07-28
Fix a bug (3.3 only) that could cause the importer to produce inconsistent stores

chrisvest:


tinwelint (approved):
... and wonderful commit messaging, Chris

PR #9735, opened 2017-07-30
Cypher Register work - projection and unknown property keys

systay:


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/ExecutionContext.scala
```diff
@@ -30,25 +30,29 @@ object ExecutionContext {
...
+  def apply(numberOfLongs: Int, numberOfRefs: Int) = new ExecutionContext(m = null, numberOfLongs, numberOfRefs)
 }
 
-case class ExecutionContext private (m: MutableMap[String, Any], numberOfLongs: Int)
+case class ExecutionContext private (m: MutableMap[String, Any], numberOfLongs: Int, numberOfRefs: Int)
   extends MutableMap[String, Any] {
 
   def copyFrom(input: ExecutionContext): Unit = {
-    if (input.numberOfLongs > numberOfLongs)
+    if (input.numberOfLongs > numberOfLongs || input.numberOfRefs > numberOfRefs)
```
alexaverbuch: Is this check necessary in the tight loop?
We will get `java.lang.ArrayIndexOutOfBoundsException` if input is larger.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/RegisteredPipeBuilder.scala
```diff
@@ -109,8 +117,11 @@ class RegisteredPipeBuilder(fallback: PipeBuilder,
             k -> runtimeExpressions.NodeFromRegister(offset)
           case LongSlot(offset, true, CTNode, _) =>
             k -> runtimeExpressions.NullCheck(offset, runtimeExpressions.NodeFromRegister(offset))
+          case RefSlot(offset, _, _, _) =>
+            k -> runtimeExpressions.ReferenceFromRegister(offset)
+
           case _ =>
-            throw new InternalException("Did not find " + k + "in the pipeline information1")
+            throw new InternalException("Did not find `" + k + "` in the pipeline information1")
```
alexaverbuch, line 131: How about: `s"Did not find '$k' in pipeline information"`?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/ProjectionRegisterPipe.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  expressions.values.foreach(_.registerOwningPipe(this))
+
+  protected def internalCreateResults(input: Iterator[ExecutionContext], state: QueryState): Iterator[ExecutionContext] = {
+    input.map {
+      ctx =>
+        expressions.foreach {
+          case (offset, expression) =>
+            val result = expression(ctx)(state)
+            ctx.setRefAt(offset, result)
```
alexaverbuch: why do we never do `setLongAt()`?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/ExecutionContext.scala
```diff
@@ -52,10 +52,11 @@ case class MapExecutionContext(m: MutableMap[String, Any])
   override def copyFrom(input: ExecutionContext): Unit = fail()
 
   override def setLongAt(offset: Int, value: Long): Unit = fail()
-
   override def getLongAt(offset: Int): Long = fail()
 
   private def fail(): Nothing = throw new InternalException("Tried using a map context as a primitive context")
+  def setRefAt(offset: Int, value: Any): Unit = fail()
```
craigtaverner: These differ from the get/setLongAt in two ways, the use of `override` and the line spacing.

craigtaverner: On searching, I find no uses either. I think these should be removed.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisteredRewriter.scala
```diff
@@ -39,6 +39,22 @@ class RegisteredRewriter(tokenContext: TokenContext) {
     val newPipelineInfo = mutable.HashMap[LogicalPlan, PipelineInformation]()
     var rewrites = Map[LogicalPlan, LogicalPlan]()
     val rewritePlanWithRegisters = topDown(Rewriter.lift {
+      // TODO: This is a hack. The rewrite framework should do it for us, but it is not going into
+      // the expressions map. Will need to revisit, just not right now.
+      case p@Projection(source, expressions) =>
```
craigtaverner: This looks very much the same as the rewriting below, except for the additional expression values rewriting, so perhaps that could be embedded in the code below?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/ProjectionRegisterPipe.scala
systay: Maybe we should.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/expressions/NodeProperty.scala
```diff
@@ -34,3 +34,20 @@ case class NodeProperty(offset: Int, token: Int) extends Expression {
 
   override def symbolTableDependencies: Set[String] = Set.empty
 }
+
+case class NodePropertyLate(offset: Int, propKey: String) extends Expression {
+
+  override def apply(ctx: ExecutionContext)(implicit state: QueryState): Any = {
+    val maybeToken = state.query.getOptPropertyKeyId(propKey)
+    if (maybeToken.isEmpty)
```
craigtaverner, line 42: Could use a map(...).getOrElse(null) here for simpler code.

craigtaverner (approved):

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisteredRewriter.scala
systay: Not sure what you mean here. Like the comment says - the case below should be enough already.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/expressions/NodeProperty.scala
systay: We should look at the produced byte code. 
In these tight loops, I am not after simple code. I'm after fast code. My concern with the `map()` alternative is that we would be creating another `Some` object.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/ProjectionRegisterPipe.scala
systay: What I meant with that is - I'm sure we can figure out logic to write longs to the long arrays, but this is a first simple step that should still move us forward.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/expressions/NodeProperty.scala
alexaverbuch: @systay @craigtaverner why not remove the `QueryContext#getOpt*()` methods altogether, and replace with `QueryContext#has*()` + `QueryContext#get*()`?
at least in the new Java runtime

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisteredRewriter.scala
```diff
@@ -39,6 +39,25 @@ class RegisteredRewriter(tokenContext: TokenContext) {
...
+       */
+      case oldPlan@Projection(_, expressions) =>
+        val information = pipelineInformation(oldPlan)
+        val rewriter = rewriteCreator(information, oldPlan)
+
+        val newExpressions = expressions collect {
+          case (column, expression) if !expression.isInstanceOf[Variable] => column -> expression.endoRewrite(rewriter)
+        }
+
+        val newPlan = oldPlan.copy(expressions = newExpressions)(oldPlan.solved)
```
craigtaverner, line 54: Shouldn't the variables be kept, just not re-written? This code looks like it re-writes the variables away entirely.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/expressions/NodeFromRegister.scala
```diff
@@ -35,3 +35,15 @@ case class NodeFromRegister(offset: Int) extends Expression {
 
   override def symbolTableDependencies: Set[String] = Set.empty
 }
+
+case class ReferenceFromRegister(offset: Int) extends Expression {
```
craigtaverner: Why is this not in its own file? Does not really make sense in the NodeFromRegister.scala file

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisteredRewriter.scala
craigtaverner: Looking down at the ProjectionRegisteredPipe I see the parameter `introducedExpression`, and modification of the incoming context instead of creation of a new one, so this makes me begin to think the above code is correct in the sense that it only needs to know about the modified columns, which are the expressions to be run during projection. I am wondering however about projections that remove columns, but that is presumably already handled in the pipeline planning, which defines all columns that should be known about at the time.

enterprise/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/RegisteredPipeBuilderTest.scala
```diff
@@ -305,7 +305,7 @@ class RegisteredPipeBuilderTest extends CypherFunSuite with LogicalPlanningTestS
     )())
   }
 
-  test("aggregation used for distinct") {
+  ignore("aggregation used for distinct") {
```
craigtaverner, line 308: Are we ignoring these because we are in an intermediate state and only want to fix the tests once some additional work is done? Would be nice for a TODO here stating what work still needs to be done. I'm guessing the TODO is aggregation?

craigtaverner (changes requested):

craigtaverner (approved):

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/expressions/ReferenceFromRegister.scala
```diff
@@ -0,0 +1,36 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher.internal.compatibility.v3_3.runtime.interpreted.expressions
+
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.ExecutionContext
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.commands.expressions.Expression
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.pipes.QueryState
+
+case class ReferenceFromRegister(offset: Int) extends Expression {
```
alexaverbuch, line 26: 👍

PR #9737, opened 2017-07-31
upgrading driver version

praveenag:


praveenag:
Making this change in `3.2` to fix flaky tests and to be forward merged.

praveenag:
Making this change in `3.2` to fix flaky tests and to be forward merged.

PR #9741, opened 2017-07-31
Disable duplicates tests in LuceneSchemaIndexUniquenessVerificationIT on power8

MishaDemianenko:
Previously we had sporadic failures of *WithDuplicates tests with
CorruptedIndexException inside of lucene on power8 that we did not find a way how to
address, this PR will disable those tests in case if we running on that
platform for now


MishaDemianenko:
<b>should be null merged into 3.3</b>

PR #9743, opened 2017-07-31
Combined schema index provider

tinwelint:
Basically a `SchemaIndexProvider` that combines two SchemaIndexProviders, one native provider with a limited set of capabilities (numbers for now) and one to fall back everything else to (the standard lucene provider for now).

The combination starts at the `SchemaIndexProvider` and extends all the way out to the `IndexPopulator`, `IndexAccessor` and its associates. For this to work a `NativeSchemaNumberIndexProvider` is also implemented here.

This PR does not yet introduce this combined provider as a loadable kernel extension so no migration or change in default provider is included right now.


community/index/src/main/java/org/neo4j/index/internal/gbptree/Layout.java
```diff
@@ -161,16 +163,34 @@ static long namedIdentifier( String name, int checksum )
...
 
     /**
+     * Typically, layout is compatible with given identifier, major and minor version if
+     * <ul>
+     *     <li>{@code layoutIdentifier == this.identifier()}</li>
+     *     <li>{@code majorVersion == this.majorVersion()}</li>
+     *     <li>{@code minorVersion == this.minorVersion()}</li>
+     * </ul>
+     * <p>
+     * Edge case: When only want to {@link GBPTree#readHeader(PageCache, File, Layout, Header.Reader)} read header} of index and not
```
MishaDemianenko: not really clear what this edge case describing, why they are equal, what is the contract of those qualities, should all layouts be compatible with those requirements?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedSchemaIndexProvider.java
```diff
@@ -0,0 +1,135 @@
...
+
+    static <T> T select( T boost, T fallback, Value... values )
+    {
+        if ( values.length > 1 )
+        {
+            // Multiple values must be handled by fallback
+            return fallback;
+        }
+
+        Value singleValue = values[0];
```
MishaDemianenko: in case if this single value is also part of composite indexes definition and now only one property is updated we should take that into account when we select proper index implementation

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedIndexPopulator.java
```diff
@@ -0,0 +1,142 @@
...
+        finally
+        {
+            fallbackPopulator.markAsFailed( failure );
+        }
+    }
+
+    @Override
+    public void includeSample( IndexEntryUpdate update )
+    {
+        boostPopulator.includeSample( update );
```
MishaDemianenko: Should we also propagate update only to correct populator right? otherwise, it will be doubled in the end?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedSchemaIndexProvider.java
```diff
@@ -0,0 +1,135 @@
...
+        if ( singleValue.valueGroup() == ValueGroup.NUMBER )
+        {
+            // It's a number, the boost can handle this
+            return boost;
+        }
+        return fallback;
+    }
+
+    static IndexSample combineSamples( IndexSample first, IndexSample other )
```
MishaDemianenko: would move that to CombineIndexSampler instead

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedSchemaIndexProvider.java
```diff
@@ -0,0 +1,135 @@
...
+import org.neo4j.kernel.impl.api.index.sampling.IndexSamplingConfig;
+import org.neo4j.kernel.impl.storemigration.StoreMigrationParticipant;
+import org.neo4j.storageengine.api.schema.IndexSample;
+import org.neo4j.values.storable.Value;
+import org.neo4j.values.storable.ValueGroup;
+
+import static java.lang.String.format;
+
+/**
+ * Exactly what this provider will be called and where it will live is undecided, so for the time being
```
MishaDemianenko: this part of javadoc is not that useful and should go

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedSchemaIndexProvider.java
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * The idea is to have a boosted provider which can handle some type of data and is faster, where all other
+ * data is managed by the fallback provider.
+ */
+public class CombinedSchemaIndexProvider extends SchemaIndexProvider
+{
+    private final SchemaIndexProvider boostProvider;
+    private final SchemaIndexProvider fallbackProvider;
+
+    public CombinedSchemaIndexProvider( SchemaIndexProvider boostProvider, SchemaIndexProvider fallbackProvider )
```
MishaDemianenko: I think already here as soon as we know that index is combined based on provided descriptor we should just return lucene index implementations

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexHeaderReader.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    public void read( ByteBuffer headerData )
+    {
+        state = headerData.get();
+        if ( state == BYTE_FAILED )
+        {
+            short messageLength = headerData.getShort();
+            byte[] failureMessageBytes = new byte[messageLength];
+            headerData.get( failureMessageBytes );
+            failureMessage = new String( failureMessageBytes, Charset.forName( "UTF-8" ) );
```
MishaDemianenko: StandardCharsets.UTF_8

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexProvider.java
```diff
@@ -0,0 +1,181 @@
...
+        return new NativeSchemaNumberIndexAccessor<>( pageCache, storeFile, layout, recoveryCleanupWorkCollector );
+    }
+
+    @Override
+    public String getPopulationFailure( long indexId ) throws IllegalStateException
+    {
+        try
+        {
+            String failureMessage = readPopulationFailure( indexId );
+            if ( failureMessage == null )
```
MishaDemianenko, line 113: do we really need to do that? maybe empty string instead?

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/combined/CombinedIndexPopulatorTest.java
```diff
@@ -0,0 +1,379 @@
...
+
+        // then
+        verifyCallFail( failure, () ->
+        {
+            combinedIndexPopulator.markAsFailed( anyString() );
+            return null;
+        } );
+    }
+
+    // includeSample
```
MishaDemianenko: looks like some parts are still missing

community/values/src/main/java/org/neo4j/values/storable/ArrayValue.java
```diff
@@ -24,7 +24,7 @@
 /**
  * Array of one of the storable primitives
  */
-abstract class ArrayValue extends Value implements SequenceValue
+public abstract class ArrayValue extends Value implements SequenceValue
```
MishaDemianenko: not sure why this became public?

MishaDemianenko (changes requested):
Looks good in general, have only a couple of comments mainly because of composite indexes.
Not sure i really like boost/fallback naming convention i would prefer if those things would be called native and lucene instead to highlight what is actually expected to be there

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexProvider.java
```diff
@@ -115,14 +121,35 @@ public String getPopulationFailure( long indexId ) throws IllegalStateException
...
+                return InternalIndexState.ONLINE;
+            case BYTE_POPULATING:
+                return InternalIndexState.POPULATING;
+            default:
+                throw new IllegalStateException( "Unexpected initial state byte value " + headerReader.state );
+            }
+        }
+        catch ( IOException e )
+        {
+            return InternalIndexState.POPULATING;
```
burqen: `LuceneSchemaIndexProvider` also do `log.error( "Failed to open index:" + indexId + ", requesting re-population.", e );`. Maybe native should as well?

Also, missing test for this behaviour.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedSchemaIndexProvider.java
tinwelint: Index updates always arrive to the index "complete", so even if there's a node N with label L and property keys A and B. Let's also say that the index for L and A and B exists and N has values in A and B... updating just A or B will still produce an update to the index which consists of both A and B. So this is a case we don't need to worry about.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedSchemaIndexProvider.java
burqen: There are a few alternatives:

1. Lucene and Combined providers living side by side.
What provider is used when an index is created will then look something like:
```
if ( compositeIndex ) { return LuceneProvider }
if ( configuredToOnlyUseLucene ) { return LuceneProvider }
return CombinedProvider
```
In this way, the Combined index will only be used when there is actually a chance that it will see updates for Native part.

2. We only have Combined provider (like what this PR assumes)
Then it will need to handle the case for composite keys as well (by sending them all to Lucene).

What route we will take is still undecided and the this code will adapt when a decision is taken.

Did this make it more clear?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedSchemaIndexProvider.java
burqen: We just talked about this but just want to type it out for completeness of this PR.

All index updates come as complete `IndexEntryUpdate` objects which contain the whole index entry that has been updated, not only the single part that is actually changed. So this should not be a problem.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexProvider.java
burqen: We chose this way to align with how `LuceneSchemaIndexProvider` does it.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedSchemaIndexProvider.java
tinwelint: There's still benefit in creating an index as "composite lucene native" instead of lucene, and that is that the native provider will over time be able to handle more and more value types so having all indexes as composite will see "automatic" benefits over time, but lucene will not.

tinwelint: If we have a configuration option and it's set to only use lucene then I don't think that the composite provider should be even involved, do you?

tinwelint: If we have a configuration option and it's set to only use lucene then I don't think that the composite provider should be even involved, do you?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedIndexPopulator.java
tinwelint: Yes, I think you're right about that.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexProvider.java
tinwelint: Yup, may be a good idea

tinwelint:
@MishaDemianenko all comments should be addressed now

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedSchemaIndexProvider.java
MishaDemianenko: i think it still makes sense to have a FusionSchemaIndexProvider but as soon as we know  based on index descriptor that index is combined we should just return pure lucene implementations of accessor's, populators etc since there is not a lot of value in wrapping those things right?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/combined/CombinedIndexPopulator.java
MishaDemianenko: this one is still there and include update in both samples

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/combined/CombinedIndexPopulatorTest.java
MishaDemianenko: those things are still not there

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexAccessor.java
```diff
@@ -0,0 +1,149 @@
...
+    {
+        return concatResourceIterators(
+                asList( nativeAccessor.snapshotFiles(), luceneAccessor.snapshotFiles() ).iterator() );
+    }
+
+    @Override
+    public void verifyDeferredConstraints( PropertyAccessor propertyAccessor )
+            throws IndexEntryConflictException, IOException
+    {
+        nativeAccessor.verifyDeferredConstraints( propertyAccessor );
```
MishaDemianenko, line 146: just a note for question, not sure it actually will work as it is now for native index

tinwelint: It works fine, because for native index is a no-op

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexPopulator.java
```diff
@@ -0,0 +1,144 @@
...
+        finally
+        {
+            lucenePopulator.markAsFailed( failure );
+        }
+    }
+
+    @Override
+    public void includeSample( IndexEntryUpdate update )
+    {
+        nativePopulator.includeSample( update );
```
MishaDemianenko: sample still included into both populators?

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexPopulatorTest.java
```diff
@@ -0,0 +1,381 @@
...
+
+        // then
+        verifyCallFail( failure, () ->
+        {
+            fusionIndexPopulator.markAsFailed( anyString() );
+            return null;
+        } );
+    }
+
+    // includeSample
```
MishaDemianenko: missing tests still

MishaDemianenko (approved):

tinwelint:
@MishaDemianenko updated

PR #9746, opened 2017-07-31
Migrate to native label index during store upgrade

MishaDemianenko:
Add additional migrator participant that will create and populate
native label index as part of store upgrade if it does not exist yet.
Migrator is fully integrated into upgrade procedure: it's separate migration step
and support population progress reporting.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/NativeLabelScanStoreMigrator.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/NativeLabelScanStoreMigrator.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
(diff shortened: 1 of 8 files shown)


MishaDemianenko:
<b>This PR does not force migration from 3.2 to 3.3 as it is now</b>

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/NativeLabelScanStoreMigrator.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public void moveMigratedFiles( File migrationDir, File storeDir, String versionToUpgradeFrom,
+            String versionToMigrateTo ) throws IOException
+    {
+        if ( nativeLabelScanStoreMigrated )
+        {
+            File nativeLabelIndex = new File( migrationDir, NativeLabelScanStore.FILE_NAME );
+            fileSystem.moveToDirectory( nativeLabelIndex, storeDir );
```
chrisvest: This won't work on blockdevice. Take a look at how the `StoreMigrator` does it.

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/NativeLabelScanStoreMigrator.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private StoreFactory getStoreFactory( File storeDir, String versionToMigrateFrom )
+    {
+        return new StoreFactory( storeDir, pageCache, fileSystem,
+                        selectForVersion( versionToMigrateFrom ), NullLogProvider.getInstance() );
+    }
+
+    private boolean migrateNativeLabelScanStore( File storeDir )
+    {
+        return !fileSystem.fileExists( new File( storeDir, NativeLabelScanStore.FILE_NAME ) );
```
chrisvest: This one is also problematic on blockdevice.

chrisvest (changes requested):
This would break store migration on blockdevice, and also a relevant integration test is failing.

MishaDemianenko:
@chrisvest updated

chrisvest (approved):

PR #9748, opened 2017-07-31
Appends tree information on exceptions leaving surface

tinwelint:
There are many places where exceptions can be thrown from inside internal
code related to GBPTree and its internals. However the surface of GBPTree
is very small and so the approach is to not pass in tree information into
all internal classes and static methods, but instead decorate exceptions
from the small number of places where these exceptions can leave the
surface. This approach is much less intrusive than the alternative.

This is how it was already, but not the entire surface was covered.
Now also SeekCursor decorates outgoing exceptions.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
@@ -46,6 +46,7 @@ import static java.lang.String.format;
 import static org.neo4j.index.internal.gbptree.Generation.generation;
 import static org.neo4j.index.internal.gbptree.Generation.stableGeneration;
 import static org.neo4j.index.internal.gbptree.Generation.unstableGeneration;
+import static org.neo4j.index.internal.gbptree.GenerationSafePointer.MIN_GENERATION;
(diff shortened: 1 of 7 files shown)


community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -685,10 +697,10 @@ private PagedFile mapWithCorrectPageSize( PageCache pageCache, File indexFile, P
...
             if ( pageSize > pageCache.pageSize() )
             {
-                throw new MetadataMismatchException( "Tree in " + indexFile.getAbsolutePath() +
-                        " was created with page size:" + pageSize +
-                        ", but page cache used to open it this time has a smaller page size:" +
-                        pageCache.pageSize() + " so cannot be opened" );
+                throw new MetadataMismatchException(
+                        " was created with page size %d, but page cache used to open it this time " +
+                        "has a smaller page size %d so cannot be opened",
+                        pageSize, pageCache.pageSize() );
```
burqen: This message could use a sentence-start.

burqen (changes requested):
Found one malformed exception message, looks good otherwise.

tinwelint:
@burqen updated

burqen (approved):

PR #9752, opened 2017-08-01
Close statement in node relationships iterable's

MishaDemianenko:
According to Node interface getRelationships* methods should return
relationship iterable's as result. In the same time, NodeProxy methods
always return ResourceIterables for corresponding calls.
As result, acquired statements will never be never released
and that will cause all sort of side effects with obsolete lucene index readers
that will be cached till the end of the transaction.
With changes introduced in this PR, the statement will be closed but
only in case if iterator of provided iterable was exhausted.


MishaDemianenko:
@tinwelint updated

tinwelint (approved):

PR #9753, opened 2017-08-01
Add Register Interpreted runtime and make it default in enterprise

eebus:


community/cypher/cypher/src/main/scala/org/neo4j/cypher/CypherRuntime.scala
```diff
@@ -25,7 +25,8 @@ case object CypherRuntime extends CypherOptionCompanion[CypherRuntime] {
 
   case object default extends CypherRuntime("default")
   case object interpreted extends CypherRuntime("interpreted")
+  case object registerInterpreted extends CypherRuntime("enterprise-interpreted")
```
craigtaverner, line 28: Should the name 'register' be here? Perhaps while the public name is 'enterprise-interpreted', the internal name is 'sleipnir'?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/PreParserOption.scala
```diff
@@ -34,6 +34,7 @@ case object GreedyPlannerOption extends PlannerPreParserOption("greedy")
 case object IDPPlannerOption extends PlannerPreParserOption("idp")
 case object DPPlannerOption extends PlannerPreParserOption("dp")
 case object InterpretedRuntimeOption extends RuntimePreParserOption("interpreted")
+case object RegisterInterpretedRuntimeOption extends RuntimePreParserOption("enterprise-interpreted")
```
craigtaverner, line 37: Should this be called 'Register'?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
```diff
@@ -405,7 +405,7 @@ return p""")
     graph.createIndex("User", "email")
 
     // when
-    val result = testWith(Configs.CommunityInterpreted, "MATCH (n:User) USING INDEX n:User(email) WHERE exists(n.email) RETURN n")
+    val result = testWith(Configs.All - Configs.Compiled, "MATCH (n:User) USING INDEX n:User(email) WHERE exists(n.email) RETURN n")
```
craigtaverner: Seems the same as Configs.Interpreted. That would be cleaner.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/BuildEnterpriseInterpretedExecutionPlan.scala
```diff
@@ -47,29 +48,33 @@ object BuildEnterpriseInterpretedExecutionPlan extends Phase[EnterpriseRuntimeCo
...
+      val pipeInfo = executionPlanBuilder.build(from.periodicCommit, logicalPlan, idMap)(pipeBuildContext, context.planContext)
+      val PipeInfo(pipe: Pipe, updating, periodicCommitInfo, fp, planner) = pipeInfo
+      val columns = from.statement().returnColumns
+      val resultBuilderFactory = DefaultExecutionResultBuilderFactory(pipeInfo, columns, logicalPlan, idMap)
+      val func = BuildInterpretedExecutionPlan.getExecutionPlanFunction(periodicCommitInfo, from.queryText, updating, resultBuilderFactory, context.notificationLogger)
+      val fingerprint = context.createFingerprintReference(fp)
+      val periodicCommig = periodicCommitInfo.isDefined
+      val indexes = logicalPlan.indexUsage
+      val execPlan = RegisteredExecutionPlan(fingerprint, periodicCommig, planner, indexes, func, pipe, context.config)
+      new CompilationState(from, Some(execPlan))
```
craigtaverner: We seem to be missing a call to the runtimeSuccessRateMonitor for success.

    runtimeSuccessRateMonitor.newPlanSeen(from.logicalPlan)

craigtaverner (changes requested):

craigtaverner (approved):

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/CypherPreParser.scala
```diff
@@ -52,28 +52,29 @@ case object CypherPreParser extends Parser with Base {
     | option("planner", "dp") ~ push(DPPlannerOption)
   )
 
-  def RuntimeOption = rule("runtime option")(
+  def RuntimeOption: Rule1[RuntimePreParserOption with Product with Serializable] = rule("runtime option")(
```
systay: `with Product with Serializable` should not be there

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/CypherPreParser.scala
```diff
@@ -52,28 +52,29 @@ case object CypherPreParser extends Parser with Base {
...
     option("runtime", "interpreted") ~ push(InterpretedRuntimeOption)
       | option("runtime", "compiled") ~ push(CompiledRuntimeOption)
+      | option("runtime", "enterprise-interpreted") ~ push(RegisterInterpretedRuntimeOption)
   )
 
-  def StrategyOption = rule("strategy option")(
+  def StrategyOption: Rule1[EagerOption.type] = rule("strategy option")(
```
systay: these `.type` should probably not be there either

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/EnterpriseRuntimeBuilder.scala
```diff
@@ -29,15 +29,25 @@ import org.neo4j.cypher.internal.frontend.v3_3.phases.{Do, If, Transformer}
...
           If[EnterpriseRuntimeContext, LogicalPlanState, CompilationState](_.maybeExecutionPlan.isEmpty) {
-            BuildEnterpriseInterpretedExecutionPlan
+            pickInterpretedExecutionPlan()
           }
 
       case Some(InterpretedRuntimeName) =>
-        BuildEnterpriseInterpretedExecutionPlan
+        BuildInterpretedExecutionPlan
+
+      case Some(EnterpriseInterpretedRuntimeName) =>
```
systay: When the user has specified enterprise runtime, we should either warn or fail if that is not possible to achieve

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/RegisteredPipeBuilder.scala
```diff
@@ -114,6 +114,12 @@ class RegisteredPipeBuilder(fallback: PipeBuilder,
       case _: Selection =>  // selection relies on inner expressions to interact with variables
         fallback.build(plan, source)
 
+      case _: OptionalExpand =>
```
systay: OptionalExpandPipe writes to execution contexts and should not be handled by the fallback.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisteredRewriter.scala
```diff
@@ -126,6 +136,18 @@ class RegisteredRewriter(tokenContext: TokenContext) {
           case _ =>
             throw new InternalException("Did not find `" + k + "` in the pipeline information")
         }
+
+      case idFunction@FunctionInvocation(_, FunctionName("id"), _, _) =>
+        idFunction
+
+      case _: FunctionInvocation =>
```
systay, line 143: As far as I know, most functions do not read or write to execution contexts, and should be fine to run in the interpreted runtime

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/RegisteredPipeBuilder.scala
```diff
@@ -141,7 +155,8 @@ class RegisteredPipeBuilder(fallback: PipeBuilder,
 
     plan match {
       case Apply(_,_) => ApplyRegisterPipe(lhs, rhs)(id)
-      case _ => fallback.build(plan, lhs, rhs)
+      case CartesianProduct(_, _) => fallback.build(plan, lhs, rhs)
```
systay, line 182: CartesianProductPipe does funky stuff with execution contexts. Are you sure this should be handled by the fallback and not just fail?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/expressions/GetDegreePrimitive.scala
```diff
@@ -0,0 +1,42 @@
...
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.commands.expressions.Expression
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.pipes.QueryState
+import org.neo4j.cypher.internal.frontend.v3_3.SemanticDirection
+
+case class GetDegreePrimitive(offset: Int, typ: Option[String], direction: SemanticDirection) extends Expression {
+
+  override def apply(ctx: ExecutionContext)(implicit state: QueryState): Any = typ match {
+    case None => state.query.nodeGetDegree(ctx.getLongAt(offset), direction)
+    case Some(t) => state.query.getOptRelTypeId(t) match {
+      case None => 0
```
systay: instead of this second match, maybe use 
```
map(relTypeId => state.query.nodeGetDegree(ctx.getLongAt(offset), direction, relTypeId))
.getOrElse(0)
```

I don't think there is any difference in the produced bytecode, and it feels more readable IMHO.

systay (changes requested):

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -180,7 +180,7 @@
     @Internal
     public static final Setting<String> cypher_runtime = setting(
             "unsupported.cypher.runtime",
-            options( "INTERPRETED", "COMPILED", DEFAULT ), DEFAULT );
+            options( "INTERPRETED", "COMPILED", "ENTERPRISE-INTERPRETED" ,  DEFAULT ), DEFAULT );
```
systay, line 183: I still think that this is a really bad name, and should not make it out to users. Can't we change it to the "turbo-runtime" until we have something better? We want to be able to move this stuff out of enterprise, and calling it "enterprise-interpreted" makes that really hard.

systay (approved):

PR #9760, opened 2017-08-02
Importer option to disallow cache allocation on heap

tinwelint:
Normally, allocating data arrays for caching in the importer tries off-heap,
falls back to on-heap and then on to additional ways of allocating.
This commit introduces a config option in the import tool to disallow
the on-heap alternative. This to have better control over heap usage
for normal working memory.

Configuration option is e.g. `--cache-on-heap=false`

### DIFF
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -235,7 +235,14 @@ public class ImportTool
                 "(advanced) Maximum memory that importer can use for various data structures and caching " +
                 "to improve performance. If left as unspecified (null) it is set to " + DEFAULT_MAX_MEMORY_PERCENT +
                 "% of (free memory on machine - max JVM memory). " +
-                "Values can be plain numbers, like 10000000 or e.g. 20G for 20 gigabyte, or even e.g. 70%." );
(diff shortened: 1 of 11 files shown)


chrisvest (approved):
Looks good.

tinwelint:
@chrisvest What we could do instead of introducing an option is to just disallow HEAP allocation and warn if heap looks unnecessarily big compared to amount of available off-heap memory. WDYT?

MishaDemianenko:
@tinwelint will that require people to change their heap/page cache setting that configured in neo4j configuration specifically for the importer and then back? or imported does not use any of those config settings today?

tinwelint:
@MishaDemianenko The import-tool (neo4j-import) is its own script, with `HEAP_SIZE` or even other options for setting heap size. Import tool currently does not consider page cache setting. I believe `neo4j-admin import` has its own means of setting heap size too, but not sure.

MishaDemianenko:
@tinwelint yeah, the `HEAP_SIZE` thingy thx! In case if we will disallow heap usage now for importer that will make that option mostly useless for importer and that would mean that in all cases where its configured/specified will not gonna make sense anymore and will make it worse now which is not a nice thing to do in a patch release

tinwelint:
@MishaDemianenko There are still other things that must make use of the heap, e.g. read input collision ids which come from actual duplicate input ids, but also from accidental collisions in the encoded input ids into longs. If/when we get rid of that from heap too it may make more sense to rethink the heap setting.

PR #9762, opened 2017-08-02
Reduces peak memory required by importer by 40%

tinwelint:
by separating import logic so that IdMapper and NodeRelationshipCache aren't
used at the same time. For this to work one additional stage is introduced
before relationship linking that goes through the imported relationship records
and counts node degrees.

Typically this separation reduces the peak memory usage by 40%.

main author: @SRaghuram

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/CalculateDenseNodesStep.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/CalculateDenseNodesStep.java
@@ -21,60 +21,43 @@ package org.neo4j.unsafe.impl.batchimport;
 
 import org.neo4j.kernel.impl.store.record.RelationshipRecord;
 import org.neo4j.unsafe.impl.batchimport.cache.NodeRelationshipCache;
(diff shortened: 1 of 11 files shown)


chrisvest:
What performance impact does this have?

tinwelint:
The additional time spent is in the node degree stage, which is bound by number of relationships. It's a stage reading from relationship store (in parallel) and incrementing numbers in a cache array, twice for each relationship (start/end node), also in parallel if need be. In my simple measurements the added stage typically performs like the relationship count stage in the end performs, which is roughly 10x faster than relationship import stage, or faster. The node/rel import stands for ~50% of the total time, relationship import being typically more dominant than node import. So ~40% of the time got 10% longer.

This means an **increase in time spent ~5%** if my math serves me well here. It correlates with my total time measurements. More data points should be collected though.

tinwelint:
Failed on unrelated flaky test `ListQueriesProcedureTest.shouldProvideAllocatedBytes`

tinwelint:
As a data point, other measurements where number of relationships is heavily dominating number of nodes, two orders of magnitude, this penalty is closer to 10%.

burqen (approved):
Seems like a worthwhile tradeoff! 🎉

PR #9764, opened 2017-08-02
Procedures now close acquired KernelStatements

burqen:
Verified in test by performing two index reads with
intermediate create. Second read should see result
of create but will not if KernelStatement has leaked,
because Lucene reader from first match will be reused
for second read.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
@@ -112,34 +112,36 @@ public class SchemaProcedure
                         RelationshipType relationshipType = relationshipTypeIterator.next();
                         String relationshipTypeGetName = relationshipType.name();
                         int relId = readOperations.relationshipTypeGetForName( relationshipTypeGetName );
-                        ResourceIterator<Label> labelsInUse = graphDatabaseAPI.getAllLabelsInUse().iterator();
 
-                        List<NodeImpl> startNodes = new LinkedList<>();
-                        List<NodeImpl> endNodes = new LinkedList<>();
-
-                        while ( labelsInUse.hasNext() )
+                        try ( ResourceIterator<Label> labelsInUse = graphDatabaseAPI.getAllLabelsInUse().iterator() )
                         {
-                            Label labelToken = labelsInUse.next();
-                            String labelName = labelToken.name();
-                            Map<String,Object> properties = new HashMap<>();
-                            NodeImpl node = getOrCreateLabel( labelName, properties, nodes );
-                            int labelId = readOperations.labelGetForName( labelName );
+                            List<NodeImpl> startNodes = new LinkedList<>();
+                            List<NodeImpl> endNodes = new LinkedList<>();
 
-                            if ( readOperations.countsForRelationship( labelId, relId, ReadOperations.ANY_LABEL ) > 0 )
+                            while ( labelsInUse.hasNext() )
                             {
-                                startNodes.add( node );
+                                Label labelToken = labelsInUse.next();
+                                String labelName = labelToken.name();
+                                Map<String,Object> properties = new HashMap<>();
+                                NodeImpl node = getOrCreateLabel( labelName, properties, nodes );
+                                int labelId = readOperations.labelGetForName( labelName );
+
+                                if ( readOperations.countsForRelationship( labelId, relId, ReadOperations.ANY_LABEL ) > 0 )
+                                {
+                                    startNodes.add( node );
+                                }
+                                if ( readOperations.countsForRelationship( ReadOperations.ANY_LABEL, relId, labelId ) > 0 )
+                                {
+                                    endNodes.add( node );
+                                }
                             }
-                            if ( readOperations.countsForRelationship( ReadOperations.ANY_LABEL, relId, labelId ) > 0 )
+                            for ( NodeImpl startNode : startNodes )
                             {
-                                endNodes.add( node );
-                            }
-                        }
-                        for ( NodeImpl startNode : startNodes )
-                        {
-                            for ( NodeImpl endNode : endNodes )
-                            {
-                                RelationshipImpl relationship =
-                                        addRelationship( startNode, endNode, relationshipTypeGetName, relationships );
+                                for ( NodeImpl endNode : endNodes )
+                                {
+                                    RelationshipImpl relationship =
+                                            addRelationship( startNode, endNode, relationshipTypeGetName, relationships );
+                                }
                             }
                         }
                     }
--- a/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
@@ -40,6 +40,7 @@ import org.neo4j.graphdb.ResourceIterator;
 import org.neo4j.graphdb.Transaction;
 import org.neo4j.kernel.api.KernelTransaction;
 import org.neo4j.kernel.api.ReadOperations;
+import org.neo4j.kernel.api.Statement;
 import org.neo4j.kernel.api.StatementTokenNameLookup;
 import org.neo4j.kernel.api.constraints.NodePropertyConstraint;
 import org.neo4j.kernel.api.index.IndexDescriptor;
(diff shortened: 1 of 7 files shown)


burqen:
Merge commit to 3.2 here https://github.com/burqen/neo4j/tree/3.2-procedures-close-kernel-statement-merge-from-3.1

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -62,82 +63,90 @@ public GraphResult buildSchemaGraph()
...
-            }
-
-            //add all relationships
+            ReadOperations readOperations = statement.readOperations();
+            StatementTokenNameLookup statementTokenNameLookup = new StatementTokenNameLookup( readOperations );
 
-            Iterator<RelationshipType> relationshipTypeIterator =
-                    graphDatabaseAPI.getAllRelationshipTypesInUse().iterator();
-            while ( relationshipTypeIterator.hasNext() )
+            try ( Transaction transaction = graphDatabaseAPI.beginTx(); )
```
craigtaverner, line 71: Do we need this inner transaction? We are already in a transaction, so this will be a dummy.

burqen: I would guess not, but cleaning this up is outside the scope of this PR.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -62,82 +63,90 @@ public GraphResult buildSchemaGraph()
...
+                        .iterator() )
                 {
-                    for ( NodeImpl endNode : endNodes )
+                    while ( relationshipTypeIterator.hasNext() )
                     {
-                        RelationshipImpl relationship = addRelationship( startNode, endNode, relationshipTypeGetName, relationships );
+                        RelationshipType relationshipType = relationshipTypeIterator.next();
+                        String relationshipTypeGetName = relationshipType.name();
+                        int relId = readOperations.relationshipTypeGetForName( relationshipTypeGetName );
+                        ResourceIterator<Label> labelsInUse = graphDatabaseAPI.getAllLabelsInUse().iterator();
```
MishaDemianenko: this resource iterator should be closed as well

enterprise/neo4j-enterprise/src/test/java/org/neo4j/kernel/builtinprocs/ProcedureResourcesIT.java
```diff
@@ -0,0 +1,211 @@
...
+    private void createIndex()
+    {
+        try ( Transaction tx = db.beginTx() )
+        {
+            db.execute( "CREATE INDEX ON " + indexDefinition );
+            tx.success();
+        }
+        try ( Transaction tx = db.beginTx() )
+        {
+            db.schema().awaitIndexesOnline( 1, TimeUnit.SECONDS );
```
MishaDemianenko: this can be not enough on latest 3.3 version

enterprise/neo4j-enterprise/src/test/java/org/neo4j/kernel/builtinprocs/ProcedureResourcesIT.java
```diff
@@ -0,0 +1,211 @@
...
+            tx.success();
+        }
+    }
+
+    private String buildProcedureQuery( String procedureName, List<Object> parameters )
+    {
+        StringBuilder sb = new StringBuilder();
```
MishaDemianenko: String joiner can be used here to simplify looping and checks, not that important in the same time

enterprise/neo4j-enterprise/src/test/java/org/neo4j/kernel/builtinprocs/ProcedureResourcesIT.java
```diff
@@ -0,0 +1,211 @@
...
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.kernel.api.proc.ProcedureSignature;
+import org.neo4j.kernel.impl.proc.Procedures;
+import org.neo4j.test.rule.DatabaseRule;
+import org.neo4j.test.rule.EnterpriseDatabaseRule;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class ProcedureResourcesIT
```
MishaDemianenko, line 51: different versions will contain different procedures so this test will need some adaptations during merges be careful there 👀

MishaDemianenko, line 51: its' also strange that this test does not catch a case where we still have one statement that is not closed because of not closed resource iterable

MishaDemianenko (changes requested):
Looks good, a couple of small comments. While this change cleaned up a lot of incorrect statements usages here and there I think we need more strict eyes on those. We need *the tracker* that will track open statements in the transaction and will fail it in the case when it will be closed without statements being closed across all our tests at least, sounds like a good idea?

burqen:
For completeness. Here is an attempt at tracking open statements: https://github.com/neo4j/neo4j/pull/9794

enterprise/neo4j-enterprise/src/test/java/org/neo4j/kernel/builtinprocs/ProcedureResourcesIT.java
burqen: Problem was there was no relationships in the dataset which meant that code path was never hit. Fixed now.

enterprise/neo4j-enterprise/src/test/java/org/neo4j/kernel/builtinprocs/ProcedureResourcesIT.java
burqen: Hmm... set to 5 just in case then.

burqen:
@MishaDemianenko Comments addressed. Will take care to adapt test during forward merge!

enterprise/neo4j-enterprise/src/test/java/org/neo4j/kernel/builtinprocs/ProcedureResourcesIT.java
MishaDemianenko: that is what i thinking about:
```
        StringJoiner stringJoiner = new StringJoiner( ",", "CALL " + procedureName + "(", ")" );
        for ( Object parameter : parameters )
        {
            stringJoiner.add( parameter.toString() );
        }
        return stringJoiner.toString();
```

enterprise/neo4j-enterprise/src/test/java/org/neo4j/kernel/builtinprocs/ProcedureResourcesIT.java
```diff
@@ -0,0 +1,229 @@
...
+        StringBuilder sb = new StringBuilder();
+        StringJoiner joiner = new StringJoiner( "," );
+        sb.append( "CALL " ).append( procedureName ).append( "(" );
+        for ( Object parameter : parameters )
+        {
+            joiner.add( parameter.toString() );
+        }
+        sb.append( joiner.toString() );
+        sb.append( ")" );
+        System.out.println( sb.toString() );
```
MishaDemianenko: in any case you do not need to print this string

MishaDemianenko (approved):
@burqen 2 small comments otherwise looks good will approve now but please take a look on them

burqen:
@MishaDemianenko Ping 👍

MishaDemianenko:
@burqen Pong, whats up?)

enterprise/neo4j-enterprise/src/test/java/org/neo4j/kernel/builtinprocs/ProcedureResourcesIT.java
burqen: Aaah, cool!

enterprise/neo4j-enterprise/src/test/java/org/neo4j/kernel/builtinprocs/ProcedureResourcesIT.java
burqen: \*facepalm\*

MishaDemianenko:
👍

PR #9770, opened 2017-08-03
Added CreateNode and EmptyResult for enterprise interpreted

SaschaPeukert:


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CreateAcceptanceTest.scala
```diff
@@ -31,7 +31,7 @@ class CreateAcceptanceTest extends ExecutionEngineFunSuite with QueryStatisticsT
 
     val query = s"CREATE (a) WITH a LOAD CSV FROM '$url' AS line CREATE (b) CREATE (a)<-[:T]-(b)"
 
-    val result = updateWithBothPlannersAndCompatibilityMode(query)
+    val result = testWithUpdate(Configs.All - Configs.Compiled - Configs.EnterpriseInterpreted - Configs.Cost2_3, query)
```
systay: We need to find a better Configs name for this.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/LernaeanTestSupport.scala
```diff
@@ -67,6 +68,7 @@ trait LernaeanTestSupport extends CypherTestSupport {
           Eagerly.immutableMapValues(m.asScala, convert)
         case l: java.util.List[_] => l.asScala.map(convert)
         case d: java.lang.Double if replaceNaNs && java.lang.Double.isNaN(d) => NanReplacement
+        case n: NodeProxy => new NodeProxy(null,0)
```
systay: What is this change for?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/CreateNodeRegisterPipe.scala
```diff
@@ -0,0 +1,88 @@
...
+import scala.collection.Map
+
+case class CreateNodeRegisterPipe(ident: String, pipelineInformation: PipelineInformation,
+                                  labels: Seq[LazyLabel], properties: Option[Expression])
+                                   (val id: Id = new Id) extends Pipe {
+
+  private val offset = pipelineInformation.getLongOffsetFor(ident)
+
+  protected def internalCreateResults(state: QueryState): Iterator[ExecutionContext] = {
+    val node: Node = state.query.createNode()
```
systay: We should not create NodeProxy objects unless we must. Here, we need to change the QueryContext trait to not return a NodeProxy - it should just return a long.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/EmptyResultRegisterPipe.scala
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher.internal.compatibility.v3_3.runtime.interpreted.pipes
+
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.ExecutionContext
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.pipes.{Pipe, PipeWithSource, QueryState}
+import org.neo4j.cypher.internal.compiler.v3_3.planDescription.Id
+
+case class EmptyResultRegisterPipe(source: Pipe)(val id: Id = new Id) extends PipeWithSource(source) {
```
systay: We only need to write `Register` specific versions of `Pipe`s iff they have direct interaction with `ExecutionContext`s - reading from them, writing to them or creating them. Since `EmptyResultPipe` does none of those things, we don't need this extra class.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/CreateNodeRegisterPipe.scala
```diff
@@ -0,0 +1,88 @@
...
+    setLabels(context, state, nodeId)
+
+    state.copyArgumentStateTo(context)
+    context.setLongAt(offset, nodeId)
+    PrimitiveLongHelper.map(PrimitiveLongCollections.singleton(nodeId), nodeId => context)
+  }
+
+  private def setProperties(context: ExecutionContext, state: QueryState, nodeId: Long) = {
+    properties.foreach { (expr) =>
+      expr(context)(state) match {
```
systay, line 54: Hmm... I'm thinking we might want to specialise this and not make a runtime check. I would bet that a large percentage of usages of CREATE come with a literal map, and so making a specialised version of that, where we have a `Map[String, Expression]` thingie to use here. we should measure, but my gut e-coli tells me it should be a nice improvement.

systay (changes requested):

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/CreateNodeRegisterPipe.scala
craigtaverner: Seems like a great idea, but perhaps done as a second round of improvements, as an optimization to this work.

alexaverbuch: @systay do you mind if that specialization is done as a separate piece of work? i.e., we merge this as is and create a card so we don't forget

community/graphdb-api/src/main/java/org/neo4j/graphdb/GraphDatabaseService.java
```diff
@@ -55,6 +55,16 @@
...
 
+    /**
+     * Creates a new node and returns it id.
+     * Please note: Neo4j reuses its internal ids when
+     * nodes and relationships are deleted, which means it's bad practice to
+     * refer to them this way. Instead, use application generated ids.
+     *
+     * @return the created nodes id.
+     */
+    Long createNodeId();
```
chrisvest, line 66: :scream:

MishaDemianenko: its strange to see more than one person to find out about that method almost simultaneously

PR #9773, opened 2017-08-04
Removes generic and some boxing in RecordAccess

tinwelint:
The key was generified and boxed all the way simply to distinguish between
long/int, which is completely unnecessary. Now it just uses long everywhere
and runs toIntExact in places where an int is expected. Saves memory as
well as cluttery code.

### DIFF
--- a/community/kernel/pom.xml
+++ b/community/kernel/pom.xml
@@ -143,6 +143,9 @@ the relevant Commercial Agreement.
             <include>org/neo4j/procedure/*</include>
             <include>org/neo4j/helpers/collection/*</include>
           </includes>
+          <excludes>
+            <exclude>org/neo4j/unsafe/batchinsert/DirectRecordAccess</exclude>
+          </excludes>
         </configuration>
       </plugin>
     </plugins>
--- a/community/kernel/pom.xml
+++ b/community/kernel/pom.xml
@@ -145,6 +145,7 @@ the relevant Commercial Agreement.
           </includes>
           <excludes>
             <exclude>org/neo4j/unsafe/batchinsert/DirectRecordAccess</exclude>
+            <exclude>org/neo4j/unsafe/batchinsert/DirectRecordAccessSet</exclude>
           </excludes>
         </configuration>
       </plugin>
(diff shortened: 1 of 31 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/RecordChanges.java
```diff
@@ -37,13 +37,13 @@
  * @param <RECORD>
```
klaren: `<KEY>`  is no longer valid in javadoc

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/DirectRecordAccess.java
```diff
@@ -236,7 +234,7 @@ public void commit()
         }
 
         List<DirectRecordProxy> directRecordProxies = new ArrayList<>( batch.values() );
-        Collections.sort(directRecordProxies, ( o1, o2 ) -> -o1.getKey().compareTo( o2.getKey() ) );
+        Collections.sort(directRecordProxies, ( o1, o2 ) -> Long.compare( -o1.getKey(), o2.getKey() ) );
```
klaren: No need to go through `Collections`. Can just call sort on list

`directRecordProxies.sort( ( o1, o2 ) -> Long.compare( -o1.getKey(), o2.getKey() ) );`

community/kernel/pom.xml
```diff
@@ -143,6 +143,9 @@ the relevant Commercial Agreement.
             <include>org/neo4j/procedure/*</include>
             <include>org/neo4j/helpers/collection/*</include>
           </includes>
+          <excludes>
+            <exclude>org/neo4j/unsafe/batchinsert/DirectRecordAccess</exclude>
```
klaren, line 147: We should probably exclude `org.neo4j.unsafe.batchinsert.DirectRecordAccessSet` too while we're at it. Or move them to the `internal` sub-package.

klaren (changes requested):
Well spotted! :+1: 
I only have 3 small comments.

community/kernel/pom.xml
tinwelint: How about marking both AND moving both :)

tinwelint:
@klaren updated

community/kernel/pom.xml
klaren: Even better :)

klaren (approved):

tinwelint:
@klaren  Classes in `internal` doesn't need to be excluded, besides I'm excluding what gets compared to the previous version.

PR #9774, opened 2017-08-04
More efficient log rotation checking

tinwelint:
Merely checking need for log rotation shows up on the radar when profiling,
especially if there are many concurrent committers. When appending there's
a natural selection of one out of all that gets to force the log.
This commit piggy-backs on that decision and lets that thread, and only
that one out of all concurrent committers, to check log rotation.

Also introduced double checked locking for checking need for log rotation
due to a lot of unnecessary hogging of that monitor in most calls,
since log rotation isn't needed very often

These two minor changes improves transaction commit concurrency

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/BatchingTransactionAppender.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/BatchingTransactionAppender.java
@@ -93,11 +93,6 @@ public class BatchingTransactionAppender extends LifecycleAdapter implements Tra
     @Override
     public long append( TransactionToApply batch, LogAppendEvent logAppendEvent ) throws IOException
     {
-        // We put log rotation check outside the private append method since it must happen before
-        // we generate the next transaction id
-        boolean logRotated = logRotation.rotateLogIfNeeded( logAppendEvent );
-        logAppendEvent.setLogRotated( logRotated );
-
         // Assigned base tx id just to make compiler happy
         long lastTransactionId = TransactionIdStore.BASE_TX_ID;
         // Synchronized with logFile to get absolute control over concurrent rotations happening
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/BatchingTransactionAppender.java
```diff
@@ -93,11 +93,6 @@ public void start() throws Throwable
     @Override
     public long append( TransactionToApply batch, LogAppendEvent logAppendEvent ) throws IOException
     {
-        // We put log rotation check outside the private append method since it must happen before
```
klaren, line 96: Can we safely remove this comment and/or intention?

tinwelint: It seems that it wasn't true, right? And the comment doesn't say why. Also it isn't like that anymore, now with the new changes.

klaren: So true, **outside** the private append method is not the same as **before** the private append method, my bad.

klaren (approved):
Nice work!

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/rotation/LogRotationImpl.java
```diff
@@ -49,18 +49,21 @@ public boolean rotateLogIfNeeded( LogAppendEvent logAppendEvent ) throws IOExcep
          * doing force (think batching of writes), such that it can't see a bad state of the writer
          * even when rotating underlying channels.
          */
-        synchronized ( logFile )
+        if ( logFile.rotationNeeded() )
```
chrisvest, line 52: For this double-checked locking to work properly, the `PhysicalLogVersionedStoreChannel.position` field needs to be volatile.

PR #9776, opened 2017-08-04
Heavily reduces contention on generating ids

tinwelint:
The main idea is to have batching id generators local to each transaction instance (which is pooled). This batching sits in between store and typical creators of records at commit time and hands out ids, grabbing batches of ids off of the stores. On clean shutdown any excess ids remaining unused will be freed back into the store before closing.

In scenarios where there are many concurrent transactions committing data, where they all co-operatively force the log then this id generator contention is the single biggest one. With these changes this contention is basically removed entirely.

The default batch size is 20, which reduces contention dramatically and should be a good value for most scenarios. The value is however controllable with a database setting.


community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
```diff
@@ -266,7 +244,15 @@ public RecordStorageEngine(
...
-        return () -> new StoreStatement( neoStores, indexReaderFactory, labelScanStore::newReader, lockService );
+        return new Supplier<StorageStatement>()
+        {
+            @Override
+            public StorageStatement get()
+            {
+                return new StoreStatement( neoStores, indexReaderFactory, labelScanStore::newReader, lockService,
+                        allocateCommandCreationContext() );
+            }
+        };
```
chrisvest: No need to expand the lambda

community/kernel/src/test/java/org/neo4j/kernel/impl/store/id/RenewableBatchIdSequenceTest.java
```diff
@@ -0,0 +1,183 @@
...
+import static java.util.Arrays.asList;
+
+import static org.neo4j.collection.primitive.PrimitiveLongCollections.EMPTY_LONG_ARRAY;
+
+public class RenewableBatchIdSequenceTest
+{
+    public static final int BATCH_SIZE = 5;
+
+    private final IdSource idSource = new IdSource();
+    private final LongConsumer excessIds = mock( LongConsumer.class );
```
chrisvest: To avoid mocking, you could make this a lambda for an `arrayList::add`. You could then inspect the list to verify the interactions.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/id/IdRangeIteratorTest.java
```diff
@@ -71,12 +74,129 @@ public void shouldUseDefragIdsFirst() throws Exception
...
         int rangeLength = 1024;
-        IdRangeIterator iterator = new IdRangeIterator( new IdRange( new long[]{7,8,9}, 1024, rangeLength ) );
+        IdRangeIterator iterator = new IdRange( new long[]{7,8,9}, 1024, rangeLength ).iterator();
+
+        // then
+        assertEquals(7, iterator.nextId());
+        assertEquals(8, iterator.nextId());
+        assertEquals(9, iterator.nextId());
+        assertEquals(1024, iterator.nextId());
+    }
```
chrisvest: Formatting.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageCommandCreationContext.java
```diff
@@ -0,0 +1,95 @@
...
+import org.neo4j.kernel.impl.transaction.state.TransactionRecordState;
+import org.neo4j.storageengine.api.CommandCreationContext;
+import org.neo4j.storageengine.api.lock.ResourceLocker;
+
+/**
+ * Holds commit data structures for creating records in a {@link NeoStores}.
+ */
+public class RecordStorageCommandCreationContext implements CommandCreationContext
+{
+    public static final int DEFAULT_ID_BATCH_SIZE = Integer.parseInt( GraphDatabaseSettings.tx_id_batch_size.getDefaultValue() );
```
chrisvest: Is this field still necessary, given we have a setting? I'd think we should always use the setting.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -467,6 +468,17 @@
...
+    @Description( "Specifies the size of id batches local to each transaction when committing. " +
+            "Committing a transaction which contains changes most often results in new data records being created. " +
+            "For each record a new id needs to be generated from an id generator. " +
+            "It's more efficient to allocate a batch of ids from the contended id generator, which the transaction " +
+            "holds and generates ids from while creating these new records. " +
+            "This setting specifies how big those batches are. " +
+            "Remaining ids are freed back to id generator on clean shutdown." )
+    @Internal
+    public static final Setting<Integer> tx_id_batch_size = setting( "unsupported.dbms.tx_id_batch_size", INTEGER,
+            "20", range( 1, 1_000 ) );
```
chrisvest: Maybe `record_id_allocation_batch_size` is more descriptive, since we are not allocating transaction ids.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/RenewableBatchIdSequences.java
```diff
@@ -0,0 +1,76 @@
...
+                if ( type.isLimitedIdStore() )
+                {
+                    // This is a token store or otherwise meta-data store, so let's not add batching for it
+                    types[type.ordinal()] = store;
+                }
+                else
+                {
+                    // This is a normal record store where id batching is beneficial
+                    LongConsumer idConsumer = id -> store.freeId( id );
+                    types[type.ordinal()] = new RenewableBatchIdSequence( store, batchSize, idConsumer );
```
chrisvest: `idConsumer` can also be defined as `store::freeId`

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/RenewableBatchIdSequences.java
```diff
@@ -58,9 +67,9 @@ public void close()
     {
         for ( StoreType type : StoreType.values() )
         {
-            if ( type.isRecordStore() )
+            if ( type.isRecordStore() && !type.isLimitedIdStore() )
             {
-                idGenerator( type ).close();
+                ((Resource)idGenerator( type )).close();
```
chrisvest: Would it make sense to instead check if the respective IdGenerator implements Resource?

chrisvest (commented):
@tinwelint There's a bunch of test failures still, and I have a handful of  cosmetic comments.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
tinwelint: Probably not

community/kernel/src/test/java/org/neo4j/kernel/impl/store/id/RenewableBatchIdSequenceTest.java
tinwelint: A nice idea

chrisvest:
@tinwelint You have a couple of compilation failures after the setting rename.

tinwelint:
There's currently testing on causal cluster with these changes taking place.

tinwelint:
Replaced by #9961

PR #9778, opened 2017-08-04
CompositeNodeKeyConstraintAcceptanceTest updated to LernaeanTestSupport

SaschaPeukert:
Also added a failWithError method to LernaeanTestSupport

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CompositeNodeKeyConstraintAcceptanceTest.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CompositeNodeKeyConstraintAcceptanceTest.scala
@@ -37,15 +37,13 @@ class CompositeNodeKeyConstraintAcceptanceTest extends ExecutionEngineFunSuite w
 
   test("should be able to create and remove single property NODE KEY") {
     // When
-    succeedWith(Configs.Procs,
-      "CREATE CONSTRAINT ON (n:Person) ASSERT (n.email) IS NODE KEY")
+    graph.execute("CREATE CONSTRAINT ON (n:Person) ASSERT (n.email) IS NODE KEY")
 
     // Then
     graph should haveConstraints("NODE_KEY:Person(email)")
 
     // When
-    succeedWith(Configs.Procs,
-      "DROP CONSTRAINT ON (n:Person) ASSERT (n.email) IS NODE KEY")
+    graph.execute("DROP CONSTRAINT ON (n:Person) ASSERT (n.email) IS NODE KEY")
 
     // Then
     graph should not(haveConstraints("NODE_KEY:Person(email)"))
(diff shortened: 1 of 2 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CompositeNodeKeyConstraintAcceptanceTest.scala
```diff
@@ -131,25 +139,30 @@ class CompositeNodeKeyConstraintAcceptanceTest extends ExecutionEngineFunSuite w
...
-      exec("CREATE CONSTRAINT ON (n:User) ASSERT (n.firstname,n.lastname) IS NODE KEY")
+      testWith(Configs.AbsolutelyAll,
+        "CREATE CONSTRAINT ON (n:User) ASSERT (n.firstname,n.lastname) IS NODE KEY")
     }
   }
 
   test("trying to add duplicate node when node key constraint exists") {
     createLabeledNode(Map("name" -> "A"), "Person")
-    exec("CREATE CONSTRAINT ON (person:Person) ASSERT (person.name) IS NODE KEY")
+    testWith(Configs.Procs,
```
craigtaverner: Should the test setup also be part of what is being tested? I think parts in 'given' should be run simply, and then the 'when/then' parts are actually tested.

SaschaPeukert: I would agree, but the 'given' parts were tested previously as well. Thus I thought there is some reason behind it. But if there is not, I happily change it

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CompositeNodeKeyConstraintAcceptanceTest.scala
```diff
@@ -31,36 +29,41 @@ import org.neo4j.test.TestEnterpriseGraphDatabaseFactory
...
+class CompositeNodeKeyConstraintAcceptanceTest extends ExecutionEngineFunSuite with CypherComparisonSupport {
 
   override protected def createGraphDatabase(config: Map[Setting[_], String] = databaseConfig()): GraphDatabaseCypherService = {
     new GraphDatabaseCypherService(new TestEnterpriseGraphDatabaseFactory().newImpermanentDatabase(config.asJava))
   }
 
   test("should be able to create and remove single property NODE KEY") {
     // When
-    exec("CREATE CONSTRAINT ON (n:Person) ASSERT (n.email) IS NODE KEY")
+    succeedWith(Configs.Procs,
```
craigtaverner: This line is really just a test setup command, not a test itself. However, written like this it is currently only testing that `CREATE CONSTRAINT` commands are disallowed with any runtime or planner or version prefixes, which is certainly out of scope for this test. So we end up running this schema command on all scenarios where all scehma commands will fail, and expect failure. This is very redundant. Let's rather have this test only test what it was meant to test.

craigtaverner (changes requested):
It feels like we might be testing a bunch of Cypher stuff for tests that are not at all about Cypher. Should we be using the new support here at all?

craigtaverner (approved):

PR #9779, opened 2017-08-04
Transaction timeout monitor

MishaDemianenko:
Introduce kernel transaction timeout monitor thread,
that will run in the background and will mark timed out transactions
as terminated in addition to existent execution guard checker.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -240,6 +240,10 @@ public abstract class GraphDatabaseSettings
     @Description("The maximum time interval of a transaction within which it should be completed.")
     public static final Setting<Long> transaction_timeout = setting( "dbms.transaction.timeout", DURATION, String.valueOf( UNSPECIFIED_TIMEOUT ) );
 
+    @Description("Configures the time interval between transaction monitor checks. Determines how often " +
+            "monitor thread will check transaction for timeout.")
+    public static final Setting<Long> transaction_monitor_check_interval = setting( "dbms.transaction.monitor.check.interval", DURATION, "10s" );
+
     @Description( "The maximum amount of time to wait for running transactions to complete before allowing "
                   + "initiated database shutdown to continue" )
     @Internal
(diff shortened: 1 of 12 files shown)


tinwelint:
Perhaps this opens op an opportunity to _only_ do this checking and remove the guarding statement operations "layer"? And increase default rate to 1s or something. Would that be a good idea?

MishaDemianenko:
@tinwelint we can remove only part that actually checks time out but can't remove the part that throwing guarded exception from transaction thread since it's actually that exception that will cause the rollback in most of the cases.

tinwelint:
Yes OK. I could imagine a future where this wouldn't be a specific GuardException, but rather a termination with a timeout classification instead. Would be nice to have one unified way of terminating/aborting a transaction. Of course this would not be part of this PR, but food for thought.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -240,6 +240,10 @@
     @Description("The maximum time interval of a transaction within which it should be completed.")
     public static final Setting<Long> transaction_timeout = setting( "dbms.transaction.timeout", DURATION, String.valueOf( UNSPECIFIED_TIMEOUT ) );
 
+    @Description("Configures the time interval between transaction monitor checks. Determines how often " +
+            "monitor thread will check transaction for timeout.")
+    public static final Setting<Long> transaction_monitor_check_interval = setting( "dbms.transaction.monitor.check.interval", DURATION, "10s" );
```
tinwelint: I don't think it hurts to lower this a little, how about 5s max?

tinwelint (approved):

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
lutovich: Should this setting maybe not allow negative values?
It might also be good to be able to turn this monitor off by setting check interval to zero.

burqen (approved):
Simple, clean, small and efficient! 🎉

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
MishaDemianenko: duration setting type does not allow negative values already,
do not know what would be the real use case for that but why not....just in case something goes really wrong...will not schedule job if 0 is specified then

MishaDemianenko:
@tinwelint default interval updated to be 5s
@lutovich update to skip job scheduling in case if check interval is 0.

PR #9780, opened 2017-08-05
Improving Cypher and BOLT interaction

pontusmelke:
Initially this was supposed to be a small change to improve the interaction between Cypher and Bolt but it sort of grew from there. Sorry about that.

Main idea is that Bolt now gets `Values` from Cypher instead of raw `Objects`, which allows for more efficient serialization without resorting to a big `instance of` switch. 

In order to achieve this the following changes were made:
- Interpreted runtime is now a `Map[String, AnyValue]` instead of a `Map[String, Any]`
- All expressions in the interpreted runtime now returns `AnyValue` instead of `Any`
- The compiled runtime now _boxes_ primitives to `AnyValue` instead of boxing to `Object` before interacting with the visitor.
- Some levels of indirections in the final `Result` has been removed.
- the `values` module has been updated in order to

NOTE: The failing docs build is due to internal changes in this PR. https://github.com/neo4j/neo4j-documentation/pull/113 will addresses these failures, so we should merge these PRs at more or less the same time.


community/bolt/src/test/java/org/neo4j/bolt/v1/transport/integration/AuthenticationIT.java
```diff
@@ -564,4 +566,9 @@ public boolean gotSpecialMessage()
             return specialMessage != null;
         }
     }
+
+    public static Map<String,AnyValue> mapValue( Object... objects )
```
craigtaverner: This method is never used. Perhaps remove it again?

community/bolt/src/test/java/org/neo4j/bolt/v1/transport/integration/TransportSessionIT.java
```diff
@@ -41,6 +41,8 @@
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
 import org.neo4j.helpers.HostnamePort;
 import org.neo4j.kernel.api.exceptions.Status;
+import org.neo4j.values.storable.Values;
```
craigtaverner: Unused imports

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/javacompat/ResultRecord.java
```diff
@@ -17,25 +17,36 @@
...
-    String getString( String key );
-
-    Number getNumber( String key );
-
-    Boolean getBoolean( String key );
-
-    Path getPath( String key );
+    private final AnyValue[] fields;
+
+    public ResultRecord( int size )
```
craigtaverner: Constructor never used.

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/javacompat/ResultRowImpl.java
```diff
@@ -37,7 +37,8 @@ public ResultRowImpl( Map<String, Object> results )
         this.results = results;
     }
 
-    public ResultRowImpl( )
+    public
+    ResultRowImpl( )
```
craigtaverner: Formatting?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v2_3/Compatibility.scala
```diff
@@ -61,25 +61,29 @@ trait Compatibility {
...
+                         preParsingNotifications: Set[org.neo4j.graphdb.Notification]) = {
     import org.neo4j.cypher.internal.compatibility.v2_3.helpers.as2_3
     val notificationLogger = new RecordingNotificationLogger
     val preparedQueryForV_2_3: Try[PreparedQuery] =
       Try(compiler.prepareQuery(preParsedQuery.statement,
-        preParsedQuery.rawStatement,
-        notificationLogger,
-        preParsedQuery.planner.name,
-        Some(as2_3(preParsedQuery.offset)), tracer))
+                                preParsedQuery.rawStatement,
```
craigtaverner, line 71: Most of the classes changes are reformatting. Makes it hard to see the real change, as well as hard to forward merge fixes.

community/values/pom.xml
```diff
@@ -34,36 +35,41 @@
...
+        the relevant Commercial Agreement.
       </comments>
     </license>
   </licenses>
 
   <dependencies>
     <dependency>
+      <groupId>org.neo4j</groupId>
+      <artifactId>neo4j-graphdb-api</artifactId>
+      <version>${project.version}</version>
```
fickludd, line 57: Could we remove this dep. by moving the converters somewhere else. This feels a non-intuitive to me. On the other hand, I guess `neo4j-graphdb-api` is very stable and doesn't drag in much else, so maybe it's fine.

community/values/src/main/java/org/neo4j/values/BaseToObjectValueWriter.java
```diff
@@ -0,0 +1,545 @@
...
+    {
+        assert !stack.isEmpty();
+        writeValue( stack.pop().value() );
+    }
+
+    @Override
+    public void writePath( NodeValue[] nodes, EdgeValue[] edges ) throws RuntimeException
+    {
+        assert nodes != null;
+        assert nodes.length > 0;
```
fickludd, line 139: ´assert edges.length == nodes.length - 1´?

community/values/src/main/java/org/neo4j/values/storable/ArrayValue.java
```diff
@@ -20,12 +20,15 @@
 package org.neo4j.values.storable;
 
 import org.neo4j.values.SequenceValue;
+import org.neo4j.values.AnyValue;
```
fickludd: not used

community/values/src/main/java/org/neo4j/values/BaseToObjectValueWriter.java
```diff
@@ -0,0 +1,545 @@
...
+import org.neo4j.values.storable.TextArray;
+import org.neo4j.values.storable.TextValue;
+import org.neo4j.values.virtual.CoordinateReferenceSystem;
+import org.neo4j.values.virtual.EdgeValue;
+import org.neo4j.values.virtual.MapValue;
+import org.neo4j.values.virtual.NodeValue;
+
+import static org.neo4j.helpers.collection.Iterators.iteratorsEqual;
+
+public abstract class BaseToObjectValueWriter<E extends Exception> implements AnyValueWriter<E>
```
fickludd: This class could use some javadoc. It took me a while to understand how it works and why we need it. Would still suggest we move `graphdb-api` related stuff out of this module, but let's to that in a separate PR in that case.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/Add.scala
```diff
@@ -20,27 +20,31 @@
...
-      case (IsList(x), IsList(y)) => x ++ y
-      case (IsList(x), y)         => x ++ Seq(y)
-      case (x, IsList(y))         => Seq(x) ++ y
-      case (x: String, y: Number) => x + y.toString
-      case (x: Number, y: String) => x.toString + y
+      case (x, y) if x == Values.NO_VALUE || y == Values.NO_VALUE => Values.NO_VALUE
+      case (x: IntegralValue, y: IntegralValue) => Values.longValue(StrictMath.addExact(x.longValue(),y.longValue()))
+      case (x: NumberValue, y: NumberValue) => Values.doubleValue(x.doubleValue() + y.doubleValue())
+      case (x: TextValue, y: TextValue) => Values.stringValue(x.stringValue() + y.stringValue())
+      case (IsList(x),  IsList(y)) => VirtualValues.concat(x, y)
```
craigtaverner: Extra space.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/Add.scala
```diff
@@ -20,27 +20,31 @@
...
 import org.neo4j.cypher.internal.compatibility.v3_3.runtime.ExecutionContext
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.helpers.{IsList, TypeSafeMathSupport}
 import org.neo4j.cypher.internal.compatibility.v3_3.runtime.pipes.QueryState
-import org.neo4j.cypher.internal.compatibility.v3_3.runtime.helpers.TypeSafeMathSupport
-import org.neo4j.cypher.internal.compiler.v3_3.helpers.IsList
 import org.neo4j.cypher.internal.frontend.v3_3.CypherTypeException
 import org.neo4j.cypher.internal.frontend.v3_3.symbols._
+import org.neo4j.values._
+import org.neo4j.values.storable._
+import org.neo4j.values.virtual.{ListValue, VirtualValues}
```
craigtaverner: unused import

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/helpers/MapSupport.scala
```diff
@@ -19,57 +19,33 @@
...
-
-    private def id(x: PropertyContainer) = n match {
-      case n: Node         => n.getId
-      case r: Relationship => r.getId
-    }
+  def isMap(x: AnyValue): Boolean = castToMap.isDefinedAt(x)
+
+  def castToMap: PartialFunction[AnyValue, MapValue] = {
+    case x: MapValue => x
+    case x: NodeValue  => x.properties()
```
craigtaverner: In the older version we made clear use of the QueryContext to access the kernel to lookup properties. This new approach implies that either the properties are contained in the node (eagerly looked up), or the Node has capability to do this lookup, much like the NodeProxy which we wanted to get away from. I like that the code is so much cleaner, I just wonder if we've moving away from the attempt to control kernel access that we had before?

community/values/src/main/java/org/neo4j/values/storable/FloatingPointArray.java
```diff
@@ -19,6 +19,8 @@
  */
 package org.neo4j.values.storable;
 
+import org.neo4j.values.AnyValue;
```
fickludd: not used

community/values/src/main/java/org/neo4j/values/storable/IntegralArray.java
```diff
@@ -19,6 +19,8 @@
  */
 package org.neo4j.values.storable;
 
+import org.neo4j.values.AnyValue;
```
fickludd: not used

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/DistanceFunction.scala
```diff
@@ -26,19 +26,20 @@ import org.neo4j.cypher.internal.compatibility.v3_3.runtime.pipes.QueryState
...
+  override def apply(ctx: ExecutionContext)(implicit state: QueryState): AnyValue = {
     // TODO: Support better calculations, like https://en.wikipedia.org/wiki/Vincenty%27s_formulae
     // TODO: Support more coordinate systems
     (p1(ctx), p2(ctx)) match {
-      case (null, _) => null
-      case (_, null) => null
-      case (geometry1: Point, geometry2: Point) => calculateDistance(geometry1, geometry2)
-      case (geometry1: Geometry, geometry2: Geometry) => calculateDistance(geometry1, geometry2)
+      case (x, y) if x == Values.NO_VALUE || y == Values.NO_VALUE => Values.NO_VALUE
+      case (geometry1: PointValue, geometry2: PointValue) => calculateDistance(geometry1, geometry2)
```
craigtaverner, line 40: We have lost support for Geometry conversion to Points. The coerceToPoint method still exists, but is no longer used.

community/values/src/test/java/org/neo4j/values/virtual/ListTest.java
```diff
@@ -198,4 +202,37 @@ public void shouldRecurseAndCoerce()
...
+                    {
+                        long l = ((LongValue) anyValue).longValue();
+                        return l == 1L || l == 4L || l == 7L;
+                    } ),
+                    VirtualValues.slice( list( -2L, 1L, 4L, 7L, 10L ), 1, 4 ),
+                    VirtualValues.drop( list( -2L, 1L, 4L, 7L ), 1 ),
+                    VirtualValues.take( list( 1L, 4L, 7L, 10L, 13L ), 3 ),
+                    VirtualValues.transform( list( 0L, 3L, 6L ),
+                            anyValue -> Values.longValue( ((LongValue) anyValue).longValue() + 1L ) ),
+                    VirtualValues.reverse( list( 7L, 4L, 1L ) )};
```
fickludd: I really like that you've added these ListValue manipulator methods. Did you consider putting them on the `ListValue` class instead?

community/values/src/test/java/org/neo4j/values/virtual/ListTest.java
```diff
@@ -198,4 +202,37 @@ public void shouldRecurseAndCoerce()
...
+    {
+        for ( ListValue list1 : lists )
+        {
+            for ( ListValue list2 : lists )
+            {
+                assertEqual( list1, list2 );
+                assertArrayEquals(
+                        format( "%s.asArray != %s.toArray", list1.getClass().getSimpleName(),
+                                list2.getClass().getSimpleName() ),
+                        list1.asArray(), list2.asArray() );
```
fickludd, line 258: Maybe adding a similar set of lists that are *not* equal would be good as a negative test?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/FilterFunction.scala
```diff
@@ -21,15 +21,22 @@ package org.neo4j.cypher.internal.compatibility.v3_3.runtime.commands.expression
...
 case class FilterFunction(collection: Expression, id: String, predicate: Predicate)
   extends NullInNullOutExpression(collection)
   with ListSupport
   with Closure {
-  def compute(value: Any, m: ExecutionContext)(implicit state: QueryState) =
-    makeTraversable(value).filter(element => predicate.isTrue(m.newWith1(id, element)  ))
+
+  def compute(value: AnyValue, m: ExecutionContext)(implicit state: QueryState) = {
+    val traversable = makeTraversable(value)
+    VirtualValues.filter(traversable, new java.util.function.Function[AnyValue, java.lang.Boolean]() {
```
craigtaverner, line 36: Would it be possible to have a lambda here for cleanliness? Pehaps not possible on the scala/java interface, I guess. Or we have a scala version of VirtualValues.filter that wraps the java version.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/IndexedInclusiveLongRange.scala
```diff
@@ -43,6 +43,7 @@ case class IndexedInclusiveLongRange(start: Long, end: Long, step: Long) extends
     }
   }
 
+  hashCode()
```
craigtaverner: What does this do?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/PointFunction.scala
```diff
@@ -19,24 +19,31 @@
...
-        null
-      } else {
-        Points.fromMap(map)
-      }
+      if (containsNull(mapCreator)) Values.NO_VALUE
+      else AnyValues.fromMap(mapCreator)
     case x => throw new CypherTypeException(s"Expected a map but got $x")
   }
 
+  private def containsNull(map: MapValue) = {
```
craigtaverner: Is there perhaps value in having this method on MapValue itself? It looks a bit ugly in scala code.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/ProjectedPath.scala
```diff
@@ -21,7 +21,9 @@ package org.neo4j.cypher.internal.compatibility.v3_3.runtime.commands.expression
 
 import org.neo4j.cypher.internal.compatibility.v3_3.runtime.ExecutionContext
 import org.neo4j.cypher.internal.compatibility.v3_3.runtime.pipes.QueryState
-import org.neo4j.graphdb.{Node, Relationship}
+import org.neo4j.values.AnyValue
+import org.neo4j.values.storable.Values
+import org.neo4j.values.virtual.{EdgeValue, ListValue, NodeValue}
```
craigtaverner: Unused imports

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/RelationshipEndPoints.scala
```diff
@@ -20,21 +20,19 @@
...
+    case v if v == Values.NO_VALUE => Values.NO_VALUE
     case value =>
-      val rel = castOrFail[Relationship](value)
-
-      if (start)
-        state.query.relationshipStartNode(rel)
-      else
-        state.query.relationshipEndNode(rel)
+      val rel = castOrFail[EdgeValue](value)
+      if (start) rel.startNode()
```
craigtaverner: As in my comment with MapUtils above, we are no longer using the QueryContext to access kernel resources, and instead using the RelationshipProxy. At the very least this would hide dbHits in the plan profile, but it also feels like we are not moving away from the use of RelationshipProxy in Cypher code.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/ShortestPathExpression.scala
```diff
@@ -136,7 +153,8 @@ case class ShortestPathExpression(shortestPathPattern: ShortestPath, predicates:
   private def addAllOrNoneNodeExpander(ctx: ExecutionContext, currentExpander: Expander, all: Boolean,
                                        predicate: Predicate, relName: String,
                                        currentNodePredicates: Seq[KernelPredicate[PropertyContainer]])
-                                      (implicit state: QueryState): (Expander, Seq[KernelPredicate[PropertyContainer]]) = {
+                                      (implicit
```
craigtaverner: Line break in odd place.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/predicates/ComparablePredicate.scala
```diff
@@ -55,29 +63,31 @@ abstract sealed class ComparablePredicate(val left: Expression, val right: Expre
...
   def isMatch(m: ExecutionContext)(implicit state: QueryState): Option[Boolean] = {
     val a1 = a(m)
     val b1 = b(m)
 
     (a1, b1) match {
-      case (null, _) => None
-      case (_, null) => None
-      case _         => Some(Equivalent(a1) == b1)
+      case (x, y) if x == Values.NO_VALUE || y == Values.NO_VALUE => None
+      case (x: ArrayValue, y: ListValue) => Some(VirtualValues.fromArray(x).equals(y))
```
craigtaverner: I wonder if it would be nicer to have ListValue and ArrayValue be able to handle equals of each others type? Here you convert always to ListValue. Is that most efficient?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/predicates/ComparablePredicate.scala
```diff
@@ -55,29 +63,31 @@ abstract sealed class ComparablePredicate(val left: Expression, val right: Expre
...
   }
 
   def isMatch(m: ExecutionContext)(implicit state: QueryState): Option[Boolean] = {
     val a1 = a(m)
     val b1 = b(m)
 
     (a1, b1) match {
-      case (null, _) => None
-      case (_, null) => None
-      case _         => Some(Equivalent(a1) == b1)
```
craigtaverner, line 72: Equivalent seemed to deal with quite a few things, including float/double comparisons and eagerisation of collections. I guess we have all the same capabilities in the values module now.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/predicates/Predicate.scala
```diff
@@ -167,52 +195,64 @@ case class PropertyExists(variable: Expression, propertyKey: KeyToken) extends P
...
       case _ => None
     }
 
   def containsIsNull = false
 
   def rewrite(f: (Expression) => Expression) = f(regexExpr.rewrite(f) match {
-    case lit: Literal => LiteralRegularExpression(lhsExpr.rewrite(f), lit)(converter)
-    case other        => RegularExpression(lhsExpr.rewrite(f), other)(converter)
-  })
+                                                   case lit: Literal => LiteralRegularExpression(lhsExpr.rewrite(f),
```
craigtaverner: This file is full of reformating changes, some of which (like this one) are quite ugly. Did we need to do this much reformatting?

craigtaverner (changes requested):

community/values/src/test/java/org/neo4j/values/virtual/ListTest.java
fickludd: And also, we should add a hash code check here, to be certain of consistent map behaviour.

community/values/src/main/java/org/neo4j/values/virtual/VirtualValues.java
```diff
@@ -39,12 +49,103 @@ private VirtualValues()
...
+        for ( ListValue list : lists )
+        {
+            totalSize += list.size();
+        }
+
+        AnyValue[] anyValues = new AnyValue[totalSize];
+        int startPoint = 0;
+        for ( ListValue list : lists )
+        {
+            System.arraycopy( list.asArray(), 0, anyValues, startPoint, list.size() );
```
fickludd: These `asArray` calls might allocate extra transient arrays depending on the underlying list implementation. Having a `ListValue.copyToArray( AnyValue[] target, int startPoint )` method instead could solve this (here and methods below).

community/values/src/main/java/org/neo4j/values/virtual/NodeValue.java
```diff
@@ -68,4 +63,107 @@ public String toString()
...
+            writer.writeNode( node.getId(), l, p );
+        }
+
+        @Override
+        public TextArray labels()
+        {
+            TextArray l = labels;
+            if ( l == null )
+            {
+                synchronized ( this )
```
fickludd, line 133: This is the only place where I see concurrency-safe lazy vals, e.g. `ListValue` and `AnyValue` hash do not do this. I guess this was fuelled by some need?

community/values/src/test/java/org/neo4j/values/storable/ValueComparisonTest.java
```diff
@@ -87,7 +84,10 @@
             Float.MAX_VALUE,
             Double.MAX_VALUE,
             Double.POSITIVE_INFINITY,
-            Double.NaN
+            Double.NaN,
+
+            // OTHER
+            null
```
fickludd, line 90: Moving NO_VALUE to the end will change behavior from `3.2` in the kernel at least. Does that matter? I don't know. Does cypher 3.2 put nulls last?

community/values/src/main/java/org/neo4j/values/storable/Values.java
```diff
@@ -42,8 +42,15 @@
...
     public static final Value MAX_NUMBER = Values.doubleValue( Double.NaN );
+    public static final Value ZERO_FLOAT = Values.doubleValue( 0.0 );
+    public static final Value ZERO_INT = Values.longValue( 0 );
     public static final Value MIN_STRING = Values.stringValue( "" );
     public static final Value MAX_STRING = Values.booleanValue( false );
+    public static final BooleanValue TRUE = Values.booleanValue( true );
+    public static final BooleanValue FALSE = Values.booleanValue( false );
+    public static final TextValue EMPTY_STRING = Values.stringValue( "" );
+    public static final DoubleValue E = Values.doubleValue( Math.E );
+    public static final DoubleValue PI = Values.doubleValue( Math.PI );
```
fickludd, line 53: Very nice!

community/values/src/main/java/org/neo4j/values/storable/Values.java
```diff
@@ -143,87 +150,87 @@ public static Value numberValue( Number number )
         throw new UnsupportedOperationException( "Unsupported type of Number " + number.toString() );
     }
 
-    public static Value longValue( long value )
+    public static LongValue longValue( long value )
```
fickludd, line 153: And been wanting to return specific types too!

community/values/src/main/java/org/neo4j/values/virtual/VirtualValues.java
```diff
@@ -19,18 +19,28 @@
...
-import org.neo4j.values.VirtualValue;
 
 /**
  * Entry point to the virtual values library.
  */
 @SuppressWarnings( "WeakerAccess" )
 public final class VirtualValues
 {
+    public static MapValue EMPTY_MAP = new MapValue( Collections.emptyMap() );
+    public static ListValue EMPTY_LIST = new ListValue.ArrayListValue( new AnyValue[0] );
```
fickludd: These guys should be `final`. Good to have around.

community/values/src/main/java/org/neo4j/values/virtual/VirtualValues.java
```diff
@@ -39,12 +49,103 @@ private VirtualValues()
...
+    }
+
+    public static ListValue transform( ListValue list, Function<AnyValue,AnyValue> transForm )
+    {
+        return new ListValue.TransformedListValue( list, transForm );
+    }
+
+    public static ListValue reverse( ListValue list )
+    {
+        return new ListValue.ReversedList( list );
```
fickludd, line 108: I like all these functional methods, but we to make a few unit tests to verify such central functionality.

community/values/src/main/java/org/neo4j/values/virtual/VirtualValues.java
```diff
@@ -80,26 +184,44 @@ public static EdgeReference edge( long id )
 
     public static PathValue path( NodeValue[] nodes, EdgeValue[] edges )
     {
+        assert nodes != null;
+        assert edges != null;
+        if ( nodes.length + edges.length % 2 == 0 )
```
fickludd: you want `if ( ( nodes.length + edges.length ) % 2 == 0 )` or maybe `nodes.length != edges.length + 1`?

community/values/src/main/java/org/neo4j/values/result/QueryResult.java
```diff
@@ -17,18 +17,38 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-package org.neo4j.bolt.v1.messaging.infrastructure;
+package org.neo4j.values.result;
```
fickludd, line 20: This move is a little bit mystic to me. I'm not saying it's bad, I would just like to understand the reason, as the interface it not clearly values related. Also some javadoc would be nice.

fickludd (changes requested):
I have been reviewing all changes in the values module, and nothing else.

This is really great work! Impressive work on specializing implementations and general improvements all around. Like the functional style methods on `ListValue`.

Been doing some nitpicking here and there, it's a big PR so there were things to say. The only crucial parts are some cases where I would like some more tests. 

Also, I'm completely fine with merging this PR and addressing comments on the values in a separate PR, to chunk the work up and avoid rebase issues and all that.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/DistanceFunction.scala
pontusmelke: No we still do that the `calculateDistance` function that takes in geometries handles that. I'll try to remove the unused method though.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/FilterFunction.scala
pontusmelke: Yeah the main target is java so here scala will have to suffer a bit. I think when we upgrade to newer version of scala it will handle java lambdas better. We could of course add some utility but that will also add indirection, I think I prefer faster over prettier in the runtime.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/predicates/ComparablePredicate.scala
pontusmelke: yes

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/predicates/ComparablePredicate.scala
pontusmelke: I think it does now, I think someone fixed that while I was on vacation. Good catch

community/values/src/main/java/org/neo4j/values/virtual/NodeValue.java
pontusmelke: My thinking is that we do create a lot of `Node` and `Relationship` - and hence now `NodeValue` and `EdgeValue` - in the interpreted runtime. Most of these object are not actually returned to the user and thus eagerly pulling in labels, types and properties eagerly would add a big cost for all these intermediate objects. In this way the labels and properties will only be fetched when bolt actually tries to serialize the objects.

In a new glorious runtime I think we should be a lot more careful about creating these intermediate objects, but I felt addressing all the cases where we create nodes and relationships unnecessarily in this PR would of had serious impact on my motivation for finishing this piece of work :)

community/values/src/test/java/org/neo4j/values/storable/ValueComparisonTest.java
pontusmelke: Yes, `null` is last, https://github.com/neo4j/neo4j/blob/3.3/community/cypher/cypher-compiler-3.3/src/main/java/org/neo4j/cypher/internal/compiler/v3_3/common/CypherOrderability.java#L131

community/values/src/main/java/org/neo4j/values/result/QueryResult.java
pontusmelke: Yes I agree. It is a little bit values related since the visitor uses values. My grand plan was to have the result being defined outside cypher but maybe it should live in its own module. I don't know, lets discuss and move in a separate PR?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/DistanceFunction.scala
pontusmelke: Or to clarify, we will coerce geometries to points earlier, whenever we try to put a geometry into the context we convert it to a point since that is the only geometry we can handle

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/IndexedInclusiveLongRange.scala
pontusmelke: it computes the `hashCode` at a random place in the code. It has now been removed.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/PointFunction.scala
pontusmelke: maybe

community/values/src/main/java/org/neo4j/values/virtual/NodeValue.java
fickludd: Aha, yes I see. What I actually was wondering about was the *concurrency-safiness*?

community/values/src/test/java/org/neo4j/values/virtual/ListTest.java
pontusmelke: yes, we have taken the static approach for other stuff so figured I'd stick with that. Maybe we can discuss this and change in a future PR if that makes more sense.

community/values/src/test/java/org/neo4j/values/virtual/ListTest.java
pontusmelke: `assertEquals` called on the line above already checks that `a == b && b ==a && a.hashCode == b.hashCode`

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/DistanceFunction.scala
craigtaverner: So you suggest we bring back the conversion later when we support Geometry again? That is possible, I just thought we had tests that would fail. Either those tests have been removed, or I'm remembering wrong.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v2_3/Compatibility.scala
pontusmelke: tried to revert most of the reformatting changes

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/expressions/DistanceFunction.scala
pontusmelke: we do support geometry but we only have points so far. If a user pass in a geometry via parameters etc we will try to convert it to a point since that is the only geometry we know about. Same as before but we only convert/fail earlier.

community/values/pom.xml
pontusmelke: Lets talk about it and move it in a separate PR?

community/values/src/main/java/org/neo4j/values/BaseToObjectValueWriter.java
pontusmelke: fixed

community/values/src/main/java/org/neo4j/values/virtual/VirtualValues.java
pontusmelke: added unit tests

community/values/src/main/java/org/neo4j/values/virtual/NodeValue.java
pontusmelke: NodeValues can be handled from multiple threads

fickludd (approved):

craigtaverner (approved):

PR #9771, opened 2017-08-03
Refactoring around topology service for sensible addressing

phughk:
This was meant to be the same PR as this one that was closed https://github.com/neo4j/neo4j/pull/9715

PR #9782, opened 2017-08-07
3.3 Changes proving there is a race condition

phughk:
https://github.com/neo4j/neo4j/pull/9771 has failing tests and we suspect it is because the `TopologyService` is being called before initialisation is finalised (therefore needs to retry).
If this PR passes tests, then we know it's because the above statement is true.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClient.java
```diff
@@ -80,6 +86,7 @@ public void onFileStreamingComplete( CompletableFuture<Long> signal, StoreCopyFi
 
     StoreId fetchStoreId( MemberId from ) throws StoreIdDownloadFailedException
     {
+        AdvertisedSocketAddress fromAddress = topologyService.findCatchupAddress( from ).orElseThrow( () -> new TopologyLookupException( from ) );
```
phughk: TODO if fails, remove this line

PR #9785, opened 2017-08-08
Refactoring around topology service for sensible addressing

phughk:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
```diff
@@ -78,6 +83,19 @@
         this.refreshPeriod = config.get( CausalClusteringSettings.cluster_topology_refresh ).toMillis();
         this.myself = myself;
         this.groups = config.get( CausalClusteringSettings.server_groups );
+        this.topologyServiceRetryStrategy = resolveStrategy();
+    }
+
+    private TopologyServiceRetryStrategy resolveStrategy()
+    {
+        Duration defaultDuration = DURATION.apply( CausalClusteringSettings.cluster_topology_refresh.getDefaultValue() );
```
martinfurmanski: You really shouldn't need to get the default value, the config framework should handle that.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/TopologyServiceMultiRetryStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.causalclustering.discovery;
+
+import java.util.Optional;
+
+import org.neo4j.causalclustering.identity.MemberId;
+import org.neo4j.helpers.AdvertisedSocketAddress;
+
+public class TopologyServiceMultiRetryStrategy extends MultiRetryStrategy<MemberId,Optional<AdvertisedSocketAddress>> implements TopologyServiceRetryStrategy
```
martinfurmanski, line 27: Do you really need these type-bound specialisations? (TopologyServiceMultiRetryStrategy, TopologyServiceNoRetriesStrategy, TopologyServiceRetryStrategy).

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/RetryStrategy.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.causalclustering.discovery;
+
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public interface RetryStrategy<I, E>
```
martinfurmanski: Some JavaDoc would be nice!

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
```diff
@@ -78,6 +83,19 @@
...
+        long delayMillis =
+                config.getValue( CausalClusteringSettings.cluster_topology_refresh.name() )
+                        .map( configValueToCast -> (Duration) configValueToCast)
+                        .orElse( defaultDuration )
+                        .toMillis() / 2;
```
martinfurmanski: Why /2 ?

phughk: The default refresh rate is every 5s, so by dividing by 2 we can shorten the delay between polls while also keeping it within the refresh window. It is also the reason for a hard-coded "3 retries"

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/TopologyServiceMultiRetryStrategy.java
phughk: I provided the base `RetryStrategy` with generics in case we want to reuse them, but I can simplify it if you'd like

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
phughk: refreshPeriod already provided on line 83

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/TopologyServiceMultiRetryStrategy.java
martinfurmanski: I am suggesting to just use the generic variants.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastClient.java
martinfurmanski: I understand this correctly you should just need to do:
config.get( CausalClusteringSettings.cluster_topology_refresh ).toMillis() / 2

phughk: Addressed in next commit

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/TopologyServiceMultiRetryStrategy.java
phughk: Agreed over slack that having these separate classes makes the usage more readable, so not addressed

digitalstain (approved):

Issue #9052, opened 2017-03-17
Merge ReflectiveProcedureCompiler and procedure-compiler logic


fbiville:
Right now, the validation is implemented twice, albeit a tad bit differently (one is reflection-based, one is [element](https://docs.oracle.com/javase/7/docs/api/javax/lang/model/element/package-summary.html)-based).

Most logic is the same and we should find a way to centralize this as much as possible.
That way, when a new kind of Cypher extension is developed, the effort to keep the compiler up-to-date would remain minimal.

Lojjs:
Hi @fbiville and sorry for the late reply. Our performance team will have a look at this and come back to you.
Best regards Louise, Neo4j Cypher team

fbiville:
No worries, the PR was at a very very early stage.

The idea would consist in extracting a separate module with a common Neo4j extension model and common validation logic (i.e. 95+% of validation cases?), re-used by the reflective validation module and the compile-time one.

Then, find a way to plug custom validation logic (5+% of cases) where relevant.

In terms of validation differences, here are the ones I can think of right now:

 - generic types can be properly validated at compile-time, they are not erased yet (so we could enforce at compile-time the use of `Map<String, Object>` instead of rawtypes or wrong generic type parameters)
 - extension uniqueness can only be 100% validated at runtime, the procedure compiler processes 1 compilation unit at a time and cannot know about the multiple extension JARs deployed against a specific Neo4j instance

fickludd:
Hi again Florent!

I think everyone agree that this makes a lot of sense, but we just have too many other prioritized things to do to get around to this in a while. If you are passionate about pushing this we will gladly guide a PR, but no-one is available to drive this from Neo4j.

Would you mind closing the issue, or do you want it to remain open?

fbiville:
Let's close it for now. I can create a new issue when I have some available bandwidth again.

PR #9788, opened 2017-08-08
[#9052][WIP] Unify two procedure validations

fbiville:
This commit centralizes the common validation between the
reflective (or runtime) procedure compiler and the compile-time
annotation processor.

The validation workflow is defined once and specific validation
functions are implemented on both sides.

### DIFF
--- a/community/pom.xml
+++ b/community/pom.xml
@@ -68,6 +68,7 @@
     <module>procedure-compiler</module>
     <module>values</module>
     <module>ssl</module>
+    <module>procedure-validation</module>
   </modules>
 
   <licenses>
--- a/community/procedure-validation/LICENSES.txt
+++ b/community/procedure-validation/LICENSES.txt
@@ -1,0 +1,4 @@
+This file contains the full license text of the included third party
+libraries. For an overview of the licenses see the NOTICE.txt file.
+
+
--- a/community/procedure-validation/NOTICE.txt
+++ b/community/procedure-validation/NOTICE.txt
@@ -1,0 +1,27 @@
+Neo4j
+Copyright © 2002-2017 Network Engine for Objects in Lund AB (referred to
+in this notice as "Neo Technology")
+   [http://neotechnology.com]
+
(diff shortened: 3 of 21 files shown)


fbiville:
This is barely the start of the PR, it is obviously **not ready for merge** at all.
If you think the timing is too tight for `3.3`, please let me know.

I basically translated `ReflectiveProcedureCompiler#compileFunction` into a generic workflow, while leaving some implementation gaps that I will try to fill when porting this workflow into `ReflectiveProcedureCompiler` first and the annotation processor.

Here is a rough TODO list:

 - [x] Create a new module with the common validation logic
 - [x] Port user-defined function validation
 - [ ] Implement the user-defined function validation logic
     - [ ] on `ReflectiveProcedureCompiler` side
     - [ ] on the annotation processor side
 - [ ] Port user-defined aggregation function validation
 - [ ] Implement the user-defined aggregation function validation logic
     - [ ] on `ReflectiveProcedureCompiler` side
     - [ ] on the annotation processor side
 - [ ] Port user-defined procedure validation
 - [ ] Implement the user-defined procedure validation logic
     - [ ] on `ReflectiveProcedureCompiler` side
     - [ ] on the annotation processor side

When this plan is executed, then the sync between the compile-time and reflective validators will be greatly eased!

Any help and feedback is more than welcome!

fickludd:
Hey @fbiville,

This seems like a reasonable refactoring to perform, but as you can tell not something which has been at top of our priority list. If you are still interested in driving this change I can support and review. Otherwise we will close this until we next have reason to work around procedures.

fbiville:
@fickludd this refactoring was indeed more critical when user-defined aggregation functions were about to be added in 3.2. If there is no new type of extension or no changes on the existing ones in the short/mid-term, we can probably close the issue for now. I do not have much bandwidth on my side at the moment.

fickludd:
@fbiville Ok, let's close it for now, and get back to it when someone has a need. Thanks for putting in the time to write this up in the first place!

PR #9789, opened 2017-08-09
Legacy indexes can be read when in read_only mode

tinwelint:
Previously simply getting a handle to a legacy index would have the transaction
think that it needed write access, by mistake. This is now changed so that
first the existence of the index is checked with only read requirement and then
if it doesn't gets escalated to create-and-get it.


tinwelint:
Replaced by https://github.com/neo4j/neo4j/pull/9792

PR #9791, opened 2017-08-09
3.3 Removed dependency of CC on backup

phughk:


phughk:
This PR should be for 3.2, re-opening for 3.2 instead.

PR #9794, opened 2017-08-10
Add possibility to track not closed kernel statements

MishaDemianenko:
Add possibility to track not closed kernel statements and
where they were opened and closed during transaction lifecycle.
In case if statement tracking is enabled transaction that has unclosed
statement will not gonna be able to commit

### DIFF
--- a/community/collections/src/main/java/org/neo4j/helpers/collection/Iterables.java
+++ b/community/collections/src/main/java/org/neo4j/helpers/collection/Iterables.java
@@ -41,6 +41,7 @@ import java.util.stream.Stream;
 import org.neo4j.function.Predicates;
 import org.neo4j.graphdb.Resource;
 import org.neo4j.graphdb.ResourceIterable;
+import org.neo4j.graphdb.ResourceIterator;
 
 import static org.neo4j.helpers.collection.Iterators.asResourceIterator;
 
@@ -447,6 +448,10 @@ public final class Iterables
      * Returns the given iterable's first element or {@code null} if no
      * element found.
      *
+     * If the {@link Iterable#iterator() iterator} created by the {@code iterable} implements {@link Resource}
+     * it will be {@link Resource#close() closed} in a {@code finally} block after the single item
+     * has been retrieved, or failed to be retrieved.
+     *
      * @param <T> the type of elements in {@code iterable}.
      * @param iterable the {@link Iterable} to get elements from.
      * @return the first element in the {@code iterable}, or {@code null} if no
@@ -454,7 +459,18 @@ public final class Iterables
      */
     public static <T> T firstOrNull( Iterable<T> iterable )
     {
-        return Iterators.firstOrNull( iterable.iterator() );
+        Iterator<T> iterator = iterable.iterator();
+        try
+        {
+            return Iterators.firstOrNull( iterator );
+        }
+        finally
+        {
+            if ( iterator instanceof Resource )
+            {
+                ((Resource) iterator).close();
+            }
+        }
     }
 
     /**
@@ -612,7 +628,10 @@ public final class Iterables
      */
     public static <T> long count( Iterable<T> iterable, Predicate<T> filter )
     {
-        return Iterators.count( iterable.iterator(), filter );
+        try ( ResourceIterator<T> iterator = Iterators.asResourceIterator( iterable.iterator() ) )
+        {
+            return Iterators.count( iterator, filter );
+        }
     }
 
     /**
--- a/community/collections/src/main/java/org/neo4j/helpers/collection/Iterables.java
+++ b/community/collections/src/main/java/org/neo4j/helpers/collection/Iterables.java
@@ -48,6 +48,9 @@ import static org.neo4j.helpers.collection.Iterators.asResourceIterator;
 public final class Iterables
 {
 
+    @SuppressWarnings( "rawtypes" )
+    private static final ResourceIterable EMPTY_RESOURCE_ITERATOR = new EmptyResourceIterable<>();
+
     private Iterables()
     {
     }
@@ -56,6 +59,11 @@ public final class Iterables
     public static <T> Iterable<T> empty()
     {
         return Collections.emptyList();
+    }
+
+    public static <T> Iterable<T> emptyResourceIterable()
+    {
+        return (Iterable<T>) EMPTY_RESOURCE_ITERATOR;
     }
 
     public static <T> Iterable<T> limit( final int limitItems, final Iterable<T> iterable )
@@ -628,9 +636,17 @@ public final class Iterables
      */
     public static <T> long count( Iterable<T> iterable, Predicate<T> filter )
     {
-        try ( ResourceIterator<T> iterator = Iterators.asResourceIterator( iterable.iterator() ) )
+        Iterator<T> iterator = iterable.iterator();
+        try
         {
             return Iterators.count( iterator, filter );
+        }
+        finally
+        {
+            if ( iterator instanceof ResourceIterator )
+            {
+                ((ResourceIterator) iterator).close();
+            }
         }
     }
 
@@ -932,5 +948,18 @@ public final class Iterables
         List<String> names = new ArrayList<>();
         EnumSet.allOf( cls ).forEach( item -> names.add( item.name() ) );
         return names;
+    }
+
+    private static class EmptyResourceIterable<T> implements ResourceIterable<T>
+    {
+        private EmptyResourceIterable()
+        {
+        }
+
+        @Override
+        public ResourceIterator<T> iterator()
+        {
+            return Iterators.emptyResourceIterator();
+        }
     }
 }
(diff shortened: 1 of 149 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
```diff
@@ -235,6 +251,12 @@ final void forceClose()
     {
         if ( referenceCount > 0 )
         {
+            if ( TRACK_STATEMENTS )
```
tinwelint: Good, yes I think the only right place to check this is in `forceClose`. 👍

tinwelint (approved):

MishaDemianenko:
@tinwelint ready for another round of review

community/collections/src/main/java/org/neo4j/helpers/collection/Iterables.java
```diff
@@ -612,7 +628,10 @@ public static String toString( Iterable<?> values, String separator )
      */
     public static <T> long count( Iterable<T> iterable, Predicate<T> filter )
     {
-        return Iterators.count( iterable.iterator(), filter );
+        try ( ResourceIterator<T> iterator = Iterators.asResourceIterator( iterable.iterator() ) )
```
tinwelint: Instead of wrapping in a resource iterator, which is potentially one unnecessary object instantiation, couldn't we just do the usual check of `Resource` in `finally`?

community/graph-algo/src/main/java/org/neo4j/graphalgo/impl/path/DijkstraBidirectional.java
```diff
@@ -174,7 +175,7 @@ public TraversalMetadata metadata()
             if ( NoneStrictMath.compare( thisState + otherSideShortest.doubleValue(), shortestSoFar.doubleValue(), epsilon ) > 0 &&
                  stopAfterLowestCost )
             {
-                return Collections.emptyList();
+                return Iterables.resourceIterable( Collections.emptyList() );
```
tinwelint: There's an `Iterators#emptyResourceIterator` method, perhaps we should introduce a `Iterables#emptyResourceIterable` too?

community/graph-algo/src/main/java/org/neo4j/graphalgo/impl/path/TraversalAStar.java
```diff
@@ -122,7 +122,8 @@ public WeightedPath findSinglePath( Node start, Node end )
                 new SelectorFactory( end, interest ) )
                 .evaluator( includeWhereEndNodeIs( end ) )
                 .traverse( start );
-        return () -> new WeightedPathIterator( lastTraverser.iterator(), costEvaluator, stopAfterLowestWeight );
+        return Iterators.asResourceIterable(
```
tinwelint, line 125: Doesn't look like this transformation is needed since `WeightedPathIterator` is now a `ResourceIterator`

community/graphdb-api/src/main/java/org/neo4j/graphdb/Node.java
```diff
@@ -78,7 +78,9 @@
      * are attached to this node, an empty iterable will be returned.
      *
      * @return all relationships attached to this node
+     * @deprecated this method should return {@link ResourceIterable} instead
      */
+    @Deprecated
```
tinwelint: I wonder if there are better ways than marking these methods as deprecated. I'd guess doing this would cause some confusion as to what to use instead.

community/kernel/src/main/java/org/neo4j/kernel/impl/traversal/BidirectionalTraversalDescriptionImpl.java
```diff
@@ -143,9 +143,20 @@ public Traverser traverse( Node start, Node end )
...
-                startNodes, endNodes ) );
+        return new DefaultTraverser( () ->
+        {
+            Resource resource = statementFactory.get();
+            try
+            {
+                return new BidirectionalTraverserIterator( resource, start, end, sideSelector, collisionPolicy,
+                        collisionEvaluator, maxDepth, startNodes, endNodes );
+            }
+            catch ( Exception e )
```
tinwelint: I prefer the `success` and `finally` approach before `catch ( Exception e )` which isn't as covering.

community/kernel/src/main/java/org/neo4j/kernel/impl/traversal/TraversalBranchImpl.java
```diff
@@ -174,7 +157,24 @@ protected TraversalBranch newNextBranch( Node node, Relationship relationship )
...
-        relationships = PRUNED_ITERATOR;
+        closeRelationships();
+        relationships = Iterators.emptyResourceIterator();
+    }
+
+    @Override
+    public void close()
+    {
+        //TODO: not sure its a right place to close parent
+        parent.close();
```
tinwelint: It feels iffy to close parent here because just because this branch is exhausted doesn't mean that parent it.

community/kernel/src/main/java/org/neo4j/kernel/impl/traversal/TraversalBranchImpl.java
```diff
@@ -174,7 +157,24 @@ protected TraversalBranch newNextBranch( Node node, Relationship relationship )
...
+        closeRelationships();
+    }
+
+    private void closeRelationships()
+    {
+        if ( relationships != null )
+        {
+            relationships.close();
+        }
```
tinwelint, line 166: could you move the assignment of the empty iterator in here instead of outside in all places?

tinwelint (changes requested):
I think the idea of pushing `ResourceIterator` from the inside and out is OK due to `asResourceIterator` checking `instanceof`. I'm unsure about the surface and if deprecation is the way to go.

All the added close calls and what not... doe they affect performance in any way?

community/graph-algo/src/main/java/org/neo4j/graphalgo/impl/path/TraversalAStar.java
MishaDemianenko: it's iterator now, but we need Iterable, that's why we need wrapping

community/kernel/src/main/java/org/neo4j/kernel/impl/traversal/BidirectionalTraversalDescriptionImpl.java
MishaDemianenko: `try-finally` will not help here since its exceptions that `BidirectionalTraverserIterator` can throw during instantiation we covering from here, in normal case this iterator will be closed later

community/graphdb-api/src/main/java/org/neo4j/graphdb/Node.java
MishaDemianenko: I can remove introduced depreciations and leave only a note that it's actually a resource iterator returned that we rely on that and API will be changed in a future version? sounds good?

community/kernel/src/main/java/org/neo4j/kernel/impl/traversal/BidirectionalTraversalDescriptionImpl.java
tinwelint: What I mean is that `success` flag w/ finally block also covers `OutOfMemoryError` or whatever and generally looks better than catch `Exception/Throwable`. Works the same way as the catch above.

What I mean is:

```
boolean succes = false;
try
{
    BidirectionalTraverserIterator result = new BidirectionalTraverserIterator( resource, start, end, sideSelector, collisionPolicy, collisionEvaluator, maxDepth, startNodes, endNodes );
    success = true;
    return result;
}
finally
{
    if ( !success )
    {
        resource.close();
    }
}

MishaDemianenko: ah, ok now i see, thx

community/graphdb-api/src/main/java/org/neo4j/graphdb/Node.java
MishaDemianenko: ok so for now i removed all introduced deprecations

community/kernel/src/main/java/org/neo4j/kernel/impl/traversal/TraversalBranchImpl.java
MishaDemianenko: I reverted all the closing of branches and paths, in all places where it's unclear whenever I close them or not and add a flag to skip open statements checks for graph algo module

community/kernel/src/main/java/org/neo4j/kernel/impl/traversal/TraversalBranchImpl.java
MishaDemianenko: done

MishaDemianenko:
@tinwelint updated

tinwelint (approved):

PR #9796, opened 2017-08-11
Importer collects relationships with missing data

tinwelint:
Previously a relationship with missing START_ID, END_ID or TYPE would immediately
fail an import, due to them being considered mandatory fields. Now they are considered
bad relationships and can be collected instead, if collecting bad relationships is
enabled and it's withing the bad tolerance.

This will allow for a more flexible input reader for imports.

### DIFF
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -443,7 +443,7 @@ public class ImportTool
             input = new CsvInput( nodeData( inputEncoding, nodesFiles ), defaultFormatNodeFileHeader(),
                     relationshipData( inputEncoding, relationshipsFiles ), defaultFormatRelationshipFileHeader(),
                     idType, csvConfiguration( args, defaultSettingsSuitableForTests ), badCollector,
-                    configuration.maxNumberOfProcessors() );
+                    configuration.maxNumberOfProcessors(), !skipBadRelationships );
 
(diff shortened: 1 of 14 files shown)


community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -443,7 +443,7 @@ public static void main( String[] incomingArguments, boolean defaultSettingsSuit
             input = new CsvInput( nodeData( inputEncoding, nodesFiles ), defaultFormatNodeFileHeader(),
                     relationshipData( inputEncoding, relationshipsFiles ), defaultFormatRelationshipFileHeader(),
                     idType, csvConfiguration( args, defaultSettingsSuitableForTests ), badCollector,
-                    configuration.maxNumberOfProcessors() );
+                    configuration.maxNumberOfProcessors(), !skipBadRelationships );
```
ragadeeshu, line 446: Any specific reason for the method parameter to be the opposite of the config? Should one of them be changed?

ragadeeshu (approved):
Looks good, just one small question.

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
tinwelint: Hmm, well... In each of the contexts I prefer them as they are. Otherwise in `CsvInput` that would look like `skipRelationshipValidation` which feels a little reversed, where the check in the implementation would be `if ( !skipRelationshipValidation ) doValidation();`

ragadeeshu:
Sounds like a valid point.

PR #9797, opened 2017-08-11
MemberID address resolution performed earlier

phughk:
Address resolution of cluster members is now performed earlier.
TopologyService has been removed from many classes that will be used in the future in the backup client.
A retry mechanism has been introduced to handle he fact that it is now possible to request an address for a member
before the table of member addresses has been refreshed


phughk:
Closed as this shouldn't even be open

PR #9798, opened 2017-08-11
Added EagerRegisterPipe

SaschaPeukert:


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/EagerRegisterPipe.scala
```diff
@@ -0,0 +1,117 @@
...
+      override def next(): ExecutionContext = {
+        val row = new ExecutionContext {
+          private val globalRefOffset = refIndex
+          private val globalPrimitiveOffset = primitiveIndex
+
+          override def setLongAt(offset: Int, value: Long): Unit = primitiveRows(offset + globalPrimitiveOffset) = value
+
+          override def setRefAt(offset: Int, value: Any): Unit = refRows(offset + globalRefOffset) = value
+
+          // if this is called from somewhere like PrimitiveExecutionContext#copyFrom the array copy is redundant
```
alexaverbuch: @systay what are your thoughts about this? also, generally about this pipe?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/EagerRegisterPipe.scala
```diff
@@ -98,12 +101,42 @@ case class EagerRegisterPipe(source: Pipe, pipelineInformation: PipelineInformat
...
+    def ++=(elements: Array[T]): Unit = {
+      while (length + elements.length > array.length)
+        grow()
+      System.arraycopy(elements, 0, array, length, elements.length)
+      length = length + elements.length
+    }
+
+    protected def newArray(size: Int): Array[T]
+
+    // NOTE: current impl doesn't reuse space of old array. Instead, it creates bigger array & copies old elements to it
```
alexaverbuch: @systay @eebus if/when reviewing PR please pay special attention to `ResizableArray`.
It may turn out you want to revert this commit, to go back to using `ArrayBuffer` (FYI, all tests still pass if this commit is reverted).
The benefit of aggregating into `Array` is we get to `arraycopy` in many places, which (presumably) becomes more beneficial with more columns.
Downside of current impl is (as in comment) every time we resize array we *copy* elements to new array. It's done with `arraycopy` so hopefully low cost, but it *does* mean that we waste space (and some time). 
In worst case this impl consumes 3x more space than necessary -- if we reused old array and "appended arrays" (as in your branch, Andres) then worst case would be 2x (when we are one element past resize border). Maybe we want to make that addition to `ResizableArray`, but it would complicate reading. cc @SaschaPeukert

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
```diff
@@ -160,6 +160,11 @@ object RegisterAllocation {
         result += (lp -> pipeline)
         pipeline
 
+      case Eager(source) =>
+        val pipeline = allocate(source,nullable,argument)
```
systay: I've been thinking about this one - shouldn't this be a pipeline breaker?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/ExecutionContext.scala
```diff
@@ -78,32 +85,32 @@ case class MapExecutionContext(m: MutableMap[String, AnyValue])
     m.foreach(f)
   }
 
-  override def +=(kv: (String, AnyValue)) = {
+  override def +=(kv: (String, AnyValue)): MapExecutionContext.this.type = {
```
systay, line 88: These types are IntelliJ artifacts and not helpful/pretty. In this case, you should just use `MapExecutionContext`.

systay (approved):

systay:
Really nice that you have taken on updating `CypherComparisonSupport` like this. 👍

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/ExecutionContext.scala
systay: Never mind me. ^ That was a lie.

PR #9799, opened 2017-08-11
Make it possible to disable tracking of heap allocation

thobe:


tinwelint (approved):
I think this looks good 👍

thobe:
Build failure looks like a flaky test, it at least looks unrelated to the changes at hand.

There are two main open questions with this change, I'd say:

1. What should the actual names of the config parameters be? I just picked two names that seemed sensible to me, but perhaps we should prefix them with `unsupported.` and annotate them as `@Internal`?
2. Should we change the defaults? There are good reasons why we might want `dbms.track_query_allocation` to be _disabled_ by default.

chrisvest:
@thobe I think the default for heap tracking should be off. If this will leave a column in the listQueries that is always empty, then we should probably avoid hiding the settings since people can clearly see that tracking heap usage is possible, but turned off. And maybe they'd like to turn it on.

thobe:
@chrisvest yes, I agree. The reason I have not done that is because we have already released versions where it is enabled by default.

Having heap tracking off would leave a null-valued column in the output of `listQueries` (as you can see in the test case).

PR #9802, opened 2017-08-11
Fix migration failures for native label scan store

klaren:
Migration in the high limit format is performed with the batch importer that end up creating an empty label scan store file, which made `NativeLabelScanStoreMigrator` think the migration had already happened.

This PR forces the `NativeLabelScanStoreMigrator` to always recreate the file even when it is present.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
@@ -180,12 +180,23 @@ public class NativeLabelScanStore implements LabelScanStore
         this.pageCache = pageCache;
         this.pageSize = pageSize;
         this.fullStoreChangeStream = fullStoreChangeStream;
-        this.storeFile = new File( storeDir, FILE_NAME );
+        this.storeFile = getLabelScanStoreFile( storeDir );
         this.singleWriter = new NativeLabelScanWriter( 1_000 );
         this.readOnly = readOnly;
         this.monitors = monitors;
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -180,7 +180,7 @@ public NativeLabelScanStore( PageCache pageCache, File storeDir, FullStoreChange
         this.pageCache = pageCache;
         this.pageSize = pageSize;
         this.fullStoreChangeStream = fullStoreChangeStream;
-        this.storeFile = new File( storeDir, FILE_NAME );
+        this.storeFile = getLabelScanStoreFile( storeDir );
```
MishaDemianenko, line 183: Since it's a store shouldn't we move this into StoreFile instead? cc @burqen

MishaDemianenko (commented):
Even knowing that this thing is actually fixing another more high-level integration test what about a more fine-grained test for introduced changes?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
burqen: It's not actually considered a store. Not found in `StoreType` or in `StoreFile`.
Maybe it _does_ make sense to _make_ it a store, but I would say that is outside the scope of this PR.

burqen:
@MishaDemianenko Test added 👍

MishaDemianenko (approved):

PR #9803, opened 2017-08-11
Reversed dependency between CC and backup

phughk:
Before this commit, CC is dependant on backup.
Applying this commit will shuffle dependencies around to be more sensible

### DIFF
--- a/enterprise/cluster/pom.xml
+++ b/enterprise/cluster/pom.xml
@@ -104,5 +104,10 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.neo4j</groupId>
+      <artifactId>neo4j-enterprise-kernel</artifactId>
+      <version>${project.version}</version>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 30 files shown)


phughk:
@martinfurmanski once you have reviewed this it is ready for merge, I have already prepared the merge commit for 3.3

enterprise/cluster/pom.xml
```diff
@@ -104,5 +104,10 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.neo4j</groupId>
+      <artifactId>neo4j-enterprise-kernel</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
andrewkerr9000, line 111: Where is this needed?

andrewkerr9000 (approved):
:+1: Although I'm not sure where the dependencies on neo4j-enterprise-kernel are needed.

enterprise/cluster/pom.xml
phughk: Used for `EnterpriseEditionSettings`

digitalstain:
We cannot merge this until the replacement tool for CC backup is in place, correct?

PR #9804, opened 2017-08-14
Native + Lucene Fusion index as default

burqen:
## Fusion index as default
Fusion index = a fusion between native and Lucene index.
Native index is built on top of GB+Tree and can only handle numbers.
Lucene index is old Lucene implementation.

Fusion index works by wrapping Native and Lucene into one logical index where all numbers are directed to Native and the rest (including all arrays and all composite keys) to Lucene.

First benchmarks show a write performance improvement of between 1.5-4x.

## Default behaviour
- All newly created indexes will be 'fusion indexes'.
- Old indexes (created with only Lucene) will not be migrated but continue as 'Lucene only' indexes. Drop and recreate to use fusion index. This is recommended for indexes on number focused properties (not composite or arrays). For indexes over Strings there is no use to move to fusion.

## Migration
Migrating from <3.3 will keep all indexes around and will work just fine as Lucene only. Drop and recreate to move to fusion index.

## Internal config to disable
An internal unsupported config option is introduced to disable fusion index as default. In this case, Lucene will be used instead. Disabling fusion index will NOT migrate already created fusion indexes. Those will have to be dropped and recreated to use the new default (Lucene).

## TL;DR
- New fusion index = native + lucene.
- Fusion is default.
- When creating new index, default is used.
- No migration is done for already existing indexes. Can be migrated to default by drop and recreate.
- Turn fusion index of with internal unsupported config option.
- Faster, Harder, Scooter.

co-author: @tinwelint 

[cl] Introduce new native index for numbers.


community/consistency-check/src/main/java/org/neo4j/consistency/internal/SchemaIndexExtensionLoader.java
```diff
@@ -0,0 +1,81 @@
...
+        public void add( CleanupJob job )
+        {
+            if ( job.needed() )
+            {
+                throw new IllegalStateException( "Consistency checker should not do recovery" );
+            }
+        }
+    }
+
+    public static final RecoveryCleanupWorkCollector RECOVERY_PREVENTING_COLLECTOR = new RecoveryPreventingCollector();
```
MishaDemianenko: Please fields first, public member and them private member

MishaDemianenko: not sure what collectors class and definition does in SchemaIndexExtensionLoader, even if it's passed as one of the parameters to instantiate extensions

community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/AllByPrioritySelectionStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.neo4j.kernel.extension.KernelExtensionUtil.servicesClassPathEntryInformation;
+
+/**
+ * Selects the candidate with highest priority (assumed to implement {@link Comparable}) and returns
+ * in {@link #select(Class, Iterable)}, but keeps the others for access too.
+ */
+public class AllByPrioritySelectionStrategy<T> implements DependencyResolver.SelectionStrategy
+{
+    @SuppressWarnings( "rawtypes" )
+    private List lowerPrioritizedCandidates = Collections.emptyList();
```
MishaDemianenko: `private List<T> lowerPrioritizedCandidates = Collections.emptyList();` then you do not need to suppress warnings

community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/AllByPrioritySelectionStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+import org.neo4j.graphdb.DependencyResolver;
+import org.neo4j.helpers.collection.Iterables;
+
+import static org.neo4j.kernel.extension.KernelExtensionUtil.servicesClassPathEntryInformation;
+
+/**
+ * Selects the candidate with highest priority (assumed to implement {@link Comparable}) and returns
+ * in {@link #select(Class, Iterable)}, but keeps the others for access too.
+ */
+public class AllByPrioritySelectionStrategy<T> implements DependencyResolver.SelectionStrategy
```
MishaDemianenko: `<T extends Comparable>` instead `<T>`

community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/AllByPrioritySelectionStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+            throw new IllegalArgumentException( "Could not resolve dependency of type: " +
+                                                type.getName() + ". " + servicesClassPathEntryInformation() );
+        }
+        Collections.sort( all, Collections.reverseOrder() );
+        R highest = (R) all.remove( 0 );
+        lowerPrioritizedCandidates = all;
+        return highest;
+    }
+
+    @SuppressWarnings( "unchecked" )
```
MishaDemianenko: no need to suppress warnings

community/kernel/src/main/java/org/neo4j/kernel/extension/dependency/AllByPrioritySelectionStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+public class AllByPrioritySelectionStrategy<T> implements DependencyResolver.SelectionStrategy
+{
+    @SuppressWarnings( "rawtypes" )
+    private List lowerPrioritizedCandidates = Collections.emptyList();
+
+    @SuppressWarnings( {"rawtypes", "unchecked"} )
+    @Override
+    public <R> R select( Class<R> type, Iterable<R> candidates ) throws IllegalArgumentException
+    {
+        List<Comparable> all = (List<Comparable>) Iterables.asList( candidates );
```
MishaDemianenko: ` List<T> all = (List<T>) Iterables.asList( candidates );` instead

community/consistency-check/src/test/java/org/neo4j/consistency/checking/full/FullCheckIntegrationTest.java
```diff
@@ -556,11 +556,13 @@ public void shouldReportNodesThatAreNotIndexed() throws Exception
...
             IndexRule indexRule = indexRuleIterator.next();
-            IndexAccessor accessor = fixture.directStoreAccess().indexes().getOnlineAccessor(
-                    indexRule.getId(), indexRule.getIndexDescriptor(), samplingConfig );
+            IndexAccessor accessor = fixture.directStoreAccess().indexes().
+                    apply( indexRule.getProviderDescriptor() ).getOnlineAccessor(
+                            indexRule.getId(), indexRule.getIndexDescriptor(), samplingConfig );
             IndexUpdater updater = accessor.newUpdater( IndexUpdateMode.ONLINE );
             updater.remove( asPrimitiveLongSet( indexedNodes ) );
             updater.close();
+            accessor.force();
```
MishaDemianenko, line 565: what's the point doing force specifically before closing accessor?

community/index/src/main/java/org/neo4j/index/internal/gbptree/InternalTreeLogic.java
```diff
@@ -387,9 +387,11 @@ private int search( PageCursor cursor, KEY key, KEY readKey, int keyCount )
     private boolean cursorIsAtExpectedLocation( PageCursor cursor )
     {
         assert currentLevel >= 0 : "Uninitialized tree logic, currentLevel:" + currentLevel;
-        assert cursor.getCurrentPageId() == levels[currentLevel].treeNodeId : "Expected cursor to be at page:" +
-                    levels[currentLevel].treeNodeId + " at level:" + currentLevel + ", but was at page:" +
-                    cursor.getCurrentPageId();
+        long currentPageId = cursor.getCurrentPageId();
+        long expectedPageId = levels[currentLevel].treeNodeId;
+        assert currentPageId == expectedPageId : "Expected cursor to be at page:" +
```
MishaDemianenko, line 392: noo, that was one of my favourite non-readable statements 😞

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -394,6 +394,10 @@
     public static final Setting<Boolean> multi_threaded_schema_index_population_enabled =
             setting( "unsupported.dbms.multi_threaded_schema_index_population_enabled", BOOLEAN, TRUE );
 
+    @Internal
+    public static final Setting<Boolean> enable_native_schema_index =
```
MishaDemianenko, line 404: surprising option

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/BatchingMultipleIndexPopulator.java
```diff
@@ -181,23 +160,10 @@ private void awaitCompletion()
      */
     private void batchUpdate( IndexPopulation population, IndexEntryUpdate<?> update )
```
MishaDemianenko: this method is not used at all now

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/BatchingMultipleIndexPopulator.java
```diff
@@ -206,18 +172,7 @@ private void flushIfNeeded( IndexPopulation population, List<IndexEntryUpdate<?>
...
-            Map.Entry<IndexPopulation,List<IndexEntryUpdate<?>>> entry = entries.next();
-            IndexPopulation population = entry.getKey();
-            List<IndexEntryUpdate<?>> updates = entry.getValue();
-            entries.remove();
-            if ( updates != null && !updates.isEmpty() )
-            {
-                flush( population, updates );
-            }
-        }
+        populations.forEach( population -> flush( population, population.takeCurrentBatch() ) );
```
MishaDemianenko: there is no need to pass second param here since population is already passed as first one

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
```diff
@@ -86,6 +89,7 @@
 
     public static final String QUEUE_THRESHOLD_NAME = "queue_threshold";
     private final int QUEUE_THRESHOLD = FeatureToggles.getInteger( getClass(), QUEUE_THRESHOLD_NAME, 20_000 );
+    private final int BATCH_SIZE = FeatureToggles.getInteger( BatchingMultipleIndexPopulator.class, BATCH_SIZE_NAME, 10_000 );
```
MishaDemianenko: feature toggle is moved but class is the same? why?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionSchemaIndexProvider.java
```diff
@@ -80,29 +78,46 @@ public IndexAccessor getOnlineAccessor( long indexId, IndexDescriptor descriptor
     @Override
     public String getPopulationFailure( long indexId ) throws IllegalStateException
     {
-        String nativeFailure = nativeProvider.getPopulationFailure( indexId );
-        String luceneFailure = luceneProvider.getPopulationFailure( indexId );
-        if ( nativeFailure != null )
+        String nativeFailure = null;
+        try
```
MishaDemianenko, line 82: what is the sense in throwing and catching useless exceptions instead of updating providers to not throw exceptions if not failed?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionSchemaIndexProvider.java
```diff
@@ -80,29 +78,46 @@ public IndexAccessor getOnlineAccessor( long indexId, IndexDescriptor descriptor
...
         {
-            return luceneFailure;
+            luceneFailure = luceneProvider.getPopulationFailure( indexId );
+        }
+        catch ( IllegalStateException e )
+        {   // Just catch
         }
+
+        if ( nativeFailure != null || luceneFailure != null )
```
MishaDemianenko, line 98: please use `StringUtils.isNotEmpty` instead

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -408,9 +410,16 @@ public void start() throws IOException
...
         life.add( recoveryCleanupWorkCollector );
 
-        schemaIndexProvider = dependencyResolver.resolveDependency( SchemaIndexProvider.class,
-                HighestSelectionStrategy.getInstance() );
-        dependencies.satisfyDependency( schemaIndexProvider );
+        AllByPrioritySelectionStrategy<SchemaIndexProvider> indexProviderSelection = new AllByPrioritySelectionStrategy<>();
+        SchemaIndexProvider defaultIndexProvider =
+                dependencyResolver.resolveDependency( SchemaIndexProvider.class, indexProviderSelection );
+
+        // todo do we really need to have this dependency?
```
MishaDemianenko, line 416: TODO

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexSamplingCancellationTest.java
```diff
@@ -1,199 +0,0 @@
-/*
```
MishaDemianenko, line 1: Not quite sure why this test was removed, we do not cancel sampling anymore?

MishaDemianenko (changes requested):
Couple of small comments here and there

community/consistency-check/src/main/java/org/neo4j/consistency/internal/SchemaIndexExtensionLoader.java
tinwelint: Well, this class is a utility used in a bunch of places in consistency checker. It loads schema index provider kernel extensions and consistency checker is not allowed to do recovery. Where should it go?

community/consistency-check/src/test/java/org/neo4j/consistency/checking/full/FullCheckIntegrationTest.java
tinwelint: That's the general contract and something that native indexing hooks in on to checkpoint properly. It's just that lucene happened to make do without it.

community/index/src/main/java/org/neo4j/index/internal/gbptree/InternalTreeLogic.java
tinwelint: Bugger, well I'd guess it wasn't the last of its kind :)

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
tinwelint: We went for it since it's easier to manage, also in tests. Now that we've gotten all fixed and working we can investigate if it would be tricky to manage this if it were a toggle instead. Hang on.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
tinwelint: Something of a confusion move at a time we were confused about multi index populator. Forgot about it... we can move it back.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionSchemaIndexProvider.java
tinwelint: I think we want to check `null` explicitly. Empty string means something else in this case.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
tinwelint: Ah, to be investigated...

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexSamplingCancellationTest.java
tinwelint: It only tested that the in-memory test index `HashBasedIndex` sample method threw exception in a particular case. I think it was mis-written to begin with.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionSchemaIndexProvider.java
tinwelint: `getPopulationFailure` is designed to throw `IllegalStateException` if called for index that isn't in `FAILED` state, sensible methinks.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
tinwelint: It was left there because that's how it was previously. Now the `SchemaIndexProviderMap` is used instead. It's not strictly required for us to remove this right now, since it may require some time and testing. We've created a card for it now.

tinwelint:
I think we need to look at `SchemaIndexHaIT` dude, it fails consisently on this branch... not very often at all otherwise. I think it's simply a kernel extension loading thingie.

community/consistency-check/src/main/java/org/neo4j/consistency/internal/SchemaIndexExtensionLoader.java
MishaDemianenko: well if you will need in any other test ever RecoveryPreventingCollector I would say that SchemaIndexExtensionLoader would be the last place where i would expect to find it, even in separate class is better imho

community/consistency-check/src/test/java/org/neo4j/consistency/checking/full/FullCheckIntegrationTest.java
MishaDemianenko: I mean if you close it without force it should do that automatically right?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
MishaDemianenko: The option still present in this class

MishaDemianenko:
@tinwelint @burqen there is one small thing still please fix that otherwise its good and approved

MishaDemianenko (approved):

tinwelint:
The build passed, although the Github <--> CI integration didn't quite get that bit. @burqen was there some issue with leaving open files behind (as shown in some benchmark)? If so perhaps we should hold off merge until we know more.

burqen:
@tinwelint It's a bit unclear. There does not seem to be any leaking files but we need some more investigation before merging.

PR #9811, opened 2017-08-14
Applies label and index updates in parallel

tinwelint:
When applying a transaction both label updates and index updates are enqueued and applied
as a batch at the end of the transaction(s). On top of this sits one WorkSync for label updates
and another for index updates, batching together work from concurrent transactions so that
only one out of all concurrently applying transactions applies all label and index updates.

Previously label updates and index updates were applied in sequence, from one transaction's
perspective in that it would submit its label update to the WorkSync and either apply them,
together with other concurrent transaction's label updates, or just sit and wait for them to
be applied by another concurrent transaction. Only after that would it do the same for its
index updates.

This commit further increases batching and parallelism by calling WorkSync#applyAsync
for label updates and moving on to index updates immediately and in the end awaiting label
updates to complete.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/command/IndexBatchTransactionApplier.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/command/IndexBatchTransactionApplier.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Supplier;
 
+import org.neo4j.concurrent.AsyncApply;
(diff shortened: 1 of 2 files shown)


MishaDemianenko:
looks good for now, will wait with approve till benchmarks run will be available

MishaDemianenko:
@tinwelint so do we have a run somewhere?

MishaDemianenko:
@tinwelint any news on this one?

tinwelint:
@MishaDemianenko not really, the benchmark ran on overtime and looked strange so stopped it. Will try to get this going again.

MishaDemianenko (approved):

PR #9812, opened 2017-08-15
Introduce `dbms.setConfigValue` to change dynamic configuration values

klaren:
Enterprise version of neo4j now has a way to change configuration values at runtime through procedures. E.g.

    CALL dbms.setConfigValue('dbms.logs.query.enabled', 'true')

The procedure takes a setting followed by a value. If the value is empty, the setting will fallback to the default value. This only works for settings marked as dynamic, which is currently only:

- `dbms.transaction.timeout`
- `dbms.logs.query.threshold`
- `dbms.logs.query.enabled`

**Note:** Changes are never persisted and if the server is restarted all changes will be lost.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
@@ -565,21 +565,20 @@ public class Config implements DiagnosticsProvider, Configuration
     }
 
     /**
+     * Updates a provided setting to a given value. This method is intended to be used for changing settings during
+     * runtime. If you want to change settings at startup, use {@link Config#augment}.
      *
-     * TODO: DRAGONS! This code works for the current dynamic variables(3), but is not generic.
-     * TODO: DRAGONS! No migration or config validation is done.
+     * @implNote No migration or config validation is done. If you need this you have to refactor this method.
      *
      * @param setting The setting to set to the specified value.
      * @param value The new value to set, passing {@code null} or empty should reset the value back to default value.
      * @throws IllegalArgumentException if the provided setting is unknown or not dynamic.
      * @throws InvalidSettingException if the value is not formatted correctly.
      */
-    public void setConfigValue( String setting, String value ) throws IllegalArgumentException, InvalidSettingException
+    public void updateDynamicSetting( String setting, String value ) throws IllegalArgumentException, InvalidSettingException
     {
         // Make sure the setting is valid and is marked as dynamic
-        Optional<ConfigValue> option =
-                configOptions.stream().map( it -> it.asConfigValues( params ) ).flatMap( List::stream )
-                        .filter( it -> it.name().equals( setting ) ).findFirst();
+        Optional<ConfigValue> option = findConfigValue( setting );
 
         if ( !option.isPresent() )
         {
@@ -615,6 +614,12 @@ public class Config implements DiagnosticsProvider, Configuration
 
             params.put( setting, value );
         }
+    }
+
+    private Optional<ConfigValue> findConfigValue( String setting )
+    {
+        return configOptions.stream().map( it -> it.asConfigValues( params ) ).flatMap( List::stream )
+                .filter( it -> it.name().equals( setting ) ).findFirst();
     }
 
     /**
--- a/community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
@@ -79,7 +79,7 @@ public class Config implements DiagnosticsProvider, Configuration
     private final Map<String,String> params = new CopyOnWriteHashMap<>(); // Read heavy workload
     private final ConfigurationMigrator migrator;
     private final List<ConfigurationValidator> validators = new ArrayList<>();
-    private final Map<String,String> overriddenDefaults = new CopyOnWriteHashMap<>(); // TODO : has bo be reapplied every update
+    private final Map<String,String> overriddenDefaults = new CopyOnWriteHashMap<>();
 
     // Messages to this log get replayed into a real logger once logging has been instantiated.
     private Log log = new BufferingLog();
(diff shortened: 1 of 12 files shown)


enterprise/kernel/src/test/java/org/neo4j/kernel/enterprise/builtinprocs/SetConfigValueProcedureTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+public class SetConfigValueProcedureTest
+{
+    @Rule
+    public final DatabaseRule db = new ImpermanentEnterpriseDatabaseRule()
+    {
+        @Override
+        protected void configure( GraphDatabaseBuilder builder )
+        {
+            builder.setConfig( cypher_hints_error, "true" );
```
MishaDemianenko: Not sure why do you need this setting here

enterprise/kernel/src/test/java/org/neo4j/kernel/enterprise/builtinprocs/SetConfigValueProcedureTest.java
```diff
@@ -0,0 +1,121 @@
...
+            db.execute( "CALL dbms.setConfigValue('unknown.setting.indeed', 'foo')" );
+        }
+        catch ( Exception e )
+        {
+            Throwable cause = Exceptions.rootCause( e );
+            assertThat( cause, instanceOf( IllegalArgumentException.class ) );
+            assertThat( cause.getMessage(), equalTo( "Unknown setting: unknown.setting.indeed" ) );
+            return;
+        }
+        fail( "Should throw IllegalArgumentException" );
```
MishaDemianenko: you can fail inside of try block instead after `db.execute` then you do not need to have `return` statement inside your catch block; also you can use `ExpectedException` rule

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -565,6 +565,59 @@ public void setLogger( Log log )
...
+     *
+     * @param setting The setting to set to the specified value.
+     * @param value The new value to set, passing {@code null} or empty should reset the value back to default value.
+     * @throws IllegalArgumentException if the provided setting is unknown or not dynamic.
+     * @throws InvalidSettingException if the value is not formatted correctly.
+     */
+    public void setConfigValue( String setting, String value ) throws IllegalArgumentException, InvalidSettingException
+    {
+        // Make sure the setting is valid and is marked as dynamic
+        Optional<ConfigValue> option =
```
MishaDemianenko: i would extract searching of config option to a separate helper function

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -565,6 +565,59 @@ public void setLogger( Log log )
...
+     *
+     * TODO: DRAGONS! This code works for the current dynamic variables(3), but is not generic.
+     * TODO: DRAGONS! No migration or config validation is done.
+     *
+     * @param setting The setting to set to the specified value.
+     * @param value The new value to set, passing {@code null} or empty should reset the value back to default value.
+     * @throws IllegalArgumentException if the provided setting is unknown or not dynamic.
+     * @throws InvalidSettingException if the value is not formatted correctly.
+     */
+    public void setConfigValue( String setting, String value ) throws IllegalArgumentException, InvalidSettingException
```
MishaDemianenko: i think we need to rename the method to highlight that it's not a generic set of config value, but a specific update of dynamic properties for runtime only (still works for non-runtime, but targeted use case is runtime update)

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -565,6 +565,59 @@ public void setLogger( Log log )
     }
 
     /**
+     *
+     * TODO: DRAGONS! This code works for the current dynamic variables(3), but is not generic.
```
MishaDemianenko: and a proper java doc please, no todos, describe instead what is does clearly

enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerTest.java
```diff
@@ -69,6 +69,7 @@
     public final FakeHeapAllocation heapAllocation = new FakeHeapAllocation();
     private long pageHits;
     private long pageFaults;
+    private long threshold = 10; //ms
```
MishaDemianenko: please rename variable instead to highlight that it's `ms` we are talking about here

MishaDemianenko (changes requested):
Couple of minor comments otherwise looks good

MishaDemianenko:
@craigtaverner can you please take a look into procedure part (we introduce new ent. only procedure in here) thx in advance

klaren:
Great feedback @MishaDemianenko ! New commits ready for review addressing it.

enterprise/kernel/src/test/java/org/neo4j/kernel/enterprise/builtinprocs/SetConfigValueProcedureTest.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Test
+    public void failIfUnknownSetting() throws Exception
+    {
+        expect.expect( new NestedThrowableMatcher( IllegalArgumentException.class ) );
+        expect.expectMessage( "Unknown setting: unknown.setting.indeed" );
+
+        db.execute( "CALL dbms.setConfigValue('unknown.setting.indeed', 'foo')" );
+        fail( "Should throw IllegalArgumentException" );
```
MishaDemianenko: you do not need those fail in the end of the test since `ExpectedException` will throw an exception if expected error was not thrown and will fail the test (same for other tests)

MishaDemianenko:
@klaren one small comment, otherwise looks fine

enterprise/kernel/src/test/java/org/neo4j/kernel/enterprise/builtinprocs/SetConfigValueProcedureTest.java
klaren: missed that, correcting

MishaDemianenko (approved):

PR #9818, opened 2017-08-16
Added acceptance tests for variable length queries

pontusmelke:


enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/VarLengthAcceptance.feature
```diff
@@ -0,0 +1,132 @@
...
+      CREATE (n1:A {name: 'n1', blocked: false}),
+             (n2:B {name: 'n2', blocked: false}),
+             (n3:B {name: 'n3', blocked: false}),
+             (n4:C {name: 'n4', blocked: true}),
+             (n1)-[:T {blocked: false}]->(n2),
+             (n2)-[:T {blocked: false}]->(n3),
+             (n3)-[:T {blocked: true}]->(n4)
+      """
+
+  Scenario: Handles checking properties on nodes in path - description 1
```
craigtaverner: Don't we usually use descriptions that describe the test, for example in this case we could replace 'description 1' with 'using in-pattern property value'?

craigtaverner: Actually this entire line could be edited, because this test does not check node properties, but relationship properties.

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/VarLengthAcceptance.feature
```diff
@@ -0,0 +1,132 @@
...
+      MATCH (a)-[:T*]->(c {blocked: false})
+      RETURN c.name
+      """
+    Then the result should be:
+      | c.name |
+      | 'n2'   |
+      | 'n3'   |
+    And no side effects
+
+  Scenario: Handles checking properties on nodes in path - description 2
```
craigtaverner: Perhaps 'description 2' should be 'using ALL() function on path node properties'

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/VarLengthAcceptance.feature
```diff
@@ -0,0 +1,132 @@
...
+      WHERE ALL(n in nodes(p) WHERE n.blocked = false)
+      RETURN c.name
+      """
+    Then the result should be:
+      | c.name |
+      | 'n2'   |
+      | 'n3'   |
+    And no side effects
+
+  Scenario: Handles checking properties on nodes in path - description 3
```
craigtaverner: Perhaps 'description 3' could be be same as previous, but add a phrase about more complex pattern

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/VarLengthAcceptance.feature
```diff
@@ -0,0 +1,132 @@
...
+      WHERE ALL(n in nodes(p) WHERE n.blocked = false)
+      RETURN c.name
+      """
+    Then the result should be:
+      | c.name |
+      | 'n2'   |
+      | 'n3'   |
+    And no side effects
+
+  Scenario: Handles checking properties on relationships in path - description 1
```
craigtaverner: I think we can assume that all descriptions could be updated, right? I'll stop commenting now.

craigtaverner (changes requested):

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/VarLengthAcceptance.feature
pontusmelke: No it doesn't but I agree, I just took the feature file as it was. Will fix naming

PR #9821, opened 2017-08-16
Use automatic port selection in TransactionGuardIntegrationTest

MishaDemianenko:
Use automatic port selection for bolt connectors instead of preliminary allocated


tinwelint:
I must say I have no idea about the validity of this change. Have you run this by someone better versed in the drivers and bolt?

MishaDemianenko:
@tinwelint this is a standard technic of using dynamically allocated port that we use for all our bolt related tests i can take it for myself completely np and will merge it after rebase

PR #9823, opened 2017-08-16
Correct documentation in neo4j.conf for causal_clustering.cluster_rou…

andrewkerr9000:
Correct value for `causal_clustering.cluster_routing_ttl` should be 5 minutes, not 5 seconds. For consistency specify as 300s everywhere.


phughk (approved):
Looks good to me

digitalstain (approved):

PR #9825, opened 2017-08-16
Unflake TestCommunication.serverStopsStreamingToDeadClient by increasing timeout

klaren:


enterprise/com/src/test/java/org/neo4j/com/ClientCrashingWriter.java
```diff
@@ -52,6 +52,7 @@ public void write( ReadableByteChannel data )
                 if ( (totalSize += size) >= crashAtSize )
                 {
                     client.stop();
+                    throw new IOException( "Fake read error" );
```
MishaDemianenko, line 55: should not be here?

klaren: that's a good question, client.stop() is called 1000+ times otherwise, which I believe makes it more falky. From what I can tell it does not change the behavior of the test and this is the only usage of it.

MishaDemianenko (approved):

PR #9827, opened 2017-08-16
Update version of commons-lang to 3.5

Mats-SX:


pom.xml
```diff
@@ -584,7 +584,7 @@
       <dependency>
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-lang3</artifactId>
-        <version>3.3.2</version>
+        <version>3.5</version>
```
MishaDemianenko, line 587: since we do an update, shouldn't we consider latest stable 3.6? or you need a specific version for reasons?

Mats-SX: I'm interested in `3.5` because Spark uses that: https://github.com/apache/spark/blob/master/pom.xml#L173 and we're using the neo4j frontend from our Spark-dependent project.

MishaDemianenko: ok, thx!

Mats-SX:
Build failure is related to downstream `NOTICE.txt` (docs module) not containing the new dependency. Should not hinder this PR from getting through -- discussing options with docs team.

EDIT: Closed by mistake: reopening.

PR #9833, opened 2017-08-17
Performs assertNoSuccessor as an assertion in GBPTree

tinwelint:
because it's somewhat costly when moving around a lot in the tree for
doing updates and have never been a real problem, it's just been added
there as additional confidence from the beginning.


burqen (changes requested):
Changes are approved but build error does not look good at all. Maybe try rebase.

PR #9842, opened 2017-08-17
Added EagerAggregationRegisterPipe

SaschaPeukert:
Also fixed an error in ExpandInto regarding start and enpoints in patterns like (a)-->(a)

### DIFF
--- a/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
+++ b/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
@@ -19,7 +19,6 @@
  */
 package org.neo4j.cypher.internal.compatibility.v3_3.runtime
 
-import org.neo4j.cypher.internal.compiler.v3_3.planner.CantCompileQueryException
 import org.neo4j.cypher.internal.compiler.v3_3.planner.logical.plans._
 import org.neo4j.cypher.internal.frontend.v3_3.ast.Expression
 import org.neo4j.cypher.internal.frontend.v3_3.symbols._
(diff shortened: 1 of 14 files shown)


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/RegisteredPipeBuilder.scala
```diff
@@ -157,6 +159,18 @@ class RegisteredPipeBuilder(fallback: PipeBuilder,
...
+      case Aggregation(_, groupingExpressions, aggregatingExpressions) if aggregatingExpressions.isEmpty =>
+        //TODO: Distinct ?
+        throw new CantCompileQueryException(s"Unsupported logical plan operator: $plan")
+
+      case Aggregation(_, groupingExpressions, aggregatingExpressions) =>
+        EagerAggregationRegisterPipe(
+          source,
+          pipeline,
+          groupingExpressions.keySet,
+          Eagerly.immutableMapValues[String, frontEndAst.Expression, AggregationExpression](aggregatingExpressions, convertExpressions(_).asInstanceOf[AggregationExpression])
```
systay, line 171: It's not clear to me why are we using Eagerly here. Maybe we can add a comment explaining why that is important?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
```diff
@@ -36,21 +35,16 @@ object RegisterAllocation {
     def allocate(lp: LogicalPlan, nullable: Boolean, argument: Option[PipelineInformation]): PipelineInformation = lp match {
       case Aggregation(source, groupingExpressions, aggregationExpressions) =>
         val oldPipeline = allocate(source, nullable, argument)
-        val newPipeline = PipelineInformation.empty
+        val newPipeline = oldPipeline.deepClone()
```
systay, line 38: I don't think this is correct. This means that the output row will contain all the input data and then some more, and that should not be true

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/EagerAggregationRegisterPipe.scala
```diff
@@ -0,0 +1,128 @@
...
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.pipes.{MutableMaps, Pipe, PipeWithSource, QueryState}
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.{ExecutionContext, PipelineInformation}
+import org.neo4j.cypher.internal.compiler.v3_3.planDescription.Id
+
+import scala.collection.mutable.{Map => MutableMap}
+
+// Eager aggregation means that this pipe will eagerly load the whole resulting sub graphs before starting
+// to emit aggregated results.
+// Cypher is lazy until it can't - this pipe will eagerly load the full match
+case class EagerAggregationRegisterPipe(source: Pipe, pipelineInformation: PipelineInformation, keyExpressions: Set[String], aggregations: Map[String, AggregationExpression])
```
systay, line 35: This class needs some work to move more logic from runtime to compile time

systay (changes requested):

systay:
Replaced by #9901

PR #9848, opened 2017-08-18
Add SemiApply to the register runtime

systay:
Plus some minor changes

### DIFF
--- a/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
+++ b/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
@@ -141,6 +141,18 @@ object RegisterAllocation {
         result += (lp -> rhsPipeline)
         rhsPipeline
 
+      case SemiApply(lhs, rhs) =>
+        val lhsPipeline = allocate(lhs, nullable, argument)
+        val rhsPipeline = allocate(rhs, nullable, Some(lhsPipeline.deepClone()))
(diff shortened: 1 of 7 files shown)


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
```diff
@@ -141,6 +141,18 @@ object RegisterAllocation {
         result += (lp -> rhsPipeline)
         rhsPipeline
 
+      case SemiApply(lhs, rhs) =>
+        val lhsPipeline = allocate(lhs, nullable, argument)
+        val rhsPipeline = allocate(rhs, nullable, Some(lhsPipeline.deepClone()))
```
fickludd, line 146: As we discussed it might be possible to omit the `deepClone()`. Worth a shot?

fickludd (approved):
This looks good to me! We could try to remove the deepCloning of lhs in the `RegisterAllocation.allocate()`, but it doesn't have to happen for this to go in. Nice with misc clean-ups.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
systay: I think we need it. The argument being sent in to the `rhs` will be used to "seed" the `rhs` pipeline, but other operators will add registers to the `rhs`-pipeline, and these new allocations should not show up on the `lhs`-pipeline.

PR #9853, opened 2017-08-18
Renamed dbms.allow_format_migration to dbms.allow_upgrade for clarification

klaren:
Configuration option `dbms.allow_format_migration` has been renamed to `dbms.allow_upgrade` since it applies to more than format migration. 

The expected behaviour is that the database should refuse to start if any changes will be made to the store that will prevent the database from starting on a previous version of neo4j.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -125,8 +125,12 @@ public class GraphDatabaseSettings implements LoadableConfig
             "older store version. " +
             "Setting this to `true` does not guarantee successful upgrade, it just " +
             "allows an upgrade to be performed." )
-    public static final Setting<Boolean> allow_store_upgrade = setting( "dbms.allow_format_migration", BOOLEAN,
-            FALSE );
+    @Deprecated
+    @ReplacedBy( "dbms.allow_upgrade" )
+    public static final Setting<Boolean> allow_store_upgrade = setting( "dbms.allow_format_migration", BOOLEAN, FALSE );
+
+    @Description( "Whether to allow an upgrade in case the current version of the database starts against an older version." )
+    public static final Setting<Boolean> allow_upgrade = setting( "dbms.allow_upgrade", BOOLEAN, FALSE );
 
     @Description( "Database record format. Enterprise edition only. Valid values: `standard`, `high_limit`. " +
                   "Default value:  `standard`." )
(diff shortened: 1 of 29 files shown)


community/neo4j/src/test/java/db/DatabaseStartupTest.java
```diff
@@ -81,7 +82,7 @@ public void startTheDatabaseWithWrongVersionShouldFailWithUpgradeNotAllowed() th
             assertTrue( ex.getCause() instanceof LifecycleException );
             assertTrue( ex.getCause().getCause() instanceof UpgradeNotAllowedByConfigurationException );
             assertEquals( "Failed to start Neo4j with an older data store version. To enable automatic upgrade, " +
-                          "please set configuration parameter \"dbms.allow_format_migration=true\"",
```
burqen, line 84: Nice catch!

community/kernel/src/main/java/org/neo4j/kernel/configuration/GraphDatabaseConfigurationMigrator.java
```diff
@@ -100,5 +100,14 @@ public void setValueWithOldSetting( String value, Map<String,String> rawConfigur
                 }
             }
         } );
+        add( new SpecificPropertyMigration( "dbms.allow_format_migration",
+                "dbms.allow_format_migration has been replaced with dbms.allow_upgrade." )
```
burqen, line 104: Nice! Can we also add a test for this?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -125,8 +125,12 @@
...
             "Setting this to `true` does not guarantee successful upgrade, it just " +
             "allows an upgrade to be performed." )
-    public static final Setting<Boolean> allow_store_upgrade = setting( "dbms.allow_format_migration", BOOLEAN,
-            FALSE );
+    @Deprecated
+    @ReplacedBy( "dbms.allow_upgrade" )
+    public static final Setting<Boolean> allow_store_upgrade = setting( "dbms.allow_format_migration", BOOLEAN, FALSE );
+
+    @Description( "Whether to allow an upgrade in case the current version of the database starts against an older version." )
+    public static final Setting<Boolean> allow_upgrade = setting( "dbms.allow_upgrade", BOOLEAN, FALSE );
```
burqen, line 133: Have you looked into what implications this change has for docs? Let's make sure that is handled before merging.

burqen (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/configuration/GraphDatabaseConfigurationMigrator.java
klaren: @burqen We definitely should, new commit available!

burqen (approved):

burqen:
@klaren Approved! Don't know if the docs_compile error is related. It does not look like it, but worth checking with docs team before merging.

PR #9857, opened 2017-08-21
Removed legacy log files from 2.2 and backwards

klaren:
The old legacy format for log files that had a version header in the file can be removed since 2.2 is no longer supported. The implication of this is that 3.3 can only read log files back to 2.3 during migration.


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/VersionAwareLogEntryReader.java
```diff
@@ -70,24 +69,12 @@ public LogEntry readLogEntry( SOURCE channel ) throws IOException
...
-                        // if the read type is negative than it is actually the log entry version
-                        // so we need to read an extra byte which will contain the type
-                        versionCode = typeCode;
-                        typeCode = channel.get();
-                    }
-                }
-                catch ( ReadPastEndException e )
-                {   // Make these exceptions slip by straight out to the outer handler
-                    throw e;
+                    return null;
```
tinwelint, line 75: Did this method previously also explicitly return `null` on `versionCode` `0`? I can't see it straight away by looking at the existing code.

tinwelint (commented):
Nice to get rid of all those old things, well done.

I have a question about a zero version code in `VersionAwareLogEntryReader`, otherwise things look good methinks.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/VersionAwareLogEntryReader.java
klaren: Not explicitly. The zero-version mapped to the old `LogEntryParsersV2_0` which mapped zero-type to `null`. This seems to be needed according to the test `shouldParseStreamOfZerosAsEmptyLogEntries()` in `VersionAwareLogEntryReaderTest`.

tinwelint: OK cheers

tinwelint (approved):

PR #9860, opened 2017-08-21
Replication benchmark procedure

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/BenchmarkResult.java
```diff
@@ -0,0 +1,39 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.causalclustering.core.replication;
+
+@SuppressWarnings( "WeakerAccess" )
+public class BenchmarkResult
+{
+    public Long totalRequests;
+    public Long totalBytes;
+    public Long timeMillis;
```
andrewkerr9000, line 27: Would a `java.time.Duration` be better than a `Long` for storing the time? You may also wish to make the fields `final` - and add `equals` and `hashcode`.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/BenchmarkProcedure.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        long runTime = System.currentTimeMillis() - startTime;
+
+        long totalRequests = 0;
+        long totalBytes = 0;
+
+        for ( Worker worker : workerArr )
+        {
+            totalRequests += worker.totalRequests;
```
andrewkerr9000: You could do `long totalRequests = Arrays.stream ( workerArr ).mapToLong( Worker::getTotalRequests ).sum();`, which would require the getter. Or you could do `BenchmarkResult result = Arrays.stream ( workerArr ).reduce( ... );`, but that may be less readable.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/BenchmarkProcedure.java
```diff
@@ -0,0 +1,157 @@
...
+public class BenchmarkProcedure
+{
+    @Context
+    public Replicator replicator;
+
+    @Context
+    public Log log;
+
+    private static long startTime;
+    private static Worker[] workerArr;
```
andrewkerr9000: Why an array rather than a list?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/CoreStateMachines.java
```diff
@@ -181,6 +200,15 @@ public void dispatch( ReplicatedLockTokenRequest lockRequest, long commandIndex,
         }
 
         @Override
+        public void dispatch( DummyRequest dummyRequest, long commandIndex, Consumer<Result> callback )
+        {
+            if ( benchmarkMachine != null )
+            {
+                benchmarkMachine.applyCommand( dummyRequest, commandIndex, callback );
+            }
```
andrewkerr9000: If `benchmarkMachine` may be null, then perhaps it should be an `Optional<BenchmarkMachine>`

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/CoreReplicatedContentMarshal.java
```diff
@@ -85,6 +87,11 @@ else if ( content instanceof DistributedOperation )
             channel.put( DISTRIBUTED_OPERATION );
             ((DistributedOperation) content).serialize( channel );
         }
+        else if ( content instanceof DummyRequest )
```
andrewkerr9000, line 90: Is there a reason we don't want to call a method on `ReplicatedContent`? If so could we use a visitor pattern instead?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/CoreReplicatedContentMarshal.java
```diff
@@ -119,6 +126,9 @@ public ReplicatedContent unmarshal0( ReadableChannel channel ) throws IOExceptio
             case DISTRIBUTED_OPERATION:
                 content = DistributedOperation.deserialize( channel );
                 break;
+            case DUMMY_REQUEST:
+                content = DummyRequest.Marshal.INSTANCE.unmarshal( channel );
+                break;
```
andrewkerr9000, line 131: We could refactor the byte identifiers to be an `enum` with the byte value as a field, and an `unmarshal` method. This switch block could then go away.

andrewkerr9000 (commented):
Looks good, but I did make a couple of style related comments. I also made some suggestions to make `CoreReplicatedContentMarshal` safer, but that's not code introduced in this commit.

Looks fine functionally.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/BenchmarkResult.java
martinfurmanski: Iirc, these are all limitations put by the procedure system, because they use reflection and only support a limited set of types.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/BenchmarkProcedure.java
martinfurmanski: No special reason really. I know the exact size of it and I don't plan to change it.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/CoreStateMachines.java
martinfurmanski: I think that is generally frowned upon though.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/CoreReplicatedContentMarshal.java
martinfurmanski: I think that could make a lot of sense! A lot of this code is quite old and could use a refactoring.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/CoreReplicatedContentMarshal.java
martinfurmanski: +1

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/BenchmarkProcedure.java
andrewkerr9000: Not a big deal at all, but I dislike Java arrays for reasons of syntax/API and `ArrayStoreException`. It is possible to create an `ArrayList` with an initial capacity.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/BenchmarkProcedure.java
andrewkerr9000: Or `long totalRequests = Arrays.stream ( workerArr ).mapToLong( worker -> worker.totalRequests ).sum()` which wouldn't require changes to `Worker`.

Stream all the things! :smile:

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/CoreStateMachines.java
andrewkerr9000: If `benchmarkMachine` should never be null then I'd get rid of the null check and let it bomb with an NPE if it's not set up correctly.

PR #9862, opened 2017-08-21
Add feature toggle for stronger index entry hashing

chrisvest:
We can alternatively make it an `unsupported.` setting, and pass its value to the `ConstraintEnforcingEntityOperations`, which could then make the decision about what hash function to use. That approach would have the advantage that we could change the default value depending on the edition, such that only HA setups would keep using the old hash function.

### DIFF
--- a/community/common/src/main/java/org/neo4j/hashing/IncrementalXXH64.java
+++ b/community/common/src/main/java/org/neo4j/hashing/IncrementalXXH64.java
@@ -20,29 +20,12 @@
 package org.neo4j.hashing;
 
 /**
- * The (our) incremental XXH64 hash function.
- * <p>
- * This hash function is based on xxHash (XXH64 variant), but modified to work incrementally on 8-byte blocks
- * instead of on 32-byte blocks. Basically, the 32-byte block hash loop has been removed, so we use the 8-byte
- * block tail-loop for the entire input.
- * <p>
- * This hash function is roughly twice as fast as its predecessor, about 30% faster than optimised murmurhash3
- * implementations though not as fast as optimised xxHash implementations due to the smaller block size.
- * It is allocation free, unlike its predecessor. And it retains most of the excellent statistical properties of
- * xxHash, failing only the "TwoBytes" and "Zeroes" keyset tests in SMHasher, passing 12 out of 14 tests.
- * <p>
- * The hasher is used by first {@link IncrementalXXH64#init(long) initialising} the state with a seed, which can be any
- * arbitrary 64-bit integer, including zero. The produced state is then recursively cycled through the
- * {@link IncrementalXXH64#update(long, long) update} function and merged with an input word as many times as desired.
- * When the input has been consumed, the final hash value is then produced from the last intermediate hash value with
- * the {@link IncrementalXXH64#finalise(long) finalise} function.
- * <p>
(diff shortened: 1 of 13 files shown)


chrisvest:
Build is green. Github just missed the signal from the Tools Compile build.

community/common/src/main/java/org/neo4j/hashing/IncrementalXXH64.java
```diff
@@ -0,0 +1,81 @@
...
+package org.neo4j.hashing;
+
+/**
+ * The (our) incremental XXH64 hash function.
+ * <p>
+ * This hash function is based on xxHash (XXH64 variant), but modified to work incrementally on 8-byte blocks
+ * instead of on 32-byte blocks. Basically, the 32-byte block hash loop has been removed, so we use the 8-byte
+ * block tail-loop for the entire input.
+ * <p>
+ * This hash function is roughly twice as fast as its predecessor, about 30% faster than optimised murmurhash3
```
MishaDemianenko: not really clear which predecessor we are talking about here

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/ResourceTypes.java
```diff
@@ -71,35 +78,52 @@ public WaitStrategy waitStrategy()
         return waitStrategy;
     }
 
+    /**
+     * The schema index entry hashing method used up until 2.2.0.
```
MishaDemianenko: That's not that really so right since `LegacyIndexProxy` still use it now?

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/ResourceTypes.java
```diff
@@ -135,4 +159,88 @@ public static ResourceType fromId( int typeId )
...
+    /**
+     * This is a stronger, full 64-bit hashing method for schema index entries that we should use by default in a
+     * future release, where we will also upgrade the HA protocol version. Currently this is indicated by the "4_x"
+     * name suffix, but any version where the HA protocol version changes anyway would be just as good an opportunity.
+     *
+     * @see IncrementalXXH64
+     */
+    public static long indexEntryResourceId_4_x( long labelId, IndexQuery.ExactPredicate... predicates )
+    {
+        long hash = IncrementalXXH64.init( 0x0123456789abcdefL );
```
MishaDemianenko: We already have HashFunction defined that is used in a HopScotch, i would say that IncrementalXXH64 should be changed to be out of the possible HashFunctions and they should be moved to commons module most probably, and that interface can be extended to include `update(hash, block)` and `finalise(hash)` so new hashing function will not gonna be glued that much to ResourceTypes

MishaDemianenko (changes requested):
Couple of small comments otherwise looks good!

community/kernel/src/main/java/org/neo4j/kernel/impl/locking/ResourceTypes.java
chrisvest: Oh, I missed that.

chrisvest:
@MishaDemianenko I believe I've addressed your comments. Please have another look.

MishaDemianenko (approved):
Looks great now!

PR #9864, opened 2017-08-21
Check only first property for uniqueness during uniqueness constraint verification

MishaDemianenko:
Before we were checking all the fields of index for uniqueness
when we were about to create uniqueness constraint. That was fine for
one property index but n case of the composite index of N properties
we would actually do verification N times while checking combinations
for first property only is more than enough.


MishaDemianenko:
@lutovich @craigtaverner can you guys please take a look, thx!

ragadeeshu (approved):

lutovich (approved):

PR #9865, opened 2017-08-21
dbms.killQueries now shows if it failed to kill a query

eebus:


community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -567,6 +567,10 @@
     public static final Setting<String> procedure_unrestricted =
             setting( "dbms.security.procedures.unrestricted", Settings.STRING, "" );
 
+    @Description("If dbms.killQueries give a verbose output, with information about whitch querys where not found." )
```
pontusmelke: Do we really need a setting for this, couldn't it just always be verbose?

pontusmelke: `whitch` -> `which`

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
```diff
@@ -312,17 +314,21 @@ private boolean isAdminProcedure( String procedureName )
...
         securityContext.assertCredentialsNotExpired();
         try
         {
             long queryId = fromExternalString( idText ).kernelQueryId();
 
-            return getActiveTransactions( tx -> executingQueriesWithId( queryId, tx ) )
-                .map( catchThrown( InvalidArgumentsException.class, this::killQueryTransaction ) );
-         }
+            Set<Pair<KernelTransactionHandle,ExecutingQuery>> querys = getActiveTransactions( tx -> executingQueriesWithId( queryId, tx ) ).collect( toSet() );
+            Config configs = resolver.resolveDependency( Config.class );
```
pontusmelke: We only need to find the setting once, right? Do not need to do it for each call to `killQuery`?

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/EnterpriseBuiltInDbmsProcedures.java
```diff
@@ -332,20 +338,30 @@ private boolean isAdminProcedure( String procedureName )
...
-                .collect( toSet() );
 
-            return getActiveTransactions( tx -> executingQueriesWithIds( queryIds, tx ) )
-                .map( catchThrown( InvalidArgumentsException.class, this::killQueryTransaction ) );
+            Set<Long> queryIds = idTexts.stream().map( catchThrown( InvalidArgumentsException.class, QueryId::fromExternalString ) ).map(
+                    catchThrown( InvalidArgumentsException.class, QueryId::kernelQueryId ) ).collect( toSet() );
+
+            Set<QueryTerminationResult> terminatedQuerys = getActiveTransactions( tx -> executingQueriesWithIds( queryIds, tx ) ).map(
+                    catchThrown( InvalidArgumentsException.class, this::killQueryTransaction ) ).collect( toSet() );
+            Config configs = resolver.resolveDependency( Config.class );
```
pontusmelke: Same here, this needs could be done once

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
eebus: The intent is to not change old behavior in a patch release for 3.2. In the forward merge, this config should be flipped to support the verbose as the default.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -567,6 +567,10 @@
     public static final Setting<String> procedure_unrestricted =
             setting( "dbms.security.procedures.unrestricted", Settings.STRING, "" );
 
+    @Description("If dbms.killQueries give a verbose output, with information about which querys where not found." )
```
OliviaYtterbrink: This description could be better, maybe `Specifies whether or not dbms.killQueries produces a verbose output, with information about which queries were not found`?

OliviaYtterbrink (approved):

PR #9866, opened 2017-08-21
GetServers returns list not map

pontusmelke:
The signature of the `getServers` procedures are wrong, they return a list
of maps and not just a map.

NOTE: This has been fixed in 3.3 so please do a null merge.


martinfurmanski:
@pontusmelke Where does this become an actual issue? Are you adding result type checking?

pontusmelke:
@martinfurmanski it has become a real issue in 3.3 where we are more strict with types, in 3.2 I guess it is mostly an error if you do inspections and list procedures etc. It is already fixed in 3.3 but I still thought it might be valuable to backport.

martinfurmanski (approved):

PR #9868, opened 2017-08-22
Reverted semi apply

eebus:


fickludd (approved):
As the merge implementation of SemiApply turned out to be broken, we temporarily revert the Slotted SemiApply in order to unbreak downstream builds.

PR #9870, opened 2017-08-22
Cluster overview procedure should not return array

pontusmelke:
The procedure `dbms.cluster.overview` was creating arrays instead of lists and arrays
are not a supported return type of procedures.


fickludd (approved):
Looks good

PR #9872, opened 2017-08-22
3.2 Enable backups over CC transaction protocol

phughk:
Currently we have a separate protocol for backups. This protocol is used in HA but also in the backup tool (neo4j-admin).
Introducing this change enables `neo4j-admin backup` to work with CC endpoints without any change of configuration.


enterprise/backup/src/main/java/org/neo4j/backup/nextgen/Application.java
```diff
@@ -0,0 +1,96 @@
+/*
```
phughk: TODO delete this

enterprise/backup/src/main/java/org/neo4j/backup/nextgen/BackupModule.java
```diff
@@ -0,0 +1,146 @@
...
+
+    private static PageCache aPageCache( FileSystemAbstraction fileSystemAbstraction )
+    {
+        PageCursorTracerSupplier pageCursorTracerSupplier = new DefaultPageCursorTracerSupplier();
+        PageCacheTracer pageCacheTracer = new DefaultPageCacheTracer();
+        int cachePageSize = 8192;
+        int maxPages = 1024;
+        PageSwapperFactory pageSwapperFactory = new SingleFilePageSwapperFactory();
+        pageSwapperFactory.setFileSystemAbstraction( fileSystemAbstraction );
+        return new MuninnPageCache( pageSwapperFactory, maxPages, cachePageSize, pageCacheTracer, pageCursorTracerSupplier );
```
chrisvest: Are you intentionally ignoring the page cache configurations in `neo4j.conf`? In doing so, we can no longer do an online backup directly to block device storage, but have to change the procedure to back up to file storage and then import onto block device storage.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/TrackingResponseHandler.java
```diff
@@ -127,6 +127,7 @@ public void onCoreSnapshot( CoreSnapshot coreSnapshot )
 
     Optional<Long> lastResponseTime()
     {
+        System.out.println( "lastResponseTime = " + lastResponseTime );
```
chrisvest: ಠ_ಠ

enterprise/backup/src/main/java/org/neo4j/backup/nextgen/BackupModule.java
phughk: Not intentionally - I will fix that, thanks!

tinwelint:
One thing which makes it a bit harder to read these changes is the lack of javadocs. Class-level docs and on some important methods would do wonders. Cheers.

enterprise/backup/src/main/java/org/neo4j/backup/BackupStrategyWrapper.java
```diff
@@ -0,0 +1,102 @@
...
+    {
+        this.backupStrategyImplementation = backupStrategyImplementation;
+        this.backupCopyService = backupCopyService;
+    }
+
+    public BackupState doBackup( OnlineBackupContext onlineBackupContext )
+    {
+        wrappedAsRuntimeException( backupStrategyImplementation::init );
+        wrappedAsRuntimeException( backupStrategyImplementation::start );
+        BackupState state = performBackupWithoutLifecycle( onlineBackupContext );
```
tinwelint: You could use `LifeSupport` to take a LifeCycle through its stages instead of doing this wrapping and manual calls, which also doesn't gracefully handles failures along the way.

enterprise/backup/src/main/java/org/neo4j/backup/StoreCopyService.java
```diff
@@ -0,0 +1,79 @@
...
+    long retrieveStore( StoreFileStreams storeFileStreams, StoreId expectedStoreId ) throws StoreCopyFailedException
+    {
+        long latestTxNumber = storeCopyClient.copyStoreFiles( fromAddress, expectedStoreId, storeFileStreams );
+        try
+        {
+            storeFileStreams.close();
+        }
+        catch ( Exception e )
+        {
+            throw new RuntimeException( e );
```
tinwelint: `Exceptions.launderedException` instead? This would not wrap `RuntimeException`s one more time.

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/DefaultPageCursorTracerSupplier.java
```diff
@@ -25,7 +25,7 @@
 
     public static final DefaultPageCursorTracerSupplier INSTANCE = new DefaultPageCursorTracerSupplier();
 
-    private DefaultPageCursorTracerSupplier()
+    public DefaultPageCursorTracerSupplier()
```
chrisvest: You can just grab the `INSTANCE` where you use this, instead of making the constructor public and creating a new one.

community/kernel/src/main/java/org/neo4j/kernel/configuration/OnlineBackupSettings.java
```diff
@@ -5,19 +5,19 @@
...
- * GNU Affero General Public License for more details.
+ * GNU General Public License for more details.
  *
- * You should have received a copy of the GNU Affero General Public License
- * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-package org.neo4j.backup;
+package org.neo4j.kernel.configuration;
```
chrisvest: Why is it moved to community? Could it not be in, say, enterprise-kernel?

enterprise/backup/src/main/java/org/neo4j/backup/BackupCommandArgumentHandler.java
```diff
@@ -0,0 +1,150 @@
...
+                    "Configuration file to supply additional configuration in. This argument is DEPRECATED." ) )
+            .withArgument( new OptionalBooleanArg( "cc-graph", true,
+                    "Perform consistency checks between nodes, relationships, properties, types and tokens." ) )
+            .withArgument( new OptionalBooleanArg( "cc-indexes", true,
+                    "Perform consistency checks on indexes." ) )
+            .withArgument( new OptionalBooleanArg( "cc-label-scan-store", true,
+                    "Perform consistency checks on the label scan store." ) )
+            .withArgument( new OptionalBooleanArg( "cc-property-owners", false,
+                    "Perform additional consistency checks on property ownership. This check is *very* expensive in " +
+                            "time and memory." ) );
```
chrisvest: I think the argument names should be extract to constants, so their definition here won't diverge from the code that looks for them to interpret their meaning. The list of constants would also serve as an easier-to-read overview of what options are available.

chrisvest: I think it would also be desirable to have an option for overriding the page cache memory setting. People run the backup tool on machines that are already running Neo4j instances, more often than they should.

chrisvest: And obviously a page cache size override argument should apply both to the page cache used for the store copy, and the one used for the consistency check.

enterprise/backup/src/main/java/org/neo4j/backup/BackupStageOutcome.java
```diff
@@ -0,0 +1,28 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.backup;
+
+public enum BackupStageOutcome
+{
+    SUCCESS,
+    WRONG_PROTOCOL,
+    TRUE_FAILURE,
+    ABSOLUTE_FAILURE;
```
chrisvest: Better names and javadocs would be nice. `SUCCESS` makes sense, and `WRONG_PROTOCOL` you can infer is a failure condition, and you even get an idea about what the problem is. But `TRUE_FAILURE` and `ABSOLUTE_FAILURE` are not very helpful, though.

chrisvest (changes requested):
I'm only a quarter through it, but I'll give you these comments already since it looks like you are still actively working on it.

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/DefaultPageCursorTracerSupplier.java
phughk: Addressed

enterprise/backup/src/main/java/org/neo4j/backup/StoreCopyService.java
phughk: Addressed

enterprise/backup/src/main/java/org/neo4j/backup/BackupStrategyWrapper.java
phughk: Addressed

community/kernel/src/main/java/org/neo4j/kernel/configuration/OnlineBackupSettings.java
phughk: Addressed

enterprise/backup/src/main/java/org/neo4j/backup/BackupCommandArgumentHandler.java
phughk: Addressed - please have a look at `BackupSupportingClassesFactory` and `HaBackup` for how this has been handled

enterprise/backup/src/main/java/org/neo4j/backup/BackupStageOutcome.java
phughk: kept `TRUE_FAILURE` as is, changed `ABSOLUTE_FAILURE` to `UNRECOVERABLE_FAILURE`, let me know if you have a better suggestion - the idea is that a true failure lets the surrounding class know that what it performed wasn't successful, but it can still try different methods (such as fallback from incremental to full, or fallback from full backup to another strategy). `UNRECOVERABLE_FAILURE` on the other hand means that there was a serious error and there cannot be any fallback.

enterprise/backup/src/main/java/org/neo4j/backup/BackupDelegator.java
```diff
@@ -0,0 +1,95 @@
...
+import org.neo4j.causalclustering.catchup.storecopy.StoreIdDownloadFailedException;
+import org.neo4j.causalclustering.identity.StoreId;
+import org.neo4j.helpers.AdvertisedSocketAddress;
+import org.neo4j.io.fs.FileSystemAbstraction;
+
+/**
+ * Simplifies the process of performing a backup over the transaction protocol by wrapping all the necessary classes
+ * and delegating methods to the correct instances.
+ */
+public class BackupDelegator
```
martinfurmanski: This is not lifecycled and I can't see how catchupClient is used in here.

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupContextLoader.java
```diff
@@ -0,0 +1,62 @@
...
+     * @param commandlineArgs Command line arguments split into an array by space
+     * @return the resolved configuration and arguments used in controlling the behaviour of the backup tool
+     */
+    public OnlineBackupContext fromCommandLineArguments( String[] commandlineArgs )
+    {
+        try
+        {
+            OnlineBackupRequiredArguments requiredArguments = backupCommandArgumentHandler.establishRequiredArguments( commandlineArgs );
+            Config config = onlineBackupCommandConfigLoader.loadConfig( requiredArguments.getAdditionalConfig() );
+            CheckConsistencyConfig checkConsistencyConfig = backupCommandArgumentHandler.establishConfigurationOverride( config );
```
martinfurmanski: Perhaps rename?

enterprise/backup/src/main/resources/META-INF/services/org.neo4j.configuration.LoadableConfig
```diff
@@ -1 +0,0 @@
-org.neo4j.backup.OnlineBackupSettings
```
martinfurmanski, line 1: delete file?

enterprise/backup/src/test/java/org/neo4j/backup/BackupFlowTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+        // then we want a predictable exception (instead of NullPointer)
+        expectedException.expect( CommandFailed.class );
+        expectedException.expectMessage( "Failed to run a backup using the available strategies. Perhaps using the wrong port?" );
+
+        // when
+        subject.performBackup( onlineBackupContext );
+    }
+
+    private void anyStrategyPasses()
```
martinfurmanski: Add annotation or delete test?

enterprise/backup/src/main/resources/META-INF/services/org.neo4j.configuration.LoadableConfig
phughk: Just checked, file was indeed deleted :+1:

enterprise/backup/src/test/java/org/neo4j/backup/BackupFlowTest.java
phughk: As it turns out, this is used in the above tests to simplify configuring strategy mocks.
I have added a minor java-doc to this method.

enterprise/backup/src/main/java/org/neo4j/backup/BackupCommandArgumentHandler.java
```diff
@@ -0,0 +1,172 @@
...
+        boolean checkGraph;
+        boolean checkIndexes;
+        boolean checkLabelScanStore;
+        boolean checkPropertyOwners;
+        try
+        {
+            // We can remove the loading from config file in 4.0
+            if ( arguments.has( "cc-graph" ) )
+            {
+                checkGraph = arguments.getBoolean( "cc-graph" );
```
chrisvest: Many opportunities to use the new constants here.

enterprise/backup/src/main/java/org/neo4j/backup/BackupCommandArgumentHandler.java
```diff
@@ -0,0 +1,172 @@
...
+        {
+            throw new IncorrectUsage( e.getMessage() );
+        }
+    }
+
+    OnlineBackupRequiredArguments establishRequiredArguments( String... args ) throws IncorrectUsage
+    {
+        try
+        {
+            OptionalHostnamePort address = toOptionalHostnamePortFromRawAddress( arguments.parse( args ).get( "from" ) );
```
chrisvest: You can use the constant for `from`

enterprise/backup/src/main/java/org/neo4j/backup/BackupCommandArgumentHandler.java
```diff
@@ -0,0 +1,172 @@
...
+            Path folder = arguments.getMandatoryPath( ARG_NAME_BACKUP_DIRECTORY );
+            String name = arguments.get( ARG_NAME_BACKUP_NAME );
+            boolean fallbackToFull = arguments.getBoolean( ARG_NAME_FALLBACK_FULL );
+            boolean doConsistencyCheck = arguments.getBoolean( ARG_NAME_CHECK_CONSISTENCY );
+            long timeout = arguments.get( ARG_NAME_TIMEOUT, TimeUtil.parseTimeMillis );
+            Optional<Path> additionalConfig = arguments.getOptionalPath( ARG_NAME_ADDITIONAL_CONFIG_DIRECTORY );
+            Path reportDir = arguments.getOptionalPath( ARG_NAME_REPORT_DIRECTORY )
+                    .orElseThrow( () -> new IllegalArgumentException( "cc-report-dir must be a path" ) );
```
chrisvest: The exception message can use the relevant constant

enterprise/backup/src/main/java/org/neo4j/backup/BackupCopyService.java
```diff
@@ -0,0 +1,104 @@
...
+    public BackupCopyService( OutsideWorld outsideWorld )
+    {
+        this.outsideWorld = outsideWorld;
+    }
+
+    public void moveBackupLocation( File oldLocation, File newLocation ) throws CommandFailed
+    {
+        try
+        {
+            outsideWorld.fileSystem().renameFile( oldLocation, newLocation );
```
chrisvest, line 49: This won't fly on block device. Shame we didn't notice this on the original implementation you copied this from.

phughk: If there is a simple work around I can address it still, that shouldn't be an issue

enterprise/backup/src/main/java/org/neo4j/backup/BackupCommandArgumentHandler.java
phughk: Addressed in next push

enterprise/backup/src/main/java/org/neo4j/backup/BackupCommandArgumentHandler.java
phughk: Addressed in next push

enterprise/backup/src/main/java/org/neo4j/backup/BackupCommandArgumentHandler.java
phughk: Addressed in next push

enterprise/backup/src/main/java/org/neo4j/backup/BackupFlowFactory.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    BackupFlow backupFlow( OnlineBackupContext onlineBackupContext, HaBackup haBackup, BackupDelegator backupDelegator )
+    {
+        ProgressMonitorFactory progressMonitorFactory = ProgressMonitorFactory.textual( outsideWorld.errorStream() );
+
+        List<BackupStrategyWrapper> strategies = Stream
+            .of(
+                new CausalClusteringBackupStrategy( backupDelegator, addressResolutionHelper ),
+                new HaBackupStrategy( haBackup, addressResolutionHelper, onlineBackupContext.getRequiredArguments().getTimeout() ) )
```
chrisvest: The HaBackupStrategy also works when we take online backup of single-instance Enterprise?

phughk: Yes, it is using the backup protocol.

enterprise/backup/src/main/java/org/neo4j/backup/BackupStrategyInterface.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.neo4j.backup;
+
+import java.io.File;
+
+import org.neo4j.helpers.OptionalHostnamePort;
+import org.neo4j.kernel.configuration.Config;
+import org.neo4j.kernel.lifecycle.Lifecycle;
+
+public interface BackupStrategyInterface extends Lifecycle
```
chrisvest: This doesn't need to outright say `Interface` in the name.

enterprise/backup/src/main/java/org/neo4j/backup/BackupStrategyWrapper.java
```diff
@@ -0,0 +1,127 @@
...
+import java.io.File;
+
+import org.neo4j.commandline.admin.CommandFailed;
+import org.neo4j.helpers.OptionalHostnamePort;
+import org.neo4j.kernel.configuration.Config;
+import org.neo4j.kernel.lifecycle.LifeSupport;
+
+class BackupStrategyWrapper
+{
+    private final BackupStrategyInterface backupStrategyImplementation;
```
chrisvest: And this field (and associated parameter) doesn't really need to say `Implementation`.

phughk: Addressed in next push

enterprise/backup/src/main/java/org/neo4j/backup/BackupStrategyInterface.java
phughk: Addressed in next push

enterprise/backup/src/main/java/org/neo4j/backup/BackupCopyService.java
chrisvest: The `org.neo4j.com.storecopy.StoreCopyClient` has an example of how it could be done. But we are clearly also missing a block device test for the `neo4j-admin backup` command.

chrisvest: The key is figuring out which files are stored on the block device, and then moving those files via the `FileMoveAction.copyViaPageCache`.

enterprise/backup/src/main/java/org/neo4j/backup/HaBackup.java
```diff
@@ -160,12 +142,11 @@ private BackupOutcome fullBackup( FileSystemAbstraction fileSystem, String sourc
...
         if ( !directoryIsEmpty( fileSystem, targetDirectory ) )
         {
-            throw new RuntimeException( "Can only perform a full backup into an empty directory but " +
-                    targetDirectory + " is not empty" );
+            throw new RuntimeException( "Can only perform a full backup into an empty directory but " + targetDirectory + " is not empty" );
         }
         long timestamp = System.currentTimeMillis();
         long lastCommittedTx = -1;
-        try ( PageCache pageCache = createPageCache( fileSystem, tuningConfiguration ) )
+        try ( PageCache pageCache = this.pageCacheOptional.orElse( createPageCache( fileSystem, tuningConfiguration ) ))
```
chrisvest: I think checkstyle in 3.3. will complain about the lack of a space at the end, there.

enterprise/backup/src/main/java/org/neo4j/backup/IdGeneratorWrapper.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.neo4j.backup;
+
+import java.io.File;
+import java.io.IOException;
+
+import org.neo4j.io.fs.FileSystemAbstraction;
+import org.neo4j.kernel.impl.store.id.IdGeneratorImpl;
+
+public class IdGeneratorWrapper
```
chrisvest: This class is awkward because it has no state and no overrides. I would have tested `ClearIdService` with an ephemeral file system instead.

phughk: Do you think it's necessary to change in this PR?

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
```diff
@@ -0,0 +1,228 @@
...
+    @Before
+    public void initialiseBackupDirectory()
+    {
+        backupDir = testDirectory.directory( "backups" );
+    }
+
+    @Test
+    public void backupCanBePerformedOverCcWithDefaultPort() throws Exception
+    {
+        assumeFalse( SystemUtils.IS_OS_WINDOWS );
```
chrisvest, line 101: Why do all these tests skip Windows?

phughk: That's what was in the original `OnlineBackupCommandIT` - not sure why

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
```diff
@@ -0,0 +1,228 @@
...
+                runBackupToolFromOtherJvmToGetExitCode( "--from", customAddress,
+                        "--cc-report-dir=" + backupDir,
+                        "--backup-dir=" + backupDir,
+                        "--name=defaultport" ) );
+        assertEquals( DbRepresentation.of( clusterDatabase( cluster ) ), getBackupDbRepresentation( "defaultport" ) );
+    }
+
+    @Test
+    @Ignore( "Need to resolved this - causal clustering should have the backup protocol disabled" )
+    public void backupCanNotBePerformedOverHa() throws Exception
```
chrisvest: The test does not look like it's starting an HA cluster.

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandConfigLoaderTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+import static java.lang.String.format;
+import static org.hamcrest.Matchers.any;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+public class OnlineBackupCommandConfigLoaderTest
+{
+    @Rule
+    public TemporaryFolder testDirectory = new TemporaryFolder();
```
chrisvest: Why `TemporaryFolder` instead of `TestDirectory`?

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
phughk: The test starts a causal cluster. There are 2 separate IT classes for OnlineBackupCommand:
OnlineBackupCommand**Cc**IT and OnlineBackupCommand**Ha**IT
This test is ignored, because we are expecting the new behaviour for CC to not have the backup protocol enabled (at all). therefore this test should fail.

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandConfigLoaderTest.java
chrisvest: I think we really should use `TestDirectory` because that will put the data in `target/test-data`, which on our build agents is more likely to be mounted on a volume that can fit the data, than `TMPDIR` is.

phughk: Addressed in next commit

enterprise/backup/src/test/java/org/neo4j/backup/util/TestHelpers.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+        catch ( Exception e )
+        {
+            if ( !exceptionClass.isInstance( e ) )
+            {
+                throw new RuntimeException( format( "Exception %s is not of type %s", e.getClass().getName(), exceptionClass.getName() ), e );
+            }
+            return (E) e;
+        }
+        throw new RuntimeException( "The code expected to fail hasn't failed" );
```
chrisvest: Maybe you want to throw `AssertionError`s instead.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
```diff
@@ -136,29 +132,6 @@ public CoreServerModule( IdentityModule identityModule, final PlatformModule pla
...
-                    servicesToStopOnStoreCopy.remove( pickBackupExtension( dataSource ) );
-                }
-
-                private OnlineBackupKernelExtension pickBackupExtension( NeoStoreDataSource dataSource )
-                {
-                    return dataSource.getDependencyResolver().resolveDependency( OnlineBackupKernelExtension.class );
-                }
-            } );
-        }
-
```
chrisvest, line 161: Is it always enabled, and the extension is registered automatically?

enterprise/backup/src/test/java/org/neo4j/backup/util/TestHelpers.java
phughk: Addressed in next push

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
phughk: It should always be disabled for CC (there is no need in maintaining it, since the neo4j-admin will be the same version as the cc running). But you are saying it is registered automatically? That is what I need to avoid (`@Ignore` in OnlineBackupCommandCcIt)

enterprise/kernel/src/test/java/org/neo4j/util/JvmRunner.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.neo4j.commandline.admin.AdminTool;
+import org.neo4j.io.proc.ProcessUtil;
+import org.neo4j.test.ProcessStreamHandler;
+import org.neo4j.test.rule.TestDirectory;
+
+public class JvmRunner
```
chrisvest: Seems pretty backup specific. Maybe just add them as more static methods to `org.neo4j.backup.util.TestHelpers`?

chrisvest (changes requested):
Did a second round of review. Still nits to pick.

enterprise/kernel/src/test/java/org/neo4j/util/JvmRunner.java
phughk: Because it contains `@Rule TestDirectory` then it is expected to be extended from. The method that doesn't require `NEO4J_HOME` depends on the TestDirectory. Also it had to be moved to enterprise/kernel because there are tests in causal clustering (if I recall) that use this as well. Its a maven tree thing essentially

chrisvest: `OnlineBackupCommandCcIT` is the only test that uses that feature of it, and it could just as well have its own `TestDirectory`.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
phughk: Resolved - CC (core+RR) no longer have the backup protocol

enterprise/backup/src/main/java/org/neo4j/backup/BackupCopyService.java
phughk: Not addressed - will create a new ticket afterwards.

chrisvest:
You got a couple of compilation failures.

chrisvest (approved):

phughk:
Conflicts are being resolved in linked PR

PR #9874, opened 2017-08-22
Add test support for 3.2 community.

sherfert:


alexaverbuch (dismissed):
'Version3_2'
 --> add more scenarios to cover all 3.2 combinations of planner/runtime

'Compiled'
 --> this is confusing, e.g., 'Compiled' only contains 3.3 scenarios, and there is no way to get all 'Compiled' combinations (e.g., with 3.2)
 --> same problem with 'CompiledSource' & 'CompiledByteCode': TestConfiguration = Scenarios.CompiledByteCode3_3'

'AbsolutelyAll'
 --> does not include AbsolutelyAll, e.g., 3.2 compiled combinations

PR #9875, opened 2017-08-22
LogEntryParser no longer accepts zeros as valid entries

klaren:
The reason the parser originally supported zeroes was due to memory mapped files having zeroes. Since those old log formats are no longer supported this behavior is no longer desired.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/token/ReplicatedTokenRequestSerializer.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/token/ReplicatedTokenRequestSerializer.java
@@ -23,6 +23,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/token/ReplicatedTokenRequestSerializer.java
```diff
@@ -45,6 +46,7 @@
 {
     private ReplicatedTokenRequestSerializer()
     {
+        throw new AssertionError();
```
digitalstain: Since we're adding an assertion here, perhaps a message describing what the assertion is (class should not be instantiated) is in order.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/tx/ReplicatedTransactionFactory.java
```diff
@@ -71,10 +71,6 @@ public static ReplicatedTransaction createImmutableReplicatedTransaction( Transa
             throw new RuntimeException( e );
         }
 
-        /*
```
digitalstain: I would keep this comment in, as it's not clear why we do the copy. Saying that we need to trim the zeroes because they are not parseable from the log entry parser and can cause errors is a nice thing to note here.

digitalstain:
There is a typo in the title - entries, not entires :)

ragadeeshu (approved):
Looks good, apart from the comments that have been made already

klaren:
@digitalstain @ragadeeshu Thanks for the feedback! New commit available addressing it.

PR #9876, opened 2017-08-22
Code cleanup

MishaDemianenko:
Remove not used code in DynamicIndexStoreView, use entry sets iterators
use correct type for numberOfCollisions in EncodingIdMapper,
do proper non handled version check in RecordStorageCommandReaderFactory


community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageCommandReaderFactory.java
```diff
@@ -48,7 +48,8 @@ public RecordStorageCommandReaderFactory()
         LogEntryVersion[] versions = LogEntryVersion.values();
         for ( LogEntryVersion version : versions )
         {
-            if ( version == null )
+            CommandReader versionReader = readers[abs( version.byteCode() )];
```
klaren, line 51: :+1:

klaren (approved):
Everything looks good!

PR #9877, opened 2017-08-23
Improved value handling in property store

pontusmelke:
- Remove `Values.of` instead create with the correct type
- Create `UTF8StringValue` that doesn't need to create a string instance when serializing
- Use `UTF8StringValue` whenever possible

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/NodeUpdates.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/NodeUpdates.java
@@ -39,7 +39,6 @@ import static java.lang.String.format;
 import static java.util.Arrays.binarySearch;
 import static org.neo4j.kernel.impl.api.index.NodeUpdates.PropertyValueType.Changed;
 import static org.neo4j.kernel.impl.api.index.NodeUpdates.PropertyValueType.NoValue;
-import static org.neo4j.kernel.impl.store.ShortArray.EMPTY_LONG_ARRAY;
 
 /**
  * Subclasses of this represent events related to property changes due to node addition, deletion or update.
@@ -48,6 +47,7 @@ import static org.neo4j.kernel.impl.store.ShortArray.EMPTY_LONG_ARRAY;
 public class NodeUpdates implements PropertyLoader.PropertyLoadSink
 {
     private final long nodeId;
+    private static final long[] EMPTY_LONG_ARRAY = new long[0];
 
     // ASSUMPTION: these long arrays are actually sorted sets
     private final long[] labelsBefore;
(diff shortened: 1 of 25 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/StorePropertyPayloadCursor.java
```diff
@@ -308,7 +317,7 @@ private ByteBuffer newBiggerBuffer( int requiredCapacity )
         return ByteBuffer.allocate( newCapacity ).order( ByteOrder.LITTLE_ENDIAN );
     }
 
-    private static Object readArrayFromBuffer( ByteBuffer buffer )
+    private static Value readArrayFromBuffer( ByteBuffer buffer )
```
fickludd: Could this be `ArrayValue`, or is the `NO_VALUE` hindering that?

community/values/src/main/java/org/neo4j/values/storable/Values.java
```diff
@@ -51,6 +51,14 @@
...
     public static final DoubleValue E = Values.doubleValue( Math.E );
     public static final DoubleValue PI = Values.doubleValue( Math.PI );
+    public static final Value EMPTY_SHORT_ARRAY = Values.shortArray( new short[0] );
+    public static final Value EMPTY_BOOLEAN_ARRAY = Values.booleanArray( new boolean[0] );
+    public static final Value EMPTY_BYTE_ARRAY = Values.byteArray( new byte[0] );
+    public static final Value EMPTY_CHAR_ARRAY = Values.charArray( new char[0] );
+    public static final Value EMPTY_INT_ARRAY = Values.intArray( new int[0] );
+    public static final Value EMPTY_LONG_ARRAY = Values.longArray( new long[0] );
+    public static final Value EMPTY_FLOAT_ARRAY = Values.floatArray( new float[0] );
+    public static final Value EMPTY_DOUBLE_ARRAY = Values.doubleArray( new double[0] );
```
fickludd: These could have more specific types if we want to.

fickludd (commented):
From my perspective this looks good. I would like to see a CoreAPI trigger build just to verify performance... assuming there is some array-value tests in there.

alexaverbuch:
@fickludd @pontusmelke I can show you which benchmarks test array/string values and help you get those running

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/NodeUpdates.java
```diff
@@ -48,6 +47,7 @@
 public class NodeUpdates implements PropertyLoader.PropertyLoadSink
 {
     private final long nodeId;
+    private static final long[] EMPTY_LONG_ARRAY = new long[0];
```
tinwelint, line 50: Just a note: there's a `PrimitiveLongCollections#EMPTY_LONG_ARRAY` too, but I won't request change in this PR.

tinwelint (approved):
At first sight it looked like there were more objects created (`Value` instances), but it looks like such instances were created higher up anyway. Making `Value` explicit all the way seems like a good idea.

PR #9879, opened 2017-08-23
Replication throttling

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/ReplicatedContent.java
```diff
@@ -24,4 +24,13 @@
...
 {
+    default boolean hasSize()
+    {
+        return false;
+    }
+
+    default long size()
+    {
+        throw new UnsupportedOperationException();
+    }
```
andrewkerr9000, line 35: Not sure about these names. Maybe `hasPayload` and `payloadSize`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/SenderService.java
```diff
@@ -78,12 +75,17 @@ public void send( AdvertisedSocketAddress to, Message message )
...
         }
         finally
         {
             serviceLock.readLock().unlock();
         }
+
+        if ( block )
+        {
+            future.awaitUninterruptibly();
+        }
```
andrewkerr9000, line 88: I'm not familiar with this (or Netty `Future`'s) code so I'm missing context, but I'd normally expect an async service to return a `Future`, allowing the calling site to decide whether to chain callbacks, block, or just ignore.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/Throttler.java
```diff
@@ -0,0 +1,74 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.causalclustering.core.replication;
+
+import org.neo4j.function.ThrowingSupplier;
+
+/**
+ * Throttles calls to underlying invocation based on available credits.
+ */
+public class Throttler
```
andrewkerr9000: Worth unit testing?

andrewkerr9000 (commented):
I can't be of much help here I'm afraid.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/SenderService.java
martinfurmanski: Good observation and I actually was going for that initially but decided to just contain it with a simple block flag. I think there is a lot of potential for future maturity to be gained in the messaging stack. I am trying to be minimally invasive here.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/Throttler.java
martinfurmanski: Definitely should be. It is on the TODO.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/ReplicatedContent.java
martinfurmanski: Yea the naming can probably be improved. I guess everything has a payload though.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/SenderService.java
andrewkerr9000: :+1:

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/Throttler.java
andrewkerr9000: :+1:

PR #9881, opened 2017-08-23
Fixed error in unwind with projection in slotted runtime

craigtaverner:
The UNWIND support is incomplete and does not work for actual queries. There are several missing pieces and this PR fixes two:
- The existance of a renaming projection will cause an uncaught exception in planning, breaking fallback to interpreted (and therefor also breaking downstream builds like docs)
- No support for ReferenceFromRegister in expression converters, fixing this gets very simple unwind queries to work with slotted runtime

More work is required to get unwind to work with more complex cases, but this PR is important to get downstream builds to pass.

### DIFF
--- a/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisteredRewriter.scala
+++ b/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisteredRewriter.scala
@@ -157,12 +157,16 @@ class RegisteredRewriter(tokenContext: TokenContext) {
         }
 
       case Variable(k) =>
-        pipelineInformation(k) match {
-          case LongSlot(offset, false, CTNode, name) => NodeFromRegister(offset, name)
-          case LongSlot(offset, true, CTNode, name) => NullCheck(offset, NodeFromRegister(offset, name))
-          case LongSlot(offset, false, CTRelationship, name) => RelationshipFromRegister(offset, name)
-          case LongSlot(offset, true, CTRelationship, name) => NullCheck(offset, RelationshipFromRegister(offset, name))
-          case RefSlot(offset, _, _, _) => ReferenceFromRegister(offset)
+        pipelineInformation.get(k) match {
+          case Some(slot) => slot match {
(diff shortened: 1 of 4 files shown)


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisteredRewriter.scala
```diff
@@ -157,12 +157,16 @@ class RegisteredRewriter(tokenContext: TokenContext) {
         }
 
       case Variable(k) =>
-        pipelineInformation(k) match {
-          case LongSlot(offset, false, CTNode, name) => NodeFromRegister(offset, name)
-          case LongSlot(offset, true, CTNode, name) => NullCheck(offset, NodeFromRegister(offset, name))
-          case LongSlot(offset, false, CTRelationship, name) => RelationshipFromRegister(offset, name)
-          case LongSlot(offset, true, CTRelationship, name) => NullCheck(offset, RelationshipFromRegister(offset, name))
-          case RefSlot(offset, _, _, _) => ReferenceFromRegister(offset)
+        pipelineInformation.get(k) match {
```
systay, line 160: Instead of `match`ing on an `Option`, you could use map and getOrElse:

```
val x = pipelineInformation.get(k).map {
        case LongSlot(offset, false, CTNode, name) => NodeFromRegister(offset, name)
        case LongSlot(offset, true, CTNode, name) => NullCheck(offset, NodeFromRegister(offset, name))
        case LongSlot(offset, false, CTRelationship, name) => RelationshipFromRegister(offset, name)
        case LongSlot(offset, true, CTRelationship, name) => NullCheck(offset, RelationshipFromRegister(offset, name))
        case RefSlot(offset, _, _, _) => ReferenceFromRegister(offset)
        case _ =>
              throw new CantCompileQueryException("Unknown type for `" + k + "` in the pipeline information")
    }
x.getOrElse(throw new CantCompileQueryException("Did not find `" + k + "` in the pipeline information"))
```

craigtaverner: Good point. I think I was rushed to get the docs build working again. Perhaps we make this cleanup part of another PR.

PR #9885, opened 2017-08-23
Compatibility spec suite slotted runtime passing UNWIND+Projection

craigtaverner:
The PR at https://github.com/neo4j/neo4j/pull/9881 causes many TCK tests to pass with slotted runtime, that previously failed. However, the PR builds do not run the TCK, so this was not noticed.

This PR just removes a bunch of new passing tests from the blacklist for slotted runtime.


craigtaverner:
Replaced with TCK updates in https://github.com/neo4j/neo4j/pull/9891

PR #9883, opened 2017-08-23
Fixed SemiApply in the slotted runtime

eebus:
This PR adds the SemiApply and AntiSemiApply operators to the slotted runtime. This fixes the previous (broken) implementation. To get SemiApply to work, this PR also adds methods for copying 
 `PrimitiveExecutionContext`s between pipelines, and specifying what part of the context to copy.

PR #9889, opened 2017-08-23
Slotted Cartesian Product

alexaverbuch:
Branched from https://github.com/neo4j/neo4j/pull/9883

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
@@ -601,7 +601,7 @@ return p""")
 
   // Not sure if TCK material -- is this test just for `columns()`?
   test("columns should be in the provided order") {
-    val result = succeedWith(Configs.AllExceptSlotted, "MATCH (p),(o),(n),(t),(u),(s) RETURN p,o,n,t,u,s")
+    val result = succeedWith(Configs.All, "MATCH (p),(o),(n),(t),(u),(s) RETURN p,o,n,t,u,s")
 
     result.columns should equal(List("p", "o", "n", "t", "u", "s"))
   }
--- a/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/EnterprisePipeBuilder.scala
+++ b/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/EnterprisePipeBuilder.scala
@@ -231,6 +231,7 @@ class EnterprisePipeBuilder(fallback: PipeBuilder,
     implicit val table: SemanticTable = context.semanticTable
 
     val id = idMap.getOrElse(plan, new Id)
+    val pipeline = pipelines(plan)
 
     plan match {
       case Apply(_, _) =>
@@ -239,6 +240,12 @@ class EnterprisePipeBuilder(fallback: PipeBuilder,
       case _: SemiApply |
            _: AntiSemiApply =>
         fallback.build(plan, lhs, rhs)
+
+      case _: CartesianProduct =>
+        val lhsPlan = plan.lhs.get
+        val lhsLongCount = pipelines(lhsPlan).numberOfLongs
+        val lhsRefCount = pipelines(lhsPlan).numberOfReferences
+        CartesianProductRegisterPipe(lhs, rhs, lhsLongCount, lhsRefCount, pipeline)(id)
 
       case _ => throw new CantCompileQueryException(s"Unsupported logical plan operator: $plan")
     }
--- a/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/PrimitiveExecutionContext.scala
+++ b/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/PrimitiveExecutionContext.scala
@@ -32,14 +32,14 @@ case class PrimitiveExecutionContext(pipeline: PipelineInformation) extends Exec
   private val longs = new Array[Long](pipeline.numberOfLongs)
   private val refs = new Array[AnyValue](pipeline.numberOfReferences)
 
-  override def copyTo(target: ExecutionContext): Unit = target match {
+  override def copyTo(target: ExecutionContext, longOffset: Int = 0, refOffset: Int = 0): Unit = target match {
     case other@PrimitiveExecutionContext(otherPipeline) =>
       if (pipeline.numberOfLongs > otherPipeline.numberOfLongs ||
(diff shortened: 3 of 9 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
```diff
@@ -29,6 +29,26 @@ import scala.collection.mutable.ArrayBuffer
 
 class MatchAcceptanceTest extends ExecutionEngineFunSuite with QueryStatisticsTestSupport with CypherComparisonSupport {
 
+
+  test("apa") {
```
craigtaverner: Is this a test to throw away?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
```diff
@@ -487,10 +507,10 @@ return p""")
         |RETURN host""".stripMargin
 
     //WHEN
-    val result = profile(query)
-
+    val result = succeedWith(Configs.Interpreted, query)
     //THEN
-    result.toList should equal(List(Map("host" -> host), Map("host" -> null)))
+    println(result.getExecutionPlanDescription)
```
craigtaverner: Temp code?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/PrimitiveExecutionContext.scala
```diff
@@ -24,29 +24,33 @@ import org.neo4j.cypher.internal.frontend.v3_3.InternalException
...
+        throw new InternalException("Tried to copy more data into less.")
+      else {
+        System.arraycopy(longs, 0, other.longs, longOffset, pipeline.numberOfLongs)
+        System.arraycopy(refs, 0, other.refs, refOffset, pipeline.numberOfReferences)
+      }
     case _ => fail()
   }
 
-  def copyFrom(input: ExecutionContext): Unit = input match {
+  def copyFrom(input: ExecutionContext, nLongs: Int, nRefs: Int): Unit = input match {
```
craigtaverner: Missing override?

craigtaverner (changes requested):

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
alexaverbuch: It's not from this PR

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/PrimitiveExecutionContext.scala
alexaverbuch: @craigtaverner I think all review comments are now addressed

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/EnterprisePipeBuilder.scala
```diff
@@ -231,12 +231,23 @@ class EnterprisePipeBuilder(fallback: PipeBuilder,
...
 
+      case _: SemiApply |
+           _: AntiSemiApply =>
+        fallback.build(plan, lhs, rhs)
+
+      case _: CartesianProduct =>
+        val lhsPlan = plan.lhs.get
+        val lhsLongCount = pipelines(lhsPlan).numberOfLongs
+        val lhsRefCount = pipelines(lhsPlan).numberOfReferences
+        CartesianProductRegisterPipe(lhs, rhs, lhsLongCount, lhsRefCount, pipeline)(id)
```
craigtaverner, line 248: The lhsLongCount and lhsRefCount would be known by the lhsCtx inside internalCreateResults, so why do we need to pass it in here?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/CartesianProductRegisterPipe.scala
```diff
@@ -19,18 +19,26 @@
...
+                                       (val id: Id = new Id) extends Pipe {
+
+  protected def internalCreateResults(state: QueryState): Iterator[ExecutionContext] = {
+    lhs.createResults(state) flatMap {
+      lhsCtx =>
+        rhs.createResults(state) map {
+          rhsCtx =>
+            val context = PrimitiveExecutionContext(pipelineInformation)
+            lhsCtx.copyTo(context)
+            rhsCtx.copyTo(context, lhsLongCount, lhsRefCount)
```
craigtaverner, line 39: The lhsLongCount and lhsRefCount would be known by the lhsCtx already, so why do we need to pass it in to the CartesianProductRegisterPipe? Would it not be better to use the CartesianProductPipe pattern of calling `lhsCtx mergeWith rhsCtx` and have the PrimitiveExecutionContext implementation of `mergeWith` understand how to merge two execution contexts like this? Then we could also remove the CartesianProductRegisterPipe class entirely.

craigtaverner (changes requested):

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/CartesianProductRegisterPipe.scala
craigtaverner: One issue I see with my idea is that the Cartesian Product pipeline is provided at register allocation time. But it is strictly a concatenation of left and right hand side pipelines, and we are doing that with mergeWith also (at the execution context level). I still feel there are ways to simplify this.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/EnterprisePipeBuilder.scala
alexaverbuch: I agree it's ugly, but even though in the pipe we get a `PrimitiveExecutionContext` it is passed in as the base `ExecutionContext`, so we'd need to do something like 
```
rhsCtx.copyTo(context, lhsCtx.asInstanceOf[PrimitiveExecutionContext].pipeline.numberOfLongs, lhsCtx.asInstanceOf[PrimitiveExecutionContext].pipeline.numberOfReferences)
```

craigtaverner (approved):

PR #9890, opened 2017-08-23
Renaming 'enterprise interpreted' to 'slotted' runtime

craigtaverner:


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ExpressionAcceptanceTest.scala
```diff
@@ -97,15 +97,15 @@ class ExpressionAcceptanceTest extends ExecutionEngineFunSuite with CypherCompar
   test("prepending item to a list should behave correctly in all runtimes") {
     val query = "CYPHER WITH {a:[1,2,3]} AS x RETURN 'a:' + x.a AS r"
 
-    val result = succeedWith(Configs.All - Configs.EnterpriseInterpreted, query)
+    val result = succeedWith(Configs.AllExceptSleipnir, query)
```
alexaverbuch: Why `Sleipnir` instead of `Slotted`?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ExpressionAcceptanceTest.scala
```diff
@@ -97,15 +97,15 @@ class ExpressionAcceptanceTest extends ExecutionEngineFunSuite with CypherCompar
...
 
     result.toList.head("r") should equal(List("a:", 1, 2, 3))
   }
 
   test("appending item to a list should behave correctly in all runtimes") {
     val query = "CYPHER WITH {a:[1,2,3]} AS x RETURN x.a + 'a:' AS r"
 
-    val result = succeedWith(Configs.All - Configs.EnterpriseInterpreted, query)
+    val result = succeedWith(Configs.AllExceptSleipnir, query)
```
alexaverbuch: Why `Sleipnir` instead of `Slotted`?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/UnionAcceptanceTest.scala
```diff
@@ -39,7 +39,7 @@ class UnionAcceptanceTest extends ExecutionEngineFunSuite with CypherComparisonS
       |M.a as B
     """.stripMargin
 
-    val result = succeedWith(Configs.Interpreted - Configs.EnterpriseInterpreted, query)
+    val result = succeedWith(Configs.CommunityInterpreted, query)
```
alexaverbuch: I know you didn't add `CommunityInterpreted` in this commit, but while you're at it...
I think it's a bit weird to use the word "Community" here, it's not a runtime.
Maybe rename `Interpreted` with `AllInterpreted` or `SlottedAndInterpreted` and `CommunityInterpreted` to `Interpreted`?

alexaverbuch (dismissed):

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ExpressionAcceptanceTest.scala
craigtaverner: For the same reason we did not change Sleipnir to Enterprise-Interpreted. We are using this as a reminder, a TODO. At the time the goal was to get 100% coverage with sleipnir. That is unlikely in the short term, so perhaps we should rename this. I also see a few place where we have RegisterAllocation, and that could also be renamed, but would prefer to do that in another PR. This PR focuses on the surface and the definition of the runtime.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/UnionAcceptanceTest.scala
craigtaverner: Good point. I did do a tiny simplification in the Configs, but could take it further. But perhaps we do it as a Configs refactoring PR instead.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/CostCompatibility.scala
```diff
@@ -62,7 +62,7 @@ case class CostCompatibility[CONTEXT <: CommunityRuntimeContext,
   override val maybeRuntimeName: Option[RuntimeName] = runtime match {
     case CypherRuntime.default => None
     case CypherRuntime.interpreted => Some(InterpretedRuntimeName)
-    case CypherRuntime.registerInterpreted => Some(EnterpriseInterpretedRuntimeName)
+    case CypherRuntime.`slotted` => Some(SlottedRuntimeName)
```
alexaverbuch: why with ``?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ExpressionAcceptanceTest.scala
alexaverbuch: I wouldn't be against renaming all occurrences of "Sleipnir" || "Registered" in the codebase with one consistent name, i.e., "Slotted". 
This PR already does a fair bit of that (e.g., test names, test scenarios, etc.), so why not do it all now.

craigtaverner: I want a PR that causes as little conflict with existing PR's, while still also achieving the goal of the PR. I drew the line at renaming internal classes that are commonly touched by ongoing work. I think phasing this in two phases will make this much easier.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/CostCompatibility.scala
craigtaverner: I believe intellij did that for me. When I saw it, I assumed there was a valid reason, but did not investigate what that was.

PR #9893, opened 2017-08-24
Default configuration in acceptance suites

pontusmelke:
Currently we always run acceptance tests with a particular runtime and/or planner configured. This means that we are not validating that the database works with default values - hence blacklisted tests might (and currently does) hide all runtime bugs in the runtimes.

Running tests with default configuration exposed a couple of bugs, both in the compiled runtime and in the slotted runtime, that is fixed in this PR as well.


pontusmelke:
~~NOTE that with these settings we currently fail on 8 tests in the TCK. We need to fix these ASAP and then merge this PR.~~

PR #9894, opened 2017-08-24
Align value module ordering with OpenCypher

fickludd:
This PR changes the ordering of storable and virtual `Value`s. This is done to align with the accepted [OpenCypher CIP defining orderability](https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc).

Changes to value ordering when comparing **storable** values:
  * Arrays are ordered before scalars

Changes to value ordering when comparing **all** values:
  * supertypes are ordered as `map`, `node`, `relationship`, `list/array`, `path`, `point`, `text`, `boolean`, `number`, `NO_VALUE`
  * lists have dictionary order, eg. `[], [1], [1,2], [2], ...` 
  * paths are ordered as lists of alternating nodes and relationships

Finally this allows us to use `AnyValues.COMPARATOR` for `SortPipe` ordering.

### DIFF
--- a/community/values/src/main/java/org/neo4j/values/SequenceValue.java
+++ b/community/values/src/main/java/org/neo4j/values/SequenceValue.java
@@ -20,14 +20,38 @@
 package org.neo4j.values;
 
 import java.util.Comparator;
+import java.util.Iterator;
+
+import static org.neo4j.values.SequenceValue.IterationPreference.RANDOM_ACCESS;
 
 /**
  * Values that represent sequences of values (such as Lists or Arrays) need to implement this interface.
  * Thus we can get an equality check that is based on the values (e.g. List.equals(ArrayValue) )
  * Values that implement this interface also need to overwrite isSequence() to return true!
+ *
+ * Note that even though SequenceValue extends Iterable iterating over the sequence using iterator() might not be the
+ * most performant method. Branch using iterationPreference() in performance critical code paths.
  */
-public interface SequenceValue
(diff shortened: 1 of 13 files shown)


community/values/src/main/java/org/neo4j/values/SequenceValue.java
```diff
@@ -52,4 +55,25 @@ default boolean equals( SequenceValue other )
...
+
+    default int compareToSequence( SequenceValue other, Comparator<AnyValue> comparator )
+    {
+        int i = 0;
+        int x = 0;
+        int length = Math.min( length(), other.length() );
+
+        while ( x == 0 && i < length )
+        {
+            x = comparator.compare( value( i ), other.value( i ) );
```
pontusmelke: I know this doesn't really change from how it used to work but not all ListValue implementations have constant time random access - `FilteredListValue` hasn't. So it would be more efficient - at least in that case - to use the `iterator` instead.

fickludd: Nice catch. Specialized both `compareTo` and `equals` now.

fickludd:
@tinwelint For future reference: should we use the `team-kernel` label instead of `kernel` or what is happening here?

tinwelint:
There's only one label now: `team-kernel`

Issue #9775, opened 2017-08-04
Neo4j 3.2.0 won't start on database files used by a 3.2.3 instance previously


amorgner:
Neo4j 3.2.0 is unable to run on database files created with Neo4j 3.2.0 after they have been used by Neo4j 3.2.3, log shows error

    Unrecognized log entry version -10. At position LogPosition{logVersion=0, byteOffset=34} and entry version null

**Neo4j Version:** 3.2.3 / 3.2.0
**Operating System:** Ubuntu 17.10  

### Steps to reproduce
1. Start Neo4j 3.2.0 (fresh installation)
2. Stop Neo4j 3.2.0
3. Start Neo4j 3.2.3 on same database files created by 3.2.0 instance
4. Stop Neo4j 3.2.3
5. Start Neo4j 3.2.0 again on same database files previously used with 3.2.3 instance

### Expected behavior
Normal database startup there's no documented format change, and 3.2.3 starts without `dbms.allow_format_migration` being enabled.

### Actual behavior
Database fails to start with the following exception stacktrace:

	2017-08-04 11:09:58.921+0000 ERROR Failed to start Neo4j: Starting Neo4j failed: Component 'org.neo4j.server.database.LifecycleManagingDatabase@156604c1' was successfully initialized, but failed to start. Please see attached cause exception. Starting Neo4j failed: Component 'org.neo4j.server.database.LifecycleManagingDatabase@156604c1' was successfully initialized, but failed to start. Please see attached cause exception.
	org.neo4j.server.ServerStartupException: Starting Neo4j failed: Component 'org.neo4j.server.database.LifecycleManagingDatabase@156604c1' was successfully initialized, but failed to start. Please see attached cause exception.
		at org.neo4j.server.exception.ServerStartupErrors.translateToServerStartupError(ServerStartupErrors.java:68)
		at org.neo4j.server.AbstractNeoServer.start(AbstractNeoServer.java:219)
		at org.neo4j.server.ServerBootstrapper.start(ServerBootstrapper.java:107)
		at org.neo4j.server.ServerBootstrapper.start(ServerBootstrapper.java:80)
		at org.neo4j.server.enterprise.EnterpriseEntryPoint.main(EnterpriseEntryPoint.java:32)
	Caused by: org.neo4j.kernel.lifecycle.LifecycleException: Component 'org.neo4j.server.database.LifecycleManagingDatabase@156604c1' was successfully initialized, but failed to start. Please see attached cause exception.
[...]

PR #9896, opened 2017-08-24
Respect dbms.allow_upgrade concerning LogEntryVersion

klaren:
During startup, we now check the version of the last appended transaction log entry and match it against the current version. This way we can detect when the `LogEntryVersion` is changed, and this is now considered an upgrade.

The error messages have also been improved when an unknown transaction log format is encountered, presenting the user with an educated guess of what actually is wrong the transaction logs.

This fixes issue #9775  by disallowing the server to fully start and append transaction log entries that the older version responsible for creating the transaction logs in the first place cannot read.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/recovery/RecoveryRequiredChecker.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/recovery/RecoveryRequiredChecker.java
@@ -27,11 +27,11 @@ import org.neo4j.io.pagecache.PageCache;
 import org.neo4j.kernel.impl.store.MetaDataStore;
 import org.neo4j.kernel.impl.store.NeoStores;
 import org.neo4j.kernel.impl.transaction.log.LogPosition;
+import org.neo4j.kernel.impl.transaction.log.LogTailScanner;
 import org.neo4j.kernel.impl.transaction.log.PhysicalLogFiles;
 import org.neo4j.kernel.impl.transaction.log.ReadableClosablePositionAwareChannel;
 import org.neo4j.kernel.impl.transaction.log.entry.LogEntryReader;
 import org.neo4j.kernel.impl.transaction.log.entry.VersionAwareLogEntryReader;
-import org.neo4j.kernel.recovery.LatestCheckPointFinder;
 import org.neo4j.kernel.recovery.PositionToRecoverFrom;
 
 import static org.neo4j.kernel.recovery.PositionToRecoverFrom.NO_MONITOR;
@@ -67,7 +67,7 @@ public class RecoveryRequiredChecker
 
         LogEntryReader<ReadableClosablePositionAwareChannel> reader = new VersionAwareLogEntryReader<>();
 
-        LatestCheckPointFinder finder = new LatestCheckPointFinder( logFiles, fs, reader );
+        LogTailScanner finder = new LogTailScanner( logFiles, fs, reader );
         return new PositionToRecoverFrom( finder, NO_MONITOR ).apply( logVersion ) != LogPosition.UNSPECIFIED;
     }
 }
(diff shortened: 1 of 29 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/recovery/RecoveryRequiredChecker.java
```diff
@@ -63,11 +63,12 @@ public boolean isRecoveryRequiredAt( File dataDir ) throws IOException
         }
 
         long logVersion = MetaDataStore.getRecord( pageCache, neoStore, MetaDataStore.Position.LOG_VERSION );
+        MetaDataStore.setRecord( pageCache, neoStore, MetaDataStore.Position.LOG_VERSION, logVersion );
```
tinwelint: Why set a new value for that record here? Isn't this a "checker" only?

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
```diff
@@ -275,12 +272,6 @@ public StoreReadLayer storeReadLayer()
         return storeLayer;
     }
 
-    @Override
-    public CommandReaderFactory commandReaderFactory()
-    {
-        return commandReaderFactory;
```
tinwelint: Removing this is counter to how it was intended. A `CommandReaderFactory` is designed to be state-less and so it should be enough with one per database instance. It may be that other places who currently instantiates a `RecordStorageCommandReaderFactory` should really be getting this instance instead, especially hot paths that read commands. I'd like that way better than to remove this method.

`VersionAwareLogEntryReader` has a default constructor which creates a new `RecordStorageCommandReaderFactory` mostly for convenience in tests. Now I see that it has been used and new instances of these readers are created in many places, even if it should always be dependent on which specific storage engine is used underneath. It's fine currently, but will obviously break if/when in the future having multiple storage engines.

You don't have to do anything to other parts of the code, but would you mind keeping this method at least (and I'll create a card for the other stuff)?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/LogEntryWriter.java
```diff
@@ -37,17 +37,34 @@
...
+        this( channel, CURRENT );
+    }
+
+    /**
+     * Create a writer that uses a different version than {@link LogEntryVersion#CURRENT}. Useful when writing test for
+     * migration/upgrade scenarios.
+     * @param channel underlying channel
+     * @param version version to put in the header
+     */
+    public LogEntryWriter( FlushableChannel channel, LogEntryVersion version )
```
tinwelint: Hmm, this constructor is a bit dangerous if used in main (by mistake or otherwise). Could it be avoided somehow?

tinwelint (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/impl/recovery/RecoveryRequiredChecker.java
klaren: Oops, left over debug code, removing.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
klaren: The main problem is that the `LogEntryReader` now needs to be created before the `RecordStorageEngine`, I'll restore the functions and add the card to the backlog for now. The `VersionAwareLogEntryReader` is only created once now in the hot path, so it's not a huge problem.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/LogEntryWriter.java
klaren: Scary constructor indeed, I'll inline the needed channel writes instead.

klaren:
@tinwelint Two new commits addressing you comments ready for reviewing.

tinwelint (approved):

tinwelint:
Just needs a rebase now

PR #9898, opened 2017-08-24
Add MergeCreateNode, ConditionalApply, AntiConditionalApply to slotted runtime

henriknyman:


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/BuildEnterpriseInterpretedExecutionPlan.scala
```diff
@@ -79,6 +79,7 @@ object BuildEnterpriseInterpretedExecutionPlan extends Phase[EnterpriseRuntimeCo
       new CompilationState(from, Some(execPlan))
     } catch {
       case e: CypherException =>
+        e.printStackTrace()
```
fickludd: Wh00t?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/pipes/ConditionalApplyRegisterPipe.scala
```diff
@@ -0,0 +1,60 @@
...
+        else {
+          val output = PrimitiveExecutionContext(pipelineInformation)
+          output.copyFrom(lhsContext)
+          Iterator.single(output)
+        }
+    }
+
+  private def condition(context: ExecutionContext) = {
+
+    def notNull(name: String): Boolean = pipelineInformation.get(name) match {
```
fickludd: I wonder if we could make a offset lookup here instead of by name? We should know the offsets at compile-time, so the map lookup could be done once then, unless I missed something.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
```diff
@@ -218,13 +218,16 @@ object RegisterAllocation {
...
+        // The variable name should already have been allocated by the NodeLeafPlan
+        incomingPipeline
+
       case CreateRelationship(_, IdName(name), startNode, typ, endNode, props) =>
         incomingPipeline.newLong(name, nullable = false, CTRelationship)
         incomingPipeline
 
       case MergeCreateRelationship(_, IdName(name), startNode, typ, endNode, props) =>
         incomingPipeline.newLong(name, nullable = false, CTRelationship)
-        incomingPipeline
```
fickludd: does `newLong()` return a `PipelineInformation`? This change breaks source code symmetry for me, I think I would prefer to keep this `incomingPipeline` for clarity, or if we remove it, remove it everywhere.

fickludd (changes requested):
Nice to get merge added. Since you need to rebase anyway, some minor clean-ups can happen, and optionally it would be nice to avoid the map-lookup at runtime.

fickludd (approved):

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/RegisterAllocation.scala
```diff
@@ -218,6 +218,10 @@ object RegisterAllocation {
         incomingPipeline.newLong(name, nullable = false, CTNode)
         incomingPipeline
 
+      case MergeCreateNode(_, IdName(name), _, _) =>
+        // The variable name should already have been allocated by the NodeLeafPlan
```
systay, line 222: Maybe we can assert that this is the case?

fickludd: Good point, but though that should not hold back this PR any longer.

PR #9904, opened 2017-08-24
Remove timeout execution guard

MishaDemianenko:
Remove timeout execution guard and guarding statement operations.
Now transaction timeout will be checked only by KernelTransactionTimeoutMonitor.
which will update termination status that will be checked by assertOpen in
OperationsFacade only.
Now KernelTransactionTimeoutMonitor will by default each 2 seconds.
Keep Guard implementations to have an ability to properly check transaction
status in any additional places.


tinwelint:
@MishaDemianenko There's a checkstyle issue as reported by the build:

`integrationtests/src/test/java/org/neo4j/TransactionGuardIntegrationTest.java:785:65: '{' at column 65 should be on a new line. [LeftCurly]`

MishaDemianenko:
@tinwelint updated

tinwelint (approved):
Nice to get rid of that layer!

PR #9901, opened 2017-08-24
Add aggregation to slotted runtime

systay:

PR #9908, opened 2017-08-25
Distinct in slotted runtime

pontusmelke:
Builds on https://github.com/neo4j/neo4j/pull/9901


systay (approved):
LGTM!

In the future though, I think it's better to only change code style of the code you are editing, not the whole file. Makes it easier to see which are the interesting changes.

systay:
DISTINCT is now part of #9901

PR #9911, opened 2017-08-25
Infrastructure for isolation levels

chrisvest:
To begin with, we are only adding the initial draft of half of an isolation level that is slightly stronger than Read Committed, to explore the effort needed to do more in this area.

None of this is exposed in the public API, and certainly won't be in 3.2.


community/common/src/test/java/org/neo4j/test/rule/RepeatRule.java
```diff
@@ -77,6 +88,10 @@ public Statement apply( Statement base, Description description )
         {
             return new RepeatStatement( repeat.times(), base );
         }
+        if ( defaultRepetitions != 0 )
```
tinwelint, line 91: It's so funny that I made almost the exact same change in #9953

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -378,24 +373,23 @@ public long nodeGetFromUniqueIndexSeek(
 
         // If we find the node - hold a shared lock. If we don't find a node - hold an exclusive lock.
         // If locks are deferred than both shared and exclusive locks will be taken only at commit time.
-        Locks.Client locks = state.locks().optimistic();
-        LockTracer lockTracer = state.lockTracer();
+        StatementLocks statementLocks = state.locks();
         long indexEntryId = indexEntryResourceId( labelId, predicates );
 
-        locks.acquireShared( lockTracer, INDEX_ENTRY, indexEntryId );
+        statementLocks.uniquenessConstraintEntryAcquireShared( indexEntryId );
```
systay, line 379: If we are taking a stab at a fundamental change in our locking model, maybe it's time to introduce `Update` locks as well, instead of doing little dance of grabbing and releasing locks?

chrisvest:
Closing because we have learned from this experience that there is more work in this than initially estimated, and we therefor need to take a second look at our priorities.

PR #9923, opened 2017-08-28
Do not record stack traces of open/closed statements by default

MishaDemianenko:
Recording stack traces takes a lot of time on build agents.
Introduce 'recordStatementsTraces' feature toggle in KernelStatement.class to
control traces recording. Disable it by default.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
@@ -79,7 +79,7 @@ import static org.neo4j.unsafe.impl.internal.dragons.FeatureToggles.toggle;
  */
 public class KernelStatement implements TxStateHolder, Statement, AssertOpen
 {
-    private static final boolean TRACK_STATEMENTS = flag( KernelStatement.class, "trackStatements", true );
+    private static final boolean TRACK_STATEMENTS = flag( KernelStatement.class, "trackStatements", false );
     private static final boolean RECORD_STATEMENTS_TRACES = flag( KernelStatement.class, "recordStatementsTraces", false );
     private static final int STATEMENT_TRACK_HISTORY_MAX_SIZE = 100;
     private static final Deque<StackTraceElement[]> EMPTY_STATEMENT_HISTORY = new ArrayDeque<>( 0 );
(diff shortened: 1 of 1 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
```diff
@@ -107,7 +110,7 @@ public KernelStatement( KernelTransactionImplementation transaction,
         this.facade = new OperationsFacade( transaction, this, procedures );
         this.executingQueryList = ExecutingQueryList.EMPTY;
         this.systemLockTracer = systemLockTracer;
-        this.statementOpenCloseCalls = TRACK_STATEMENTS ? new ArrayDeque<>() : null;
+        this.statementOpenCloseCalls = TRACK_STATEMENTS ? new ArrayDeque<>() : EMPTY_STATEMENT_HISTORY;
```
klaren: Shouldn't this be `RECORD_STATEMENTS_TRACES` as conditional?
I think returning null would be clearer. See comment below.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
```diff
@@ -78,8 +79,10 @@
  */
 public class KernelStatement implements TxStateHolder, Statement, AssertOpen
 {
-    private static final boolean TRACK_STATEMENTS = flag( KernelStatement.class, "trackStatements", true );
+    private static final boolean TRACK_STATEMENTS = flag( KernelStatement.class, "trackStatements", false );
```
klaren, line 82: There is no information about `TRACK_STATEMENTS`, a simple comment would suffice. 
I also think it would be valid to note that `RECORD_STATEMENTS_TRACES` does nothing without `TRACK_STATEMENTS`.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelStatement.java
```diff
@@ -338,6 +349,10 @@ private void recordOpenCloseMethods()
 
         private static String buildMessage( Deque<StackTraceElement[]> openCloseTraces )
         {
+            if ( openCloseTraces.isEmpty() )
```
klaren, line 352: Could check `openCloseTraces != null || openCloseTraces.isEmpty()` here and removed `EMPTY_STATEMENT_HISTORY`.

klaren (commented):
Just a few clarifications.

PR #9926, opened 2017-08-29
Validate type of fields added to legacy index.

ragadeeshu:
Without this validation, a transaction could commit, but fail to write to
disk if the type of the field was invalid.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
+++ b/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
@@ -27,8 +27,11 @@ import org.apache.lucene.analysis.core.LowerCaseFilter;
 import org.apache.lucene.analysis.core.WhitespaceAnalyzer;
 import org.apache.lucene.analysis.core.WhitespaceTokenizer;
 import org.apache.lucene.document.Document;
+import org.apache.lucene.index.DocValuesType;
+import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.IndexCommit;
 import org.apache.lucene.index.IndexableField;
+import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.SnapshotDeletionPolicy;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.Sort;
@@ -42,6 +45,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
```diff
@@ -111,6 +115,45 @@ public String toString()
     private boolean closed;
     private LuceneFilesystemFacade filesystemFacade;
     private IndexReferenceFactory indexReferenceFactory;
+    private Map<IndexIdentifier,Map<String,DocValuesType>> indexTypeMap;
```
MishaDemianenko, line 119: I would assume that this map should be concurrent since it will be used by multiple threads and type information should be available for all of those

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
```diff
@@ -111,6 +115,45 @@ public String toString()
...
+        {
+            expectedType = DocValuesType.SORTED_NUMERIC;
+            expectedTypeName = "numbers";
+        }
+        else
+        {
+            expectedType = DocValuesType.SORTED_SET;
+            expectedTypeName = "strings";
+        }
+        Map<String,DocValuesType> stringDocValuesTypeMap = indexTypeMap.get( identifier );
```
MishaDemianenko: indexTypeMap.computeIfAbsent?

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
```diff
@@ -111,6 +115,45 @@ public String toString()
...
+        // If the index searcher has never been loaded, we need to load it now to populate the map.
+        if ( stringDocValuesTypeMap == null )
+        {
+            getIndexSearcher( identifier );
+            stringDocValuesTypeMap = indexTypeMap.get( identifier );
+        }
+        DocValuesType actualType;
+        synchronized ( stringDocValuesTypeMap )
```
MishaDemianenko: i would expect `stringDocValuesTypeMap` map to be concurrent as well

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
```diff
@@ -111,6 +115,45 @@ public String toString()
...
+        {
+            actualType = stringDocValuesTypeMap.get( key );
```
MishaDemianenko: stringDocValuesTypeMap.putIfAbsent should be used instead of synch section

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
```diff
@@ -281,6 +325,15 @@ synchronized IndexReference syncGetIndexSearcher( IndexIdentifier identifier )
             {
                 indexReference = indexReferenceFactory.createIndexReference( identifier );
                 indexSearchers.put( identifier, indexReference );
+                HashMap<String,DocValuesType> fieldTypes = new HashMap<>();
```
MishaDemianenko: mentioned earlier, it should be some sort of conc. map

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
MishaDemianenko, line 119: this map is never cleaned, i would expect that we will remove index information from it when the index is removed

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
```diff
@@ -281,6 +325,15 @@ synchronized IndexReference syncGetIndexSearcher( IndexIdentifier identifier )
...
+                for ( LeafReaderContext leafReaderContext : indexReference.getSearcher().getTopReaderContext().leaves() )
+                {
+                    for ( FieldInfo fieldInfo : leafReaderContext.reader().getFieldInfos() )
+                    {
+                        fieldTypes.put( fieldInfo.name, fieldInfo.getDocValuesType() );
```
MishaDemianenko, line 327: also not quite sure when we should remove field type info from map. Would expect that in normal circumstances people do not have lots dynamic fields in index, but maybe some point for future

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
```diff
@@ -111,6 +115,45 @@ public String toString()
     private boolean closed;
     private LuceneFilesystemFacade filesystemFacade;
     private IndexReferenceFactory indexReferenceFactory;
+    private Map<IndexIdentifier,Map<String,DocValuesType>> indexTypeMap;
+
+    public void assertValidType( String key, Object value, IndexIdentifier identifier )
```
MishaDemianenko: would be nice to move this method below constructor

MishaDemianenko (changes requested):
Couple of comments

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
ragadeeshu: Sadly not really applicable here, since we need to load the index searcher, and that will populate the map for us.

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneDataSource.java
ragadeeshu: Since lucene only cleans the internal map when removing the index entirely, we should to the same.

ragadeeshu:
@MishaDemianenko forgot to ping you, sorry. This is updated

MishaDemianenko (approved):
Approved, only one question have you checked how big performance impact does this change introduce?

PR #9928, opened 2017-08-29
Fix variable scope bug with aggregation, foreach and sorting

henriknyman:
For certain queries with aggregation expressions in combination with
child scopes (FOREACH) and ORDER BY there was a problem with variable
scope resulting from the interaction of the AST rewriters
normalizeWithClauses, expandStar and isolateAggregation with semantic
analysis.
The expandStar rewriter would rewrite PragmaWithout clauses to explicit
With clauses using variable positions that would later be outdated by
the second semantic analysis that follows rewriting, which would break
the correctness of that semantic analysis.


pontusmelke:
Nice find!

craigtaverner (approved):

craigtaverner:
Closing, and we can re-open again if we feel we need it.

PR #9937, opened 2017-08-29
Modularize enterprise cypher

systay:


systay:
[Physical planner](https://trello.com/c/hHRe6bwF/45-physical-planner)

systay:
[Move runtime to its own package](https://trello.com/c/Owef5Szg/9-move-runtime-to-its-own-package)

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/TopDownWithState.scala
```diff
@@ -1,69 +0,0 @@
...
-package org.neo4j.cypher.internal.compatibility.v3_3.runtime
-
-import org.neo4j.cypher.internal.frontend.v3_3.Foldable._
-import org.neo4j.cypher.internal.frontend.v3_3.Rewritable._
-import org.neo4j.cypher.internal.frontend.v3_3.Rewriter
-
-import scala.annotation.tailrec
-import scala.collection.mutable
-
-object TopDownWithState {
```
fickludd, line 29: What was this, some old forgotten code that's no longer used?

fickludd (approved):
Much appreciated step towards better cypher code structure. Like it.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/TopDownWithState.scala
systay: Yep. I added it because we would need it for registered rewriter (or so I thought), but then I solved that with a different approach and forgot to delete this class.

enterprise/cypher/slotted-runtime/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/interpreted/PrimitiveExecutionContextTest.scala
```diff
@@ -20,6 +20,7 @@
 package org.neo4j.cypher.internal.compatibility.v3_3.runtime.interpreted
 
 import org.neo4j.cypher.internal.compatibility.v3_3.runtime.PipelineInformation
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.slotted.PrimitiveExecutionContext
```
fickludd: Should rename the test package too, this import shouldn't be needed.

fickludd (changes requested):
This looks good, except we should also rename the test package to `slotted`.

sherfert (approved):

PR #9942, opened 2017-08-30
Increased timeout to reduce flakyness

eebus:


enterprise/cypher/cypher/src/test/scala/org/neo4j/cypher/ExecutionEngineTest.scala
```diff
@@ -889,7 +889,7 @@ order by a.COL1""")
     createNode()
 
     // when
-    timeOutIn(2, TimeUnit.SECONDS) {
+    timeOutIn(20, TimeUnit.SECONDS) {
```
alexaverbuch: Can we simply delete this test? Looks like it was initially introduced for an infinite loop in the planner.  cc @systay

alexaverbuch (changes requested):

alexaverbuch (approved):

PR #9946, opened 2017-08-30
Create and check for existence for relationship simultaneous

eebus:
Deletion might cause a race condition when deletion the same relationship concurrent. 
We now check for existence and creation at the same time, removing the race.

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/pipes/IdSeekIterator.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/pipes/IdSeekIterator.scala
@@ -49,7 +49,8 @@ abstract class IdSeekIterator[T <: PropertyContainer]
   private def computeNextEntity(): T = {
     while (entityIds.hasNext) {
       val id = asLongEntityId(entityIds.next())
-      return operations.getByIdIfExists(id.longValue()).getOrElse(null.asInstanceOf[T])
+      val maybeEntity = operations.getByIdIfExists(id.longValue())
+      if(maybeEntity.isDefined) return maybeEntity.get
     }
     null.asInstanceOf[T]
   }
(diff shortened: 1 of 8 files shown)


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/pipes/IdSeekIterator.scala
```diff
@@ -49,9 +49,7 @@ abstract class IdSeekIterator[T <: PropertyContainer]
   private def computeNextEntity(): T = {
     while (entityIds.hasNext) {
       val id = asLongEntityId(entityIds.next())
-      if (operations.exists(id.longValue()))
-        return operations.getById(id.longValue())
-
+      return operations.getByIdIfExists(id.longValue()).getOrElse(null.asInstanceOf[T])
```
craigtaverner: Perhaps a logic error here. Previously we only returned if the entity existed, otherwise moved on to the next one. Now we return even if it does not exist, and return null. To match older behaviour perhaps we should have something like:

```
   val maybeEntity = operations.getByIdIfExists(id.longValue())
   if(maybeEntity.isDefined) return maybeEntity.get
```

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_3/TransactionBoundQueryContext.scala
```diff
@@ -413,6 +413,14 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
 
     override def exists(id: Long): Boolean =
       transactionalContext.statement.readOperations().nodeExists(id)
+
+    override def getByIdIfExists(id: Long): Option[Node] = {
+      try {
+        Some(entityAccessor.newNodeProxyById(id))
```
craigtaverner: This does not check for existence.

craigtaverner (changes requested):

craigtaverner (approved):

PR #9947, opened 2017-08-30
Fix IndexProviderCompatibilityTestSuite

burqen:
... by introducing possibility to suffix test directory rule.

When running test suite in parallel, the same test directories
will be shared between different test suits because they are
generated from owning class (which is similar) and test name.

By adding a unique suffix per test suite we avoid this directory
sharing.

### DIFF
--- a/community/io/src/test/java/org/neo4j/test/rule/PageCacheAndDependenciesRule.java
+++ b/community/io/src/test/java/org/neo4j/test/rule/PageCacheAndDependenciesRule.java
@@ -24,6 +24,7 @@ import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 import org.neo4j.io.fs.FileSystemAbstraction;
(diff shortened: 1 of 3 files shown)


community/io/src/test/java/org/neo4j/test/rule/PageCacheAndDependenciesRule.java
```diff
@@ -45,17 +46,19 @@
 
     public PageCacheAndDependenciesRule()
     {
-        this( () -> new EphemeralFileSystemRule() );
+        this( () -> new EphemeralFileSystemRule(), fs -> TestDirectory.testDirectory( fs ) );
```
klaren, line 49: Can be simplified to:
`this( EphemeralFileSystemRule::new, TestDirectory::testDirectory );`

community/kernel/src/test/java/org/neo4j/kernel/api/index/IndexProviderCompatibilityTestSuite.java
```diff
@@ -72,6 +73,9 @@ public void setup()
 
         public Compatibility( IndexProviderCompatibilityTestSuite testSuite, IndexDescriptor descriptor )
         {
+            this.pageCacheAndDependenciesRule = new PageCacheAndDependenciesRule( DefaultFileSystemRule::new,
+                    fs -> TestDirectory.testDirectory( null, fs, "-" + testSuite.getClass().getSimpleName().substring( 0, 3 ) ) );
```
klaren, line 77: 3 characters might be a bit optimistic for future cases. If you just want it to run fine today 1 would suffice :)

klaren (changes requested):
Just two small comments for the changes. There is, however, a "hidden" test directory rule in `UniqueConstraintCompatibility` rows 988 to 997 that's easy to miss.

community/io/src/test/java/org/neo4j/test/rule/PageCacheAndDependenciesRule.java
burqen: `TestDirectory` will create it's own default file system if not provided with one and we want to use the same one as the page cache does

klaren: The proposed simplification does not change behavior

MishaDemianenko:
@burqen you do not need to do all of that prefix thingy, you only need to pass `getClass()` to internal TestDirectory and it will create correct non shared folder for you. There is no need to pass _unique_ prefix to something that already should be unique.

burqen:
This will be replaced by fix from @klaren . Closing

PR #9951, opened 2017-08-31
Remove dependency on graphdb-api in values

pontusmelke:
The `values` module was depending on `neo4j-graphdb-api` in order to be able to create wrapped node and relationship proxies. This dependency is now removed and the wrapping is now done in the top level cypher module instead of in the values module.


community/values/src/test/java/org/neo4j/values/storable/BufferValueWriter.java
```diff
@@ -157,7 +156,7 @@ public void writeString( char value )
     @Override
     public void writeUTF8( byte[] bytes, int offset, int length ) throws RuntimeException
     {
-        buffer.add( UTF8.decode( bytes, offset, length ) );
+        buffer.add( new String( bytes, offset, length, StandardCharsets.UTF_8 ) );
```
fickludd, line 159: Interesting... what is the reason?

fickludd (approved):
Looks good to me, this is a sensible restructuring to remove dependencies.

fickludd:
[Should values depend on graphdb-api?](https://trello.com/c/77xJf6um/16-should-values-depend-on-graphdb-api)

fickludd:
[Where should QueryResult live?](https://trello.com/c/vbOAF4GI/15-where-should-queryresult-live)

community/values/src/test/java/org/neo4j/values/storable/BufferValueWriter.java
pontusmelke: values doesn't doesn't have that utility as a dependency anymore. `UTF8.decode` just does that very thing.

PR #9953, opened 2017-08-31
Proper index recovery

tinwelint:
This commit changes the way recovery works, with the main driving force to
have index updates be replayed as they were during normal transaction
application before the crash or otherwise unclean shutdown.

Previously index recovery was special and worked by recording which nodes
that were changed in recovered transaction followed by a complete reindex
of those nodes in all existing indexes. Besides the downside of requiring
special recovery code it also put on the index the requirement of being able
to efficiently delete index entries about a certain node id, i.e. to be able to
efficiently do lookups on node id. This worked in the Lucene case, but doesn't
work for the native number index, which only keys on the property values.
  Index recovery worked this way because the act of producing index updates
from a transaction in many cases requires reading data from the store.
It would be incorrect and often impossible to read from the store while it was
recovering.

Now this has changed so that indexes participate in and see normal index updates
during recovery. The enabler for this is to first do a reverse recovery where
all the "before" versions of all transactions since the last checkpoint are
applied in reverse order onto the neostore only. This takes the store back in
time to how it looked at the checkpoint and into a consistent state.
From there a forward recovery is performed where indexes participate as they
would in normal transaction application.

Interesting to note is that the changes, and with some additional changes,
has made recovery both faster, due to:

- The previous reindexing in the end was rather slow.
- LockService was used when applying recovered transactions, to no benefit.

... and less memory consuming, due to:

- Lowering of recovery transaction queue size from 10k --> 100, where there
  was no observed performance benefit having the queue so large. The amount
  of memory required to hold 10k large transactions could be gigabytes.
- Memory efficient implementation of reverse transaction cursor.

So recovery is now more straight forward, faster and more memory efficient.
And above all, it works nicely with our native schema indexes.


community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/FlippableIndexProxy.java
```diff
@@ -402,6 +404,10 @@ public void flip( Callable<Void> actionDuringFlip, FailedIndexProxyFactory failu
             {
                 actionDuringFlip.call();
                 this.delegate = flipTarget.create();
+                if ( started )
+                {
+                    this.delegate.start();
```
klaren, line 409: It's not clear why we need to call `delegate.start()` twice

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/PopulatingIndexProxy.java
```diff
@@ -184,11 +183,5 @@ public String toString()
         public void close() throws IOException, IndexEntryConflictException
         {
         }
-
```
klaren, line 187: We could remove `PopulatingIndexUpdater`

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/command/IndexBatchTransactionApplier.java
```diff
@@ -65,24 +62,25 @@
 
     private List<NodeLabelUpdate> labelUpdates;
     private IndexUpdates indexUpdates;
+    private long txId;
```
klaren: Unused

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/EagerlyReversedTransactionCursor.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.List;
+
+import org.neo4j.kernel.impl.transaction.CommittedTransactionRepresentation;
+
+/**
+ * Eagerly exhausts a {@link TransactionCursor} and allows moving through it in reverse order.
+ * The idea is that this should only be done for a subset of a bigger transaction log stream, typically
+ * for one log file.
+ *
+ * For reversing a transaction log consisting of multiple log files {@link FileWiseReversedTransactionCursor}
```
klaren: `FileWiseReversedTransactionCursor` does not exists

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReadAheadChannel.java
```diff
@@ -29,14 +29,16 @@
...
+
 /**
  * A buffering implementation of {@link ReadableClosableChannel}. This class also allows subclasses to read content
  * spanning more than one file, by properly implementing {@link #next(StoreChannel)}.
  * @param <T> The type of StoreChannel wrapped
  */
 public class ReadAheadChannel<T extends StoreChannel> implements ReadableClosableChannel, PositionableChannel
 {
-    public static final int DEFAULT_READ_AHEAD_SIZE = 1024 * 4;
+    public static final int DEFAULT_READ_AHEAD_SIZE = toIntExact( kibiBytes( 4 ) );
```
klaren, line 41: :+1:

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
```diff
@@ -0,0 +1,177 @@
...
+        chunkStartOffsetIndex = offsetCursor;
+        totalSize = channel.position();
+
+        return offsets;
+    }
+
+    @Override
+    public boolean next() throws IOException
+    {
+        while ( !exhausted() )
```
klaren: `if` ?

community/kernel/src/main/java/org/neo4j/kernel/recovery/DefaultRecoverySPI.java
```diff
@@ -74,19 +73,21 @@ public LogPosition getPositionToRecoverFrom() throws IOException
...
+    }
 
-        transactionsToApply = new TransactionQueue( 10_000, ( first, last ) -> storageEngine.apply( first, RECOVERY ) );
-        recoveryVisitor = new RecoveryVisitor( transactionsToApply );
-
-        return recoveryVisitor;
+    @Override
+    public RecoveryApplier getRecoveryApplier( TransactionApplicationMode mode ) throws Exception
+    {
+        transactionsToApply = new TransactionQueue( 100, ( first, last ) -> storageEngine.apply( first, mode ) );
```
klaren: we don't need `transactionsToApply` as field anymore

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/GivenTransactionCursor.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+
+import org.neo4j.kernel.impl.transaction.CommittedTransactionRepresentation;
+
+public class GivenTransactionCursor implements TransactionCursor
+{
+    private int index = -1;
+    private final CommittedTransactionRepresentation[] transactions;
+
+    public GivenTransactionCursor( CommittedTransactionRepresentation... transactions )
```
klaren: can be `private`

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/ReversedMultiFileTransactionCursorTest.java
```diff
@@ -0,0 +1,173 @@
...
+        @SuppressWarnings( "unchecked" )
+        ThrowingFunction<LogPosition,TransactionCursor,IOException> result = mock( ThrowingFunction.class );
+        AtomicLong txId = new AtomicLong( 0 );
+        CommittedTransactionRepresentation[][] logs = new CommittedTransactionRepresentation[transactionCounts.length][];
+        for ( int logVersion = 0; logVersion < transactionCounts.length; logVersion++ )
+        {
+            logs[logVersion] = transactions( transactionCounts[logVersion], txId );
+        }
+
+        when( result.apply( any( LogPosition.class ) ) ).thenAnswer( new Answer<TransactionCursor>()
```
klaren: Can be replaced with a lambda

community/kernel/src/test/java/org/neo4j/kernel/RecoveryIT.java
```diff
@@ -210,6 +267,443 @@ public void recoveryShouldFixPartiallyAppliedSchemaIndexUpdates()
         }
     }
 
+    @SuppressWarnings( "rawtypes" )
```
klaren, line 270: Not needed

community/kernel/src/test/java/org/neo4j/test/TestGraphDatabaseFactory.java
```diff
@@ -233,17 +232,17 @@ public GraphDatabaseService newDatabase( @Nonnull Config config )
...
     }
 
-    static class TestGraphDatabaseFacadeFactory extends GraphDatabaseFacadeFactory
+    public static class TestGraphDatabaseFacadeFactory extends GraphDatabaseFacadeFactory
     {
         private final TestGraphDatabaseFactoryState state;
         private final boolean impermanent;
 
-        TestGraphDatabaseFacadeFactory( TestGraphDatabaseFactoryState state, boolean impermanent )
+        public TestGraphDatabaseFacadeFactory( TestGraphDatabaseFactoryState state, boolean impermanent )
```
klaren: `protected` is enough

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/schema/populator/NonUniqueDatabaseIndexPopulatingUpdaterTest.java
```diff
@@ -224,7 +204,7 @@ public void additionsDeliveredToIndexWriter() throws Exception
         String expectedString1 = documentRepresentingProperties( (long) 1, "foo" ).toString();
         String expectedString2 = documentRepresentingProperties( (long) 2, "bar" ).toString();
         String expectedString3 = documentRepresentingProperties( (long) 3, "qux" ).toString();
-        String expectedString4 = documentRepresentingProperties( (long) 4, "git", "bit" ).toString();
```
klaren: We can remove the other unnecessary casts, for symmetry :)

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/schema/populator/NonUniqueDatabaseIndexPopulatingUpdaterTest.java
```diff
@@ -247,7 +227,7 @@ public void changesDeliveredToIndexWriter() throws Exception
 
         String expectedString1 = documentRepresentingProperties( (long) 1, "after1" ).toString();
         String expectedString2 = documentRepresentingProperties( (long) 2, "after2" ).toString();
-        String expectedString3 = documentRepresentingProperties( (long) 3, "bit", "after2" ).toString();
```
klaren: Same here

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/inmemory/UpdateCapturingIndexProvider.java
```diff
@@ -0,0 +1,90 @@
...
+
+public class UpdateCapturingIndexProvider extends SchemaIndexProvider
+{
+    private final SchemaIndexProvider actual;
+    private final Map<Long,UpdateCapturingIndexAccessor> indexes = new ConcurrentHashMap<>();
+    private final Map<Long,Collection<IndexEntryUpdate>> initialUpdates;
+
+    public UpdateCapturingIndexProvider( SchemaIndexProvider actual, Map<Long,Collection<IndexEntryUpdate>> initialUpdates )
+    {
+        super( actual.getProviderDescriptor(), actual.priority() );
```
klaren: A copy constructor in `SchemaIndexProvider` might be more appropriate than having a public getter method

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/EagerlyReversedTransactionCursor.java
```diff
@@ -0,0 +1,87 @@
...
+ * should be used (it will use this class internally though).
+ *
+ * @see FileWiseReversedTransactionCursor
```
tinwelint: Oh I see that somehow these class references weren't updated properly.

klaren (changes requested):
Heavy stuff! :sunglasses: 
Very nice with the simplification during recovery. I just have a few comments related to the code.

Although not part of this PR, I think we should make the `process` method in `IndexUpdater` use a wildcard instead of raw type.

I also think it would make sense to set `started = false` in the stop methods where the `started` variable was introduced.

community/common/src/test/java/org/neo4j/test/rule/RepeatRule.java
```diff
@@ -73,9 +95,9 @@ public void evaluate() throws Throwable
     public Statement apply( Statement base, Description description )
     {
         Repeat repeat = description.getAnnotation( Repeat.class );
-        if ( repeat != null )
+        if ( repeat != null || defaultTimes > 1 )
         {
-            return new RepeatStatement( repeat.times(), base );
+            return new RepeatStatement( max( repeat != null ? repeat.times() : 1, defaultTimes ), base, description );
```
burqen: I would expect that the annotation would overrule the default, even if it is smaller.

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexUpdater.java
```diff
@@ -38,6 +37,4 @@
 
     @Override
     void close() throws IOException, IndexEntryConflictException;
-
-    void remove( PrimitiveLongSet nodeIds ) throws IOException;
```
burqen, line 42: Good riddance! 🎉

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
```diff
@@ -366,25 +367,32 @@ protected BatchTransactionApplierFacade applier( TransactionApplicationMode mode
...
 
         // Perform the application
         return new BatchTransactionApplierFacade(
                 appliers.toArray( new BatchTransactionApplier[appliers.size()] ) );
     }
 
+    private LockService lockService()
+    {
+        return started ? lockService : NO_LOCK_SERVICE;
+    }
```
burqen: The reason we do not need the `lockService` before start is because recovery is single threaded, right?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/command/NeoStoreTransactionApplier.java
```diff
@@ -95,37 +99,36 @@ else if ( command.getRelId() != -1 )
             lockGroup.add( lockService.acquireRelationshipLock( command.getRelId(), LockService.LockType.WRITE_LOCK ) );
         }
 
-        // track the dynamic value record high ids
-        // update store
-        neoStores.getPropertyStore().updateRecord( command.getAfter() );
+        PropertyRecord select = version.select( command );
+        neoStores.getPropertyStore().updateRecord( select );
```
burqen: Inline this to use same patter everywhere.

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
```diff
@@ -82,11 +99,27 @@ public void init() throws Throwable
         }
 
         monitor.recoveryRequired( recoveryFromPosition );
+        spi.startRecovery();
+
+        // Backwards for neo store only
+        long checkpointTxId = TransactionIdStore.BASE_TX_ID;
+        try (   TransactionCursor transactionsToRecover = spi.getTransactionsInReverseOrder( recoveryFromPosition );
+                RecoveryApplier recoveryVisitor = spi.getRecoveryApplier( REVERSE_RECOVERY ); )
```
burqen: Three spaces here may be a little over the top.

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
```diff
@@ -82,11 +99,27 @@ public void init() throws Throwable
...
+
+        monitor.reverseStoreRecoveryCompleted( checkpointTxId );
 
+        // Forward with all appliers
         LogPosition recoveryToPosition;
         CommittedTransactionRepresentation lastTransaction = null;
-        Visitor<CommittedTransactionRepresentation,Exception> recoveryVisitor = spi.startRecovery();
-        try ( TransactionCursor transactionsToRecover = spi.getTransactions( recoveryFromPosition ) )
+        try (   TransactionCursor transactionsToRecover = spi.getTransactions( recoveryFromPosition );
+                RecoveryApplier recoveryVisitor = spi.getRecoveryApplier( RECOVERY ); )
```
burqen: Alright, now we got no more spaces to spare, used up the entire buffer of spaces we had. And winter is coming... How will we manage.

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
```diff
@@ -82,11 +99,27 @@ public void init() throws Throwable
...
+
+        // Backwards for neo store only
+        long checkpointTxId = TransactionIdStore.BASE_TX_ID;
+        try (   TransactionCursor transactionsToRecover = spi.getTransactionsInReverseOrder( recoveryFromPosition );
+                RecoveryApplier recoveryVisitor = spi.getRecoveryApplier( REVERSE_RECOVERY ); )
+        {
+            while ( transactionsToRecover.next() )
+            {
+                recoveryVisitor.visit( transactionsToRecover.get() );
+                checkpointTxId = transactionsToRecover.get().getCommitEntry().getTxId();
```
burqen: We are only using `checkpointTxId` to report to monitor that "this is how far back I reversed the store", right? A better name for that variable would be something like `reversedToTxId`.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/FlippableIndexProxy.java
tinwelint: Because the line above reassigns `delegate`, which hasn't seen the `start()` call before.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/PopulatingIndexProxy.java
tinwelint: Hmm, how could we do that?

burqen (changes requested):
My first review round is finished.

I really like those changes! Code is well documented which is awesome 👍 
Most comments are about readability, formatting or straight up questions about my understanding. 

I have not yet reviewed:
- `ReversedMultiFileTransactionCursor`
- `ReversedSingleFileTransactionCursor`
- Most tests

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
tinwelint: I think you're correct

community/kernel/src/test/java/org/neo4j/kernel/RecoveryIT.java
tinwelint: Well, it hides warnings about IndexEntryUpdate tho

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/schema/populator/NonUniqueDatabaseIndexPopulatingUpdaterTest.java
tinwelint: Picky, but it's good to be :)

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/PopulatingIndexProxy.java
klaren: Introducing a new class to save 3 lines of code just seems like an unnecessary indirection. But it doesn't hurt :)

community/common/src/test/java/org/neo4j/test/rule/RepeatRule.java
tinwelint: I would too

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
tinwelint: Just wanted to align both those. Easier on the eye imho.

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
tinwelint: Right, since it's the last checkpoint or the first transaction id ever. Or `lowestRecoveredTxId`?

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
tinwelint: No likey my feng shui?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/PopulatingIndexProxy.java
tinwelint: Hah, I thought you meant removing the updating alltogether :) I didn't introduce that class, or did I?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
```diff
@@ -0,0 +1,177 @@
...
+        while ( transactionCursor.next() )
+        {
+            if ( offsetCursor == offsets.length )
+            {   // Grow
+                offsets = Arrays.copyOf( offsets, offsetCursor * 2 );
+            }
+            offsets[offsetCursor++] = startOffset;
+            long nextStartOffset = channel.position();
+            assert nextStartOffset > startOffset : "It looks like this channel moved over to another version";
+            startOffset = channel.position();
```
burqen: Is there a reason for not using `nextStartOffset` when reassigning `startOffset`? Could `channel.position()` have been changed?

Also, the `assert`, is that just a sanity check while testing?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
```diff
@@ -0,0 +1,177 @@
...
+ * per chunk "move". This is all internal, so from the outside it simply reverses a transaction log.
+ * The memory overhead compared to reading a log in the natural order is almost negligible.
+ *
+ * This cursor currently only works for a single log file, such that the given {@link ReadAheadLogChannel} should not be
+ * instantiated with a {@link LogVersionBridge} moving it over to other versions when exhausted. For reversing a whole
+ * log stream consisting of multiple log files have a look at {@link ReversedMultiFileTransactionCursor}.
+ *
+ * @see ReversedMultiFileTransactionCursor
+ */
+class ReversedSingleFileTransactionCursor implements TransactionCursor
```
burqen: Sweet javadoc! Especially being consistent with using the same vocabulary used in the code, `sketch out` 🎉 

What would make it even better is a piece of ascii art explaining the variables, `chunkStartOffsetIndex`, `chunkLength`, `chunkCursor`, etc.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
```diff
@@ -0,0 +1,177 @@
...
+        // Start at lowOffsetIndex - 1 and count backwards until almost reaching the chunk size
+        long highOffset = chunkStartOffsetIndex == offsetsLength ? totalSize : offsets[chunkStartOffsetIndex];
+        int newLowOffsetIndex = chunkStartOffsetIndex;
+        while ( newLowOffsetIndex > 0 )
+        {
+            long deltaOffset = highOffset - offsets[--newLowOffsetIndex];
+            if ( deltaOffset > CHUNK_SIZE )
+            {   // We've come too far
+                if ( chunkStartOffsetIndex - newLowOffsetIndex > 1 )
+                {   // Don't include last this transaction, since it wasn't the only one
```
burqen: This sentence could use some love.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
```diff
@@ -0,0 +1,177 @@
...
+    @Override
+    public CommittedTransactionRepresentation get()
+    {
+        return chunkTransactions.item( chunkLength - chunkCursor );
+    }
+
+    @Override
+    public LogPosition position()
+    {
+        throw new UnsupportedOperationException( "Perhaps not required" );
```
burqen: This guy could use a better error message.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/command/Command.java
```diff
@@ -91,6 +91,28 @@ public static Mode fromRecordState( AbstractBaseRecord record )
         }
     }
 
+    public enum Version
```
klaren: A javadoc about what it is and what it's used for would be nice

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedMultiFileTransactionCursor.java
```diff
@@ -0,0 +1,143 @@
...
+        while ( currentLogTransactionCursor == null || !currentLogTransactionCursor.next() )
+        {
+            // We've run out of transactions in this log version, back up to a previous one
+            currentVersion--;
+            if ( currentVersion < backToPosition.getLogVersion() )
+            {
+                return false;
+            }
+
+            closeCurrent();
```
burqen, line 116: When `currentVersion < backToPosition.getLogVersion()` we will leave current cursor open. I guess it is safe for us to rely on `this.close()` to close it. Just double checking.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/inmemory/InMemoryIndexProvider.java
```diff
@@ -114,4 +114,17 @@ public InMemoryIndexProvider snapshot()
         }
         return new InMemoryIndexProvider( priority, copy );
     }
+
+    public boolean dataEquals( InMemoryIndexProvider other )
+    {
+        for ( Map.Entry<Long, InMemoryIndex> entry : indexes.entrySet() )
```
burqen: Formatting

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexingServiceTest.java
```diff
@@ -682,84 +677,16 @@ private boolean allInState( IndexingService indexing, InternalIndexState state,
...
             {
-                throw new UnsupportedOperationException();
-            }
-
-            @Override
-            public void collectUpdatedNodeIds( final PrimitiveLongSet target )
-            {
-                target.addAll( nodeIds.iterator() );
+                List<IndexEntryUpdate<LabelSchemaDescriptor>> updates = new ArrayList<>();
+                nodeIds.visitKeys( nodeId -> updates.add( IndexEntryUpdate.add( 0, index.schema(), Values.of( 1 ) ) ) );
```
burqen: Surprised we don't use `nodeId` here instead of `0`. We will generate identical index updates for each `nodeId` in set, all of which target node zero.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexingServiceTest.java
```diff
@@ -682,84 +677,16 @@ private boolean allInState( IndexingService indexing, InternalIndexState state,
...
-        IndexUpdates updates = nodeIdsAsIndexUpdates( nodeIds );
-        indexing.apply( updates );
-        life.start();
-
-        // Then
-        assertTrue( "applyingRecoveredData was not called", applyingRecoveredDataCalled.get() );
-        assertTrue( "recoveredUpdatesApplied was not called", recoveredUpdatesAppliedCalled.get() );
-    }
-
     private IndexUpdates nodeIdsAsIndexUpdates( PrimitiveLongSet nodeIds )
```
burqen: This could take vararg argument instead.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/EagerlyReversedTransactionCursor.java
```diff
@@ -0,0 +1,87 @@
...
+ * for one log file.
+ *
+ * For reversing a transaction log consisting of multiple log files {@link ReversedMultiFileTransactionCursor}
+ * should be used (it will use this class internally though).
+ *
+ * @see ReversedMultiFileTransactionCursor
+ */
+class EagerlyReversedTransactionCursor implements TransactionCursor
+{
+    private final List<CommittedTransactionRepresentation> txs = new ArrayList<>();
```
klaren, line 40: Wouldn't a stack/linked list be more appropriate?

tinwelint: More garbage and slower tho, so I don't see the benefit really.

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursorTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        // when
+        CommittedTransactionRepresentation[] readTransactions = readAllFromReversedCursor();
+
+        // then
+        assertTransactionRange( readTransactions, txId, TransactionIdStore.BASE_TX_ID );
+    }
+
+    @Test
+    public void shouldHandleLargeTransactions() throws Exception
```
burqen, line 104: May I also suggest adding an `IT` that test `VeryLarge` transactions? I mean the kind of transactions that touch every node in the store.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/EagerlyReversedTransactionCursor.java
tinwelint: Also, this class is a fallback for cases where the channel isn't a `ReadAheadLogChannel` so I'm not even sure this is ever needed. Could likely be removed at some point when/if we can verify that fact.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
```diff
@@ -0,0 +1,177 @@
...
+        long startOffset = channel.position();
+        while ( transactionCursor.next() )
+        {
+            if ( offsetCursor == offsets.length )
+            {   // Grow
+                offsets = Arrays.copyOf( offsets, offsetCursor * 2 );
+            }
+            offsets[offsetCursor++] = startOffset;
+            long nextStartOffset = channel.position();
+            assert nextStartOffset > startOffset : "It looks like this channel moved over to another version";
```
klaren: I don't believe this is a valid check and relies too much on the underlying channel. The `transactionCurson` has a method `position()` that returns the current log version.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
```diff
@@ -0,0 +1,177 @@
...
+ */
+class ReversedSingleFileTransactionCursor implements TransactionCursor
+{
+    // Should this be passed in or extracted from the read-ahead channel instead?
+    private static final int CHUNK_SIZE = ReadAheadChannel.DEFAULT_READ_AHEAD_SIZE;
+
+    private final ReadAheadLogChannel channel;
+    private final TransactionCursor transactionCursor;
+    // Should be generally large enough to hold transactions in a chunk, where one chunk is the read-ahead size of ReadAheadLogChannel
+    private final ArrayCollection<CommittedTransactionRepresentation> chunkTransactions = new ArrayCollection<>( 20 );
```
klaren: Would prefer to see `ArrayDeque` with `push() pop() peek()` to ensure it's LIFO. The performance gain with `ArrayCollection` is negligible comparing to readability.

klaren (commented):
A few more comments, 2 was accidentally added as single comments instead of as review comments.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/EagerlyReversedTransactionCursor.java
klaren: `ArrayDeque` would have the same impact as an ArrayList, but you don't need the private int to keep track of index. But true, I could not find a case where this fallback would be used either.

tinwelint: Right, `ArrayDeque` looks like a good candidate actually. But let's keep it as is since it will probably go away... I think it should.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
tinwelint: That check is just an assertion to spot cases where you may have passed in a channel with some log version bridge so that it traverses multiple log versions. The check isn't fully covering tho, but I don't think it's invalid. Or perhaps we could check the log version set on the channel instead, to see that it stays the same.

klaren: I could imagine a couple of perverse cases with strange log rotation settings that would cause this to fail. Checking the log version on the channel seems like a wonderful idea :+1:

community/kernel/src/test/java/org/neo4j/kernel/RecoveryIT.java
```diff
@@ -210,6 +267,443 @@ public void recoveryShouldFixPartiallyAppliedSchemaIndexUpdates()
...
+    @Test
+    public void shouldSeeSameIndexUpdatesDuringRecoveryAsFromNormalIndexApplication() throws Exception
```
burqen, line 272: This is some sweet testing!

community/kernel/src/test/java/org/neo4j/kernel/RecoveryIT.java
```diff
@@ -210,6 +267,443 @@ public void recoveryShouldFixPartiallyAppliedSchemaIndexUpdates()
...
+        // then
+        assertEquals( lastCommittedTxIdBeforeRecovered, lastCommittedTxIdAfterRecovered );
+        assertTrue( indexStateAtCrash.dataEquals( indexStateAtLastCheckPoint /*which then participated in recovery*/ ) );
+        assertSameUpdates( updatesAtCrash, updatesAfterRecovery );
+        db.shutdown();
+        crashedFs.close();
+    }
+
+    @Test
+    public void shouldSeeTheSameRecordsAtCheckpointAsAfterReverseRecovery() throws Exception
```
burqen, line 343: And also this! 👍

burqen (changes requested):
Second part of first review round finished! 🎉 

There are some small things to look at, but overall this looks very good!

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
burqen: Auto format will remove the extra spaces from both lines. At least for me 😄  doesn't work for you?

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
burqen: Copy from above: Auto format will remove the extra spaces from both lines. At least for me 😄  doesn't work for you?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
tinwelint: Right OK, although this has just changed, removing `chunkLength` and `chunkCursor`. I love ASCII art and can provide some if it still makes sense.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
tinwelint: Nope, it's a mistake. The assert has changed to a real check+exception of log version at start and end to make sure the channel doesn't span multiple log versions underneath.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
tinwelint: Right, changed it and actually removed that condition because it doesn't improve anything regarding reading performance, just increases complexity.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedMultiFileTransactionCursor.java
tinwelint: Yup, that was the idea

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexingServiceTest.java
tinwelint: I'm unsure what that test did even... mostly checking that any update was generated. Changed it anyway.

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursorTest.java
tinwelint: Oh OK, define very large then.... 10 gigabytes?

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
tinwelint: I'd guess we're passed point of diminishing returns when discussing spaces, but will change the spaces for this one

tinwelint: `Ctrl+Shift+F` in Eclipse puts them aligned and in the right place it seems :)

klaren (approved):
I'm happy with the current changes :tada:

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursorTest.java
burqen: Yep, that should be enough!

burqen:
@tinwelint Looks good, approved!

burqen (approved):

burqen:
Let's wait for it to go green though

tinwelint:
It's green enough

PR #9954, opened 2017-08-31
Free id keeper enhancement

martinfurmanski:


community/io/src/main/java/org/neo4j/io/fs/OffsetChannel.java
```diff
@@ -0,0 +1,164 @@
...
+    @Override
+    public void flush() throws IOException
+    {
+        delegate.flush();
+    }
+
+    @Override
+    public long write( ByteBuffer[] srcs, int offset, int length ) throws IOException
+    {
+        return delegate.write( srcs, offset, length );
```
MishaDemianenko, line 132: not sure if it matters, but this delegation does not really follow the same offset pattern. Maybe keep it as unsupported operation then?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
```diff
@@ -22,90 +22,122 @@
...
+     * when IDs from old runs get reused and newly freed IDs are put on top
+     * of the stack. During a clean shutdown the gap will be compacted away.
+     *
+     * During an aggressive run a gap is never formed since batches of free
+     * IDs are flushed on top of the stack (end of file) and also read in
+     * from the top of the stack.
      */
-    private long maxReadPosition;
-    private long readPosition; // the place from where we read. Always <= maxReadPosition
+    private long prePosition;
```
MishaDemianenko: I would call it something like `initialPosition` instead

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
```diff
@@ -22,90 +22,122 @@
...
+
+    private long freeIdCount;
+
+    /**
+     * Keeps the position where batches of IDs will be flushed out to.
+     * This can be viewed as being put on top of a stack.
      */
-    private final boolean aggressiveReuse;
-    private long defraggedIdCount;
+    private long stackPosition;
```
MishaDemianenko, line 63: and this I would call writerPosition

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
```diff
@@ -22,90 +22,122 @@
...
 public class FreeIdKeeper implements Closeable
 {
-    public static final long NO_RESULT = -1;
-    public static final int ID_ENTRY_SIZE = Long.BYTES;
-    private final LinkedList<Long> freeIds = new LinkedList<>();
-    private final LinkedList<Long> readFromDisk = new LinkedList<>();
+    private static final int ID_ENTRY_SIZE = Long.BYTES;
+
+    private final List<Long> freeIds = new ArrayList<>();
+    private final List<Long> readFromDisk = new ArrayList<>();
```
MishaDemianenko: looking into how those used, maybe ArrayDeque will much better here

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
```diff
@@ -116,155 +148,165 @@ else if ( defraggedIdCount > 0 && canReadMoreIdBatches() )
...
+            throw new UnderlyingStorageException( "Failed reading free id batch", e );
+        }
     }
 
-    /*
-     * After this method returns, if there were any entries found, they are placed in the readFromDisk list and the
-     * readPosition is updated accordingly.
-     */
-    private void readIdBatch()
+    private boolean readIdBatch0() throws IOException
```
MishaDemianenko, line 169: Not really a fun of `method0` patterns, maybe `methodImpl` instead?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
```diff
@@ -116,155 +148,165 @@ else if ( defraggedIdCount > 0 && canReadMoreIdBatches() )
...
-            int bytesRead;
-            do
+            // there is no compaction to be done
+            return;
+        }
+
+        long writePosition = stackPosition;
+        long readPosition = prePosition;
+        int nBytes;
+        do
```
MishaDemianenko: not quite sure how it works for cases when write position is higher then read position

MishaDemianenko (changes requested):
Great simplification, only have couple of comments

community/io/src/main/java/org/neo4j/io/fs/OffsetChannel.java
martinfurmanski: This offset is an offset into the buffer. There is no need to offset the position since this method does not have a position parameter. Whatever the position was from before will be used.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
martinfurmanski: It's quite common.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
martinfurmanski: Better in what way?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
martinfurmanski: Stack is more descriptive.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
martinfurmanski: same same

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
martinfurmanski: That cannot happen because this is only for the regular mode. The stack can only grow and the reader can only consume down towards 0. The initial state is that they are equal and if they still are when closing/compacting then there is an early return. In all other cases the batches will be moved from prePosition to stackPosition.

community/io/src/main/java/org/neo4j/io/fs/OffsetChannel.java
MishaDemianenko: Javadoc states that this channel is a channel with offset, this method and all other that performs operations on ByteBuffer[] are ignoring those which can be confusing

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
MishaDemianenko: we are interesting in first/last elements most of the time and ArrayDeque will have handy methods that will do exactly that, in addition to that based on Deque interface readers can already understand what kind of operations we are doing with those list of ids

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
MishaDemianenko: ¯\_(ツ)_/¯

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
MishaDemianenko: i know, it does force me to like it, he

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
MishaDemianenko: yeah now i see, would you mind to add that additional assertion still into that method that `stackPosition` should be less than `prePosition`

community/io/src/main/java/org/neo4j/io/fs/OffsetChannel.java
martinfurmanski: I think it is very possible that it is slightly confusing from implementation, but not from the interface though. It should be clear from the interface that the write methods that do not get position as a parameter use the current position. The current position is properly offset by any methods that adjust the position or get the position as a secondary parameter.

This class was just an attempt to improve upon having to think about offsets within the FreeIdKeeper implementation. It was unnecessary to have that extra complexity in there. I don't particularly like the OffsetChannel as such, but it is better than the alternative.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
martinfurmanski: Definitely!

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
martinfurmanski: Cool, I will update.

PR #9956, opened 2017-08-31
Corrupted logs resilient recovery

MishaDemianenko:
Allow recovery process to restore store with corrupted transaction logs.
Transaction logs can be corrupted in various scenarios, most frequently
that happens when out of disk space error occurred.
This PR introduces additional resilience in log reading during recovery
that will allow restoring as many transactions as the database can read.
Any further logs after the first nonreadable transaction are ignored and nonreadable as well.
All unreadable transactions logs will be truncated at the end of recovery.

To avoid cases when recovery can remove transactions logs during downgrade
that can happen by mistake additional "force" option introduced.
That will not gonna allow to remove logs with the unknown version by default.


tinwelint:
So these changes makes latest checkpoint finder log instead of throwing on broken log entries. Will recovery also be able handle this? Also any chance something like [LenientInvalidLogEntryHandler.java](https://github.com/neo4j/neo4j/blob/3.3/tools/src/main/java/org/neo4j/tools/dump/LenientInvalidLogEntryHandler.java) would be handy?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/LogTailScanner.java
```diff
@@ -147,28 +174,46 @@ private LogTailInformation update() throws IOException
...
+        long firstTxIdAfterCheckPoint = LogTailInformation.NO_TRANSACTION_ID;
+        if ( startEntryAfterCheckPoint )
+        {
+            try
+            {
+                firstTxIdAfterCheckPoint = extractFirstTxIdAfterPosition( target, fromVersionBackwards );
+            }
+            catch ( LogReadingException lre )
+            {
+                //
```
tinwelint: Should we silently accept such failure always, or only on "force"? This is changing default behaviour isn't it?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/LogTailScanner.java
```diff
@@ -147,28 +174,46 @@ private LogTailInformation update() throws IOException
...
-                startEntryAfterCheckPoint = extractFirstTxIdAfterPosition( target, version ) !=
-                        LogTailInformation.NO_TRANSACTION_ID;
+                try
+                {
+                    startEntryAfterCheckPoint =
+                            extractFirstTxIdAfterPosition( target, version ) != LogTailInformation.NO_TRANSACTION_ID;
+                }
+                catch ( LogReadingException lre )
+                {
+                    startEntryAfterCheckPoint = true;
```
tinwelint: Should we silently accept such failure always, or only on "force"? This is changing default behaviour isn't it?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/ReversedSingleFileTransactionCursor.java
```diff
@@ -89,14 +96,21 @@
...
+                }
+                offsets[offsetCursor++] = startOffset;
+                startOffset = channel.position();
             }
-            offsets[offsetCursor++] = startOffset;
-            startOffset = channel.position();
+        }
+        catch ( Throwable t )
+        {
+            log.warn( buildReadErrorMessage( offsets, offsetCursor, logVersion ), t );
```
tinwelint: It looks like the sketching just aborts, leaving some variables not properly initialized and still not throwing. Can't this make the upcoming reverse reading start at the wrong place?

tinwelint (changes requested):
There's something about the logging approach that makes me shrug. All of a sudden there are critical log reading classes which may fail for various reasons and still not throw, where the only way to find that out outside the class is to look at log messages. In such internal and critical classes we should not use this opaque. Also if failing, swallowing and just continue the execution may leave some state in unexpected state and I don't think that's properly tested, nor easily made to work for all cases.

To remove the hard dependency on logging internally another way would be to pass in a monitor receiving the exceptions and it could decide to fail or log or perhaps both, because I do think that the default behaviour should absolutely be to fail and only be made to log and continue (dangerously) on an explicit force flag setting.

MishaDemianenko:
@tinwelint do not really understand roots of your concern at all, extensive logging used only to preserve as much as possible information into log files for future analysis or investigation nothing more and there is nothing wrong with it I can see only tons of advantages from that.
All of the readers/locators/cursors use either position/tail information/presence or absence of record with properly recorded location to signal validity, presence and read ability of record and that is more than enough to treat result of log parsing, especially to treat last succesfull position.
Once again there is no hard dependency on logging anywhere, ever.

MishaDemianenko:
@tinwelint should be ready for a new round

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -228,6 +235,8 @@ boolean applicable( DiagnosticsPhase phase )
     }
 
     public static final String DEFAULT_DATA_SOURCE_NAME = "nioneodb";
+    private final boolean failOnCorruptedLogFiles = FeatureToggles.flag( NeoStoreDataSource.class,
+            "failOnCorruptedLogFiles", false );
```
tinwelint, line 239: A default value of `false` is potentially unsafe isn't it? I see this recovery feature as something like:

- You do recovery and for some reason it fails
- You realize that there may be some corruption and perhaps do some preparation work, backup, copy whatever
- You do recovery with this flag to force attempt recovery anyway

community/kernel/src/main/java/org/neo4j/kernel/impl/logging/SimpleLogService.java
```diff
@@ -27,6 +27,12 @@
     private final LogProvider userLogProvider;
     private final LogProvider internalLogProvider;
 
+    public SimpleLogService( LogProvider commonLogProvider )
+    {
+        this.userLogProvider = commonLogProvider;
```
tinwelint: So this constructor should work differently than the one below? The one below wraps in `DuplicatingLogProvider`. If they should differ then please document and explain exactly why and which you should use in different scenarios.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/reverse/ReversedTransactionCursorMonitor.java
```diff
@@ -0,0 +1,25 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.transaction.log.reverse;
+
+public interface ReversedTransactionCursorMonitor
```
tinwelint, line 22: I'm usually much in favor of having these interfaces inside the class that uses them... much easier to understand IMHO.

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
```diff
@@ -38,116 +37,94 @@
  */
 public class Recovery extends LifecycleAdapter
 {
-    public interface Monitor
```
tinwelint, line 41: I'm not a fan of moving out interfaces that are very tied to its parent outside. Why this refactoring?

community/kernel/src/main/java/org/neo4j/kernel/recovery/TransactionLogPruner.java
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Transaction log pruner used during recovery to truncate all the logs after some specified position, that
+ * recovery threats as corrupted or non-readable.
+ * Transaction log file specified by provided log position will be truncated to provided length, any
+ * subsequent files will be removed.
+ * Any removed or modified log content will be stored in separate corruption logs archive for further analysis and as
+ * an additional safety option to have the possibility to fully restore original logs in a faulty case.
+ */
+public class TransactionLogPruner
```
tinwelint: Terminology-wise this collides with transaction log pruning, i.e. removing old logs that have been rotated out. A better term would be something about truncation, wouldn't it?

tinwelint (changes requested):
Had some comments. Good with the added monitor approach. I'm not a fan of the accidental refactorings around pulling out internal interfaces to their own classes and renaming them. I think that's a matter of taste too, isn't it? It may better to have that as separate PR, if at all, to discuss that there instead of going back and forth about these changes that aren't actually required.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
MishaDemianenko: i do not think we should ask user confirmation for recovery it should just work, also consider scenarios like cloud, docker containers auto restarts, copying of tx logs over network etc - that should not involve any user interactions at all. And since we have a first safety mechanic to backup files that will be truncated and the fact that we doing that in brand new version and will have time to catch possible problems there i think it should be "false" by default

MishaDemianenko:
Those renamings where not that accidental, some of them were squashed into the first commit, which is a bad thing thats true.  All other I have tried to do in different commits, but in general, it will be quite hard to extract them. Those changes allowed to split and separate different parts of recovery and extract interfaces and services that are supplied to recovery or by different components to recovery into a separate interfaces/classes/entities That in the same moment to simplify a lot of things. And since it's a new version I think it's good to rename some sort of old things and to do a cleanup. And if this change will be backported ever those changes will not gonna be backported.
But that indeed can be a thing to avoid and extract into separate PRs instead. Noted.

MishaDemianenko:
@tinwelint updated with renaming and some javadoc clarifications

tinwelint:
Yeah, actually you're right about the default of it, that it's going to be a problem for some scenarios like cloud and that it should "just work".

community/kernel/src/main/java/org/neo4j/kernel/recovery/CorruptedLogsTruncator.java
```diff
@@ -0,0 +1,167 @@
...
+import org.neo4j.io.fs.FileSystemAbstraction;
+import org.neo4j.io.fs.StoreChannel;
+import org.neo4j.kernel.impl.transaction.log.LogPosition;
+import org.neo4j.kernel.impl.transaction.log.PhysicalLogFiles;
+
+import static java.lang.String.format;
+
+/**
+ * Transaction log truncator used during recovery to truncate all the logs after some specified position, that
+ * recovery threats as corrupted or non-readable.
```
tinwelint: `threats` --> `treats`

community/kernel/src/main/java/org/neo4j/kernel/recovery/CorruptedLogsTruncator.java
```diff
@@ -0,0 +1,167 @@
...
+            {
+                byteBuffer.flip();
+                destination.write( byteBuffer.array(), byteBuffer.position(), byteBuffer.remaining() );
+                byteBuffer.clear();
+            }
+        }
+        destination.closeEntry();
+    }
+
+    private boolean isNotLastLogFile( long recoveredTransactionLogVersion )
```
tinwelint: Let's try and use positive naming/results, like `isLastLogFile`

community/kernel/src/main/java/org/neo4j/kernel/recovery/LogTailScanner.java
```diff
@@ -248,26 +258,71 @@ public LogTailInformation getTailInformation() throws UnderlyingStorageException
...
+        {
+            this.id = id;
+        }
+
+        public boolean isFailure()
+        {
+            return failure;
+        }
+
+        public void setFailure( boolean failure )
```
tinwelint: I don't see why this class needs to have publically accessable setters instead of final fields and set them in constructor.

community/kernel/src/main/java/org/neo4j/kernel/recovery/CorruptedLogsTruncator.java
```diff
@@ -0,0 +1,167 @@
...
+ * Transaction log truncator used during recovery to truncate all the logs after some specified position, that
+ * recovery threats as corrupted or non-readable.
+ * Transaction log file specified by provided log position will be truncated to provided length, any
+ * subsequent files will be removed.
+ * Any removed or modified log content will be stored in separate corruption logs archive for further analysis and as
+ * an additional safety option to have the possibility to fully restore original logs in a faulty case.
+ */
+public class CorruptedLogsTruncator
+{
+    public static final String CORRUPTED_TX_LOGS_FOLDER_NAME = "corrupted-tx-logs";
```
tinwelint: Should this rhyme with how the transaction log files are currently called? They're currently called `neostore.transaction.db.v???`

tinwelint (commented):
Had some more minor comments... I think it generally looks good.

MishaDemianenko:
@tinwelint comments addressed

tinwelint (approved):
Awesome!

PR #9958, opened 2017-09-01
Rename 'legacy index' to 'explicit index'

klaren:


eebus (commented):
I have looked all the .scala files and they are approved.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_2/Compatibility.scala
```diff
@@ -146,8 +146,8 @@ trait Compatibility[C <: CompilerContext] {
         case SchemaIndexScanUsage(identifier, label, propertyKey) =>
           val labelId = transactionalContext.readOperations.labelGetForName(label)
           schemaIndexUsage(identifier, labelId, label, propertyKey)
-        case LegacyNodeIndexUsage(identifier, index) => legacyIndexUsage(identifier, "NODE", index)
-        case LegacyRelationshipIndexUsage(identifier, index) => legacyIndexUsage(identifier, "RELATIONSHIP", index)
+        case LegacyNodeIndexUsage(identifier, index) => explicitIndexUsage(identifier, "NODE", index)
```
ragadeeshu, line 149: This is a part of 3.2 cypher and thus unchangeable?

ragadeeshu (commented):
Looks good overall. There is a mention of a "legacy auto index" in the javadoc for org.neo4j.kernel.api.explicitindex.AutoIndexOperations, is this correct? Also, I can imagine that this is quite a big change for docs? Are they in the loop?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_2/Compatibility.scala
klaren: Yes, it would involve some rather unpleasant updates to the old cypher compiler. And this is the only bridge that's needed.

klaren:
@ragadeeshu Good question, I thought that legacy referred to the automatic indexing, not the index it self. Could be wrong though. And yes, the documentation team is in the loop.

ragadeeshu (approved):

PR #9961, opened 2017-09-01
Heavily reduces contention on generating ids

tinwelint:
The main idea is to have batching id generators local to each transaction instance (which is pooled).
This batching sits in between store and typical creators of records at commit time and hands out ids,
grabbing batches of ids off of the stores. On clean shutdown any excess ids remaining unused will be
freed back into the store before closing.

In scenarios where there are many concurrent transactions committing data, where they all
co-operatively force the log then this id generator contention is the single biggest one.
With these changes this contention is basically removed entirely.

The batch size is controllable with a (not yet fully finalized) database setting `unsupported.dbms.record_id_batch_size`.


tinwelint:
Replaces #9776

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/ReplicatedIdGenerator.java
```diff
@@ -116,23 +118,40 @@ public synchronized long nextId()
...
+        if ( idBatch.totalSize() > 0 )
+        {
+            return idBatch;
+        }
+        IdRange range = idQueue.nextIdBatch( size );
+        if ( range.totalSize() == 0 )
+        {
+            acquireNextIdBatch();
+            range = idQueue.nextIdBatch( size );
+        }
```
MishaDemianenko: 'nextId()' call as last step incrementing id generator highId, shouldn't we also do it here?

community/cypher/cypher/src/test/java/org/neo4j/cypher/internal/javacompat/NotificationAcceptanceTest.java
```diff
@@ -59,7 +59,7 @@ public void shouldNotifyWhenUsingCypher3_1ForTheRulePlannerWhenCypherVersionIsTh
         InputPosition position = new InputPosition( 20, 1, 21 );
 
         // then
-        assertThat( result.getNotifications(), contains( RULE_PLANNER_UNAVAILABLE_FALLBACK.notification( position ) ) );
+        assertThat( result.getNotifications(), Matchers.<Notification>contains( RULE_PLANNER_UNAVAILABLE_FALLBACK.notification( position ) ) );
```
MishaDemianenko, line 62: not quite sure why those matchers were changes since it works with static imports

enterprise/kernel/src/test/java/org/neo4j/graphdb/store/id/IdReuseTest.java
```diff
@@ -47,7 +48,17 @@
 {
     @Rule
     public DatabaseRule dbRule = new EnterpriseDatabaseRule()
-            .withSetting( EnterpriseEditionSettings.idTypesToReuse, IdType.NODE + "," + IdType.RELATIONSHIP );
+            .withSetting( EnterpriseEditionSettings.idTypesToReuse, IdType.NODE + "," + IdType.RELATIONSHIP )
+            // Set tx id batching to 1 since there are sequential id testing which gets bitten by this inner working:
```
MishaDemianenko: i do not think this comment is helpful at all even now, and it will be less helpful over time when things will be updated

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -459,8 +460,17 @@
...
+            "Committing a transaction which contains changes most often results in new data records being created. " +
+            "For each record a new id needs to be generated from an id generator. " +
+            "It's more efficient to allocate a batch of ids from the contended id generator, which the transaction " +
+            "holds and generates ids from while creating these new records. " +
+            "This setting specifies how big those batches are. " +
+            "Remaining ids are freed back to id generator on clean shutdown." )
+    @Internal
+    public static final Setting<Integer> record_id_batch_size = setting( "unsupported.dbms.record_id_batch_size", INTEGER,
+            "20", range( 1, 1_000 ) );
+
```
MishaDemianenko: Suddenly description of `forced_kernel_id` setting disappeared.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -459,8 +460,17 @@
...
+    @Description( "Specifies the size of id batches local to each transaction when committing. " +
+            "Committing a transaction which contains changes most often results in new data records being created. " +
+            "For each record a new id needs to be generated from an id generator. " +
+            "It's more efficient to allocate a batch of ids from the contended id generator, which the transaction " +
+            "holds and generates ids from while creating these new records. " +
+            "This setting specifies how big those batches are. " +
+            "Remaining ids are freed back to id generator on clean shutdown." )
+    @Internal
+    public static final Setting<Integer> record_id_batch_size = setting( "unsupported.dbms.record_id_batch_size", INTEGER,
+            "20", range( 1, 1_000 ) );
```
MishaDemianenko: is 20 good enough value as default? also, do we need to limit it to 1000 as max?

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
```diff
@@ -266,6 +245,12 @@ public StoreReadLayer storeReadLayer()
     }
 
     @Override
+    public RecordStorageCommandCreationContext allocateCommandCreationContext()
```
MishaDemianenko, line 248: not quite sure that `RecordStorageEngine` should be the factory for those; do not have any good candidates at the same time, except maybe store statement/transaction itself

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
```diff
@@ -283,10 +268,13 @@ public void createCommands(
     {
         if ( txState != null )
         {
-            RecordChangeSet recordChangeSet = new RecordChangeSet( loaders );
-            TransactionRecordState recordState = new TransactionRecordState( neoStores, integrityValidator,
-                    recordChangeSet, lastTransactionIdWhenStarted, locks,
-                    relationshipCreator, relationshipDeleter, propertyCreator, propertyDeleter );
+            // We can make this cast here because we expected that the storageStatement passed in here comes from
```
MishaDemianenko, line 271: do not think this assumption is correct, you can easily create a subclass of the current engine and return something else instead. We should just add method createTransactionRecordState into the interface and in the worst case, we will just change it in future since that is fully internal.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdContainer.java
```diff
@@ -232,6 +232,24 @@ public long getReusableId()
...
+        while ( count < maxSize )
+        {
+            long id = freeIdKeeper.getId();
+            if ( id == FreeIdKeeper.NO_RESULT )
+            {
+                break;
+            }
+            tmpIdArr[count++] = id;
+        }
+        long[] defragIdArr = new long[count];
```
MishaDemianenko: `Arrays.copyOfRange(tmpIdArr, 0, count)` instead?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdRangeIterator.java
```diff
@@ -71,12 +76,44 @@ public long next()
...
+            sizeLeft -= numberOfDefrags;
+        }
+
+        long rangeStart = 0;
+        int rangeLength = 0;
+        int rangeOffset = currentRangeOffset();
+        int rangeAvailable = length - rangeOffset;
+        if ( sizeLeft > 0 && rangeAvailable > 0 )
+        {
+            rangeStart = start + rangeOffset;
```
MishaDemianenko, line 99: why do we do`+ rangeOffset` part here? isn't range start/length independent from defrag ids and we can just start from `start` here instead?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/RenewableBatchIdSequence.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.function.LongConsumer;
+
+import org.neo4j.graphdb.Resource;
+
+import static org.neo4j.kernel.impl.store.id.IdRangeIterator.VALUE_REPRESENTING_NULL;
+
+/**
+ * An {@link IdSequence} which does internal batching by using another {@link IdSequence} as source of batches.
+ * Meant to be used by a single thread during its life time only.
```
MishaDemianenko: javadoc here claims that class is single threaded only

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/RenewableBatchIdSequence.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public class RenewableBatchIdSequence implements IdSequence, Resource
+{
+    private final IdSequence source;
+    private final int batchSize;
+    private final LongConsumer excessIdConsumer;
+    private IdSequence currentBatch;
+    private boolean closed;
+
+    public RenewableBatchIdSequence( IdSequence source, int batchSize, LongConsumer excessIdConsumer )
```
MishaDemianenko: constructor can be package private

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/RenewableBatchIdSequence.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+        this.source = source;
+        this.batchSize = batchSize;
+        this.excessIdConsumer = excessIdConsumer;
+    }
+
+    /**
+     * It's dangerous to potentially have multiple concurrent calls to close w/ regards to freeing excessive ids.
```
MishaDemianenko: javadoc here claims that it's dangerous, and method is synchronized those two assumptions/theories should be aligned

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/RenewableBatchIdSequences.java
```diff
@@ -0,0 +1,75 @@
...
+        return types[type.ordinal()];
+    }
+
+    @Override
+    public void close()
+    {
+        for ( StoreType type : StoreType.values() )
+        {
+            IdSequence generator = idGenerator( type );
+            if ( generator instanceof RenewableBatchIdSequence )
```
MishaDemianenko, line 69: What if we will make all IdSequence's resources instead so we can avoid this cast?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StandardDynamicRecordAllocator.java
```diff
@@ -0,0 +1,55 @@
...
+        return dataSize;
+    }
+
+    @Override
+    public DynamicRecord nextRecord()
+    {
+        return allocateRecord( idGenerator.nextId() );
+    }
+
+    public static DynamicRecord allocateRecord( long id )
```
MishaDemianenko, line 48: this is looking really strange: we own id generator and only we should be responsible for next id generation in the same time there is a public method that can accept id from somebody else.
And single usage says that that method can be avoided

community/kernel/src/main/java/org/neo4j/storageengine/api/StorageStatement.java
```diff
@@ -157,4 +157,24 @@
...
+    /**
+     * Reserves a relationship id for future use to store a relationship. The reason for it being exposed here is that
+     * internal ids of nodes and relationships are publicly accessible all the way out to the user.
+     * This will likely change in the future though.
+     *
+     * @return a reserved relationship id for future use.
+     */
+    long reserveRelationship();
+
+    CommandCreationContext getCommandCreationContext();
```
MishaDemianenko: javadoc is missing

community/kernel/src/test/java/org/neo4j/kernel/impl/core/TestCrashWithRebuildSlow.java
```diff
@@ -158,7 +161,8 @@ public void crashAndRebuildSlowWithDynamicStringDeletions() throws Exception
                 nodes.add( node );
                 if ( previous != null )
                 {
-                    previous.createRelationshipTo( node, MyRelTypes.TEST );
+                    Relationship rel = previous.createRelationshipTo( node, MyRelTypes.TEST );
+                    System.out.println( rel );
```
MishaDemianenko: 🦅

community/primitive-collections/src/main/java/org/neo4j/collection/pool/MarshlandPool.java
```diff
@@ -108,40 +108,37 @@ public void release( T obj )
         // Return it locally if possible
         LocalSlot<T> localSlot = puddle.get();
 
-        if ( localSlot.object == null )
-        {
-            localSlot.set( obj );
+        if ( localSlot.release( obj ) )
```
MishaDemianenko:
```
if ( !localSlot.release( obj ) )
 {   // Fall back to the delegate pool
            pool.release( obj );
 }
```

MishaDemianenko (changes requested):
Couple of comments, also not sure that MarshlandPool changes are covered by tests at all

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/ReplicatedIdGenerator.java
tinwelint: @martinfurmanski wdyt?

community/cypher/cypher/src/test/java/org/neo4j/cypher/internal/javacompat/NotificationAcceptanceTest.java
tinwelint: Some compilation issue in my Eclipse environment. Not sure why, but I don't think this change hurts... if it now works for everyone then all good, right?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/id/ReplicatedIdGenerator.java
martinfurmanski: @tinwelint Sounds like a good idea

enterprise/kernel/src/test/java/org/neo4j/graphdb/store/id/IdReuseTest.java
tinwelint: Yupp, will remove it

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
tinwelint: 20 is some empirically chosen value... it roughly amounts to 1/20 of id contention, at least something in that area. Too big will cause too many ids in flight and lost on crash. Too few will not reduce id contention enough and the management around the id batches will be overkill. Also the 1000 limit, I'm guess we should, if anything, limit it to something less because I'm thinking about id batches size is clustering environments... which I believe is in those sizes and these tx batch sizes should not be as mig as those.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
tinwelint: How odd, fixed now

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
tinwelint: It has everything to do with the `StorageEngine` abstraction. I see no other candidate

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
tinwelint: I think this is a common pattern for explicitly allocating instances (which is a good thing) and pass in instead of having the method allocate and manage itself. We own this code completely so it's never a practical problem.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdRangeIterator.java
tinwelint: Do you see a case not covered by `IdRangeIteratorTest`?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StandardDynamicRecordAllocator.java
tinwelint: It's a static method allocating a `DynamicRecord` with a certain id, setting its `created` and `inUse` flags. What's strage with that? :)

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/RenewableBatchIdSequences.java
tinwelint: I'd say that's an unnecessary broadening of that interface for the purpose of this single use case. This cast feels less intrusive/odd than changing IsSequence itself imho.

tinwelint:
Replied and updated, have another look (I know there's failing tests, very few, left) but still.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
MishaDemianenko: yep it should not be less than 20 for sure i was thinking about making it bigger by default and put limitations to something like 100_000 or so, for cases when your workload are batches of different sorts, even if you have a single machine or you do not really care about lost ids

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
MishaDemianenko: do not think that casting with a long comment is a _common pattern_ or either practice of some good sort

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StandardDynamicRecordAllocator.java
MishaDemianenko: The Strange thing is not what it does, by why it publicly exposed in a current way.
StandardDynamicRecordAllocator has a reference to `idGenerator` and responsible for generating of correct `nextRecord` with an appropriate id that should be allocated by `idGenerator`. First strange thing - who, if not id generator, can generate a correct id for you. Another one is it's single usage in AbstractDynamicStore: `StandardDynamicRecordAllocator.allocateRecord( nextId() )` where that next id will be generated from idGenerator as well, so why to expose it at all

MishaDemianenko (approved):
A couple of minor comments but do not think those are that important to block approval. But feel free to react on those :)
Build is retriggered now

tinwelint:
@MishaDemianenko the build was green. Wanted to squash the commits about MarshlandPool into one before merge so did so... this will give another run in CI too.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
tinwelint: You may be correct about the upper limit. Although I'm fairly certain that the contention point in the majority of cases (mind that in most cases this id generator contention point isn't among the top points at all) a batch size of 10-20 will reduce it such that it will not be observable in any profiling. We've decided to have batch size 1 for 3.2 as default, but I'd opt for a smallish size in 3.3 and we'll see how it works from there.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StandardDynamicRecordAllocator.java
tinwelint: It's not a single usage, there's also `StandardDynamicRecordAllocator` (instance) and its subclass `RelativeIdRecordAllocator`, isn't there?

PR #9967, opened 2017-09-04
Increased timeout in test

tinwelint:


ragadeeshu (approved):
10 seconds might be a bit short, but if it works, I guess it works.

tinwelint:
Mmh, you may be right. Otoh it's 10x higher than before. Ah well, merging anyway :)

PR #9968, opened 2017-09-04
Fixes issues in CountsTrackerTest

tinwelint:


klaren (approved):
Nice catch with the race condition as well!

PR #9969, opened 2017-09-04
Properly awaits condition in test instead of sleeping

tinwelint:


community/neo4j/src/test/java/org/neo4j/tracers/PageCacheCountersIT.java
```diff
@@ -95,13 +94,12 @@ public void pageCacheCountersAreSumOfPageCursorCounters() throws Exception
         long initialFlushes = pageCacheTracer.flushes();
 
         startNodeCreators( nodeCreators, nodeCreatorFutures );
-        TimeUnit.MILLISECONDS.sleep( 500 );
+        while ( pageCacheTracer.pins() == 0 || pageCacheTracer.faults() == 0 || pageCacheTracer.unpins() == 0 )
```
burqen, line 97: I think we should still have a timeout on test it self to avoid infinite hang, something like 10 min or so.

burqen (changes requested):
Nice! Got one suggestion for improvement.

community/neo4j/src/test/java/org/neo4j/tracers/PageCacheCountersIT.java
tinwelint: Alright, I can add a 1 minute timeout perhaps

tinwelint:
Updated

burqen (approved):

PR #9970, opened 2017-09-04
Removes timeout from test because there's no reason to have it

tinwelint:
That test will just have to take as long as it takes

### DIFF
--- a/community/neo4j/src/test/java/org/neo4j/index/LegacyIndexTest.java
+++ b/community/neo4j/src/test/java/org/neo4j/index/LegacyIndexTest.java
@@ -38,12 +38,10 @@ import static org.neo4j.helpers.collection.MapUtil.stringMap;
 
 public class LegacyIndexTest
 {
-    private static final long TEST_TIMEOUT = 40_000;
-
     @Rule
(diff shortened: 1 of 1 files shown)


community/neo4j/src/test/java/org/neo4j/index/LegacyIndexTest.java
```diff
@@ -38,12 +38,10 @@
 
 public class LegacyIndexTest
 {
-    private static final long TEST_TIMEOUT = 40_000;
-
     @Rule
     public TestDirectory directory = TestDirectory.testDirectory();
 
-    @Test( timeout = TEST_TIMEOUT )
+    @Test
```
MishaDemianenko, line 44: test timeout was there to verify that performance of legacy index population is more or less decent for cases when we have a lot of fields in the index. With proposed change timeout is gone making this test totally useless. Not sure its a good thing

tinwelint:
Oh I see... hmm but the test is a bit fragile then. I remember this test now. Now with the parallel execution the test times can vary really really much, so this should perhaps be verified in another way, no?

tinwelint:
I'm closing this one... we should think about rewriting this test.

PR #9971, opened 2017-09-04
faster hashcode for ArrayValue

pontusmelke:
By implementing the hash code as a dot product we can achieve higher parallelizability which makes some hash code computations roughly two times faster.

### DIFF
--- a/community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
+++ b/community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
@@ -24,7 +24,7 @@ import java.util.Arrays;
 
 /**
  * Static methods for computing the hashCode of primitive numbers and arrays of primitive numbers.
- *
+ * <p>
  * Also compares Value typed number arrays.
  */
 @SuppressWarnings( "WeakerAccess" )
@@ -32,6 +32,24 @@ public final class NumberValues
 {
     private NumberValues()
     {
+    }
+
+    /*
+     * Using the fact that the hashcode ∑x_i * 31^(i-1) can be expressed as
+     * a dot product, [v_1, v_2, v_2, ...] • [1, 31, 31^2,...]. By expressing
+     * it in that way the compiler is smart enough to better parallelize the
+     * computation of the hash code.
+     */
+    private static final int MAX_LENGTH = 10000;
+    private static final int[] COEFFICIENTS = new int[MAX_LENGTH + 1];
+
+    static
+    {
+        COEFFICIENTS[0] = 1;
+        for ( int i = 1; i <= MAX_LENGTH; ++i )
+        {
+            COEFFICIENTS[i] = 31 * COEFFICIENTS[i - 1];
+        }
     }
 
     private static final long NON_DOUBLE_LONG = 0xFFE0_0000_0000_0000L; // doubles are exact integers up to 53 bits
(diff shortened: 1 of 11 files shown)


community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
```diff
@@ -49,53 +67,110 @@ public static int hash( double number )
...
+     * This is a slightly silly optimization but by turning the computation
+     * of the hashcode into a dot product we trick the jit compiler to use SIMD
+     * instructions and performance doubles.
+     */
     public static int hash( byte[] values )
     {
-        int result = 1;
-        for ( byte value : values )
+        final int max = values.length;
+        int result = COEFFICIENTS[max];
```
chrisvest: What if `values.length` is longer than `COEFFICIENTS.length`?

pontusmelke: Hmm...I just changed that. But you're of course right, then catastrophe entails...

community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
```diff
@@ -34,6 +34,26 @@ private NumberValues()
...
+     * it in that way the compiler is smart enough to better parallelize the
+     * computation of the hash code.
+     */
+    static final int MAX_LENGTH = 10000;
+    private static final int[] COEFFICIENTS = new int[MAX_LENGTH + 1];
+
+    static
+    {
+        //We are defining the coefficient vector backwards, [1, 31, 31^2,...]
+        //makes it easier and faster do find the starting position later
```
alexaverbuch: makes it easier and faster TO find the starting position later

community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
```diff
@@ -36,15 +36,17 @@ private NumberValues()
 
     /*
      * Using the fact that the hashcode ∑x_i * 31^(i-1) can be expressed as
-     * a dot product, [v_1, v_2, v_2, ...] • [1, 31, 31^2,...]. By expressing
+     * a dot product, [1, v_1, v_2, v_2, ..., v_n] • [31^n, 31^{n-1}, ..., 31, 1]. By expressing
```
alexaverbuch: `1, v_1, v_2, v_2` --> `1, v_1, v_2, v_3`?

community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
```diff
@@ -49,53 +69,74 @@ public static int hash( double number )
...
-        return (int)(bits ^ (bits >>> 32));
+        return (int) (bits ^ (bits >>> 32));
     }
 
+    /*
+     * This is a slightly silly optimization but by turning the computation
+     * of the hashcode into a dot product we trick the jit compiler to use SIMD
+     * instructions and performance doubles.
+     */
     public static int hash( byte[] values )
```
lutovich, line 88: Can we use `Arrays.hashCode(byte[])` and other overloads instead or writing code ourselves?
Presumably they are very optimized and can become intrinsics at some point.

alexaverbuch:
nice!

couple questions:
  1. did you run JMH benchmarks against this branch? it contains NodeHashJoin, Distinct aggregation, and maybe more benchmarks that make use of hashcode.
  2. I guess you moved `TextValues.hash( char[] value )` & `TextValues.hash( String[] value )` to `NumberValues` so `COEFFICIENTS` could be reused? 
I think that's sensible, but it's not intuitive that "NumberValues" contains text-specific code. Maybe move these methods to `ValuesHashcodeUtil` or some such instead?

community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
pontusmelke: Yes the hashcode is for `a = [v_1, v_2,..., v_n]` is

```
int result = 1;
for (int element : a) {
    result = 31 * result + element;
}
```
which is equivalent to

```
1 * 31^n + v_1 * 31^{n-1} + ... + v_{n-1} * 31 + v_n * 1 = [1, v_1,...,v_n] • [31^n, ..., 31^1, 31^0]
```

community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
pontusmelke: They are not intrinsics at the moment and more than twice as slow both in Java 8 and Java 9. When Java10 is on the horizon we should revisit.

pontusmelke:
@alexaverbuch 

1. Yes this has very limited effect since we a rarely use primitive arrays
2. Yes I moved them but benchmarking showed no improvements for `Object[]` so I should move it back again.

community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
lutovich: They do not need a major release to change code or add intrinsics ;)
See https://bugs.openjdk.java.net/browse/JDK-8186203 and https://bugs.openjdk.java.net/browse/JDK-8059113. It just feels strange to me that computing hash code for arrays is our bottleneck.

community/values/src/test/java/org/neo4j/values/storable/NumberValuesTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Test
+    public void shouldHandleBigIntArrays()
+    {
+        int[] big = new int[NumberValues.MAX_LENGTH + 1];
+        int[] slightlySmaller = new int[MAX_LENGTH];
+        Arrays.fill( big, 1337 );
+        Arrays.fill( slightlySmaller, 1337 );
+
+        assertThat( hash( big ), equalTo( hash( slightlySmaller ) ) );
```
systay: I can see that this is an outcome of the way we cap at MAX_LENGTH, but I'm not sure why we need to specify this behaviour with unit tests.

pontusmelke: Yeah I guess you're right. Wanted to make sure I got it right, but maybe that assertion has served its role and can be removed

systay:
I updated the unit tests - we need two arrays that contain the same values but as longs/ints to return the same hash-code, right?

pontusmelke:
@systay 👍

community/values/src/main/java/org/neo4j/values/storable/NumberValues.java
pontusmelke: It is probably not a major bottleneck but since our interpreted runtime is basically one big hashmap these things do matter :)

We have jmh-benchmarks set up for hashcodes so we should pick up if `Arrays.hashCode` are made faster all of a sudden. But even so we cannot use them naively everywhere since we have different equality semantics than that of java. For example, given 

```
ByteArray bytes = Values.byteArray( new byte[]{-42, 42} );
LongArray longs = Values.longArray( new long[]{-42L, 42L} );
```

then in our semantics we expect `bytes.equals(longs) == true` and `bytes.hashCode() == longs.hashCode()`. This is not the case for java equality or java hashcodes.

lutovich: Fair enough, I haven't thought about array equality.

PR #9974, opened 2017-09-04
Improve error message for duration settings

klaren:
The error message for a duration now includes valid units.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/helpers/TimeUtil.java
+++ b/community/kernel/src/main/java/org/neo4j/helpers/TimeUtil.java
@@ -24,7 +24,9 @@ import java.util.function.Function;
 
 public final class TimeUtil
 {
-    public static final TimeUnit DEFAULT_TIME_UNIT = TimeUnit.SECONDS;
+    private static final TimeUnit DEFAULT_TIME_UNIT = TimeUnit.SECONDS;
+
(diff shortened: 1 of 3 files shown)


burqen (changes requested):
Nice changes! I don't think they work all the way though, since `Settings.apply` only pass along message from the "closest" cause.
```
catch ( IllegalArgumentException e )
{
    throw new InvalidSettingException( name(), value, e.getMessage() );
}
```
Verified with "test":
```
@Test
public void badDuration() throws Exception
{
    Setting<Duration> setting = buildSetting( "foo.bar", DURATION ).build();
    try
    {
        setting.apply( map( stringMap( "foo.bar", "hej" ) ) );
    }
    catch ( InvalidSettingException e )
    {
        e.printStackTrace();
    }
}
```
That prints stack trace:
```
org.neo4j.graphdb.config.InvalidSettingException: Bad value 'hej' for setting 'foo.bar': Error parsing duration setting

	at org.neo4j.kernel.configuration.Settings$DefaultSetting.apply(Settings.java:1299)
	at org.neo4j.kernel.configuration.Settings$DefaultSetting.apply(Settings.java:1202)
	at org.neo4j.kernel.configuration.SettingsTest.badDuration(SettingsTest.java:289)
```

burqen (approved):

burqen:
@klaren Looks super!

jjaderberg (approved):
:+1:

PR #9977, opened 2017-09-04
Update message on store format

mariascharin:


mariascharin:
@MishaDemianenko Please review the updated description of dbms.record_format. In particular, please check that the call
public static final Setting<String> record_format = setting( "dbms.record_format", Settings.STRING, "standard" );
is correct. Thanks!

mariascharin:
@MishaDemianenko OK?

MishaDemianenko (approved):

PR #9978, opened 2017-09-04
Fusion index use pure Lucene for composite indexes

burqen:
By checking how many property keys given IndexDescriptor has.


tinwelint (approved):
Nice

burqen:
Waiting for benchmark run to verify fix.

burqen:
Benchmarks suggest this makes close to no difference. Closing.

PR #9980, opened 2017-09-04
Fusion index enabled by default

tinwelint:
Fusion index is a combination of Lucene+native indexing. The native index
can currently only handle numbers, more efficiently than Lucene can.
Other values still go to Lucene.

Existing indexes will not automatically be migrated, but instead new
indexes will be created with the fusion index provider.


burqen:
Running HA-robustness on this branch. Status expected in three hours.

PR #9984, opened 2017-09-05
Make ImpermanentDatabaseRule use random directory

klaren:


tinwelint:
I do believe this change is fine and good, just wondering... was there a real problem with this? My understanding is that each test suite creates its own ephemeral file system and so should never collide with each other, even if the directory was the same.

klaren:
@tinwelint You're completely right, something went horribly wrong in the test that revealed this issue. Haven't encountered it again since so I'm gonna close this until it can be produced repeatable.

PR #9985, opened 2017-09-05
Re-enables previous accidental disable of parallel index population

tinwelint:
Now in BatchingMultipleIndexPopulator the index scan takes advantage of
the parallel execution available by its executor. In a previous commit this
was accidentally disabled.


burqen:
Waiting for benchmark run to verify fix.

PR #9986, opened 2017-09-05
Better handling of parameters from Bolt

pontusmelke:
When reading data from the wire we have full type information so instead of sending parameters as a `Map<String, Object>` which Cypher then turns into a `MapValue` we send a `MapValue` directly.

NOTE: This refactoring will fail the docs build so disregard that build failure for now. In order for docs build we need to also merge https://github.com/neo4j/neo4j-documentation/pull/122


enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerIT.java
```diff
@@ -227,7 +228,7 @@ public void shouldLogParametersWhenNestedMap() throws Exception
         List<String> logLines = readAllLines( logFilename );
         assertEquals( 1, logLines.size() );
         assertThat( logLines.get( 0 ), endsWith( String.format(
-                " ms: %s - %s - {props: {name: 'Roland', position: 'Gunslinger', followers: [Jake, Eddie, Susannah]}}"
+                " ms: %s - %s - {props: {name: 'Roland', position: 'Gunslinger', followers: ['Jake', 'Eddie', 'Susannah']}}"
```
henriknyman, line 231: This looks more correct now. I guess it could potentially affect somebody's custom query log parsing script, but that's life.

henriknyman (approved):
Some seemingly needless code reformatting, but looks good in general.

PR #9988, opened 2017-09-05
Order writes to native schema number index

tinwelint:
Because it's more sympathetic to the GBPTree writer.


tinwelint:
Doesn't seem to make any difference. Holding off for the time being.

PR #9991, opened 2017-09-05
Fix copyright regex for windows builds.

ragadeeshu:
For some reason windows thinks that there is an extra character? This gobbles that up.


tinwelint (approved):
Cool, I don't understand exactly why, but if it works it works. And it's optional too, goody.

PR #9993, opened 2017-09-05
Fix side effects validation in TCK implementation

Mats-SX:
Side effects were not checked at all in the TCK implementation. In https://github.com/opencypher/openCypher/pull/236, side effects calculation was changed for the TCK, with a Cypher-based approach for measuring the side effect metrics. This PR introduces support for validating side effects in accordance to the standard definition.

Unfortunately, the TCK has not been released in an updated version that adheres all scenarios to these changes yet, and as such many scenarios have been blacklisted in this PR. These will be un-blacklisted again once the TCK dependency can be updated.


Mats-SX:
@fickludd It would be great if someone from the Cypher team could be assigned this PR.

enterprise/cypher/compatibility-spec-suite/src/test/resources/blacklists/compatibility-23.txt
```diff
@@ -185,3 +185,40 @@ In-query procedure call should fail if one of the argument expressions uses an a
 Comparing strings and integers using > in an AND'd predicate
 Comparing strings and integers using > in a OR'd predicate
 Comparing across types yields null, except numbers
+
+// rule planner/eagerness bug
+Combine MATCH, WITH and CREATE
+
+// TCK side effects are incorrectly specified -- need update
```
craigtaverner, line 192: Does this mean side effects are hard to test for in 2.3, but otherwise these tests would pass? Or does 2.3 have different side effects, and that is perhaps a problem with 2.3 that will not be fixed? In either case is there value in not testing side effects in 2.3?

enterprise/cypher/compatibility-spec-suite/src/test/resources/blacklists/compatibility-31.txt
```diff
@@ -10,3 +10,37 @@ In-query call to procedure with explicit arguments that drops all result fields
 // procedure call bugs
 Standalone call to procedure with argument of type INTEGER accepts value of type FLOAT
 In-query call to procedure with argument of type INTEGER accepts value of type FLOAT
+
+// TCK side effects are incorrectly specified -- need update
```
craigtaverner, line 14: Not testing side effects in 2.3 is not as serious as not testing side effects in 3.1.

enterprise/cypher/spec-suite-tools/src/main/scala/cypher/feature/parser/SideEffects.scala
```diff
@@ -0,0 +1,155 @@
...
+        v.copy(relsDeleted = values.get(index).toInt)
+      case (v, (ADDED_LABELS, index)) =>
+        v.copy(labelsCreated = values.get(index).toInt)
+      case (v, (DELETED_LABELS, index)) =>
+        v.copy(labelsDeleted = values.get(index).toInt)
+      case (v, (ADDED_PROPERTIES, index)) =>
+        v.copy(propsCreated = values.get(index).toInt)
+      case (v, (DELETED_PROPERTIES, index)) =>
+        v.copy(propsDeleted = values.get(index).toInt)
+      case (sideEffect, _) => throw InvalidFeatureFormatException(
```
craigtaverner: Did you mean to report on the accumulator here? Seems to make more sense to `case (_, sideEffect)` or `case (_, (sideEffect, _))`

enterprise/cypher/spec-suite-tools/src/main/scala/cypher/feature/parser/SideEffects.scala
```diff
@@ -0,0 +1,155 @@
...
+  private val relsQuery =
+    s"""$prefix MATCH ()-[r]->() RETURN id(r) AS rel"""
+
+  private val labelsQuery =
+    s"""$prefix MATCH (n)
+       |UNWIND labels(n) AS label
+       |RETURN DISTINCT label""".stripMargin
+
+  private val propsQuery =
+    s"""$prefix MATCH (n)
```
craigtaverner, line 125: Would it not be easier to `CALL db.propertyKeys()` ?

craigtaverner: Although I suspect that will not show deleted properties, right?

enterprise/cypher/compatibility-spec-suite/src/test/resources/blacklists/default.txt
```diff
@@ -9,3 +9,37 @@ In-query call to procedure with argument of type INTEGER accepts value of type F
 // TODO tck enforces specific order of `keys`, remove from blacklist when tck has been updated.
 Copying properties from literal map with ON CREATE
 Copying properties from literal map with ON MATCH
+
+// TCK side effects are incorrectly specified -- need update
```
craigtaverner, line 13: OK. I'm starting to understand. All blacklists have these same lines, so all these tests need their side effects corrected. I guess that is followup work?

craigtaverner (changes requested):

enterprise/cypher/compatibility-spec-suite/src/test/resources/blacklists/default.txt
craigtaverner: I also notice no blacklist for 3.2, so it seems 3.3 TCK is missing support for testing against 3.2? Perhaps a separate PR for that would be good.

enterprise/cypher/compatibility-spec-suite/src/test/resources/blacklists/compatibility-23.txt
Mats-SX: No, these blacklistings (as you touch on in a later comment) are there because the scenarios themselves are incorrect and requires updating. This is done in this PR: https://github.com/opencypher/openCypher/pull/250

Now, the zero-state here is not very different; if you inspect the code currently in `3.3`, side effects are not validated at all (as I mention in the PR description). After introducing correct validation, many scenarios were found to be incorrectly specified (specifies the wrong expectations) and just very few were found to actually pose a problem (bug) in one of the configurations (I believe only in 2.3).

As I see it, there are two viable strategies available here: merge this now and update the TCK dependency as soon as it comes out, effectively un-blacklisting the vast majority of scenarios that have been blacklisted in this PR; or wait with merging this PR until that has happened and make updating the dependency another commit on this PR.

Merging now has the benefit of this PR not falling behind `3.3` and needing rebasing/maintenance. Merging later has the benefit of not blacklisting scenarios, potentially hiding future bugs until the scenarios can be un-blacklisted again. (The scenario is already blacklisted, so introducing a change that makes it "fail more" won't be seen.)

enterprise/cypher/spec-suite-tools/src/main/scala/cypher/feature/parser/SideEffects.scala
Mats-SX: These queries are part of the standard TCK definitions. See https://github.com/opencypher/openCypher/tree/master/tck#observability-of-side-effects

The idea behind that is to formulate the TCK in terms of standard Cypher only. Currently, there are no procedures that are part of the standard.

enterprise/cypher/compatibility-spec-suite/src/test/resources/blacklists/default.txt
Mats-SX: Good point! That configuration seems to be missing, indeed. It is out of scope for this PR though.

craigtaverner (approved):

PR #9995, opened 2017-09-06
Changes schema index directory structure for fusion index

tinwelint:
The lucene schema index provider puts its indexes in:

  `<db>/schema/index/lucene/<indexId>/...`

this commit doesn't change this, but the fusion index provider,
i.e. the combination of lucene+native gets a new structure driven by the need
to one day be able to do efficient background migration of schema indexes
and for convenience in the wild. The new structure looks like this:

  `<db>/schema/index/<providerKey>/<providerVersion>/<indexId>/...`

Including the sub-provider directory and specifically for this fusion index (for indexId 5):

  `<db>/schema/index/lucene_native/0.1/5/lucene/...`
  `<db>/schema/index/lucene_native/0.1/5/native/...`

This has two benefits:

- all index data related to one index lives in one directory
- indexes on disk are also versions, just like providers are

Since the fusion index also uses a lucene schema index provider internally,
this structure has been made pluggable and given to a SchemaIndexProvider
at construction time. See IndexDirectoryStructure. This class has all knowledge
about the base schema root directory and various directory structure implementations
collected in one place.

NOTE This change does not require any migration, except for databases recently
created, where the fusion index is enabled by default. Although no releases
has had the fusion index in them so that is not a problem.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDirectoryStructure.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDirectoryStructure.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api.index;
+
+import java.io.File;
+
+import org.neo4j.kernel.api.index.SchemaIndexProvider.Descriptor;
+
+import static org.neo4j.io.fs.FileUtils.path;
+
+/**
+ * Dictates how directory structure looks for a {@link SchemaIndexProvider} and its indexes. Generally there's a
+ * {@link #rootDirectory() root directory} which contains all index directories in some shape and form.
+ * For getting a directory (which must be a sub-directory to the root directory) for a particular index there's the
+ * {@link #directoryForIndex(long)} method.
+ *
+ * These instances are created from a {@link Factory} which typically gets passed into a {@link SchemaIndexProvider} constructor,
+ * which then a {@link IndexDirectoryStructure} given its {@link Descriptor}.
+ */
+public abstract class IndexDirectoryStructure
+{
+    /**
+     * Creates an {@link IndexDirectoryStructure} for a {@link Descriptor} for a {@link SchemaIndexProvider}.
+     */
+    public interface Factory
+    {
+        IndexDirectoryStructure forProvider( SchemaIndexProvider.Descriptor descriptor );
+    }
+
+    private static class SubDirectoryByIndexId extends IndexDirectoryStructure
+    {
+        private final File providerRootFolder;
+
+        private SubDirectoryByIndexId( File providerRootFolder )
+        {
+            this.providerRootFolder = providerRootFolder;
+        }
+
+        @Override
+        public File rootDirectory()
+        {
+            return providerRootFolder;
+        }
+
+        @Override
+        public File directoryForIndex( long indexId )
+        {
+            return path( providerRootFolder, String.valueOf( indexId ) );
+        }
+    }
+
+    /**
+     * Returns the base schema index directory, i.e.
+     *
+     * <pre>
+     * &lt;db&gt;/schema/index/
+     * </pre>
+     *
+     * @param databaseStoreDir database store directory, i.e. {@code db} in the example above, where e.g. {@code nodestore} lives.
+     * @return the base directory of schema indexing.
+     */
+    static final File baseSchemaIndexFolder( File databaseStoreDir )
+    {
+        return path( databaseStoreDir, "schema", "index" );
+    }
+
+    /**
+     * @param databaseStoreDir store directory of database, i.e. {@code db} in the example above.
+     * @return {@link Factory} for creating {@link IndexDirectoryStructure} returning directories looking something like:
+     *
+     * <pre>
+     * &lt;db&gt;/schema/index/&lt;providerKey&gt;/&lt;indexId&gt;/
+     * </pre>
+     */
+    public static Factory directoriesByProviderKey( File databaseStoreDir )
+    {
+        return descriptor -> new SubDirectoryByIndexId(
+                path( baseSchemaIndexFolder( databaseStoreDir ), fileNameFriendly( descriptor.getKey() ) ) );
+    }
+
+    /**
+    * @param databaseStoreDir store directory of database, i.e. {@code db} in the example above.
+    * @return {@link Factory} for creating {@link IndexDirectoryStructure} returning directories looking something like:
+    *
+    * <pre>
+    * &lt;db&gt;/schema/index/&lt;providerKey&gt;-&lt;providerVersion&gt;/&lt;indexId&gt;/
+    * </pre>
+    */
+    public static Factory directoriesByProvider( File databaseStoreDir )
+    {
+        return descriptor -> new SubDirectoryByIndexId(
+                path( baseSchemaIndexFolder( databaseStoreDir ), fileNameFriendly( descriptor ) ) );
+    }
+
+    /**
+     * @param directoryStructure existing {@link IndexDirectoryStructure}.
+     * @return a {@link Factory} returning an already existing {@link IndexDirectoryStructure}.
+     */
+    public static Factory given( IndexDirectoryStructure directoryStructure )
+    {
+        return descriptor -> directoryStructure;
+    }
+
+    /**
+     * Useful when combining multiple {@link SchemaIndexProvider} into one.
+     *
+     * @param parentStructure {@link IndexDirectoryStructure} of the parent.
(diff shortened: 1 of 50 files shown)


community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDirectoryStructure.java
```diff
@@ -0,0 +1,208 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.api.index;
```
burqen, line 20: Am I mistaken or is this package not public API? If so, this class should be moved somewhere else.

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDirectoryStructure.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+    };
+
+    /**
+     * Useful for some in-memory index providers or similar.
+     */
+    public static final Factory NONE = descriptor -> NO_DIRECTORY_STRUCTURE;
+
+    /**
+     * Returns root directory. Must be above all sub-directories returned from {@link #directoryForIndex(long)}.
```
burqen: "Must be above" could be changed to "Must be parent to"

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDirectoryStructure.java
```diff
@@ -0,0 +1,208 @@
...
+import static org.neo4j.io.fs.FileUtils.path;
+
+/**
+ * Dictates how directory structure looks for a {@link SchemaIndexProvider} and its indexes. Generally there's a
+ * {@link #rootDirectory() root directory} which contains all index directories in some shape and form.
+ * For getting a directory (which must be a sub-directory to the root directory) for a particular index there's the
+ * {@link #directoryForIndex(long)} method.
+ *
+ * These instances are created from a {@link Factory} which typically gets passed into a {@link SchemaIndexProvider} constructor,
+ * which then a {@link IndexDirectoryStructure} given its {@link Descriptor}.
```
burqen: Bring this sentence back to a consistent state, please.

Nice javadoc otherwise 👍

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionSchemaIndexProvider.java
```diff
@@ -49,10 +50,11 @@
     private final SchemaIndexProvider luceneProvider;
     private final Selector selector;
 
-    public FusionSchemaIndexProvider( SchemaIndexProvider nativeProvider, SchemaIndexProvider luceneProvider, Selector selector,
+    public FusionSchemaIndexProvider( IndexDirectoryStructure.Factory directoryStructure,
```
burqen: May I suggest putting `directoryStructure` as last parameter to align better with `super`

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexProvider.java
```diff
@@ -48,18 +49,25 @@
 {
     public static final String KEY = "native";
     public static final Descriptor NATIVE_PROVIDER_DESCRIPTOR = new Descriptor( KEY, "1.0" );
+    private static final String INDEX_FILE_NAME = "index";
```
burqen: Would be nice to also have index id as part of file name for easier readability by humans.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexProvider.java
```diff
@@ -48,18 +49,25 @@
...
+
     private final PageCache pageCache;
-    private final File nativeSchemaIndexBaseDir;
     private final Log log;
     private final RecoveryCleanupWorkCollector recoveryCleanupWorkCollector;
     private final boolean readOnly;
 
     public NativeSchemaNumberIndexProvider( PageCache pageCache, File storeDir, LogProvider logging,
```
burqen: Can be package-private

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexProvider.java
```diff
@@ -48,18 +49,25 @@
...
             RecoveryCleanupWorkCollector recoveryCleanupWorkCollector, boolean readOnly )
     {
-        super( NATIVE_PROVIDER_DESCRIPTOR, 0 );
+        this( pageCache, IndexDirectoryStructure.directoriesByProvider( storeDir ), logging,
+                recoveryCleanupWorkCollector, readOnly );
+    }
+
+    public NativeSchemaNumberIndexProvider( PageCache pageCache, IndexDirectoryStructure.Factory directoryStructure, LogProvider logging,
+            RecoveryCleanupWorkCollector recoveryCleanupWorkCollector, boolean readOnly )
```
burqen: Quite a few double constructors and factory methods are added in this PR to accommodate for `IndexDirectoryStructure`. I think it would be cleaner to just inline the `directoriesByProvider( storeDir )` method instead.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/NativeLuceneFusionSchemaIndexProviderFactory.java
```diff
@@ -74,14 +78,22 @@ public static FusionSchemaIndexProvider newInstance( PageCache pageCache, File s
             LogProvider logProvider, Config config, OperationalMode operationalMode,
             RecoveryCleanupWorkCollector recoveryCleanupWorkCollector )
     {
+        IndexDirectoryStructure.Factory childDirectoryStructure = subProviderDirectoryStructure( storeDir );
         boolean readOnly = isReadOnly( config, operationalMode );
         NativeSchemaNumberIndexProvider nativeProvider =
-                new NativeSchemaNumberIndexProvider( pageCache, storeDir, logProvider, recoveryCleanupWorkCollector, readOnly );
-        LuceneSchemaIndexProvider luceneProvider = LuceneSchemaIndexProviderFactory.create( fs, storeDir, logProvider, config,
+                new NativeSchemaNumberIndexProvider( pageCache, childDirectoryStructure, logProvider, recoveryCleanupWorkCollector, readOnly );
+        LuceneSchemaIndexProvider luceneProvider = LuceneSchemaIndexProviderFactory.create( fs, childDirectoryStructure, logProvider, config,
```
burqen: Long lines

tinwelint:
@burqen you had some comments, didn't you? Where are they?

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDirectoryStructure.java
```diff
@@ -0,0 +1,208 @@
...
+     * Returns the base schema index directory, i.e.
+     *
+     * <pre>
+     * &lt;db&gt;/schema/index/
+     * </pre>
+     *
+     * @param databaseStoreDir database store directory, i.e. {@code db} in the example above, where e.g. {@code nodestore} lives.
+     * @return the base directory of schema indexing.
+     */
+    static final File baseSchemaIndexFolder( File databaseStoreDir )
```
burqen: Does not need to be final.

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDirectoryStructure.java
```diff
@@ -0,0 +1,208 @@
...
+     * @param parentStructure {@link IndexDirectoryStructure} of the parent.
+     * @return {@link Factory} creating {@link IndexDirectoryStructure} looking something like:
+     *
+     * <pre>
+     * &lt;db&gt;/schema/index/.../&lt;indexId&gt;/&lt;childProviderKey&gt;/
+     * </pre>
+     */
+    public static Factory directoriesBySubProviderKey( IndexDirectoryStructure parentStructure )
+    {
+        return new Factory()
```
burqen: Can be lambda

burqen (changes requested):
Looks sweet! Gonna be awesome with a more separated directory structure.
Got a few comments, most of which we have already talked about.

I would prefer if we could combine `provider_key` and `provider_version` into one directory name instead of separating them into two directories.
In that way we could treat sub-provider directories the same as "first" provider directory.

Instead of:
`<db>/schema/index/lucene_native/0.1/5/lucene/...`
`<db>/schema/index/lucene_native/0.1/5/native/...`
It would look like:
`<db>/schema/index/lucene_native_0.1/5/lucene_?/...`
`<db>/schema/index/lucene_native_0.1/5/native_0.3/...`

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDirectoryStructure.java
tinwelint: The kernel API isn't public, so that's fine

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexDirectoryStructure.java
tinwelint: Missing a `creates` right?

tinwelint:
Updated

tinwelint:
Hmm recent rebase exposed an issue where the index directories under the fusion provider aren't deleted when dropping an index. This should be looked into first.

burqen (approved):

burqen:
@tinwelint Changes look good 👍  but build state looks a bit more 👎

burqen (changes requested):
Second thought, let's not approve this until build failure is solved.

tinwelint:
@burqen updated and should fix the last test failure.

burqen (approved):

PR #9997, opened 2017-09-06
3.3 slotted map project

alexaverbuch:


systay (approved):

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/SlottedRewriter.scala
```diff
@@ -121,29 +121,28 @@ class SlottededRewriter(tokenContext: TokenContext) {
...
+            if (nullable)
+              NullCheck(offset, propExpression)
+            else
+              propExpression
 
-        if (slot.nullable)
-          NullCheck(slot.offset, propExpression)
-        else
-          propExpression
+          case RefSlot(offset, _, _, _) => prop.copy(map = ReferenceFromSlot(offset))(prop.position)
```
sherfert, line 144: I didn't know case classes have a copy method until now. Nice.

PR #10001, opened 2017-09-06
Removes resolves of SchemaIndexProvider directly

tinwelint:
Instead forces use of SchemaIndexProviderMap where care has to be taken for
multiple providers, which is good to consider. In many cases, especially in tests
it's still just about the default index, which can easily be retrieved from that map.

This exposed (now fixed) an issue in StoreSizeBean where it didn't include index sizes
from all providers when calculating schema index size.


klaren (approved):
Nice catch of the issue in `StoreSizeBean`!

PR #10003, opened 2017-09-06
Remove rawtype usage in KernelExtension

klaren:


community/kernel/src/main/java/org/neo4j/kernel/impl/util/DependenciesProxy.java
```diff
@@ -31,31 +31,26 @@
...
 
     /**
      * Create a dynamic proxy that implements the given interface and backs invocation with lookups into the given
      * dependency resolver.
-     *
-     * @param dependencyResolver
-     * @param dependenciesInterface
-     * @param <T>
-     * @return
      */
```
ragadeeshu: Maybe we could fill out the missing parts of the javadoc instead?

ragadeeshu (approved):
Looks good, one optional comment

klaren:
@ragadeeshu addressed!

PR #10004, opened 2017-09-06
Upgrading Shiro and commons-digester version

eebus:


eebus:
Replace by https://github.com/neo4j/neo4j/pull/10019

PR #10008, opened 2017-09-07
Remove module neo4j-monitor-logging

klaren:
The module **neo4j-monitor-logging** appears to be unused. This PR removes the module and a few methods that this module has the only consumer of.


tinwelint (approved):
Nice to get rid off, I never understood its purpose anyway

PR #10010, opened 2017-09-07
Propagate error description for upgrade not allowed exceptions

klaren:


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/LogVersionUpgradeChecker.java
```diff
@@ -47,11 +47,8 @@ public static void check( LogTailScanner tailScanner, Config config ) throws Upg
             LogEntryVersion latestLogEntryVersion = tailScanner.getTailInformation().latestLogEntryVersion;
             if ( latestLogEntryVersion != null && LogEntryVersion.moreRecentVersionExists( latestLogEntryVersion ) )
             {
-                String message = String.format(
-                        "The version your upgrading to is using a new transaction log format. This is a non-reversible " +
-                                "upgrade and you wont be able to downgrade after starting. To allow upgrade, please set " +
-                                "configuration parameter \"%s=true\"",
-                        GraphDatabaseSettings.allow_upgrade.name() );
+                String message = "The version your upgrading to is using a new transaction log format. This is a " +
```
tinwelint: `your` --> `you're`

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/LogVersionUpgradeChecker.java
```diff
@@ -47,11 +47,8 @@ public static void check( LogTailScanner tailScanner, Config config ) throws Upg
...
+                        "non-reversible upgrade and you wont be able to downgrade after starting";
```
tinwelint: Also would be helpful to see the version (both existing as well as current), right?

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/UpgradeNotAllowedByConfigurationException.java
```diff
@@ -25,14 +25,19 @@
...
-                "Failed to start Neo4j with an older data store version. "
-                        + "To enable automatic upgrade, please set configuration parameter \"%s=true\"",
-                GraphDatabaseSettings.allow_upgrade.name() ) );
+        super( baseMessage() );
+    }
+
+    private static String baseMessage()
+    {
+        return String.format(
+                "Neo4j cannot be started, because the database files require upgrading and upgrades are " +
```
tinwelint: `upgrades are`, dunno about the plural there. Perhaps `upgrading is`... just taking extra care since it's public surface. WDYT?

tinwelint (changes requested):
High changed line count to comments ratio, but wouldn't it be better if there was one place where the note about the config was mentioned and specific places where the upgrade is noticed mentions specific details, but not that... otherwise all places will have to mention that config in a consistent way.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/LogVersionUpgradeChecker.java
klaren: If you believe that numbers like -9 will make the users less confused we could add that.

tinwelint (approved):
A completely misread the changes the previous review round. This looks dandy!

PR #10011, opened 2017-09-07
Improve log assertion in BoltFailuresIT

lutovich:
Test `boltServerLogsRealErrorWhenDriverIsClosedWithRunningTransactions` was flaky because it did not properly wait for error message. It asserts that particular error message appears in logs after a driver transaction failure. This failure can be observed earlier then error is logged.

This PR makes test wait up to 1 minute for the message to appear.


tinwelint (approved):
I think that makes sense. Great work

PR #10013, opened 2017-09-07
refactor comparison support

Lojjs:


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -95,209 +95,153 @@ trait CypherComparisonSupport extends CypherTestSupport {
...
 
     lastResult
   }
 
-  private def assertNoOverlap(expectedSuccessFrom: TestConfiguration, ignoreScenarios: TestConfiguration) = {
-    val hasOverlap = expectedSuccessFrom.scenarios.exists(ignoreScenarios.scenarios.contains(_))
-    if (hasOverlap)
-      fail("Set of configurations that were expected to succeed and the set of configs that should be ignored are overlapping")
+  private def extractFirstScenario(config: TestConfiguration, compareResults: TestConfiguration, comparePlans: TestConfiguration): TestScenario = {
+    val scenariosToChooseFrom = (compareResults.scenarios.isEmpty, comparePlans.scenarios.isEmpty) match {
```
sherfert: Since we are assuming `compareResults` and `comparePlans` are a subset of `config`,
can we not always intersect all three (and fail if that intersection is empty)?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -95,209 +95,153 @@ trait CypherComparisonSupport extends CypherTestSupport {
...
     }
 
     lastResult
   }
 
-  private def assertNoOverlap(expectedSuccessFrom: TestConfiguration, ignoreScenarios: TestConfiguration) = {
-    val hasOverlap = expectedSuccessFrom.scenarios.exists(ignoreScenarios.scenarios.contains(_))
-    if (hasOverlap)
-      fail("Set of configurations that were expected to succeed and the set of configs that should be ignored are overlapping")
+  private def extractFirstScenario(config: TestConfiguration, compareResults: TestConfiguration, comparePlans: TestConfiguration): TestScenario = {
```
sherfert: Can we change the parameter name `config` to `expectedSuccessFrom` or similar?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -95,209 +95,153 @@ trait CypherComparisonSupport extends CypherTestSupport {
...
-          None
-        }
-
+        executeScenario(thisScenario, query, expectedSuccessFrom.scenarios.contains(thisScenario), executeBefore, params)
     }
 
     firstScenario.prepare()
     executeBefore()
-    val lastResult = innerExecute(s"CYPHER ${firstScenario.preparserOptions} $query", params.toMap)
+    val lastResult = innerExecute(s"CYPHER ${firstScenario.preparserOptions} $query", params)
```
sherfert: `firstScenario` and `lastResult` seem inconsistent. They're both the scenario/result you're comparing against, aren't they?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -95,209 +95,153 @@ trait CypherComparisonSupport extends CypherTestSupport {
...
+                            ignoreResults: TestConfiguration = Configs.Empty,
+                            ignorePlans: TestConfiguration = Configs.AllRulePlanners,
+                            executeBefore: () => Unit = () => {},
+                            params: Map[String, Any] = Map.empty): InternalExecutionResult = {
+    val compareResults = expectedSuccessFrom -  ignoreResults
+    val comparePlans = expectedSuccessFrom - ignorePlans
+    val firstScenario = extractFirstScenario(expectedSuccessFrom, compareResults, comparePlans)
+
+    // let this throw to enable catching exceptions outside
+    firstScenario.prepare()
```
sherfert: I commented this block out and ran all acceptance and all compatibility tests. No failures. Can you point me to the test(s) that require this to be here?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/AggregationAcceptanceTest.scala
```diff
@@ -40,8 +42,9 @@ class AggregationAcceptanceTest extends ExecutionEngineFunSuite with CypherCompa
                    |RETURN id(selectedFriendship) AS friendshipId, selectedFriendship.propFive AS propertyValue""".stripMargin
     val params = "param" -> 3
```
sherfert: Now that `executeWith` accepts a map of params, we should probably define this right here as a map already.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -457,6 +405,8 @@ object CypherComparisonSupport {
 
   object Runtimes {
 
+    val all = Runtimes(CompiledBytecode, CompiledSource, Slotted, Interpreted, ProcedureOrSchema, Default)
```
sherfert: In `Planner.all` and `Versions.all` Alex and I explicitly excluded `Default`, since one of them would be the default anyway. Is that required to be included here?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NodeIndexSeekByRangeAcceptanceTest.scala
```diff
@@ -928,19 +926,14 @@ class NodeIndexSeekByRangeAcceptanceTest extends ExecutionEngineFunSuite with Cy
 
     val query = "MATCH (n:Label) WHERE n.prop >= '1' AND n.prop > 10 RETURN n.prop AS prop"
 
-    an[IllegalArgumentException] should be thrownBy {
```
sherfert: Is that `IllegalArgumentException` not thrown anymore or do we not assert it is thrown any more?

sherfert (changes requested):
This is a nice step forward, I like it! The comments below are just minor things that can enhance readability of the code or questions.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NodeIndexSeekByRangeAcceptanceTest.scala
OliviaYtterbrink: this was wrongly removed

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
OliviaYtterbrink: `test("should refuse to execute index seeks using inequalities over different types")` in `NodeIndexSeekByRangeAcceptanceTest` is reliant on this being here (when it has been reverted to actually check for the exception again)

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
OliviaYtterbrink: There is no `Versions.all`, and `Default` in runtimes is actually different from specifying one

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
OliviaYtterbrink: 👍

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
OliviaYtterbrink: Sometimes the plans aren't comparable but the results are, intersecting these we would then end up with an empty set when we just want the intersection between `compareResults` and `config`

sherfert (approved):

PR #10016, opened 2017-09-07
Improved GBPTree file handling

burqen:
Introduce interface GBPTreeFileUtil that gathers file operations
used inside of components built on top of GBPTree, label index
and schema index.

Implementation decide how those file operations should be carried out.
Specifically if it should be done using normal file system or file system
provided by page cache.


tinwelint (approved):
Looks nice and clean! Don't get the abstract test class, it's as if you wanted to do an abstract with two subclasses testing each util impl, but mid-air changed direction and went for parameterized.

It's fine, though. I'm OK with it.

burqen:
@tinwelint Blockdevice also inherit this test. Will write a comment about that and merge.

PR #10018, opened 2017-09-07
moving to threadsafe api for date formatting. fixing feedback

praveenag:


community/logging/src/main/java/org/neo4j/logging/FormattedLog.java
```diff
@@ -134,6 +131,18 @@ public Builder withLogLevel( Level level )
         }
 
         /**
+         * Use the specified function
+         *
+         * @param dateTimeFunction the log level to use as a default
```
lutovich: Javadoc looks incorrect

community/logging/src/main/java/org/neo4j/logging/FormattedLog.java
```diff
@@ -198,19 +207,17 @@ public FormattedLog toPrintWriter( PrintWriter writer )
...
+                    dateTimeFormatter, dateTimeFunction );
         }
     }
 
-    private final Supplier<Date> currentDateSupplier;
     private final Supplier<PrintWriter> writerSupplier;
-    private final TimeZone timezone;
-    private final DateFormat dateFormat;
-    private final Object lock;
+    final TimeZone timezone;
```
lutovich: Imho it is better to keep these fields private and pass them separately to `FormattedLogger` constructor instead of `this`

community/logging/src/main/java/org/neo4j/logging/FormattedLog.java
```diff
@@ -41,23 +40,21 @@
...
     public static class Builder
     {
-        private TimeZone timezone = UTC;
+        private TimeZone timezone = FormattedLogger.UTC;
         private Object lock = this;
         private String category;
         private Level level = Level.INFO;
         private boolean autoFlush = true;
-        private DateFormat dateFormat = SIMPLE_DATE_FORMAT;
+        private Function<TimeZone, ZonedDateTime> dateTimeFunction = FormattedLogger.DEFAULT_CURRENT_DATE_TIME;
```
lutovich: Can we make this function `ZoneId -> ZonedDateTime`? It looks like implementation always calls `TimeZone#toZoneId()` and we can avoid this conversion.

It will then be possible to change `timezone` field to `zoneId` and use existing `ZoneOffset.UTC` constant.

integrationtests/src/test/java/org/neo4j/bolt/BoltMessageLoggingIT.java
```diff
@@ -148,7 +148,16 @@ public void shouldWriteErrorsToCustomFileWhenConfigured() throws IOException
         db.setConfig( bolt_logging_enabled, TRUE );
         db.setConfig( bolt_log_filename, customBoltLogFile.toString() );
         db.ensureStarted();
-        driver = newDriver();
+        try
+        {
+            driver = newDriver();
+        }
+        catch ( Exception e )
```
lutovich: Should this `try-catch` be removed?

community/logging/src/main/java/org/neo4j/logging/FormattedLog.java
lutovich: Actually, I think it will be simpler to expose only following methods:

```
#withDateTimeFormatter(DateTimeFormatter formatter);
#withTimeSupplier(Supplier<TemporalAccessor> timeSupplier);
```

formatter can include `ZoneId` and do the conversion. Then `FormattedLogger#time()` can simply call:
`formatter.format(timeSupplier.get())` and no `with*TimeZone()` methods are needed.

lutovich (commented):
@praveenag changes generally look good, couple things can be simplified

lutovich (approved):
@praveenag changes look fine. One small thing about javadocs. Deprecation tag should be `@deprecated` all lowercase not `@Deprecated`, otherwise it might not render correctly in the resulting webpage.

praveenag:
Merging after checking with @glindroth about random build failures.

PR #10019, opened 2017-09-07
Upgraded Shiro and commons-digester

eebus:
changelog: Upgraded shiro-core version to 1.4.0 and  commons-digester to  commons-digester3 3.2. 
This upgrade is due to a security problem with commons-beanutils-1.8.3. See https://nvd.nist.gov/vuln/detail/CVE-2014-0114 for more details.

### DIFF
--- a/enterprise/security/pom.xml
+++ b/enterprise/security/pom.xml
@@ -70,7 +70,7 @@
     <dependency>
       <groupId>org.apache.shiro</groupId>
       <artifactId>shiro-core</artifactId>
-      <version>1.3.2</version>
+      <version>1.4.0</version>
     </dependency>
     <dependency>
       <groupId>com.github.ben-manes.caffeine</groupId>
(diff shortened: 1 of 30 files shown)


enterprise/security/pom.xml
```diff
@@ -70,7 +70,7 @@
     <dependency>
       <groupId>org.apache.shiro</groupId>
       <artifactId>shiro-core</artifactId>
-      <version>1.3.2</version>
+      <version>1.4.0</version>
```
lassewesth: Don't do this. look at how versions are centrally managed for all other dependencies

klaren (commented):
Changes look good. However, since this will affect our costumers using embedded I think we should mark this as a changelog item and describe it's updated due to a security issue.

pom.xml
```diff
@@ -549,18 +578,30 @@
         <scope>compile</scope>
         <exclusions>
           <exclusion>
-            <artifactId>commons-digester</artifactId>
-            <groupId>commons-digester</groupId>
+            <artifactId>commons-digester3</artifactId>
+            <groupId>commons-digester3</groupId>
```
MishaDemianenko: no sure commons-configuration have both dependencies digester and digester3, so most probably you do not need to change this exclusion

PR #10022, opened 2017-09-08
Store highId backwards scan happens only when necessary

digitalstain:
During startup, the store needs to determine the high id it has. The source
 of that truth is in two places - the id generator file and the store, which
 via a backwards scan can determine (after recovery) which record is the
 highest in use. This last operation happened always and the result was
 compared with id generator and if the id generator was clean, it was
 effectively thrown away. In a store that is large and has lots of
 deleted records at the end, that backwards scan can take a long time
 and its result is probably not used after all.
This change makes that backwards scan happen only if the id generator
 is found to be unreliable, which currently means just created. The
 result is reduced startup time after a clean shutdown, without loss
 of safety.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/MetaDataStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/MetaDataStore.java
@@ -610,18 +610,23 @@ public class MetaDataStore extends CommonAbstractStore<MetaDataRecord,NoStoreHea
 
     private void setRecord( Position position, long value )
     {
-        long id = position.id;
-
         // We need to do a little special handling of high id in neostore since it's not updated in the same
         // way as other stores. Other stores always gets updates via commands where records are updated and
         // the one making the update can also track the high id in the event of recovery.
         // Here methods can be called directly, for example setLatestConstraintIntroducingTx where it's
         // unclear from the outside which record id that refers to, so here we need to manage high id ourselves.
-        setHighestPossibleIdInUse( id );
+        // Furthermore, we want the store to have a fixed size to accommodate the fixed number of fields it has. But,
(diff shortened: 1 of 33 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/store/MetaDataStore.java
```diff
@@ -610,18 +610,23 @@ private void scanAllFields( int pf_flags, Visitor<PageCursor,IOException> visito
...
         // unclear from the outside which record id that refers to, so here we need to manage high id ourselves.
-        setHighestPossibleIdInUse( id );
+        // Furthermore, we want the store to have a fixed size to accommodate the fixed number of fields it has. But,
+        // there is no guarantee that the field with the largest id will be called in a particular run. That means
+        // that to make sure we always have a proper max id, we must always set it to the highest possible value
+        // according to the enumeration of fields
+        Position[] values = Position.values();
+        long highestPossibleId = values[ values.length - 1 ].id;
+
+        setHighestPossibleIdInUse( highestPossibleId );
```
MishaDemianenko, line 625: this thing is fine for scope of this PR, but it still looks like additional useless works since it should be enough just to set that id once during initialization if required at all for metadata store.

MishaDemianenko (approved):
One minor, not really PR related note. Approved!

community/kernel/src/main/java/org/neo4j/kernel/impl/store/MetaDataStore.java
digitalstain: Correct. What i couldn't easily reason about was what happens on upgrade paths and whether i would then need to add code there as well. I took the safe route of always updating, which is what happened before anyway, so there is no performance regression in any way, just the same repetitive work but with the proper value. We should revisit the metadata store startup process eventually.

PR #10023, opened 2017-09-08
Revert usage of UTF8StringValue

pontusmelke:
Using UTF8StringValue adds an extra copy to the core API and shows up as
regression in benchmarks. Reverts that change until we can come up with a
way of pleasing both Bolt and core API.


tinwelint (approved):
👍

PR #10025, opened 2017-09-09
Does not delete store files on failed import

tinwelint:
The store files are very likely in an invalid state, but could be able to start
if opening a db on them, which would see very strange and inconsistent data.
Therefore logs failure/success and also tells user about this fact.


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
```diff
@@ -251,6 +253,11 @@ public void doImport( Input input ) throws IOException
                 nodeLabelsCache.close();
             }
         }
+
+        if ( success )
+        {
+            log.info( "Import completed successfully, took " + Format.duration( totalTimeMillis ) + ". " + storeUpdateMonitor );
+        }
```
chrisvest: I don't think we need the success variable, since the `catch ( Throwable t )` above will throw a laundered exception on failure, and thus avoiding this code path.

chrisvest (approved):
Functionally this is fine. Just one minor nit to pick, if you please.

PR #10026, opened 2017-09-10
Disable parameter extraction for GraphUrls for now

boggle:
It is getting in our way in CAPS and we don't need extraction there


Mats-SX:
CI build error is an unrelated HA test -- safe to merge.

PR #10036, opened 2017-09-11
Bump ancestor pom version to 46

jjaderberg:
to get rid of broken docs-maven-plugin.
This is a bump from 42 to 46, in which version the broken plugin was dropped. I don't have a perfect grasp of changes between 42..45, which are so far only used with 3.3 branch.


jjaderberg:
Let's do from 3.3 instead, because Lasse is a 🐔

PR #10038, opened 2017-09-11
Init high id in metadata store only once

MishaDemianenko:
Initialize high id in metadata store only once when store is open
instead of previous behaviour when that was happening constantly.


tinwelint (approved):
Oh, I thought this was about updating `highId` in general in stores on `nextId` and `nextIdBatch`. This is simpler than that. Nice anyway.

PR #10042, opened 2017-09-12
Allow not comparing plans again in CypherComparisonSupport

sherfert:
Comparing plans is not always desired and especially when porting tests
form NewPlannerTestSupport we might want to increase coverage for
result comparison but not plan comparison. Thus, this commit changes
the API to allow not comparing plans at all.


alexaverbuch (changes requested):
The compare plans capability was initially introduced as a means of asserting that all configurations' plans had some characteristic, e.g., "all planners create a plan with an index seek". 
The way that was done is to perform this check on one configuration's plan, and to then assert that all configurations have the same plan.

Asserting plan equality/similarity is usually too strict of an assert. It would be nice if instead of performing plan similarity `CypherComparisonSupport` provided a way to pass in a predicate (e.g., `(LogicalPlan)=>Boolean`) that specified just the check that needs to be performed (e.g., "plan contains index seek")

sherfert:
Now that we have the predicates PR ( https://github.com/neo4j/neo4j/pull/10061 ), I am closing this one.

PR #10050, opened 2017-09-12
Update DatabaseSchemaState backing map

MishaDemianenko:
Use plain CHM instead of map guarded by fair ReentrantReadWriteLock.
Update api a bit to provide expected atomicity.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/DatabaseSchemaState.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/DatabaseSchemaState.java
@@ -19,10 +19,8 @@
  */
 package org.neo4j.kernel.impl.api;
 
-import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


burqen (approved):
Nice cleanup!

community/kernel/src/main/java/org/neo4j/kernel/impl/api/DatabaseSchemaState.java
```diff
@@ -36,104 +34,38 @@
...
-        lock.writeLock().lock();
-        try
-        {
-            state = replacement;
-        }
-        finally
-        {
-            lock.writeLock().unlock();
-        }
+        return (V) state.computeIfAbsent( key, (Function<Object, Object>) creator );
```
tinwelint, line 56: That's a nice one-line replacement for all that cruft!

PR #10051, opened 2017-09-12
Not used dependencies declaration cleanup

MishaDemianenko:


tinwelint (approved):
Looks good. Did some light git history browsing and these were really old dependencies and mostly test dependencies.

PR #10052, opened 2017-09-12
Remove usage of JimFS

MishaDemianenko:
Remove JimFS usage, update couple of tests that was using it to use
DefaultFileSystemAbstraction instead.


chrisvest (approved):
👍

PR #10054, opened 2017-09-13
Make query state an explicit argument for interpreted expressions

systay:


systay:
I felt like we had too few PRs that need conflict resolving and rebasing, so here you are.

PR #10056, opened 2017-09-13
Do not raise error on missing context graphs

Mats-SX:
Regular Cypher queries (that do not handle anything pertaining
to multiple graphs) are supposed to be unaffected by multiple
graphs features, and simply use the ambient graph for all
context graph purposes.

### DIFF
--- a/community/cypher/frontend-3.3/src/main/scala/org/neo4j/cypher/internal/frontend/v3_3/ast/GraphReturnItems.scala
+++ b/community/cypher/frontend-3.3/src/main/scala/org/neo4j/cypher/internal/frontend/v3_3/ast/GraphReturnItems.scala
@@ -99,7 +99,8 @@ final case class GraphReturnItems(includeExisting: Boolean, items: Seq[GraphRetu
           newSourceName.map(source => newTargetName.map(target => ContextGraphs(source, target)).getOrElse(ContextGraphs(source, source)))
         optNewContextGraphs match {
           case Some(newContextGraphs) => s.updateContextGraphs(newContextGraphs)
-          case None => Left(SemanticError("No context graphs available", position))
+            // If there are no context graphs that's fine -- use the ambient graph
+          case None => Right(s)
         }
       }
     }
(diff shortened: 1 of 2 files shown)


community/cypher/frontend-3.3/src/main/scala/org/neo4j/cypher/internal/frontend/v3_3/ast/GraphReturnItems.scala
```diff
@@ -99,7 +99,8 @@ final case class GraphReturnItems(includeExisting: Boolean, items: Seq[GraphRetu
           newSourceName.map(source => newTargetName.map(target => ContextGraphs(source, target)).getOrElse(ContextGraphs(source, source)))
         optNewContextGraphs match {
           case Some(newContextGraphs) => s.updateContextGraphs(newContextGraphs)
-          case None => Left(SemanticError("No context graphs available", position))
+            // If there are no context graphs that's fine -- use the ambient graph
```
boggle, line 102: But we want this to be an error if it happens regularly - it would only be ok for the leading with, wouldn't it? Or are you suggesting a change in semantics - no context graph specified, back to ambient?

boggle: BTW we need to think this though for UNION, too, I think.

Mats-SX: For a query such as the one in the test, how does the context graph information flow?

```
MATCH (m) // no context graphs -- use ambient
WITH n // no context graphs still, as no graph return items -- this PR argues use ambient
RETURN n // no context graphs -- use ambient
```

The current model fails at the `WITH` -- there are no context graphs. This query should not fail, though. As I see it, there are two approaches:

a) Like this PR; no context graphs means ambient graph
b) Explicitly infer ambient graph as context graph if the leading clause is not a `WITH` with existing graph return items

The case you're describing is impossible to construct; there's no syntax that says 'project zero graphs', f.e. like `GRAPHS -`. You can however project two (non-context) graphs without explicitly setting new source and target; that's still an error and the other test in this PR makes sure that is still covered.

boggle: Cool, thanks!

PR #10057, opened 2017-09-13
Extracts log reading for tooling into more reusable class

tinwelint:
It's now much easier to write a quick transaction log stream analyzer.
DumpLogicalLog also uses this analyzer internally.

Also actually tests this pretty important functionality.

### DIFF
--- a/tools/src/main/java/org/neo4j/tools/dump/TransactionLogAnalyzer.java
+++ b/tools/src/main/java/org/neo4j/tools/dump/TransactionLogAnalyzer.java
@@ -1,0 +1,202 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


tools/src/main/java/org/neo4j/tools/dump/TransactionLogAnalyzer.java
```diff
@@ -0,0 +1,202 @@
...
+     * @param fileSystem {@link FileSystemAbstraction} to find the files on.
+     * @param storeDirOrLogFile {@link File} pointing either to a directory containing transaction log files, or directly
+     * pointing to a single transaction log file to analyze.
+     * @param invalidLogEntryHandler {@link InvalidLogEntryHandler} to pass in to the internal {@link LogEntryReader}.
+     * @param monitor {@link Monitor} receiving call-backs for all {@link Monitor#transaction(LogEntry[]) transactions},
+     * {@link Monitor#checkpoint(CheckPoint, LogPosition) checkpoints} and {@link Monitor#logFile(File, long) log file transitions}
+     * encountered during the analysis.
+     * @throws IOException on I/O error.
+     */
+    public static void analyze( FileSystemAbstraction fileSystem, File storeDirOrLogFile,
```
burqen, line 109: Can be package-private

burqen (approved):
Looks nice!

burqen:
@tinwelint I think a rebase will cure your build-problem. It's related to name change `legacy` -> `explicit`.

tinwelint:
OK I'll do that.

tools/src/main/java/org/neo4j/tools/dump/TransactionLogAnalyzer.java
tinwelint: It's a publically available tool in that component, so just because your IDE tells you that it _can_ be package private doesn't mean it _should_ :)

PR #10014, opened 2017-09-07
Settings validation for bloom extension

ragadeeshu:
Based on: https://github.com/neo4j/neo4j/pull/9888

PR #9888, opened 2017-08-23
Fulltext addon for bloom

ragadeeshu:

PR #10060, opened 2017-09-13
Populate the bloom fulltext index on startup

chrisvest:
This builds on #9888, and together with #10014 should give us functional completeness, though we'd ideally like to only rebuild the index when it's strictly necessary, e.g. if relevant configurations have changed.

### DIFF
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
@@ -79,6 +79,11 @@ public class FulltextProvider implements AutoCloseable
         db.registerTransactionEventHandler( fulltextTransactionEventUpdater );
     }
 
+    public void awaitPopulation() throws IOException
+    {
+        applier.writeBarrier().await();
+    }
+
     /**
      * Closes the provider and all associated resources.
      */
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
@@ -27,6 +27,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.kernel.AvailabilityGuard;
 import org.neo4j.logging.Log;
 
 /**
@@ -50,12 +51,13 @@ public class FulltextProvider implements AutoCloseable
      * Creates a provider of fulltext indices for the given database. This is the entry point for all fulltext index operations.
      * @param db Database that this provider should work with.
      * @param log For logging errors.
+     * @param availabilityGuard Used for waiting with populating the index until the database is available.
      */
-    public FulltextProvider( GraphDatabaseService db, Log log )
+    public FulltextProvider( GraphDatabaseService db, Log log, AvailabilityGuard availabilityGuard )
     {
         this.db = db;
         this.log = log;
-        applier = new FulltextUpdateApplier( log );
+        applier = new FulltextUpdateApplier( log, availabilityGuard );
         applier.start();
         fulltextTransactionEventUpdater = new FulltextTransactionEventUpdater( this, log, applier );
         nodeProperties = new HashSet<>();
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
@@ -53,7 +53,7 @@ public class FulltextProvider implements AutoCloseable
      * @param db Database that this provider should work with.
      * @param log For logging errors.
      * @param availabilityGuard Used for waiting with populating the index until the database is available.
-     * @param scheduler
+     * @param scheduler For background work.
      */
     public FulltextProvider( GraphDatabaseService db, Log log, AvailabilityGuard availabilityGuard,
                              JobScheduler scheduler )
@@ -84,6 +84,13 @@ public class FulltextProvider implements AutoCloseable
         db.registerTransactionEventHandler( fulltextTransactionEventUpdater );
     }
 
+    /**
+     * Wait for the asynchronous background population, if one is on-going, to complete.
+     *
+     * Such population, where the entire store is scanned for data to write to the index, will be started if the index
+     * needs to recover after an unclean shut-down, or a configuration change.
+     * @throws IOException If it was not possible to wait for the population to finish, for some reason.
+     */
     public void awaitPopulation() throws IOException
     {
         applier.writeBarrier().awaitCompletion();
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
@@ -83,7 +83,7 @@ public class FulltextProvider implements AutoCloseable
 
     public void awaitPopulation() throws IOException
     {
-        applier.writeBarrier().await();
+        applier.writeBarrier().awaitCompletion();
     }
 
     /**
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
@@ -29,6 +29,7 @@ import java.util.Set;
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.kernel.AvailabilityGuard;
 import org.neo4j.logging.Log;
+import org.neo4j.scheduler.JobScheduler;
 
 /**
  * Provider class that manages and provides fulltext indices. This is the main entry point for the fulltext addon.
@@ -52,12 +53,14 @@ public class FulltextProvider implements AutoCloseable
      * @param db Database that this provider should work with.
      * @param log For logging errors.
      * @param availabilityGuard Used for waiting with populating the index until the database is available.
+     * @param scheduler
      */
-    public FulltextProvider( GraphDatabaseService db, Log log, AvailabilityGuard availabilityGuard )
+    public FulltextProvider( GraphDatabaseService db, Log log, AvailabilityGuard availabilityGuard,
+                             JobScheduler scheduler )
     {
         this.db = db;
         this.log = log;
-        applier = new FulltextUpdateApplier( log, availabilityGuard );
+        applier = new FulltextUpdateApplier( log, availabilityGuard, scheduler );
         applier.start();
         fulltextTransactionEventUpdater = new FulltextTransactionEventUpdater( this, log, applier );
         nodeProperties = new HashSet<>();
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
@@ -35,7 +35,6 @@ import org.neo4j.logging.Log;
 public class FulltextProvider implements AutoCloseable
 {
     public static final String LUCENE_FULLTEXT_ADDON_INTERNAL_ID = "__lucene__fulltext__addon__internal__id__";
-    private static FulltextProvider instance;
     private final GraphDatabaseService db;
     private final Log log;
     private final FulltextTransactionEventUpdater fulltextTransactionEventUpdater;
@@ -45,6 +44,7 @@ public class FulltextProvider implements AutoCloseable
     private final Set<WritableFulltext> writableRelationshipIndices;
     private final Map<String,LuceneFulltext> nodeIndices;
     private final Map<String,LuceneFulltext> relationshipIndices;
+    private final FulltextUpdateApplier applier;
 
     /**
      * Creates a provider of fulltext indices for the given database. This is the entry point for all fulltext index operations.
@@ -55,7 +55,9 @@ public class FulltextProvider implements AutoCloseable
     {
         this.db = db;
         this.log = log;
-        fulltextTransactionEventUpdater = new FulltextTransactionEventUpdater( this, log );
+        applier = new FulltextUpdateApplier( log );
+        applier.start();
+        fulltextTransactionEventUpdater = new FulltextTransactionEventUpdater( this, log, applier );
         db.registerTransactionEventHandler( fulltextTransactionEventUpdater );
         nodeProperties = new HashSet<>();
         relationshipProperties = new HashSet<>();
@@ -72,6 +74,7 @@ public class FulltextProvider implements AutoCloseable
     public void close()
     {
         db.unregisterTransactionEventHandler( fulltextTransactionEventUpdater );
+        applier.stop();
         nodeIndices.values().forEach( luceneFulltextIndex ->
         {
             try
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
@@ -58,13 +58,25 @@ public class FulltextProvider implements AutoCloseable
         applier = new FulltextUpdateApplier( log );
         applier.start();
         fulltextTransactionEventUpdater = new FulltextTransactionEventUpdater( this, log, applier );
-        db.registerTransactionEventHandler( fulltextTransactionEventUpdater );
         nodeProperties = new HashSet<>();
         relationshipProperties = new HashSet<>();
         writableNodeIndices = new HashSet<>();
         writableRelationshipIndices = new HashSet<>();
         nodeIndices = new HashMap<>();
         relationshipIndices = new HashMap<>();
+    }
+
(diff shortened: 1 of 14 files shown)


enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
```diff
@@ -65,13 +68,32 @@ public FulltextProvider( GraphDatabaseService db, Log log )
...
+    }
+
     /**
      * Closes the provider and all associated resources.
      */
     @Override
     public void close()
     {
         db.unregisterTransactionEventHandler( fulltextTransactionEventUpdater );
+        applier.stop();
```
MishaDemianenko, line 106: Will it be impossible to stop while population is not completed right?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
```diff
@@ -65,13 +68,32 @@ public FulltextProvider( GraphDatabaseService db, Log log )
...
+        for ( WritableFulltext index : writableRelationshipIndices )
+        {
+            applier.populateRelationships( index, db );
+        }
+        db.registerTransactionEventHandler( fulltextTransactionEventUpdater );
+    }
+
+    public void awaitPopulation() throws IOException
+    {
+        applier.writeBarrier().await();
```
MishaDemianenko: it would be enough and much simpler to keep a map of indexes with their statuses (2 indexes)

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
```diff
@@ -65,13 +68,32 @@ public FulltextProvider( GraphDatabaseService db, Log log )
...
+            applier.populateNodes( index, db );
+        }
+        for ( WritableFulltext index : writableRelationshipIndices )
+        {
+            applier.populateRelationships( index, db );
+        }
+        db.registerTransactionEventHandler( fulltextTransactionEventUpdater );
+    }
+
+    public void awaitPopulation() throws IOException
```
MishaDemianenko: what is intended lifecycle of this call: who should call it? should w serve updates while indexes are not populated?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextTransactionEventUpdater.java
```diff
@@ -87,77 +86,32 @@ public Object beforeCommit( TransactionData data ) throws Exception
     @Override
     public void afterCommit( TransactionData data, Object state )
     {
-        //update node indices
         try
         {
+            List<BinaryLatch> completions = new ArrayList<>();
```
MishaDemianenko: what do we gaining by generating lots independent small updates? why not batch them to at least one per index?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
```diff
@@ -0,0 +1,365 @@
...
+        {
+            for ( PropertyEntry<E> propertyEntry : propertyEntries )
+            {
+                if ( index.properties().contains( propertyEntry.key() ) )
+                {
+                    long entityId = propertyEntry.entity().getId();
+                    Map<String,Object> allProperties = state.get( entityId );
+                    if ( allProperties == null || allProperties.isEmpty() )
+                    {
+                        index.getIndexWriter().deleteDocuments( newTermForChangeOrRemove( entityId ) );
```
MishaDemianenko, line 129: this thing still creates index writer each time

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
```diff
@@ -0,0 +1,365 @@
...
+            log.error( "Interrupted before " + workerThread.getName() + " could shut down.", e );
+        }
+    }
+
+    private interface FulltextIndexUpdate
+    {
+        Pair<WritableFulltext,BinaryLatch> applyUpdateAndReturnIndex() throws IOException;
+    }
+
+    private static class ApplierThread extends Thread
```
MishaDemianenko: We create our of thread here, maybe use Neo4jJobScheduler instead?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
```diff
@@ -0,0 +1,365 @@
...
+            this.log = log;
+            this.availabilityGuard = availabilityGuard;
+            setDaemon( true );
+        }
+
+        @Override
+        public void run()
+        {
+            waitForDatabaseToBeAvailable();
+            Set<WritableFulltext> refreshableSet = Collections.newSetFromMap( new IdentityHashMap<>() );
```
MishaDemianenko: why set can't be simple HashSet, for example? why do we need identity hash map in here?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+        private void waitForDatabaseToBeAvailable()
+        {
+            boolean isAvailable;
+            do
+            {
+                isAvailable = availabilityGuard.isAvailable( 100 );
+            }
+            while ( !isAvailable && !availabilityGuard.isShutdown() );
```
MishaDemianenko: not sure how useful `!availabilityGuard.isShutdown()` here since available it will be only in case if it's not shut down already and in the worst case, it can be shut down a moment after you check that shutdown flag anyway

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
```diff
@@ -0,0 +1,365 @@
...
+import org.neo4j.helpers.collection.Pair;
+import org.neo4j.kernel.AvailabilityGuard;
+import org.neo4j.kernel.api.impl.schema.writer.PartitionedIndexWriter;
+import org.neo4j.logging.Log;
+
+import static org.neo4j.kernel.api.impl.fulltext.LuceneFulltextDocumentStructure.documentForPopulation;
+import static org.neo4j.kernel.api.impl.fulltext.LuceneFulltextDocumentStructure.documentRepresentingProperties;
+import static org.neo4j.kernel.api.impl.fulltext.LuceneFulltextDocumentStructure.newTermForChangeOrRemove;
+
+class FulltextUpdateApplier
```
MishaDemianenko: why do we use only single threaded population here?  take a look how `BatchingMultipleIndexPopulator` does it for example if you did not yet

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/LuceneFulltextDocumentStructure.java
```diff
@@ -62,6 +62,14 @@ static Field encodeValueField( String propertyKey, Value value )
         return LuceneFulltextFieldEncoding.encodeField( propertyKey, value );
     }
 
+    public static Document documentForPopulation( long entityId, Map<String,Object> properties )
```
MishaDemianenko: don't think you need it as well since our standard index population know how to overcome that limitation already

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomFulltextIndexBenchmarks.java
```diff
@@ -0,0 +1,177 @@
...
+import org.neo4j.graphdb.Transaction;
+import org.neo4j.graphdb.factory.GraphDatabaseBuilder;
+import org.neo4j.test.TestGraphDatabaseFactory;
+import org.neo4j.test.rule.TestDirectory;
+import org.neo4j.test.rule.fs.DefaultFileSystemRule;
+import org.neo4j.test.rule.fs.FileSystemRule;
+
+@Ignore( "These are rudimentary benchmarks, but implemented via the jUnit framework to make them easy to run " +
+         "from an IDE." )
+public class BloomFulltextIndexBenchmarks
```
MishaDemianenko, line 40: Since it's a benchmark it should not live here at all, no reason to introduce dead code in the repo

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/LuceneFulltextUpdaterTest.java
```diff
@@ -666,17 +686,85 @@ public void shouldNotReturnNonMatches() throws Exception
...
+        {
+            fulltextFactory.createFulltextIndex( "nodes", NODES, singletonList( "prop" ), provider );
+            fulltextFactory.createFulltextIndex( "relationships", RELATIONSHIPS, singletonList( "prop" ), provider );
+            provider.init();
+            provider.awaitPopulation();
+
+            try ( ReadOnlyFulltext reader = provider.getReader( "nodes", NODES ) )
+            {
+
+                PrimitiveLongIterator hello = reader.query( "hello" );
```
MishaDemianenko, line 783: it's super hard to understand what this test is actually testing because of zebras of asserts, can those be grouped extracted in some way to be able to understand whats going on without  👓  and 🍷

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
```diff
@@ -0,0 +1,365 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.neo4j.concurrent.BinaryLatch;
```
MishaDemianenko, line 39: While it's understandable why you use latches I do not feel comfortable with that level of abstraction that escapes from this class, I think this should be simplified since we do not need that level of precision (per update) anyway and we can be quite relaxed in general about queue status

MishaDemianenko (changes requested):
The first round of review completed!
Would be great if we will update PR description a bit to describe what actually going on inside of this change request.
It's not really clear so far for me how clients should use provided procedures in terms of indexes availability and how we can guard them against miss use, and other questions here and there.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
chrisvest: As it stands, yeah. We have to pull the population out as an explicit life cycle phase, for it to work well with the recovery as well, if we allow clean shut-down mid-population.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
chrisvest: I think it's an important architectural flexibility that we can submit updates asynchronously, and then wait for their completion. It has not yet been decided if we want the index to be eventually consistent or synchronous. We don't have to expose the concrete BinaryLatch, though. I can hide that behind an interface.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextTransactionEventUpdater.java
chrisvest: The applier batches as it consumes from the queue, but you want to put the loop into the applier? We can do that.

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomFulltextIndexBenchmarks.java
chrisvest: We'll delete them when we the architecture is more settled.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
chrisvest: I'll add some javadoc.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
chrisvest: It's missing infrastructure for indexing relationships, so it's not an integration we can do at this point.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
chrisvest: The writer is cached in the `index` now.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/LuceneFulltextDocumentStructure.java
chrisvest: It makes no observable difference to performance to build a list of capturing lambdas instead, which we'd have to do because the call needs to know the size of the iterator up front. So I'm not sure it's worth the trouble in this case.

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/LuceneFulltextUpdaterTest.java
chrisvest: Let's do that in #10079 where we've already made significant changes to this file.

chrisvest:
@MishaDemianenko I think I've addressed all your comments.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextTransactionEventUpdater.java
MishaDemianenko: yeah basically generate only one update that ges to applier queue at least per index per commit

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
MishaDemianenko: still open question right?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
MishaDemianenko: ok

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
MishaDemianenko: hm you can still start multiple threads (maybe not in thins PR) to perfrom index population to make it times faster then from single thread

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
MishaDemianenko: still open question here

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/LuceneFulltextDocumentStructure.java
MishaDemianenko: it's more a question to have common way of doing things and avoid bicycles here and there that do the same thing but on their own special way

MishaDemianenko:
@chrisvest @ragadeeshu couple of open questions still, most of the comments solved already by new commits

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
chrisvest: Yeah, I've created a card for it.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/LuceneFulltextDocumentStructure.java
chrisvest: Alright.

chrisvest:
@MishaDemianenko Removed the document allocation.

MishaDemianenko (approved):

PR #10062, opened 2017-09-13
Latest netty and support OpenSSL

martinfurmanski:


community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslSystemSettings.java
```diff
@@ -0,0 +1,39 @@
...
+import org.neo4j.graphdb.config.Setting;
+
+import static org.neo4j.kernel.configuration.Settings.options;
+import static org.neo4j.kernel.configuration.Settings.setting;
+
+/**
+ * System-wide settings for SSL.
+ */
+@Description( "System-wide settings for SSL." )
+public class SslSystemSettings implements LoadableConfig
```
lutovich, line 35: Maybe simply `SslSettings`? Otherwise "ssl system" is slightly confusing.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslSystemSettings.java
```diff
@@ -0,0 +1,39 @@
...
+{
+    @Description( "SSL provider" )
+    public static final Setting<SslProvider> ssl_provider = setting( "dbms.ssl.provider", options( SslProvider.class ), SslProvider.JDK.name() );
```
lutovich: Maybe worth inventing our own enum to not depend on what netty provides?
Otherwise users might request provider that's not in the classpath or is unstable, like `OPENSSL_REFCNT`.

lutovich (commented):
@martinfurmanski looks good, made two comments.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslSystemSettings.java
martinfurmanski: Well, it is to distinguish it from other ssl settings. Simply SslSettings is not good enough.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslSystemSettings.java
martinfurmanski: I have no strong opinion really. On one side it seems like unnecessary duplication since Netty should at least be stable on what it supports. Getting rid of the possibility of OPENSSL_REFCOUNT would be the main benefit for me.

lutovich: Elements of `SslProvider` might refer to providers that require separate classpath dependency, like `OPENSSL`. So it might happen users configure what we do not have in the classpath, right?

I think it is probably safer to duplicate this bit.

martinfurmanski: Sure, but that will not be protected from either with this enum or a separate enum.

martinfurmanski: To be clear: OpenSSL is not included as a dependency in this PR, so it cannot be enabled successfully without the user supplying a library suitable for its platform.

martinfurmanski: @lutovich I updated this to be netty_ssl_provider since it really has nothing to do with the JCE/JSSE providers. This also makes it more sensical to keep the netty enum.

lutovich: Oh I did not notice that OpenSSL provider dependency has scope `test`. Maybe we should include it and use OpenSSL by default if we know it is faster?

martinfurmanski: It might be a reasonable option, but it is not so flexible. The only thing which has a complete dependency for inclusion is a statically linked uber jar for osx/linux/win with google's fork of openssl called boringssl.

Security-conscious users might want to choose their own version of openssl and even use the version which they have installed on the system, but that requires a completely different netty-tcnative dep.

There is also some intrinsic value in the JDK implementation since it f.ex. can't have memory leaks and buffer overflow issues, so it is not necessarily a bad default even if it performs horribly.

PR #10063, opened 2017-09-13
fix remoteAddress displayed based on feedback

praveenag:


community/bolt/src/main/java/org/neo4j/bolt/logging/BoltMessageLoggerImpl.java
```diff
@@ -51,11 +53,21 @@
...
+    {
+        return () -> remoteAddress( channel );
+    }
+
     BoltMessageLoggerImpl( BoltMessageLog messageLog, Channel channel )
     {
+        this( messageLog, channel, remoteAddressSupplier( channel ) );
+    }
+
+    BoltMessageLoggerImpl( BoltMessageLog messageLog, Channel channel, Supplier<String> remoteAddressSupplier )
```
lutovich: I think having supplier of remote address is an overkill, it makes code more complex. We can simply accept a `Channel` and mock it in tests, if needed.

lutovich (commented):
@praveenag changes look good. I think small simplification can be made.

community/bolt/src/main/java/org/neo4j/bolt/logging/BoltMessageLoggerImpl.java
praveenag: I recognise that this is really ugly. But the reason I did this because it is impossible to mock a final method as explained [here](https://github.com/mockito/mockito/wiki/What's-new-in-Mockito-2#mock-the-unmockable-opt-in-mocking-of-final-classesmethods)

community/bolt/src/main/java/org/neo4j/bolt/logging/BoltMessageLoggerImpl.java
```diff
@@ -191,7 +203,11 @@ private String getCorrelationId()
 
     private static String remoteAddress( Channel channel )
     {
-        return channel.remoteAddress().toString();
+        InetSocketAddress remoteAddress = (InetSocketAddress) channel.remoteAddress();
```
technige: Could this cast fail?

praveenag: It would never fail because we always create a SocketChannel. Having said that Never say Never. I have fixed it by handling that as an edge case.

PR #10065, opened 2017-09-14
Handle nested lists for UNWIND in compiled runtime

pontusmelke:
Fix a bug where we weren't properly handling types in nested lists.

### DIFF
--- a/enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/LogicalPlanConverter.scala
+++ b/enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/LogicalPlanConverter.scala
@@ -526,6 +526,10 @@ object LogicalPlanConverter {
         case CypherCodeGenType(symbols.ListType(innerType), ListReferenceType(innerReprType))
           if RepresentationType.isPrimitive(innerReprType) =>
           (CypherCodeGenType(innerType, innerReprType), UnwindPrimitiveCollection(opName, collection))
+        case CypherCodeGenType(symbols.ListType(innerType), ListReferenceType(innerReprType)) =>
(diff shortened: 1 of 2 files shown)


enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/LogicalPlanConverter.scala
```diff
@@ -526,6 +526,10 @@ object LogicalPlanConverter {
         case CypherCodeGenType(symbols.ListType(innerType), ListReferenceType(innerReprType))
           if RepresentationType.isPrimitive(innerReprType) =>
           (CypherCodeGenType(innerType, innerReprType), UnwindPrimitiveCollection(opName, collection))
+        case CypherCodeGenType(symbols.ListType(innerType), ListReferenceType(innerReprType)) =>
+          (CypherCodeGenType(innerType, innerReprType), ir.UnwindCollection(opName, collection))
+        case CypherCodeGenType(symbols.ListType(innerType), _) =>
```
sherfert: Now that case is duplicate... that is probably not intentional?

pontusmelke: No not quite, in the case where the inner type is primitive we use `UnwindPrimitiveCollection` and in the other we use ` ir.UnwindCollection`

sherfert: Sorry, I should have been more specific. Lines 531 and 533.

pontusmelke: yes you are right, that was a mistake :)

sherfert (approved):

PR #10068, opened 2017-09-14
Fix NullPointerException in ExecutionPlanWrapper.

sherfert:
In rare situations, the wrapping of older query plans would fail
because a transaction had been closed. By eagerly using the transaction,
we make sure not to fail like this.

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_2/Compatibility.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_2/Compatibility.scala
@@ -89,77 +89,21 @@ trait Compatibility[C <: CompilerContext] {
         // Log notifications/warnings from planning
         planImpl.notifications(planContext).foreach(notificationLogger.log)
 
-        (new ExecutionPlanWrapper(planImpl, transactionalContext,preParsingNotifications, pos3_2), extractedParameters)
+        val searchMonitor = kernelMonitors.newMonitor(classOf[IndexSearchMonitor])
+        val executionPlanWrapper = new ExecutionPlanWrapper(
+          planImpl,
+          transactionalContext,
+          preParsingNotifications,
+          pos3_2,
+          searchMonitor,
+          executionMonitor)
+        (executionPlanWrapper, extractedParameters)
       }
 
       override protected val trier: Try[BaseState] = preparedSyntacticQueryForV_3_2
     }
   }
 
-  class ExecutionPlanWrapper(inner: ExecutionPlan_v3_2, transactionalContext: TransactionalContextWrapperV3_3,
-                             preParsingNotifications: Set[org.neo4j.graphdb.Notification], offSet: frontend.v3_2.InputPosition)
-    extends ExecutionPlan {
-
-    private val searchMonitor = kernelMonitors.newMonitor(classOf[IndexSearchMonitor])
-
-    private def queryContext(transactionalContext: TransactionalContextWrapperV3_3) = {
-      val ctx = new TransactionBoundQueryContext(TransactionalContextWrapperV3_2(transactionalContext.tc))(
-        searchMonitor)
-      new ExceptionTranslatingQueryContext(ctx)
-    }
-
-    def run(transactionalContext: TransactionalContextWrapperV3_3, executionMode: CypherExecutionMode,
-            params: Map[String, Any]): Result = {
-      val innerExecutionMode = executionMode match {
-        case CypherExecutionMode.explain => ExplainModev3_2
-        case CypherExecutionMode.profile => ProfileModev3_2
-        case CypherExecutionMode.normal => NormalModev3_2
-      }
-      exceptionHandler.runSafely {
-        val innerParams = typeConversions.asPrivateMap(params)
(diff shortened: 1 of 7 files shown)


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_2/CostCompatibility.scala
```diff
@@ -53,7 +53,7 @@ case class CostCompatibility[C <: CompilerContext, T <: Transformer[C, Compilati
       case CypherRuntime.default => None
       case CypherRuntime.interpreted => Some(InterpretedRuntimeName)
       case CypherRuntime.compiled => Some(CompiledRuntimeName)
-      case _ => throw new IllegalArgumentException("Runtime is not supported in Cypher 2.3")
+      case _ => throw new IllegalArgumentException("Runtime is not supported in Cypher 3.2")
```
pontusmelke, line 56: oops...

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_2/Compatibility.scala
```diff
@@ -89,84 +89,97 @@ trait Compatibility[C <: CompilerContext] {
...
-        case LegacyRelationshipIndexUsage(identifier, index) => explicitIndexUsage(identifier, "RELATIONSHIP", index)
-      }.asJava)
-    }
 
-    override def run(transactionalContext: TransactionalContextWrapperV3_3, executionMode: CypherExecutionMode, params: MapValue): Result = {
-      var map: mutable.Map[String, Any] = mutable.Map[String, Any]()
-      params.foreach(new BiConsumer[String, AnyValue] {
-        override def accept(t: String, u: AnyValue): Unit = map.put(t, valueHelper.fromValue(u))
-      })
+  private def queryContext(transactionalContext: TransactionalContextWrapperV3_3) = {
```
pontusmelke, line 126: Why do we need to change 3.2 so much more than 3.1 and 2.3?

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_2/CompatibilityTest.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.mockito.stubbing.Answer
+import org.neo4j.cypher.internal.compiler.v3_2.executionplan.{SchemaIndexSeekUsage, ExecutionPlan => ExecutionPlan_v3_2}
+import org.neo4j.cypher.internal.frontend.v3_2.InputPosition
+import org.neo4j.cypher.internal.frontend.v3_3.test_helpers.CypherFunSuite
+import org.neo4j.cypher.internal.spi.v3_2.TransactionBoundQueryContext.IndexSearchMonitor
+import org.neo4j.cypher.internal.spi.v3_3.{TransactionalContextWrapper => TransactionalContextWrapperV3_3}
+import org.neo4j.kernel.api.ReadOperations
+import org.neo4j.kernel.impl.query.QueryExecutionMonitor
+
+class CompatibilityTest extends CypherFunSuite {
```
pontusmelke, line 33: Tests for 2.3 and 3.1 as well?

pontusmelke (commented):
Nice. Is it possible to add some integration test as well that is not using mocks? There was one query on the trello card

```
test("should refuse to execute index seeks using inequalities over different types") {
    // Given
    createLabeledNode(Map("prop" -> 15), "Label")
    createLabeledNode(Map("prop" -> "1"), "Label")

    (1 to 400).foreach { _ =>
      createLabeledNode("Label")
    }
    graph.createIndex("Label", "prop")

    val query = "MATCH (n:Label) WHERE n.prop >= '1' AND n.prop > 10 RETURN n.prop AS prop"


    an[IllegalArgumentException] should be thrownBy {
      executeWith(Configs.Empty, query).toList
    }
    executeWith(Configs.Interpreted, s"EXPLAIN $query") should use(IndexSeekByRange.name)
  }
```

could we add that as integration test?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_2/Compatibility.scala
sherfert: The reason is that ExecutionPlanWrapper was defined as in inner class of Compatibility (across all versions). This is not really needed and made it impossible for us to write the unit test. We decided to extract it to its own class for 3.2, in order to be able to test it. 

We could do the same for the other versions, but we could also do that on demand, if needed, which I think makes sense.

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_2/CompatibilityTest.scala
sherfert: The error did only affect version 3.2. The test in this fashion would not make much sense for the other versions, since
`override val plannerInfo` does not use `transactionalContext` there, and the whole test is centered around the mock for that object.

PR #10071, opened 2017-09-14
Make sure built-in procedures close resources when done

systay:
The first two commits are backports from 3.3 and should be forward-merged with a null commit. The last commit is applicable to both 3.2 and 3.3

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
@@ -68,8 +68,10 @@ public class BuiltInProcedures
         Statement statement = tx.acquireStatement();
         try
         {
-            // Ownership of the reference to the acquired statement is transfered to the returned iterator stream
-            return TokenAccess.LABELS.inUse( statement ).map( LabelResult::new ).stream();
+            // Ownership of the reference to the acquired statement is transfered to the returned iterator stream,
+            // but we still want to eagerly consume the labels, so we can catch any exceptions,
+            List<LabelResult> labelResults = asList( TokenAccess.LABELS.inUse( statement ).map( LabelResult::new ) );
+            return labelResults.stream();
         }
         catch ( Throwable t )
         {
@@ -85,8 +87,11 @@ public class BuiltInProcedures
         Statement statement = tx.acquireStatement();
         try
         {
-            // Ownership of the reference to the acquired statement is transfered to the returned iterator stream
-            return TokenAccess.PROPERTY_KEYS.inUse( statement ).map( PropertyKeyResult::new ).stream();
+            // Ownership of the reference to the acquired statement is transfered to the returned iterator stream,
+            // but we still want to eagerly consume the labels, so we can catch any exceptions,
+            List<PropertyKeyResult> propertyKeys =
+                    asList( TokenAccess.PROPERTY_KEYS.inUse( statement ).map( PropertyKeyResult::new ) );
+            return propertyKeys.stream();
         }
         catch ( Throwable t )
         {
@@ -102,8 +107,11 @@ public class BuiltInProcedures
         Statement statement = tx.acquireStatement();
         try
         {
-            // Ownership of the reference to the acquired statement is transfered to the returned iterator stream
-            return TokenAccess.RELATIONSHIP_TYPES.inUse( statement ).map( RelationshipTypeResult::new ).stream();
+            // Ownership of the reference to the acquired statement is transfered to the returned iterator stream,
+            // but we still want to eagerly consume the labels, so we can catch any exceptions,
+            List<RelationshipTypeResult> relationshipTypes =
+                    asList( TokenAccess.RELATIONSHIP_TYPES.inUse( statement ).map( RelationshipTypeResult::new ) );
+            return relationshipTypes.stream();
         }
         catch ( Throwable t )
         {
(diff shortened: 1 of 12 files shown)


community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
```diff
@@ -67,17 +67,20 @@
...
     @Description( "List all property keys in the database." )
     @Procedure( name = "db.propertyKeys", mode = READ )
     public Stream<PropertyKeyResult> listPropertyKeys()
     {
+
         try ( Statement statement = tx.acquireStatement() )
         {
-            return TokenAccess.PROPERTY_KEYS.inUse( statement ).map( PropertyKeyResult::new ).stream();
+            // Ownership of the reference to the acquired statement is transfered to the returned iterator stream
+            return asList( TokenAccess.PROPERTY_KEYS.inUse( statement ).map( PropertyKeyResult::new ) ).stream();
```
tinwelint: Yup, this eagerness here is the preferred way to go

community/kernel/src/main/java/org/neo4j/kernel/impl/api/TokenAccess.java
```diff
@@ -143,7 +143,11 @@ private static boolean hasAny( Iterator<?> iter )
         @Override
         public void close()
         {
-            statement.close();
```
tinwelint, line 146: Is this strictly necessary or just being over the top safe?

community/kernel/src/main/java/org/neo4j/kernel/impl/proc/ReflectiveProcedureCompiler.java
```diff
@@ -671,7 +671,11 @@ public boolean hasNext() throws ProcedureException
             @Override
             public void close() throws Exception
             {
-                closeable.close();
+                if ( closeable != null )
```
tinwelint, line 674: Is this strictly necessary or just being over the top safe?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/TokenAccess.java
systay: This was back-ported from the 3.3 branch. Not sure - what do you think, @henriknyman?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ProcedureCallSupportAcceptanceTest.scala
```diff
@@ -87,4 +87,26 @@ class ProcedureCallSupportAcceptanceTest extends ProcedureCallAcceptanceTest {
       map("one", "alpha", "newTwo", "beta")
     ))
   }
+
+  test("should be able to execute union of multiple token accessing procedures") {
```
thobe, line 91: This test fails with:

```
org.neo4j.graphdb.NotInTransactionException: The statement has been closed.
```

thobe:
Just to verify my understanding of this changeset. What is supposed to fix the issue is mainly:

- [Line 102 in DefaultExecutionResultBuilderFactory](https://github.com/neo4j/neo4j/pull/10071/files#diff-9e953639cb3fbc71d1fa8d742723cce0R102)
- [Line 102 in ProcedureExecutionResult](https://github.com/neo4j/neo4j/pull/10071/files#diff-1257d865b16a46ad191c188d40ca02ccR102), and
- [Line 54 in CompiledExecutionResult](https://github.com/neo4j/neo4j/pull/10071/files#diff-123545a8c821d04dbcd74857dc4c0545R54)

Then there are changes to some of the procedures to ensure that they finish their reads before the statement is closed.

Is that correct?

systay:
We are trying to make the test `should be able to execute union of multiple token accessing procedures` go green, and those are the changes needed.

henriknyman:
@thobe The three things you mention are a secondary issue when we run in PROFILE mode. The main issue is that the three built-in procedures could not close the statement while returning a lazy stream.

henriknyman:
@tinwelint The reason we have to transfer ownership of the statement into `TokenAccess` is because that is the way the Core API works. The statement will be closed when the iterator/stream is exhausted.
We could make the built-in procedures eager if we are worried about leaking a statement in case of an exception, but we cannot do try-with-resource since the statement will be closed by exhausting the stream in the happy case.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/TokenAccess.java
henriknyman: Hmm I don't remember.. it could be the latter.

tinwelint:
@henriknyman OK right, so that will only get closed in the happy case (no exception and stream fully consumed) if letting the stream manage the closing of the resource. This eagerness feels better and is typically on very small number of tokens.

tinwelint (approved):

thobe:
@systay / @henriknyman: this is still not where it needs to be:

```
java.lang.AssertionError: Failed on procedure CALL db.propertyKeys()
    at org.neo4j.kernel.builtinprocs.ProcedureResourcesIT.verifyProcedureCloseAllAcquiredKernelStatements(ProcedureResourcesIT.java:118)
    at org.neo4j.kernel.builtinprocs.ProcedureResourcesIT.allProcedures(ProcedureResourcesIT.java:82)
```

PR #10074, opened 2017-09-14
Revert "Changes how MarshlandPool works with inner transactions"

tinwelint:
This reverts commit a966220998c8b8b5ba870e923aadd4090cc1a3ba.

This wasn't a strictly necessary change, mostly made to not require changing many tests relying on certain id ordering.

### DIFF
--- a/community/graph-algo/src/main/java/org/neo4j/graphalgo/impl/path/ShortestPath.java
+++ b/community/graph-algo/src/main/java/org/neo4j/graphalgo/impl/path/ShortestPath.java
@@ -162,7 +162,7 @@ public class ShortestPath implements PathFinder<Path>
                 goOneStep( endData, startData, hits, startData, stopAsap );
             }
             Collection<Hit> least = hits.least();
-            return least != null ? filterPaths( hitsToPaths( least, start, end, stopAsap ) ) : Collections.emptyList();
+            return least != null ? filterPaths( hitsToPaths( least, start, end, stopAsap, maxResultCount ) ) : Collections.emptyList();
         }
     }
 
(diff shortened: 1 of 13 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathAcceptanceTest.scala
```diff
@@ -558,6 +558,10 @@ class ShortestPathAcceptanceTest extends ExecutionEngineFunSuite with NewPlanner
     val a2 = createLabeledNode("A")
     val a3 = createLabeledNode("A")
     val a4 = createLabeledNode("A")
+    println( a1 )
```
MishaDemianenko: should not be here

community/graph-algo/src/main/java/org/neo4j/graphalgo/impl/path/ShortestPath.java
```diff
@@ -162,7 +162,7 @@ private void resolveMonitor( Node node )
                 goOneStep( endData, startData, hits, startData, stopAsap );
             }
             Collection<Hit> least = hits.least();
-            return least != null ? filterPaths( hitsToPaths( least, start, end, stopAsap ) ) : Collections.emptyList();
+            return least != null ? filterPaths( hitsToPaths( least, start, end, stopAsap, maxResultCount ) ) : Collections.emptyList();
```
MishaDemianenko, line 165: would be nice to see test for this change

MishaDemianenko (changes requested):
Couple of small comments

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathAcceptanceTest.scala
tinwelint: Argh

community/graph-algo/src/main/java/org/neo4j/graphalgo/impl/path/ShortestPath.java
tinwelint: Well, TestShortestPath tests this. It's now green.

tinwelint: Since we're gonna run one more build w/ the println removal anyway, I can create an addition test perhaps

MishaDemianenko (approved):

PR #10075, opened 2017-09-15
Use online index sampling when populating from Label index

burqen:
When label index was implemented on top of Lucene, the data during scan of label index
could be obsolete resulting in multiple updates for same node during population.
Indexes themselves could handle this, but samplers could not. Therefore sampling was done
as a separate step after population was done.
Since Label index is implemented on top of GBPTree now, this is not longer necessary.

Expectation is that this should improve performance during index population.


burqen:
It seems number index population see some benefit, but strings see small regression.
http://alacrity.neohq.net/?tab=benchmarksOverview&benchmarkGroup=Core%2520API&neo4jVersion=3.3.0&neo4jVersion=burqen%25203.3-allow-index-sampling-with-label-scan&benchmark=CreateIndexNonUnique.createIndex_%2528index%252CSCHEMA%2529&date=19-8-2017

Adding the code simplification into the mix, I personally think it looks worth merging.

MishaDemianenko (approved):

PR #10079, opened 2017-09-15
Special characters and the Fulltext Index Add-On

chrisvest:
Builds on #10060.

### DIFF
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/SimpleFulltextReader.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/SimpleFulltextReader.java
@@ -35,6 +35,8 @@ import org.neo4j.kernel.api.impl.index.collector.DocValuesCollector;
 import org.neo4j.kernel.api.impl.index.partition.PartitionSearcher;
 import org.neo4j.kernel.api.impl.schema.reader.IndexReaderCloseException;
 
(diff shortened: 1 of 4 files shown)


enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextUpdateApplier.java
```diff
@@ -149,7 +149,7 @@ private BinaryLatch enqueuePopulateIndex( WritableFulltext index, GraphDatabaseS
         {
             PartitionedIndexWriter indexWriter = index.getIndexWriter();
             String[] indexedPropertyKeys = index.properties().toArray( new String[0] );
-            try ( Transaction tx = db.beginTx() )
+            try ( Transaction ignore = db.beginTx( 10, TimeUnit.HOURS ) )
```
tinwelint: Couldn't this applier be started after recovery instead? To have this timeout _should_ be fine, but it's better to avoid timeouts altogether and change ordering, if possible.

tinwelint (commented):
Looks nice. Had one life cycle comment about recovery vs applier start (apparently shows up as outdated).

chrisvest:
@MishaDemianenko dropped the test simplifications here.

chrisvest:
@tinwelint Please take another look, since a few more changes have gone in since you looked at it last.

tinwelint (approved):
Oh OK my comment was on the PR this was based on...

PR #10083, opened 2017-09-18
Never namespace graph variables

Mats-SX:
Graph variables are only introduced in a projection clause, and have
duplication of names checked at that point.

- Do not re-use variable instance when normalising


boggle (approved):
Looks good!

PR #10087, opened 2017-09-18
Label and schema index log internal recovery

burqen:
Log report of crashed pointer cleanup of underlying GBPTree to label and schema index.

This logging was accidentally removed from label index at time when lucene label scan store was removed. This PR re-introduces it and also add recovery logging for native schema index.

### DIFF
--- a/community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
+++ b/community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
@@ -226,12 +226,14 @@ public class ConsistencyCheckService
         } ) );
 
         // Bootstrap kernel extensions
+        Monitors monitors = new Monitors();
         LifeSupport life = new LifeSupport();
         KernelExtensions extensions = life.add( instantiateKernelExtensions( storeDir,
                 fileSystem, config, new SimpleLogService( logProvider, logProvider ), pageCache,
                 RECOVERY_PREVENTING_COLLECTOR,
                 // May be enterprise edition, but in consistency checker we only care about the operational mode
-                COMMUNITY ) );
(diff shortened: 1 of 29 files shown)


tinwelint (changes requested):
I'm very much in favor of never passing in explicit `Log` into anything, instead doing such by means of `Monitor`... always. NativeLabelScanStore does this and I do think that this should be the case here too. WDYT?

burqen:
@tinwelint We already had `log` in there for this guy

```
    @Override
    public InternalIndexState getInitialState( long indexId, IndexDescriptor descriptor )
    {
        try
        {
            NativeSchemaIndexHeaderReader headerReader = new NativeSchemaIndexHeaderReader();
            GBPTree.readHeader( pageCache, nativeIndexFileFromIndexId( indexId ), new ReadOnlyMetaNumberLayout(),
                    headerReader );
            switch ( headerReader.state )
            {
            case BYTE_FAILED:
                return InternalIndexState.FAILED;
            case BYTE_ONLINE:
                return InternalIndexState.ONLINE;
            case BYTE_POPULATING:
                return InternalIndexState.POPULATING;
            default:
                throw new IllegalStateException( "Unexpected initial state byte value " + headerReader.state );
            }
        }
        catch ( IOException e )
        {
            log.error( "Failed to open index:" + indexId + ", requesting re-population.", e );
            return InternalIndexState.POPULATING;
        }
    }
```

So I took that shortcut. You make a good point though. Let's do it with monitors instead.

burqen:
@tinwelint Updated schema index to use monitor for logging

tinwelint (approved):
Looks sharp!

burqen:
@tinwelint Updated with satisfied dependencies for consistency checker

PR #10090, opened 2017-09-19
Extracts log reading for tooling into more reusable class

tinwelint:
Replaces #10057


burqen:
@tinwelint Build does not look very happy. Could be something that is not fully setup for new `3.4` branch perhaps?

PR #10093, opened 2017-09-19
Only perform check on conflicting mode when `@PerformsWrites` is present

OliviaYtterbrink:
Fixes an issue where the compiler would complain about conflicting modes even when not using `@PerformsWrites`

### DIFF
--- a/community/procedure-compiler/integration-tests/src/test/java/org/neo4j/tooling/procedure/ProcedureTest.java
+++ b/community/procedure-compiler/integration-tests/src/test/java/org/neo4j/tooling/procedure/ProcedureTest.java
@@ -80,6 +80,21 @@ public class ProcedureTest
     }
 
     @Test
+    public void calls_procedures_with_different_modes_returning_void()
+    {
+        try ( Driver driver = GraphDatabase.driver( graphDb.boltURI(), configuration() );
+                Session session = driver.session() )
+        {
+            session.run( "CALL " + procedureNamespace + ".performsWrites()" );
+            session.run( "CALL " + procedureNamespace + ".defaultMode()" );
+            session.run( "CALL " + procedureNamespace + ".readMode()" );
+            session.run( "CALL " + procedureNamespace + ".writeMode()" );
+            session.run( "CALL " + procedureNamespace + ".schemaMode()" );
+            session.run( "CALL " + procedureNamespace + ".dbmsMode()" );
+        }
+    }
+
+    @Test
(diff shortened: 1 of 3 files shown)


community/procedure-compiler/integration-tests/src/test/java/org/neo4j/tooling/procedure/ProcedureTest.java
```diff
@@ -80,6 +80,21 @@ public void calls_procedures_with_simple_input_type_returning_void()
     }
 
     @Test
+    public void calls_procedures_with_different_modes_returning_void()
```
craigtaverner, line 83: Does this test actually test anything? Do session.run() methods throw exceptions on failure?

community/procedure-compiler/processor/src/main/java/org/neo4j/tooling/procedure/visitors/ProcedureVisitor.java
```diff
@@ -102,7 +101,11 @@ public ProcedureVisitor( Types typeUtils, Elements elementUtils, boolean ignores
 
     private Stream<CompilationMessage> validatePerformsWriteUsage( ExecutableElement executableElement )
     {
-        return performsWriteVisitor.visit( executableElement );
+        if ( executableElement.getAnnotation( PerformsWrites.class ) != null )
```
craigtaverner, line 104: Do we still support a 'PerformsWrites' annotation, and if so, should it not behave the same as `mode=WRITE`? This method seems specific to only the deprecated version?

craigtaverner (changes requested):

community/procedure-compiler/integration-tests/src/test/java/org/neo4j/tooling/procedure/ProcedureTest.java
OliviaYtterbrink: Yes, it will throw exception if you can't execute the query.

community/procedure-compiler/processor/src/main/java/org/neo4j/tooling/procedure/visitors/ProcedureVisitor.java
OliviaYtterbrink: `PerformsWrites` is deprecated but not removed and can't be paired up with `Procedure.mode` at all. `PerformsWrites` is indeed the same as specifying `mode=WRITE`.
The validation is only required when using the deprecated annotation.

craigtaverner (approved):

PR #10096, opened 2017-09-19
Make bloom extension loadable, disabled by default

ragadeeshu:
Based on #10060 
Also verifies that the consistency checker does not crash.

### DIFF
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomKernelExtensionFactory.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomKernelExtensionFactory.java
@@ -22,6 +22,7 @@ package org.neo4j.kernel.api.impl.fulltext.integrations.bloom;
 import java.io.File;
 
 import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.helpers.Service;
 import org.neo4j.io.fs.FileSystemAbstraction;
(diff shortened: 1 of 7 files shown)


enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomKernelExtensionFactory.java
```diff
@@ -41,7 +41,7 @@
 public class BloomKernelExtensionFactory extends KernelExtensionFactory<BloomKernelExtensionFactory.Dependencies>
 {
 
-    private static final String SERVICE_NAME = "bloom";
+    static final String SERVICE_NAME = "bloom";
```
klaren: Could still be private right? Or even inlined?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomKernelExtensionFactory.java
```diff
@@ -41,7 +41,7 @@
 public class BloomKernelExtensionFactory extends KernelExtensionFactory<BloomKernelExtensionFactory.Dependencies>
```
klaren, line 43: Other factories have `@Service.Implementation( KernelExtensionFactory.class )`, not sure if we should adopt such black magic though...

klaren (approved):
Two small comments otherwise this looks good!

PR #10105, opened 2017-09-20
Fix store copy of growing files

martinfurmanski:
This could lead to protocol errors where file chunks were
continued to be sent even though the "last" one had already
been sent.

This does not change the protocol as such even though it would
be beneficial to get rid of the dual usage of the encodedLength
field. An encodedLength of less than the maximum size is
interpreted as the last chunk on the receiving end.

This was problematic with growing files which would encode a
length less than the maximum size and yet had a possibility
to send additional chunks for files which had grown.

In particular this became a common problem because the first
chunk was prefetched during the creation of the FileSender
which could be a very long time before it actually was
sent. This could be many minutes or even hours. For the counts
store which has two files that are alternately truncated, it
would be a common case that one file was initially prefetched
while empty but no longer so when it became time to send it.

This would generally be harmless though, because the subsequent
recovery process would repair the state.

The first chunk is now no longer prefetched during the creation
of the FileSender, but rather on the first invocation of
readChunk. The end condition is also handled more cleanly now
and additional data will not be attempted to be read if the end
condition has been met and a last-chunk has been created.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/FileChunk.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/FileChunk.java
@@ -23,9 +23,6 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.Objects;
 
-import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufAllocator;
-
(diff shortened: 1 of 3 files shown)


davidegrohmann (approved):

martinfurmanski:
@davidegrohmann I updated the PR with an additional test and an easier to read implementation for all corner cases.

PR #10106, opened 2017-09-20
Fixes issues w/ concurrent schema changes

tinwelint:
Individual commits tells the story, but generally this is about various issues with concurrent schema changes, such that the loaded representation of the schema may end up in state of permanently missing some updates, and also temporary reading temporarily broken schema.

Some of these issues has existed since the introduction of schema, but the more severe ones was introduced with the global schema lock split.

The `ConcurrentCreateDropIndexIT` test could previously run into all these issues in one or two runs, consistently, so could reproduce very reliably. For that reason I opted for adding only that test for this.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
@@ -172,27 +172,27 @@ public class IndexingService extends LifecycleAdapter implements IndexingUpdateS
     @Override
     public void init()
     {
-        IndexMap indexMap = indexMapRef.indexMapSnapshot();
-
-        Map<InternalIndexState, List<IndexLogRecord>> indexStates = new EnumMap<>( InternalIndexState.class );
-        for ( IndexRule indexRule : indexRules )
+        indexMapRef.modify( indexMap ->
         {
-            IndexProxy indexProxy;
-
-            long indexId = indexRule.getId();
-            IndexDescriptor descriptor = indexRule.getIndexDescriptor();
-            SchemaIndexProvider.Descriptor providerDescriptor = indexRule.getProviderDescriptor();
-            SchemaIndexProvider provider = providerMap.apply( providerDescriptor );
-            InternalIndexState initialState = provider.getInitialState( indexId, descriptor );
-            indexStates.computeIfAbsent( initialState, internalIndexState -> new ArrayList<>() )
-                    .add( new IndexLogRecord( indexId, descriptor ) );
-
-            log.debug( indexStateInfo( "init", indexId, initialState, descriptor ) );
-            switch ( initialState )
+            Map<InternalIndexState, List<IndexLogRecord>> indexStates = new EnumMap<>( InternalIndexState.class );
+            for ( IndexRule indexRule : indexRules )
             {
+                IndexProxy indexProxy;
+
+                long indexId = indexRule.getId();
+                IndexDescriptor descriptor = indexRule.getIndexDescriptor();
+                SchemaIndexProvider.Descriptor providerDescriptor = indexRule.getProviderDescriptor();
+                SchemaIndexProvider provider = providerMap.apply( providerDescriptor );
+                InternalIndexState initialState = provider.getInitialState( indexId, descriptor );
+                indexStates.computeIfAbsent( initialState, internalIndexState -> new ArrayList<>() )
+                .add( new IndexLogRecord( indexId, descriptor ) );
+
+                log.debug( indexStateInfo( "init", indexId, initialState, descriptor ) );
+                switch ( initialState )
+                {
                 case ONLINE:
                     indexProxy =
-                        indexProxyCreator.createOnlineIndexProxy( indexId, descriptor, providerDescriptor );
+                    indexProxyCreator.createOnlineIndexProxy( indexId, descriptor, providerDescriptor );
                     break;
                 case POPULATING:
                     // The database was shut down during population, or a crash has occurred, or some other sad thing.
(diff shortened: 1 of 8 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -437,7 +437,7 @@ private void apply( Iterable<IndexEntryUpdate<LabelSchemaDescriptor>> updates, I
      * it is *vital* that it is stable, and handles errors very well. Failing here means that the entire db
      * will shut down.
      */
-    public void createIndexes( IndexRule... rules ) throws IOException
+    public synchronized void createIndexes( IndexRule... rules ) throws IOException
```
chrisvest: Only the `createIndexes` and the `dropIndex` synchronise on the `IndexingService` monitor lock as far as I can see. This means the state in here is inconsistently synchronised. If we are only worried about concurrent updates to the `indexMapRef`, then we should synchronise on that, and do so consistently. This synchronisation could, for instance, move into the `IndexMapReference` class and be exposed as an atomic `update` method that takes a lambda, or something like that.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
```diff
@@ -55,16 +52,16 @@
...
 
-    private final Map<SchemaDescriptor,IndexDescriptor> indexDescriptors = new HashMap<>();
-    private final PrimitiveIntObjectMap<Set<IndexDescriptor>> indexDescriptorsByLabel = Primitive.intObjectMap();
+    private final Map<SchemaDescriptor,IndexDescriptor> indexDescriptors = new ConcurrentHashMap<>();
+    private final Map<Integer,Set<IndexDescriptor>> indexDescriptorsByLabel = new ConcurrentHashMap<>();
     private final ConstraintSemantics constraintSemantics;
 
     private final Map<Class<?>,Object> dependantState = new ConcurrentHashMap<>();
-    private final PrimitiveIntObjectMap<List<IndexDescriptor>> indexByProperty = Primitive.intObjectMap();
+    private final Map<Integer,List<IndexDescriptor>> indexByProperty = new ConcurrentHashMap<>();
```
chrisvest, line 65: I think we should have a card for looking into an implementation of this based on snapshotting and atomic multi-value updates.

community/kernel/src/test/java/org/neo4j/graphdb/schema/ConcurrentCreateDropIndexIT.java
```diff
@@ -0,0 +1,110 @@
...
+
+public class ConcurrentCreateDropIndexIT
+{
+    private static final String KEY = "key";
+
+    @Rule
+    public final DatabaseRule db = new ImpermanentDatabaseRule();
+
+    @Test
+    public void shouldTest() throws Throwable
```
chrisvest: How about `concurrentCreatingAndDroppingOfIndexesShouldNotInterferre`?

community/kernel/src/test/java/org/neo4j/graphdb/schema/ConcurrentCreateDropIndexIT.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        race.go();
+
+        // THEN they should all be observed as dropped in the end
+        try ( Transaction tx = db.beginTx() )
+        {
+            assertEquals( 0, asList( db.schema().getIndexes() ).size() );
+            tx.success();
+        }
+    }
```
chrisvest, line 116: I think it would be good to also have a middle phase where threads/2 indexes are dropped concurrently with another threads/2 indexes being created, asserting that the index count ends up unchanged.

chrisvest (changes requested):
@tinwelint First round completed. You have a couple of failing tests, and I have a few comments.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
tinwelint: I think so too.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
tinwelint: Maybe. Which are the other usages that should have been synchronized?

community/kernel/src/test/java/org/neo4j/graphdb/schema/ConcurrentCreateDropIndexIT.java
tinwelint: Forgot to rename, it started out as a quick-just-get-this-reproduced thing :)

community/kernel/src/test/java/org/neo4j/graphdb/schema/ConcurrentCreateDropIndexIT.java
tinwelint: Good idea, I can add that

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
chrisvest: It's not that `init` and `start` _need_ to synchronise. Rather, I'd like the synchronisation to be consistent, e.g. by pushing the responsibility down to the `IndexMapReference`, such that there is no way to get it wrong.

tinwelint: Right OK, that's of course never a bad idea. I can have a look at that suggestion about a lambda, I like it.

tinwelint:
@chrisvest something like this?

chrisvest (approved):
Very nice 👍

PR #10109, opened 2017-09-20
Removed grandparent and parent-central POMs

klaren:
This is a large refactoring of the POM structure which removes a lot of cruft that has been accumulated through the years.

 - Most notably the lifecycle now aligns with the regular maven lifecycle.
 - The -DrunITs is removed and integration tests are now opt-out instead of opt-in.
 - Duplicate declarations in child POMs have moved up to the parent POMs instead.
 - All eclipse m2e declarations are removed.
 - Old obsolete properties no longer used are removed.

Javadoc generation and packaging builds are most probably broken. But to keep this change smaller I opted to merge this now and fix that later. The release of 3.4 is quite far away so there should be time to fix this.

This builds, which is more than expected :)


community/neo4j/pom.xml
```diff
@@ -228,7 +226,7 @@ the relevant Commercial Agreement.
           </execution>
           <execution>
             <id>create-javadocs</id>
-            <phase>${attach-javadoc-phase}</phase>
+            <phase>none</phase> <!-- TODO: fix -->
```
MishaDemianenko, line 229: TODO

community/bolt/pom.xml
```diff
@@ -163,11 +163,11 @@
...
-      <dependency>
-          <groupId>commons-lang</groupId>
-          <artifactId>commons-lang</artifactId>
-          <version>RELEASE</version>
-          <scope>test</scope>
-      </dependency>
+    <dependency>
+      <groupId>commons-lang</groupId>
+      <artifactId>commons-lang</artifactId>
+      <version>RELEASE</version>
```
MishaDemianenko, line 169: i think while we here we should change it to particular version number or even better would be to switch it to commons-lang3 which other modules use

enterprise/cluster/pom.xml
```diff
@@ -16,10 +15,8 @@
   <url>http://components.neo4j.org/${project.artifactId}/${project.version}</url>
 
   <properties>
```
MishaDemianenko, line 17: its a time to fix really nice copy pasted short name of this module

enterprise/cypher/cypher/pom.xml
```diff
@@ -45,25 +45,12 @@
 
   <properties>
     <version-package>cypher.internal</version-package>
-    <scala.version>2.11.11</scala.version>
-    <scala.binary.version>2.11</scala.binary.version>
     <maven.javadoc.failOnError>false</maven.javadoc.failOnError>
```
MishaDemianenko, line 48: do we still need this one?

community/cypher/cypher/README.txt
```diff
@@ -33,5 +33,5 @@ To get the Cypher project working in Eclipse (with the above plugins installed):
 * Use File -> Import... -> Existing Maven Projects to import the project.
 * After the import has finished, right-click on the project and
   -> Maven -> Update Project Configuration...
-* Make sure to set the Scala version to 2.10.* for the projects.
+* Make sure to set the Scala version to 2.11.* for the projects.
```
MishaDemianenko, line 36: would be nice if we can kill this file at all, its obsolete anyway and describing super old versions

enterprise/neo4j-enterprise/pom.xml
```diff
@@ -269,7 +267,7 @@
           </execution>
           <execution>
             <id>create-javadocs</id>
-            <phase>${attach-javadoc-phase}</phase>
+            <phase>none</phase> <!-- TODO: fix -->
```
MishaDemianenko, line 270: one more here

pom.xml
```diff
@@ -523,6 +713,18 @@
         <artifactId>httpcore</artifactId>
         <version>4.4.5</version>
       </dependency>
+      <dependency>
+        <groupId>com.google.jimfs</groupId>
```
MishaDemianenko: we do not need it anymore

MishaDemianenko (changes requested):
Couple of comments otherwise great step into a clean poms future)

community/bolt/pom.xml
klaren: Will create a card for this to cleanup common-lang dependencies

MishaDemianenko (approved):

community/cypher/cypher/README.txt
Mats-SX: @craigtaverner @systay @fickludd fyi

enterprise/cypher/pom.xml
```diff
@@ -51,248 +48,16 @@
...
-                      <goal>enforce</goal>
-                  </goals>
-                  <configuration>
-                      <rules>
-                          <requireFilesDontExist>
-                              <files>
-                                  <file>cypher-compiler-3.2</file>
-                                  <file>cypher-compiler-3.1</file>
-                                  <file>cypher-compiler-3.0</file>
-                                  <file>cypher-compiler-2.3</file>
```
Mats-SX, line 122: Why is this removed? It can happen if you switch branches or fail at forward-merging, and has been known to prevent bad check-ins.

pom.xml
```diff
@@ -2,26 +2,75 @@
...
   <artifactId>parent</artifactId>
   <version>3.4.0-SNAPSHOT</version>
-
   <packaging>pom</packaging>
   <name>Neo4j</name>
   <description>Neo4j Graph Database</description>
+  <inceptionYear>2002</inceptionYear>
+
+  <organization>
+    <name>The Neo4j Graph Database Project</name>
```
Mats-SX, line 15: Wow, really? I guess this was copied from a parent, but we may want to change this @magnusvejlstrup

enterprise/cypher/pom.xml
klaren: It's already defined in the parent POM and is inherited. All cypher modules inherits from  `community/cypher/pom.xml`, even enterprise.

pom.xml
klaren: It's only used when generating copyright information in our javadoc since we don't use the site maven target anymore, i.e. `Copyright © 2002–2017 The Neo4j Graph Database Project. All rights reserved.`. Should maybe be changed to Neo4j Inc. now?

PR #10110, opened 2017-09-20
Ignore election performance IT

martinfurmanski:
Was already ignored for the rapid conditions and are now ignored
for the normal conditions as well, since running builds in parallel
has a significant effect on the outcome.

The intention is to move this to a proper benchmarking environment
eventually. The tests are left in place so that one can run them
manually until then.

The test still exists unignored on older branches where parallel
testing is not utilized.


digitalstain (approved):
There is value in this being run manually so we shouldn't remove it yet. It should be taken care of though when a proper benchmarking framework for CC is in place.

PR #10045, opened 2017-09-12
Update cost model constants for 3.3.

Lojjs:
Eagerness multiplier kept from 3.2.

PR #10111, opened 2017-09-20
Revert "Merge pull request #10045 from Lojjs/3.3-cost-model"

fickludd:
This reverts a cost model update, because it introduced a regression in LDBC Query 12. The regression was caused by less optimal planning, due to the changed relative costs of some operators.


fickludd:
Both test failures seem to be flaky tests.

PR #10118, opened 2017-09-20
Avoid stack overflow on IBM JDK in case of panic during import

MishaDemianenko:
Since IBM JDK 8 does not check for circular references in exception
current panic mechanic in import tool can cause stack overflow
when printing exception stack trace.
Current PR change panic handling to not wrap original
source of panic into new RuntimeException and will avoid setting
panic exception being suppressed by panic exception.


chrisvest (approved):

chrisvest:
I wish there was a way to entirely eliminate this class of bugs, but I can't think of any 😕

PR #10119, opened 2017-09-21
Report recovery progress

MishaDemianenko:
Allow recovery to report ongoing progress, so users can have indication
of how much work is already done and how much left for recovery to restore
database and make it available for new transactions again.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
@@ -143,7 +143,9 @@ public class Recovery extends LifecycleAdapter
     {
         long numberOfTransactionToRecover =
                 getNumberOfTransactionToRecover( recoveryStartInformation, lastReversedTransaction );
-        progressReporter.start( numberOfTransactionToRecover << 1 );
+        // since we will process each transaction twice (doing reverse and direct detour) we need to
+        // multiply number of transactions that we want to recover by 2 to be able to report correct progress
+        progressReporter.start( numberOfTransactionToRecover * 2 );
     }
 
     private void reportProgress()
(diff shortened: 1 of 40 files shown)


MishaDemianenko:
Please note that it's based on #9956 and will be updated later to target 3.4

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
```diff
@@ -117,14 +126,35 @@ public void init() throws Throwable
...
         startupStatistics.setNumberOfRecoveredTransactions( numberOfRecoveredTransactions );
         monitor.recoveryCompleted( numberOfRecoveredTransactions );
     }
+
+    private void initProgressReporter( RecoveryStartInformation recoveryStartInformation,
+            CommittedTransactionRepresentation lastReversedTransaction )
+    {
+        long numberOfTransactionToRecover =
+                getNumberOfTransactionToRecover( recoveryStartInformation, lastReversedTransaction );
+        progressReporter.start( numberOfTransactionToRecover << 1 );
```
tinwelint: This is due to both reverse and forward recovery, right? Please comment that here and also the `<< 1` is unnecessary, can might as well do `* 2` for better readability.

tinwelint (commented):
It looks like you've mostly changed class names and that this progress logging already existed somehow. Why wasn't whatever was there enough?

community/kernel/src/main/java/org/neo4j/kernel/recovery/Recovery.java
MishaDemianenko: sure, I can do * 2 if you think it's more readable, correct since we will process each transaction twice we need to double expected number of transaction to process.

MishaDemianenko:
@tinwelint we already had progress reporter for migration. It was called "Section" before since it was a subpart of the migration process. I extracted that away from migration so now we can report progress of any long running things.
Before we only have messages "Recovery started", "Recovery completed" without any indication of a progress. And that can take a while to recover a database

MishaDemianenko:
@tinwelint updated

tinwelint:
Thanks, although my main point was that this *2 should also get a comment there in the code as to help anyone reading that code later.

MishaDemianenko:
@tinwelint updated

tinwelint (approved):

PR #10120, opened 2017-09-21
Removes optimization to not checkpoint GBPTree on no changes

tinwelint:
This solves an issue which would be observed as a tree looking like non-clean
even after a clean db shutdown. The main reason for this is that skipping
the checkpoint would not await the background clean job and so closing the tree
could potentially be done before the clean job was done. This would prevent
the tree from being marked as clean on close.

Read-only tools like consistency checker asserts that no recovery is required
before starting and so would fail on a db containing such a tree.

### DIFF
--- a/community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
+++ b/community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
@@ -855,11 +855,34 @@ public class GBPTreeTest
         CleanJobControlledMonitor monitor = new CleanJobControlledMonitor();
         try ( GBPTree<MutableLong,MutableLong> index = index().with( monitor ).with( cleanupWork ).build() )
         {
-            // WHEN
-            // Cleanup not finished
+            // WHEN cleanup not finished
             Future<?> cleanup = executor.submit( throwing( cleanupWork::start ) );
             monitor.barrier.awaitUninterruptibly();
             index.writer().close();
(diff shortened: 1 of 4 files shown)


community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
```diff
@@ -1231,6 +1231,27 @@ public void shouldNotCheckpointOnCloseIfNoChangesHappened() throws Exception
...
+        CheckpointCounter checkpointCounter = new CheckpointCounter();
+
+        // WHEN
+        try ( GBPTree<MutableLong,MutableLong> index = index().with( checkpointCounter ).build() )
+        {
+            checkpointCounter.reset();
+            for ( int i = 0; i < 2; i++ )
+            {
+
+            }
```
burqen: Empty loop

burqen (changes requested):
Two `DumpGBPTree` seems a little excessive. Keep one, dump the other ;)

There is a test called `GBPTreeTest.cleanJobShouldLockOutCheckpoint`.
We should add test `GBPTreeTest.cleanJobShouldLockOutCheckpointOnNoUpdate`. Just copy test and remove `index.writer().close();`

tinwelint:
There... amended the removal of the duplicate dumper into the first commit. Fixed the other things in a new commit.

burqen (approved):

PR #10123, opened 2017-09-21
Increase election timeout for id reuse IT

martinfurmanski:
This is not a perfect solution. The test probably needs to be
rewritten and deep reacher into the stack for some assumptions
on when the assertion can be trusted.


digitalstain (approved):
This requires a rewrite to make the assertion more robust. However, that requires substantial effort and we want the coverage this test provides for now.

PR #10125, opened 2017-09-21
Make sure that if the database fails to start for some reason…

chrisvest:
…then the fulltext index add-on will not prevent shut down

### DIFF
--- a/enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomIT.java
+++ b/enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomIT.java
@@ -19,6 +19,7 @@
  */
 package org.neo4j.kernel.api.impl.fulltext.integrations.bloom;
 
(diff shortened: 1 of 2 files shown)


enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomIT.java
```diff
@@ -341,6 +343,41 @@ public void shouldBeAbleToRunConsistencyCheck() throws Exception
...
+        }
+        catch ( Exception e )
+        {
+            // Good
+        }
+        finally
+        {
+            if ( !dir.setReadable( true ) )
+            {
+                System.err.println( "!!! Failed to make " + dir + " writable again !!!" );
```
MishaDemianenko: Oo

chrisvest: I didn't make it an assertion in case it would shadow another throwable from the try-clause.

ragadeeshu (approved):

PR #10128, opened 2017-09-21
Shade ASM library to avoid version conflicts in embedded deployments

chrisvest:


fickludd (approved):
LGTM. I didn't double check that this does indeed hide `org.objectweb.asm`, but I guess you did :).

fickludd:
Tests are flaky, cardified fixing those, and muted in TC. Will merge this on freeze manager approval.

PR #10129, opened 2017-09-21
Allow most modules to be compiled with java 9

MishaDemianenko:
Compilations is not possible so far for
procedure-compiler modules because of dependencies that rely on tools.jar
and all scala modules. All other modules are compilable.


MishaDemianenko:
Should be updates after #10109 will be merged to simplify poms updates.

lutovich (commented):
LGTM!
Newly required generics are sad :(

PR #10135, opened 2017-09-22
Make topology address exception checked

martinfurmanski:
Many code paths treat unexpected exceptions as fatal.


andrewkerr9000 (approved):
LGTM

PR #10138, opened 2017-09-25
Fix pom version-package

pontusmelke:
Not sure what this does but I'm pretty sure it was wrong.


craigtaverner:
Having discussed this with Anton, he is refactoring the poms in a way that will remove this problem entirely. These lines are part of OSGi support which has not been enabled since 1.8, so that is why we saw no symptoms from this being wrong. His suggestion is close this PR, and let his refactoring deal with the cleanup.

PR #10140, opened 2017-09-25
Additional procedures for bloom fulltext

ragadeeshu:


enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomProcedures.java
```diff
@@ -55,41 +59,89 @@ public void awaitPopulation() throws IOException
         provider.awaitPopulation();
     }
 
+    @Description( "Returns the property keys indexed by the bloom addon" )
+    @Procedure( name = "db.fulltext.bloomFulltextProperties", mode = READ )
+    public Stream<PropertyOutput> bloomFulltextProperties() throws Exception
+    {
+        return provider.getProperties(BLOOM_NODES, NODES).stream().map( PropertyOutput::new );
```
chrisvest: Checkstyle spaces.

ragadeeshu: Fixed

ragadeeshu:
@chrisvest updated

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
```diff
@@ -220,6 +222,28 @@ public ReadOnlyFulltext getReader( String identifier, FulltextIndexType type ) t
         }
     }
 
+    public Set<String> getProperties( String identifier, FulltextIndexType type )
+    {
+        return applyToMatchingIndex( identifier, type, LuceneFulltext::getProperties );
+    }
+
+    private <E> E applyToMatchingIndex( String identifier, FulltextIndexType type, Function<LuceneFulltext,E> getProperties )
```
chrisvest: This `getProperties` function should have a more generic name, I think.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/LuceneFulltext.java
```diff
@@ -34,13 +34,15 @@
...
 import org.neo4j.kernel.api.impl.schema.writer.PartitionedIndexWriter;
+import org.neo4j.kernel.api.index.InternalIndexState;
 
 class LuceneFulltext extends AbstractLuceneIndex
 {
     private final Analyzer analyzer;
     private final String identifier;
     private final FulltextProvider.FulltextIndexType type;
     private final Set<String> properties;
+    private InternalIndexState state;
```
chrisvest: Perhaps this needs to be volatile, just to be safe?

chrisvest (commented):
I had a couple of comments. If you just fix those up, it's good to go in, I think. 👍

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
ragadeeshu: whoops

chrisvest (approved):

jexp:
Why do they have fulltext twice in the name? and bloom?
Perhaps align with the Cypher team on naming like they did for the explicit indexes, so it is consistent.
/cc @craigtaverner 

db.fulltext.bloomFulltextStatus -> db.fulltext.status

chrisvest:
@jexp It's not meant to be a user-facing feature. It's only for the bloom project, so we don't to contaminate the namespace if we want to do a fully featured, general purpose fulltext index later.

PR #10144, opened 2017-09-25
Clean up arguments in StandardInternalExecutionResult for testing

sherfert:
The method toEagerResultForTestingOnly used to add runtimes and
planner arguments always, even when they already existed. This PR
changes that to only add runtime, if it is not defined yet. The value
for planner was hardcoded to COST, this was entirely removed.

Even though this PR touches a file in main, it should influence test only.

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/executionplan/StandardInternalExecutionResult.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/executionplan/StandardInternalExecutionResult.scala
@@ -109,7 +109,7 @@ abstract class StandardInternalExecutionResult(context: QueryContext, runtime: R
      * NOTE: This should ony be used for testing, it creates an InternalExecutionResult
      * where you can call both toList and dumpToString
      */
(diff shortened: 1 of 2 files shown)


OliviaYtterbrink (approved):

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/executionplan/StandardInternalExecutionResult.scala
```diff
@@ -122,12 +122,16 @@ abstract class StandardInternalExecutionResult(context: QueryContext, runtime: R
 
       override protected def createInner: util.Iterator[util.Map[String, Any]] = result.iterator()
 
-      override def executionPlanDescription(): InternalPlanDescription =
-        self.executionPlanDescription()
-          .addArgument(Planner(planner.toTextOutput))
-          .addArgument(PlannerImpl(planner.name))
```
henriknyman, line 128: What happens with the Planner argument? Is that set somewhere else? I think benchmarking relies on it being set.

OliviaYtterbrink: This change only applies to internal testing through `RewindableExecutionResult`.

PR #10148, opened 2017-09-26
Add more options for querying the bloom addon

ragadeeshu:


enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomProcedures.java
```diff
@@ -57,27 +57,37 @@ public void awaitPopulation() throws IOException
...
-            return queryAsStream( terms, indexReader );
+            return queryAsStream( terms, indexReader, fuzzy, matchAll );
         }
     }
 
     @Description( "Queries the bloom index for relationships" )
     @Procedure( name = "db.fulltext.bloomFulltextRelationships", mode = READ )
-    public Stream<Output> bloomFulltextRelationships( @Name( "terms" ) List<String> terms ) throws Exception
+    public Stream<Output> bloomFulltextRelationships( @Name( "terms" ) List<String> terms, @Name( "fuzzy" ) boolean fuzzy,
+            @Name( "Require that all terms are matched" ) boolean matchAll ) throws Exception
```
chrisvest: You could set `defaultValue`s for the `fuzzy` and `matchAll` parameters. Just a thought.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomProcedures.java
```diff
@@ -57,27 +57,37 @@ public void awaitPopulation() throws IOException
...
     }
 
-    private Stream<Output> queryAsStream( List<String> terms, ReadOnlyFulltext indexReader )
+    private Stream<Output> queryAsStream( List<String> terms, ReadOnlyFulltext indexReader, boolean fuzzy, boolean matchAll )
     {
-        PrimitiveLongIterator primitiveLongIterator = indexReader.fuzzyQuery( terms.toArray( new String[0] ) );
+        PrimitiveLongIterator primitiveLongIterator;
+        if ( fuzzy )
+        {
+            primitiveLongIterator = indexReader.fuzzyQuery( matchAll, terms.toArray( new String[0] ) );
```
chrisvest: This `terms.toArray( new String[0] )` is looking a bit silly. Especially when all we do with the array further down, is turning it into a `Stream`. In a future PR we could just pass the list through.

chrisvest (approved):
Couple of comments, but it looks good.

PR #10149, opened 2017-09-26
Make the ForsetiFalseDeadlockTest more robust

chrisvest:
Previously the test would fail on any false deadlock that popped out.
Now, when a false deadlock is detected, the test instead does a more thorough
check, to make sure that the false deadlock failure rate is 2% or less.

The 2% was chosen because that roughly matches the flaky failure rate we have
been observing on our build systems so far.

### DIFF
--- a/enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiFalseDeadlockTest.java
+++ b/enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiFalseDeadlockTest.java
@@ -293,6 +293,33 @@ public class ForsetiFalseDeadlockTest
 
     private void loopRunTest( int testRuns ) throws InterruptedException, java.util.concurrent.ExecutionException
     {
+        List<Throwable> exceptionList = new ArrayList<>();
+        loopRun( testRuns, exceptionList );
+
+        if ( !exceptionList.isEmpty() )
+        {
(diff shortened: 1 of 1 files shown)


enterprise/kernel/src/test/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiFalseDeadlockTest.java
```diff
@@ -293,6 +293,33 @@ public void testMildlyForFalseDeadlocks() throws Exception
...
+
+        if ( !exceptionList.isEmpty() )
+        {
+            // We saw exceptions. Run it 99 more times, and then verify that our false deadlock rate is less than 2%.
+            int additionalRuns = testRuns * 99;
+            loopRun( additionalRuns, exceptionList );
+            double totalRuns = additionalRuns + testRuns;
+            double failures = exceptionList.size();
+            double failureRate = failures / totalRuns;
+            if ( failureRate > 0.2 )
```
MishaDemianenko: should it be 0.02 instead? failure reporting threshold is more than 20%?

chrisvest: Oh shoot, you're right!

chrisvest:
@MishaDemianenko I made the 1-byte change. I guess we can merge on green?

MishaDemianenko:
@chrisvest correct

MishaDemianenko (approved):

PR #10150, opened 2017-09-26
Fixed performance regression in calling procedures

craigtaverner:


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/executionplan/procs/ProcedureExecutionResult.scala
```diff
@@ -95,9 +95,8 @@ class ProcedureExecutionResult(context: QueryContext,
   override def accept[EX <: Exception](visitor: QueryResultVisitor[EX]): Unit = {
     executionResults.foreach { res =>
       val fieldArray = new Array[AnyValue](indexResultNameMappings.size)
-      var i = 0
-      for ((pos, _, typ) <- indexResultNameMappings) {
-        fieldArray(i) = typ match {
+      for (pos <- indexResultNameMappings.indices) {
+        fieldArray(pos) = indexResultNameMappings(pos)._3 match {
```
pontusmelke: is this a bottleneck? If not I think the old way looks nicer, not a fan of `._3`

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/executionplan/StandardInternalExecutionResult.scala
```diff
@@ -98,9 +99,13 @@ abstract class StandardInternalExecutionResult(context: QueryContext, runtime: R
   override def accept[E <: Exception](visitor: ResultVisitor[E]): Unit = {
     accept(new QueryResultVisitor[E] {
       override def visit(record: QueryResult.Record): Boolean = {
-        val row = new MapBasedRow
-        row.map = fieldNames().zip(record.fields().map(context.asObject)).toMap
-        visitor.visit(row)
+        val names = fieldNames()
```
pontusmelke: Probably not super important but `val names = fieldNames() ` can be done outside `visit`

pontusmelke (approved):
Nice, had some comments that are not terribly important.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/executionplan/procs/ProcedureExecutionResult.scala
craigtaverner: Amazingly the profiling showed this to be 2x faster for this part of the profile. It turned out that the call stack created by scala for the tuple foreach was much, much deeper than the one for the int range foreach. No idea why, but the performance change was very real.

So we did not do this whimsically.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/executionplan/StandardInternalExecutionResult.scala
craigtaverner: And I wonder if the fact that fieldNames is now a `val` would get this optimised by the compiler the same in all cases?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/executionplan/StandardInternalExecutionResult.scala
```diff
@@ -97,10 +97,14 @@ abstract class StandardInternalExecutionResult(context: QueryContext, runtime: R
 
   override def accept[E <: Exception](visitor: ResultVisitor[E]): Unit = {
     accept(new QueryResultVisitor[E] {
+      val names = fieldNames()
       override def visit(record: QueryResult.Record): Boolean = {
-        val row = new MapBasedRow
-        row.map = fieldNames().zip(record.fields().map(context.asObject)).toMap
-        visitor.visit(row)
+        val fields = record.fields()
+        val mapData = new mutable.AnyRefMap[String, Any](names.length)
```
pontusmelke, line 103: We do the same thing in `PipeExecutionResult`, it should be just as inefficient there. Can we apply the same fix there?

craigtaverner: Good point. I've added a new commit for this. I also profiled it with `MATCH (n) RETURN n` and in that case the previous code used 40% of the result generation time to convert the results to external types with this method, while after the commit the code uses 20% of the time. So, as before, we double the performance of this specific algorithm.

PR #10152, opened 2017-09-27
Fix bug in planning PruningVarExpand

fickludd:
Distinct is planned as an operator now and is recognized properly by the pruningVarExpand rewriter.

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/VarLengthExpandQueryPlanAcceptanceTest.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/VarLengthExpandQueryPlanAcceptanceTest.scala
@@ -42,7 +42,6 @@ class VarLengthExpandQueryPlanAcceptanceTest extends ExecutionEngineFunSuite wit
   }
 
   test("Plan pruning var expand on distinct var-length match") {
-    setUp("To")
     val query = "MATCH (a)-[*1..2]->(c) RETURN DISTINCT c"
     val result = executeWithCostPlannerAndInterpretedRuntimeOnly(query)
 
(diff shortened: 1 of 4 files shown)


systay:
Nice find. Thanks for fixing my breakage... 🥇 

If I were to wish for one more thing, it would be to add at least one test to `PruningVarExpanderTest`

fickludd:
@systay The blame should be on both of us for never adding these integration tests in the first place, but now it should be all good :).

Updated the `PruningVarExpanderTest` to actually test using relevant input plans, good call.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/VarLengthExpandQueryPlanAcceptanceTest.scala
```diff
@@ -28,56 +28,101 @@ class VarLengthExpandQueryPlanAcceptanceTest extends ExecutionEngineFunSuite wit
...
-        || |                     +----------------+------+---------+-----------------+-------------------+----------------------+------------------+----------------------------------+
-        || +Filter               |              0 |    1 |       1 |               0 |                 0 |               0.0000 | e                | e.name = {  AUTOSTRING1}         |
-        || |                     +----------------+------+---------+-----------------+-------------------+----------------------+------------------+----------------------------------+
-        || +NodeByLabelScan      |              1 |    1 |       2 |               0 |                 0 |               0.0000 | e                | :To                              |
-        |+-----------------------+----------------+------+---------+-----------------+-------------------+----------------------+------------------+----------------------------------+
-        |""".stripMargin)
+    result.executionPlanDescription() should useOperatorWithText("VarLengthExpand(All)", "(e)<-[:*..4]-(a)")
+    result.executionPlanDescription() should useOperatorWithText("NodeByLabelScan", ":To")
   }
```
sherfert, line 34: Since we're porting everything to CypherComparisonSupport, does it make sense to write these new tests using that framework from the beginning?

systay: I think we should move the tests over, but at the same time it would be nice to disconnect the two work streams

PR #10157, opened 2017-09-27
Log about only successful termination of transactions outside safe zone

MishaDemianenko:


burqen (approved):
How did we not do this in the first place? Nice catch! 🎉

PR #10160, opened 2017-09-27
Fix comparison bug in slotted runtime

henriknyman:
SlottedRewriter should apply PrimitiveEquals only on variables allocated
to LongSlots.

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MiscAcceptanceTest.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MiscAcceptanceTest.scala
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.internal.cypher.acceptance
+
+import org.neo4j.cypher.ExecutionEngineFunSuite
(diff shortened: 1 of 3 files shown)


enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/SlottedRewriter.scala
```diff
@@ -146,8 +146,10 @@ class SlottedRewriter(tokenContext: TokenContext) {
       case e@Equals(Variable(k1), Variable(k2)) => // TODO: Handle nullability
         val slot1 = pipelineInformation(k1)
         val slot2 = pipelineInformation(k2)
-        if (slot1.typ == slot2.typ)
+        if (slot1.typ == slot2.typ && PipelineInformation.isLongSlot(slot1)) {
```
pontusmelke, line 149: Shouldn't we check both slots here and not just have an assertion, can't we get a Reference Node and a Slot Node here? Or maybe we are guaranteed to not have any reference node left here?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MiscAcceptanceTest.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+package org.neo4j.internal.cypher.acceptance
+
+import org.neo4j.cypher.ExecutionEngineFunSuite
+import org.neo4j.internal.cypher.acceptance.CypherComparisonSupport._
+
+class MiscAcceptanceTest extends ExecutionEngineFunSuite with CypherComparisonSupport {
+
+  // This test verifies a bugfix in slotted runtime
+  test("should be able to compare integers") {
```
pontusmelke, line 28: Could maybe be in the acceptance suite?

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/SlottedRewriter.scala
henriknyman: If they have the same type I assumed they should have the same slot type or something is wrong with slot allocation.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MiscAcceptanceTest.scala
henriknyman: There is a test that fails on this already in compatibility spec suite but it uses path expressions so it cannot be run on slotted in 3.3.

pontusmelke: I meant the acceptance suite, I thought we prefer to add new tests there as feature files instead of using scala tests

PR #10164, opened 2017-09-28
Support RollUpApply and PathExpression in slotted runtime

henriknyman:


enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/RollUpApplySlottedPipe.scala
```diff
@@ -0,0 +1,84 @@
...
+  extends PipeWithSource(lhs) {
+
+  private val collectionSlot = pipelineInformation.get(collectionName).get
+
+  private val getValueToCollectFunction = {
+    val expression: Expression = identifierToCollect._2
+    (state: QueryState) => (ctx: ExecutionContext) => expression(ctx, state)
+  }
+
+  private val hasNullValuePredicates: Seq[(ExecutionContext) => Boolean] =
```
systay, line 42: I think I would prefer this to be done outside this file and have this file focus on the runtime aspects. When reading this code I want to be in a runtime frame of mind, and here we are mixing between the two.

Pretty sure this is highly subjective, so feel free to ignore me. This is just my humble opinion.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/RollUpApplySlottedPipe.scala
```diff
@@ -0,0 +1,84 @@
...
+    nullableIdentifiers.toSeq.map { elem =>
+      val elemSlot = pipelineInformation.get(elem)
+      elemSlot match {
+        case Some(LongSlot(offset, true, _, _)) => { (ctx: ExecutionContext) => ctx.getLongAt(offset) == -1 }
```
systay: We know which of these slots are nullable. No need to check for null if it is not nullable.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/RollUpApplySlottedPipe.scala
```diff
@@ -0,0 +1,84 @@
...
+      case RefSlot(offset, _, _, _) =>
+        (ctx: ExecutionContext, value: AnyValue) => ctx.setRefAt(offset, value)
+      case _ => throw new InternalError("Expected collection to be allocated to a ref slot")
+    }
+  }
+
+  override protected def internalCreateResults(input: Iterator[ExecutionContext], state: QueryState) = {
+    input.map {
+      ctx =>
+        val outputRow = PrimitiveExecutionContext(pipelineInformation)
```
systay: No need to create a new row and copy the incoming data into it. We should just write to the incoming `ctx`.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/RollUpApplySlottedPipe.scala
```diff
@@ -0,0 +1,84 @@
...
+        ctx.copyTo(outputRow)
+        if (hasNullValue(ctx)) {
+          setCollectionInRow(outputRow, NO_VALUE)
+        }
+        else {
+          val innerState = state.withInitialContext(ctx)
+          val innerResults: Iterator[ExecutionContext] = rhs.createResults(innerState)
+          val collection = VirtualValues.list(innerResults.map(getValueToCollectFunction(state)).toArray: _*)
+          setCollectionInRow(outputRow, collection)
```
systay: Instead of using a method value, I think we should calculate the offset to an int at constructor time and just `outputRow.setRefAt` here. This is in the tight loop, so minimizing virtual calls is a good idea.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/SlottedPipeBuilder.scala
```diff
@@ -274,28 +274,28 @@ class SlottedPipeBuilder(fallback: PipeBuilder,
...
-          case _ =>
-            throw new InternalException(s"Did not find `$k` in the pipeline information")
-        }
-    }
+    val runtimeColumns: Seq[(String, commandExpressions.Expression)] =
+      columns.map(createProjectionForIdentifier(pipelineInformation1))
     runtimeColumns
   }
 
+  private def createProjectionForIdentifier(pipelineInformation1: PipelineInformation)(identifier: String) = {
```
systay, line 283: How about just `slotInformation`? That `1` itches a little... ;)

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -271,12 +271,17 @@ object SlotAllocation {
...
+        val newPipeline = lhsPipeline.seedClone()
         rhsPipeline.foreachSlot {
           case (k, slot) =>
-            cartesianProductPipeline.add(k, slot)
+            newPipeline.add(k, slot)
         }
-        cartesianProductPipeline
+        newPipeline
+
+      case RollUpApply(_, _, collectionName, _, _) =>
```
systay, line 282: Could we beef up `SlotAllocationTest` to match? We've been pretty good with the test coverage there, would be nice to keep it up.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -271,12 +271,17 @@ object SlotAllocation {
...
+        val newPipeline = lhsPipeline.seedClone()
```
systay: I don't think we need to do a `seedClone()` here. Pretty sure we should just add one reference slot to the `lhs` pipeline. For each incoming row, we're producing exactly one row, and the only difference between the incoming and outgoing row is that the outgoing one has one additional refslot, nothing else.

enterprise/cypher/slotted-runtime/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/RollUpApplyPipeTest.scala
```diff
@@ -0,0 +1,185 @@
...
+import org.neo4j.cypher.internal.frontend.v3_4.symbols._
+import org.neo4j.cypher.internal.frontend.v3_4.test_helpers.CypherFunSuite
+import org.neo4j.cypher.internal.spi.v3_4.{Operations, QueryContext}
+import org.neo4j.graphdb.Node
+import org.neo4j.helpers.ValueUtils
+import org.neo4j.values.storable.Values
+import org.neo4j.values.storable.Values.NO_VALUE
+import org.neo4j.values.virtual.VirtualValues
+
+class RollUpApplySlottedPipeTest extends CypherFunSuite with PipeTestSupport with FakeEntityTestSupport {
```
systay, line 40: Nice test coverage! 👍

systay (changes requested):
Nice work. I think we are there functionality-wise. My comments are mostly performance improvements.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
fickludd: yes, l think `lhs` should continue as well.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/expressions/SlottedExpressionConverters.scala
```diff
@@ -61,7 +63,44 @@ object SlottedExpressionConverters extends ExpressionConverter {
...
+
+      case MultiRelationshipPathStep(relExpression, SemanticDirection.BOTH, next) =>
+        multiUndirectedRelationshipProjector(toCommandExpression(relExpression, self).get, project(next))
+
+      case NilPathStep =>
+        nilProjector
+    }
+
+    val projector = project(e.step)
+//    val dependencies = e.step.dependencies.map(_.asInstanceOf[Variable].name) // TODO: Verify that we never need this after slot rewriting
```
fickludd: There a TODO here. Is that acted on, or cardified?

fickludd (changes requested):
From what I understand, this looks good in general. As andres point out, we could avoid copying rows from the left though, which would be nice to fix.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/SlottedPipeBuilder.scala
henriknyman: It was already there, I just extracted a method :) Renaming it something different just because there is an outer instance with the same type seems equally weird IMHO.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
henriknyman: Yes

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/RollUpApplySlottedPipe.scala
henriknyman: I don't really have a strong opinion, but I chose to humbly ignore yours :) Currently there is a mix of two styles across slotted pipe implementations, so maybe we could agree on one style, document it and unify the code as a separate concern.

fickludd (approved):

PR #10165, opened 2017-09-28
Fix bug in detach delete, which could cause deadlocks

fickludd:
This bug was caused by TwoPhaseNodeForRelationshipLocking not
guaranteeing to acquire locks in ascending node id order. This was
tested, but the 3 tested nodes coincidentally were provided in the
correct order by that iterator, causing the test to pass.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
@@ -111,10 +111,7 @@ class TwoPhaseNodeForRelationshipLocking
 
     private void lockAllNodes( KernelStatement state, long[] nodeIds )
     {
-        for ( long nodeId : nodeIds )
-        {
-            state.locks().optimistic().acquireExclusive( state.lockTracer(), ResourceTypes.NODE, nodeId );
-        }
+        state.locks().optimistic().acquireExclusive( state.lockTracer(), ResourceTypes.NODE, nodeIds );
     }
 
     private void unlockAllNodes( KernelStatement state, long[] nodeIds  )
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
@@ -19,11 +19,14 @@
  */
 package org.neo4j.kernel.impl.api;
 
+import java.util.Arrays;
+
 import org.neo4j.collection.primitive.Primitive;
-import org.neo4j.collection.primitive.PrimitiveLongIterator;
+import org.neo4j.collection.primitive.PrimitiveArrays;
 import org.neo4j.collection.primitive.PrimitiveLongSet;
 import org.neo4j.cursor.Cursor;
 import org.neo4j.function.ThrowingConsumer;
+import org.neo4j.kernel.api.exceptions.EntityNotFoundException;
 import org.neo4j.kernel.api.exceptions.KernelException;
 import org.neo4j.kernel.impl.api.operations.EntityReadOperations;
 import org.neo4j.kernel.impl.locking.ResourceTypes;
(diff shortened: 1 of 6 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
```diff
@@ -82,24 +81,48 @@ void lockAllNodesAndConsumeRelationships( long nodeId, final KernelStatement sta
         while ( retry );
     }
 
-    private void lockAllNodes( KernelStatement state )
+    private void collectAndSortNodeIds( long nodeId, KernelStatement state ) throws EntityNotFoundException
```
tinwelint, line 84: Is it common to do detach delete for multiple nodes in one transaction? I'm thinking about the added primitive long set and array created for each such request and potential reuse of those data structures

community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
```diff
@@ -82,24 +81,48 @@ void lockAllNodesAndConsumeRelationships( long nodeId, final KernelStatement sta
...
 
-    private void unlockAllNodes( KernelStatement state )
+    private void unlockAllNodes( KernelStatement state, long[] nodeIds  )
     {
-        PrimitiveLongIterator iterator = nodeIds.iterator();
-        while ( iterator.hasNext() )
+        for ( long nodeId : nodeIds )
         {
-            state.locks().optimistic().releaseExclusive( ResourceTypes.NODE, iterator.next() );
+            state.locks().optimistic().releaseExclusive( ResourceTypes.NODE, nodeId );
```
tinwelint, line 121: You could use one call since the id argument is a vararg

community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
```diff
@@ -82,24 +81,48 @@ void lockAllNodesAndConsumeRelationships( long nodeId, final KernelStatement sta
...
+
+    private void lockAllNodes( KernelStatement state, long[] nodeIds )
     {
-        PrimitiveLongIterator nodeIdIterator = nodeIds.iterator();
-        while ( nodeIdIterator.hasNext() )
+        for ( long nodeId : nodeIds )
         {
-            state.locks().optimistic()
-                    .acquireExclusive( state.lockTracer(), ResourceTypes.NODE, nodeIdIterator.next() );
+            state.locks().optimistic().acquireExclusive( state.lockTracer(), ResourceTypes.NODE, nodeId );
```
tinwelint: You could use one call since the id argument is a vararg

tinwelint (commented):
Looks good, I'd guess there's no easy way of avoiding the added hit for creating the node id set and sorting those. I just had some comments for improvements.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
fickludd: Actually, for some reason `releaseExclusive` is not varargs. Maybe that should be changed, but because it has 7 implementations I would rather not do it in this PR.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/TwoPhaseNodeForRelationshipLocking.java
fickludd: Hmm, I don't know how common that could be. If we start reusing objects here we should think about concurrency etc. so we store them in the correct place, which could be annoying. On the other hand, the created objects currently should be very short lived and probably thrown out on the first young GC. I think it will be fine to keep them.

tinwelint: Yup, agree it may add complexity/other issues.

tinwelint (approved):

PR #10167, opened 2017-09-28
Bugfix that nested lists can not be used with max or min

sherfert:
This also adds missing acceptance tests for the max aggregation function. Now they should work according to 
https://github.com/opencypher/openCypher/blob/master/cip/1.accepted/CIP2016-06-14-Define-comparability-and-equality-as-well-as-orderability-and-equivalence.adoc

changelog: Change the behaviour of min and max to not throw exceptions when aggregating values other than numbers and strings.


pontusmelke (changes requested):
Nicely done!

Since this PR changes behavior, make sure we add a note about it in the changelog. Ping me if you're unsure of how to do that.

pontusmelke (approved):

petraselmer:
This is great. Only one suggestion: I would recommend adding a test or two, for completeness and full clarity, showing `max()` in lists with more than one element - such as `[23, 45]` and also with `null` in the list.

sherfert:
I will add another commit then!

systay:
[Cypher max() on list expressions](https://trello.com/c/4DbxCHkI/75-cypher-max-on-list-expressions)

PR #10168, opened 2017-09-28
Handle INACCESSIBLE credentials

OliviaYtterbrink:
Solves division by 0

### DIFF
--- a/community/security/src/main/java/org/neo4j/server/security/auth/Credential.java
+++ b/community/security/src/main/java/org/neo4j/server/security/auth/Credential.java
@@ -63,10 +63,10 @@ public class Credential
         return byteEquals( passwordHash, hash( salt, password ) );
     }
 
-
     /**
      * <p>Utility method that replaces Arrays.equals() to avoid timing attacks.
-     * The length of the loop executed will always be the length of the given password</p>
+     * The length of the loop executed will always be the length of the given password.
+     * Remember {@link #INACCESSIBLE} credentials should still execute loop for the length of given password.</p>
      *
      * @param actual the actual password
      * @param given password given by the user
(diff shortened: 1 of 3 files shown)


community/security/src/main/java/org/neo4j/server/security/auth/Credential.java
```diff
@@ -83,13 +84,21 @@ private boolean byteEquals( byte[] actual, byte[] given )
             return false;
         }
         boolean result = true;
+        boolean accessible = true;
```
lutovich, line 86: Maybe rename this to smth like `actualArrayIsNotEmpty`? Otherwise `accessible` does not make sense for byte arrays.

community/security/src/main/java/org/neo4j/server/security/auth/Credential.java
```diff
@@ -106,7 +115,9 @@ public boolean equals( Object o )
 
         Credential that = (Credential) o;
 
-        return byteEquals( this.salt, that.salt ) && byteEquals( this.passwordHash, that.passwordHash );
+        boolean saltEquals = byteEquals( this.salt, that.salt );
+        boolean passwordEquals = byteEquals( this.passwordHash, that.passwordHash );
+        return saltEquals && passwordEquals;
```
lutovich, line 122: Was this condition split for readability or to always execute both branches? Maybe add a comment if it is the later

community/security/src/main/java/org/neo4j/server/security/auth/Credential.java
praveenag: This is to handle to accessible case as explained in the comment.

community/security/src/main/java/org/neo4j/server/security/auth/Credential.java
praveenag: Yes this was done to always to execute both branches. We'll add the comment.

eebus (approved):

PR #10169, opened 2017-09-28
Enable hazelcast logging through custom factory

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
```diff
@@ -26,6 +26,7 @@
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.logging.Level;
```
andrewkerr9000, line 29: Is this only for Hazelcast or are we using java logging elsewhere?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/discovery/HazelcastLogging.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class HazelcastLogging implements LoggerFactory
+{
+    // there is no constant in the hazelcast library for this
+    private static final String HZ_LOGGING_CLASS_PROPERTY = "hazelcast.logging.class";
+
+    private static LogProvider logProvider;
+    private static Level minLevel;
+
+    static void enable( LogProvider logProvider, Level minLevel )
```
andrewkerr9000, line 37: So static mutable methods are to control the fields of objects instantiated by Hazelcast? I hadn't thought of that possibility.

If hazelcast has some means of injecting arbitrary config then for an alternative implementation this class could read config instead of statics.

andrewkerr9000 (approved):
LGTM

PR #10178, opened 2017-09-29
Make the fulltext add-on index applier more robust wrt. to various failure modes

chrisvest:


ragadeeshu (commented):
Looks very good overall. I'm very curious about the choice to implements stubs instead of using mocks?

chrisvest:
@ragadeeshu To avoid encouraging testing with mocks. Mocks encourages over-specification, and tighter coupling between tests and irrelevant implementation details. Tests should only depend on observable behaviour of a components public API.

ragadeeshu (approved):

PR #10183, opened 2017-10-02
Encrypted Backups (3.2)

phughk:
Backups over CC may be encrypted


phughk:
Closed as this isn't going into 3.2

PR #10184, opened 2017-10-02
Fixing a broken test for patterns in case statements

sherfert:


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/PatternComprehensionAcceptanceTest.scala
```diff
@@ -135,8 +135,9 @@ class PatternComprehensionAcceptanceTest extends ExecutionEngineFunSuite with Cy
     val query = "MATCH (n:Start) RETURN [p = (n)-->() | p] AS list"
 
     val result = executeWith(expectedToSucceed, query)
-
-    result.toList should equal(List(Map("list" -> List(PathImpl(n1, r, n2)))))
+    graph.inTx {
```
craigtaverner: The requirement that actions on the results after pulling them could still require a transaction seems to be a behaviour change that might have consequences for users, not just in tests. I think we need a better way. Since the relationship values are created after looking up in the relationship store, they could, in principle, be created with all four fields (id, start, type, end) and therefore not require a later lookup (or transaction). It would be a bigger change, but a safer one.

craigtaverner (changes requested):

community/kernel/src/main/java/org/neo4j/helpers/BaseToObjectValueWriter.java
```diff
@@ -33,6 +33,7 @@
 import org.neo4j.graphdb.PropertyContainer;
 import org.neo4j.graphdb.Relationship;
 import org.neo4j.graphdb.spatial.Point;
+import org.neo4j.graphdb.traversal.Paths;
```
craigtaverner: Unused import

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/PatternComprehensionAcceptanceTest.scala
```diff
@@ -135,7 +135,6 @@ class PatternComprehensionAcceptanceTest extends ExecutionEngineFunSuite with Cy
     val query = "MATCH (n:Start) RETURN [p = (n)-->() | p] AS list"
 
     val result = executeWith(expectedToSucceed, query)
-
```
craigtaverner: Unnecessary change

craigtaverner (changes requested):

craigtaverner (approved):

PR #10185, opened 2017-10-02
Fix javadoc filters

klaren:
Javadoc needs `sourcepath` specified for now, however, this breaks the aggregate report.
Will be fixed if this following issue is resolved: https://issues.apache.org/jira/browse/MJAVADOC-497

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,7 +175,7 @@
             <detectLinks>false</detectLinks>
             <detectOfflineLinks>false</detectOfflineLinks>
             <quiet>true</quiet>
-            <sourcepath>${project.basedir}/src/main/java</sourcepath>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -177,6 +175,7 @@
             <detectLinks>false</detectLinks>
             <detectOfflineLinks>false</detectOfflineLinks>
             <quiet>true</quiet>
+            <sourcepath>${project.basedir}/src/main/java</sourcepath>
```
MishaDemianenko: i think you should be able to reference that using sourceDirectory property as described here http://maven.apache.org/pom.html#Resources

MishaDemianenko:
@klaren one comment otherwise good to merge

pom.xml
klaren: Great suggestion, works like a charm.

PR #10187, opened 2017-10-02
3.4 dependency check

praveenag:
* Create owasp security report for dependencies.
* Upgrade dependent libraries to latest versions.

### DIFF
--- a/community/bolt/LICENSES.txt
+++ b/community/bolt/LICENSES.txt
@@ -17,6 +17,7 @@ Apache Software License, Version 2.0
   Lucene Memory
   Lucene QueryParsers
   Netty/All-in-One
+  Objenesis
   opencsv
   parboiled-core
   parboiled-scala
--- a/pom.xml
+++ b/pom.xml
@@ -889,12 +889,12 @@
       <dependency>
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-nop</artifactId>
-        <version>1.8.0-beta0</version>
+        <version>1.7.25</version>
       </dependency>
       <dependency>
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-api</artifactId>
-        <version>1.8.0-beta0</version>
+        <version>1.7.25</version>
       </dependency>
 
       <!-- testing -->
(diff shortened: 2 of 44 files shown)


pom.xml
```diff
@@ -864,17 +884,17 @@
...
         <groupId>com.github.ben-manes.caffeine</groupId>
         <artifactId>caffeine</artifactId>
-        <version>2.3.3</version>
+        <version>2.6.1</version>
       </dependency>
       <dependency>
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-nop</artifactId>
-        <version>1.7.22</version>
+        <version>1.8.0-beta0</version>
```
MishaDemianenko: beta0? I do not think we should have any beta dependencies in the product

package-depencies.txt
```diff
@@ -0,0 +1,5217 @@
+[INFO] Scanning for projects...
```
MishaDemianenko: also I do not understand why do we need this file in a repo?

MishaDemianenko (changes requested):

package-depencies.txt
praveenag: Oh that was added by mistake. Let me remove it.

pom.xml
praveenag: I agree. That missed my initial sweep. I will update it to the latest patch`1.7.25`.

praveenag:
@MishaDemianenko How about now?

pom.xml
```diff
@@ -1025,17 +1045,17 @@
...
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-lang3</artifactId>
-        <version>3.5</version>
+        <version>3.7</version>
       </dependency>
       <dependency>
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-text</artifactId>
-        <version>1.1</version>
+        <version>1.2</version>
```
MishaDemianenko: i remember this one should be the same as in caps project and 1.1 was chosen exactly for that reason not so long time ago

community/bolt/LICENSES.txt
```diff
@@ -15,6 +15,7 @@ Apache Software License, Version 2.0
   Lucene Memory
   Lucene QueryParsers
   Netty/All-in-One
+  Objenesis
```
MishaDemianenko, line 20: do you know where Objenesis comes from now? I remember we killed it some time ago. Do we really need it back or we can exclude it?

MishaDemianenko (changes requested):
couple of comments

praveenag:
Merged after blessings from @MishaDemianenko 🙏

PR #10188, opened 2017-10-02
Manage bloom indexed keys by procedure

ragadeeshu:


enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProvider.java
```diff
@@ -244,6 +234,29 @@ public InternalIndexState getState( String identifier, FulltextIndexType type )
         return applyToMatchingIndex( identifier, type, LuceneFulltext::getState );
     }
 
+    public void drop( String identifier, FulltextIndexType type ) throws IOException
+    {
+        if ( type == FulltextIndexType.NODES )
+        {
+            nodeIndices.remove( identifier ).drop();
+            writableNodeIndices.remove( identifier ).close();
```
chrisvest: We need to protect these hashmaps from concurrent access.

chrisvest (changes requested):
Looks fine, except we need to protect the maps in the provider, now.

ragadeeshu:
@chrisvest updated

chrisvest (approved):

PR #10196, opened 2017-10-03
Change aux package to util

fickludd:
'aux' is apparently reserved by windows, and project
containing a 'aux' directory cannot be checked out
by windows users!


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_4/codegen/GeneratedMethodStructure.scala
```diff
@@ -61,7 +62,7 @@ object GeneratedMethodStructure {
 }
 
 
-class GeneratedMethodStructure(val fields: Fields, val generator: CodeBlock, aux: AuxGenerator, tracing: Boolean = true,
+class GeneratedMethodStructure(val fields: Fields, val generator: CodeBlock, util: AuxGenerator, tracing: Boolean = true,
```
henriknyman: This variable name didn't really need to change

fickludd: ...so windows doesn't complain about variables named `aux` you think? :)

PR #10197, opened 2017-10-03
Atomic schema cache state updates

MishaDemianenko:
Previously schema state update was non atomic and readers
where observing partial cache updates.
PR changes that behaviour by preventing concurrent updates of cache,
performing atomic cache state update.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
@@ -19,13 +19,15 @@
  */
 package org.neo4j.kernel.impl.api.store;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.StampedLock;
 import java.util.function.Function;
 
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
```diff
@@ -40,209 +42,279 @@
...
-    private final Map<Long,ConstraintRule> constraintRuleById = new ConcurrentHashMap<>();
-    private final Set<ConstraintDescriptor> constraints = newSetFromMap( new ConcurrentHashMap<>() );
-
-    private final Map<SchemaDescriptor,IndexDescriptor> indexDescriptors = new ConcurrentHashMap<>();
-    private final Map<Integer,Set<IndexDescriptor>> indexDescriptorsByLabel = new ConcurrentHashMap<>();
-    private final ConstraintSemantics constraintSemantics;
-
-    private final Map<Class<?>,Object> dependantState = new ConcurrentHashMap<>();
-    private final Map<Integer,List<IndexDescriptor>> indexByProperty = new ConcurrentHashMap<>();
+    private final Lock cacheUpdateLock = new StampedLock().asWriteLock();
```
chrisvest, line 52: Why this particular lock instead of, say, `new ReentrantLock()`?

chrisvest (approved):
Just one question.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/SchemaCache.java
MishaDemianenko: Since we do not need reentrancy part and need simple single entry lock, decided to go for this one. But it can be any lock or semaphore in principle

PR #10199, opened 2017-10-04
3.4 cypher util auto formatting

fickludd:
We have devised a strategy for slowly adding formatting rules to scala code. It was decided that applying them on the whole code base at once was bad because it would make investigating the git history much harder. Instead we can initially add it to new modules as we are creating them, and add CI checks on one module at a time.

This PR adds some auto-formatting rules for [scalafmt](http://scalameta.org/scalafmt/), and applies them on the cypher utils module. What is left to do is to add style-checking for this module in maven.

### DIFF
--- a/community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/NonEmptyList.scala
+++ b/community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/NonEmptyList.scala
@@ -33,22 +33,23 @@ object NonEmptyList {
   def apply[T](first: T, tail: T*): NonEmptyList[T] =
     loop(Last(first), tail.iterator).reverse
 
-  def newBuilder[T]: mutable.Builder[T, Option[NonEmptyList[T]]] = new mutable.Builder[T, Option[NonEmptyList[T]]] {
-    private var vecBuilder = Vector.newBuilder[T]
+  def newBuilder[T]: mutable.Builder[T, Option[NonEmptyList[T]]] =
+    new mutable.Builder[T, Option[NonEmptyList[T]]] {
+      private var vecBuilder = Vector.newBuilder[T]
 
-    override def +=(elem: T): this.type = {
-      vecBuilder += elem
-      this
-    }
+      override def +=(elem: T): this.type = {
+        vecBuilder += elem
+        this
+      }
 
-    override def result(): Option[NonEmptyList[T]] = {
-      vecBuilder.result().toNonEmptyListOption
-    }
+      override def result(): Option[NonEmptyList[T]] = {
+        vecBuilder.result().toNonEmptyListOption
+      }
 
-    override def clear(): Unit = {
-      vecBuilder.clear()
+      override def clear(): Unit = {
+        vecBuilder.clear()
+      }
     }
-  }
 
   implicit def canBuildFrom[T] = new CanBuildFrom[Any, T, Option[NonEmptyList[T]]] {
     def apply(from: Any) = newBuilder[T]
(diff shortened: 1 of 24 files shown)


community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/NonEmptyList.scala
```diff
@@ -141,7 +145,7 @@ sealed trait NonEmptyList[+T] {
   final def filter[X >: T](f: X => Boolean): Option[NonEmptyList[T]] =
     foldLeft[Option[NonEmptyList[T]]](None) {
       case (None, elem) => if (f(elem)) Some(Last(elem)) else None
-      case (acc@Some(nel), elem) => if (f(elem)) Some(Fby(elem, nel)) else acc
+      case (acc @ Some(nel), elem) => if (f(elem)) Some(Fby(elem, nel)) else acc
```
henriknyman, line 148: Nice!

community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/NonEmptyList.scala
```diff
@@ -262,41 +274,71 @@ sealed trait NonEmptyList[+T] {
...
+      acc: Either[
+        (NonEmptyList[A], Option[NonEmptyList[B]]),
+        (Option[NonEmptyList[A]], NonEmptyList[B])
+      ]
+  ): Either[
+    (NonEmptyList[A], Option[NonEmptyList[B]]),
+    (Option[NonEmptyList[A]], NonEmptyList[B])
+  ] =
     (value, acc) match {
       case (Left(elem), Left((lefts, optRights))) => Left((Fby(elem, lefts), optRights))
```
henriknyman, line 313: Can you make it break all these lines after => ? I think that would increase debuggability. Alternatively contextually break all the lines if it decides to break at least one (similar to what it does with comma separated lists).

community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/symbols/TypeRange.scala
```diff
@@ -42,16 +46,16 @@ case class TypeRange(lower: CypherType, upper: Option[CypherType]) {
...
       val newUpper = upper.fold(that.upper)(t => Some(that.upper.fold(t)(_ leastUpperBound t)))
       if (newUpper.isDefined && !(newLower isAssignableFrom newUpper.get))
         None
       else
         Some(TypeRange(newLower, newUpper))
-  }
+    }
 
-  def covariant = copy(upper = None)
+  def covariant: TypeRange = copy(upper = None)
```
henriknyman, line 58: Automatic type annotation on public methods is very useful

community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/symbols/TypeSpec.scala
```diff
@@ -166,10 +183,21 @@ class TypeSpec private (private val ranges: Seq[TypeRange]) extends Equals {
...
+  ): StringBuilder = {
     val strings = toStrings
     if (strings.length > 1)
-      strings.dropRight(1).addString(b, start, sep, "").append(lastSep).append(strings.last).append(end)
+      strings
+        .dropRight(1)
+        .addString(b, start, sep, "")
+        .append(lastSep)
+        .append(strings.last)
+        .append(end)
```
henriknyman, line 200: This is nice

henriknyman (approved):
I kind of like most of what its doing here, even though there are some code constructs not covered by this module that I wonder how it would treat. It is also hard to get a feel for how strict it is by just looking at the result.

community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/symbols/TypeRange.scala
fickludd: Oh sorry, I added that manually =P. But it if useful though =).

PR #10202, opened 2017-10-04
Fix online backup of read-only instances with schema and legacy indexes.

### LINKED ISSUE #10089: Database backup fails in read-only mode
The command `neo4j-admin backup` fails if the database being backed up is in read-only mode.

**Neo4j Version:** 3.2.2 (Enterprise)
**Operating System:** Tested with the database running on Ubuntu 16.04, neo4j-admin was tested from Ubuntu 16.04 and Windows 10 v1703
**API:** neo4j-admin

### Steps to reproduce
1. Run neo4j-admin backup, pointed at a running database (either remotely or from the same machine)

### Expected behavior
It should create a backup of the running database

### Actual behavior
It fails with the following output:
`Doing full backup...`
`command failed: Backup failed: Unable to force ContractCheckingIndexProxy -> OnlineIndexProxy[accessor:org.neo4j.kernel.api.impl.schema.LuceneIndexAccessor@ae4d8a8, descriptor:Index( GENERAL, :label[39](property[51]) )]`

chrisvest:
Taking an online backup of a read-only instance would previously throw a variety of exceptions, depending on its schema/legacy indexes. This fixes #10089.

This also fixes a plurality of bugs where create and drop of legacy indexes are concurrent with commits of changes to those indexes. These bugs had snuck in because the relevant test didn't have a proper test-name, and thus the test was never run by our build system.

Also fixed in this PR, is a bug in _relationship_ legacy indexes, where changes could be committed with different document types. This was already fixed for node indexes a long time ago, but that fix was apparently never applied to the relationship indexes.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneCommandApplier.java
+++ b/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneCommandApplier.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.neo4j.kernel.api.exceptions.legacyindex.LegacyIndexNotFoundKernelException;
 import org.neo4j.kernel.impl.api.TransactionApplier;
 import org.neo4j.kernel.impl.index.IndexCommand;
 import org.neo4j.kernel.impl.index.IndexCommand.AddNodeCommand;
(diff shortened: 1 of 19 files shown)


community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneCommandApplier.java
```diff
@@ -72,46 +73,67 @@ public boolean visitIndexAddRelationshipCommand( AddRelationshipCommand command
...
+            // not be able to recover, which is why this check is here.
+            if ( value != null )
+            {
+                context.ensureWriterInstantiated();
+                context.indexType.addToDocument( context.getDocument( entityId, true ).document, key, value );
+            }
+        }
+        catch ( LegacyIndexNotFoundKernelException ignore )
+        {
+            // Pretend the index never existed.
```
tinwelint, line 96: Useful for recovery, or when could an index just disappear like that?

tinwelint (commented):
Looks great. I have one little question.

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneCommandApplier.java
chrisvest: As far as I can tell, there is locking to make legacy index configuration changes mutually exclusive with updates to those indexes, so legacy indexes can basically be dropped at any time.

What happens here is that we ignore updates that were slated for indexes that are dropped concurrently with our commit. Previously, this would have thrown some sort of RuntimeException.

tinwelint: I'm absolutely certain you meant to say _...there is NO locking to make legacy index..._. OK then I understand.

tinwelint (approved):

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/LuceneCommandApplier.java
chrisvest: Yes, that's exactly what I meant to say. Somehow that important little detail didn't make it from my brain to my fingers.

PR #10206, opened 2017-10-04
Close AbstractTraverserIterator resource only once

fickludd:
As statements are reference counted, it is not idea to let the AbstractTraverserIterator close it's resource (statement) more than once. This was exposed with the increased strictness of asserting that statements are open before operations.

changelog: Fix a bug that would cause exceptions with the message `The statement has been closed.` when using the traversal API from inside procedures.


fickludd:
@burqen it would be nice to merge this soon so we might get it into 3.3.0-RC2 (and therefore hopefully 3.3.0). Otherwise we might have partially crippled apoc in 3.3.

burqen (approved):

PR #10211, opened 2017-10-04
Change database panic message.

MishaDemianenko:
Previous message was confusing and was asking people to do "tx recovery/restart"
"Tx" was confusing and some people tried to find those methods in a transaction
objects in driver/kernel/etc api.
Also it can be quite problematic to do only recovery on your own.
New panic message:
Critical error has encountered (see database logs for more details). Please restart the database.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
@@ -28,8 +28,8 @@ import static org.neo4j.helpers.Exceptions.withCause;
 
 public class DatabaseHealth
 {
-    private static final String panicMessage = "Database has encountered some problem, "
-            + "please perform necessary action (tx recovery/restart)";
+    private static final String panicMessage = "Critical error has encountered (see database logs for more details). " +
+            "Please restart the database.";
     private static final Class<?>[] CRITICAL_EXCEPTIONS = new Class[]{OutOfMemoryError.class};
 
     private volatile boolean healthy = true;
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
```diff
@@ -28,8 +28,8 @@
 
 public class DatabaseHealth
 {
-    private static final String panicMessage = "Database has encountered some problem, "
-            + "please perform necessary action (tx recovery/restart)";
+    private static final String panicMessage = "Critical error has encountered (see database logs for more details). " +
+            "Please restart the database.";
```
chrisvest: It's at least missing a "been" in the sentence.

I would phrase it `The database has encountered a critical error, and needs to be restarted. Please see the debug.log file for details.`

chrisvest (commented):
I think the wording could be improved.

community/kernel/src/main/java/org/neo4j/kernel/internal/DatabaseHealth.java
MishaDemianenko: There is a small problem with referencing just 'debug.log' in a message since if you on a driver level you need a highlight that error is in server logs and not in a driver logs of any sort.
WDYT about: "The database has encountered a critical error, and needs to be restarted. Please see database logs for more details."?

chrisvest: Sounds good 👍

MishaDemianenko:
@chrisvest updated

chrisvest (approved):

PR #10217, opened 2017-10-05
Allow usage of system timezone for database logs

MishaDemianenko:
Previously all database logs where using UTC only for time in all logs.
This PR will allow to change behaviour and use system timezone for all the logs
using newly introduced "dbms.logs.timezone" database setting.

Current possible setting values are:
UTC(default value) - log in UTC;
SYSTEM - log in a system default timezone.

### DIFF
--- a/community/consistency-check/src/test/java/org/neo4j/consistency/ConsistencyCheckToolTest.java
+++ b/community/consistency-check/src/test/java/org/neo4j/consistency/ConsistencyCheckToolTest.java
@@ -25,7 +25,6 @@ import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
-import org.mockito.stubbing.Answer;
 
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
@@ -103,7 +102,7 @@ public class ConsistencyCheckToolTest
             Mockito.when( service.runFullConsistencyCheck( any( File.class ), any( Config.class ),
                     any( ProgressMonitorFactory.class ), any( LogProvider.class ), any( FileSystemAbstraction.class ),
                     eq( false ), any( CheckConsistencyConfig.class ) ) )
-                    .then( (Answer<ConsistencyCheckService.Result>) invocationOnMock ->
+                    .then( invocationOnMock ->
                     {
                         LogProvider provider = invocationOnMock.getArgumentAt( 3, LogProvider.class );
                         provider.getLog( "test" ).info( "testMessage" );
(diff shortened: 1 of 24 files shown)


community/consistency-check/src/test/java/org/neo4j/consistency/ConsistencyCheckToolTest.java
```diff
@@ -83,6 +94,38 @@ public void runsConsistencyCheck() throws Exception
...
+    public void consistencyCheckerLogUseSystemTimezoneIfConfigurable() throws Exception
+    {
+        TimeZone defaultTimeZone = TimeZone.getDefault();
+        try
+        {
+            ConsistencyCheckService service = mock( ConsistencyCheckService.class );
+            Mockito.when( service.runFullConsistencyCheck( any( File.class ), any( Config.class ),
+                    any( ProgressMonitorFactory.class ), any( LogProvider.class ), any( FileSystemAbstraction.class ),
+                    eq( false ), any( CheckConsistencyConfig.class ) ) )
+                    .then( (Answer<ConsistencyCheckService.Result>) invocationOnMock ->
```
burqen: Cast is reduntant

community/consistency-check/src/test/java/org/neo4j/consistency/ConsistencyCheckToolTest.java
```diff
@@ -177,6 +220,24 @@ public void failWhenStoreWasNonCleanlyShutdown() throws Exception
         runConsistencyCheckToolWith( fs.get(), storeDirectory.graphDbDir().getAbsolutePath() );
     }
 
+    private void checkLogRecordTimeZone( ConsistencyCheckService service, String[] args, int i, String s )
```
burqen: Could use better param names here

community/server/src/main/java/org/neo4j/server/AbstractNeoServer.java
```diff
@@ -362,6 +363,7 @@ private void setUpHttpLogging() throws IOException
 
         AsyncRequestLog requestLog = new AsyncRequestLog(
                 dependencyResolver.resolveDependency( FileSystemAbstraction.class ),
+                config.get( GraphDatabaseSettings.log_timezone ).getZoneId(),
```
klaren: Static import, for consistency? :)

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/PlatformModule.java
```diff
@@ -278,6 +278,7 @@ protected LogService createLogService( LogProvider userLogProvider )
             builder.withLevel( debugContext, Level.DEBUG );
         }
         builder.withDefaultLevel( config.get( GraphDatabaseSettings.store_internal_log_level ) );
+        builder.withTimeZone( config.get( GraphDatabaseSettings.log_timezone ).getZoneId() );
```
klaren: Could be chained.

burqen (changes requested):
I think this looks good. Just some small tweaks needed.

community/server/src/main/java/org/neo4j/server/ServerBootstrapper.java
```diff
@@ -172,6 +172,7 @@ protected abstract NeoServer createNeoServer( Config config, GraphDatabaseDepend
     private static LogProvider setupLogging( Config config )
     {
         LogProvider userLogProvider = FormattedLogProvider.withoutRenderingContext()
+                            .withZoneId( config.get( GraphDatabaseSettings.log_timezone ).getZoneId() )
```
klaren, line 175: The field `log` is assigned a non-timezone logger first, not sure if this will pose a problem or not.

klaren (commented):
Looks good! 3 small comments.

community/server/src/main/java/org/neo4j/server/ServerBootstrapper.java
MishaDemianenko: That is for cases where we fail to parse provided configuration while we still need to provide log to the user, for those cases we can't provide valid configuration of logger and just use defaults

MishaDemianenko:
@burqen @klaren updated

burqen (approved):

PR #10220, opened 2017-10-06
Fixes edge-case with merging two nodes in GBPTree

tinwelint:
There was an edge case where, if max key count in leaves were odd, would read the wrong key to bubble upwards to the parent(s) when merging a node into its right sibling. This would in turn make the wrong structural changes to the tree, resulting in a negative key count in a parent, instead of actually removing the expected key from another parent. This would later result in reading garbage data (recognized on read by throwing exception saying something like reading BROKEN/BROKEN pointer pair) when reading data from that node.

This PR fixes this by reading the correct key on merge, also doing the read _before_ merging the nodes together, as to make the code easier to understand. Existing trees affected by this will have to be deleted and rebuilt.


MishaDemianenko:
[Fusion Index error after detach delete all nodes.](https://trello.com/c/6lcgajK8/83-fusion-index-error-after-detach-delete-all-nodes)

burqen (approved):

PR #10224, opened 2017-10-08
graphGetProperty return NO_VALUE not null

pontusmelke:
Method `graphGetProperty` was returning a `null` instead of `NO_VALUE`
which causes a `NullPointerException`.


fickludd (approved):
How annoying that I missed this case... thanks for cleaning up.

PR #10226, opened 2017-10-09
3.4 - Resolved Consistency check errors in backup

phughk:
Before applying this patch, the neo4j-admin backup tool can fail
performing a backup. This is due to the omission of the step to recover
a store once it has been copied. The solution is starting a temporary
embedded database after every full backup.


enterprise/backup/src/main/java/org/neo4j/backup/BackupFlow.java
```diff
@@ -22,17 +22,21 @@
...
 import org.neo4j.consistency.ConsistencyCheckService;
 import org.neo4j.consistency.checking.full.ConsistencyFlags;
 import org.neo4j.helpers.progress.ProgressMonitorFactory;
+import org.neo4j.io.pagecache.PageCache;
 import org.neo4j.kernel.configuration.Config;
+import org.neo4j.kernel.internal.GraphDatabaseAPI;
 import org.neo4j.logging.LogProvider;
 
 import static java.lang.String.format;
+import static org.neo4j.backup.BackupProtocolService.startTemporaryDb;
```
martinfurmanski: Some imports here look superfluous.

enterprise/backup/src/main/java/org/neo4j/backup/AbstractBackupSupportingClassesFactory.java
```diff
@@ -106,6 +116,11 @@ private static BackupDelegator backupDelegator( RemoteStore remoteStore, CatchUp
         return new BackupDelegator( remoteStore, catchUpClient, storeCopyClient, new ClearIdService( new IdGeneratorWrapper() ) );
     }
 
+    private static PageCache createPageCache( FileSystemAbstraction fileSystemAbstraction, Config config )
```
martinfurmanski, line 105: Wrapping just one call seems a bit superfluous. You could just import ConfigurableStandalonePageCacheFactory.createPageCache and the code line would be exactly the same.

enterprise/backup/src/main/java/org/neo4j/backup/AbstractBackupSupportingClassesFactory.java
```diff
@@ -77,8 +84,11 @@
      */
     BackupSupportingClasses createSupportingClassesForBackupStrategies( Config config )
     {
-        PageCache pageCache = ConfigurableStandalonePageCacheFactory.createPageCache( fileSystemAbstraction, config );
-        return new BackupSupportingClasses( backupDelegatorFormConfig( pageCache, config ), haFromConfig( pageCache ) );
+        PageCache pageCache = createPageCache( fileSystemAbstraction, config );
+        return new BackupSupportingClasses(
+                backupDelegatorFormConfig( pageCache, config ),
```
andrewkerr9000: `From` rather than `Form`?

enterprise/backup/src/main/java/org/neo4j/backup/BackupFlow.java
andrewkerr9000: Unneeded imports?

enterprise/backup/src/main/java/org/neo4j/backup/BackupProtocolService.java
```diff
@@ -93,14 +93,12 @@
...
 
-    static final String DIFFERENT_STORE_MESSAGE = "Target directory contains full backup of a logically different store.";
-
     private final Supplier<FileSystemAbstraction> fileSystemSupplier;
     private final LogProvider logProvider;
     private final Log log;
     private final OutputStream logDestination;
     private final Monitors monitors;
-    private final Optional<PageCache> pageCacheOptional;
+    private final PageCache pageCache;
```
andrewkerr9000, line 104: Does pageCache need to be closed?

enterprise/backup/src/main/java/org/neo4j/backup/BackupProtocolService.java
```diff
@@ -109,8 +107,8 @@
 
     BackupProtocolService( OutputStream logDestination )
     {
-        this( DefaultFileSystemAbstraction::new, FormattedLogProvider.toOutputStream( logDestination ), logDestination,
-                new Monitors(), null );
+        this( DefaultFileSystemAbstraction::new, FormattedLogProvider.toOutputStream( logDestination ), logDestination, new Monitors(),
```
andrewkerr9000, line 113: The other constructor has `pageCache` annotated with `@Nullable` which isn't true any more.

enterprise/backup/src/main/java/org/neo4j/backup/ExistingBackupWithDifferentStoreException.java
```diff
@@ -0,0 +1,40 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.backup;
+
+import java.io.File;
+
+import org.neo4j.kernel.impl.store.MismatchingStoreIdException;
+
+class ExistingBackupWithDifferentStoreException extends RuntimeException
```
andrewkerr9000: If this is being caught should it be a `RuntimeException`?

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
```diff
@@ -137,6 +148,44 @@ public void backupCanNotBePerformedOverBackupProtocol() throws Exception
...
+    public void dataIsInAUsableStateAfterBackup() throws Exception
+    {
+        // given database exists
+        Cluster cluster = startCluster( recordFormat );
+
+        // and the database has indexes
+        createIndexes( cluster.getDbWithAnyRole( Role.LEADER ).database() );
+
+        // and the database is being populated
+        AtomicBoolean continueFlagReference = new AtomicBoolean( false );
```
andrewkerr9000: Should this start as `true`?

andrewkerr9000 (commented):
Mostly minor comments, but there may be a bug in a test case.

enterprise/backup/src/main/java/org/neo4j/backup/BackupProtocolService.java
phughk: The page cache isn't opened here, so closing also doesn't happen here. Additionally closing it here will cause the other strategies not to work as they rely on the same page cache.

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
```diff
@@ -157,9 +206,9 @@ public static DbRepresentation createSomeData( GraphDatabaseService db )
     {
         try ( Transaction tx = db.beginTx() )
```
martinfurmanski: Cluster has a coreTx(..) method which would be good to use here and in other places because it can cope with leader-switches. This is one of the main reasons for tests being flaky.

enterprise/backup/src/main/java/org/neo4j/backup/ExistingBackupWithDifferentStoreException.java
phughk: Changed to a `Exception`, thanks!

enterprise/backup/src/main/java/org/neo4j/backup/BackupProtocolService.java
phughk: Addressed

enterprise/backup/src/main/java/org/neo4j/backup/ExistingBackupWithDifferentStoreException.java
```diff
@@ -0,0 +1,40 @@
...
+{
+    static final String DIFFERENT_STORE_MESSAGE = "Target directory contains full backup of a logically different store.";
+    private static final String BACKUP_LOCATION_MESSAGE = "Directory `%s` contains full backup of a logically different store.";
+
+    ExistingBackupWithDifferentStoreException( MismatchingStoreIdException mismatchingStoreIdException )
+    {
+        super( DIFFERENT_STORE_MESSAGE, mismatchingStoreIdException );
+    }
+
+    ExistingBackupWithDifferentStoreException( ExistingBackupWithDifferentStoreException cause, File backupLocation )
```
martinfurmanski: Should the cause really be of the same type?

enterprise/backup/src/main/java/org/neo4j/backup/ExistingBackupWithDifferentStoreException.java
```diff
@@ -0,0 +1,40 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.backup;
+
+import java.io.File;
+
+import org.neo4j.kernel.impl.store.MismatchingStoreIdException;
+
+class ExistingBackupWithDifferentStoreException extends Exception
```
martinfurmanski: Could you perhaps avoid introducing ExistingBackupWithDifferentStoreException altogether and just create MismatchingStoreIdExceptions, perhaps with a supporting class/factory if you want to factor out the messages? My thinking is that you for example won't have to catch both exceptions in BackupFlow.java and check the cause() for special handling.

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
phughk: Reworded and corrected to `true` for clarity

enterprise/backup/src/main/java/org/neo4j/backup/BackupFlow.java
phughk: Addressed

enterprise/backup/src/main/java/org/neo4j/backup/AbstractBackupSupportingClassesFactory.java
phughk: Addressed

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
```diff
@@ -137,6 +148,44 @@ public void backupCanNotBePerformedOverBackupProtocol() throws Exception
...
+        createIndexes( cluster.getDbWithAnyRole( Role.LEADER ).database() );
+
+        // and the database is being populated
+        AtomicBoolean populateDatabaseFlag = new AtomicBoolean( true );
+        new Thread( () -> repeatedlyPopulateDatabase( clusterLeader( cluster ).database(), populateDatabaseFlag ) )
+                .start(); // populate db with number properties etc.
+        oneOffShutdownTasks.add( () -> populateDatabaseFlag.set( false ) ); // kill thread
+
+        // then backup is successful
+        String ip = TestHelpers.backupAddressCc( clusterLeader( cluster ).database() );
```
martinfurmanski: Not really an ip. Perhaps address or hostnamePort?

martinfurmanski (changes requested):
Looks mostly good to me, just a few minor comments.

enterprise/backup/src/main/java/org/neo4j/backup/BackupFlow.java
phughk: Addressed

enterprise/backup/src/main/java/org/neo4j/backup/AbstractBackupSupportingClassesFactory.java
phughk: Addressed

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
phughk: Addressed

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
phughk: Addressed

enterprise/backup/src/main/java/org/neo4j/backup/ExistingBackupWithDifferentStoreException.java
phughk: Addressed - removed

enterprise/backup/src/main/java/org/neo4j/backup/ExistingBackupWithDifferentStoreException.java
phughk: Addressed - removed `ExistingBackupWithDifferentStoreException` class

phughk:
Changes addressed

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandHaIT.java
```diff
@@ -97,27 +124,65 @@ public void makeSureBackupCanBePerformedWithCustomPort() throws Exception
...
+    {
+        while ( continueFlagReference.get() )
+        {
+            createSomeData( db );
+        }
+    }
+
+    private void createIndexes( GraphDatabaseService db )
+    {
+        try ( Transaction tx = db.beginTx() )
```
martinfurmanski, line 182: This also needs coreTx(..).

enterprise/kernel/src/test/java/org/neo4j/util/TestHelpers.java
```diff
@@ -76,7 +80,15 @@ public static int runBackupToolFromOtherJvmToGetExitCode( File neo4jHome, String
         return new ProcessStreamHandler( process, false ).waitForResult();
     }
 
-    public static String backupAddress( GraphDatabaseAPI graphDatabase )
+    public static String backupAddressCc( GraphDatabaseAPI graphDatabase )
+    {
+        ListenSocketAddress hostnamePort = graphDatabase.getDependencyResolver().resolveDependency( Config.class ).get(
+                listenAddress( "causal_clustering.transaction_listen_address", 6000 ) );
```
martinfurmanski, line 86: I think you should be fine with just config.get( CausalClusteringSettings.xxx ).

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
```diff
@@ -149,21 +198,28 @@ private Cluster startCluster( String recordFormat )
...
+                Node node = coreGraphDatabase.createNode( label );
+                node.setProperty( PROP_NAME, "Neo" );
+                node.setProperty( PROP_RANDOM, Math.random() * 10000 );
+                coreGraphDatabase.createNode().createRelationshipTo( node, RelationshipType.withName( "KNOWS" ) );
+                transaction.success();
+            } );
+        }
+        catch ( Exception e )
+        {
+            e.printStackTrace();
```
martinfurmanski: Just let the exception through. The coreTx(..) method will retry on any transient failures. Any other exceptions should be considered issues and bubble up to fail the test.

martinfurmanski (changes requested):

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandHaIT.java
phughk: This is in the HA part of the integration tests, therefore it doesn't have `coreTx`; Not addressed

enterprise/kernel/src/test/java/org/neo4j/util/TestHelpers.java
phughk: Actually this would cause a circular dependency; It's surprising that the string works though.

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandCcIT.java
phughk: Actually this code is called within a Thread, so it is more convenient to wrap it here than to have the method declare it throws and have to wrap it anyway in the Thread declaration. Not addressed, but used `launderedException` instead of `printStackTrace`

enterprise/backup/src/main/java/org/neo4j/backup/BackupCopyService.java
```diff
@@ -57,6 +61,10 @@ public void moveBackupLocation( File oldLocation, File newLocation ) throws Comm
     boolean backupExists( File destination )
     {
         File[] listFiles = outsideWorld.fileSystem().listFiles( destination );
+        List<String> filenames = Stream.of( Optional.ofNullable( listFiles ).orElse( new File[]{} ) )
+                .map( File::getName )
+                .collect( Collectors.toList() );
+        System.out.format( "List of backup files for `%s` are: %s\n", destination, filenames );
```
phughk: TODO delete

martinfurmanski (approved):

PR #10228, opened 2017-10-09
Fix NOTICE and LICENSES files to match new browser

klaren:
Licenses in the browser do not propagate automatically. This PR updates the licenses that are currently in use. This process should be automated in the future so it's not caught by accident a few weeks before release.

We also had a few libraries with dual licenses, this PR also resolves those.


community/licensecheck-config/src/main/resources/licensing-requirements-browser.xml
```diff
@@ -22,93 +22,660 @@
...
+    <artifact id="d3" name="D3.js">
+      <license>BSD License</license>
+    </artifact>
+
+    <!-- Auto-generated from yarn -->
+    <artifact id="ansi-escapes.js" name="ansi-escapes.js">
       <license>MIT</license>
     </artifact>
-    <artifact id="angular-local-storage" name="angular-local-storage">
+    <artifact id="ansi-regex.js" name="ansi-regex.js">
```
MishaDemianenko, line 55: some of the artifacts like "ansi-regex.js" are duplicated now

MishaDemianenko (commented):
Looks correct-ish, only one small comment about duplicates

PR #10242, opened 2017-10-10
CsvInput verifies id spaces before starting import

tinwelint:
such that it will fail fast if there are relationship headers specifying
id spaces which no node header has will fail explaining just that instead of
treating all those relationships as bad.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/Groups.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/Groups.java
@@ -19,6 +19,7 @@
  */
 package org.neo4j.unsafe.impl.batchimport.input;
 
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -49,8 +50,7 @@ public class Groups
         {
             if ( global != globalMode.booleanValue() )
             {
-                throw new IllegalStateException( "Mixing specified and unspecified group belongings " +
-                        "in a single import isn't supported" );
+                throw mixingOfGroupModesException();
             }
         }
 
@@ -65,5 +65,37 @@ public class Groups
             byName.put( name, group = new Group.Adapter( nextId++, name ) );
         }
         return group;
+    }
+
+    private IllegalStateException mixingOfGroupModesException()
+    {
+        return new IllegalStateException( "Mixing specified and unspecified group belongings " +
+                "in a single import isn't supported" );
+    }
+
+    public synchronized Group get( String name )
+    {
+        boolean global = name == null;
+        if ( globalMode != null && global != globalMode.booleanValue() )
+        {
+            throw mixingOfGroupModesException();
+        }
+
+        if ( name == null )
+        {
+            return Group.GLOBAL;
+        }
+
+        Group group = byName.get( name );
+        if ( group == null )
+        {
+            throw new HeaderException( "Group '" + name + "' not found. Available groups are: " + groupNames() );
+        }
+        return group;
+    }
+
+    private String groupNames()
+    {
+        return Arrays.toString( byName.keySet().toArray( new String[byName.keySet().size()] ) );
     }
 }
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/Groups.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/Groups.java
@@ -48,7 +48,7 @@ public class Groups
         }
         else
         {
-            if ( global != globalMode.booleanValue() )
+            if ( global != globalMode )
             {
                 throw mixingOfGroupModesException();
             }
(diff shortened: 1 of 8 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/Groups.java
```diff
@@ -66,4 +66,36 @@ public synchronized Group getOrCreate( String name )
...
+    private IllegalStateException mixingOfGroupModesException()
+    {
+        return new IllegalStateException( "Mixing specified and unspecified group belongings " +
+                "in a single import isn't supported" );
+    }
+
+    public synchronized Group get( String name )
+    {
+        boolean global = name == null;
+        if ( globalMode != null && global != globalMode.booleanValue() )
```
klaren: `.booleanValue()` is not needed

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInput.java
```diff
@@ -90,17 +98,74 @@ public CsvInput(
...
+                {
+                    Header header = relationshipHeaderFactory.create( dataStream, config, idType );
+                    verifyRelationshipHeader( header, Type.START_ID, dataStream.sourceDescription() );
+                    verifyRelationshipHeader( header, Type.END_ID, dataStream.sourceDescription() );
+                }
+            }
+        }
+        catch ( IOException e )
+        {
+            throw new UncheckedIOException( e );
```
klaren, line 144: :+1:

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/input/GroupsTest.java
```diff
@@ -50,4 +52,109 @@ public void shouldHandleConcurrentGetOrCreate() throws Throwable
...
     }
+
+    @Test
+    public void shouldFailOnMixedGroupModeInGetOrCreate() throws Exception
+    {
+        // given
+        Groups groups = new Groups();
+        groups.getOrCreate( null );
+
+        try
```
klaren: Would prefer to have expected exceptions in the added tests since nothing is invoked after the throw.

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInputTest.java
```diff
@@ -872,6 +885,95 @@ public void shouldSkipRelationshipValidationIfToldTo() throws Exception
...
+        try
+        {
+            // when
+            new CsvInput( data, defaultFormatNodeFileHeader(), relationshipData(), defaultFormatRelationshipFileHeader(), IdType.INTEGER,
+                    COMMAS, mock( Collector.class ), getRuntime().availableProcessors(), false );
+            fail( "Should not parse" );
+        }
+        catch ( InputException e )
+        {
+            // then
```
klaren, line 903: Same here

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInputTest.java
```diff
@@ -1005,6 +1107,51 @@ private static CharReadable charReader( String data )
...
+        CapturingDataFactories( Function<Configuration,CharReadable> factory, Decorator<ENTITY> decorator )
+        {
+            this.factory = factory;
+            this.decorator = decorator;
+        }
+
+        @Override
+        public Iterator<DataFactory<ENTITY>> iterator()
+        {
+            return Iterators.iterator( new DataFactory<ENTITY>()
```
klaren: Great place for a lambda ;)
```java
          return Iterators.iterator( config -> new Data<ENTITY>()
            {
                @Override
                public CharReadable stream()
                {
                    return last = factory.apply( config );
                }

                @Override
                public Decorator<ENTITY> decorator()
                {
                    return decorator;
                }
            } );
```

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/Groups.java
```diff
@@ -66,4 +66,36 @@ public synchronized Group getOrCreate( String name )
...
         return group;
     }
+
+    private IllegalStateException mixingOfGroupModesException()
+    {
+        return new IllegalStateException( "Mixing specified and unspecified group belongings " +
+                "in a single import isn't supported" );
+    }
+
+    public synchronized Group get( String name )
```
klaren, line 76: It's a bit hard to follow the flow, but it looks like this method was introduced to help `verifyRelationshipHeader()` assert that the header is there. Wouldn't a `Groups.isPresent(String)` make more sense and eliminate the duplicate code?

klaren (changes requested):
Fail fast and early! :tada:

I just have a few small comments.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/Groups.java
tinwelint: `isPresent` would be implemented almost like `get` wouldn't it?

klaren: That is the case, yes, disregard my previous comment :)

klaren (approved):

PR #10246, opened 2017-10-10
Shutdown index population jobs during indexing service shutdown

MishaDemianenko:
Shutdown indexing population jobs by canceling them and waiting to
finish ongoing population before closing all indexes down.
This change prevent ongoing index populations to throw exceptions because
of already closed indexes.
Also change order of stop/shutdown for label scan and indexing service
since indexes can use label scan store for indexes population

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobController.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobController.java
@@ -28,7 +28,7 @@ import org.neo4j.scheduler.JobScheduler;
 
 import static org.neo4j.scheduler.JobScheduler.Groups.indexPopulation;
 
-public class IndexPopulationJobController
+class IndexPopulationJobController
 {
     private final Set<IndexPopulationJob> populationJobs =
             Collections.newSetFromMap( new ConcurrentHashMap<IndexPopulationJob,Boolean>() );
@@ -39,7 +39,7 @@ public class IndexPopulationJobController
         this.scheduler = scheduler;
     }
 
-    public void stop() throws ExecutionException, InterruptedException
+    void stop() throws ExecutionException, InterruptedException
     {
         for ( IndexPopulationJob job : populationJobs )
         {
@@ -47,13 +47,13 @@ public class IndexPopulationJobController
         }
     }
 
-    public void startIndexPopulation( IndexPopulationJob job )
+    void startIndexPopulation( IndexPopulationJob job )
     {
         populationJobs.add( job );
         scheduler.schedule( indexPopulation, new IndexPopulationJobWrapper( job, this ) );
     }
 
-    public void indexPopulationCompleted( IndexPopulationJob populationJob )
+    void indexPopulationCompleted( IndexPopulationJob populationJob )
     {
         populationJobs.remove( populationJob );
     }
@@ -68,8 +68,7 @@ public class IndexPopulationJobController
         private IndexPopulationJob indexPopulationJob;
         private IndexPopulationJobController jobController;
 
-        IndexPopulationJobWrapper( IndexPopulationJob indexPopulationJob,
-                IndexPopulationJobController jobController )
+        IndexPopulationJobWrapper( IndexPopulationJob indexPopulationJob, IndexPopulationJobController jobController )
         {
             this.indexPopulationJob = indexPopulationJob;
             this.jobController = jobController;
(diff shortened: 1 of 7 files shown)


community/neo4j/src/test/java/schema/IndexPopulationIT.java
```diff
@@ -132,6 +137,42 @@ public void indexCreationDoNotBlockWritesOnOtherLabel() throws ExecutionExceptio
...
+    }
+
+    private void prePopulateDatabase( GraphDatabaseService database, Label testLabel, String propertyName )
+    {
+        for ( int j = 0; j < 100; j++ )
+        {
+            for ( int i = 0; i < 100; i++ )
+            {
+                try ( Transaction transaction = database.beginTx() )
+                {
```
burqen: Should inner loop be inside transaction?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobController.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Collections;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+
+import org.neo4j.scheduler.JobScheduler;
+
+import static org.neo4j.scheduler.JobScheduler.Groups.indexPopulation;
+
+public class IndexPopulationJobController
```
burqen: This class and all public methods can be package-private

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobController.java
```diff
@@ -0,0 +1,91 @@
...
+        return populationJobs;
+    }
+
+    private static class IndexPopulationJobWrapper implements Runnable
+    {
+        private IndexPopulationJob indexPopulationJob;
+        private IndexPopulationJobController jobController;
+
+        IndexPopulationJobWrapper( IndexPopulationJob indexPopulationJob,
+                IndexPopulationJobController jobController )
```
burqen: No need to break up into multiple lines

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobController.java
```diff
@@ -0,0 +1,91 @@
...
+        for ( IndexPopulationJob job : populationJobs )
+        {
+            job.cancel().get();
+        }
+    }
+
+    public void startIndexPopulation( IndexPopulationJob job )
+    {
+        populationJobs.add( job );
+        scheduler.schedule( indexPopulation, new IndexPopulationJobWrapper( job, this ) );
```
burqen: I think there is a race in `IndexingService` between `createIndexes` and `shutdown` where population jobs can be added after `IndexPopulationJobController.stop` is called. Result will be that we can leak population jobs.

It might be that this is guarded for higher up in the stack. But I think this should be guarded for explicitly. Either here or in `IndexingService`.

burqen (changes requested):
Changes look good overall but I think I've found a race.

burqen:
@MishaDemianenko Test failure on IndexingServiceTest is possible to reproduce locally on your branch. Looks related.

community/neo4j/src/test/java/schema/IndexPopulationIT.java
MishaDemianenko: for some reason two for loops that the way of doing a lot of nodes, combined into one

burqen:
Race is handled by letting all transactions finish before shutdown. Moving on.

community/neo4j/src/test/java/schema/IndexPopulationIT.java
burqen: More is always better, right ;)

MishaDemianenko:
@burqen updated

community/neo4j/src/test/java/schema/IndexPopulationIT.java
```diff
@@ -159,16 +159,13 @@ public void shutdownDatabaseDuringIndexPopulations()
 
     private void prePopulateDatabase( GraphDatabaseService database, Label testLabel, String propertyName )
     {
-        for ( int j = 0; j < 100; j++ )
+        for ( int j = 0; j < 10_000; j++ )
         {
-            for ( int i = 0; i < 100; i++ )
+            try ( Transaction transaction = database.beginTx() )
```
burqen: I think batching here would be a good idea. Or just create all of the 10000 nodes in one transaction. Do as you please with this.

burqen (approved):

community/neo4j/src/test/java/schema/IndexPopulationIT.java
MishaDemianenko: i think it does not matter

PR #10247, opened 2017-10-10
Deprecate Present in favor of CompletableFuture.completedFuture

MishaDemianenko:
Mark Present as deprecated only since it's part of public API.
Will be removed as part of next major release.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/helpers/FutureAdapter.java
+++ b/community/kernel/src/main/java/org/neo4j/helpers/FutureAdapter.java
@@ -20,6 +20,7 @@
 package org.neo4j.helpers;
 
 import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


community/kernel/src/main/java/org/neo4j/helpers/FutureAdapter.java
```diff
@@ -63,21 +71,23 @@ public V get()
...
         public V get( long timeout, TimeUnit unit )
         {
             return value;
         }
     }
 
+    /**
+     * This method will be deleted as par tof next major release. Please use {@link CompletableFuture#complete(Object)}
+     * instead.
+     */
```
burqen: "par tof"  -> "part of"

community/kernel/src/main/java/org/neo4j/helpers/FutureAdapter.java
```diff
@@ -43,6 +46,11 @@ public boolean isCancelled()
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * This class will be deleted as par tof next major release. Please use {@link CompletableFuture#complete(Object)}
+     * instead.
+     */
```
burqen: "par tof" -> "part of"

burqen (changes requested):
Approved after fixing comments 🎉

burqen:
@MishaDemianenko `ClusterSeedingIT` and `BackupCoreIT` fail locally for me on your branch. Vanilla `3.4` is fine. Most likely this has nothing to do with you changes, but a rebase might be worth it.

MishaDemianenko:
@burqen duck typing updated

burqen (approved):

PR #10249, opened 2017-10-10
Fix shortest path in with clause bug

### LINKED ISSUE #8201: `shortestPath()` returns strange object when used as a function in an expression
I misapplied `shortestPath` in a `WITH` statement expression and found some very strange behavior in the resulting object; it seems to behave as a collection of Paths that do not behave like Paths _or_ Nodes/Relationships in Cypher, though in the python driver it is properly hydrated as a Path (and the individual collection items as Nodes or Relationships, respectively).
- **Neo4j version** : 3.0.3, 3.0.6, 3.1BETA1
- **Operating system** : Ubuntu 14.04
- **API/Driver** : Web frontend and `neo4j-python-driver`
## Actual Behavior

screenshot 2016-10-17 at 11 29 26 pm
Here is the initial result of the "incorrect" but valid use of `shortestPath`; it appears to be a regular Path. However, if you try to run `Nodes()` on the object, you get this error:
screenshot 2016-10-17 at 11 29 51 pm
So it appears to be a Collection. However, in `python`, this object is returned as a Path, not a Collection. It does support indexing in Cypher, though:
screenshot 2016-10-17 at 11 38 23 pm
In `python`, these objects are correctly hydrated as Nodes or Relationships, so when the Collection is broken up in Cypher, it seems to be a Collection of the Path elements. However, if one attempts to use `Labels()` on one of these elements (or `Type()` if a relationship), we get an error:
screenshot 2016-10-17 at 11 39 46 pm
It is also impossible to access the attributes of these objects, and most Path functions do not work on them (though `Length()` will return 1 for any of them). So the original object returned from `shortestPath()` in an expression operates like a collection, but the elements don't behave in Cypher like anything (though they are returned with enough data that the `python` driver can correctly identify them, turning the Collection into a Path and treating the Collection objects as Nodes or Relationships).
## Expected behavior

Given the increase in function syntax use coming in 3.1, it seems like `shortestPath()` (whose syntax is identical to a function call) should either generate an ordinary path when used in an expression, or should throw an error so that new users know to only use them in a `MATCH` statement.

Lojjs:
Semantic check claimed shortest path returned a list
Unwind didn't work for the Scala iterator returned by allShortestPaths
This PR fix the two bugs above and adds some testing

The new acceptance tests must be ported to CypherComparisonSupport when forward merging

Changelog: Fixes #8201 so shortestPath and allShortestPaths have the right type in WITH clauses

### DIFF
--- a/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathAcceptanceTest.scala
+++ b/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathAcceptanceTest.scala
@@ -31,7 +31,7 @@ class ShortestPathAcceptanceTest extends ExecutionEngineFunSuite with NewPlanner
   var nodeC: Node = _
   var nodeD: Node = _
 
-  override def databaseConfig = Map(GraphDatabaseSettings.forbid_shortestpath_common_nodes -> "false")
+  override def databaseConfig() = Map(GraphDatabaseSettings.forbid_shortestpath_common_nodes -> "false")
 
   override protected def initTest(): Unit = {
     super.initTest()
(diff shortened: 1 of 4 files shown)


sherfert (approved):
Very isolated bugfixes and good unit test. I am not 100 % sure the acceptance tests cover all interesting cases (e.g. shortestPaths in a list, in a map, etc.), but it is definitely a step forward.

spacecowboy:
@Lojjs just FYI, the changelog will not make `#8201` into a link. You need to do that explicitly in the message if you want to link to the original issue.

PR #10252, opened 2017-10-10
 Added UnionSlottedPipe

SaschaPeukert:


enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/SlottedPipeBuilder.scala
```diff
@@ -355,12 +355,22 @@ class SlottedPipeBuilder(fallback: PipeBuilder,
       case Union(_, _) =>
         val lhsInfo = pipelines(lhs.id)
         val rhsInfo = pipelines(rhs.id)
-
-        UnionSlottedPipe(lhs, rhs, lhsInfo, rhsInfo, pipeline, computeUnionSlots(lhsInfo, rhsInfo, pipeline))(id = id)
+        //val lhsOverlap = overlaps(pipeline, lhsInfo) // since pipeline is constructed from lhs the overlap will always be true
```
pontusmelke: This is not true, `MATCH (n) RETURN n UNION RETURN 42 AS n`
will (at least in the future) have a long slot for n in the lhs but a refslot in the outgoing pipeline

pontusmelke: We will need to have something like

```
val lhsOverlap = overlaps(pipeline, lhsInfo)
val rhsOverlap = overlaps(pipeline, rhsInfo)
if (lhsOverlap && rhsOverlap) FastUnionSlottedPipe(....)
else if (lhsOverlap) FastLeftUnionSlottedPipe(...)
else if (rhsOverlap) FastRightUnionSlottedPipe(...)
else UnionSlottedPipe(...)
```

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/UnionSlottedPipe.scala
```diff
@@ -24,19 +24,10 @@ import org.neo4j.cypher.internal.compatibility.v3_4.runtime.slotted.PrimitiveExe
 import org.neo4j.cypher.internal.compatibility.v3_4.runtime.{ExecutionContext, PipelineInformation}
 import org.neo4j.cypher.internal.v3_4.logical.plans.LogicalPlanId
 
-case class UnionSlottedPipe(lhs: Pipe, rhs: Pipe, lhsInfo: PipelineInformation, rhsInfo: PipelineInformation,
+case class UnionSlottedPipe(lhs: Pipe, rhs: Pipe, rhsInfo: PipelineInformation,
```
pontusmelke: UnionSlottedPipe needs both lhsInfo and rhsInfo

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/UnionSlottedPipe.scala
```diff
@@ -45,12 +36,8 @@ case class UnionSlottedPipe(lhs: Pipe, rhs: Pipe, lhsInfo: PipelineInformation,
     new Iterator[ExecutionContext] {
       override def hasNext: Boolean = left.hasNext || right.hasNext
 
-      override def next(): ExecutionContext = if (left.hasNext) {
-        val in = left.next()
-        val out = PrimitiveExecutionContext(unionInfo)
-        mapSlots.foreach(f => f(in, out, lhsInfo, state))
-        out
-      } else {
+      override def next(): ExecutionContext = if (left.hasNext) left.next()
```
pontusmelke: This is wrong in the general case the left-hand side also might need to be transformed

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/SlottedPipeBuilder.scala
```diff
@@ -351,7 +352,79 @@ class SlottedPipeBuilder(fallback: PipeBuilder,
...
+        else
+          UnionSlottedPipe(lhs, rhs, rhsInfo, pipeline, computeUnionSlots(lhsInfo, rhsInfo, pipeline))(id = id)
+
       case _ => throw new CantCompileQueryException(s"Unsupported logical plan operator: $plan")
     }
   }
+
+  private def overlaps(source: PipelineInformation, toCheck: PipelineInformation):Boolean = {
+    source.mapSlot {
+      case (k, s) => s == toCheck.get(k).get
```
henriknyman: I wonder if == is a too strict condition here? E.g. if one side has the type of a RefSlot as a CTString and the other as CTAny they could still be used as is with the fast Union? Also nullable should not matter.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/SlottedPipeBuilder.scala
```diff
@@ -351,7 +352,79 @@ class SlottedPipeBuilder(fallback: PipeBuilder,
...
+    //find columns where output is a reference slot but where the input is a long slot
+    def findColums(input: PipelineInformation) = {
+      val slots = out.mapSlot {
+        case (k, slot: RefSlot) => input.get(k).get match {
+          case ls: LongSlot => Some(ls)
+          case _ => None
+        }
+        case _ => None
+      }.flatten.toIndexedSeq
+      slots.zip(createProjectionsForResult(slots.map(_.name), input).map(_._2)).toMap
```
henriknyman: Maybe this could be broken up in some named vals for readability

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/SlottedPipeBuilder.scala
```diff
@@ -351,7 +352,79 @@ class SlottedPipeBuilder(fallback: PipeBuilder,
...
+    }
+    //Compute how to convert input to output pipeline
+    val expressions = findColums(lhsInfo) ++ findColums(rhsInfo)
+    val mapSlots: Iterable[(ExecutionContext, ExecutionContext, PipelineInformation, QueryState) => Unit] = out.mapSlot {
+      case (k, v: LongSlot) => (in, out, info, _) =>
```
thobe: In what way is the `info` passed in to these functions different from the `rhsInfo` we already have in scope?

Looking at the code it certainly looks like it would be the same. We pass in the same `rhsInfo` to both the `computeUnionSlots` function as we store in the `UnionSlottedPipe` object, and the `rhsInfo` stored in `UnionSlottedPipe` is passed unmodified to the functions that are called `f` there, where it ends up as `info` here.

If we instead just use `rhsInfo` here, we don't have to have `rhsInfo` available in `UnionSlottedPipe`. It also means we can move more of the decision making that happens here to compiletime. Essentially we don't need the `PipelineInformation` object referenced at all at runtime, since we can do the lookups when `computeUnionSlots` runs, and have the mappers it returns only know the actual slot offsets it is mapping (and not how to look up those slots - because that has already been done.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/UnionSlottedPipe.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.neo4j.cypher.internal.compatibility.v3_4.runtime.slotted.pipes
+
+import org.neo4j.cypher.internal.compatibility.v3_4.runtime.pipes.{Pipe, QueryState}
+import org.neo4j.cypher.internal.compatibility.v3_4.runtime.slotted.PrimitiveExecutionContext
+import org.neo4j.cypher.internal.compatibility.v3_4.runtime.{ExecutionContext, PipelineInformation}
+import org.neo4j.cypher.internal.v3_4.logical.plans.LogicalPlanId
+
+case class UnionSlottedPipe(lhs: Pipe, rhs: Pipe,
+                            lhsMapping: (ExecutionContext, QueryState) => ExecutionContext,
```
systay: I really like this solution. Great stuff!

If I could wish for one thing - maybe we should create a trait or type alias instead of writing out ` (ExecutionContext, QueryState) => ExecutionContext`?

enterprise/cypher/slotted-runtime/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/UnionSlottedPipeTest.scala
```diff
@@ -0,0 +1,226 @@
...
+    // Then
+    result should equal(
+      List(Map("x" -> edgeValue(42)), Map("x" -> edgeValue(43))))
+  }
+
+  test("should handle one long slot and one relationship slot") {
+    // Given
+    val lhsInfo = PipelineInformation.empty.newLong("x", nullable = false, CTNode)
+    val rhsInfo = PipelineInformation.empty.newLong("x", nullable = false, CTRelationship)
+    val out = PipelineInformation.empty.newReference("x", nullable = false, CTAny)
```
systay, line 144: <3 So cool...

pontusmelke:
Think we fixed all comments. Happy? @henriknyman

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/UnionSlottedPipe.scala
```diff
@@ -37,82 +35,8 @@ case class UnionSlottedPipe(lhs: Pipe, rhs: Pipe,
...
-                               (val id: LogicalPlanId = LogicalPlanId.DEFAULT) extends Pipe {
-  // This pipe is only applicable if the ordering of the variables on the lhs equals the ordering on the rhs
-
-  protected def internalCreateResults(state: QueryState): Iterator[ExecutionContext] = {
-    val left = lhs.createResults(state)
-    val right = rhs.createResults(state)
-    new Iterator[ExecutionContext] {
-      override def hasNext: Boolean = left.hasNext || right.hasNext
-      override def next(): ExecutionContext = if (left.hasNext) left.next()
-      else right.next()
```
henriknyman: It seems to me that these specialized "Fast" pipes would be faster in the applicable cases, since it avoids a lot of row copying. It also seems to me to be the most common cases. Did you delete them mainly to simplify the code? Did you benchmark any differences? The same RowMapping optimization could also be applied to these specializations.

henriknyman (commented):
It seems to me that the specialized "Fast" pipes that you had in a previous commit would be faster in the applicable cases, since it avoids a lot of row copying. It also seems to me to be the most common cases. Did you delete them mainly to simplify the code? Did you benchmark any differences? The same RowMapping optimization could also be applied to these specializations.

henriknyman:
Oh, sorry, I misread the `computeUnionMapping` method because of a bit weird indentation. You are just forwarding the row without copy if possible, just in a more generalized solution now. The simplifiaction to the code base is probably worth the extra function application (which is easily inlineable). Very nice.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/SlottedPipeBuilder.scala
```diff
@@ -351,7 +354,97 @@ class SlottedPipeBuilder(fallback: PipeBuilder,
...
+      //longslot
+      case (k, s: LongSlot) => s == in.get(k).get
+      //For refslot is is ok that types etc differs, just make sure
+      //they have the same offset
+      case (k, s: RefSlot) => s.offset == in.get(k).get.offset
+    }.forall(_ == true)
+
+    //If we overlap we can just pass the result right throug
+    if (overlaps) (incoming: ExecutionContext, _: QueryState) => incoming
+    else {
```
henriknyman, line 410: This block is missing indentation

henriknyman (approved):

PR #10255, opened 2017-10-11
Renamed all remaining occurrences of EnterpriseInterpreted to Slotted

SaschaPeukert:


enterprise/cypher/compatibility-spec-suite/src/test/java/cypher/CompatibilitySpecSuiteTest.java
```diff
@@ -149,7 +149,7 @@ public static void teardown()
             tags = { "~@pending" },
             strict = true
     )
-    public static class CostInterpretedEnterprise extends Base
+    public static class CostSlotted extends Base
```
craigtaverner, line 152: Were there no references to this? If so, perhaps it can be deleted?

craigtaverner (approved):

enterprise/cypher/compatibility-spec-suite/src/test/java/cypher/CompatibilitySpecSuiteTest.java
SaschaPeukert: Don't we need that to specifically run Compatibility tests on Slotted?

thobe: The "reference" is through the fact that this is a suite of tests, so all nested classes are taken as test classes.

PR #10256, opened 2017-10-11
Transitive closure in where clauses

pontusmelke:
For queries like

```
MATCH (n)-->(m) WHERE n.prop = m.prop AND m.prop = 42 RETURN *
```

we can safely rewrite the query to

```
MATCH (n)-->(m) WHERE n.prop = 42 AND m.prop = 42 RETURN *
```

Also rewrites queries of the form

```
MATCH (n)-->(m) WHERE m.prop = n.prop AND m.prop = 42 RETURN *
```

which is turned into

```
MATCH (n)-->(m) WHERE m.prop = 42 AND n.prop = 42 RETURN *
```

### DIFF
--- a/community/cypher/cypher-compiler-3.4/src/test/scala/org/neo4j/cypher/internal/compiler/v3_4/ast/rewriters/TransitiveClosureTest.scala
+++ b/community/cypher/cypher-compiler-3.4/src/test/scala/org/neo4j/cypher/internal/compiler/v3_4/ast/rewriters/TransitiveClosureTest.scala
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher.internal.compiler.v3_4.ast.rewriters
+
+import org.neo4j.cypher.internal.compiler.v3_4.phases.LogicalPlanState
+import org.neo4j.cypher.internal.compiler.v3_4.planner.AstRewritingTestSupport
+import org.neo4j.cypher.internal.compiler.v3_4.test_helpers.ContextHelper
+import org.neo4j.cypher.internal.frontend.v3_4.ast.Query
+import org.neo4j.cypher.internal.frontend.v3_4.ast.rewriters.{rewriteEqualityToInPredicate, transitiveClosure}
+import org.neo4j.cypher.internal.util.v3_4.test_helpers.CypherFunSuite
+
+class TransitiveClosureTest extends CypherFunSuite with AstRewritingTestSupport {
+
+  test("MATCH (a)-->(b) WHERE a.prop = b.prop AND b.prop = 42") {
+    shouldRewrite(
+      "MATCH (a)-->(b) WHERE a.prop = b.prop AND b.prop = 42",
+      "MATCH (a)-->(b) WHERE a.prop = 42 AND b.prop = 42")
+  }
+
+  test("MATCH (a)-->(b) WHERE b.prop = a.prop AND b.prop = 42") {
+    shouldRewrite(
+      "MATCH (a)-->(b) WHERE a.prop = b.prop AND b.prop = 42",
+      "MATCH (a)-->(b) WHERE a.prop = 42 AND b.prop = 42")
+  }
+
+  test("MATCH (a)-->(b) WHERE a.prop = b.prop OR b.prop = 42") {
+    shouldNotRewrite("MATCH (a)-->(b) WHERE a.prop = b.prop OR b.prop = 42")
+  }
+
+  test("MATCH (a)-->(b) WHERE a.prop = b.prop AND b.prop = b.prop2 AND b.prop2 = 42") {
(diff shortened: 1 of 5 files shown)


community/cypher/cypher-compiler-3.4/src/test/scala/org/neo4j/cypher/internal/compiler/v3_4/ast/rewriters/TransitiveClosureTest.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  test("MATCH (a)-->(b) WHERE a.prop = b.prop AND b.prop = 42") {
+    shouldRewrite(
+      "MATCH (a)-->(b) WHERE a.prop = b.prop AND b.prop = 42",
+      "MATCH (a)-->(b) WHERE a.prop = 42 AND b.prop = 42")
+  }
+
+  test("MATCH (a)-->(b) WHERE b.prop = a.prop AND b.prop = 42") {
+    shouldRewrite(
+      "MATCH (a)-->(b) WHERE a.prop = b.prop AND b.prop = 42",
```
craigtaverner: Test title differs in equality order, but actual test content does not and is identical to previous test.

craigtaverner (approved):
Cool solution. One test might like a small change.

community/cypher/cypher-compiler-3.4/src/test/scala/org/neo4j/cypher/internal/compiler/v3_4/ast/rewriters/TransitiveClosureTest.scala
pontusmelke: You might almost think I just copy-pasted without thinking about it...

community/cypher/frontend-3.4/src/main/scala/org/neo4j/cypher/internal/frontend/v3_4/ast/rewriters/transitiveClosure.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    def apply(that: AnyRef): AnyRef = instance.apply(that)
+
+    private val instance: Rewriter = topDown(Rewriter.lift {
+      case where@Where(expression) =>
+        recurse(where, w => w.endoRewrite(whereRewriter))
+    })
+
+    @tailrec
+    private def recurse(in: Where, f: Where => Where): Where = {
```
systay: Isn't this what `fixedPoint` already does?

community/cypher/frontend-3.4/src/main/scala/org/neo4j/cypher/internal/frontend/v3_4/ast/rewriters/transitiveClosure.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  override def instance(ignored: BaseContext): Rewriter = transitiveClosureRewriter
+
+  override def postConditions: Set[Condition] = Set.empty
+
+  private case object transitiveClosureRewriter extends Rewriter {
+
+    def apply(that: AnyRef): AnyRef = instance.apply(that)
+
+    private val instance: Rewriter = topDown(Rewriter.lift {
```
systay: Is there a reason we do this `topDown` instead of `bottomUp`? I was under the impression that `bottomUp` is preferable from a performance perspective, if either work.

community/cypher/frontend-3.4/src/main/scala/org/neo4j/cypher/internal/frontend/v3_4/ast/rewriters/transitiveClosure.scala
```diff
@@ -0,0 +1,105 @@
...
+      case where@Where(expression) =>
+        recurse(where, w => w.endoRewrite(whereRewriter))
```
systay: For other AST rewriters, we do the fixedPoint recursion on the outside, so rewriters can help each other. We could do that by moving this into the lateAstRewriting phase. Is that possible?

community/cypher/frontend-3.4/src/main/scala/org/neo4j/cypher/internal/frontend/v3_4/ast/rewriters/transitiveClosure.scala
pontusmelke: Probably I was looking for something called `iterateUntilConvergence` but it was probably renamed years ago.

community/cypher/frontend-3.4/src/main/scala/org/neo4j/cypher/internal/frontend/v3_4/ast/rewriters/transitiveClosure.scala
pontusmelke: It is probably possible, however I wanted to do the rewrite before we rewrite `Equals` to `In`.

systay (changes requested):
Really cool! Just a few tweaks.

systay:
I would also like to point out that this does not solve another class of transitive closures. This query:
```
MATCH (a), (b) WHERE a = b RETURN a, b
```

is currently solved with this plan:

```
+-----------------+----------------+-----------+-------+
| Operator        | Estimated Rows | Variables | Other |
+-----------------+----------------+-----------+-------+
| +ProduceResults |              0 | a, b      |       |
| |               +----------------+-----------+-------+
| +ValueHashJoin  |              0 | a -- b    | a = b |
| |\              +----------------+-----------+-------+
| | +AllNodesScan |              1 | b         |       |
| |               +----------------+-----------+-------+
| +AllNodesScan   |              1 | a         |       |
+-----------------+----------------+-----------+-------+
```

We could be plan this better than that. This PR is a great step forward. Just wanted to remind us that there is more transitive stuff we can do.

community/cypher/frontend-3.4/src/main/scala/org/neo4j/cypher/internal/frontend/v3_4/ast/rewriters/transitiveClosure.scala
systay: Ah, makes sense. :+1:

pontusmelke:
@systay Agree, handling th

pontusmelke:
@systay I agree but that is not really transitive closure. We should probably just rewrite equality of nodes or relationships to  projections, i.e.
`a = b => WITH b AS a`?

PR #10262, opened 2017-10-11
Better error message when not is parsed as a function.

### LINKED ISSUE #2556: NOT is not parsed properly when nested inside expressions without parentheses
To reproduce just try to parse:

RETURN true = not(42 = 32)

it will fail since NOT is parsed as a function invocation rather than the boolean negation operator.

sherfert:
changelog: Fixes #2556 so that a sensible error message is generated when "not" is parsed as a function.


Lojjs (approved):
Looks fine to me, simple and clean solution.

PR #10263, opened 2017-10-11
Test multiple index iterators open on Kernel API level

burqen:
Verify behaviour of multiple index query iterators on same index in same transaction.

```
Query type:
- Exact
- Exist
- Range
For:
- Single number
- Single string
- Composite number
- Composite string
With: 
- Nested iterators
- Interleaved iterators
- Iterators exhausted back to back
```

### DIFF
--- a/community/lucene-index/src/test/java/org/neo4j/kernel/impl/api/OperationsFacadeSchemaIndexIteratorTest.java
+++ b/community/lucene-index/src/test/java/org/neo4j/kernel/impl/api/OperationsFacadeSchemaIndexIteratorTest.java
@@ -54,6 +54,7 @@ import org.neo4j.test.rule.EmbeddedDatabaseRule;
 import org.neo4j.test.rule.RandomRule;
 
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.collection.IsIterableContainingInAnyOrder.containsInAnyOrder;
 import static org.junit.Assert.assertThat;
 
 /**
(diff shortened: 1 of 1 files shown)


community/lucene-index/src/test/java/org/neo4j/kernel/impl/api/OperationsFacadeSchemaIndexIteratorTest.java
```diff
@@ -0,0 +1,819 @@
...
+                expected.add( i );
+            }
+            assertSameContent( actual, expected );
+        }
+
+        void assertSameContent( List<Long> actual, List<Long> expected )
+        {
+            actual.sort( Long::compareTo );
+            expected.sort( Long::compareTo );
+            assertThat( actual, is( expected ) );
```
klaren: You could use `IsIterableContainingInAnyOrder.containsInAnyOrder` instead.

community/lucene-index/src/test/java/org/neo4j/kernel/impl/api/OperationsFacadeSchemaIndexIteratorTest.java
```diff
@@ -0,0 +1,819 @@
...
+        {
+            // when
+            ReadOperations readOperations = statement.readOperations();
+            PrimitiveLongIterator iter1 = indexCoordinator.queryExists( readOperations );
+            List<Long> actual1 = new ArrayList<>();
+            PrimitiveLongIterator iter2 = indexCoordinator.queryExists( readOperations );
+            List<Long> actual2 = new ArrayList<>();
+
+            // Interleave
+            while ( iter1.hasNext() && iter2.hasNext() )
```
klaren: I would extract the random order exhaustion to a method, line 298-318

klaren (approved):
Looks good! :+1: 

I have two small suggestions, take it or leave it :)

community/lucene-index/src/test/java/org/neo4j/kernel/impl/api/OperationsFacadeSchemaIndexIteratorTest.java
```diff
@@ -0,0 +1,819 @@
...
+            }
+
+            // then
+            indexCoordinator.assertRangeResult( actual1 );
+            indexCoordinator.assertRangeResult( actual2 );
+            tx.success();
+        }
+    }
+
+    // multipleIteratorsNotNestedExists
```
klaren: Leftovers

PR #10265, opened 2017-10-11
Return correct results for parameter maps in predicates

### LINKED ISSUE #1525: MATCH with properties inconsistency in REST API
Hi,
I was trying the new MATCH with properties cypher feature in Neo4j 2.0 RC1 using the REST API. 

I noticed that if I explicitly hard coded the properties in the cypher query it worked as expected. But I passed the properties as a parameter it doesn't work. This behaviour seems to be inconsistent with the way other cypher queries and params work.

A very basic program in Python to illustrate the issue and the output is [here](https://gist.github.com/ducky427/7579861).

sherfert:
changelog: Fixes #1525. Comparing nodes and relationships to maps
in predicates returns correct results now.

### DIFF
--- a/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
+++ b/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
@@ -23,6 +23,7 @@ import org.neo4j.cypher._
 import org.neo4j.cypher.internal.compiler.v3_1.commands.expressions.PathImpl
 import org.neo4j.graphdb._
 import org.neo4j.helpers.collection.Iterators.single
+import org.neo4j.helpers.collection.MapUtil.map
 
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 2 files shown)


pontusmelke:
I don't think it should allowed to do compare a node and map for equality

systay:
I'm with @pontusmelke on this one. Don't see why we would want to do this.

thobe:
This behaviour is **not** what was desired in #1525. The desired behaviour in that ticket it to be able to match using all properties (in the `allProps` map parameter) when using a query like this: `MATCH (n{allProps}) RETURN n`.

The behaviour implemented in this pull request is wrong, and in violation of the specification of `=`.

Since any change addressing #1525 will be substantially different from what is done here, and since what has been implemented here is not going to get merged, I'm closing this PR.

craigtaverner:
The behaviour coded in this feature was decided on in a discussion between Tobias and Chris in https://github.com/neo4j/neo4j/pull/1542, not in #1525. The fact that this PR was associated with #1525 is misleading, as the desired behavior in that card was rejected in that card, but instead a discussion ensued in https://github.com/neo4j/neo4j/pull/1542 where the behaviour coded for this PR was in fact explicitly decided on by Tobias and Chris. Of course, that decision could be revoked.

PR #10267, opened 2017-10-11
Enable metrics by default. Make reports csv files rotatable.

MishaDemianenko:
Previously metrics were disabled by default, this PR changes that
and enable them by default.
Also previously report csv files were not limited in size in any way
and that cause problems since over time they become quite big.
This PR will allow configure csv file rotation in a way how rotation
is configurable already for any kind of log file.

### DIFF
--- a/enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
+++ b/enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
@@ -28,12 +28,17 @@ import org.neo4j.graphdb.config.Setting;
 import org.neo4j.helpers.HostnamePort;
 
 import static org.neo4j.kernel.configuration.Settings.BOOLEAN;
+import static org.neo4j.kernel.configuration.Settings.BYTES;
 import static org.neo4j.kernel.configuration.Settings.DURATION;
 import static org.neo4j.kernel.configuration.Settings.FALSE;
 import static org.neo4j.kernel.configuration.Settings.HOSTNAME_PORT;
+import static org.neo4j.kernel.configuration.Settings.INTEGER;
 import static org.neo4j.kernel.configuration.Settings.STRING;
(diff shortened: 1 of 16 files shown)


enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
```diff
@@ -130,6 +135,15 @@
                   "the CSV files." )
     public static Setting<Duration> csvInterval = setting( "metrics.csv.interval", DURATION, "3s" );
 
+    @Description( "The file size in bytes at which the csv files will auto-rotate. If set to zero then no " +
+            "rotation will occur. Accepts a binary suffix `k`, `m` or `g`." )
+    public static final Setting<Long> csvRotationThreshold = buildSetting( "metrics.csv.rotation.size",
+            BYTES, "20m" ).constraint( range( 0L, Long.MAX_VALUE ) ).build();
+
+    @Description( "Maximum number of history files for the csv files." )
+    public static final Setting<Integer> csvMaxArchives = buildSetting( "metrics.csv.rotation.keep_number",
```
tinwelint: Roughly how long will these defaults keep metrics before being rotated out completely?

A quick calculation, assuming 100 bytes per logged line, yields roughly 7 days worth of data in 20MB. So ~50 days worth of data in there.

Even given that I see no reason for having the rotation threshold that low. 20MB is not that much, I think might as well be 100 or more imho. Most editors can handle 100 MB or a couple, can't they?

tinwelint (approved):
Looks good. One concern about the default rotation threshold being a bit small.

enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
MishaDemianenko: sure, let's do them 100m by default for now if that will be too big since we can have a lot of metrics files and all of them will consume space we can change it later on

PR #10269, opened 2017-10-12
Ability to provide import arguments as -f argument

tinwelint:
Doing imports with lots of arguments and input files may lead to a long command line to provide in the terminal. This can now be simplified by placing all arguments in a file and pointing the importer to that file using the `-f` argument.


ragadeeshu (approved):

tinwelint:
@ragadeeshu This is such a small chance so it could be for 3.3 even, wdyt?

ragadeeshu:
Sure

tinwelint:
Did it

PR #10271, opened 2017-10-12
Update fields in legacy indexes when removing keys.

ragadeeshu:
Not doing this could potentially lead to incorrect index query results
after removal of an indexed key.

cl: Fix a bug where an entity could be temporarily excluded from some index query results if another key indexed on the same entity were removed.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/IndexType.java
+++ b/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/IndexType.java
@@ -374,6 +374,7 @@ public abstract class IndexType
         else
         {
             removeFieldsFromDocument( document, key, value );
+            restoreSortFields( document );
         }
     }
 
--- a/community/lucene-index/src/test/java/org/neo4j/index/LegacyIndexTest.java
+++ b/community/lucene-index/src/test/java/org/neo4j/index/LegacyIndexTest.java
@@ -30,20 +30,23 @@ import org.neo4j.graphdb.Relationship;
 import org.neo4j.graphdb.RelationshipType;
 import org.neo4j.graphdb.Transaction;
 import org.neo4j.graphdb.index.Index;
+import org.neo4j.graphdb.index.IndexHits;
 import org.neo4j.graphdb.index.RelationshipIndex;
+import org.neo4j.index.lucene.QueryContext;
 import org.neo4j.index.lucene.ValueContext;
 import org.neo4j.test.rule.DatabaseRule;
-import org.neo4j.test.rule.ImpermanentDatabaseRule;
+import org.neo4j.test.rule.EmbeddedDatabaseRule;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 public class LegacyIndexTest
 {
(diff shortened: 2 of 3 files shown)


MishaDemianenko:
It looks like this PR needs a bit more broad description and test

tinwelint (changes requested):
Some testing should be added for this.

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/IndexType.java
```diff
@@ -374,6 +374,7 @@ final void removeFromDocument( Document document, String key, Object value )
         else
         {
             removeFieldsFromDocument( document, key, value );
+            restoreSortFields( document );
```
tinwelint, line 377: Is this a fix for something unrelated to the original PR changes?

community/lucene-index/src/test/java/org/neo4j/index/LegacyIndexTest.java
```diff
@@ -139,6 +142,114 @@ public void shouldThrowIllegalArgumentChangingTypeOfFieldOnNodeIndex()
     }
 
     @Test
+    public void shouldBeAbleToAddNodesAfterRemovalOfKey()
```
tinwelint, line 145: This test is green w/o your changes, is that expected? How do we know that your changes solves the issue?

community/lucene-index/src/test/java/org/neo4j/index/LegacyIndexTest.java
```diff
@@ -139,6 +142,114 @@ public void shouldThrowIllegalArgumentChangingTypeOfFieldOnNodeIndex()
...
+
+        db.shutdownAndKeepStore();
+        db.getGraphDatabaseAPI();
+
+        //should be able to add more stuff to the index
+        try ( Transaction tx = db.beginTx() )
+        {
+            Node node = db.createNode();
+            Index<Node> nodeIndex = db.index().forNodes( indexName );
+            nodeIndex.add( node, "key", "hej" );
```
tinwelint, line 175: did you mean to use `keydelete` here as key?

community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/legacy/IndexType.java
ragadeeshu: Not unrelated. This is currently the only _known_ cause for the case handled in the first commit. One of the tests in this PR tests for this bug explicitly.

community/lucene-index/src/test/java/org/neo4j/index/LegacyIndexTest.java
ragadeeshu: This test is made green by the first commit in this PR.

community/lucene-index/src/test/java/org/neo4j/index/LegacyIndexTest.java
ragadeeshu: No, I want to verify the that the indextype of `key` is correct.

community/lucene-index/src/test/java/org/neo4j/index/LegacyIndexTest.java
tinwelint: Missed the change to the db rule. All good here.

tinwelint (approved):

PR #10274, opened 2017-10-12
Fixed bug in shortestPath with predicate depending on multiple patterns

craigtaverner:
If the predicate had dependencies on path variables from both shortestPath
patterns, then we were applying the predicate to both, even it is was only
associated with the one. This is fixed by using the same `hasDependenciesMet`
function used for normal pattern-predicate association.


community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathEdgeCasesAcceptanceTest.scala
```diff
@@ -0,0 +1,148 @@
...
+import org.neo4j.cypher.{ExecutionEngineFunSuite, NewPlannerTestSupport}
+import org.neo4j.graphdb.Node
+
+class ShortestPathEdgeCasesAcceptanceTest extends ExecutionEngineFunSuite with NewPlannerTestSupport {
+
+  test("GH #5803 query should work with shortest path") {
+    def createTestGraph() = {
+      // First create some background graph to get the stats right
+      var prev: Node = null
+      0 to 16 foreach { row =>
```
pontusmelke: Not important: 

```
(0 to 16).map(_ => createNode()).sliding(2).foreach {
  case Seq(n1,n2) => relate(n1,n2)
}
```

community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathEdgeCasesAcceptanceTest.scala
```diff
@@ -0,0 +1,148 @@
...
+                    |create (_31814)-[:`SE`]->(_31820)
+                    |create (_31814)-[:`SE`]->(_31813)
+                    |create (_31814)-[:`SE`]->(_31806)
+                    |create (_31815)-[:`SE`]->(_31810)
+                    |create (_31816)-[:`SE`]->(_31815)
+                    |create (_31817)-[:`SE`]->(_31816)
+                    |create (_31818)-[:`SE`]->(_31817)
+                    |create (_31819)-[:`SE`]->(_31818)
+                    |create (_31820)-[:`SE`]->(_31813)""".stripMargin
+      eengine.execute(query, Map.empty[String, Any])
```
pontusmelke: maybe use the GDS directly

```
graph.execute(query)
```

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/planner/logical/steps/planShortestPaths.scala
```diff
@@ -36,8 +36,12 @@ case object planShortestPaths {
            (implicit context: LogicalPlanningContext): LogicalPlan = {
 
     val variables = Set(shortestPaths.name, Some(shortestPaths.rel.name)).flatten
+    def predicateAppliesToShortestPath(p: Predicate) =
```
pontusmelke, line 39: Can we have a comment here explaining the logic of the below rows? I don't think I fully understand it now and future me will surely be clueless.

pontusmelke (changes requested):
Nice. Had some very minor comments.

pontusmelke (approved):

PR #10275, opened 2017-10-12
Removed name of slots and changed projection to produce LongSlots

SaschaPeukert:


henriknyman (dismissed):
Removing name from slots makes for a horrible debugging experience

systay:
We removed the name from slot because we now can have two different variables sharing the same slot. The name information is still available in the `PipelineInformation`.

henriknyman:
@systay I understand it was because of the multiple aliased names, but I still think we need to at least project this information in the implementation of  `PrimitiveExecutionContext.iterator` (that is used for debugging) or I strongly believe it will slow down development too much. Also currently there is no backward map from slots to names so the information is not easily available. 

We could either keep the name in the slot as a Set or Seq or add a separate map from slot to set of names. (Even keeping the first name would be better than not fixing this since it would work in >95% of cases).

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/PipelineInformation.scala
```diff
@@ -134,38 +139,51 @@ object PipelineInformation {
...
-    case LongSlot(_, nullable, typ, _) => newLong(key, nullable, typ)
-    case RefSlot(_, nullable, typ, _) => newReference(key, nullable, typ)
+    case LongSlot(_, nullable, typ) => newLong(key, nullable, typ)
+    case RefSlot(_, nullable, typ) => newReference(key, nullable, typ)
   }
 
-  def seedClone(): PipelineInformation = {
-    new PipelineInformation(this.slots, numberOfLongs, numberOfReferences)
+  def breakPipelineAndClone(): PipelineInformation = {
+    new PipelineInformation(this.slots.clone(), numberOfLongs, numberOfReferences)
```
henriknyman: We are losing alias information here

henriknyman:
We still need to deduplicate aliased slots in some places. I have a WIP-commit for that which I will try to finish off. After that somebody else should review again since there are a lot of changes.

systay (changes requested):
Nice changes! Could you please add a few tests to make sure the bugs you have fixed don't reappear in the future?

PR #10279, opened 2017-10-13
Kernel API implementation, Chapter I: Reads

fickludd:
Reads.

This PR supplies all read functionality of the Kernel API, with the exception of parallel scans. This includes access to nodes, relationships, traversals, all properties, and all index queries.

This PR does not consider transaction state. That functionality will be added incrementally as we start consuming the Kernel API. 

Note that some changes are included to the API itself, to align it more with the existing APIs, which will make swapping APIs easier. Also, API level tests are implemented in the API module, to make sure that we are not using internals in testing, and therefore also allow us to run the same tests against both NeoLE and Neo4j.


enterprise/runtime/neole/src/test/java/org/neo4j/internal/store/prototype/neole/NeoLETestSupport.java
```diff
@@ -0,0 +1,85 @@
...
+                graphDb.shutdown();
+            }
+        }
+        this.storeDir = storeDir;
+    }
+
+    @Override
+    public void beforeEachTest()
+    {
+        assumeThat( "x86_64", equalTo( System.getProperty( "os.arch" ) ) );
```
thobe: This is backwards, it should be 

    assumeThat( System.getProperty( "os.arch" ), equalTo( "x86_64" ) );

thobe:
For some reason assumption violations are reported as test failures. But there seems to be other failures in those same test cases as well, but for some reason there are no stack traces (at least not that I can find), making it really hard to discern what went wrong, since of course it works on my machine...

Although, perhaps we should just remove NeoLE? It's a prototype and not part of the product. Sure it'll serve a purpose as a guide for what we would like to refactor towards later on, but surely we should be able to dig it out from history then...

pontusmelke:
I think I'm 👍 with what @thobe is saying. We are already seeing that we have problems keeping the prototype up-to-date, better to have it live outside the main code base then.

Squash e8cf11b and c8cfaaa would be nice.

fickludd:
Ok, i have removed NeoLE completely from the product now. We can bring it back when we see the need, in Neo4j or some other repo. Time will tell.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexQuery.java
```diff
@@ -206,10 +200,16 @@ public IndexQueryType type()
...
-        public boolean test( Value value )
+        public boolean acceptsValue( Value value )
         {
             return value != null && value != Values.NO_VALUE;
         }
+
+        @Override
+        public boolean acceptsValueAt( PropertyCursor property )
+        {
+            return true;
```
burqen, line 211: Why do we not use `return acceptsValue( property.propertyValue() )`?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexCursorProgressor.java
```diff
@@ -0,0 +1,57 @@
...
+import org.neo4j.internal.kernel.api.LabelSet;
+import org.neo4j.values.storable.Value;
+
+public interface IndexCursorProgressor
+{
+    boolean next();
+
+    void close();
+
+    interface NodeValueCursor
```
MishaDemianenko: not sure why all those interfaces live here

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexCursorProgressor.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.newapi;
+
+import org.neo4j.internal.kernel.api.LabelSet;
+import org.neo4j.values.storable.Value;
+
+public interface IndexCursorProgressor
```
MishaDemianenko: maybe we can make it AutoCloseable?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Lazy.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.function.Supplier;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Encapsulates a double checked locking pattern for lazy initialization.
+ * TODO: move this to some util place or something.
+ */
+public final class Lazy<T>
```
MishaDemianenko: No need for this class: we already have Suppliers::lazySingleton, or LazySingleReference, most probably something else as well.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -0,0 +1,189 @@
...
+{
+    private final Read read;
+    private final RecordCursor<DynamicRecord> labelCursor;
+    private PageCursor pageCursor;
+    private long next;
+    private long highMark;
+
+    NodeCursor( Read read )
+    {
+        super( -1 );
```
MishaDemianenko: NO_ID

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -0,0 +1,189 @@
...
+        {
+            pageCursor.close();
+            pageCursor = null;
+        }
+        reset();
+    }
+
+    private void reset()
+    {
+        setId( next = NO_ID );
```
MishaDemianenko: i do not think we need to save this line better make it more readable

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
```diff
@@ -0,0 +1,402 @@
...
+    private long next;
+    private int block;
+    ByteBuffer buffer;
+    private PageCursor page;
+    private PageCursor stringPage;
+    private PageCursor arrayPage;
+
+    public PropertyCursor( Read read )
+    {
+        super( -1 );
```
MishaDemianenko: NO_ID

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Store.java
```diff
@@ -0,0 +1,336 @@
...
+import org.neo4j.storageengine.api.schema.LabelScanReader;
+import org.neo4j.string.UTF8;
+import org.neo4j.values.storable.ArrayValue;
+import org.neo4j.values.storable.TextValue;
+import org.neo4j.values.storable.Values;
+
+import static org.neo4j.kernel.impl.store.record.AbstractBaseRecord.NO_ID;
+import static org.neo4j.kernel.impl.store.record.RecordLoad.NORMAL;
+
+class Store extends Read implements Token
```
MishaDemianenko: i find naming of this class is super confusing taking into account presence of real stores in the same module that has nothing to do with this "store"

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/TempKernel.java
```diff
@@ -0,0 +1,142 @@
...
+import org.neo4j.kernel.impl.api.KernelTransactions;
+import org.neo4j.kernel.impl.storageengine.impl.recordstorage.RecordStorageEngine;
+import org.neo4j.kernel.internal.GraphDatabaseAPI;
+
+class TempKernel implements Kernel, Session
+{
+    private final Transaction tx;
+    private final Cursors cursors;
+
+    TempKernel( GraphDatabaseAPI db )
```
MishaDemianenko, line 48: would assume i do not need to look into this one

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
```diff
@@ -407,6 +407,23 @@ public boolean isInUse( long id )
     }
 
     /**
+     * DANGER: make sure to always close this cursor.
+     */
+    public PageCursor openPageCursor( long id )
```
MishaDemianenko: i do not think *DANGER* is enough to prevent errors and people from forgetting to close cursors we need to introduce technic that will allow trace not closed cursors and where they came from

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/IndexReader.java
```diff
@@ -91,5 +91,48 @@ public boolean hasFullNumberPrecision( IndexQuery... predicates )
...
+    {
+        try
+        {
+            // the actual property key is ignored in the implementation, so we can pass in whatever...
+            // it is actually ok for the index implementation to ignore the property key under these two assumptions:
+            // 1. That all queries is "well formed" - i.e. we never ask for a key not in the index.
+            // 2. For compound indexes all nodes have all properties assigned.
+            // While we violate 1. here, we are at least "well intended", we don't actually care about what the key is.
+            // 2. holds because compound indexes are only created through node keys.
+            IndexQuery.ExistsPredicate scan = IndexQuery.exists( -1 );
```
MishaDemianenko: maybe it makes sense to create a constant index query for those cases instead of creating a new each time here

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
```diff
@@ -19,7 +19,25 @@
  */
 package org.neo4j.internal.kernel.api;
 
-public interface Transaction extends Read, Write
+public interface Transaction
```
MishaDemianenko: Not sure e why do we need yet another transaction that has nothing to do with transactions that we already have

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RandomRelationshipTraversalCursorTestBase.java
```diff
@@ -30,69 +29,58 @@
...
-import static org.neo4j.graphdb.factory.GraphDatabaseSettings.dense_node_threshold;
 
-public class RandomRelationshipTraversalCursorTest
+public abstract class RandomRelationshipTraversalCursorTestBase<G extends KernelAPIReadTestSupport>
+        extends KernelAPIReadTestBase<G>
 {
     private static final int N_TRAVERSALS = 10_000;
     private static int N_NODES = 100;
     private static int N_RELATIONSHIPS = 1000;
     private static Random random = new Random( 666 );
```
MishaDemianenko: not sure that we need to use the same seed always, number of the beast is of cause can be treacherous but still, random seed looks more appropriate

MishaDemianenko (commented):
Made a quick review since it's impossible to do non-quick for that huge pr. Some comments here and there.
Also, in general, most of the code really lacking javadocs and it's, even more, harder sometimes to get purpose or the idea without surfing around.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
fickludd: Hmm, I don't know which existing interface you suggest that we use instead? `org.neo4j.graphdb.Transaction` is public, and `KernelTransaction` is on a lower level as far as I understand.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexCursorProgressor.java
fickludd: The define the 3 types of counter parts of the `IndexCursorProgressor` implementations. Where should they live? I think it would be possible to embed the `IndexCursorProgressor` logic directly into the cursors, but that would some at a cost of making these classes bigger, and it would also introduce some code redundancy.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
fickludd: Right, we definitely need support for minimizing the risk of bugs due to non-closed cursor. Both PageCursors and API cursors. I made a card for improving that.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/TempKernel.java
fickludd: That's right, it's only used as the access point for testing. The plan is to merge this class with `org.neo4j.kernel.impl.api.Kernel` in later work.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexCursorProgressor.java
burqen: After implementing those interfaces I think it makes sense and is pretty comfortable that they live together like this.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RandomRelationshipTraversalCursorTestBase.java
```diff
@@ -40,7 +40,8 @@
     private static final int N_TRAVERSALS = 10_000;
     private static int N_NODES = 100;
     private static int N_RELATIONSHIPS = 1000;
-    private static Random random = new Random( 666 );
+    private static long seed = (new Random()).nextInt();
+    private static Random random = new Random( seed );
```
burqen, line 44: May I suggest using `RandomRule` instead?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexQuery.java
```diff
@@ -32,6 +32,12 @@
 public abstract class IndexQuery
 {
     /**
+     * Scans all entries in the index.
+     */
+    @SuppressWarnings( "StaticInitializerReferencesSubClass" )
+    public static ExistsPredicate SCAN = new ExistsPredicate( -1 );
```
burqen: Magic number, please introduce static variable.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexReference.java
```diff
@@ -23,6 +23,11 @@
  * This class is really a renamed org.neo4j.kernel.api.schema.IndexDescriptor. Maybe it will compose an
  * IndexDescriptor with an offset for faster index lookup.
  */
-public class IndexReference
+public interface IndexReference
```
burqen: Include something about expected lifecycle of instances in javadoc. Can they be cached for example?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
```diff
@@ -19,7 +19,25 @@
  */
 package org.neo4j.internal.kernel.api;
 
-public interface Transaction extends Read, Write
```
burqen, line 22: Nice that this guy no longer extend `Read` and `Write`. That was super confusing!

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
```diff
@@ -19,7 +19,25 @@
...
+public interface Transaction
 {
+    long READ_ONLY_TRANSACTION = 0;
```
burqen: This value implicitly depend on `TransactionIdStore.BASE_TX_ID`. I would prefer if we could make that explicit or at the very least document it.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Write.java
```diff
@@ -24,7 +24,7 @@
 /**
  * Defines the write operations of the Kernel API.
  */
-interface Write
+public interface Write
```
burqen, line 27: This interface and it's method could use some javadoc. Especially for return values of the property methods.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/KernelAPIReadTestBase.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.rules.TemporaryFolder;
+
+import java.io.IOException;
+
+import org.neo4j.graphdb.GraphDatabaseService;
+
+@SuppressWarnings( "WeakerAccess" )
+public abstract class KernelAPIReadTestBase<G extends KernelAPIReadTestSupport>
```
burqen: Can we please document the interaction between `KernelAPIReadTestBase` and `KernelAPIReadTestSupport` to make it easier for others to expand test suite.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/KernelAPIWriteTestBase.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.rules.TemporaryFolder;
+
+import java.io.IOException;
+
+import org.neo4j.graphdb.GraphDatabaseService;
+
+@SuppressWarnings( "WeakerAccess" )
+public abstract class KernelAPIWriteTestBase<G extends KernelAPIWriteTestSupport>
```
burqen: And javadoc for those guys as well.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/LowLevelPropertyCursorTestBase.java
```diff
@@ -0,0 +1,189 @@
...
+            assertThat( cursor.propertyKey(), equalTo( NO_SUCH_PROPERTY ) );
+            assertThat( cursor.propertyValue(), equalTo( Values.NO_VALUE ) );
+
+            assertException( cursor::booleanValue );
+            assertException( cursor::stringValue );
+            assertException( cursor::longValue );
+            assertException( cursor::doubleValue );
+
+            // Should we wait with specialized predicates?
+//            assertFalse( cursor.valueEqualTo( 0L ) );
```
burqen: Is this part forgotten or just on hold for now?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/LowLevelPropertyCursorTestBase.java
```diff
@@ -0,0 +1,189 @@
...
+    public void shouldParseInteger()
+    {
+        long X = 2L;
+        Map<Integer,Object> map = new HashMap<>();
+        map.put( 1, X );
+        PropertyCursor cursor = withValues( map );
+
+        assertTrue( cursor.next() );
+        assertThat( cursor.propertyKey(), equalTo( 1 ) );
+        assertThat( cursor.stringValue(), equalTo( X ) );
```
burqen: `cursor.longValue()`?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/LowLevelPropertyCursorTestBase.java
```diff
@@ -0,0 +1,189 @@
...
+    public void shouldParseFloat()
+    {
+        double X = 2.0;
+        Map<Integer,Object> map = new HashMap<>();
+        map.put( 1, X );
+        PropertyCursor cursor = withValues( map );
+
+        assertTrue( cursor.next() );
+        assertThat( cursor.propertyKey(), equalTo( 1 ) );
+        assertThat( cursor.stringValue(), equalTo( X ) );
```
burqen: `cursor.doubleValue()`?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/LowLevelPropertyCursorTestBase.java
```diff
@@ -0,0 +1,189 @@
...
+    public void shouldParseBoolean()
+    {
+        boolean X = false;
+        Map<Integer,Object> map = new HashMap<>();
+        map.put( 1, X );
+        PropertyCursor cursor = withValues( map );
+
+        assertTrue( cursor.next() );
+        assertThat( cursor.propertyKey(), equalTo( 1 ) );
+        assertThat( cursor.stringValue(), equalTo( X ) );
```
burqen: `cursor.booleanValue()`?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/LowLevelPropertyCursorTestBase.java
```diff
@@ -0,0 +1,189 @@
...
+import org.neo4j.values.storable.BufferValueWriter;
+import org.neo4j.values.storable.Values;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public abstract class LowLevelPropertyCursorTestBase
```
burqen: Should we not also test array values in this class?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+        try ( Transaction tx = graphDb.beginTx() )
+        {
+            graphDb.createNode( label( "Node" ) ).setProperty( "prop", "one" );
+            graphDb.createNode( label( "Node" ) ).setProperty( "prop", "two" );
+            graphDb.createNode( label( "Node" ) ).setProperty( "prop", "two" );
+            graphDb.createNode( label( "Node" ) ).setProperty( "prop", "three" );
+            graphDb.createNode( label( "Node" ) ).setProperty( "prop", "three" );
+            graphDb.createNode( label( "Node" ) ).setProperty( "prop", "three" );
+            graphDb.createNode( label( "Node" ) ).setProperty( "prop", 4 );
```
burqen: We should also add:
- duplicate entries for numbers
- boolean values
- floating point values

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexCursorProgressor.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.newapi;
+
+import org.neo4j.internal.kernel.api.LabelSet;
+import org.neo4j.values.storable.Value;
+
+public interface IndexCursorProgressor extends AutoCloseable
```
burqen: This interface together with nested counter parts could really use some javadoc. Without looking at the existing implementations it is very hard to figure out how they play together (and even after looking it sill gives me headaches 😅  ).

That being said, when you have figured out how those interfaces work they are pretty nice!

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/IndexReader.java
```diff
@@ -91,5 +93,47 @@ public boolean hasFullNumberPrecision( IndexQuery... predicates )
...
+        {
+            cursor.done();
+        }
     };
+
+    default void query( IndexCursorProgressor.NodeValueCursor cursor, IndexQuery... query )
+    {
+        try
+        {
+            cursor.initialize( new NodeValueIndexProgressor( query( query ), cursor ), null );
```
burqen: Why do we use `null` for property keys here? I know that `NodeValueIndexProgressor` doesn't serve any values but that should not be the concern of this default implementation. Through `query` we have access to the property key ids, let's use them.

Using `NodeValueIndexCursor.numberOfProperties()` to determine if values are possible to read directly from index does not work because it depend on what capability the underlying index implementation has and what values are stored. The capabilities for string may not be the same as for number for example. Therefore we can't know if values are available in index or not when we initialize `NodeValueCursor`. I'm working on extending the API for this purpose, so no work need to be done to accommodate for that in this PR. Just wanted to give some context about above comment.

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/IndexReader.java
```diff
@@ -91,5 +93,47 @@ public boolean hasFullNumberPrecision( IndexQuery... predicates )
...
+    {
+        try
+        {
+            // the actual property key is ignored in the implementation, so we can pass in whatever...
+            // it is actually ok for the index implementation to ignore the property key under these two assumptions:
+            // 1. That all queries is "well formed" - i.e. we never ask for a key not in the index.
+            // 2. For compound indexes all nodes have all properties assigned.
+            // While we violate 1. here, we are at least "well intended", we don't actually care about what the key is.
+            // 2. holds because compound indexes are only created through node keys.
+            cursor.initialize( new NodeValueIndexProgressor( query( SCAN ), cursor ), null );
```
burqen: Read comment above

burqen (changes requested):
Glad to see this API taking shape and coming to life 👍 

As a general comment, not only on this PR but on this module as a whole: The lack of javadoc concerns me. As this module is being written from scratch with the whole context in place from start we should really take this opportunity to make the javadoc super clear. Especially since this will be the main umbrella under which we work for (most likely) quite a long time. A good start would be javadoc for all interfaces in `org.neo4j.internal.kernel.api`. This documentation work does not necessarily need to be part of this PR but I would like to see this moving forward.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
MishaDemianenko: It's completely fine and understandable that you need some kind of new interface for this kind of facade but just named it specifically what kind of Transaction it is to avoid confusions since it would be hard to read code around different kinds of transactions. CypherTransaction, etc would be better

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexCursorProgressor.java
MishaDemianenko: if you need to implement an interface to make sense out of it it just better not to do that. Do not think that is any criteria in any way.

@fickludd it looks like those 3 Cursor classes have something in common - `initialization`and `done` parts that can be extracted into some sort of closeable cursor interface. 
For me, those 3 cursors look like some kind of Value cursors and IndexCursorProgressor would be the last place where I will gonna take a look into what kind of cursors do we have in case i want to implement new kind of cursor and make it comfortable in the existing  hierarchy. I would put them completely in a separate on in some sort of cursors helper class.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/PropertyUtil.java
```diff
@@ -0,0 +1,86 @@
...
+import org.neo4j.values.storable.ArrayValue;
+import org.neo4j.values.storable.Values;
+
+public class PropertyUtil
+{
+    public static ArrayValue readArrayFromBuffer( ByteBuffer buffer )
+    {
+        if ( buffer.limit() <= 0 )
+        {
+            throw new IllegalStateException( "Given buffer is empty" );
```
chrisvest, line 38: So the buffer was empty. What did you expect instead, and why? This error message can be improved.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/PropertyUtil.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        byte typeId = buffer.get();
+        buffer.order( ByteOrder.BIG_ENDIAN );
```
chrisvest, line 42: Why do we go out of our way to ensure the buffer is in BE mode?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/PropertyUtil.java
```diff
@@ -0,0 +1,86 @@
...
+                {   // Fallback to the generic approach, which is a slower
+                    Bits bits = Bits.bitsFromBytes( buffer.array(), buffer.position() );
+                    int length = ((buffer.limit() - buffer.position()) * 8 - (8 - bitsUsedInLastByte)) / requiredBits;
+                    return type.createArray( length, bits, requiredBits );
+                }
+            }
+        }
+        finally
+        {
+            buffer.order( ByteOrder.LITTLE_ENDIAN );
```
chrisvest, line 83: The buffer might not originally have been in LE mode. If we have to do this dance, then we should restore the buffer to whatever `buffer.order()` it had originally.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/PropertyUtil.java
```diff
@@ -0,0 +1,86 @@
...
+import org.neo4j.kernel.impl.store.PropertyType;
+import org.neo4j.kernel.impl.store.ShortArray;
+import org.neo4j.kernel.impl.util.Bits;
+import org.neo4j.string.UTF8;
+import org.neo4j.values.storable.ArrayValue;
+import org.neo4j.values.storable.Values;
+
+public class PropertyUtil
+{
+    public static ArrayValue readArrayFromBuffer( ByteBuffer buffer )
```
chrisvest, line 34: This feels misplaced. Similar code already has a home in `DynamicArrayStore`, and a bit in `PropertyStore`.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    @Override
+    ArrayValue array( PropertyCursor cursor, long reference, PageCursor page )
+    {
+        ByteBuffer buffer =
+                cursor.buffer = readDynamic( propertyStore.getArrayStore(), reference, cursor.buffer, page );
+        buffer.flip();
+        return PropertyUtil.readArrayFromBuffer( buffer );
+    }
```
chrisvest, line 302: These two methods, `string` and `array` are not matching the abstraction level of the other methods in this class. `PropertyCursor` already has methods that match these. I think this logic should be moved in that direction.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
chrisvest: I'd like to be more explicit and call this method `openPageCursorForReading`.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexQuery.java
fickludd: Because the fact that we have a `PropertyCursor` (pointing to a valid property record), means the the property exists, ergo the predicate is fulfilled.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Write.java
fickludd: Added javadocs.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/RecordStore.java
```diff
@@ -115,6 +116,8 @@
      */
     RECORD getRecord( long id, RECORD target, RecordLoad mode ) throws InvalidRecordException;
 
+    void getRecordByCursor( long id, RECORD target, RecordLoad mode, PageCursor cursor ) throws InvalidRecordException;
```
chrisvest: Please add some javadocs to this new API. For instance, it'd be good to know that the record id is understood to be absolute with reference to the store, and that the cursor will be moved to the relevant page if it isn't already there. You can reference `getRecord` for the record load mode behaviour.

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/IndexReader.java
```diff
@@ -91,5 +91,18 @@ public boolean hasFullNumberPrecision( IndexQuery... predicates )
...
+            cursor.done();
+        }
     };
+
+    default void query(
+            IndexCursorProgressor.NodeValueCursor cursor,
+            IndexQuery... query ) throws IndexNotApplicableKernelException
+    {
+        cursor.initialize( new NodeValueIndexProgressor( query( query ), cursor ), null );
+    }
```
chrisvest: I think this definition should move up next to the other method definitions in the interface. It also needs javadocs.

The `EMPTY` implementation just calls `done` on the cursor, while the default implementation only calls `initialize` and never calls `done`. I find this inconsistency in life cycle calls quite puzzling. The `EMPTY` really ought to be fine with the default implementation, and I don't understand why the default implementation is not calling `done`, since that appears to be something that this method _can_ do. This is very strongly indicative that we have a bug somewhere in here.

chrisvest: I'd like to insist that this is cleared up before merging.

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/ReadTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertThat( invertReference( NO_ID ), equalTo( (long) NO_ID ) );
+    }
+
+    @Test
+    public void shouldInvertTheInvert()
+    {
+        ThreadLocalRandom random = ThreadLocalRandom.current();
+        for ( int i = 0; i < 1000; i++ )
+        {
+            //TODO what is the real upper limit
```
chrisvest: The upper limit is record format specific, and given by `RecordFormat.getMaxId()`.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
```diff
@@ -0,0 +1,444 @@
...
+
+    static long removeFilteringFlag( long reference )
+    {
+        return reference & ~FILTER_MASK;
+    }
+
+    static boolean needsFiltering( long reference )
+    {
+        return (reference & FILTER_MASK) != 0L;
+    }
```
chrisvest: Is this "filtering" stuff related to the `invertReference`? If so, could we move these static methods and the `FILTER_MASK` to an interface with some top-level documentation that describe how it all fits together?

Also, this "filtering" concept isn't telling me anything about what is going on or what this is used for. Can we find a more descriptive name?

The javadocs on `invertReference` is pretty good, though.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexCursor.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.internal.kernel.api;
+
+/**
+ * A cursor for accessing data in an explicit index.
+ */
+interface ExplicitIndexCursor
```
chrisvest: The javadocs is wrong. This isn't even a cursor. I think we should call this something like `ExplicitIndexSearchResultSummary` or something like that.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/RelationshipIndexCursor.java
```diff
@@ -36,6 +36,5 @@
 
     long targetNodeReference();
 
-//  long relationshipReference(); // not sure relationships will have independent references, so exposing this might
-// be leakage.
+    long relationshipReference(); // will relationships have independent references? exposing it is leakage!
```
chrisvest: For relationships to be _indexed_, they will most likely need to have ids in some form.

chrisvest: The `long relationshipReference()` method in `RelationshipDataAccessor`, on the other hand… In a future version we might have relationships that don't actually have ids. If this new kernel API is meant to be stable for the foreseeable future, we should perhaps make that return `OptionalLong` instead? Alternatively we can dedicate a special value for signalling when a relationship exists, but has no id.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/SchemaRead.java
```diff
@@ -0,0 +1,28 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.internal.kernel.api;
+
+/**
+ * Surface for getting schema information, such as fetching specific indexes or constraints.
+ */
+public interface SchemaRead
+{
+    IndexReference index( int label, int... properties );
```
chrisvest: What if there is no such index? What about constraints?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Token.java
```diff
@@ -47,4 +47,8 @@
     void propertyKeyCreateForName( String propertyKeyName, int id ) throws SomeException;
 
     void relationshipTypeCreateForName( String relationshipTypeName, int id ) throws SomeException;
+
+    int nodeLabel( String name );
+
+    int propertyKey( String name );
```
chrisvest: So I guess these are non-creating getters? What if those labels and property keys don't exist? Some javadocs would be nice. Also, `SomeException` needs to be _something else_.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
```diff
@@ -19,7 +19,25 @@
...
 
+    long commit();
```
chrisvest: I _guess_ the long returned here is either the transaction id, or the `READ_ONLY_TRANSACTION` constant?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
chrisvest: In that case I think the dependency should go in the direction such that `BASE_TX_ID` is defined in terms of this value, so that in order to understand this publicly exposed constant, you don't have to first understand some deeply internal constant.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/DeepRelationshipTraversalCursorTestBase.java
```diff
@@ -0,0 +1,141 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.neo4j.graphdb.RelationshipType.withName;
+
+public abstract class DeepRelationshipTraversalCursorTestBase<G extends KernelAPIReadTestSupport>
+        extends KernelAPIReadTestBase<G>
+{
+    private static long three_root;
+    private static int expected_total, expected_unique;
```
chrisvest, line 40: I think this is against our code style.

chrisvest: Quite a lot of the new tests are doing this.

chrisvest (commented):
I had a few comments. Most of them I won't mind if you'd prefer to defer them to another PR, but the `IndexReader` stuff I'd like to see fixed in this one. Also the `PropertyUtil` stuff is a bit iffy.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/DeepRelationshipTraversalCursorTestBase.java
fickludd: It used to be against the code style, but there was agreement to revert that rule :).

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RandomRelationshipTraversalCursorTestBase.java
fickludd: Cool I didn't know about that! I prefer not to use it here though, since there is only one test and therefore not so much code duplication, while `RandomRule` would introduce a whole new dependency on the commons test-jar. If we need more test utils later, maybe this is an option.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
fickludd: I chose to only comment on this so far. I'm now sure how this will be used, and if we were to change the internal store transaction ids at some point, we could always map ids to match the Kernel API. Let's not overthink this yet.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
fickludd: Problem is, it's just a transaction. Nothing cyphery about it. It's in essence a `KernelTransaction`, but that name is also taken. I cannot find a better name, if anyone else does, let's improve this later.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexCursorProgressor.java
fickludd: Had a lengthy discussion with @burqen about this interface. The `Cursor` interfaces here really have nothing to do with cursors, so we renamed them. Added a lot of javadocs. This is really an internal index interface, so I also moved it next to `IndexReader` to signal that.

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/ReadTest.java
fickludd: That creates an interesting dependency circus if we were to encode it. I just hardcoded the highlimit limit and made a weak reference.

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/IndexReader.java
fickludd: This indeed seems strange, I let EMPTY fall back to the default implementation as you suggest. I don't think the `IndexReader` should ever call done() on the client, but rather leave that to the `Progressor`. 

I experimented with ways to make this more type-safe, but they all seemed to make the implementation increasingly contrived. Maybe we can think about this later?

fickludd:
@MishaDemianenko @burqen @chrisvest I think I addressed all you concerns. Thank you for keeping the code great. The only thing I didn't act on is `PropertyUtil`. I agree that the method is weirdly implemented, but we only moved it from `StorePropertyPayloadCursor`, and cleaning it up is something I just don't think is the purpose of this PR.

chrisvest (approved):

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/SchemaRead.java
fickludd: Eh, what about constraints? Cypher doesn't really care about them apart from create/drop, so let's add any constrainty methods on need?

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/IndexProgressor.java
```diff
@@ -0,0 +1,144 @@
...
+ */
+package org.neo4j.storageengine.api.schema;
+
+import org.neo4j.internal.kernel.api.LabelSet;
+import org.neo4j.values.storable.Value;
+
+/**
+ * The index progressor is a cursor like class, which allows controlled progression through the entries of an index.
+ * In contrast to a cursor, the progressor does not hold value state, but rather attempts to write the next entry to a
+ * Client. The client can them accept the entry, in which case next() returns, or reject it, in which case the
```
burqen, line 28: Client. The client can ~them~ then accept the entry, ~~in which case next() returns~~,  or reject it, in which case the

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/IndexProgressor.java
```diff
@@ -0,0 +1,144 @@
...
+ *                 client ---- next() --> progressor
+ *                        <-- accept() --
+ *                                 :false
+ *                        <-- accept() --
+ *                                 :false
+ *                        <-- accept() --
+ *                                  :true
+ *                 client <--------------
+ *   <-----------
+ */
```
burqen, line 48: Freakin' sweet diagram! This interface cluster is so much more understandable now 🎉

burqen (approved):
Found a weird sentence in javadoc, otherwise it looks like all my concerns have been addressed. Approved 🎉

PR #10287, opened 2017-10-16
Generate positive random bytes to get expected error in RecoveryCorruptedTransactionLogIT

MishaDemianenko:
Use random rule generator only, generate random positive bytes to be
able to fail always version selection.


community/kernel/src/test/java/org/neo4j/kernel/RecoveryCorruptedTransactionLogIT.java
```diff
@@ -454,7 +454,7 @@ private void truncateBytesFromLastLogFile( long bytesToTrim ) throws IOException
         }
     }
 
-    private void addRandomLongsToLastLogFile() throws IOException
+    private void addRandomLongsToLastLogFile( Supplier<Byte> byteSource ) throws IOException
```
burqen: Method name is no longer accurate. `addRandomByteToLastLogFile`

burqen (approved):
Approved with updated method name

PR #10288, opened 2017-10-16
More verbose logging for dynamic configuration changes

klaren:
Logging now includes the origin of the change and an indication of whether the value is default or not.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
@@ -572,10 +572,11 @@ public class Config implements DiagnosticsProvider, Configuration
      *
      * @param setting The setting to set to the specified value.
      * @param newValue The new value to set, passing {@code null} or empty should reset the value back to default value.
+     * @param origin The source of the change, e.g. {@code dbms.setConfigValue()}.
      * @throws IllegalArgumentException if the provided setting is unknown or not dynamic.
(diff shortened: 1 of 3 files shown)


community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -615,14 +616,17 @@ public void updateDynamicSetting( String setting, String newValue ) throws Illeg
                 validator.validate( newEntry, ignore -> {} ); // Throws if invalid
             }
 
-            oldValue = getDefaultValueOf( setting );
+            oldValue = getConfiguredValueOf( setting );
 
-            params.put( setting, newValue );
+            if ( params.put( setting, newValue ) == null )
```
MishaDemianenko, line 621: this assumption is not really correct since it can be the case that some property was reset to be null and that is not a default value of right?

MishaDemianenko (commented):
Single small question

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
klaren: Due to the way lookup works, `null` is treated as the default value. So technically you cannot set `null` as a value and have the setting return that unless the default value is `null`. See `org.neo4j.kernel.configuration.Settings.DefaultSetting#apply`.

MishaDemianenko (approved):

PR #10290, opened 2017-10-16
Fix bug in SingleRowSlottedPipe

henriknyman:
If a SingleRow appears on the right-hand side of a logical plan
it needs to copy the arguments from the query state, or we could
end up using uninitialized slots.


henriknyman:
The only way we have been able to trigger this is with a merge relationship. Currently a merge relationship plan will not compile with slotted runtime on 3.3, and it will use the fallback, so maybe we do not need to fix this.

henriknyman:
This pull request has somehow got a broken connection to the actual branch, so I have to reopen a new one https://github.com/neo4j/neo4j/pull/10303.

PR #10291, opened 2017-10-16
Fix bug in UnwindSlottedPipe

henriknyman:
We need to also copy the first incoming row since it can have a
different slot configuration than the pipeline that contains
the Unwind.


henriknyman:
This is only an issue if the evaluation of the collection expression is using additional slots on the new pipeline, which is if it has a scope expression (e.g. extract, list comprehension etc.) or nested plan expression. Those expressions does not currently compile with the slotted runtime on 3.3 and it will use the fallback, so maybe we do not need to fix this.

fickludd:
This will be fixed when we properly support slotted scope expressions

PR #10292, opened 2017-10-16
Storing Points

craigtaverner:
This work provides initial support for storing Point geometries in the store. This involved some cleanup of the existing Point type support, and the PointValue system in particular. Some thinking has gone into future support for non-Point geometries, but only Point is supported at this point. Some key parts of this work:
- Use ValueWriter in PropertyStore for non-array types
- Use public interface for point in PointValue and contained types since the 3.3 removal of the runtime from the compiler model in Cypher
- Serialization/deserialization of 2D and 3D points (cannot support higher dimensions without moving to dynamic store, or switching precision to float, which is also considered in the format).

The key capabilities that would allow us to remove the 'not ready for merge' flag would be full stack support for assigning and reading Points as properties of nodes, using Cypher and Core API.


community/kernel/src/main/java/org/neo4j/helpers/BaseToObjectValueWriter.java
```diff
@@ -290,17 +285,43 @@ public void remove()
...
+    // TODO remove the next five methods, once this class moved to a non-public API location
+    @Deprecated
     public void beginPoint( CoordinateReferenceSystem coordinateReferenceSystem ) throws RuntimeException
     {
-        stack.push( new PointWriter( coordinateReferenceSystem ) );
+        throw new UnsupportedOperationException();
     }
 
-    @Override
+    @Deprecated
```
pontusmelke: Just delete no need to deprecate

community/values/src/main/java/org/neo4j/values/storable/PointValue.java
```diff
@@ -0,0 +1,138 @@
...
+package org.neo4j.values.storable;
+
+import java.util.Arrays;
+
+import org.neo4j.values.utils.PrettyPrinter;
+import org.neo4j.values.virtual.CoordinateReferenceSystem;
+
+import static java.lang.String.format;
+
+public class PointValue extends ScalarValue
```
pontusmelke: I know you badly want it but `Point` is not a storable value yet and shouldn't live here.

tinwelint:
@craigtaverner there's no description on this PR. What is it about, what drove it, how does it affect things/storage? It's super hard to read a PR w/o getting primed with those sorts of contexts. Thanks.

tinwelint:
How final is the format? Will it change soon or ongoing? This puts our versioning to the test... how do we version this?

craigtaverner:
@tinwelint The format is certainly not final. We are now at the stage where we would like feedback from kernel developers and anyone with interest in store format specification. The current state of the PR is a first draft of what we think makes sense, but should certainly be discussed further before finalizing.

craigtaverner:
@tinwelint Discussions with @digitalstain last week indicate this format is backwards compatible (additive, upgrades require no store change) and partly forwards compatible (can downgrade as long as no-one added points to the store). We should decide how to (or if we need to) reflect this in the store version number.

community/kernel/src/main/java/org/neo4j/kernel/api/index/ArrayEncoder.java
```diff
@@ -130,10 +131,17 @@ public void writeString( char value )
...
 
         @Override
-        public void writeString( char[] value, int offset, int length )
-        {
-            builder.append( value, offset, length );
-            builder.append( '|' );
+        public void writePoint( CoordinateReferenceSystem crs, double[] coordinate ) throws RuntimeException
+        {
+            throw new UnsupportedOperationException( "Arrays of points are not yet supported." );
+            // Possible implementation:
```
MishaDemianenko: this comment section should go if it's not supported

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,164 @@
...
+                    int dimension = getDimension( firstBlock );
+                    if ( dimension > 3 )
+                    {
+                        return PropertyType.BLOCKS_USED_FOR_BAD_TYPE_OR_ENCODING;
+                    }
+                    return 1 + dimension;
+                }
+            };
+
+    public final int gtype;
```
MishaDemianenko: not sure why this field is public

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,164 @@
...
+{
+    GEOMETRY_POINT( 0 )
+            {
+                @Override
+                public Value decode( CoordinateReferenceSystem crs, int dimension, long[] valueBlocks, int offset )
+                {
+                    double[] coordinate = new double[dimension];
+                    for ( int i = 0; i < dimension; i++ )
+                    {
+                        coordinate[i] = Double.longBitsToDouble( valueBlocks[i + 1 + offset] );
```
MishaDemianenko, line 70: no need to do +1 inside in the loop just do iteration from 1 till <= dimension

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,164 @@
...
+            };
+
+    public final int gtype;
+
+    GeometryType( int gtype )
+    {
+        this.gtype = gtype;
+    }
+
+    public abstract Value decode( CoordinateReferenceSystem crs, int dimension, long[] valueBlocks, int offset );
```
MishaDemianenko: I can see that those 2 methods are abstract and we have only one type that overriding it. Do we plan to add more?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,164 @@
...
+
+    // TODO this might not work this way for Polygons or other geometries that do not
+    // know their length just from the firstBlock. Since we will probably use dynamic property
+    // stores, it might be that the answer here is just 1.
+    public abstract int calculateNumberOfBlocksUsedForGeometry( long firstBlock );
+
+    private static int getGeometryType( long firstBlock )
+    {
+        return (int) ((firstBlock & 0x00000000F0000000L) >> 28);
```
MishaDemianenko: all those magic numbers and masks should be extracted as constants with proper naming for future readers convenience

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,164 @@
...
+    public final int gtype;
+
+    GeometryType( int gtype )
+    {
+        this.gtype = gtype;
+    }
+
+    public abstract Value decode( CoordinateReferenceSystem crs, int dimension, long[] valueBlocks, int offset );
+
+    // TODO this might not work this way for Polygons or other geometries that do not
```
MishaDemianenko: who/when we should address this todo?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+        else
+        {
+            return geometryType.calculateNumberOfBlocksUsedForGeometry( firstBlock );
+        }
+    }
+
+    private static GeometryType find( int gtype )
+    {
+        for ( GeometryType type : GeometryType.values() )
```
MishaDemianenko: this for loop can be more effective in a couple of ways: "GeometryType.values()" are always the same so you do not need to create a new array each time; you can use your target gtype as an index to access required element directly without need to check each of arrays element for match in a for loop

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,164 @@
...
+
+    public static long[] encodePoint( int keyId, CoordinateReferenceSystem crs, double[] coordinate ) throws IllegalArgumentException
+    {
+        if ( coordinate.length > 3 )
+        {
+            // One property block can only contains at most 4x8 byte parts, one for header and 3 for coordinates
+            throw new UnsupportedOperationException( "Points with more than 3 dimensions are not supported in the PropertyStore" );
+        }
+
+        long keyAndType = keyId | (((long) (PropertyType.GEOMETRY.intValue()) << StandardFormatSettings.PROPERTY_TOKEN_MAXIMUM_ID_BITS));
```
MishaDemianenko: please use some local variable for StandardFormatSettings.PROPERTY_TOKEN_MAXIMUM_ID_BITS takes a lot of space and please described desired format somewhere in java doc

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,164 @@
...
+        int gtype = getGeometryType( firstBlock );
+        int dimension = getDimension( firstBlock );
+
+        if ( isFloatPrecision( firstBlock ) )
+        {
+            throw new UnsupportedOperationException( "Float precision is unsupported in Geometry properties" );
+        }
+        if ( dimension > 3 )
+        {
+            throw new UnsupportedOperationException( "Points with more than 3 dimensions are not supported in the PropertyStore" );
```
MishaDemianenko: i would just say that "Points with more than 3 dimensions are not supported" since most of the supported/unsupported things are property format related and those can be quite different independently from type of store.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,164 @@
...
+        return decode( block.getValueBlocks(), 0 );
+    }
+
+    public static Value decode( long[] valueBlocks, int offset )
+    {
+        long firstBlock = valueBlocks[offset];
+        int gtype = getGeometryType( firstBlock );
+        int dimension = getDimension( firstBlock );
+
+        if ( isFloatPrecision( firstBlock ) )
```
MishaDemianenko, line 256: i can see this check is used only to say that it's unsupported. Is that how it was planned?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyStore.java
```diff
@@ -226,91 +229,155 @@ public void encodeValue( PropertyBlock block, int keyId, Value value )
...
-        else if ( asObject instanceof Integer )
+        else
         {
-            setSingleBlockValue( block, keyId, PropertyType.INT, ((Integer) asObject).longValue() );
+            value.writeTo( new PropertyBlockValueWriter( block, keyId, stringAllocator ) );
         }
-        else if ( asObject instanceof Boolean )
+    }
+
+    private static class PropertyBlockValueWriter implements ValueWriter<IllegalArgumentException>
```
MishaDemianenko: we tried to use property block value writer here before but benchmarks where showing degradation because of that please double check that this can be done now

enterprise/runtime/neole/src/main/java/org/neo4j/internal/store/prototype/neole/PropertyCursor.java
```diff
@@ -83,6 +84,12 @@
      *                                                                                 0 means 64, other values "normal"
      *            [xxxx,xxxx] [xxxx,xxxx] [xxxx,    ] [    ,    ] payload(+ maybe in next block) (0xFFFF_FF00_0000_0000)
      *                                                            bits are densely packed, bytes torn across blocks
+     * POINT:     [    ,    ] [    ,    ] [    ,    ] [    ,    ] [xxxx,type][K][K][K] geometry subtype
```
MishaDemianenko: please move this format description to a place where that format is defined as well and reference this class or vica versa in javadoc

community/values/src/main/java/org/neo4j/values/storable/CRSTable.java
```diff
@@ -17,36 +17,43 @@
...
-public enum CoordinateReferenceSystem
+public enum CRSTable
 {
-    Cartesian( "cartesian", 7203, "http://spatialreference.org/ref/sr-org/7203/" ),
-    WGS84( "WGS-84", 4326, "http://spatialreference.org/ref/epsg/4326/" );
-
+    CUSTOM( "custom", 0 ),
+    EPSG( "epsg", 1 ),
+    SR_ORG( "sr-org", 2 );
     public final String name;
```
MishaDemianenko: same here as before why those field a republic?

community/values/src/main/java/org/neo4j/values/storable/CRSTable.java
```diff
@@ -17,36 +17,43 @@
...
-    {
-        return code;
+        this.tableId = tableId;
     }
 
-    public String type()
+    public static CRSTable find( int tableId )
     {
-        return name;
+        for ( CRSTable table : CRSTable.values() )
```
MishaDemianenko: also same lookup pattern as described before

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -0,0 +1,102 @@
...
+public class CoordinateReferenceSystem implements CRS
+{
+    public static final CoordinateReferenceSystem Cartesian = new CoordinateReferenceSystem( "cartesian", CRSTable.SR_ORG, 7203 );
+    public static final CoordinateReferenceSystem WGS84 = new CoordinateReferenceSystem( "WGS-84", CRSTable.EPSG, 4326 );
+
+    public static CoordinateReferenceSystem get( int tableId, int code )
+    {
+        CRSTable table = CRSTable.find( tableId );
+        // TODO this might break, since we are changing the name. Use the function below?
+        return new CoordinateReferenceSystem( table.name + "-" + code, table, code );
```
MishaDemianenko: not sure why do we need to create new instances of reference system here? we have only fixed number of those systems, right?

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -0,0 +1,102 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.values.storable;
+
+import org.neo4j.graphdb.spatial.CRS;
+
+public class CoordinateReferenceSystem implements CRS
```
MishaDemianenko: it really looks like this class should be enum instead?

community/values/src/main/java/org/neo4j/values/storable/Values.java
```diff
@@ -253,6 +257,40 @@ public static ShortArray shortArray( short[] value )
...
+    public static PointValue pointValue( CoordinateReferenceSystem crs, double... coordinate )
+    {
+        return new PointValue( crs, coordinate );
+    }
+
+    public static PointValue point( Point point )
+    {
+        // An optimization could be to do an instanceof PointValue check here
+        // and in that case just return the casted argument.
+        List<Double> coordinate = point.getCoordinate().getCoordinate();
```
MishaDemianenko, line 285: ".getCoordinate().getCoordinate()" looks quite strange

community/values/src/main/java/org/neo4j/values/storable/Values.java
```diff
@@ -253,6 +257,40 @@ public static ShortArray shortArray( short[] value )
...
+        for ( int i = 0; i < coords.length; i++ )
+        {
+            coords[i] = coordinate.get( i );
+        }
+        return new PointValue( crs( point.getCRS() ), coords );
+    }
+
+    public static CoordinateReferenceSystem crs( CRS crs )
+    {
+        if ( crs.getHref().equals( CoordinateReferenceSystem.WGS84.href ) )
```
MishaDemianenko: would prefer to have condition flipped to have constants on the left side to avoid NPEs in any case independently from crs that can be passed here ever

community/values/src/main/java/org/neo4j/values/utils/PrettyPrinter.java
```diff
@@ -152,16 +152,22 @@ public void writePath( NodeValue[] nodes, EdgeValue[] edges )
...
-        stack.push( new PointWriter( coordinateReferenceSystem ) );
-    }
-
-    @Override
-    public void endPoint()
-    {
-        assert !stack.isEmpty();
-        append( stack.pop().done() );
+        append( "{geometry: {type: \"Point\", coordinates: [" );
+        for ( int i = 0; i < coordinate.length; i++ )
```
MishaDemianenko: Arrays.toString( coordinate ) and you will have array [a, b, c] without need to put commas yourself and being careful about that

MishaDemianenko (changes requested):
In addition to already mentioned things would really like to see perf results for this change and also more integration tests with storing all sorts of coordinates: negative, big values, NANs if possible, etc

MishaDemianenko:
Also important note we need to bump store format versions in order to prevent people with older versions (without points) trying to read new property types

enterprise/runtime/neole/src/main/java/org/neo4j/internal/store/prototype/neole/PropertyCursor.java
fickludd: We are actually about to nuke NeoLE from the neo4j code base, so this is definitely not the future-proof place to document format changes. Maybe somewhere around the `PropertyRecordFormat`?

craigtaverner: We should move all the format descriptions here to a better place. We only added the point format here because it was the only place where the formats are currently described in a reasonable way. Perhaps that move is done as part of the 'nuke neole' work?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
craigtaverner: Yes :-)

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
craigtaverner: We would like to do this soon, but there are some unanswered questions around polygon indexing which we need to answer.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
craigtaverner: We copied the conventions we saw elsewhere in the code where it was easy to read the meaning of the value like this. If we change this place, perhaps we should change all other cases in PropertyStore where this happens?

MishaDemianenko: if you have time to do that that would be really nice

community/values/src/main/java/org/neo4j/values/storable/CRSTable.java
craigtaverner: I always use public fields for enums because they are immutable. Seems a common pattern, but we can do it differently with accessor methods if that is preferred for stylistic reasons.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
craigtaverner: Then we need a `-1` inside the loop. Replacing `+1` with `-1` does not feel like a win to me.

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
craigtaverner: There are a few thousand CRS out there, normally defined by libraries. We expect to either:
* Include an appropriate dependency to populate this table
* Allow users to include such dependencies to extend this table for their domain

Right now we will not expose that capability, and we could have an enum, but it feels better to have the code sympathetic to the direction we expect it to go.

community/values/src/main/java/org/neo4j/values/storable/Values.java
craigtaverner: It does indeed. And sadly this is part of the public interface since 3.0. Personally I think that interface was not perfectly designed and would like to change it. In more ways than this one. Current policy is to only make changes on major releases. I wonder if we could, for a possibly entirely unused interface like this, motivate to make changes on minor releases? Worth discussing.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
MishaDemianenko: you absolutely right, nevermind

community/values/src/main/java/org/neo4j/values/storable/CRSTable.java
MishaDemianenko: yeah i would prefer access methods

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
MishaDemianenko: ok sounds reasonable

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
sherfert: This is something open for discussion. We wanted to have the opportunity that users can trade precision for storage compactness. But we do not want to implement this right now.

enterprise/runtime/neole/src/main/java/org/neo4j/internal/store/prototype/neole/PropertyCursor.java
sherfert: Moved to PropertyStore

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
sherfert: The answer is 1 for all Geometries storing in a dynamic store. That is also the default in PropertyType

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,335 @@
...
+                    }
+                    else
+                    {
+                        throw new RuntimeException(
+                                "Point array with unexpected type. Actual:" + dataValue.getClass().getSimpleName() + ". Expected: FloatingPointArray." );
+                    }
+                }
+            };
+
+    public final int gtype;
```
MishaDemianenko: comment about public field still valid

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
MishaDemianenko: if it will not be implemented right know what is the point of keeping that check?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyStore.java
MishaDemianenko: This comment is still valid. Any benchmarks data?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
```diff
@@ -0,0 +1,335 @@
...
+            bytes[4] = (byte) (crs.code >> 8 & 0xFFL);
+            bytes[5] = (byte) (crs.code & 0xFFL);
+        }
+
+        static GeometryHeader fromArrayHeaderBytes( byte[] header )
+        {
+            int geometryType = Byte.toUnsignedInt( header[1] );
+            int dimension = Byte.toUnsignedInt( header[2] );
+            int crsTableId = Byte.toUnsignedInt( header[3] );
+            int crsCode = Byte.toUnsignedInt( header[5] ) + (Byte.toUnsignedInt( header[4] ) << 8);
```
MishaDemianenko: can we please update line to "(Byte.toUnsignedInt( header[4] ) << 8) + Byte.toUnsignedInt( header[5] ) "

community/kernel/src/test/java/org/neo4j/kernel/impl/store/TestArrayStore.java
```diff
@@ -100,6 +103,15 @@ public void longArrayPropertiesShouldBeBitPacked() throws Exception
     }
 
     @Test
+    public void doubleArrayPropertiesShouldBeBitPacked() throws Exception
+    {
+        assertBitPackedArrayGetsCorrectlySerializedAndDeserialized( new double[]{Double.longBitsToDouble( 0x1L ), Double.longBitsToDouble( 0x7L )},
```
MishaDemianenko: please reformat those long lines in this test

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -0,0 +1,128 @@
...
+        else
+        {
+            // We expect this path only to get executed once we allow writing custom CRS's to the store
+            return new CoordinateReferenceSystem( table.getName() + "-" + code, table, code );
+        }
+    }
+
+    public static CoordinateReferenceSystem get( CRS crs )
+    {
+        if ( crs == null || crs.getHref() == null )
```
MishaDemianenko: since get by href is now null safe `crs.getHref()` is not mandatory anymore.

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -0,0 +1,128 @@
...
+        {
+            throw new IllegalArgumentException( "Invalid CRS: " + crs );
```
MishaDemianenko: maybe Objects.requireNonNull maybe use instead

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -0,0 +1,128 @@
...
+        {
+            return CoordinateReferenceSystem.WGS84;
+        }
+        else if ( CoordinateReferenceSystem.Cartesian.href.equals( href ) )
+        {
+            return CoordinateReferenceSystem.Cartesian;
+        }
+        else
+        {
+            throw new UnsupportedOperationException( "Unknown CRS: " + href );
```
MishaDemianenko: would say that its more illegal argument

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    public final String name;
```
MishaDemianenko: all of the fields should be private

community/values/src/main/java/org/neo4j/values/storable/CRSTable.java
```diff
@@ -0,0 +1,97 @@
...
+    private final String name;
+    private final int tableId;
+
+    CRSTable( String name, int tableId )
+    {
+        this.name = name;
+        this.tableId = tableId;
+    }
+
+    private static final CRSTable[] TYPES = CRSTable.values();
```
MishaDemianenko: please move all static stuff before a instance related stuff, here and in other new enums

community/values/src/main/java/org/neo4j/values/storable/PointValue.java
```diff
@@ -0,0 +1,203 @@
...
+            }
+        }
+        return true;
+    }
+
+    @Override
+    protected boolean eq( Object other )
+    {
+        return other != null &&
+               (
```
MishaDemianenko: strange formatting

MishaDemianenko (changes requested):
Have couple of minor comments on my second round

community/kernel/src/main/java/org/neo4j/kernel/impl/store/GeometryType.java
sherfert: As a reminder for later. Do you suggest to remove and simply assume it will be doublePrecision for now?

community/values/src/main/java/org/neo4j/values/storable/PointValue.java
sherfert: I actually cannot think of a more compact formatting that is still quite readable.

community/values/src/test/java/org/neo4j/values/storable/ValueComparisonTest.java
```diff
@@ -38,6 +40,16 @@
...
     private Object[] objs = new Object[]{
+            // POINTS
+            pointValue( CoordinateReferenceSystem.WGS84, -1.0, -1.0 ),
+            pointValue( CoordinateReferenceSystem.WGS84, 1.0, 1.0 ),
+            pointValue( CoordinateReferenceSystem.WGS84, 1.0, 2.0 ),
+            pointValue( CoordinateReferenceSystem.WGS84, 2.0, 1.0 ),
+            pointValue( CoordinateReferenceSystem.Cartesian, -1.0, -1.0 ),
+            pointValue( CoordinateReferenceSystem.Cartesian, 1.0, 1.0 ),
+            pointValue( CoordinateReferenceSystem.Cartesian, 1.0, 2.0 ),
+            pointValue( CoordinateReferenceSystem.Cartesian, 2.0, 1.0 ),
```
fickludd: Just got curious about this work... I think there should be point-arrays here somewhere too?

community/values/src/test/java/org/neo4j/values/storable/PointTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class PointTest
+{
+    @Test
+    public void cartesianShouldEqualItself()
+    {
+        assertEqual( pointValue( CoordinateReferenceSystem.Cartesian, 1.0, 2.0 ), pointValue( CoordinateReferenceSystem.Cartesian, 1.0, 2.0 ) );
+        assertEqual( pointValue( CoordinateReferenceSystem.Cartesian, -1.0, 2.0 ), pointValue( CoordinateReferenceSystem.Cartesian, -1.0, 2.0 ) );
+        assertEqual( pointValue( CoordinateReferenceSystem.Cartesian, -1.0, -2.0 ), pointValue( CoordinateReferenceSystem.Cartesian, -1.0, -2.0 ) );
+        assertEqual( pointValue( CoordinateReferenceSystem.Cartesian, 0.0, 0.0 ), pointValue( CoordinateReferenceSystem.Cartesian, 0.0, 0.0 ) );
```
fickludd: This would be a nice spot for static import.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/DynamicArrayStore.java
```diff
@@ -34,18 +34,58 @@
...
+ *             <li>Byte 2: The dimension of the geometry (currently only 2 or 3 dimensions are supported)</li>
+ *             <li>Byte 3: Coordinate Reference System Table id: {@link CRSTable}</li>
+ *             <li>Bytes 4-5: 16bit short Coordinate Reference System code: {@link CoordinateReferenceSystem}</li>
+ *         </ul>
+ *         The format of the body is specific to the type of Geometry being stored:
+ *         <ul>
+ *             <li>Points: Stored as double[] using the same format as primitive arrays above, starting with the 3 byte header (see above)</li>
+ *         </ul>
+ *     </li>
+ * </ul>
```
fickludd, line 84: Nice!

community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyStore.java
sherfert: Yes. No regressions from what we can see. An image is attached to the trello card.

community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigratorTest.java
```diff
@@ -205,6 +207,13 @@ public void shouldNotMigrateFilesForVersionsWithSameCapability() throws Exceptio
         assertFalse( progressReporter.started );
     }
 
+    @Test
+    public void shouldOpenStore()
```
MishaDemianenko: what do we testing here?

MishaDemianenko (approved):

PR #10295, opened 2017-10-17
Proper fix for "not" error message if used right of "="

sherfert:
This PR adds to the already merged
https://github.com/neo4j/neo4j/pull/10262/ that did not fix the problem
in its entirety.


Lojjs (approved):
Good addition to make it case-insensitive

PR #10297, opened 2017-10-17
Fix deadlock between index close and index add

burqen:
`this` is instance of `LuceneDataSource`

T1 want to delete index 1
T2 want to add to index 1

T1 `closeIndex` synchronized on `this`
T2 `getIndexSearcher` synchronize on searcher for index 1
T1 BLOCKED on `searcher.dispose` on searcher for index 1,
   which internally synchronize on searcher. Held by T2.
T2 BLOCKED on `syncGetIndexSearcher`, synchronized on `this`.
   Held by T1.
DEADLOCK

Instead of synchronizing on found searcher, T2 synchronize
on `this` avoiding grabbing locks in reverse order.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/explicit/LuceneDataSource.java
+++ b/community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/explicit/LuceneDataSource.java
@@ -298,7 +298,7 @@ public class LuceneDataSource extends LifecycleAdapter
         {
             return syncGetIndexSearcher( identifier );
         }
-        synchronized ( searcher )
+        synchronized ( this )
         {
             /*
              * We need to get again a reference to the searcher because it might be so that


community/lucene-index/src/main/java/org/neo4j/index/impl/lucene/explicit/LuceneDataSource.java
```diff
@@ -298,7 +298,7 @@ IndexReference getIndexSearcher( IndexIdentifier identifier ) throws ExplicitInd
         {
             return syncGetIndexSearcher( identifier );
         }
-        synchronized ( searcher )
+        synchronized ( this )
```
tinwelint, line 301: It looks like the path into this `synchronized` block is the most common one. Won't this change increase contention a bit too much?

burqen: @tinwelint I don't see any other simple way around it. This class is pretty much the all-star team of concurrency control with `synchronized` blocks both on self and inner instances that themselves also have `synchronized` methods, one concurrent map, two additional concurrent collections hidden inside `IndexClockCache` and then also a `ReentrantReadWriteLock` that can be grabbed from outside.

Untangling this seems like a bigger risk to me, right now.

burqen:
Need a better fix. Leave as is for now.

PR #10298, opened 2017-10-17
Fix bug for create/merge relationship in slotted runtime

henriknyman:
The slotted pipe builder should not mix up long slots and
reference slots for create/merge relationship.

We currently only support long slots in these pipes, so getting nodes
in reference slots should generate a compile time error and fall back
to interpreted runtime.

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CreateAcceptanceTest.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CreateAcceptanceTest.scala
@@ -91,4 +91,25 @@ class CreateAcceptanceTest extends ExecutionEngineFunSuite with QueryStatisticsT
     result.executionPlanDescription()
   }
 
+  //Not TCK material
+  // This test exposed a bug in the slotted runtime where it could mix up long slots with ref slots
+  test("should not accidentally create relationship between the wrong nodes") {
+    val a = createLabeledNode("A")
+    val b = createLabeledNode("B")
+
+    val query =
+      """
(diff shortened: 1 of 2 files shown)


SaschaPeukert (approved):

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CreateAcceptanceTest.scala
```diff
@@ -91,4 +91,25 @@ class CreateAcceptanceTest extends ExecutionEngineFunSuite with QueryStatisticsT
     result.executionPlanDescription()
   }
 
+  //Not TCK material
+  // This test exposed a bug in the slotted runtime where it could mix up long slots with ref slots
+  test("should not accidentally create relationship between the wrong nodes") {
```
thobe, line 96: Testing for specific bugs are usually called _"Regression testing"_. Perhaps we should have a separate (set of) test class(es) for regression tests, because it feels a bit out of place here...

henriknyman:
Opened as https://github.com/neo4j/neo4j/pull/10305 instead

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CreateAcceptanceTest.scala
henriknyman: That would be kind of nice to separate those. But we haven't had it so far in the scala acceptance tests so they are already scattered with regression tests for specific issues. Maybe we can do this as a separate test refactoring?

PR #10302, opened 2017-10-17
Removes contention on reporting bad data in importer

tinwelint:
By using AsyncEvents in BadCollector. The reporting isn't made faster, instead
most of the work is placed outside the thread reporting the bad data.
This allows the importer threads to execute at basically full speed and
the bad data will end up in the log eventually.


ragadeeshu (approved):
Looks nice. could more work be delegated to the logger, or is the remaining in-thread work already negligible?

tinwelint:
It feels like it's doing the minimal amount of work on the executing thread already. Some performance measurements showed a contention down to 1/6th from before the change on my 4-core (8 hyper threads) machine running w/ 8 threads.

Issue #4582, opened 2015-05-07
Isn't TransactionTemplate missing the failure handling?


tkroman:
I really like how easy it became to handle non-fatal exceptions now that the `TransactionTemplate` was introduced, but the `Monitor` class seems to have `failed()` method, which isn't used anywhere, while another (`failure()`) is used in case a `Throwable` is caught. Is it the intended behavior? Do you have some plans for the future? I think it would be useful to place a call to `failed()` right before the `break` happens (in case an exception is not accepted by the predicate).
https://github.com/neo4j/neo4j/blob/2.2.1/community/kernel/src/main/java/org/neo4j/helpers/TransactionTemplate.java#L157

You might have completely different intentions about the usage, so I'm just asking if the fact that this method is unused is normal or just something someone had overlooked.

I would be glad to implement the feature myself along with a couple of tests in case I'm right.

klaren:
This looks like a bug in the implementation, I've created a PR to fix it. `failed()` is called after all the retries have failed or the thread is interrupted. Since this bug fix changes behavior, it will only go into the next release and not be backported.

Thanks for reporting!

PR #10309, opened 2017-10-18
TransactionTemplate failed callback now called

klaren:
`TransactionTemplate` has a monitor with method `failed()` that was never called. Now it's called when all the retires have failed. 

Fixes issue #4582

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/helpers/TransactionTemplate.java
+++ b/community/kernel/src/main/java/org/neo4j/helpers/TransactionTemplate.java
@@ -132,8 +132,7 @@ public class TransactionTemplate
         } );
     }
 
-    public <T> T execute( Function<Transaction, T> txFunction )
-            throws TransactionFailureException
+    public <T> T execute( Function<Transaction, T> txFunction ) throws TransactionFailureException
     {
         Throwable txEx = null;
         for ( int i = 0; i < retries; i++ )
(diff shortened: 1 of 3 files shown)


chrisvest (changes requested):
I would like to see a test that shows the existing and new behaviours working.

klaren:
@chrisvest Ready for another round.

community/kernel/src/main/java/org/neo4j/helpers/TransactionTemplate.java
```diff
@@ -100,29 +125,47 @@ public TransactionTemplate( GraphDatabaseService gds, Monitor monitor, int retri
...
+        Objects.requireNonNull( gds );
         return new TransactionTemplate( gds, monitor, retries, backoff, retryPredicate );
     }
 
     public TransactionTemplate retries( int retries )
     {
+        if ( retries < 1 )
+        {
+            throw new IllegalArgumentException( "Number of retries must be greater than or equal to 1" );
+        }
```
chrisvest: Oh, so the number of retries actually includes the first try as well.

community/kernel/src/main/java/org/neo4j/helpers/TransactionTemplate.java
```diff
@@ -81,13 +93,26 @@ public void retrying()
...
+     * <p>
+     * Default exceptions to retry on is everything except {@link Error} and {@link TransactionTerminatedException}.
+     */
     public TransactionTemplate()
     {
-        this( null, new Monitor.Adapter(), 0, 0, any(
-                Predicates.<Throwable>instanceOf( Error.class ),
-                Predicates.<Throwable>instanceOf( TransactionTerminatedException.class ) ).negate() );
+        this( null, new Monitor.Adapter(), 0, 0,
+                ex -> !Error.class.isInstance( ex ) && !TransactionTerminatedException.class.isInstance( ex ) );
```
chrisvest: This violates the invariant that `retries` should be greater than 0.

community/kernel/src/main/java/org/neo4j/helpers/TransactionTemplate.java
```diff
@@ -81,13 +93,26 @@ public void retrying()
...
+     *
+     * @param gds graph database to execute on.
+     * @param monitor a monitor that can react to events.
+     * @param retries number of retries to try before failing.
+     * @param backoff milliseconds to wait between each retry.
+     * @param retryPredicate what {@link Throwable}'s to retry on.
+     */
     public TransactionTemplate( GraphDatabaseService gds, Monitor monitor, int retries,
                                 long backoff, Predicate<Throwable> retryPredicate )
     {
```
chrisvest, line 123: This doesn't check any invariants.

community/kernel/src/main/java/org/neo4j/helpers/TransactionTemplate.java
klaren: Passing 0 will result in an instant failure in `execute` due to the for loop. I should probably patch that now that you mention it.

klaren:
@chrisvest I think everything is addressed now.

chrisvest (approved):
Looks good!

PR #10313, opened 2017-10-18
Make the `dbms.checkpoint.iops.limit` setting dynamic so it can be changed at runtime with the `dbms.setConfigValue(setting, value)` procedure

chrisvest:


community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -571,42 +576,29 @@ public void setLogger( Log log )
      * @implNote No migration or config validation is done. If you need this you have to refactor this method.
      *
      * @param setting The setting to set to the specified value.
-     * @param newValue The new value to set, passing {@code null} or empty should reset the value back to default value.
+     * @param update The new value to set, passing {@code null} or the empty string should reset the value back to default value.
      * @throws IllegalArgumentException if the provided setting is unknown or not dynamic.
      * @throws InvalidSettingException if the value is not formatted correctly.
      */
-    public void updateDynamicSetting( String setting, String newValue ) throws IllegalArgumentException, InvalidSettingException
+    public void updateDynamicSetting( String setting, String update )
```
klaren, line 583: After more closer inspections I believe we have to synchronize the whole method. `newValue` and `oldValue` pair still races and the main reason we log the change is to know what's the current state is. Without the synchronization, the log messages are not guaranteed to reflect that since the order of log messages is not coupled with the order of the changes.

community/kernel/src/test/java/org/neo4j/kernel/configuration/ConfigTest.java
```diff
@@ -364,18 +372,79 @@ public void augmentDefaults() throws Exception
...
-        config.updateDynamicSetting( MyDynamicSettings.boolSetting.name(), "true" );
-        config.updateDynamicSetting( MyDynamicSettings.boolSetting.name(), "" );
+        config.updateDynamicSetting( settingName, "false" );
+        config.updateDynamicSetting( settingName, "true" );
+        config.updateDynamicSetting( settingName, "" );
 
-        verify( log ).info("Setting changed: '%s' changed from '%s' to '%s'", MyDynamicSettings.boolSetting.name(), "true", "false" );
-        verify( log ).info("Setting changed: '%s' changed from '%s' to '%s'", MyDynamicSettings.boolSetting.name(), "false", "true" );
-        verify( log ).info("Setting changed: '%s' changed from '%s' to '%s'", MyDynamicSettings.boolSetting.name(), "true", "true" );
+        InOrder order = inOrder( log );
```
klaren, line 386: Nice catch :+1:

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/transaction/log/checkpoint/ConfigurableIOLimiter.java
```diff
@@ -32,41 +32,82 @@
 
 public class ConfigurableIOLimiter implements IOLimiter
 {
-    private static final AtomicIntegerFieldUpdater<ConfigurableIOLimiter> disableCountUpdater =
-            AtomicIntegerFieldUpdater.newUpdater( ConfigurableIOLimiter.class, "disabledCount" );
+    private static final AtomicLongFieldUpdater<ConfigurableIOLimiter> stateUpdater =
```
klaren, line 35: Do we still need this? `state` could as well be an `AtomicLong` since we always access if through the updater, right?
Or line 133 should just get the variable without the updater
```
133:     long state = this.state;
```
since I'm guessing the original reason to use the updater is to get rid of the overhead from the `get()` call.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/transaction/log/checkpoint/ConfigurableIOLimiter.java
```diff
@@ -32,41 +32,82 @@
...
+        long oldState;
+        long newState;
         if ( iops == null || iops < 1 )
         {
-            iopq = NO_LIMIT;
-            disabledCount = 1; // This permanently disables IO limiting
+            do
+            {
+                oldState = stateUpdater.get( this );
+                int disabledCounter = getDisabledCounter( state );
```
klaren: I think it would be cleaner to derive the new value from the `oldState` just retrieved. But that's just a matter of taste, feel free to ignore.

klaren (changes requested):
Overall this looks good, nice addition to the dynamic settings! :tada:

The issue related to the dynamic setting update is not directly linked with this PR but since synchronization was intruded here we could as well discuss it now.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/transaction/log/checkpoint/ConfigurableIOLimiter.java
chrisvest: It's to avoid the object indirection of the AtomicLong.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/transaction/log/checkpoint/ConfigurableIOLimiter.java
chrisvest: Good catch. It was not supposed to be like that.

chrisvest:
@klaren I believe I've addressed you comments. Please have another look.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/transaction/log/checkpoint/ConfigurableIOLimiter.java
```diff
@@ -116,13 +158,31 @@ public long maybeLimitIO( long previousStamp, int recentlyCompletedIOs, Flushabl
     @Override
     public void disableLimit()
     {
-        disableCountUpdater.getAndIncrement( this );
+        long currentState;
+        long newState;
+        do
+        {
+            currentState = stateUpdater.get( this );
```
MishaDemianenko, line 165: it's not a problem now but it can be in future: if we will have 2 calls to disable limit we will need to have 2 calls to enable limit which I would say is a bit unexpected behaviour

MishaDemianenko (approved):
Only one small note

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/transaction/log/checkpoint/ConfigurableIOLimiter.java
chrisvest: That is by design, actually. This call is used to temporarily disable the limit for the purpose of speeding up flushing. For instance, copying a store over the network requires a recent check-point, which means that the check-point process now suddenly comes under pressure to complete as soon as possible. If, say, we have both an online back-up being taken, and a new member joins the cluster, then two store copy processes will be started and they will both call `disableLimit`, and when they have their check-point they will both `enableLimit`. This way, the check-pointer won't slow down again until the last urgent need for a check-point has been serviced.

It might not be obvious from the method name, but the javadocs explains it.

MishaDemianenko: Makes sense thx!

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/transaction/log/checkpoint/ConfigurableIOLimiter.java
klaren: That's what I mean, you can omit all the `stateUpdater.get()` calls and just read it with an assignment. Then you could get rid of the `SuppressWarning` as well.

chrisvest: There's not really an indirection with `stateUpdater` because it's `static final`. I like the symmetry in the code as is.

klaren (approved):
Changes looks good!

PR #10314, opened 2017-10-18
Don't delete known_host

pontusmelke:
This is no longer necessary.


lassewesth:
Approved @pontusmelke.

lutovich:
@pontusmelke should the target branch be 3.1?

pontusmelke:
@lutovich The test isn't there in 3.0 so this is as far back as it can go?

lutovich:
@pontusmelke target branch is 3.4 right now. I'll change it to 3.1.

pontusmelke:
@lutovich Ah sorry, my bad!

PR #10318, opened 2017-10-18
Procedure to list all transactions currently executing and visible to the user.

MishaDemianenko:
Introduce new "dbms.listTransactions" procedure that allows listing of
all active transactions, observe their statuses and consumed resources.
You can see how procedure result looks like on attached screenshot.
Result table contains the following information:
"transactionId" - unique transaction id.
"username" - transaction creator user name.
"metaData" - transactional metadata.
"startTime" - transaction start time.
"protocol" - a protocol that was used to create the transaction. It currently comes from the query that transaction is executing so it can be empty if the transaction does not have a query to execute.
"clientAddress" - client that created transaction. Same as above: can be empty if there is no current query.
"requestUri" - client origin request URI that started the transaction. Same as above: can be empty if there is no current query.
"currentQueryId" - current executing query id.
"currentQuery" - current executing query text.
"activeLockCount" - transaction active lock count.
"status" - transaction status. Possible values are: Running, Terminated with reason: 'termination reason', Blocked by: 'list of transactions that block execution'.
"resourceInformation" - blocking resource information. Contains description what transaction is waiting for, if it's blocked. 
"elapsedTimeMillis" - a number of elapsed milliseconds after transaction start.
"cpuTimeMillis" - a number of milliseconds it CPU has to spend to execute the current transaction.
"waitTimeMillis" - a number of milliseconds transaction was waiting for various resources.
"idleTimeMillis" - a number of milliseconds current transaction was idle. (Elapsed time minus CPU time millis and waiting time millis)
"allocatedBytes" - a number of bytes that transaction allocated.
"pageHits" - a number of page cache hits for the current transaction.
"pageFaults" - a number of page cache faults for the current transaction.


MishaDemianenko:
screen shot 2017-10-19 at 12 34 23

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/DefaultPageCursorTracer.java
```diff
@@ -64,10 +66,12 @@ public void reportEvents()
         if ( hits > 0 )
         {
             pageCacheTracer.hits( hits );
+            historicalHits = Math.addExact( historicalHits, hits );
         }
         if ( faults > 0 )
         {
             pageCacheTracer.faults( faults );
+            historicalFaults = Math.addExact( historicalFaults, faults );
```
chrisvest: Is it better to throw an exception than to overflow in these cases?

chrisvest: I don't think signed long overflow is going to be a concern for these cases in practice.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -720,9 +720,20 @@ private NeoStoreKernelModule buildKernel( TransactionAppender appender,
...
+        CpuClock cpuClock = CpuClock.NOT_AVAILABLE;
+        if ( config.get( GraphDatabaseSettings.track_query_cpu_time ) )
+        {
+            cpuClock = CpuClock.CPU_CLOCK;
+        }
+        HeapAllocation heapAllocation = HeapAllocation.NOT_AVAILABLE;
+        if ( config.get( GraphDatabaseSettings.track_query_allocation ) )
+        {
+            heapAllocation = HeapAllocation.HEAP_ALLOCATION;
+        }
```
chrisvest, line 670: Maybe we should have a card about making these settings dynamic. What do you think?

community/kernel/src/main/java/org/neo4j/kernel/api/query/ExecutingQueryStatus.java
```diff
@@ -51,4 +55,22 @@ boolean isPlanning()
     {
         return false;
     }
+
+    /**
+     * Is query waiting on a locks
+     * @return true if waiting on a locks, false otherwise
```
chrisvest: It reads better if we don't put `a` in front of `locks` here.

community/kernel/src/main/java/org/neo4j/kernel/api/query/ExecutingQueryStatus.java
```diff
@@ -51,4 +55,22 @@ boolean isPlanning()
...
+     */
+    boolean isWaitingOnLocks()
+    {
+        return false;
+    }
+
+    /**
+     * List of locks query is waiting on. Will be empty for all of the statuses except
+     * @return list of locks query is waiting on, empty list if query is not waiting.
```
chrisvest: Will be empty for all of the statuses exception what?

community/kernel/src/main/java/org/neo4j/kernel/impl/util/MonotonicCounter.java
```diff
@@ -32,9 +32,17 @@ static MonotonicCounter newAtomicMonotonicCounter()
...
-                return value.incrementAndGet();
+                int initialValue;
+                int incrementedValue;
+                do
+                {
+                    initialValue = value.intValue();
+                    incrementedValue = initialValue < 0 ? 0 : initialValue + 1;
+                }
+                while ( !value.compareAndSet( initialValue, incrementedValue ) );
+                return incrementedValue;
```
chrisvest: This will produce a sequence of 1, 2, 3 … Integer.MAX_VALUE, Integer.MIN_VALUE, 0, 1, 2, 3 …

Looks like a bug. Why not just keep the old implementation?

enterprise/kernel/src/main/java/org/neo4j/kernel/enterprise/builtinprocs/TransactionId.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final String USER_TRANSACTION_ID_PREFIX = "transaction-";
+
+    private TransactionId()
+    {
+    }
+
+    static String ofTransactionHandle( KernelTransactionHandle transaction )
+    {
+        return USER_TRANSACTION_ID_PREFIX + transaction.getUserTransactionId();
+    }
```
chrisvest: Maybe this can be a `name` function or something like that, on the `KernelTransactionHandle`. The name of this class clashes with another one in `kernel.impl.store`.

chrisvest (changes requested):
I had a couple of comments.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
MishaDemianenko: cardified

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/DefaultPageCursorTracer.java
MishaDemianenko: yeah exception here is not nice and anyway agree with you that it should not be the case ever anyway

community/kernel/src/main/java/org/neo4j/kernel/impl/util/MonotonicCounter.java
MishaDemianenko: bug fixed, test introduced. I want to make sure that tx ids or query ids are always correct and guarantee that they will never be negative, even in this unlikely, but still possible case.

MishaDemianenko:
@chrisvest updated

community/kernel/src/main/java/org/neo4j/kernel/impl/util/MonotonicCounter.java
MishaDemianenko: no, you are obviously right that's too unrealistic as well will restore original counter thingy

chrisvest (approved):

PR #10328, opened 2017-10-19
Cleanups around Config and friends

chrisvest:
The purpose of this exercise is greater enforcement of having just _one_ `Config` instance at runtime, since with dynamic settings it is now more important that everyone uses the same `Config` instance.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdGeneratorImpl.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdGeneratorImpl.java
@@ -262,11 +262,33 @@ public class IdGeneratorImpl implements IdGenerator
         IdContainer.createEmptyIdFile( fs, fileName, highId, throwIfFileExists );
     }
 
+    /**
+     * Read the high-id count from the given id-file.
+     *
+     * Note that this method should only be used when the file is not currently in use by an IdGenerator, since this
+     * method does not take any in-memory state into account.
+     *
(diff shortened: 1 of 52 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdGeneratorImpl.java
```diff
@@ -267,6 +267,11 @@ public static long readHighId( FileSystemAbstraction fileSystem, File file ) thr
         return IdContainer.readHighId( fileSystem, file );
     }
 
+    public static long readDefragCount( FileSystemAbstraction fileSystem, File file ) throws IOException
```
klaren: I would like to see some javadoc specifying the limitations here, i.e. it's not safe to run when the `FreeIdKeeper` is active since it has internal buffers not accounted for.

klaren (approved):
Looking good! Only one small comment.

Another solution to the `Listeners` problem would be to remove the annotations instead. That would actually make me a lot happier ;)

I see no reason for this going into 3.3, but I don't see any reason for it not to. Do whatever is easier for you.

klaren (dismissed):
Misclick

chrisvest:
I addressed Anton's comments. I think it'd be nice to get the `ListenersTest` fixed up in 3.3, and the rest is pretty harmless. What do you think, @MishaDemianenko ?

MishaDemianenko (approved):
@chrisvest agree, let's fix it in 3.3

PR #10331, opened 2017-10-20
Remove the logical plan Argument

systay:
At run time, we do not differentiate between SingleRow and Argument. So,
our code can be simplified by just removing Argument and not have
special planner code for it


enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/SlotAllocation.scala
```diff
@@ -113,15 +113,12 @@ object SlotAllocation {
...
-      case _: Argument =>
-        argument.getOrElse(throw new InternalException("Found argument without Apply"))
-
       case leaf: NodeLogicalLeafPlan =>
         val pipeline = argument.getOrElse(PipelineInformation.empty)
         pipeline.newLong(leaf.idName.name, nullable, CTNode)
         pipeline
 
-      case SingleRow() =>
+      case _:SingleRow =>
```
craigtaverner, line 121: Previous behaviour for Apply was to throw an exception here 'Found argument without Apply'. That has been changed. Is that correct? I guess that assertion was only to catch presumably impossible cases where Argument was used like SingleRow, and now Argument can be used like SingleRow, so this differentiation disappears.

craigtaverner (approved):

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/SlotAllocation.scala
systay: Yep.

systay:
Replaced by #10362

PR #10332, opened 2017-10-20
3.3 list comparability

Lojjs:
Changed behaviour of list comparisons with null to align with openCypher

### DIFF
--- a/community/values/src/main/java/org/neo4j/values/storable/ArrayValue.java
+++ b/community/values/src/main/java/org/neo4j/values/storable/ArrayValue.java
@@ -20,6 +20,7 @@
 package org.neo4j.values.storable;
 
 import java.util.Iterator;
+import java.util.NoSuchElementException;
 
 import org.neo4j.values.AnyValue;
 import org.neo4j.values.SequenceValue;
@@ -51,7 +52,11 @@ public abstract class ArrayValue extends Value implements SequenceValue
             @Override
             public AnyValue next()
             {
-                return value( offset );
+                if ( !hasNext() )
+                {
+                    throw new NoSuchElementException();
+                }
+                return value( offset++ );
             }
         };
     }
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NullListAcceptanceTest.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NullListAcceptanceTest.scala
@@ -1,0 +1,172 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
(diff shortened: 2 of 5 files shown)


community/values/src/main/java/org/neo4j/values/storable/ArrayValue.java
```diff
@@ -51,7 +52,11 @@ public boolean hasNext()
             @Override
             public AnyValue next()
             {
-                return value( offset );
+                if ( !hasNext() )
+                {
+                    throw new NoSuchElementException();
+                }
+                return value( offset++ );
```
craigtaverner, line 59: This seems to fix a serious bug. The fact that it was not noticed before makes me wonder if this is untested, and perhaps even unused code?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NullListAcceptanceTest.scala
```diff
@@ -0,0 +1,172 @@
...
+
+class NullListAcceptanceTest extends ExecutionEngineFunSuite with CypherComparisonSupport {
+
+  // Changed behaviour to comply with opencypher in 3.3
+  val nullInListConfigOld: TestConfiguration = TestConfiguration(V2_3 -> V3_2, Planners.all, Runtimes.all) + Configs.AllRulePlanners
+
+  // Comparison between lists and non-lists
+
+  test("equality between list and literal should return null") {
+    val query = "WITH [1, 2] AS l1, 'foo' AS l2 RETURN l1 = l2 AS res"
```
craigtaverner, line 34: Why is this null and not false?

craigtaverner (changes requested):

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NullListAcceptanceTest.scala
Lojjs: According to the CIP: "Lists are incomparable to any value that is not also a list." Incomparable seems to be used equivalent with "should evaluate to null"

community/values/src/main/java/org/neo4j/values/storable/ArrayValue.java
Lojjs: Definitely, me and @OliviaYtterbrink was also surprised that this had not been discovered before. We have talked with @fickludd about it (he has written this class) and he also thought it was a serious bug.

fickludd: This bug could be very bad, and it's great that it was found. The code clearly lacks some tests, but I don't think it's unused.

fickludd:
Me and @Lojjs just discussed whether it would make sense to move comparability to the values module, either encapsulated in it's own class as now, or as a new method on `AnyValue`, eg. `AnyValue.compareTo( AnyValue other )`

OliviaYtterbrink:
This has been replaced with https://github.com/neo4j/neo4j/pull/10411

PR #10333, opened 2017-10-20
Remove non existent services for test META-INF files.

MishaDemianenko:


klaren (approved):
Nice catch! We should probably just remove all those files and rely on the `Service` annotations to do the magic in a future PR though.

PR #10340, opened 2017-10-23
Cleanup HostnamePortTest URI creation assumptions.

MishaDemianenko:
Cleanup incorrect assumption about ability to create URI for any host
and port if scheme is not specified.
Since URI should always have scheme defined and according to RFC 3986
it should be defined as
"scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )"
And hosts that start with a digit will break that assumption.


MishaDemianenko:
suppressed by #10365

PR #10341, opened 2017-10-24
 Introduce enum for modes in which StoreChannel can be opened using FSA

MishaDemianenko:
Replace string constants by the newly introduced enum.


chrisvest (approved):
Good stuff.

PR #10347, opened 2017-10-26
Fix a bug that prevented page cache metrics from the Causal Cluster replication from being reported

chrisvest:
Ensure that page cache events from the core state applier (the data replication in Causal Cluster) publishes the page curser trace numbers


MishaDemianenko:
@chrisvest while reporting part of this pr is fine can you please point me to the place where CommandApplicationProcess or thread around it is using page cursors to make sure we have anything to report ever?

martinfurmanski (commented):
I don't get how it all wires together, but I'll trust you! An explanation would be nice though.

MishaDemianenko (approved):
Looks good except the question I have asked before? who will gonna update those counters in the first place?

chrisvest:
Looks like your question is really good, Misha. The pieces are not quite connecting, so more work is needed. Including adding an integration test.

chrisvest:
@martinfurmanski @MishaDemianenko I rewrote the commit, so it's all different from what you looked at before. Please give it another look-over. At least now there's an integration test that verifies that it actually works now. Also, paging metrics for the startup process are reported now.

MishaDemianenko (approved):

PR #10349, opened 2017-10-27
Transactional LogFiles abstraction

MishaDemianenko:
Group all of the transaction logs operations and functionality under newly introduced LogFiles interface.
Hide previously leaked abstraction of log file from modules that have nothing to do with transaction logs.
Introduce LogFilesBuilder to simplify configuration and access to log files for external modules.


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/BatchingTransactionAppender.java
```diff
@@ -59,7 +61,7 @@
 
     private final AtomicReference<ThreadLink> threadLinkHead = new AtomicReference<>( ThreadLink.END );
     private final TransactionMetadataCache transactionMetadataCache;
-    private final LogFile logFile;
+    private LogFile logFile;
```
tinwelint: Looks like it could still be `final`

tinwelint (commented):
I like it and better naming of those classes now. Had some minor comments

PR #10352, opened 2017-10-28
Unicode cleanup

pontusmelke:
Many of the cypher string functions were not handling unicode code points properly. 

For example:
- `size(string)` was returning the number of 16 bit numbers required for UTF16 encoding (`char[].length`)  and not the number of unicode code points. For example `RETURN size("💩")`
 was returning 2 instead of 1.
- `trim` only removed ascii whitespace characters whereas `ltrim` and `rtrim` removed all whitespace characters. The following invariant was thus not always true `trim(string) == ltrim(rtrim(string))`
- `substring` acted on 16 bit integers instead of code points meaning e.g `RETURN substring("🤓💩👻", 1, 1) != 💩` but instead some random string.
- `left` and `right` also was not working on code points but instead on the underlying `char[]`

This PR changes so that all cypher string functions deals with unicode code points instead of on items in the underlying `char[]`. Also adds dedicated methods for the above functions as well as `computeHash` and `compareTo` on `UTF8StringValue` meaning that we don't need to serialize java Strings in order to perform these operations.


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/Add.scala
```diff
@@ -25,8 +25,7 @@ import org.neo4j.cypher.internal.compatibility.v3_4.runtime.pipes.QueryState
 import org.neo4j.cypher.internal.util.v3_4.CypherTypeException
 import org.neo4j.cypher.internal.util.v3_4.symbols._
 import org.neo4j.values._
-import org.neo4j.values.storable.StringValue.UTF8StringValue
-import org.neo4j.values.storable._
+import org.neo4j.values.storable.{UTF8StringValue, _}
```
thobe: isn't this the same as `import org.neo4j.values.storable._` (unless some other package from which we import `_` also defines a `UTF8StringValue` - which they do not...)

Actually, what I'd really would prefer would be if we stopped using `_` imports altogether.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctions.scala
```diff
@@ -107,28 +122,24 @@ case class SubstringFunction(orig: Expression, start: Expression, length: Option
...
+      case text: TextValue =>
+        // if start goes off the end of the string, let's be nice and handle that.
+        val incomingLength = text.length
+        val startVal = Math.min(incomingLength, asInt(start(m, state)).value())
+
+        // if length goes off the end of the string, let's be nice and handle that.
+        val lengthVal = length match {
+          case None => incomingLength - startVal
+          case Some(func) => Math.min(incomingLength - startVal, asInt(func(m, state)).value())
+        }
```
thobe: Could we perhaps move this logic to the inside of `TextValue.substring`?
That way we can make it more efficient - since we wouldn't need to compute the length ahead of time for example...

Given that we want the methods on the `Value` types to support our use case for them anyhow, it seems like the right thing to do, that way we will be helped by having that logic implemented in one place when we use it from different runtime implementations.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctions.scala
```diff
@@ -177,14 +188,14 @@ case class SplitFunction(orig: Expression, separator: Expression)
     val separatorVal = asString(separator(m, state))
 
     if (origVal == null || separatorVal == null) {
-      Values.NO_VALUE
+      NO_VALUE
     } else {
       if (separatorVal.length > 0) {
-        VirtualValues.fromArray(Values.stringArray(split(Vector.empty, origVal, 0, separatorVal).toArray:_*))
```
thobe: This seems incredibly convoluted with conversions back and forth between scala iterables and java arrays. Even if the scala compiler by some weird chance happens to be able to optimize through this, I think we can do a great deal to improve the readability for human beings. How about we move the implementation of `split` to a method on `TextValue`?

But most likely that's too much for this particular pull request, since it is unrelated to the change at hand.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctions.scala
```diff
@@ -209,15 +220,20 @@ case class LeftFunction(orig: Expression, length: Expression)
...
-    val origVal = asString(orig(m, state))
-    val startVal = 0
-    val expectedLength = asInt(length(m, state)).value()
-    // if length goes off the end of the string, let's be nice and handle that.
-    val lengthVal = if (origVal.length < expectedLength + startVal)
-      origVal.length
-    else
-      expectedLength
-    Values.stringValue(origVal.substring(startVal, startVal + lengthVal))
+    val value = orig(m, state)
```
thobe: How does this overloading of `value` work? `value` is already defined as a parameter - so it at least confuses me as a reader of the code.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctions.scala
```diff
@@ -209,15 +220,20 @@ case class LeftFunction(orig: Expression, length: Expression)
...
+      case NO_VALUE => NO_VALUE
+      case origVal: TextValue =>
+        val startVal = 0
+        val expectedLength = asInt(length(m, state)).value()
+        // if length goes off the end of the string, let's be nice and handle that.
+        val lengthVal = if (origVal.length < expectedLength + startVal)
+          origVal.length
+        else
+          expectedLength
+        origVal.substring(startVal, startVal + lengthVal)
```
thobe: Again, I think that it is likely better to make this a method on `TextValue`.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctions.scala
```diff
@@ -209,15 +220,20 @@ case class LeftFunction(orig: Expression, length: Expression)
...
-    val expectedLength = asInt(length(m, state)).value()
-    // if length goes off the end of the string, let's be nice and handle that.
-    val lengthVal = if (origVal.length < expectedLength + startVal)
-      origVal.length
-    else
-      expectedLength
-    Values.stringValue(origVal.substring(startVal, startVal + lengthVal))
+    val value = orig(m, state)
+    value match {
+      case NO_VALUE => NO_VALUE
```
thobe: Is this not what is handled by `NullInNullOutExpression`?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctions.scala
```diff
@@ -232,12 +248,17 @@ case class RightFunction(orig: Expression, length: Expression)
   extends NullInNullOutExpression(orig) with NumericHelper {
 
   override def compute(value: AnyValue, m: ExecutionContext, state: QueryState): AnyValue = {
-    val origVal = asString(orig(m, state))
-    // if length goes off the end of the string, let's be nice and handle that.
-    val lengthVal = if (origVal.length < asInt(length(m, state)).value()) origVal.length
-    else asInt(length(m, state)).value()
-    val startVal = origVal.length - lengthVal
-    Values.stringValue(origVal.substring(startVal, startVal + lengthVal))
+    val value = orig(m, state)
```
thobe: Redefines `value` - it is also a parameter...

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctions.scala
```diff
@@ -232,12 +248,17 @@ case class RightFunction(orig: Expression, length: Expression)
...
+    value match {
+      case NO_VALUE => NO_VALUE
+      case origVal: TextValue =>
+        // if length goes off the end of the string, let's be nice and handle that.
+        val lengthVal = if (origVal.length < asInt(length(m, state)).value()) origVal.length
+        else asInt(length(m, state)).value()
+        val startVal = origVal.length - lengthVal
+        origVal.substring(startVal, startVal + lengthVal)
```
thobe: Again, how about we make this a method on `TextValue`?

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctionsTest.scala
```diff
@@ -54,7 +54,7 @@ class StringFunctionsTest extends CypherFunSuite {
     left("hello", 8) should equal(stringValue("hello"))
     left(null, 8) should equal(expectedNull)
     intercept[CypherTypeException](left(1042, 2))
-    intercept[StringIndexOutOfBoundsException](left("hello", -4))
+    intercept[IndexOutOfBoundsException](left("hello", -4))
```
thobe: Why are we changing this exception type? If we wanted to we should be able to use the same type as before, should we not?

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctionsTest.scala
```diff
@@ -119,6 +121,7 @@ class StringFunctionsTest extends CypherFunSuite {
     ltrim("  hello") should equal(stringValue("hello"))
     ltrim("  hello  ") should equal(stringValue("hello  "))
     ltrim(null) should equal(expectedNull)
+    ltrim("\u2009㺂࿝鋦毠\u2009") should equal(stringValue("㺂࿝鋦毠\u2009"))//Contains `thin space`
```
thobe: Could we do some sort of exhaustive testing for this? I.e. generate strings padded with all possible whitespaces?

community/values/src/main/java/org/neo4j/values/storable/CharValue.java
```diff
@@ -96,6 +97,39 @@ public int length()
...
+    {
+        assert start == 0;
+        assert end == 1;
+        return this;
+    }
+
+    @Override
+    public TextValue trim()
+    {
+        if ( Character.isWhitespace( value ) )
```
thobe, line 114: One thing I was thinking about, is that perhaps we should make CharValue encapsulate a (32 bit) codepoint rather than a (16 bit) `char`.

(I couldn't make that comment on the relevant line of this PR, so I added it to the place that made me think of it)

community/values/src/main/java/org/neo4j/values/storable/StringValue.java
```diff
@@ -55,7 +52,56 @@ public boolean equals( String x )
...
+        {
+            return 0;
+        }
+        int h = 1;
+        final int length = value.length();
+        for ( int offset = 0; offset < length; )
+        {
+            final int codePoint = value.codePointAt( offset );
+            h = 31 * h + codePoint;
+            offset += Character.charCount( codePoint );
```
thobe: I typically put the increment in the header of the loop. It requires declaring the `codePoint` variable there too, but I think it is worth it for the clarity of having the progression of the loop declared in the header.

community/values/src/main/java/org/neo4j/values/storable/StringValue.java
```diff
@@ -55,7 +52,56 @@ public boolean equals( String x )
     @Override
     public int computeHash()
     {
-        return value().hashCode();
+        //NOTE that we are basing the hash code on code points instead of char[] values.
```
thobe: That's a very interesting observation.

community/values/src/main/java/org/neo4j/values/storable/UTF8StringValue.java
```diff
@@ -0,0 +1,403 @@
...
+
+/*
+ * Just as a normal StringValue but is backed by a byte array and does string
+ * serialization lazily when necessary.
+ *
+ */
+public final class UTF8StringValue extends StringValue
+{
+    //0111 1111, used for removing HIGH BIT from byte
+    private static final int HIGH_BIT_MASK = 127;
```
thobe: How about using binary literals, and a javadoc comment:

    /** Used for removing the high order bit from byte. */
    private static final int HIGH_BIT_MASK = 0b0111_1111;

community/values/src/main/java/org/neo4j/values/storable/UTF8StringValue.java
```diff
@@ -0,0 +1,403 @@
...
+    //0100 000, used for detecting non-continuation bytes, e.g. 10xx xxxx
+    private static final int NON_CONTINUATION_BIT_MASK = 64;
```
thobe: Same thing here, I think this would look better using a javadoc comment and a binary literal:

    /** Used for detecting non-continuation bytes. For example {@code 0b10xx_xxxx}. */
    private static final int NON_CONTINUATION_BIT_MASK = 0b0100_0000;

community/values/src/main/java/org/neo4j/values/storable/UTF8StringValue.java
```diff
@@ -0,0 +1,403 @@
...
+    @Override
+    public <E extends Exception> void writeTo( ValueWriter<E> writer ) throws E
+    {
+        writer.writeUTF8( bytes, offset, length );
+    }
+
+    @Override
+    public boolean equals( Value value )
+    {
+        if ( value instanceof org.neo4j.values.storable.UTF8StringValue )
```
thobe: It seems strange that we would need a fully qualified name here.

community/values/src/main/java/org/neo4j/values/storable/UTF8StringValue.java
```diff
@@ -0,0 +1,403 @@
...
+    {
+        String s = value;
+        if ( s == null )
+        {
+            synchronized ( this )
+            {
+                s = value;
+                if ( s == null )
+                {
+                    s = value = new String( bytes, offset, length, StandardCharsets.UTF_8 );
```
thobe: I typically write `value = s = ...` - I'm not sure it matters, but that order seems more likely not to cause yet another volatile load.

community/values/src/main/java/org/neo4j/values/storable/StringValue.java
```diff
@@ -55,7 +52,56 @@ public boolean equals( String x )
...
+        {
+            final int codePoint = value.codePointAt( offset );
+            h = 31 * h + codePoint;
+            offset += Character.charCount( codePoint );
+        }
+        return h;
+    }
+
+    @Override
+    public TextValue substring( int start, int end )
```
thobe: I'm wondering if we should push the implementation of these methods down to the leaf types, so instead of `StringValue` providing a default implementation that `UTF8StringValue` overrides, we would have `UTF8StringValue` and `StringWrappingStringValue` each simply provide their own implementation.

thobe (approved):
Looks good. I had a few remarks, but most of those are better addressed in a separate changeset.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctions.scala
pontusmelke: I agree. I stopped at methods that requires regex, since that is trickier to implement for `UTF8StringValue`. We could of course just have a default implementation that uses strings, but leave that for later.

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctionsTest.scala
pontusmelke: I have added scalacheck test in `TextValueSpecification` thought it was better to do the exhaustive testing on the underlying values instead of on a specific runtime implementation

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/commands/expressions/StringFunctionsTest.scala
pontusmelke: Here we are relying on exceptions from the underlying implementation, it used to be an error coming form `java.lang.String.substring` now it is an error coming from invalid access of an array. Both are equally bad in my opinion we should throw something we control instead of just passing through underlying errors, but since `StringIndexOutOfBoundsException` extends `IndexOutOfBoundsException` I thought that this at least didn't make things any worse, and we could postpone a real fix to later work?

community/values/src/main/java/org/neo4j/values/storable/StringValue.java
pontusmelke: Do you mean like

```
 for ( int offset = 0, codePoint = value.codePointAt( offset ); offset < length;codePoint = value.codePointAt( offset ), offset += Character.charCount( codePoint ) )
{
  h = 31 * h + codePoint;
}
```
looks a little bit weird?

thobe: No, like this:

    for ( int offset = 0, codePoint; offset < length; offset += Character.charCount( codePoint ) )
    {
        codePoint = value.codePointAt( offset );
        h = 31 * h + codePoint;
    }

community/values/src/main/java/org/neo4j/values/storable/CharValue.java
pontusmelke: Yes let's do that as separate work

pontusmelke:
Think I did most of your suggestions the rest we can do as separate work. I have added cards for all of them i think.

community/values/src/main/java/org/neo4j/values/storable/UTF8StringValue.java
```diff
@@ -159,18 +159,18 @@ public int computeHash()
     @Override
     public TextValue substring( int start, int end )
     {
-        assert start > 0;
-        assert end > start && end < length();
+        int s = Math.min( start, length() );
+        int e = Math.min( end, length() );
```
thobe: Hmm... this still computes the length. And *twice* non the less. So no improvements there...

thobe (commented):
Have you verified that the removal of matching on `NO_VALUE` is safe? - my comment on that was more of a question than a statement.

Too bad that we didn't get any algorithmic improvement of the `substring` method by pushing the length down. Was this not possible, or is it just deferred?

fickludd:
I really enjoy seeing the coding, the thorough review, and the discussions in this PR. Also the meta process of cardifying vs. implementing change suggestions. Great work @pontusmelke and @thobe.

community/values/src/main/java/org/neo4j/values/storable/Values.java
```diff
@@ -48,7 +48,7 @@
     public static final Value MAX_STRING = Values.booleanValue( false );
     public static final BooleanValue TRUE = Values.booleanValue( true );
     public static final BooleanValue FALSE = Values.booleanValue( false );
-    public static final TextValue EMPTY_STRING = Values.stringValue( "" );
```
thobe, line 53: Interesting...

Really `Values.stringValue( value )` should be made to return `StringValue.EMPTY` if `value.isEmpty()`.

But that's for another changeset, and this is a good change in any case.

community/values/src/main/java/org/neo4j/values/storable/UTF8StringValue.java
```diff
@@ -158,21 +158,21 @@ public int computeHash()
...
     {
-        int s = Math.min( start, length() );
-        int e = Math.min( s + length, length() );
-        byte[] values = bytes;
-        if ( s >= e )
+        if ( start < 0 || length < 0)
         {
-            return StringValue.EMTPY;
+            throw new IndexOutOfBoundsException( "Cannot handle negative start index nor negative length" );
         }
```
thobe: We could possibly do a quick return of `StringValue.EMPTY` here if `length == 0`.

thobe (approved):
Very nice.

community/values/src/main/java/org/neo4j/values/storable/UTF8StringValue.java
```diff
@@ -0,0 +1,445 @@
...
+            hash = 31 * hash + codePoint;
+        }
+
+        return hash;
+    }
+
+    @Override
+    public TextValue substring( int start, int length )
+    {
+        if ( start < 0 || length < 0)
```
thobe: The style god demands that you sacrifice a whitespace and add it before the closing paren on this line.

community/values/src/main/java/org/neo4j/values/storable/UTF8StringValue.java
```diff
@@ -0,0 +1,445 @@
...
+                i++;
+                b = (byte) (b << 1);
+            }
+            count++;
+        }
+        if ( byteEnd < 0 )
+        {
+            byteEnd = len;
+        }
+        if (byteStart < 0 )
```
thobe: The style god demands that you sacrifice a whitespace and add it after the opening paren on this line.

community/values/src/test/java/org/neo4j/values/storable/UTF8StringValueTest.java
```diff
@@ -59,9 +141,65 @@ public void shouldHandleOffset()
...
+        UTF8StringValue value = utf8Value( "hello".getBytes( UTF_8 ) );
+
+        // Expect
+        exception.expect( IndexOutOfBoundsException.class );
+
+        // When
+        value.substring( 4, -3 );
     }
 }
+
```
thobe: The style god frowns upon the double blank lines and refuses to let you compile this code.

thobe (commented):
`checkstyle` had some comments on the code as well...

PR #10356, opened 2017-10-30
Move parameter extraction to a later phase

pontusmelke:
Parameter extraction and constant folding was not aligned meaning that for expressions
where constants were folded, like `RETURN 1 +3`, we didn't get any autoparameterization.


community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/CypherCompilerAstCacheAcceptanceTest.scala
```diff
@@ -127,6 +127,24 @@ class CypherCompilerAstCacheAcceptanceTest extends CypherFunSuite with GraphData
     counter.counts should equal(CacheCounts(hits = 1, misses = 1, flushes = 1))
   }
 
+  test("should not care about value in cache") {
```
thobe: I think we can come up with a better test title than this. This title could mean the exact opposite of what the test is asserting (although it could also mean what it is supposed to - it is simply too vague).

How about something like `Constant values in query should use same plan`?

thobe (approved):
Looks good to me. I only have a minor remark on test naming, but I'm guessing that problem is much wider than this particular case...

PR #10358, opened 2017-10-30
Turn off compiled runtime for multiple aggregations

pontusmelke:
The compiled runtime is not doing the right thing when having multiple
aggregations, e.g.
```
MATCH (a)-->(b)
RETURN a.prop, count(a.p1), count(b.p2)
```

This PR disables the compiled runtime from queries that contains more than one aggregation.

### DIFF
--- a/enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/LogicalPlanConverter.scala
+++ b/enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/LogicalPlanConverter.scala
@@ -31,7 +31,7 @@ import org.neo4j.cypher.internal.compiler.v3_2.planner.logical.plans._
 import org.neo4j.cypher.internal.compiler.v3_2.planner.logical.{SortDescription, plans}
 import org.neo4j.cypher.internal.compiler.v3_2.planner.{CantCompileQueryException, logical}
 import org.neo4j.cypher.internal.frontend.v3_2.Foldable._
-import org.neo4j.cypher.internal.frontend.v3_2.ast.Expression
+import org.neo4j.cypher.internal.frontend.v3_2.ast.{Expression, FunctionInvocation}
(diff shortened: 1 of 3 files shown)


enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/LogicalPlanConverter.scala
```diff
@@ -74,10 +75,16 @@ object LogicalPlanConverter {
...
-    case _: plans.Limit => true
-    //top is limit + sort
-    case _: plans.Top => true
+  private def aggregationNotSupported(p: plans.Aggregation) = {
+    val notValid = p.treeExists {
+      case _: plans.Limit => true
+      //top is limit + sort
+      case _: plans.Top => true
+    }
+    notValid && p.aggregationExpression.values.toList.treeCount {
```
craigtaverner: I would have thought this should be `||` instead of `&&` so that either of the failing conditions can cause the compiled runtime to not be used.

enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/codegen/LogicalPlanConverter.scala
```diff
@@ -127,7 +134,8 @@ object LogicalPlanConverter {
     override def consume(context: CodeGenContext, child: CodeGenPlan) = {
       val produceResultOpName = context.registerOperator(produceResults)
       val projections = produceResults.columns.map(c =>
-        c -> ExpressionConverter.createMaterializeExpressionForVariable(c)(context)).toMap
+                                                     c -> ExpressionConverter
```
craigtaverner: Was this reformatting necessary?

craigtaverner (changes requested):

pontusmelke:
@craigtaverner Nice catch, fixed it and removed the unnecessary reformatting

jexp:
@pontusmelke do we really need to turn it off instead of fixing it?

craigtaverner (approved):

PR #10359, opened 2017-10-30
Fixes bug in pattern matching around pruning var expand

systay:
When both sides of a variable path where already available,
pruning var-expand should not be used, since it disregards the already
bound end-side.


Lojjs (approved):
You seems to have missed adding the test in the compiled blacklist, fixed that for you.
Otherwise: 
Looks fine as fast as I can judge!

PR #10360, opened 2017-10-30
 Allow transaction logs to be located in a separate folder.

MishaDemianenko:
Previously transaction logs were always located in the same directory as a store.
This PR will allow changing that by specifying "logical_logs_location"
with a path to a folder where transaction logs should be stored.
This will allow placing transaction logs on a separate dedicated drive.

CC integration tests will always put transaction logs in separate directories.
HA should be able to copy store from another instance with custom
transactional logs directory and place them into the correct directory
for that specific instance.
Backup will put transaction logs together with store files and restore command will properly put them into configured transaction logs folder

### DIFF
--- a/enterprise/backup/src/main/java/org/neo4j/restore/RestoreDatabaseCommand.java
+++ b/enterprise/backup/src/main/java/org/neo4j/restore/RestoreDatabaseCommand.java
@@ -23,30 +23,35 @@ import java.io.File;
 import java.io.IOException;
 
 import org.neo4j.commandline.admin.CommandFailed;
+import org.neo4j.graphdb.factory.GraphDatabaseSettings;
 import org.neo4j.io.fs.FileSystemAbstraction;
 import org.neo4j.kernel.configuration.Config;
+import org.neo4j.kernel.impl.transaction.log.files.LogFiles;
+import org.neo4j.kernel.impl.transaction.log.files.LogFilesBuilder;
 import org.neo4j.kernel.impl.util.Validators;
 
 import static java.lang.String.format;
 import static org.neo4j.commandline.Util.checkLock;
-import static org.neo4j.dbms.DatabaseManagementSystemSettings.database_path;
+import static org.neo4j.graphdb.factory.GraphDatabaseSettings.database_path;
 
 public class RestoreDatabaseCommand
 {
     private FileSystemAbstraction fs;
     private final File fromPath;
     private final File databaseDir;
+    private final File transactionLogsDirectory;
     private String databaseName;
     private boolean forceOverwrite;
 
-    public RestoreDatabaseCommand( FileSystemAbstraction fs, File fromPath, Config config,
-                                   String databaseName, boolean forceOverwrite )
+    public RestoreDatabaseCommand( FileSystemAbstraction fs, File fromPath, Config config, String databaseName,
+            boolean forceOverwrite )
     {
         this.fs = fs;
         this.fromPath = fromPath;
         this.databaseName = databaseName;
         this.forceOverwrite = forceOverwrite;
         this.databaseDir = config.get( database_path ).getAbsoluteFile();
+        this.transactionLogsDirectory = config.get( GraphDatabaseSettings.logical_logs_location ).getAbsoluteFile();
     }
 
     public void execute() throws IOException, CommandFailed
@@ -75,6 +80,18 @@ public class RestoreDatabaseCommand
         checkLock( databaseDir.toPath() );
 
         fs.deleteRecursively( databaseDir );
-        fs.copyRecursively( fromPath, databaseDir );
+        if ( !databaseDir.equals( transactionLogsDirectory ) )
+        {
+            fs.deleteRecursively( transactionLogsDirectory );
+        }
+        LogFiles backupLogFiles = LogFilesBuilder.logFilesBasedOnlyBuilder( fromPath, fs ).build();
+        File[] files = fromPath.listFiles();
+        if ( files != null )
+        {
+            for ( File file : files )
+            {
+                fs.copyToDirectory( file, backupLogFiles.isLogFile( file ) ? transactionLogsDirectory : databaseDir );
+            }
+        }
     }
 }
--- a/enterprise/backup/src/main/java/org/neo4j/restore/RestoreDatabaseCommand.java
+++ b/enterprise/backup/src/main/java/org/neo4j/restore/RestoreDatabaseCommand.java
@@ -32,6 +32,7 @@ import org.neo4j.kernel.impl.util.Validators;
 
 import static java.lang.String.format;
 import static org.neo4j.commandline.Util.checkLock;
+import static org.neo4j.commandline.Util.isSameOrChildFile;
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.database_path;
 
 public class RestoreDatabaseCommand
@@ -80,7 +81,8 @@ public class RestoreDatabaseCommand
         checkLock( databaseDir.toPath() );
 
         fs.deleteRecursively( databaseDir );
-        if ( !databaseDir.equals( transactionLogsDirectory ) )
+
+        if ( !isSameOrChildFile( databaseDir, transactionLogsDirectory ) )
         {
             fs.deleteRecursively( transactionLogsDirectory );
         }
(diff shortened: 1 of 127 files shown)


MishaDemianenko:
@spacecowboy here is a PR that we talked about would be great if you can also take a look at it, thx in advance!

enterprise/backup/src/main/java/org/neo4j/restore/RestoreDatabaseCommand.java
```diff
@@ -75,6 +80,18 @@ public void execute() throws IOException, CommandFailed
         checkLock( databaseDir.toPath() );
 
         fs.deleteRecursively( databaseDir );
-        fs.copyRecursively( fromPath, databaseDir );
+        if ( !databaseDir.equals( transactionLogsDirectory ) )
```
spacecowboy: We should make sure that the paths are canonicalized (and symlinks are resolved) before doing this comparison. Ponder if for example:

```
dbDir = /a/b/c
txDir = /a/b/c/d/..
```

comparison would fail, and we'd try to delete the same dir twice. 

The canonicalization might be done elsewhere but it's best to be explicit *here* in my opinion

spacecowboy (changes requested):
First, it seems the PR description is out of date because absolute paths are allowed now and not only relative paths? (which of course it the right choice anyway)

Second, I had one opinion about the code itself (see comment).

## Things that should not need extra attention because having it as a regular config option is enough:

- Docker
- Anything packaging related

## Things that will need attention (separate PR or here):

- `neo4j-admin dump` needs to take transaction logs into account https://github.com/neo4j/neo4j/blob/fdb849ee398b9ffac009b5f09da4ad4b92457a03/community/dbms/src/main/java/org/neo4j/commandline/dbms/DumpCommand.java

- `neo4j-admin load` needs to take transaction logs into account https://github.com/neo4j/neo4j/blob/fdb849ee398b9ffac009b5f09da4ad4b92457a03/community/dbms/src/main/java/org/neo4j/commandline/dbms/LoadCommand.java

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/TransactionLogCatchUpWriter.java
```diff
@@ -52,14 +53,19 @@
     private long lastTxId = -1;
     private long expectedTxId;
 
-    TransactionLogCatchUpWriter( File storeDir, FileSystemAbstraction fs, PageCache pageCache,
-            LogProvider logProvider, long fromTxId, boolean asPartOfStoreCopy ) throws IOException
+    TransactionLogCatchUpWriter( File storeDir, FileSystemAbstraction fs, PageCache pageCache, Config config,
+            LogProvider logProvider, long fromTxId, boolean asPartOfStoreCopy, boolean keepTxLogsInStoreDir ) throws IOException
```
tinwelint, line 57: Would it be enough to only let the config own this decision, or do we explicitly need this `keepTxLogsInStoreDir` flag?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
```diff
@@ -264,6 +267,19 @@ public void registerEditionSpecificProcedures( Procedures procedures ) throws Ke
...
+    private LogFiles buildLocalDatabaseLogFiles( PlatformModule platformModule, FileSystemAbstraction fileSystem,
+            File storeDir )
+    {
+        try
+        {
+            return LogFilesBuilder.activeFilesBuilder( storeDir, fileSystem, platformModule.pageCache ).withConfig( config ).build();
+        }
+        catch ( IOException e )
+        {
+            throw new RuntimeException( e );
```
tinwelint, line 279: The caller throws `KernelException` right? Would it be better to transform the `IOException` into that instead?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/EnterpriseReadReplicaEditionModule.java
```diff
@@ -375,4 +381,17 @@ private static TopologyServiceRetryStrategy resolveStrategy( Config config )
...
+    private LogFiles buildLocalDatabaseLogFiles( PlatformModule platformModule, FileSystemAbstraction fileSystem,
+            File storeDir, Config config )
+    {
+        try
+        {
+            return LogFilesBuilder.activeFilesBuilder( storeDir, fileSystem, platformModule.pageCache ).withConfig( config ).build();
+        }
+        catch ( IOException e )
+        {
+            throw new RuntimeException( e );
```
tinwelint, line 394: Same here about potentially making it into a `KernelException`

enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyClient.java
```diff
@@ -333,6 +339,7 @@ private GraphDatabaseService newTempDatabase( File tempStore )
                 .setConfig( "dbms.backup.enabled", Settings.FALSE )
                 .setConfig( GraphDatabaseSettings.logs_directory, tempStore.getAbsolutePath() )
                 .setConfig( GraphDatabaseSettings.keep_logical_logs, Settings.TRUE )
+                .setConfig( GraphDatabaseSettings.logical_logs_location, tempStore.getAbsolutePath() )
```
tinwelint, line 342: What's the incentive to explicitly set the `logical_logs_location` to the store dir? (a) It's the default already and (b) why override it?

MishaDemianenko:
@spacecowboy thx a lot! PR updated!

tinwelint (changes requested):

spacecowboy (approved):

spacecowboy:
Dump and Load looks good to me

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/TransactionLogCatchUpWriter.java
MishaDemianenko: we need to set it explicitly since it can be the case that now we do whole store copy that will do transaction pulling in the end and we can't use the same shared config since its hidden under RemoteStore abstraction which is the same for case when we do whole store copy or only transaction pulling (in first case we need to put logs into correct configured place and in the second into the same directory)

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
MishaDemianenko: sorry not sure I get it, its invoked from the constructor of EnterpriseCoreEditionModule and it does not throw an exception.
what if I will create some custom exception for those two cases?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/EnterpriseReadReplicaEditionModule.java
MishaDemianenko: same as above

enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyClient.java
MishaDemianenko: Default value is based on "database_path" which is based on data_directory that is based on neo4j_home. And they are not defined for temp database and they don't need to be defined there also since in temp db we have a flat structure where we just put the store and tx logs into a directory without setting up data directory, db name directory, etc.
That's why we overwrite it with exact location of temp store directory

MishaDemianenko:
@tinwelint commented

MishaDemianenko:
@tinwelint ping ping

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/EnterpriseCoreEditionModule.java
tinwelint: Github shows diffs in mysterious ways, ignore this comment

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/EnterpriseReadReplicaEditionModule.java
tinwelint: Github shows diffs in mysterious ways, ignore this comment

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/TransactionLogCatchUpWriter.java
tinwelint: OK gotcha

tinwelint (approved):

PR #10361, opened 2017-10-31
Don't access database for impossible predicates

pontusmelke:
For queries like `MATCH (n) WHERE 1 > 5 RETURN n.prop` we can deduce that the result will be empty without touching the database. This PR accomplish that by rewriting `Selection` to `DropResult` for the cases where we know that there will be no results.

Also removes unnecessary selections coming from queries like `WHERE 1 > 0`.


systay (approved):
:1st_place_medal: 
Merge as soon as build is green

systay:
@pontusmelke that last commit looks fishy. did you mean to remove a few test files?

PR #10362, opened 2017-10-31
Remove the logical plan Argument

systay:
At run time, we do not differentiate between SingleRow and Argument. So,
our code can be simplified by just removing Argument and not have
special planner code for it


systay:
Moved the changes to the 3.4 branch

craigtaverner (approved):

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/PatternExpressionImplementationAcceptanceTest.scala
```diff
@@ -337,8 +337,8 @@ class PatternExpressionImplementationAcceptanceTest extends ExecutionEngineFunSu
 
     executeWith(Configs.Interpreted, "MATCH (n:A) RETURN (n)-[:HAS]->() as p",
       planComparisonStrategy = ComparePlansWithAssertion((planDescription) => {
-        planDescription.find("Argument") shouldNot be(empty)
-        planDescription.cd("Argument").arguments should equal(List(EstimatedRows(1)))
+        planDescription.find("SingleRow") shouldNot be(empty)
```
craigtaverner: How can we test for SingleRow when plans never print this in the plan description?

craigtaverner (changes requested):

craigtaverner (approved):

PR #10369, opened 2017-10-31
3.2 raft pre voting

andrewkerr9000:
Includes introduction of a visitor for raft messages used by message handling and marshalling (not unmarshalling as that is against the enum value, could be done with another visitor...). The visitor is safer (cannot forget to handle a message, cannot forget last break/else, no casting) and can be used in inheritance, but is a little more verbose and may be counter intuitive to someone not familiar with the pattern.

changelog: Implement a pre-election stage in Raft protocol
This should reduce the incidence of unnecessary elections caused by a disruptive follower, for example rejoining the cluster after a network partition.
This is a breaking change and must be explicitly enabled with `causal_clustering.enable_pre_voting=true`.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
@@ -37,6 +37,26 @@ import static org.neo4j.causalclustering.core.consensus.RaftMessages.Type.PRUNE_
 
 public interface RaftMessages
 {
+
+    interface Handler<T, E extends Exception>
+    {
+        T handle( Vote.Request request ) throws E;
+        T handle( Vote.Response response ) throws E;
+        T handle( PreVote.Request request ) throws E;
+        T handle( PreVote.Response response ) throws E;
+        T handle( AppendEntries.Request request ) throws E;
+        T handle( AppendEntries.Response response ) throws E;
+        T handle( Heartbeat heartbeat ) throws E;
+        T handle( LogCompactionInfo logCompactionInfo ) throws E;
+        T handle( HeartbeatResponse heartbeatResponse ) throws E;
+        T handle( Timeout.Election election ) throws E;
+        T handle( Timeout.Heartbeat heartbeat ) throws E;
+        T handle( NewEntry.Request request ) throws E;
+        T handle( NewEntry.BatchRequest batchRequest ) throws E;
+        T handle( PruneRequest pruneRequest ) throws E;
+    }
+
+    // Position is used to identify messages. Changing order will break upgrade paths.
     enum Type
     {
         VOTE_REQUEST,
@@ -59,12 +79,16 @@ public interface RaftMessages
         NEW_BATCH_REQUEST,
 
         PRUNE_REQUEST,
+
+        PRE_VOTE_REQUEST,
+        PRE_VOTE_RESPONSE,
     }
 
     interface RaftMessage extends Message
     {
         MemberId from();
         Type type();
+        <T, E extends Exception> T dispatch( Handler<T, E> handler ) throws E;
     }
 
     class Directed
@@ -140,6 +164,12 @@ public interface RaftMessages
             }
 
             @Override
+            public <T,E extends Exception> T dispatch( Handler<T,E> handler ) throws E
+            {
+                return handler.handle( this );
+            }
+
+            @Override
             public boolean equals( Object o )
             {
                 if ( this == o )
@@ -203,6 +233,12 @@ public interface RaftMessages
             }
 
             @Override
+            public <T,E extends Exception> T dispatch( Handler<T,E> handler ) throws E
+            {
+                return handler.handle( this );
+            }
+
+            @Override
             public boolean equals( Object o )
             {
                 if ( this == o )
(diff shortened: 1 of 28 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Follower.java
```diff
@@ -80,62 +78,239 @@ private static void handleLeaderLogCompaction( ReadableRaftState ctx, Outcome ou
...
+    private abstract static class Visitor implements RaftMessages.Visitor<Outcome, IOException>
+    {
+
+        protected final ReadableRaftState ctx;
+        protected final Log log;
+        protected final Outcome outcome;
+
+        public Visitor( ReadableRaftState ctx, Log log )
+        {
+
```
martinfurmanski: superfluous space here and other places

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
```diff
@@ -37,6 +37,26 @@
 
 public interface RaftMessages
 {
+
+    interface Visitor<T, E extends Exception>
```
martinfurmanski: Maybe call this a Handler or MessageHandler? And the respective handler methods handle( .. )?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
```diff
@@ -59,12 +79,16 @@
...
+
+        PRE_VOTE_REQUEST,
+        PRE_VOTE_RESPONSE,
     }
 
     interface RaftMessage extends Message
     {
         MemberId from();
         Type type();
+        <T, E extends Exception> T visit( Visitor<T, E> visitor ) throws E;
```
martinfurmanski: I usually call this "dispatch", as in dispatch message onto correct handler.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Candidate.java
```diff
@@ -38,124 +38,178 @@
...
-        Outcome outcome = new Outcome( CANDIDATE, ctx );
+        return message.visit( new Visitor( ctx, log ) );
+    }
 
-        switch ( message.type() )
-        {
-            case HEARTBEAT:
-            {
-                RaftMessages.Heartbeat req = (RaftMessages.Heartbeat) message;
+    private static class Visitor implements RaftMessages.Visitor<Outcome, IOException>
```
martinfurmanski: I wonder if this static inner class could be merged with the containing class.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Candidate.java
```diff
@@ -38,124 +38,178 @@
...
             }
+            return outcome;
+        }
 
-            default:
-                break;
+        @Override
+        public Outcome apply( RaftMessages.PreVote.Request request ) throws IOException
+        {
+            return outcome;
```
martinfurmanski: It's really good that we expose places were we do nothing, so that it doesn't get hidden.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/state/RaftState.java
```diff
@@ -47,32 +47,32 @@
...
 
     private MemberId leader;
     private Set<MemberId> votesForMe = new HashSet<>();
+    private Set<MemberId> preVotesForMe = new HashSet<>();
     private Set<MemberId> heartbeatResponses = new HashSet<>();
     private FollowerStates<MemberId> followerStates = new FollowerStates<>();
     private long leaderCommit = -1;
     private long commitIndex = -1;
     private long lastLogIndexBeforeWeBecameLeader = -1;
+    private boolean isPreElection = true; // Needed to form a cluster
```
martinfurmanski: Why is this needed? Is this for unit tests? Perhaps unit tests should generally run with pre-election support off then? ITs etc should be able to form clusters even with the pre-election protocol extension in effect.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/marshalling/RaftMessageEncoder.java
```diff
@@ -53,25 +53,67 @@ protected synchronized void encode( ChannelHandlerContext ctx,
         channel.putInt( message.type().ordinal() );
         memberMarshal.marshal( message.from(), channel );
 
-        if ( message instanceof RaftMessages.Vote.Request )
+        message.visit( new Visitor( marshal, memberMarshal, channel ) );
+    }
+
+    private static class Visitor implements RaftMessages.Visitor<Void, Exception>
```
martinfurmanski: Perhaps a better name than just Visitor. EncodingHandler?

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/PreElectionIT.java
```diff
@@ -0,0 +1,105 @@
...
+import org.neo4j.test.causalclustering.ClusterRule;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+public class PreElectionIT
+{
+    private long electionTimeout = 1;
```
martinfurmanski: Maybe just String electionTimeout = "1s"?

martinfurmanski (commented):
Looks good overall, but a few minor discussion points.

martinfurmanski:
CI runs indicate that files need to get the correct licenses in place. `mvn license:format` also probably fails `mvn checkstyle:check`

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
andrewkerr9000: I'll rename to `Handler`. The outer class name gives context as to what it is handling.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
andrewkerr9000: :+1:

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Candidate.java
andrewkerr9000: ReadableRaftState and Log are constructor parameters for the inner visitor class. If it weren't a class they would have to be passed into the dispatch method, which would make the visitor interface less general or rather more complex.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Follower.java
andrewkerr9000: :+1:

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/state/RaftState.java
andrewkerr9000: This is needed for integration tests. If a cluster member is not in pre-election state when pre-election support is enabled then it won't give a positive pre-vote response to a pre-vote request, and a real election won't occur.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/marshalling/RaftMessageEncoder.java
andrewkerr9000: The outer class name spaces the visitor, so it's clear that it refers to raft message encoding. I've renamed it to Handler.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/PreElectionIT.java
andrewkerr9000: I was going to reuse it as a long at some point.

I'm still working on these tests.

andrewkerr9000:
Squashed, hopefully ready to merge now.

Integration tests proving troublesome on TC. I've increased the timeout to 5 (!) minutes. One test passes on timeout, so will run for 5 (!) minutes.

andrewkerr9000:
Soak tests all pass reliably

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/state/RaftState.java
martinfurmanski: Hmm.. I think I still don't get why this needs to be set. The pre-election state should not have to be set to give a positive pre-vote response? And an instance seeking an election should be able to reach the isPreElection status quite quickly right?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/state/RaftState.java
```diff
@@ -46,32 +46,39 @@
...
         this.voteStorage = voteStorage;
         this.membership = membership;
         this.entryLog = entryLog;
         this.inFlightCache = inFlightCache;
-        log = logProvider.getLog( getClass() );
+        this.supportPreVoting = supportPreVoting;
+        this.log = logProvider.getLog( getClass() );
+
+        // Initial state
+        this.isPreElection = supportPreVoting;
```
andrewkerr9000, line 81: Hopefully this is a bit clearer

martinfurmanski (approved):

PR #10371, opened 2017-11-01
Calling properties on a null node without exception

### LINKED ISSUE #10353: "Expected to be a property container" error for certain Cypher statements after 3.1.2 -> 3.3.0 upgrade
I discovered that certain cypher queries were no longer working after an upgrade from 3.1.2 -> 3.3.0. the failures were the same with all clients
**Neo4j Version:** 3.3.0
**Operating System:** OSX / CentOS  
**API:** neo4j-rb / neo4j browser

### Steps to reproduce
1. Queries of the following form previously worked without issue:
```
MATCH (n:LBL1 {prop0:'foo'})-[]->(o:LBL1 WHERE EXISTS(o.prop1) WITH o OPTIONAL MATCH (o)-[:rel1]->(p:LBL2) WITH p OPTIONAL MATCH (p)-[:rel2]->(e:LBL3) WHERE p.prop2 IS NOT NULL RETURN p
```
2. This was working without issue in Neo4j 3.1.2. However it fails immediately in 3.3.0 with the following error:
```
Neo.ClientError.Statement.TypeError
Expected p to be a property container.
```
3. Removal of the final WHERE seems to fix the issue:
```
...WHERE p.prop2 IS NOT NULL...
```
4. It appears the combination of OPTIONAL MATCH + WITH + WHERE is causing the issue. 

### Expected behavior
Existing CYPHER queries should run without issue on 3.3.0, unless otherwise stated in the upgrade notes

### Actual behavior
Existing Cypher queries break with cryptic errors when they match a certain criteria

Lojjs:
Changelog: Fixes #10353 so that calling `exists` or `IS NOT null` on a property of a null node evaluates to null rather than throwing an exception

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/predicates/Predicate.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/predicates/Predicate.scala
@@ -157,6 +157,7 @@ case class PropertyExists(variable: Expression, propertyKey: KeyToken) extends P
       propertyKey.getOptId(state.query).exists(state.query.relationshipOps.hasProperty(pc.id, _)))
     case IsMap(map) => Some(map(state.query).get(propertyKey.name) != Values.NO_VALUE)
     case null => None
+    case v: Value if v.asObject() == null => None
     case _ => throw new CypherTypeException("Expected " + variable + " to be a property container.")
   }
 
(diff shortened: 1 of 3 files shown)


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/predicates/Predicate.scala
```diff
@@ -157,6 +157,7 @@ case class PropertyExists(variable: Expression, propertyKey: KeyToken) extends P
       propertyKey.getOptId(state.query).exists(state.query.relationshipOps.hasProperty(pc.id, _)))
     case IsMap(map) => Some(map(state.query).get(propertyKey.name) != Values.NO_VALUE)
     case null => None
```
pontusmelke: This should be 
```
case NO_VALUE => None
```
you can remove the `case null` probably just missed when refactored.

pontusmelke (changes requested):
Nice, just a tiny change

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/commands/predicates/Predicate.scala
Lojjs: Fixed in new commit

pontusmelke (approved):

PR #10374, opened 2017-11-02
Procedure to clear query caches

SaschaPeukert:
In certain (mostly distributed) scenarios it might be useful to be able to clear the query caches. This procedure now enables the admin to do it

changelog: Added procedure to clear query caches.


enterprise/cypher/cypher/src/main/java/org/neo4j/cypher/internal/javacompat/EnterpriseCypherEngineProvider.java
```diff
@@ -64,6 +66,14 @@ protected QueryExecutionEngine createEngine( Dependencies deps, GraphDatabaseAPI
         EnterpriseCompatibilityFactory compatibilityFactory =
                 new EnterpriseCompatibilityFactory( inner, queryService, kernelAPI, monitors, logProvider );
         deps.satisfyDependency( compatibilityFactory );
+        try
+        {
+            deps.resolveDependency( Procedures.class ).registerProcedure( QueryEngineProcedures.class );
```
pontusmelke: Shouldn't this be done here https://github.com/neo4j/neo4j/blob/3.4/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/EditionModule.java#L79

SaschaPeukert: I would have loved to place it there, but this would have introduced a dependency from kernel to cypher

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/javacompat/QueryEngineProcedures.java
```diff
@@ -0,0 +1,69 @@
...
+    @Context
+    public GraphDatabaseAPI graph;
+
+    @Context
+    public SecurityContext context;
+
+    @Description( "Clears all query caches." )
+    @Procedure( name = "dbms.clearQueryCaches", mode = DBMS )
+    public Stream<StringResult> clearAllQueryCaches()
+    {
```
OliviaYtterbrink: Why not place this procedure in the `BuiltInDbmsProcedures`?

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/javacompat/QueryEngineProcedures.java
```diff
@@ -0,0 +1,69 @@
...
+        if ( !context.isAdmin() )
```
OliviaYtterbrink: Also check that the credentials haven't expired.

OliviaYtterbrink (changes requested):
Good feature, just some things to look at.

enterprise/cypher/cypher/src/main/java/org/neo4j/cypher/internal/javacompat/EnterpriseCypherEngineProvider.java
OliviaYtterbrink: If you move the procedure you wouldn't need this procedure loading which would streamline it a bit.

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/javacompat/QueryEngineProcedures.java
SaschaPeukert: Same reason, I cannot move the procedure registration - I would create a dependency from Kernel (`BuiltInDbmsProcedures`) to Cypher (`QueryExecutionEngine`)

OliviaYtterbrink (approved):

PR #10378, opened 2017-11-03
New cache for in-flight Raft entries

martinfurmanski:
A highly efficient cache which only can keep a consecutive range of raft log entries
cached, allowing for a very simple and efficient circular buffer to be used. The cache
also allows putting a maximum bound on the amount of entries as well as their total combined
size.

The cache also exposes metrics so that cache hits, misses and total data can be observed.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
@@ -28,6 +28,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.logging.Level;
 
+import org.neo4j.causalclustering.core.consensus.log.cache.InFlightCacheFactory;
 import org.neo4j.configuration.Description;
 import org.neo4j.configuration.Internal;
 import org.neo4j.configuration.LoadableConfig;
@@ -123,6 +124,19 @@ public class CausalClusteringSettings implements LoadableConfig
     public static final Setting<List<AdvertisedSocketAddress>> initial_discovery_members =
             setting( "causal_clustering.initial_discovery_members", list( ",", ADVERTISED_SOCKET_ADDRESS ),
                     NO_DEFAULT );
+
+    @Description( "Type of in-flight cache." )
+    public static final Setting<InFlightCacheFactory.Type> in_flight_cache_type =
+            setting( "causal_clustering.in_flight_cache.type", options( InFlightCacheFactory.Type.class, true ),
+                    InFlightCacheFactory.Type.CONSECUTIVE.name() );
+
+    @Description( "The maximum number of entries in the in-flight cache." )
+    public static final Setting<Integer> in_flight_cache_max_entries =
+            setting( "causal_clustering.in_flight_cache.max_entries", INTEGER, "1024" );
+
+    @Description( "The maximum number of bytes in the in-flight cache." )
+    public static final Setting<Long> in_flight_cache_max_bytes =
+            setting( "causal_clustering.in_flight_cache.max_bytes", BYTES, "2G" );
 
     public enum DiscoveryType
     {
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.causalclustering.core.consensus.log.cache;
+
+import static java.lang.Math.floorMod;
+import static java.util.Arrays.fill;
+
+/**
+ * <pre>
+ * Design
+ *
+ * S: start index
+ * E: end index
+ *
+ * When S == E the buffer is empty.
+ *
+ * Examples:
+ *
+ *              S
+ *              v
+ * Empty   [ | | | | | | ]
+ *              ^
+ *              E
+ *
+ *
+ *                S
+ *                v
+ * Size 2  [ | | | | | | ]
+ *                    ^
+ *                    E
+ *
+ *
+ *                 S
+ *                 v
+ * Full   [ | | | | | | ]
+ *               ^
+ *               E
+ *
+ * New items are put at the current E, and then E is moved one step forward (circularly).
+ * The item at E is never a valid item.
+ *
+ * If moving E one step forward moves it onto S
+ * - then it knocks an element out
+ * - and S is also moved one step forward
+ *
+ * The S element has index 0.
+ * Removing an element moves S forward (circularly).
+ *
+ * @param <V> type of elements.
+ */
+public class CircularBuffer<V>
+{
+    private final int arraySize; // externally visible capacity is arraySize - 1
+    private Object[] elementArr;
+
+    private int S;
+    private int E;
+
+    CircularBuffer( int capacity )
+    {
+        if ( capacity <= 0 )
+        {
+            throw new IllegalArgumentException( "Capacity must be > 0." );
+        }
+
+        this.arraySize = capacity + 1; // 1 item as sentinel (can't hold entries)
+        this.elementArr = new Object[arraySize];
+    }
+
+    /**
+     * Clears the underlying buffer and fills the provided eviction array with all evicted elements.
+     * The provided array must have the same capacity as the circular buffer.
+     *
+     * @param evictions Caller-provided array for evictions.
+     */
+    public void clear( V[] evictions )
+    {
+        if ( evictions.length != arraySize - 1 )
+        {
+            throw new IllegalArgumentException( "The eviction array must be of the same size as the capacity of the circular buffer." );
+        }
+
+        int i = 0;
+        while ( S != E )
+        {
+            //noinspection unchecked
+            evictions[i++] = (V) elementArr[S];
+            S = pos( S, 1 );
+        }
+
+        S = 0;
+        E = 0;
+
+        fill( elementArr, null );
+    }
+
+    private int pos( int base, int delta )
+    {
+        return floorMod( base + delta, arraySize );
+    }
+
+    /**
+     * Append to the end of the buffer, possibly overwriting the
+     * oldest entry.
+     *
+     * @return any knocked out item, or null if nothing was knocked out.
+     */
+    public V append( V e )
+    {
+        elementArr[E] = e;
+        E = pos( E, 1 );
+        if ( E == S )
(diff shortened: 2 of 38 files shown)


lutovich:
@martinfurmanski we have Caffeine caching library in classpath right now. Cypher uses it for query plan caching. I'd recommend using it for this use-case as well instead of rolling out our own cache.

martinfurmanski:
@lutovich What makes you think it fits this use case?

lutovich:
@martinfurmanski this use-case is a cache of Raft entries and Caffeine is caching library. What makes you think it does not fit?

martinfurmanski:
@lutovich I have no idea if it fits or not, I've never used it. Caches can be very different.

martinfurmanski:
@lutovich As discussed off-line, I tried an implementation based on Caffeine and I ran into a few problems although it is almost a good fit. One problem is that I cannot support truncate() in a good way. The best approximation would be to just use invalidateAll(). The similar method prune() I could support by instead just using invalidate( key ) instead in the interface, since that is actually the ultimate usage. Caffeine does have support for weights of cache entries, but it has no way for me to retrieve a weight estimation similar to estimatedSize(), which is an element count estimation. This means that the cache size metric I have currently hooked in would not be possible.

ben-manes:
I don't know if Caffeine is a better, since you can leverage LSN ranges for an optimization. That means the frequency/recency algorithm may be less optimal than offset logic. But I can offer a small clarification.

> Caffeine does have support for weights of cache entries, but it has no way for me to retrieve a weight estimation similar to estimatedSize(), which is an element count estimation.

Please see `Cache.policy()` for methods that are only applicable for certain configurations. This way they do not pollute the API with unsupported methods or over complicate it. `Policy.Eviction#weightedSize()` in this case.

martinfurmanski:
@ben-manes Thanks for the clarification!

I find the weak-reference aspect very appealing as well, but I don't think I can live with instance equality. Any ideas on how to generally deal with that? Another cache in front from long -> Long comes to mind.

ben-manes:
Technically we could offer normal equality for weak keys as an option, without much overhead I think. We would just need to parameterize the internal methods and offer a `WeakKeyReference` that uses `equals`. I don't know what we'd call this flag or if it would be generally useful. It is kind of like Guava's `Interners.newWeakInterner()`.

Otherwise I don't have any great ideas. I know custom equality for keys/values has been requested, but cannot be done efficiently without forking the hash table. Those asks were when arrays are stored in the cache.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
```diff
@@ -124,6 +125,19 @@
             setting( "causal_clustering.initial_discovery_members", list( ",", ADVERTISED_SOCKET_ADDRESS ),
                     NO_DEFAULT );
 
+    @Description( "Type of in-flight cache." )
```
digitalstain, line 128: Generally settings are public static final. This is more of a consistency thing but i find it nice.

martinfurmanski: My mistake for avoiding the tried and true copy-n-paste!

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/ConsecutiveCache.java
```diff
@@ -0,0 +1,128 @@
...
+    private long firstIndex()
+    {
+        return endIndex - circle.size() + 1;
+    }
+
+    void put( long idx, V e, V[] evictions )
+    {
+        if ( idx < 0 )
+        {
+            throw new IllegalArgumentException( "Index must be >= 0" );
```
digitalstain: For debugging purposes, it is useful to also provide the actual value passed. Same goes for line 74 below.

digitalstain (commented):
Looks good to me.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * The S element has index 0.
+ * Removing an element moves S forward (circularly).
+ *
+ * @param <V> type of elements.
+ */
+public class CircularBuffer<V>
+{
+    private final int arraySize; // externally visible capacity is arraySize - 1
+    private Object[] elementArr;
```
andrewkerr9000, line 71: I think if you used an `ArrayList<V>` instead of an array you'd avoid the need to cast from `Object` to `V`.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
```diff
@@ -0,0 +1,180 @@
...
+    private int pos( int base, int delta )
+    {
+        return floorMod( base + delta, arraySize );
+    }
+
+    /**
+     * Append to the end of the buffer, possibly overwriting the
+     * oldest entry.
+     *
+     * @return any knocked out item, or null if nothing was knocked out.
```
andrewkerr9000, line 123: Maybe return an `Optional<V>` instead of a nullable?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
```diff
@@ -0,0 +1,180 @@
...
+    {
+        //noinspection unchecked
+        return (V) elementArr[pos( S, idx )];
+    }
+
+    public V remove()
+    {
+        if ( S == E )
+        {
+            return null;
```
andrewkerr9000, line 153: Likewise, maybe an `Optional.empty()`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/ConsecutiveCache.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+        else
+        {
+            circle.clear( evictions );
+            circle.append( e );
+            endIndex = idx;
+        }
+    }
+
+    V get( long idx )
```
andrewkerr9000, line 71: I'm going to suggest an Optional again.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
```diff
@@ -0,0 +1,180 @@
...
+        this.elementArr = new Object[arraySize];
+    }
+
+    /**
+     * Clears the underlying buffer and fills the provided eviction array with all evicted elements.
+     * The provided array must have the same capacity as the circular buffer.
+     *
+     * @param evictions Caller-provided array for evictions.
+     */
+    public void clear( V[] evictions )
```
andrewkerr9000, line 93: Why shouldn't this return a collection, rather than mutate its parameter?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/ConsecutiveInFlightCache.java
```diff
@@ -0,0 +1,168 @@
...
+    {
+        return cache.size();
+    }
+
+    private long sizeOf( RaftLogEntry entry )
+    {
+        return entry.content().hasSize() ? entry.content().size() : 0;
+    }
+
+    private void processEvictions()
```
andrewkerr9000, line 152: If we only care about the size in bytes of evicted entries should we return them at all? We could have the removal methods in the cache return the size, or to keep it generic have it accept map and reduce lambdas and return the output.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/InFlightCacheFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.neo4j.causalclustering.core.CausalClusteringSettings.in_flight_cache_max_bytes;
+import static org.neo4j.causalclustering.core.CausalClusteringSettings.in_flight_cache_max_entries;
+
+public class InFlightCacheFactory
+{
+    public static InFlightCache create( Config config, Monitors monitors )
+    {
+        Type type = config.get( CausalClusteringSettings.in_flight_cache_type );
+
+        switch ( type )
```
andrewkerr9000: This logic could be included in the enum as an overridden method

andrewkerr9000 (commented):
Looks great. I made comments, nothing major, of which the most significant is that I'd prefer a returned Optional to a nullable.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
martinfurmanski: It's purely for performance.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
martinfurmanski: Yes, but ArrayList would have to instead. The intention is to really cement the fixed length design.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
martinfurmanski: I don't necessarily dislike it, but it doesn't mix as neatly with the low-level design for me. I mostly view it as a matter of taste here.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/ConsecutiveInFlightCache.java
martinfurmanski: I don't see much harm in returning them. It is purely a separation of concerns for me. The circular buffer deals with the circular aspect and simple operations around that. The consecutive cache deals with aspects related to caching a sequence of entries. And lastly the in-flight cache is the adapter dealing with with top-level binding to the RaftLogEntry type, keeping track of sizes, synchronization, and any etc.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/InFlightCacheFactory.java
martinfurmanski: Good idea, will help me get rid of that pesky default case.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
andrewkerr9000: Would have to what?

I guess it's a choice between hiding the growability of a List or the lack of type safety of an Array in CircularBuffer.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
andrewkerr9000: In that case you could copy into `evictions` with a few calls to `System.arrayCopy` instead of in a for-loop, but it may not be worth writing the extra logic.

andrewkerr9000 (approved):
LGTM

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/log/cache/CircularBuffer.java
martinfurmanski: I meant that ArrayList would have to cast instead internally.

PR #10379, opened 2017-11-03
Add possibility to generate test coverage reports with jacoco

MishaDemianenko:
Introduce new "testCoverage" maven profile that will allow to generate
separate and aggregated test coverage and reports for each module.
Small cleanup.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@
     <lucene.version>5.5.0</lucene.version>
     <bouncycastle.version>1.53</bouncycastle.version>
     <generate-config-docs-phase>prepare-package</generate-config-docs-phase>
-    <test.runner.jvm.settings.additional></test.runner.jvm.settings.additional>
+    <test.runner.jvm.settings.additional/>
     <test.runner.jvm.settings>-Xmx2G -XX:+UseG1GC -XX:-OmitStackTraceInFastThrow -XX:+HeapDumpOnOutOfMemoryError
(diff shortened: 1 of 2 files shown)


klaren (approved):

pom.xml
```diff
@@ -686,6 +686,77 @@
...
+                            </goals>
+                        </execution>
+                        <execution>
+                            <id>merge</id>
+                            <phase>verify</phase>
+                            <goals>
+                                <goal>merge</goal>
+                            </goals>
+                            <configuration>
+                                <destFile>${project.build.directory}/jacoco-aggregated.exec</destFile>
```
lutovich, line 726: Should `jacoco-aggregated.exec` be added to `.gitignore`?

lutovich (commented):
Change looks good to me. Couple questions/suggestions:
 1) should usage example and output be documented somewhere? maybe in commit message?
 2) looks like jacoco maven plugin will not work with `DsequentialTests` because it sets `forkCount` to zero. do we care about this?

pom.xml
MishaDemianenko: project build directory is a target directory and those are ignored anyway already

MishaDemianenko:
@lutovich its only first step, the final goal is to have something that will report automatically on a PR what is coverage of a new code or diff in coverages with and without PR. So people will not gonna need to do anything themselves - thats why it's not that documented sort of things.
for 2 - correct I do not think we care much since our forking model works quite well

PR #10380, opened 2017-11-06
Fixing EXPAND INTO regression

pontusmelke:
This PR fixes a minor regression seen after the introduction of `AnyValue` into the runtime.

### DIFF
--- a/community/cypher/interpreted-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/ExpandIntoPipeTest.scala
+++ b/community/cypher/interpreted-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/ExpandIntoPipeTest.scala
@@ -69,22 +69,13 @@ class ExpandIntoPipeTest extends CypherFunSuite with PipeTestSupport {
           case _ =>
             val nodeId = invocationOnMock.getArgument[Long](0)
             val dir = invocationOnMock.getArgument[SemanticDirection](1)
-            if (nodeId == 0) {
-              dir match {
-                case SemanticDirection.INCOMING => Iterator.empty
-                case SemanticDirection.OUTGOING =>
-                  Iterator(relationship1, relationship2, relationship3, selfRelationship)
-
-                case SemanticDirection.BOTH => Iterator(relationship1, relationship2, relationship3, selfRelationship)
-              }
+            (nodeId, dir) match {
+              case (0, SemanticDirection.INCOMING) => Iterator.empty
+              case (0, _) => Iterator(relationship1, relationship2, relationship3, selfRelationship)
+              case (2, SemanticDirection.OUTGOING) => Iterator.empty
+              case (2, _) => Iterator(relationship1)
+              case (id, d)=> throw new AssertionError(s"I am only a mocked hack, not a real database. I have no clue about $id and $d")
             }
-            else if (nodeId == 2) {
-              dir match {
-                case SemanticDirection.INCOMING => Iterator(relationship1)
-                case SemanticDirection.OUTGOING => Iterator.empty
-                case SemanticDirection.BOTH => Iterator(relationship1)
-              }
-            } else ???
         }
       }
     })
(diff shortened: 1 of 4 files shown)


community/cypher/interpreted-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/ExpandIntoPipeTest.scala
```diff
@@ -60,13 +60,31 @@ class ExpandIntoPipeTest extends CypherFunSuite with PipeTestSupport {
...
+                case SemanticDirection.BOTH => Iterator(relationship1, relationship2, relationship3, selfRelationship)
+              }
+            }
+            else if (nodeId == 2) {
+              dir match {
+                case SemanticDirection.INCOMING => Iterator(relationship1)
+                case SemanticDirection.OUTGOING => Iterator.empty
+                case SemanticDirection.BOTH => Iterator(relationship1)
+              }
+            } else ???
```
fickludd: I think I'm not understanding something here... where is this new mocking used?

community/kernel/src/main/java/org/neo4j/kernel/impl/util/NodeProxyWrappingNodeValue.java
```diff
@@ -103,7 +105,7 @@ public MapValue properties()
         MapValue m = properties;
         if ( m == null )
         {
-            synchronized ( this )
+            synchronized ( propLock )
```
fickludd: I remember us talking about decoupling synchronization on this class, but when I tried to completely remove synchronization it still didn't solve the regression according to the benchmarks I ran, way back i september or something. Back then I probably messed something up, but I would love to see some benchmark result to confirm this fix!

community/kernel/src/main/java/org/neo4j/kernel/impl/util/RelationshipProxyWrappingEdgeValue.java
```diff
@@ -107,6 +108,20 @@ public NodeValue endNode()
...
+        if ( node instanceof NodeProxyWrappingNodeValue )
+        {
+            Node proxy = ((NodeProxyWrappingNodeValue) node).nodeProxy();
+            return ValueUtils.fromNodeProxy( relationship.getOtherNode( proxy ) );
+        }
+        else
+        {
+           return super.otherNode( node );
+        }
+    }
```
fickludd, line 122: Or is this it? Curious.

fickludd (changes requested):
Great that you are fixing this! 

It's somewhat hard for me to understand how this PR improves by just looking directly at code. Would like an explanation or some benchmark result to boost my confidence.

community/cypher/interpreted-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/ExpandIntoPipeTest.scala
pontusmelke: The test used to always return the same relationships regardless of input node so we got cases where the returned relationship neither started nor ended on the provided node. The test kind of worked since we were swallowing the null value but we no longer do that. So I changed the test to return the correct relationship given the input node and direction. The `???` is just lazy I guess, I could throw an assertion error instead perhaps?

community/kernel/src/main/java/org/neo4j/kernel/impl/util/NodeProxyWrappingNodeValue.java
pontusmelke: No this should be reverted, forgot about it. I reverted it in `RelationshipProxyWrappingValue`
but forgot about this one

community/kernel/src/main/java/org/neo4j/kernel/impl/util/RelationshipProxyWrappingEdgeValue.java
pontusmelke: No it is the line above that calls down to `proxy.getOtherNode` that removes the regression.

pontusmelke:
@fickludd 

Before:
```
ExpandInto.executePlan interpreted  2982.709 ± 75.934  ms/op
```
before


After:
```
ExpandInto.executePlan interpreted   2652.338 ± 58.353  ms/op
```
after


Also tidied up the hacky test a little bit and removed the changes in `NodeProxyWrappingNodeValue`

fickludd (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/util/RelationshipProxyWrappingEdgeValue.java
fickludd: Awesome.

PR #10381, opened 2017-11-06
Move string functions into TextValue

pontusmelke:
Instead of having `StringFunction.scala` implementing string functions we move the implementation of the rest of the string functions down to the `TextValue` implementation. All these methods currently require some form of regular expression support which is why these methods are added on the abstract class using the `value()` method which forces a string to be materialized.

### DIFF
--- a/community/cypher/interpreted-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/StringFunctionsTest.scala
+++ b/community/cypher/interpreted-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/StringFunctionsTest.scala
@@ -19,12 +19,11 @@
  */
 package org.neo4j.cypher.internal.runtime.interpreted.commands.expressions
 
-import org.neo4j.cypher.internal.runtime.interpreted.ExecutionContext
-import org.neo4j.cypher.internal.runtime.interpreted.QueryStateHelper
+import org.neo4j.cypher.internal.runtime.interpreted.{ExecutionContext, QueryStateHelper}
 import org.neo4j.cypher.internal.util.v3_4.CypherTypeException
 import org.neo4j.cypher.internal.util.v3_4.test_helpers.CypherFunSuite
 import org.neo4j.values.storable.Values
-import org.neo4j.values.storable.Values.{EMPTY_STRING, stringValue}
+import org.neo4j.values.storable.Values.{EMPTY_STRING, stringArray, stringValue}
 
 class StringFunctionsTest extends CypherFunSuite {
 
@@ -158,5 +157,17 @@ class StringFunctionsTest extends CypherFunSuite {
     reverse(null) should equal(expectedNull)
     reverse("\r\n") should equal(stringValue("\n\r"))
     reverse("\uD801\uDC37") should equal(stringValue("\uD801\uDC37"))
+  }
+
+  test("splitTests") {
+    def split(x: Any, y: Any) = SplitFunction(Literal(x), Literal(y))(ExecutionContext.empty, QueryStateHelper.empty)
+
+    split("HELLO", "LL") should equal(stringArray("HE", "O"))
+    split("Separating,by,comma,is,a,common,use,case", ",") should equal(stringArray("Separating", "by", "comma", "is", "a", "common", "use", "case"))
+    split("hello", "X") should equal(stringArray("hello"))
+    split("hello", null) should equal(expectedNull)
+    split(null, "hello") should equal(expectedNull)
+    split(null, null) should equal(expectedNull)
+    intercept[CypherTypeException](split(1024, 10))
   }
 }
--- a/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
+++ b/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
@@ -62,6 +62,12 @@ public abstract class StringValue extends TextValue
     }
 
     @Override
+    public TextValue toUpper()
+    {
+        return new StringWrappingStringValue( value().toUpperCase() );
+    }
+
+    @Override
     public Object asObjectCopy()
     {
         return value();
--- a/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
+++ b/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
@@ -68,6 +68,14 @@ public abstract class StringValue extends TextValue
     }
 
     @Override
+    public TextArray split( String separator )
+    {
+        String asString = value();
+        String[] split = asString.split( separator );
+        return Values.stringArray( split );
+    }
+
+    @Override
     public Object asObjectCopy()
     {
         return value();
--- a/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
+++ b/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
@@ -56,6 +56,12 @@ public abstract class StringValue extends TextValue
     }
 
     @Override
+    public TextValue toLower()
+    {
+        return new StringWrappingStringValue( value().toLowerCase() );
+    }
+
+    @Override
     public Object asObjectCopy()
     {
         return value();
--- a/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
+++ b/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
@@ -70,6 +70,7 @@ public abstract class StringValue extends TextValue
     @Override
     public TextArray split( String separator )
     {
+        assert separator != null;
         String asString = value();
         //Cypher has different semantics for the case where the separator
         //is exactly the value, in cypher we expect two empty arrays
@@ -85,7 +86,17 @@ public abstract class StringValue extends TextValue
     @Override
     public TextValue replace( String find, String replace )
     {
+        assert find != null;
+        assert replace != null;
+
         return Values.stringValue( value().replace( find, replace ) );
+    }
+
+    @Override
+    public TextValue reverse()
+    {
+        StringBuilder stringBuilder = new StringBuilder( value() );
+        return Values.stringValue( stringBuilder.reverse().toString() );
     }
 
     @Override
@@ -132,7 +143,7 @@ public abstract class StringValue extends TextValue
             }
             k += Character.charCount( c1 );
         }
-        return length() - other.length();
+        return len1 - len2;
     }
 
     static TextValue EMTPY = new StringValue()
@@ -176,7 +187,7 @@ public abstract class StringValue extends TextValue
         @Override
         public int compareTo( TextValue other )
         {
-            return Integer.compare( 0, other.length() );
+            return -other.length();
         }
 
         @Override
--- a/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
+++ b/community/values/src/main/java/org/neo4j/values/storable/StringValue.java
@@ -76,6 +76,12 @@ public abstract class StringValue extends TextValue
     }
 
     @Override
+    public TextValue replace( String find, String replace )
+    {
+        return Values.stringValue( value().replace( find, replace ) );
+    }
+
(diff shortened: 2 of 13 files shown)


community/values/src/main/java/org/neo4j/values/storable/Values.java
```diff
@@ -44,7 +44,7 @@
     public static final Value MAX_NUMBER = Values.doubleValue( Double.NaN );
     public static final Value ZERO_FLOAT = Values.doubleValue( 0.0 );
     public static final Value ZERO_INT = Values.longValue( 0 );
-    public static final Value MIN_STRING = Values.stringValue( "" );
+    public static final Value MIN_STRING =  StringValue.EMTPY;
```
fickludd: Extra whitespace

community/cypher/interpreted-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/StringFunctionsTest.scala
```diff
@@ -159,4 +158,16 @@ class StringFunctionsTest extends CypherFunSuite {
...
+    def split(x: Any, y: Any) = SplitFunction(Literal(x), Literal(y))(ExecutionContext.empty, QueryStateHelper.empty)
+
+    split("HELLO", "LL") should equal(stringArray("HE", "O"))
+    split("Separating,by,comma,is,a,common,use,case", ",") should equal(stringArray("Separating", "by", "comma", "is", "a", "common", "use", "case"))
+    split("hello", "X") should equal(stringArray("hello"))
+    split("hello", null) should equal(expectedNull)
+    split(null, "hello") should equal(expectedNull)
+    split(null, null) should equal(expectedNull)
+    intercept[CypherTypeException](split(1024, 10))
+  }
```
fickludd, line 172: As these TextValue methods are now moved to the values module, we also have to move/add tests for this functionality to the values module.  If we don't we could leave these methods untested when we remove the interpreted runtime for example.

community/values/src/main/java/org/neo4j/values/storable/StringValue.java
```diff
@@ -56,6 +56,39 @@ public boolean equals( String x )
...
+        return new StringWrappingStringValue( value().toUpperCase() );
+    }
+
+    @Override
+    public TextArray split( String separator )
+    {
+        String asString = value();
+        //Cypher has different semantics for the case where the separator
+        //is exactly the value, in cypher we expect two empty arrays
+        //where as java returns an empty array
```
fickludd: Thank you for adding this comment, that is a quite surprising behavior otherwise.

fickludd (changes requested):
This change make a lot of sense. I have two things that I would like to have fixed before merging:
1) remove that binary jfr recording please :)
2) make test follow the implementation into the values module

community/cypher/interpreted-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/StringFunctionsTest.scala
pontusmelke: Added more tests to `TextValueSpecification` plus added a new test `TextValueTest` that more or less test the same thing as this test.

pontusmelke:
@fickludd 
- 1 whoops!
- 2 added more tests

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
```diff
@@ -0,0 +1,141 @@
...
+    {
+        return asList( Values::stringValue, s -> utf8Value( s.getBytes( StandardCharsets.UTF_8 ) ) );
+    }
+
+    @Test
+    public void replace()
+    {
+        assertThat( value.apply( "hello" ).replace( "l", "w" ), equalTo( value.apply( "hewwo" ) ) );
+        assertThat( value.apply( "hello" ).replace( "ell", "ipp" ), equalTo( value.apply( "hippo" ) ) );
+        assertThat( value.apply( "hello" ).replace( "a", "x" ), equalTo( value.apply( "hello" ) ) );
```
fickludd, line 58: Worth putting in replacing with empty string?

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
```diff
@@ -0,0 +1,141 @@
...
+    {
+        assertThat( value.apply( "hello" ).substring( 2, 5 ), equalTo( value.apply( "llo" ) ) );
+        assertThat( value.apply( "hello" ).substring( 4, 5 ), equalTo( value.apply( "o" ) ) );
+        assertThat( value.apply( "hello" ).substring( 1, 3 ), equalTo( value.apply( "ell" ) ) );
+        assertThat( value.apply( "hello" ).substring( 8, 5 ), equalTo( StringValue.EMTPY ) );
+        assertThat( value.apply( "0123456789" ).substring( 1 ), equalTo( value.apply( "123456789" ) ) );
+        assertThat( value.apply( "0123456789" ).substring( 5 ), equalTo( value.apply( "56789" ) ) );
+        assertThat( value.apply( "0123456789" ).substring( 15 ), equalTo( StringValue.EMTPY ) );
+        assertThat( value.apply( "\uD83D\uDE21\uD83D\uDE21\uD83D\uDE21" ).substring( 1, 1 ),
+                equalTo( value.apply( "\uD83D\uDE21" ) ) );
```
fickludd: This assert is not very powerful. Because the same char is repeated one cannot tell which one was selected.

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
```diff
@@ -0,0 +1,141 @@
...
+        exception.expect( IndexOutOfBoundsException.class );
+        value.apply( "hello" ).substring( -4, 2 );
+    }
+
+    @Test
+    public void toLower()
+    {
+        assertThat( value.apply( "HELLO" ).toLower(), equalTo( value.apply( "hello" ) ) );
+        assertThat( value.apply( "Hello" ).toLower(), equalTo( value.apply( "hello" ) ) );
+        assertThat( value.apply( "hello" ).toLower(), equalTo( value.apply( "hello" ) ) );
```
fickludd, line 87: Could try empty string?

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Test
+    public void toUpper()
+    {
+        assertThat( value.apply( "HELLO" ).toUpper(), equalTo( value.apply( "HELLO" ) ) );
+        assertThat( value.apply( "Hello" ).toUpper(), equalTo( value.apply( "HELLO" ) ) );
+        assertThat( value.apply( "hello" ).toUpper(), equalTo( value.apply( "HELLO" ) ) );
```
fickludd, line 96: empty string?

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Test
+    public void reverse()
+    {
+        assertThat( value.apply( "Foo" ).reverse(), equalTo( value.apply( "ooF" ) ) );
+        assertThat( value.apply( "" ).reverse(), equalTo( StringValue.EMTPY ) );
+        assertThat( value.apply( " L" ).reverse(), equalTo( value.apply( "L " ) ) );
+        assertThat( value.apply( "\r\n" ).reverse(), equalTo( value.apply( "\n\r" ) ) );
+        assertThat( value.apply( "\uD801\uDC37" ).reverse(), equalTo( value.apply( "\uD801\uDC37" ) ) );
```
fickludd, line 135: We cannot see it actually reversing for UTF here. It is valuable to test the single char, but maybe also try a real word?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/TextValueSpecification.scala
```diff
@@ -68,6 +68,40 @@ object TextValueSpecification extends Properties("TextValue") with Configuration
...
+    val utf8 = utf8Value(x.getBytes(StandardCharsets.UTF_8))
+    val split = x.split(find)
+    if (x != find) {
+      stringArray(split: _*) == value.split(find) &&
+        value.split(find) == utf8.split(find)
+    } else {
+      value.split(find) == utf8.split(find) && value.split(find) == stringArray("", "")
+    }
+
+  }}
```
fickludd: I **do** appreciate this randomized testing!

fickludd (changes requested):
It's great that tests were added to the values module. I just have some nitpicks and suggestions on some of the assertions (which are quite good in general), and then we are done =).

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
pontusmelke: What do you mean by _reversing for UTF_? The last test is testing that we keep the code point `\uD801\uDC37` intact.

fickludd: Aye, but if you have two code points, they should be in reverse order?

pontusmelke:
@fickludd I will add more tests but I was under the impression that you wanted me to port the existing tests? Just to make sure you have seen it, we also have the scalacheck test in `TestValueSpecification`, that test actually gives me more confidence than just adding a few more cases in this file.

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
pontusmelke: No it is 𐐷,  http://unicode.scarfboy.com/?s=U%2B10437. \uD801 is a surrogate

fickludd: Ok, I'm not communicating well.

I understand that the current test check that the single code point `\uD801\uDC37` is not reversed to `\uDC37\uD801`, which would be a completely different unicode thing.

What I'm trying to suggest is to also add a multi-codepoint unicode string, to make sure that we actually can reverse these. E.g. `\uDC37\uD801\uDC2D\uD801 => \uDC2D\uD801\uDC37\uD801` or something.

fickludd:
@pontusmelke I saw `TestValueSpecification` and I agree that those generated test give a lot of credibility to the implementation. It's just annoying that they're at the other end of the code base (in cypher acceptance tests). I would prefer if the tests in the module of the implementation guaranteed it's correctness (at least for something basic like values). 

But if you do not agree we can merge it as is. This is a big improvement compared to before.

fickludd (approved):

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
pontusmelke: done below

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
pontusmelke: done below

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
pontusmelke: below

community/values/src/test/java/org/neo4j/values/storable/TextValueTest.java
pontusmelke: Something like this?

fickludd: 👍

PR #10384, opened 2017-11-06
Turn off sort and top after stand-alone limit in compiled runtime.

Lojjs:
Executed, but with the wrong result before. 
Same behaviour in 3.3 confirmed, should be forward merged all the way to 3.4

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/resources/blacklists/cost-compiled.txt
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/resources/blacklists/cost-compiled.txt
@@ -157,6 +157,9 @@ Accessing a map with null should return null
 //SkipLimitAcceptance.feature
 Combining LIMIT and aggregation
 Using a optional match after aggregation and before an aggregation
+Limit before sort should work except in compiled
+Limit before top should work except in compiled
+Limit before distinct should work except in compiled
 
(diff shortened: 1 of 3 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/resources/blacklists/cost-compiled.txt
```diff
@@ -157,6 +157,9 @@ Accessing a map with null should return null
 //SkipLimitAcceptance.feature
 Combining LIMIT and aggregation
 Using a optional match after aggregation and before an aggregation
+Limit before sort should work except in compiled
+Limit before top should work except in compiled
+Limit before distinct should work except in compiled
```
henriknyman: I think these are a bit weird names for acceptance tests. It should not be a requirement that a certain implementation should not work.

Lojjs: You've got a point there... Will be boarding my plane soon but will fix the names when I arrive home.

Lojjs: Totally missed your comment btw but Andres just pinged me about it

PR #10389, opened 2017-11-07
Asynchronously terminated transactions now immediately release their locks, so others can make progress

chrisvest:
Previously, asynchronously terminated transactions would only release their locks after checking the termination flag,
realising that the transaction had been asynchronously terminated, and then closing their transaction.
This process required that the terminated threads would cooperate and clean up the resources they are holding.
One such resource, a particularly precious one, is the locks acquired by the given transaction.
If a transaction never went into the kernel, it would never discover that it had been terminated, and thus could hold
on to their locks practically indefinitely, in turn blocking other transactions stuck on those locks from making
process.

The reason we did it this way, is because transactions might need to also grab locks during commit, and we cannot allow
asynchronous termination of transactions that have already started their commit process. Transactions that have started
committing will presumably relinquish their resource pretty soon anyway.

So to do this differently, to solve both the problem of letting killed transactions relinquish their locks sooner, and
to avoid killing transactions during commit, a new lock client state called "PREPARE" is introduced.
The lock client is moved to the PREPARE state when the transaction enters the PREPARE phase. Once the lock client is
in PREPARE, it can no longer be moved to the STOPPED state via asynchronous termination.
In the PREPARE state, the lock client will still allow new locks to be taken and released, and the client will move to
the STOPPED state when it is closed.

This mechanism protects the lock client and the commit process, and it allows us to release all locks held by a lock
client, when we asynchronously terminate with the stop() method it before the prepare phase.
This means that transaction timeout is now also effective on transactions that are stuck waiting for network traffic,
or other resources external to the database itself, for instance.
And it means that terminating a transaction will now immediately allow any other transactions, that are otherwise stuck
waiting for locks held by the terminated transaction, to make progress.

In this process, a bug has also been fixed in the listQueries procedure, where the lock count would come from whichever
lock client was used to *plan* the query, and not the lock client associated with the running transaction. The reason
this bug happened, is that the `StackingQueryRegistrationOperations` eagerly grabbed the lock client from the current
statement locks, when creating a lambda for getting the current lock count. The fix for that is to make the lambda
capture the `statement` reference, instead of the reference coming out of the `statement.locks()` call.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StackingQueryRegistrationOperations.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StackingQueryRegistrationOperations.java
@@ -73,7 +73,7 @@ public class StackingQueryRegistrationOperations implements QueryRegistrationOpe
         String threadName = thread.getName();
         ExecutingQuery executingQuery =
                 new ExecutingQuery( queryId, clientConnection, statement.username(), queryText, queryParameters,
-                        statement.getTransaction().getMetaData(), statement.locks()::activeLockCount,
+                        statement.getTransaction().getMetaData(), () -> statement.locks().activeLockCount(),
                         statement.getPageCursorTracer(),
                         threadId, threadName, clock, cpuClock, heapAllocation );
         registerExecutingQuery( statement, executingQuery );
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionTimeoutMonitorIT.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionTimeoutMonitorIT.java
@@ -123,8 +123,10 @@ public class KernelTransactionTimeoutMonitorIT
         }
         while ( !proceed );
 
-        Thread.sleep( 150 ); // locker should be stopped by now
-        assertFalse( lockerDone.get() ); // but still blocked on the latch
+        Thread.sleep( 150 ); // locker transaction is definitely past its allocated time now
+        // make sure it's terminated; we call this explicitly so we don't have to wait for the scheduler to run
+        database.resolveDependency( KernelTransactionTimeoutMonitor.class ).run();
+        assertFalse( lockerDone.get() ); // but the thread should still be blocked on the latch
         // Yet we should be able to proceed and grab the locks they once held
         try ( Transaction tx = database.beginTx() )
         {
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionTimeoutMonitorIT.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionTimeoutMonitorIT.java
@@ -27,8 +27,11 @@ import org.junit.rules.ExpectedException;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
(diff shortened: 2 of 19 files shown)


enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
```diff
@@ -700,14 +714,19 @@ public void stop()
...
+
+    private void releaseAllLocks()
+    {
         if ( hasLocks )
         {
             releaseAllClientLocks();
             clearWaitList();
             hasLocks = false;
         }
-        clientPool.release( this );
```
MishaDemianenko, line 710: not sure that removal of this line was planned

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionTimeoutMonitorIT.java
```diff
@@ -80,6 +92,52 @@ public void terminateExpiredTransaction() throws Exception
...
+        } );
+
+        boolean proceed;
+        do
+        {
+            proceed = nodeLockAcquired.get();
+        }
+        while ( !proceed );
+
+        Thread.sleep( 150 ); // locker should be stopped by now
```
MishaDemianenko: that's not really reliable timeout for build agents

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StackingQueryRegistrationOperations.java
```diff
@@ -73,7 +73,7 @@ public ExecutingQuery startQueryExecution(
         String threadName = thread.getName();
         ExecutingQuery executingQuery =
                 new ExecutingQuery( queryId, clientConnection, statement.username(), queryText, queryParameters,
-                        statement.getTransaction().getMetaData(), statement.locks()::activeLockCount,
+                        statement.getTransaction().getMetaData(), () -> statement.locks().activeLockCount(),
```
MishaDemianenko, line 76: i think we have more cases where this will not gonna problems with the current statement and ExecutingQuery object, let's talk about them f2f tomorrow

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionTimeoutMonitorIT.java
MishaDemianenko: we need to make it more cloud agents reliable

MishaDemianenko (commented):
Looks good in general, have a couple of minor comments. Will not mark as approved for now to not forget to talk about it.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/enterprise/lock/forseti/ForsetiClient.java
chrisvest: Right, that should have been added back into `close()`. Good catch.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StackingQueryRegistrationOperations.java
chrisvest: I search around a bit and couldn't find any other cases.

MishaDemianenko (approved):

PR #10396, opened 2017-11-08
Prevent ExplicitIndexProxy from closing statment multiple times

burqen:
When using getSingle in ExplicitIndexWrapHits statement would be
closed twice because of Iterators.singleOrNull closing provided
iterator if it is a resource, which ExplicitIndexWrapHits is.

We now keep track of close state to prevent it being closed
multiple times and thereby closing statement for others that
might be using it.


MishaDemianenko (changes requested):
Current changeset looks good, but in addition to what was already done i think we need to adapt AbstractIndexHits::getSingle to not close resource iterator twice

burqen:
@MishaDemianenko Thank you! Comment addressed.

MishaDemianenko (approved):

PR #10400, opened 2017-11-08
Add bolt dependency to enterprise, add neo4j-cypher-runtime-util test dependency

MishaDemianenko:


enterprise/neo4j-enterprise/pom.xml
```diff
@@ -90,6 +89,11 @@
       <artifactId>neo4j-security-enterprise</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.neo4j</groupId>
+      <artifactId>neo4j-bolt</artifactId>
+      <version>${project.version}</version>
```
klaren, line 95: There is another neo4j-bolt dependency defined below with scope test that can be removed now since scope compile also implies test.

jjaderberg:
@MishaDemianenko Are you sure this is for enterprise only? I had the same failure for community.

MishaDemianenko:
@jjaderberg sure community also has that problem will fix it there as well, thx!

MishaDemianenko:
@klaren updated

klaren (approved):
Looks good!

PR #10401, opened 2017-11-08
Double time for ExplicitIndexTest, cleanup deprecated things

MishaDemianenko:


burqen:
@MishaDemianenko From your change set I don't see how the test failure in `TestWhiteListJavaWrapper` can be related. But this is the only time this test has failed and it also fails locally. I don't see this test failure on `3.4 HEAD`, but I see it after rebasing your branch on `3.4 HEAD`. So my conclusion is that there is some more work for you to do here ;)

MishaDemianenko:
@burqen yeah, his past is hard and future is unclear, let me see whats wrong with it! thx )

MishaDemianenko:
@burqen here it is

community/server/src/main/java/org/neo4j/server/scripting/UserScriptClassWhiteList.java
```diff
@@ -64,6 +64,8 @@ private UserScriptClassWhiteList()
         safe.add( Node.class.getName() );
         safe.add( Relationship.class.getName() );
         safe.add( RelationshipType.class.getName() );
+        // can't list antonymous inner classes, using instance to get name of its class
+        safe.add( RelationshipType.withName( "classSource" ).getClass().getName() );
```
burqen: "antonymous"?

burqen (approved):
Not sure I understand the comment. Do with it as you please.

community/server/src/main/java/org/neo4j/server/scripting/UserScriptClassWhiteList.java
MishaDemianenko: hehe, for a moment he was special )

PR #10402, opened 2017-11-08
Experimental Cypher support for scaling queries to multiple threads

systay:


enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/VarLengthExpandSlottedPipe.scala
```diff
@@ -25,16 +25,17 @@ import org.neo4j.cypher.internal.util.v3_4.InternalException
...
+import org.neo4j.cypher.internal.v3_4.logical.plans.LogicalPlanId
 import org.neo4j.graphdb.Relationship
 import org.neo4j.kernel.impl.api.RelationshipVisitor
 import org.neo4j.kernel.impl.api.store.RelationshipIterator
 import org.neo4j.kernel.impl.util.ValueUtils
 
 import scala.collection.mutable
 
-case class VarLengthExpandSlottedPipe(source: Pipe,
+case class        VarLengthExpandSlottedPipe(source: Pipe,
```
pontusmelke: Weird white space

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/EnterpriseCompatibilityFactory.scala
```diff
@@ -44,8 +47,15 @@ class EnterpriseCompatibilityFactory(inner: CompatibilityFactory, graph: GraphDa
       case (CypherPlanner.rule, _) => inner.create(spec, config)
 
       case _ =>
+        val settings = graph.getDependencyResolver.resolveDependency(classOf[Config])
+        val morselSize = settings.get( GraphDatabaseSettings.cypher_morsel_size)
+        val workers = settings.get( GraphDatabaseSettings.cypher_worker_count)
+        val dispatcher =
+          if (workers == 1) new SingleThreadedExecutor(morselSize)
+          else if (workers == 0) new ForkJoinPoolExecutor(morselSize, Runtime.getRuntime.availableProcessors() * 2)
```
jakewins: Would we be able to have cypher get an executor from https://github.com/neo4j/neo4j/blob/3.4/community/common/src/main/java/org/neo4j/scheduler/JobScheduler.java#L73 ? 

I know it's just moving the same behavior elsewhere, but it seems useful to try to keep thread creation in a single place to ease future expansion of control of jobs.

Note that cypher can still get it's own pool just like this one, it just would get it from the central JobScheduler, see `BoltKernelExtension` for how bolt does this (eg. it depends on getting an Executor or whatever, and just uses the JobScheduler to instantiate one)

henriknyman: Why * 2? Runtime.getRuntime.availableProcessors() should include hyper threads so doubling the number of workers will just give us more context switching?

systay: You are correct. Don't know what a good value is. Our very small benchmarks do not show a performance drop when we have too many workers, but that does not mean we should. We should probably change this to what you say, but I would love to investigate and have numbers to show us.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/predicates/ComparablePredicate.scala
```diff
@@ -33,15 +33,18 @@ abstract sealed class ComparablePredicate(val left: Expression, val right: Expre
...
 
-    if (l == Values.NO_VALUE || r == Values.NO_VALUE) None
+    val res = if (l == Values.NO_VALUE || r == Values.NO_VALUE) None
     else (l, r) match {
       case (d: FloatingPointValue, _) if d.doubleValue().isNaN => None
       case (_, d: FloatingPointValue) if d.doubleValue().isNaN => None
-      case (n1: NumberValue, n2: NumberValue) => Some(compare(AnyValues.COMPARATOR.compare(n1, n2)))
+      case (n1: NumberValue, n2: NumberValue) =>
+        val i = AnyValues.COMPARATOR.compare(n1, n2)
+        Some(compare(i))
```
fickludd: debugging-remainder?

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/EnterpriseCompatibilityFactory.scala
systay: @jakewins Our JobScheduler does not support bounded threadpools, from what I can see. I think it's important to be able to cap the number of worker threads we use for this. Do you still think we should use `Neo4jJobScheduler` for this?

enterprise/cypher/morsel-runtime/pom.xml
```diff
@@ -0,0 +1,253 @@
...
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.neo4j</groupId>
+      <artifactId>neo4j-cypher-slotted-runtime</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.neo4j</groupId>
+      <artifactId>neo4j-codegen</artifactId>
```
fickludd: codegen? warum?

enterprise/cypher/morsel-runtime/pom.xml
```diff
@@ -0,0 +1,253 @@
...
+    <!-- neo4j-cypher re-include excluded (with correct version) -->
+
+    <dependency>
+      <groupId>org.parboiled</groupId>
+      <artifactId>parboiled-scala_2.11</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>net.sf.opencsv</groupId>
+      <artifactId>opencsv</artifactId>
+    </dependency>
```
fickludd: Are there deps really relevant for this runtime?

enterprise/cypher/morsel-runtime/pom.xml
```diff
@@ -0,0 +1,253 @@
...
+      <artifactId>asm-util</artifactId>
+      <version>${asm.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.ow2.asm</groupId>
+      <artifactId>asm-analysis</artifactId>
+      <version>${asm.version}</version>
+      <scope>test</scope>
+    </dependency>
```
fickludd: Same with these?

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -280,15 +280,17 @@ object SlotAllocation {
            _: ConditionalApply =>
         rhsPipeline
 
-      case _: CartesianProduct =>
-        val newPipeline = lhsPipeline.breakPipelineAndClone()
+      case _: CartesianProduct |
+           _: NodeHashJoin =>
```
fickludd: Is this sneaky addition of `NodeHashJoin` intentional? Wondering about tests etc?

fickludd (commented):
Found some misc things while browsing through here. In general it is awesome you made all these foundations for parallelization and vectorization. 

One thing I'm a bit unhappy about, is that by adding the `Morsel` runtime to the regular testing frameworks, it is suddenly everyone's problem to maintain it. Curious to see if that becomes a problem, I guess we'll find out.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
systay: Oops. Not by design!

enterprise/cypher/morsel-runtime/pom.xml
systay: That's what you get for copy-pasta poms...

systay:
@fickludd re: https://github.com/neo4j/neo4j/pull/10402#pullrequestreview-75423074

You are absolutely correct. The same can be said by just adding the code to the repo - if we now change one of many traits that are used by the morsel runtime, we need to update the prototype code as well as the rest of the code.

I think this is the right way of handling it for now, but I'm not at all sure. I think we should keep an eye on this cost, and cut it out if it becomes a burden.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/EnterpriseCompatibilityFactory.scala
jakewins: @systay There is `threadFactory(..)`, which lets you be in charge of creating threads; however, since you're using ForkJoinPool, it looks like you can't create a ForkJoinPool from a raw ThreadFactory; you'd need to amend the `JobScheduler` with a `forkJoinThreadFactory()` or `forkJoinThreadPool(int limit)` or some such.

So: Yes, you're right :( But! If you refactor `JobScheduler` to support bounded ForkJoinPools, then others that want to use ForkJoin things can benefit :) You can do that either by replicating the pattern used by `threadFactory()`, by `executor()`, or by amending the `Group` definition to build in rules - eg. "group `CypherWorkers` can only have N workers". The latter would be closer in line with long-term plans; eg. the `JobScheduler` decides how to divide up resources:

    public static Group cypherWorkers = new Group( "CypherWorker", maxThreads(..) );

By using the JobScheduler for this, we retain a single view of thread creation in the DB - and all these Cypher threads end up following the same naming scheme in thread dumps, which feels good to my OCD.

pontusmelke:
@fickludd re:re: #10402 (review)
Morsel is not part of `Configs.All` which means that it should only be a burden if you add something to the prototype - in that case you will manually have to add `+ Configs.Morsel` for any queries that are now supported. In this way I hope the burden will mostly be on the people working on the prototype rather than on everyone.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/Neo4jJobScheduler.java
```diff
@@ -67,11 +76,29 @@ public Executor executor( final Group group )
...
+    private ExecutorService createNewWorkStealingExecutor( Group group, int parallelism )
+    {
+        final ForkJoinPool.ForkJoinWorkerThreadFactory factory = pool ->
+        {
+            final ForkJoinWorkerThread worker = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread( pool );
+            worker.setName( group.threadName( new HashMap<>() ) );
+            return worker;
+        };
+
+        return new ForkJoinPool( parallelism, factory, null, false );
```
chrisvest, line 99: There's a ForkJoinPool constructor that takes a worker name prefix. I think that'll be simpler than building this custom thread factory.

chrisvest: Oh wait, that constructor is private.

PR #10403, opened 2017-11-08
Store format update with point storage capabilities

sherfert:


craigtaverner:
This is not the right way to do it.

PR #10406, opened 2017-11-08
Abandoned and timed out transactions are not released by bolt server

ali-ince:
When client transactions are not closed appropriately, bolt server continues to hold related transactions even if the server is configured to time out transactions. This PR addresses this problem and validates the transaction (if any) and releases it in case it is timed out. With releasing, the transaction state machine is reset - closing any open statement handles and transactions.

In case the client tries to interact with the server, the recorded termination status message is returned to the client as a failure message.

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
@@ -137,13 +137,13 @@ public class TransactionStateMachine implements StatementProcessor
 
         if ( tx != null )
         {
-            Optional<Status> status = tx.getReasonIfTerminated();
+            Optional<Status> statusOpt = tx.getReasonIfTerminated();
 
-            if ( status.isPresent() )
+            if ( statusOpt.isPresent() )
             {
-                if ( status.get().code().classification().rollbackTransaction() )
+                if ( statusOpt.get().code().classification().rollbackTransaction() )
                 {
-                    ctx.pendingTerminationNotice = status;
+                    ctx.pendingTerminationNotice = statusOpt.get();
 
                     reset();
                 }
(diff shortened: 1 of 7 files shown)


community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
```diff
@@ -112,10 +118,7 @@ public void reset() throws TransactionFailureException
 
     private void after()
     {
-        if ( ctx.currentTransaction != null )
```
digitalstain, line 115: It is safe to remove the null check because it wasn't needed to begin with - is that correct?

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
```diff
@@ -386,6 +422,8 @@ private static BoltResultHandle executeQuery( MutableTransactionState ctx, SPI s
         /** The current transaction, if present */
         KernelTransaction currentTransaction;
 
+        Optional<Status> pendingTerminationNotice = Optional.empty();
```
lutovich: Maybe it's better to have this field without `Optional`? It is only used using `#isPresent()` + `#get()` which is essentially a null-check.

lutovich (approved):
LGTM, one small comment

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
ali-ince: Besides its redundancy, when an error occurs in `streamResult` call - the call may begin with a valid `ctx.currentTransaction` but end with a `null` one which will leave that specific transaction left as referenced by ThreadLocal variables.

I removed the null check as a `remove the ThreadLocal reference at all times` kind of behavior.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
ali-ince: You're right. Changing :)

pratheepkn:
In our 3.2.5 seeing large number of established connection threads.. Is it related to this?
neo4j>  call dbms.listQueries() yield queryId,username,query;
+--------------------------------------------------------------------------------------+
| queryId       | username | query                                                     |
+--------------------------------------------------------------------------------------+
| "query-36322" | "neo4j"  | "  call dbms.listQueries() yield queryId,username,query;" |
+--------------------------------------------------------------------------------------+

But in OS level..
[neo4j@xxx neo4j]$ netstat -na | grep :7687 | grep ESTABLISHED | wc -l
7140
should go for latest?

ali-ince:
Hi @pratheepkn,

This PR only includes resource usage fixes when transactions are timed out / killed.

Apart from this, it may be more related to your client code - what driver/version are you using?

Thanks.

PR #10407, opened 2017-11-08
Windows service does not start when in a directory with spaces in name

ali-ince:
If neo4j is installed in a directory with space(s) in name (like C:\Program Files\neo4j), although service is reported to be installed successfully it does not start. With this fix, scripts are updated to quote all arguments passed to prunsrv.

changelog: Fixed broken service installation for paths with spaces on Windows


spacecowboy:
@ali-ince thx for the pr. I'll review and test it on Monday

spacecowboy:
@ali-ince I tested it and verified that it indeed fixes the problem. I also added a changelog entry to the description so that it is included in the changelogs for the next round of releases. Thanks for your help here.

In addtion I also backported the fix to 3.1

PR #10410, opened 2017-11-09
Reducing Cypher queries

sherfert:
This PR contains my DWYW project to reduce Cypher queries.

The method CypherReductionSupport.reduceQuery is the central piece of that work.


OliviaYtterbrink (approved):
I really like this, LGTM

PR #10411, opened 2017-11-09
Fixes equality for lists in accordance with OpenCypher.

OliviaYtterbrink:
changelog: Fixes equality for lists in accordance with OpenCypher.

Before this change, `[1,2,3] = [1,2,null]` returned `false`. Now, this comparison will return `null` which makes equality more consistent in it's behaviour.

### DIFF
--- a/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
+++ b/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
@@ -191,12 +191,52 @@ public abstract class CompiledConversionUtils
             return null;
         }
 
+        boolean lhsIsSeq = lhs instanceof  List<?>;
+        boolean rhsIsSeq = rhs instanceof List<?>;
+
+        // Can't compare a literal to a list
+        if ( (lhsIsSeq && !rhsIsSeq) || (!lhsIsSeq && rhsIsSeq) )
+        {
+            return null;
+        }
+
+        if ( lhsIsSeq && rhsIsSeq )
+        {
+            List<?> lhsList = (List<?>) lhs;
+            List<?> rhsList = (List<?>) rhs;
+            boolean foundNull = false;
+
+            // Different length lists can't be equal
+            if (lhsList.size() != rhsList.size())
+            {
+                return false;
+            }
+
+            for ( int i = 0; i < lhsList.size(); i++ )
+            {
+                Object obj1 = lhsList.get( i );
+                Object obj2 = rhsList.get( i );
+
+                Boolean objEquality = equals( obj1, obj2 );
+                if( objEquality == null )
+                {
+                    foundNull = true;
+                } else if ( !objEquality ) {
+                    return false;
+
+                }
+            }
+            if (foundNull) {
+                return null;
+            }
+            return true;
+        }
+
         if ( (lhs instanceof NodeIdWrapper && !(rhs instanceof NodeIdWrapper)) ||
              (rhs instanceof NodeIdWrapper && !(lhs instanceof NodeIdWrapper)) ||
              (lhs instanceof RelationshipIdWrapper && !(rhs instanceof RelationshipIdWrapper)) ||
              (rhs instanceof RelationshipIdWrapper && !(lhs instanceof RelationshipIdWrapper)) )
         {
-
             throw new IncomparableValuesException( lhs.getClass().getSimpleName(), rhs.getClass().getSimpleName() );
         }
 
--- a/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
+++ b/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
@@ -35,6 +35,7 @@ import java.util.stream.DoubleStream;
 import java.util.stream.IntStream;
 import java.util.stream.LongStream;
 
+import org.neo4j.cypher.internal.compatibility.v3_3.runtime.helpers.ValueConversion;
 import org.neo4j.cypher.internal.compiler.v3_3.spi.NodeIdWrapper;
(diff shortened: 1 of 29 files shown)


community/values/src/main/java/org/neo4j/values/AnyValue.java
```diff
@@ -52,4 +52,6 @@ public boolean isSequenceValue()
     {
         return false; // per default Values are no SequenceValues
     }
+
+    public abstract Boolean ternaryEquals( Object other );
```
fickludd: Is it possibly so that this could take a `Value` instead of an `Object`?

community/values/src/test/java/org/neo4j/values/utils/AnyValueTestUtil.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+package org.neo4j.values.utils;
+
+import org.neo4j.values.AnyValue;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class AnyValueTestUtil
```
fickludd, line 28: This class makes a lot of sense, like it!

fickludd (commented):
Got interested and stumbled on a down-typing suggestion. Looks good.

systay (approved):
Looks really good. Let's merge when green

community/values/src/main/java/org/neo4j/values/storable/FloatingPointValue.java
```diff
@@ -69,6 +69,12 @@ public int compareTo( FloatingPointValue other )
     }
 
     @Override
+    public boolean isNaN()
+    {
+        return this.doubleValue() != this.doubleValue();
```
sherfert: I think `Double.isNaN` would be clearer. But not strictly necessary, this can be understood too.

community/values/src/main/java/org/neo4j/values/storable/Value.java
```diff
@@ -50,6 +59,30 @@
...
     public abstract boolean equals( String[] x );
 
+    @Override
+    public Boolean ternaryEquals( AnyValue other )
+    {
+        if ( other == null || other == NO_VALUE )
+        {
+            return null;
+        }
+        if ( other instanceof SequenceValue && this.isSequenceValue() )
```
sherfert: Inconsistent check for `this` and `other`

community/values/src/test/java/org/neo4j/values/utils/AnyValueTestUtil.java
```diff
@@ -0,0 +1,71 @@
...
+                String.format( "%s should be equal %s", a.getClass().getSimpleName(), b.getClass().getSimpleName() ),
+                a.ternaryEquals( b ) );
+        assertTrue(
+                String.format( "%s should be equal %s", a.getClass().getSimpleName(), b.getClass().getSimpleName() ),
+                b.ternaryEquals( a ) );
+        assertTrue( String.format( "%s should have same hashcode as %s", a.getClass().getSimpleName(),
+                b.getClass().getSimpleName() ), a.hashCode() == b.hashCode() );
+    }
+
+    public static void assertEqualValues( AnyValue a, AnyValue b )
```
sherfert, line 48: What is the difference between `assertEqual` and `assertEqualValues` ?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NullListAcceptanceTest.scala
```diff
@@ -0,0 +1,159 @@
...
+
+    val result = executeWith(Configs.All, query, expectedDifferentResults = nullInListConfigOld)
+
+    result.toList should equal(List(Map("res" -> null)))
+  }
+
+  // IN with null
+
+  test("IN with different length lists should return false despite nulls") {
+    val query = "WITH [1] AS l1, [1, null] AS l2 RETURN l1 IN [l2] AS res"
```
sherfert, line 95: Can we change that to `WITH [1] AS l1, [[1, null]] AS l2 RETURN l1 IN l2 AS res` to make it consistent with the other tests?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NullListAcceptanceTest.scala
```diff
@@ -0,0 +1,159 @@
...
+
+    val result = executeWith(Configs.All, query)
+
+    result.toList should equal(List(Map("res" -> false)))
+  }
+
+  test("IN should return true if match despite nulls") {
+    val query = "WITH 3 AS l1, [1, null, 3] AS l2 RETURN l1 IN l2 AS res"
+
+    val result = executeWith(Configs.Interpreted, query)
```
sherfert, line 105: What is the difference in the queries that suddenly Compiled does not support them?

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
```diff
@@ -184,23 +185,38 @@ public int hashCode()
...
+                throw new IncomparableValuesException( lhs.getClass().getSimpleName(), rhs.getClass().getSimpleName() );
+            }
+            return lhs.equals( rhs );
         }
 
-        return CompiledEquivalenceUtils.equals( lhs, rhs );
+        AnyValue lhsValue = lhs instanceof AnyValue ? (AnyValue) lhs : ValueConversion.asValue( lhs );
+        AnyValue rhsValue = rhs instanceof AnyValue ? (AnyValue) rhs : ValueConversion.asValue( rhs );
+
+        return lhsValue.ternaryEquals( rhsValue );
```
sherfert, line 219: It seems the whole `CompiledEquivalenceUtils` is dead code now?

sherfert (commented):
Nice work! I have some minor comments and understanding questions.

PR #10412, opened 2017-11-10
Split should return list not array

pontusmelke:
Accidentally made `split` return an array instead of a list in the embedded
API.


fickludd (approved):
Looks good!

PR #10413, opened 2017-11-10
Value and order capabilities for schema indexes

burqen:
Builds on top of https://github.com/neo4j/neo4j/pull/10279

Introduce infrastructure to be able to get exact values and defined result order from index queries.

Implementation of capabilities is limited to single property number indexes and only ascending order.


community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/UnionIndexCapability.java
```diff
@@ -0,0 +1,69 @@
...
+        this.capabilities = capabilities;
+    }
+
+    @Override
+    public IndexOrder[] order( ValueGroup... valueGroups )
+    {
+        Set<IndexOrder> orderCapability = new HashSet<>();
+        for ( IndexCapability capability : capabilities )
+        {
+            Arrays.stream( capability.order( valueGroups ) ).forEach( orderCapability::add );
```
klaren:
```
orderCapability.addAll( Arrays.asList( capability.order( valueGroups ) ) );
```
would be easier to follow.

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/MockStore.java
```diff
@@ -25,9 +25,9 @@
 
 import org.neo4j.collection.primitive.Primitive;
 import org.neo4j.collection.primitive.PrimitiveLongObjectMap;
+import org.neo4j.internal.kernel.api.*;
```
klaren: Star imports not allowed, I thought that checkstyle caught those.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -85,51 +96,73 @@ public void shouldPerformExactLookup() throws Exception
...
+            assertFoundNodesAndValue( node, uniqueIds, expectValue, strThree1, strThree2, strThree3 );
 
             // when
-            read.nodeIndexSeek( index, node, IndexQuery.exact( prop, 6 ) );
+            expectValue = index.value( ValueGroup.NUMBER );
+            read.nodeIndexSeek( index, node, null, IndexQuery.exact( prop, 1 ) );
 
             // then
-            assertNodes( node, uniqueIds, num6 );
+            // todo continue here
```
klaren: todo

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -312,3 +485,4 @@ private long nodeWithProp( GraphDatabaseService graphDb, Object value )
         return node.getId();
     }
 }
+
```
klaren: 2 new line at end of file

klaren (dismissed):
Looks really good! I just have a few small comments.

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/MockStore.java
burqen: It does apparently https://build.neohq.net/viewLog.html?buildId=2383763&buildTypeId=PullRequests_BranchPackaging&tab=buildLog&_focus=4287

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/NodeValueIndexCursor.java
```diff
@@ -23,6 +23,37 @@
...
  * Cursor for scanning the property values of nodes in a schema index.
+ * <p>
+ * Usage pattern:
+ * <pre><code>
+ *     int nbrOfProps = cursor.numberOfProperties();
+ *     int[] propertyKeys = new int[nbrOfProps];
+ *     for ( int i = 0; i < nbrOfProps; i++ )
+ *     {
+ *         propertyKeys[i] = cursor.propertyKey( i );
+ *     }
```
fickludd: Dunno if we actually need the `propertyKeys` array here?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexValueCapability.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.internal.kernel.api;
+
+public enum IndexValueCapability
+{
+    YES( 3 ),
+    MAYBE( 2 ),
```
fickludd: Would love to know that `MAYBE` means? For some values? Could be `PARTIAL` instead?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -64,6 +69,12 @@ void createTestGraph( GraphDatabaseService graphDb )
             strThree3 = nodeWithProp( graphDb, "three" );
             nodeWithProp( graphDb, false );
             boolTrue = nodeWithProp( graphDb, true );
+            nodeWithProp( graphDb, 3 ); // Purposely mix ordering
```
fickludd, line 72: nice!

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -85,51 +96,73 @@ public void shouldPerformExactLookup() throws Exception
...
         int label = token.nodeLabel( "Node" );
         int prop = token.propertyKey( "prop" );
-        IndexReference index = schemaRead.index( label, prop );
+        CapableIndexReference index = schemaRead.index( label, prop );
         try ( NodeValueIndexCursor node = cursors.allocateNodeValueIndexCursor();
               PrimitiveLongSet uniqueIds = Primitive.longSet() )
         {
             // when
-            read.nodeIndexSeek( index, node, IndexQuery.exact( prop, "zero" ) );
+            IndexValueCapability expectValue = index.value( ValueGroup.TEXT );
```
fickludd: This line i confusing to me. `index.value( ... )` reads like a query or value-access, not a capability-listing. And is this really an `expectValue`?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -157,15 +191,16 @@ public void shouldPerformStringSuffixSearch() throws Exception
         // given
         int label = token.nodeLabel( "Node" );
         int prop = token.propertyKey( "prop" );
-        IndexReference index = schemaRead.index( label, prop );
+        CapableIndexReference index = schemaRead.index( label, prop );
+        IndexValueCapability stringCapability = index.value( ValueGroup.TEXT );
```
fickludd: I like `stringCapability` much better!

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -280,15 +311,157 @@ public void shouldPerformIndexScan() throws Exception
         // given
         int label = token.nodeLabel( "Node" );
         int prop = token.propertyKey( "prop" );
-        IndexReference index = schemaRead.index( label, prop );
+        CapableIndexReference index = schemaRead.index( label, prop );
+        IndexValueCapability wildcardCapability = index.value( new ValueGroup[]{null} );
```
fickludd: WDYT about passing an empty var-args here instead of `new ValueGroup[]{null}`?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -280,15 +311,157 @@ public void shouldPerformIndexScan() throws Exception
...
         try ( NodeValueIndexCursor node = cursors.allocateNodeValueIndexCursor();
               PrimitiveLongSet uniqueIds = Primitive.longSet() )
         {
             // when
-            read.nodeIndexScan( index, node );
+            read.nodeIndexScan( index, node, null );
```
fickludd: Also we could add a `IndexOrder.ARBITRARY` or `ANY_ORDER` or something instead of passing `null`.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberHitIndexCursorProgressor.java
```diff
@@ -23,61 +23,59 @@
...
- * @param <KEY> type of {@link SchemaNumberKey}.
- * @param <VALUE> type of {@link SchemaNumberValue}.
- */
-public class NumberHitIterator<KEY extends SchemaNumberKey, VALUE extends SchemaNumberValue>
-        extends PrimitiveLongCollections.PrimitiveLongBaseIterator
+public class NumberHitIndexCursorProgressor<KEY extends SchemaNumberKey, VALUE extends SchemaNumberValue> implements IndexProgressor
 {
     private final RawCursor<Hit<KEY,VALUE>,IOException> seeker;
-    private final Collection<RawCursor<Hit<KEY,VALUE>,IOException>> toRemoveFromWhenExhausted;
+    private final NodeValueClient cursor;
```
fickludd: `client` now, post-rebase =)

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexReader.java
```diff
@@ -100,6 +107,42 @@ public PrimitiveLongIterator query( IndexQuery... predicates ) throws IndexNotAp
...
+            return;
+        }
+
+        if ( predicates[0] instanceof NumberRangePredicate )
+        {
+            nativeReader.query( cursor, indexOrder, predicates[0] );
+            return;
+        }
+
+        // todo: There will be no ordering of the node ids here. Is this a problem?
```
fickludd, line 135: If we provide an `ExistsPredicate` with a specified `IndexOrder` we need to fail if we cannot serve it?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexReference.java
```diff
@@ -0,0 +1,71 @@
...
+import org.neo4j.internal.kernel.api.IndexCapability;
+import org.neo4j.internal.kernel.api.IndexOrder;
+import org.neo4j.internal.kernel.api.IndexValueCapability;
+import org.neo4j.values.storable.ValueGroup;
+
+class IndexReference implements org.neo4j.internal.kernel.api.CapableIndexReference
+{
+    private final int label;
+    private final int[] properties;
+    private final boolean unique;
```
fickludd: It somehow feels wrong to create a new index meta class that is a `IndexDescriptor` + capabilities. Could we do composition instead? We don't have to fix it now, just venting opinion.

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/ReadTestSupport.java
```diff
@@ -21,22 +21,34 @@
...
 
 class ReadTestSupport implements KernelAPIReadTestSupport
 {
+    private final Map<Setting,String> settings = new HashMap<>();
     private GraphDatabaseService db;
 
+    void addSetting( Setting setting, String value )
+    {
+        settings.put( setting, value );
+    }
```
fickludd, line 44: This is a nice addition!

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/SimpleNodeValueClient.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.newapi;
+
+import org.neo4j.storageengine.api.schema.IndexProgressor;
+import org.neo4j.values.storable.Value;
+
+public class SimpleNodeValueClient implements IndexProgressor.NodeValueClient
```
fickludd: As an alternative, this could live together with `IndexProgressor` in `org.neo4j.storageengine.api.schema`

fickludd (changes requested):
This is great. Got some opportunities where readability could improve, but only nitpicks.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexValueCapability.java
burqen: Exactly. I like `PARTIAL` 👍

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/NodeValueIndexCursor.java
burqen: Tried to illustrate that you can get targeted property ids from cursor and if we are missing `ValueCapability` then we can do 
```
 *             for ( int i = 0; i < nbrOfProps; i++ )
 *             {
 *                 values[i] = getPropertyValueFromStore( cursor.nodeReference(), propertyKeys[i] );
 *             }
```
But that was not very clear apparently 👍 
Thinking about it now it seems to just obscure the pattern that I'm actually trying to communicate.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
burqen: Change this to `IndexValueCapability valueCapability = index.valueCapability( ValueGroup.TEXT );`

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
burqen: The reason we want to use an explicit null value in array is because of composite indexes where you can have something like `new ValueGroup[]{null,ValueGroup.TEXT}`, indicating that first part of key could be of any value group and the second part is known to be a string.

Do you still think it is reasonable to use empty array for single property wildcard? I can see the use for that syntactic sugar but at the same time I don't like multiple ways of doing the same thing.

I guess another alternative would be to add another `ValueGroup.WILDCARD` or use `ValueGroup.NO_VALUE` for that purpose. But doesn't seem right to pollute `Values` module like that.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
burqen: We want `IndexOrder` to be used both as the answer to "what different ordering are you capable of?" and as the means of requesting order, "I want order X for my result." That makes interpretation of `ARBITRARY` and `ANY_ORDER` a bit fuzzy.

May `UNORDERED` could be a good fit.

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/SimpleNodeValueClient.java
burqen: I don't see any use for this class outside of tests, so let's keep it out of production code.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexReference.java
burqen: Yes, I agree it's a bit awkward. From what we discussed yesterday we might want to split up the code path in two.
```
public interface SchemaRead
{
    IndexDescription index( int label, int... properties );

    IndexCapability indexCapability( IndexDescription indexDescription );
}
```
Let's address this issue at that point.

burqen:
@fickludd @klaren Really sweet feedback! :D I think I've addressed all of your comments now.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
fickludd: We have a `ValueGroup.NO_VALUE` already. Maybe we should add `ValueGroup.ANY`?

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/SimpleNodeValueClient.java
fickludd: Yes... I'm suggesting to put it in `community/kernel/src/test/org/neo4j/storageengine/api/schema`?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/NodeValueIndexCursor.java
fickludd: `values[i] = getPropertyValueFromStore( cursor.nodeReference(), cursor.propertyKey( i ) )`?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
burqen: I'm gonna try that 🎉

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/SimpleNodeValueClient.java
burqen: Haha, alright. Sounds good!

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
fickludd: Hmm, `read.nodeIndexScan( index, node, UNORDERED )` reads like we explicitly want the results to be out of order. Maybe overload the method `read.nodeIndexScan( index, node )`? Or `read.nodeIndexScan( index, node, ANY_ORDER )` would read ok, but that crashes with the capability use. 

Meh. Maybe keep the null then :)

burqen:
@fickludd Another attempt 😄

community/kernel/src/main/java/org/neo4j/kernel/api/schema/index/IndexDescriptor.java
```diff
@@ -41,7 +40,7 @@
  * This will be renamed to IndexDescriptor, once the old org.neo4j.kernel.api.schema.IndexDescriptor is completely
```
MishaDemianenko: since you already here can you please clean up this comment

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionSchemaIndexProvider.java
```diff
@@ -126,6 +130,27 @@ public InternalIndexState getInitialState( long indexId, IndexDescriptor descrip
...
+    {
+        IndexCapability nativeCapability = nativeProvider.getCapability( indexDescriptor );
+        IndexCapability luceneCapability = luceneProvider.getCapability( indexDescriptor );
+        return new UnionIndexCapability( nativeCapability, luceneCapability )
+        {
+            @Override
+            public IndexOrder[] orderCapability( ValueGroup... valueGroups )
+            {
+                // No order capability when combining results from different indexes
+                if ( valueGroups.length == 1 && valueGroups[0] == ValueGroup.UNKNOWN )
```
MishaDemianenko, line 143: shouldn't we have the same check for value capabilities?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexProvider.java
```diff
@@ -193,4 +204,41 @@ public boolean compatibleWith( long layoutIdentifier, int majorVersion, int mino
...
     }
+
+    private static class NativeIndexCapability implements IndexCapability
+    {
+        @Override
+        public IndexOrder[] orderCapability( ValueGroup... valueGroups )
+        {
+            if ( support( valueGroups ) )
+            {
+                return new IndexOrder[]{IndexOrder.ASCENDING};
```
MishaDemianenko: we do not really need new arrays each time here right?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexProvider.java
```diff
@@ -193,4 +204,41 @@ public boolean compatibleWith( long layoutIdentifier, int majorVersion, int mino
...
+        }
+
+        private boolean singleWildcard( ValueGroup[] valueGroups )
+        {
+            return valueGroups.length == 1 && valueGroups[0] == ValueGroup.UNKNOWN;
+        }
+
+        private boolean support( ValueGroup[] valueGroups )
+        {
+            return valueGroups.length == 1 && ValueGroup.NUMBER.equals( valueGroups[0] );
```
MishaDemianenko: please use == for enum comparisons like its done 2 lines above

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexReader.java
```diff
@@ -106,37 +113,64 @@ public long countIndexedNodes( long nodeId, Value... propertyValues )
     @Override
     public PrimitiveLongIterator query( IndexQuery... predicates ) throws IndexNotApplicableKernelException
     {
+        NodeValueIterator nodeValueIterator = new NodeValueIterator();
+        query( nodeValueIterator, null, predicates );
```
MishaDemianenko: can we avoid passing null here? and introduce constant like IndexOrder.NONE for that instead?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NodeValueIterator.java
```diff
@@ -0,0 +1,71 @@
...
+import org.neo4j.storageengine.api.schema.IndexProgressor;
+import org.neo4j.values.storable.Value;
+
+/**
+ * A {@link IndexProgressor} + {@link IndexProgressor.NodeValueClient} combo presented as a {@link PrimitiveLongIterator}.
+ */
+public class NodeValueIterator extends PrimitiveLongCollections.PrimitiveLongBaseIterator
+        implements IndexProgressor.NodeValueClient, Resource
+{
+    private volatile boolean closed;
```
MishaDemianenko: why do we need this field to be volatile? do we expect multithreaded close calls?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NodeValueIterator.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public boolean acceptNode( long reference, Value... values )
+    {
+        return next( reference );
+    }
+
+    @Override
+    public void close()
```
MishaDemianenko, line 63: NodeValueIterator iterator used only in NativeSchemaNumberIndexReader as PrimitiveLongIterator which is not Resource how this can be reliably closed?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexReference.java
```diff
@@ -0,0 +1,71 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.newapi;
+
+import org.neo4j.internal.kernel.api.IndexCapability;
+import org.neo4j.internal.kernel.api.IndexOrder;
+import org.neo4j.internal.kernel.api.IndexValueCapability;
+import org.neo4j.values.storable.ValueGroup;
+
+class IndexReference implements org.neo4j.internal.kernel.api.CapableIndexReference
```
MishaDemianenko: we have IndexReference class that implements IndexReference interface would be nicer to rename first one

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexQuery.java
```diff
@@ -175,6 +176,13 @@ public boolean acceptsValueAt( PropertyCursor property )
         return acceptsValue( property.propertyValue() );
     }
 
+    /**
+     * todo Is this useful?
```
MishaDemianenko: not sure not sure
also it looks like we never return null from here

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexValueCapability.java
```diff
@@ -0,0 +1,47 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.internal.kernel.api;
+
+public enum IndexValueCapability
```
MishaDemianenko: would be really nice to get a description what it is about, also what those constants means (maybe having description about value capability in general will be enough)

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexValueCapability.java
```diff
@@ -0,0 +1,47 @@
...
+{
+    YES( 3 ),
+    PARTIAL( 2 ),
+    NO( 1 );
+
+    /**
+     * Higher order indicate a higher capability.
+     */
+    int order;
```
MishaDemianenko: private

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/IndexValueCapability.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Positive result if this capability is higher than other.
+     * Negative result if this capability is lower that other.
+     * Zero if this has same capability as other.
+     */
+    public int compare( IndexValueCapability other )
+    {
+        return order - other.order;
```
MishaDemianenko: Integer.compare

MishaDemianenko (changes requested):
A couple of comments here and there.

Also as a side note, i think we need to do something with tons of indexes related value objects that exist there in various reincarnations: descriptors, meta, references - it's too much they all do almost the same things, we only make api and reading experience more and more complex

community/kernel/src/main/java/org/neo4j/kernel/api/schema/index/IndexDescriptor.java
fickludd: Hehe, whoops!

fickludd (approved):

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
burqen: @MishaDemianenko Please read this discussion regarding passing `null` as `IndexOrder`.
We flipped back and forth and landed on `null` in the end, but I'm willing to go another lap ;)

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaNumberIndexReader.java
burqen: Please read thread above with @fickludd . (I'm absolutely willing to consider this again)

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NodeValueIterator.java
burqen: Nope. Legacy. Nice catch!

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NodeValueIterator.java
burqen: I'm working on a PR that will make `IndexReader` return `PrimitiveLongResourceIterator` instead. But right now we rely on this snippet from `NodeValueIndexProgressor`
```
    @Override
    public void close()
    {
        if ( ids instanceof Resource )
        {
            ((Resource) ids).close();
        }
    }
```

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionSchemaIndexProvider.java
burqen: We need fusion to make this decision about ordering because he's responsible for combining the results and thus affect the ordering. The value capability can be left completely to underlying implementations and are combined with `UnionIndexCapability`.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/IndexReference.java
burqen: Suggestions: `IndexHandle`, `CapableIndexReference` (same as interface), `IndexReferenceImpl`, ... hmm... do you have any suggestions?

burqen:
@MishaDemianenko Thanks for the feedback! I've addressed all comments except for
 - `IndexReference` class name
 - Using `null` as `IndexOrder`

Please come back to me about those 😄

burqen:
@MishaDemianenko Last two issues are now addessed

MishaDemianenko (approved):

PR #10415, opened 2017-11-10
Forwarding message in constructor calls of internal CypherException

sherfert:
For some subclasses of the version specific CypherException, the
message was not forwarded to the super constructor.

During the forward merge to 3.3, you have to add the same change to `IndexHintException`


craigtaverner (changes requested):
I think we need some acceptance tests to demonstrate the desired new behaviour.

sherfert:
I just tried to write a (failing) acceptance test. The error messages were passed onto the surface correctly also _before_ this PR. The reason is that the changes are in the internal exception, and the _mapping_ to public exception forwarded the message correctly.

Ergo, this only affects debugging (which is also the place I encountered the problem)

sherfert:
Moved to 3.4: #10442

PR #10418, opened 2017-11-10
One kernel to rule them all

fickludd:
The Kernel API introduced some new interfaces around how to interact with the Kernel and create transactions etc. Of course, we already had some interfaces for this, so this PR start the process of joining them up together. 

This is still intermediary state, but now the regular lifecycle-attached Kernel implements the Kernel API, and we can use it as the entry-point. This let's us create stable calling code, and implement benchmarks etc.

Moving forward, the `KernelTransaction` and `org.neo4j.internal.kernel.api.Transaction` interfaces will be merged. Also `InwardKernel` and `org.neo4j.internal.kernel.api.Kernel`. 

Note that docs are fixed in https://github.com/neo4j/neo4j-documentation/pull/154. Merge together.


community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
```diff
@@ -22,23 +22,31 @@
...
 
     /**
-     * Commit this transaction.
-     * @return the id of the commited transaction, or {@link READ_ONLY_TRANSACTION} if this was a read only transaction.
+     * Marks this transaction as successful. When this transaction later gets {@link #close() closed}
+     * its changes, if any, will be committed. If this method hasn't been called or if {@link #failure()}
+     * has been called then any changes in this transaction will be rolled back as part of {@link #close() closing}.
      */
-    long commit();
+    void success();
```
chrisvest, line 48: Looks like the `READ_ONLY_TRANSACTION` constant is no longer used. We already have `READ_ONLY` (and `ROLLBACK`) in `KernelTransaction` anyway.

But supposedly the transaction id is still useful, right? I'm thinking that maybe this interface shouldn't be `AutoCloseable` afterall.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
```diff
@@ -74,16 +74,19 @@
...
+                    Supplier<ExplicitIndexTransactionState> explicitIndexes )
     {
-        read = engine.storeReadLayer();
-        statement = read.newStatement();
-        NeoStores stores = engine.testAccessNeoStores();
+        this.read = engine.storeReadLayer();
+        this.statement = statement; // use provided statement, to assert no leakage
+        this.explicitIndexes = Suppliers.lazySingleton( explicitIndexes );
+
+        StoreHolder stores = engine.stores();
```
chrisvest: The `StorageEngine` should ideally not expose direct access to the underlying stores, but instead hide them behind interfaces such as `StoreReadLayer`, and eventually `org.neo4j.internal.kernel.api.Read` I suppose.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NewKernel.java
```diff
@@ -0,0 +1,62 @@
...
+    public CursorFactory cursors()
+    {
+        return cursors;
+    }
+
+    @Override
+    public org.neo4j.internal.kernel.api.Session beginSession( SecurityContext securityContext )
+    {
+        return new KernelSession( engine, kernel, securityContext );
+    }
```
chrisvest: Java supports covariant return types, so you can make this method return `KernelSession` without breaking the type safety of the interface. Would be easier on the eye.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageEngine.java
```diff
@@ -399,6 +400,12 @@ public void satisfyDependencies( DependencySatisfier satisfier )
     }
 
     @Override
+    public StoreHolder stores()
+    {
+        return neoStores;
+    }
```
chrisvest: I hope this goes away in the future.

enterprise/kernel/src/test/java/org/neo4j/kernel/enterprise/builtinprocs/TestKernelTransaction.java
```diff
@@ -0,0 +1,191 @@
...
+import org.neo4j.internal.kernel.api.security.SecurityContext;
+import org.neo4j.kernel.api.KernelTransaction;
+import org.neo4j.kernel.api.Statement;
+import org.neo4j.kernel.api.exceptions.Status;
+import org.neo4j.kernel.api.exceptions.TransactionFailureException;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class TestKernelTransaction implements KernelTransaction
```
chrisvest: This is a `StubKernelTransaction`, not a test.

chrisvest (commented):
I had a few comments. Things still look very in-between.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
fickludd: Yes you are right, I merged the constants.

The transaction ID seems only to be used by `KernelTransaction` users via `KernelTransaction.CloseListener` (which is implemented internally to unbind a transaction from it's Thread), via `KernelTransaction.getTransactionId()` which is used to implement the public `org.neo4j.graphdb.event.TransactionData` interface, and via `KernelTransaction.closeTransaction()` which is not used apart from a few tests. In the first two cases we are not in the position to actually close the transaction, so having the `long commit()` method would not help. In the last case we are actually given this exact functionality, but it isn't used. Let's keep the AutoCloseable approach for now?

fickludd:
@chrisvest We think we managed to plug the abstraction hole of stores leaking everywhere. WDYT?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
```diff
@@ -206,43 +197,44 @@ public int propertyKey( String name )
     @Override
     PageCursor nodePage( long reference )
     {
-        return nodeStore.openPageCursorForReading( reference );
+//        return read.nodes().openPageCursor( reference );
```
chrisvest: Commented out code.

chrisvest (approved):

pontusmelke:
@fickludd ~have you prepared a PR for the docs build? Please do so we can avoid red builds.~ Sorry can't read apparently

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Session.java
```diff
@@ -19,12 +19,30 @@
...
 {
-    Transaction beginTransaction();
+    /**
+     * Begin new explicit transaction.
+     *
+     * @return The new transaction
+     */
+    default Transaction beginTransaction() throws KernelException
+    {
+        return beginTransaction( Transaction.Type.explicit );
```
burqen: I don't think default implementations should be used in this abstract class kind of style. Implementors are capable of doing this themselves.

community/kernel/src/main/java/org/neo4j/kernel/api/KernelTransaction.java
```diff
@@ -92,10 +87,6 @@
          */
         void notify( long txId );
     }
-
-    long ROLLBACK = -1;
-    long READ_ONLY = 0;
-
     /**
```
burqen: Missing new line between previous method and javadoc

burqen (changes requested):
Thank you for the cleanly separated commits! 🎉 
Changes look good to me. Two small comments.

fickludd:
@burqen thanks for noticing, I actually detoured slightly just for that reason =). Requested changes are changed.

burqen (approved):

PR #10426, opened 2017-11-13
Add NodeHashJoin as an operator for the slotted runtime

systay:


fickludd:
I haven't review this properly, but from the production code changes I think this PR will not hinder or have to be changed due to the `SlotMachine` issues we have discussed. So let's merge this in when you approve @pontusmelke.

PR #10428, opened 2017-11-13
3.4 disable double[] and float[] bit-compaction

craigtaverner:
The DynamicArrayStore bit-compacts all arrays by trimming leading 0's, down to the size of the least trimmable element in the array. However, in the case of double[] and float[] this is a lot of work for very little gain (often double trims from 64 bits to only 63 or 62). This change causes the DynamicArrayStore to skip bit-compaction entirely for double[] and float[], and instead byte align the values, which should be faster.

We have done benchmarks for adding properties using the jmh-benchmarks and see some performance improvements.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/DynamicArrayStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/DynamicArrayStore.java
@@ -93,7 +93,7 @@ public class DynamicArrayStore extends AbstractDynamicStore
         }
     }
 
-    private static byte[] createBitCompactedArray(ShortArray type, Object array )
+    private static byte[] createBitCompactedArray( ShortArray type, Object array )
     {
         Class<?> componentType = array.getClass().getComponentType();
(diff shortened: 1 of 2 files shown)


tinwelint:
Just to be clear, this bit compaction doesn't change the format, right? My understanding is that existing bit-compacted float/double arrays can still be read.

craigtaverner:
Correct, the format is identical. It will claim to have bit-compacted to 64 bits, and will store all the same data that would occur in that case. The reading code is unchanged. Only the writing code has changed to simply skip the compaction algorithm, and then write to the old format.

The compaction algorithm is still used for short arrays that can fit into the PropertyStore, but for doubles that means arrays no longer than length=3. The big compaction does not help there either, but I did not remove it.

One further idea I had was an alternative bit compaction for doubles that splits the exponent and the fraction and bit compacts them separately. For doubles that are close to integers, this could lead to good compression. But that would require changing the store format, with all the consequences that implies. So that has not been done.

tinwelint (approved):

PR #10429, opened 2017-11-13
Do not use primitive equals on refs

pontusmelke:
In some case we end up with an `Equals` where one of the variables is
a longslot and one is a projected refslot. Comparing these using
`PrimitiveEquals` will not work, and what actually happened is that we
failed on a scala assert before doing the actual comparison.

changelog: Fixed an error in slotted runtime where a runtime failure happened when a reference node was being compared to a primitive node

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/LdbcAcceptanceTest.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/LdbcAcceptanceTest.scala
@@ -22,6 +22,7 @@ package org.neo4j.internal.cypher.acceptance
 import org.neo4j.cypher.ExecutionEngineFunSuite
 import org.neo4j.cypher.internal.compatibility.v3_3.runtime.planDescription.InternalPlanDescription
 import org.neo4j.cypher.internal.compatibility.v3_3.runtime.planDescription.InternalPlanDescription.Arguments.EstimatedRows
+import org.neo4j.internal.cypher.acceptance.CypherComparisonSupport.Configs
 
 /**
   * Runs the 14 LDBC queries and checks so that the result is what is expected.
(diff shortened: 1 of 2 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/LdbcAcceptanceTest.scala
```diff
@@ -72,6 +73,34 @@ class LdbcAcceptanceTest extends ExecutionEngineFunSuite with CypherComparisonSu
...
+                      |ORDER BY postCount DESC, tagName ASC
+                      |LIMIT {3}""".stripMargin
+    eengine.execute(LdbcQueries.Query4.createQuery, LdbcQueries.Query4.createParams)
+
+
+    val params: Map[String, Any] = Map("1" -> 1, "2" ->  "tag1-ᚠさ丵פش", "3" -> 10)
+
+    val result =
+    // when
+      executeWith(Configs.CommunityInterpreted, ldbcQuery, params = params)
```
Lojjs, line 98: Configs.CommunityInterpreted means that slotted is not supposed to succeed for the query. Is this really what you intended @pontusmelke ?

pontusmelke: Yes it was picking it up before but failed, now it will use interpreted instead. It is not much work to get it to support `Equals(longslot, refslot)` but I thought it was more prudent to disable in 3.3 and add support for it in 3.4.

Lojjs: That sounds reasonable. Otherwise I think it looks fine. Should I forward merge it to 3.4 for now and then you will add the support for ` Equals(longslot,  refslot)`later?

Lojjs (approved):

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/LdbcAcceptanceTest.scala
pontusmelke: yes that will happen as separate work.

PR #10430, opened 2017-11-13
Cleaner toString for paths in results

craigtaverner:
Since values were introduced, doing a path.toString on paths returned from results would give an ugly inner class description from BaseToObjectValueWriter. This PR uses the pattern seen in RelationshipProxy to provide either a detailed toString (when within transactions) or a simplified version (with unknowns replaced by `?`) when not in transactions.


ragadeeshu (approved):

MishaDemianenko:
@jexp please correct me if I wrong here we had some issues in the past from users that actually using this output in one way or another right? As far as I remember it is so and we need to include this PR in a release notes.

OliviaYtterbrink (approved):

PR #10435, opened 2017-11-14
Make sure all instances are properly closed in Cluster test class

RagnarW:
This adds an error handler to avoid instance shutdown to be aborted
if any exception is thrown. The error handler will collect any
exceptions caught during instance shutdown and then throw after
shutdowns are completed. The thrown exception will be a RuntimeException
with any additional exceptions added as suppressed.

### DIFF
--- a/enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/Cluster.java
+++ b/enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/Cluster.java
@@ -178,8 +178,8 @@ public class Cluster
     {
         try ( ErrorHandler errorHandler = new ErrorHandler( "Error when trying to shutdown cluster" ) )
         {
-            shutdownCoreMembers( errorHandler );
             shutdownReadReplicas( errorHandler );
+            shutdownCoreMembers( errorHandler );
         }
     }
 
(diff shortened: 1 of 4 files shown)


enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/Cluster.java
```diff
@@ -174,21 +174,37 @@ public ReadReplica addReadReplicaWithIdAndMonitors( @SuppressWarnings( "SamePara
...
     }
 
-    public void shutdown() throws ExecutionException, InterruptedException
+    public void shutdown()
     {
-        shutdownReadReplicas();
-        shutdownCoreMembers();
+        try ( ErrorHandler errorHandler = new ErrorHandler( "Error when trying to shutdown cluster" ) )
+        {
+            shutdownCoreMembers( errorHandler );
```
martinfurmanski: You switched the order of cores and read replicas here.

martinfurmanski (changes requested):

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/discovery/Cluster.java
RagnarW: Thanks. I'll update

martinfurmanski (approved):

PR #10440, opened 2017-11-14
Introduce wait for getting Leader in RaftReplicator.

andrewkerr9000:
Without a wait new TimeoutException and NoLeaderFoundException would be
generated constantly until a Leader has been elected.

Suggestions for an appropriate RetryStrategy welcomed. As is, will retry every half second.

This isn't an easy thing to test.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/replication/RaftReplicator.java
```diff
@@ -92,6 +97,8 @@ public RaftReplicator( LeaderLocator leaderLocator, MemberId me,
             catch ( NoLeaderFoundException e )
             {
                 log.debug( "Could not replicate operation " + operation + " because no leader was found. Retrying.", e );
+                wait( leaderTimeout.getMillis() );
```
lutovich: Maybe `Thread.sleep(leaderTimeout.getMillis())`?

andrewkerr9000:
Replaced `wait` for `sleep`, added unit test (that would have failed with `wait`)

phughk (approved):

PR #10442, opened 2017-11-14
Forwarding message in constructor calls of internal CypherException

sherfert:
For some subclasses of the version specific CypherException, the
message was not forwarded to the super constructor.


craigtaverner (changes requested):
Just wondering why you did not make the same change to `LoadCsvStatusWrapCypherException`?

sherfert:
Actually, in the public exception `LoadCsvStatusWrapCypherException`, they distinguish between `cause.message` and `extraInfo`. Thus I didn't rename it. But I am now forwarding that in the internal one.

craigtaverner (approved):

PR #10443, opened 2017-11-14
Verbose page cache flush tracer

MishaDemianenko:
Introduce verbose page cache flush tracer that will
log page cache events that happen during page cache flush together with
some additional information: flush speed, number of pages affected by flush
etc.
Useful for investigation purposes when there is a need to see how much pages
affected, how much data is flushed, how fast flush is actually progress.

### DIFF
--- a/enterprise/kernel/src/main/java/org/neo4j/kernel/monitoring/tracing/VerboseTracerFactory.java
+++ b/enterprise/kernel/src/main/java/org/neo4j/kernel/monitoring/tracing/VerboseTracerFactory.java
@@ -17,43 +17,28 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
-package org.neo4j.kernel.stresstests.transaction.checkpoint.tracers;
+package org.neo4j.kernel.monitoring.tracing;
 
+import org.neo4j.helpers.Service;
 import org.neo4j.io.pagecache.tracing.PageCacheTracer;
-import org.neo4j.kernel.impl.transaction.tracing.CheckPointTracer;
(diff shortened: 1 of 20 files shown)


enterprise/kernel/src/main/java/org/neo4j/kernel/monitoring/tracing/VerboseTracerFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.neo4j.helpers.Service;
+import org.neo4j.io.pagecache.tracing.PageCacheTracer;
+import org.neo4j.kernel.impl.util.JobScheduler;
+import org.neo4j.kernel.monitoring.Monitors;
+import org.neo4j.logging.Log;
+import org.neo4j.time.SystemNanoClock;
+
+@Service.Implementation( TracerFactory.class )
+public class VerboseTracerFactory extends DefaultTracerFactory
```
chrisvest, line 30: This won't be available unless you add it to a META-INF/services file.

MishaDemianenko: unless you rely on ServiceProcessor or any other auto generator to do that automatically:) he

chrisvest: Are you sure that's still plugged in and running?

MishaDemianenko: yep, will gonna add integration test that uses verbose tracer

MishaDemianenko:
@chrisvest integration test for newly introduced tracer added

chrisvest (approved):

PR #10449, opened 2017-11-15
Consistent import of neo4j 2.3.X (to be 2.3.11)

sherfert:
There was 2.3.11 in one location and 2.3.10 in three locations.

### DIFF
--- a/enterprise/cypher/cypher/pom.xml
+++ b/enterprise/cypher/cypher/pom.xml
@@ -164,7 +164,7 @@
     <dependency>
       <groupId>org.neo4j</groupId>
       <artifactId>neo4j-cypher-compiler-2.3</artifactId>
-      <version>2.3.10</version>
+      <version>2.3.11</version>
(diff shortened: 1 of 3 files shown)


enterprise/cypher/cypher/pom.xml
```diff
@@ -164,7 +164,7 @@
     <dependency>
       <groupId>org.neo4j</groupId>
       <artifactId>neo4j-cypher-compiler-2.3</artifactId>
-      <version>2.3.10</version>
+      <version>2.3.11</version>
```
MishaDemianenko, line 167: why not declare them in one place in a root pom and only import in all those poms instead of changing all of them always?

Lojjs: We like your thinking @MishaDemianenko but will do that as a separate PR ( @sherfert has made a card at our wall for it)

Lojjs (approved):

PR #10450, opened 2017-11-15
Renaming and javadocs around SlotAllocation

fickludd:
PipelineInformation is now SlotConfiguration

### DIFF
--- a/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotConfiguration.scala
+++ b/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotConfiguration.scala
@@ -26,17 +26,15 @@ import org.neo4j.cypher.internal.v3_4.logical.plans.LogicalPlan
 
 import scala.collection.{immutable, mutable}
 
-object PipelineInformation {
-  def empty = new PipelineInformation(mutable.Map.empty, 0, 0)
+object SlotConfiguration {
+  def empty = new SlotConfiguration(mutable.Map.empty, 0, 0)
 
-  def apply(slots: Map[String, Slot], numberOfLongs: Int, numberOfReferences: Int): PipelineInformation = {
-
+  def apply(slots: Map[String, Slot], numberOfLongs: Int, numberOfReferences: Int): SlotConfiguration = {
     val stringToSlot = mutable.Map(slots.toSeq: _*)
-    new PipelineInformation(stringToSlot, numberOfLongs, numberOfReferences)
-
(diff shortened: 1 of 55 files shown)


henriknyman:
Couldn't we do this together with an introduce `PhysicalPlan` PR to minimize the merge pain?

fickludd:
@henriknyman I suggest to introduce `PhysicalPlan` in https://github.com/neo4j/neo4j/pull/10460, which builds on this PR. Is that an OK solution?

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotConfiguration.scala
```diff
@@ -159,25 +68,24 @@ class PipelineInformation(private val slots: mutable.Map[String, Slot],
...
-  def add(key: String, slotInformation: Slot): Unit = slotInformation match {
+  def add(key: String, slot: Slot): Unit = slot match {
     case LongSlot(_, nullable, typ) => newLong(key, nullable, typ)
     case RefSlot(_, nullable, typ) => newReference(key, nullable, typ)
   }
 
-  def breakPipelineAndClone(): PipelineInformation = {
-    val newPipeline = new PipelineInformation(this.slots.clone(), numberOfLongs, numberOfReferences)
+  def copy(): SlotConfiguration = {
+    val newPipeline = new SlotConfiguration(this.slots.clone(), numberOfLongs, numberOfReferences)
```
pontusmelke, line 180: Maybe would be nicer to have alias and slotAlias part of constructor,
something like

```
class SlotConfiguration(private val slots: mutable.Map[String, Slot],
                          val initialNumberOfLongs: Int,
                          val initialNumberOfReferences: Int,
                          aliases: mutable.Set[String] = mutable.Set(),
                          slotAliases: SlotMap = new SlotMap
                ) 
```
where
```
type SlotMap =  mutable.HashMap[Slot, mutable.Set[String]] with mutable.MultiMap[Slot, String]
```

systay:
This has been annoying me for some time (I realise it's mostly my fault). Thanks for taking it on!

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotConfiguration.scala
fickludd: I find this change attractive, but wasn't very happy with the result when I tried to do it. Maybe we can change that in a next step? I get ugliness when cloning the `SlotMap` type.

fickludd:
@systay it's been annoying me too! I just had to fix it before continuing.

PR #10452, opened 2017-11-15
Avoid creating the query.log file if query logging is not actually enabled, make the `dbms.logs.query.rotation.size` and the `dbms.logs.query.rotation.keep_number` settings dynamic, so they can be updated at runtime.

chrisvest:


ragadeeshu (commented):
Looks good. But do we test dynamic changing of `log_queries_rotation_threshold`?

ragadeeshu (approved):

PR #10456, opened 2017-11-15
Fix security logging with multiple auth providers

henriknyman:
Improves security logging when multiple auth providers are
configured.

If a user is successfully authenticated by one auth provider we now
do not log authentication error messages to the security log from
another auth provider.

Also fixes a bug where we could get successful login messages in the
security log upon authenentication failure with multiple auth providers
enabled.

### DIFF
--- a/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
+++ b/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
@@ -101,10 +101,26 @@ class MultiRealmAuthManager implements EnterpriseAuthAndUserManager
         {
             securityContext = new StandardEnterpriseSecurityContext(
                     this, (ShiroSubject) securityManager.login( null, token ) );
-            if ( logSuccessfulLogin )
+            AuthenticationResult authenticationResult = securityContext.subject().getAuthenticationResult();
+            if ( authenticationResult == AuthenticationResult.SUCCESS )
             {
-                securityLog.info( securityContext, "logged in" );
+                if ( logSuccessfulLogin )
+                {
+                    securityLog.info( securityContext, "logged in" );
+                }
             }
+            else if ( authenticationResult == AuthenticationResult.PASSWORD_CHANGE_REQUIRED )
+            {
+                securityLog.info( securityContext, "logged in (password change required)" );
+            }
+            else
+            {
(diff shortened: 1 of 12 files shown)


enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
```diff
@@ -128,20 +144,28 @@ public EnterpriseSecurityContext login( Map<String,Object> authToken ) throws In
...
+                        escape( token.getPrincipal().toString() ),
+                        cause != null && cause.getMessage() != null ? " (" + cause.getMessage() + ")" : "" );
                 throw new AuthProviderFailedException( e.getCause().getMessage(), e.getCause() );
             }
             securityContext = new StandardEnterpriseSecurityContext( this,
                     new ShiroSubject( securityManager, AuthenticationResult.FAILURE ) );
-            securityLog.error( "[%s]: failed to log in: invalid principal or credentials",
-                    escape( token.getPrincipal().toString() ) );
+            Throwable cause = e.getCause();
+            Throwable causeCause = e.getCause() != null ? e.getCause().getCause() : null;
```
OliviaYtterbrink, line 164: could replace `e.getCause()` with `cause`

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
```diff
@@ -101,10 +101,26 @@ public EnterpriseSecurityContext login( Map<String,Object> authToken ) throws In
...
+            }
+            else if ( authenticationResult == AuthenticationResult.PASSWORD_CHANGE_REQUIRED )
+            {
+                securityLog.info( securityContext, "logged in (password change required)" );
             }
+            else
+            {
+                String errorMessage = ((StandardEnterpriseSecurityContext.NeoShiroSubject) securityContext.subject())
+                        .getAuthenticationFailureMessage();
+                securityLog.error( "[%s]: failed to log in: %s", escape( token.getPrincipal().toString() ), errorMessage );
```
OliviaYtterbrink, line 120: could use `securityLog.error( securityContext, "failed to log in: %s", errorMessage );`

OliviaYtterbrink (approved):
Good change, just some minor comments.

PR #10457, opened 2017-11-15
Refresh raft timeouts on message recieve

andrewkerr9000:
Introduce pre handling for raft messages that will refresh election timeouts on heartbeat and append entries requests.

This is not on the same thread as the main raft message handling, which may block to download a snapshot and thus fail to process heartbeats.

changelog: Raft heartbeat messages refresh election timeouts on receive rather than on process.
This addresses unnecessary elections caused by the raft message processing being delayed and failing to reset election timeouts in a timely manner.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/ConsensusModule.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/ConsensusModule.java
@@ -20,6 +20,7 @@
 package org.neo4j.causalclustering.core.consensus;
 
 import java.io.File;
+import java.time.Duration;
 
 import org.neo4j.causalclustering.core.CausalClusteringSettings;
 import org.neo4j.causalclustering.core.EnterpriseCoreEditionModule;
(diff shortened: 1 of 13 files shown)


andrewkerr9000:
One thing I'm not sure about is exposing RaftState outside of RaftMachine and sharing between threads. It's needed for the current term. I've limited the interface to ReadableRaftState, but wondering if I should restrict the interface to term-only.

phughk (approved):

andrewkerr9000:
A significant minority of soak tests are failing with `org.neo4j.driver.v1.exceptions.ServiceUnavailableException`. 
http://ec2-34-252-232-77.eu-west-1.compute.amazonaws.com:21000/load-generator/workload/61aa9/config/results

martinfurmanski (approved):
Would be nice if CoreServerModule was just referencing a single RaftMessageHandler and was further unaware of any details and not instantiating anything Raft specific. Perhaps even the BatchingMessageHandler is debatable.

The code for checking clusterId is now duplicated in both LeaderAvailabilityHandler and RaftMessageHandler.

I think above two points means we need some refactoring, but it doesn't have to be part of this PR. Overall the functionality looks good.

PR #10458, opened 2017-11-16
Simplify slot allocation

henriknyman:
Makes allocating a slot for a variable an idempotent operation,
so that individual steps in the slot allocation pass does not need
to worry about if a slot has already been allocated for a variable
within the same pipepline by a previous step.

The slot may have slightly different nullability and specific type as
long as it is compatible.
- If nullability differs, the nullability is conservatively increased.
- If types differs (but are compatible) the least specific type
is chosen.

If the slots are incompatible a compile-time exception is still thrown.

This also fixes CartesianProduct, ValueHashJoin and NodeHashJoin to work
with shared argument slots, and a bug in slot allocation for arguments.


enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotConfiguration.scala
```diff
@@ -189,21 +188,73 @@ class SlotConfiguration(private val slots: mutable.Map[String, Slot],
...
+        throw new InternalException(s"Tried overwriting already taken variable name $key as $slot (was: ${existingSlot})")
+      }
+      // Reuse the existing (compatible) slot
+      unifyTypeAndNullability(key, existingSlot, slot)
+    }
+    else {
+      slots.put(key, slot)
+      slotAliases.addBinding(slot, key)
+      numberOfLongs = numberOfLongs + 1
+    }
```
fickludd: An alternative way on handling this option would be

```
slots.get(key) match {
  case Some(existingSlot) =>
      ...
  case None =>
      ...
}
```

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotConfiguration.scala
```diff
@@ -189,21 +188,73 @@ class SlotConfiguration(private val slots: mutable.Map[String, Slot],
...
+        throw new InternalException(s"Tried overwriting already taken variable name $key as $slot (was: ${existingSlot})")
+      }
+      // Reuse the existing (compatible) slot
+      unifyTypeAndNullability(key, existingSlot, slot)
+    }
+    else {
+      slots.put(key, slot)
+      slotAliases.addBinding(slot, key)
+      numberOfReferences = numberOfReferences + 1
+    }
```
fickludd: See `Option[Slot]` match above.

enterprise/cypher/physical-planning/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotConfigurationTest.scala
```diff
@@ -0,0 +1,152 @@
...
+    val clone: SlotConfiguration = slots.copy()
+    slots should equal(clone)
+
+    // when
+    slots.newReference("a", nullable = false, CTNode)
+
+    // then
+    slots("a") should equal(RefSlot(0, nullable = false, CTNode))
+    clone.get("a") shouldBe empty
+    clone.numberOfReferences should equal(0)
```
fickludd: I really like this test. It would be even better though if it had some more assertions of that the longs are preserved, and that aliases are still the same.

fickludd (changes requested):
I really like this change in `SlotConfiguration`. And especially the type inference. Also really like the test coverage. Just requesting some more assertions on the deepClone test, then I'm happy!

enterprise/cypher/physical-planning/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotConfigurationTest.scala
henriknyman: I just moved this test, but sure, I'll add some more assertions

fickludd: nice!

fickludd (approved):

systay:
Hmm... I must be missing something here. When would two different operators in the same pipeline want to declare the same variable? I don't understand the context of when this would be used.

fickludd:
@systay this is not common but it happened in a few places in the spike if I remember correctly. On merge relationship it happens every time because of the explicit branching and locking.

Today I discovered that this PR does not play nice with the `ValueHashJoin` so we'll have to think about that some more before merging this.

PR #10463, opened 2017-11-16
Add `--pagecache` parameter to the `neo4j-admin backup` command, with a default value of 8 MiB that now overrides the page cache setting in `neo4j.conf`.

chrisvest:
This parameter makes it possible to specify the page cache memory size
for the backup program; both for writing the copied files, and for doing
the consistency check afterwards.

This setting is 8 MiBs by default, and it always overrides what the
`neo4j.conf` otherwise specifies. The reason for this change in override
priority is that people too often run the online backup command from the
same machine as they are taking the backup from. In those cases,
`neo4j.conf` will be specifying a page cache memory size that is
intended for a running Neo4j database on a dedicated machine, and this
is an unsuitable setting for the backup command – especially when
running on the same machine that is supposedly dedicated to the running
Neo4j database.

This commit needs to be null-merged into 3.4 and redone, because the
oneline backup command code has been substantially changed in the 3.4
branch.

For this reason, and because the change is both small and difficult to
observe and verify via testing, I have decided to not add tests for
these changes. Testing should be added to the 3.4 changes.

changelog: [3.2, 3.3] Add `--pagecache` parameter to the `neo4j-admin backup` command, with a default value of 8 MiB that now overrides the page cache setting in `neo4j.conf`.

### DIFF
--- a/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommand.java
@@ -38,6 +38,7 @@ import org.neo4j.commandline.arguments.common.OptionalCanonicalPath;
 import org.neo4j.consistency.ConsistencyCheckService;
 import org.neo4j.consistency.ConsistencyCheckSettings;
 import org.neo4j.consistency.checking.full.CheckConsistencyConfig;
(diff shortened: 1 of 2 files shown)


MishaDemianenko (approved):
Awesome! :octocat:

spacecowboy (approved):
double approve!

PR #10464, opened 2017-11-16
Fixed broken tests related to block devices

phughk:
Adds support/fixes tests for performing backups on blockdevices

### DIFF
--- a/enterprise/backup/src/main/java/org/neo4j/backup/BackupCopyService.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/BackupCopyService.java
@@ -21,36 +21,45 @@ package org.neo4j.backup;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Optional;
+import java.util.Iterator;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
+import org.neo4j.com.storecopy.FileMoveAction;
+import org.neo4j.com.storecopy.FileMoveProvider;
 import org.neo4j.commandline.admin.CommandFailed;
-import org.neo4j.commandline.admin.OutsideWorld;
+import org.neo4j.io.pagecache.PageCache;
+import org.neo4j.kernel.configuration.Config;
 
 import static java.lang.String.format;
+import static org.neo4j.graphdb.factory.GraphDatabaseSettings.logs_directory;
+import static org.neo4j.graphdb.factory.GraphDatabaseSettings.store_internal_log_path;
 
 public class BackupCopyService
 {
     private static final int MAX_OLD_BACKUPS = 1000;
 
-    private final OutsideWorld outsideWorld;
+    private final PageCache pageCache;
 
-    public BackupCopyService( OutsideWorld outsideWorld )
+    private final FileMoveProvider fileMoveProvider;
+
+    public BackupCopyService( PageCache pageCache, FileMoveProvider fileMoveProvider )
     {
-        this.outsideWorld = outsideWorld;
+        this.pageCache = pageCache;
+        this.fileMoveProvider = fileMoveProvider;
     }
 
     public void moveBackupLocation( File oldLocation, File newLocation ) throws CommandFailed
     {
         try
         {
-            outsideWorld.fileSystem().renameFile( oldLocation, newLocation );
+            Iterator<FileMoveAction> moves = fileMoveProvider.traverseGenerateMoveActions( oldLocation ).iterator();
+            while ( moves.hasNext() )
+            {
+                moves.next().move( newLocation );
+            }
         }
         catch ( IOException e )
         {
@@ -58,9 +67,15 @@ public class BackupCopyService
         }
     }
 
+    public void clearLogs( File neo4jHome )
+    {
+        File logsDirectory = Config.defaults( logs_directory, neo4jHome.getPath() ).get( store_internal_log_path );
+        logsDirectory.delete();
+    }
+
     boolean backupExists( File destination )
     {
-        File[] listFiles = outsideWorld.fileSystem().listFiles( destination );
+        File[] listFiles = pageCache.getCachedFileSystem().listFiles( destination );
         return listFiles != null && listFiles.length > 0;
     }
 
@@ -74,6 +89,13 @@ public class BackupCopyService
         return findAnAvailableBackupLocation( desiredBackupLocation, "%s.temp.%d" );
     }
 
+    /**
+     * Given a desired file name, find an available name that is similar to the given one that doesn't conflict with already existing backups
+     *
+     * @param file desired ideal file name
+     * @param pattern pattern to follow if desired name is taken (requires %s for original name, and %d for iteration)
+     * @return the resolved file name which can be the original desired, or a variation that matches the pattern
+     */
     private File findAnAvailableBackupLocation( File file, String pattern )
     {
         if ( backupExists( file ) )
@@ -84,7 +106,7 @@ public class BackupCopyService
 
             return availableAlternativeNames( file, pattern )
                     .peek( countNumberOfFilesProcessedForPotentialErrorMessage )
-                    .filter( f -> !backupExists(f) )
+                    .filter( f -> !backupExists( f ) )
                     .findFirst()
                     .orElseThrow( () -> new RuntimeException(
                             String.format( "Unable to find a free backup location for the provided %s. Number of iterations %d", file, counter.get() ) ) );
@@ -100,8 +122,7 @@ public class BackupCopyService
 
     private static File alteredBackupDirectoryName( String pattern, File directory, int iteration )
     {
-        return directory
-                .toPath()
+        return directory.toPath()
                 .resolveSibling( format( pattern, directory.getName(), iteration ) )
                 .toFile();
     }
--- a/enterprise/backup/src/main/java/org/neo4j/backup/BackupStrategyCoordinatorFactory.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/BackupStrategyCoordinatorFactory.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.neo4j.com.storecopy.FileMoveProvider;
 import org.neo4j.commandline.admin.OutsideWorld;
 import org.neo4j.consistency.ConsistencyCheckService;
 import org.neo4j.helpers.progress.ProgressMonitorFactory;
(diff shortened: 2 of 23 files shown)


phughk:
Related to https://github.com/neo-technology/neo4j-blockdevice/pull/49

enterprise/backup/src/main/java/org/neo4j/backup/BackupCopyService.java
```diff
@@ -82,9 +104,8 @@ private File findAnAvailableBackupLocation( File file, String pattern )
             final AtomicLong counter = new AtomicLong( 0 );
             Consumer<File> countNumberOfFilesProcessedForPotentialErrorMessage = generatedBackupFile -> counter.getAndIncrement();
 
-            return availableAlternativeNames( file, pattern )
-                    .peek( countNumberOfFilesProcessedForPotentialErrorMessage )
-                    .filter( f -> !backupExists(f) )
+            return availableAlternativeNames( file, pattern ).peek( countNumberOfFilesProcessedForPotentialErrorMessage )
+                    .filter( f -> !backupExists( f ) )
```
andrewkerr9000: I preferred the old formatting. Likewise lines 118 and 123.

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommandBuilder.java
```diff
@@ -0,0 +1,361 @@
...
+import org.neo4j.io.fs.DefaultFileSystemAbstraction;
+import org.neo4j.kernel.configuration.Config;
+
+import static java.lang.String.format;
+
+/**
+ * This class encapsulates the information needed to perform an online backup against a running Neo4j instance
+ * configured to act as a backup server.
+ * <p>
+ * All backup methods return the same instance, allowing for chaining calls.
```
andrewkerr9000, line 51: Not sure the last line is necessary - it's a builder, I'd expect a fluent API.

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommandBuilder.java
```diff
@@ -0,0 +1,361 @@
...
+     */
+    private String argFrom()
+    {
+        if ( host == null && port == null )
+        {
+            return "";
+        }
+        String address = String.join( ":",
+                Optional.ofNullable( host ).orElse( "" ),
+                Optional.ofNullable( port ).map( port -> Integer.toString( port ) ).orElse( "" ) );
```
andrewkerr9000, line 215: Might be cleaner to store host, port (and other fields) as Optional rather than wrap them here. Intellij will complain, but it's wrong :)

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommandBuilder.java
```diff
@@ -0,0 +1,361 @@
...
+     */
+    private static String[] args( String... args )
+    {
+        List<String> cleanedArgs = Stream.of( args )
+                .filter( Objects::nonNull )
+                .filter( not( String::isEmpty ) )
+                .collect( Collectors.toList() );
+        String[] returnArgs = new String[cleanedArgs.size()];
+        cleanedArgs.toArray( returnArgs );
+        return returnArgs;
```
andrewkerr9000, line 344: Can create an array from a stream directly:  `Stream.of( args ).toArray( String[]::new )`

enterprise/com/src/main/java/org/neo4j/com/storecopy/FileMoveProvider.java
```diff
@@ -0,0 +1,134 @@
...
+     */
+    public Stream<FileMoveAction> traverseGenerateMoveActions( File dir )
+    {
+        return traverseGenerateMoveActions( dir, dir );
+    }
+
+    /**
+     * Copies <b>the contents</b> from the directory to the base target path.
+     * <p>
+     * This is confusing, best read the tests
```
andrewkerr9000: Maybe include an example of usage in Javadoc?

enterprise/com/src/main/java/org/neo4j/com/storecopy/FileMoveProvider.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private List<File> listFiles( File dir )
+    {
+        List<File> pageCacheFiles = asList( safeArray( pageCache.getCachedFileSystem().listFiles( dir ) ) );
+        List<File> fsFiles = safeArray( dir.listFiles() );
+        return Stream.of( pageCacheFiles, fsFiles ).flatMap( List::stream ).collect( Collectors.toList() );
+    }
+
+    private List<File> safeArray( File[] files )
```
andrewkerr9000, line 142: `nullSafeArray` may be better name? There may also be an appropriate place for this in `community/common`.

andrewkerr9000 (commented):
Looks good, my comments are all on style. I'm not familiar enough with the backup though to confidently say there are no issues.

enterprise/backup/src/main/java/org/neo4j/backup/BackupCopyService.java
phughk: Addressed

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommandBuilder.java
phughk: Removed file (included in separate branch)

enterprise/backup/src/main/java/org/neo4j/backup/OnlineBackupCommandBuilder.java
phughk: Removed file (included in separate branch)

enterprise/com/src/main/java/org/neo4j/com/storecopy/FileMoveProvider.java
phughk: Addressed

enterprise/backup/src/main/java/org/neo4j/backup/BackupStrategyCoordinatorFactory.java
```diff
@@ -23,47 +23,54 @@
...
-            PageCache pageCache )
+    /**
+     * Construct a wrapper of supported backup strategies
+     *
+     * @param onlineBackupContext the input of the backup tool, such as CLI arguments, config etc.
+     * @param backupProtocolService the underlying backup implementation for HA and single node instances
+     * @param backupDelegator the backup implementation used for CC backups
+     * @param pageCache the page cache used moving files
+     * @return strategy coordinator that handles the which backup strategies are tried and establishes if a backup was successful or not
+     */
```
chrisvest, line 61: A+ javadocing of parameters, mate.

enterprise/backup/src/main/java/org/neo4j/backup/BackupStrategyWrapper.java
```diff
@@ -74,25 +78,47 @@
...
         {
+            log.info( "Previous backup found, trying incremental backup." );
             PotentiallyErroneousState<BackupStageOutcome> state =
                     backupStrategy.performIncrementalBackup( userSpecifiedBackupLocation, config, userSpecifiedAddress );
             boolean fullBackupWontWork = BackupStageOutcome.WRONG_PROTOCOL.equals( state.getState() );
             boolean incrementalWasSuccessful = BackupStageOutcome.SUCCESS.equals( state.getState() );
 
             if ( fullBackupWontWork || incrementalWasSuccessful )
             {
+                backupCopyService.clearLogs( backupLocation );
```
chrisvest, line 92: What is this `clearLogs` business about?

enterprise/backup/src/main/java/org/neo4j/backup/ParametrisedOutsideWorld.java
```diff
@@ -0,0 +1,138 @@
...
+
+import org.neo4j.commandline.admin.OutsideWorld;
+import org.neo4j.io.IOUtils;
+import org.neo4j.io.fs.DefaultFileSystemAbstraction;
+import org.neo4j.io.fs.FileSystemAbstraction;
+
+/**
+ * An outside world where you can pick and choose which input/output are dummies.
+ */
+public class ParametrisedOutsideWorld implements OutsideWorld
```
chrisvest, line 35: This sounds like it should be in `src/test` instead of `src/main`.

enterprise/com/src/main/java/org/neo4j/com/storecopy/FileMoveActionInformer.java
```diff
@@ -0,0 +1,26 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.com.storecopy;
+
+@FunctionalInterface
+public interface FileMoveActionInformer
```
chrisvest, line 23: I'd call it a predicate instead of an informer.

enterprise/com/src/main/java/org/neo4j/com/storecopy/FileMoveProvider.java
```diff
@@ -0,0 +1,158 @@
...
+        return Stream.concat( moveFiles, traverseDirectories );
+    }
+
+    private boolean isFile( File file )
+    {
+        if ( fileMoveActionInformer.shouldBeManagedByPageCache( file.getName() ) )
+        {
+            return !pageCache.getCachedFileSystem().isDirectory( file );
+        }
+        return file.isFile();
```
chrisvest, line 123: I think you should have a `FileSystemAbstraction` in here, for doing things to files that are not-managed-by-the-page-cache.

enterprise/com/src/main/java/org/neo4j/com/storecopy/FileMoveProvider.java
```diff
@@ -0,0 +1,158 @@
...
+    private List<File> listFiles( File dir )
+    {
+        List<File> pageCacheFiles = asList( safeArray( pageCache.getCachedFileSystem().listFiles( dir ) ) );
+        List<File> fsFiles = safeArray( dir.listFiles() );
+        return Stream.of( pageCacheFiles, fsFiles ).flatMap( List::stream ).collect( Collectors.toList() );
+    }
+
+    private List<File> safeArray( File[] files )
+    {
+        return Arrays.asList( Optional.ofNullable( files ).orElse( new File[]{} ) );
```
chrisvest, line 144: Not sure this is simpler than just having null-checking conditional branching.

enterprise/com/src/main/java/org/neo4j/com/storecopy/FileMoveProvider.java
```diff
@@ -0,0 +1,158 @@
...
+        if ( fileMoveActionInformer.shouldBeManagedByPageCache( file.getName() ) )
+        {
+            return pageCache.getCachedFileSystem().isDirectory( file );
+        }
+        return file.isDirectory();
+    }
+
+    private List<File> listFiles( File dir )
+    {
+        List<File> pageCacheFiles = asList( safeArray( pageCache.getCachedFileSystem().listFiles( dir ) ) );
```
chrisvest, line 137: I don't think you need to `asList` the result of `safeArray`, since it already returns a list.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
```diff
@@ -0,0 +1,258 @@
...
+        subject.traverseGenerateMoveActions( sourceFile, sourceParent ).forEach( moveToDirectory( targetParent ) );
+
+        // then
+        assertEquals( "Garbage data", readFromFile( targetFile ) );
+    }
+
+    private interface RunnableThrowable
+    {
+        void run() throws Throwable;
+    }
```
chrisvest, line 98: This is the same as `org.neo4j.function.ThrowingAction`, so you can use that instead.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+    private Supplier<RuntimeException> failure()
+    {
+        return () -> new RuntimeException( "Fail" );
+    }
+
+    private List<File> safeList( File dir )
+    {
+        return Arrays.asList( Optional.ofNullable( dir ).map( File::listFiles ).orElse( new File[]{} ) );
+    }
```
chrisvest, line 223: This is a copy of the same-named method you have in the product code. It makes the same mistake of not using a FileSystemAbstraction.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
```diff
@@ -0,0 +1,258 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class FileMoveProviderTest
+{
+    private DefaultFileSystemAbstraction defaultFileSystemAbstraction = new DefaultFileSystemAbstraction();
+    private EphemeralFileSystemAbstraction ephemeralFileSystemAbstraction = new EphemeralFileSystemAbstraction();
```
chrisvest, line 58: This works. It _should_ also work if you used two distinct `EphemeralFileSystemAbstraction` instances. That way you'll notice suspicious code patterns, such as the one in `safeList`.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
```diff
@@ -0,0 +1,258 @@
...
+        StoreChannel storeChannel = pageCache.getCachedFileSystem().create( aPageCacheFile );
+        storeChannel.write( ByteBuffer.allocate( 20 ).putChar( 'a' ).putChar( 'b' ) );
+
+        // and some of these files are handled by the page cache
+        when( fileMoveActionInformer.shouldBeManagedByPageCache( any() ) ).thenReturn( false );
+        when( fileMoveActionInformer.shouldBeManagedByPageCache( eq( aPageCacheFile.getName() ) ) ).thenReturn( true );
+
+        // when the files are copied to target location
+        List<FileMoveAction> moveActions = subject.traverseGenerateMoveActions( parentDirectory, parentDirectory )
+                .collect( Collectors.toList() );//.forEach( moveToDirectory( targetDirectory ) );
```
chrisvest: Got some commented out code, there.

enterprise/ha/src/main/java/org/neo4j/kernel/ha/cluster/member/ClusterMember.java
```diff
@@ -67,6 +67,7 @@ public URI getHAUri()
 
     public String getHARole()
     {
+        if (true) return roles.keySet().toString(); // TODO delete this
```
chrisvest: I think you should address this TODO.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
```diff
@@ -0,0 +1,258 @@
...
+        try
+        {
+            BufferedWriter bw = new BufferedWriter( new FileWriter( output ) );
+            bw.write( input );
+            bw.close();
+        }
+        catch ( IOException e )
+        {
+            throw new RuntimeException( e );
+        }
```
chrisvest, line 273: In test code, I usually just let exceptions bubble out. Not having the try-catch clauses improves the signal-to-noise ratio of the test code, because the test doesn't really care if it fails with a RuntimeExceptions that wraps the IOExceptions, or the IOException itself.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
chrisvest: This applies to most of these helper methods, it seems.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+        // and correct files are copied over the page cache
+        File expectedPageCacheCopy = new File( targetDirectory, aPageCacheFile.getName() );
+        assertTrue( expectedPageCacheCopy.toString(), pageCache.getCachedFileSystem().fileExists( expectedPageCacheCopy ) );
+    }
+
+    private Supplier<RuntimeException> failure()
+    {
+        return () -> new RuntimeException( "Fail" );
+    }
```
chrisvest, line 218: Looks like this is unused.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
chrisvest: Looks like you are not actually sing this method.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
```diff
@@ -0,0 +1,258 @@
...
+    {
+        runnableFromThrowable( file::mkdirs ).run();
+        return file;
+    }
+
+    private File createFile( File file )
+    {
+        runnableFromThrowable( file::createNewFile ).run();
+        return file;
+    }
```
chrisvest, line 260: If you just let the tests throw IOException, you don't need to do this dance with lambdas and exceptions.

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+    public PageCacheRule pageCacheRule = new PageCacheRule();
+    private PageCache pageCache;
+    private FileMoveActionInformer fileMoveActionInformer;
+
+    @Before
+    public void setup()
+    {
+        pageCache = pageCacheRule.getPageCache( ephemeralFileSystemAbstraction );
+        fileMoveActionInformer = mock( FileMoveActionInformer.class );
```
chrisvest, line 73: One way to avoid mocking in this case, would be to implement this as a method reference to a `Set::contains` method.

chrisvest (commented):
I had some comments.

phughk:
@chrisvest will merge these changes into his PR

PR #10469, opened 2017-11-17
Use resource iterators for index reads

burqen:
While moving label index and schema index from lucene to native we introduced iterators with the need to be closed. Up until now, those iterators was only closed on exhaustion or when the source (IndexReader or LabelScanReader) was closed.

This PR bubbles the shift from `PrimitiveLongIterator` to `PrimitiveLongResourceIterator` up from `IndexReader` and `LabelScanReader` to `ReadOperations` and `GraphDatabaseFacade`.

Note that no changes are made to cypher.


community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NodeValueIterator.java
```diff
@@ -66,6 +66,7 @@ public void close()
         {
             closed = true;
             progressor.close();
+            progressor = null;
```
ragadeeshu, line 69: Why do we need to null the progressor here?

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/NodeValueIndexProgressor.java
```diff
@@ -38,7 +38,7 @@ public boolean next()
     {
         while ( ids.hasNext() )
         {
-            if ( client.acceptNode( ids.next(), null ) )
+            if ( client.acceptNode( ids.next(), (Value[]) null ) )
```
ragadeeshu, line 41: Are we explicitly typecasting null here?

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveLongResourceCollections.java
```diff
@@ -0,0 +1,139 @@
...
+        public void close()
+        {
+            if ( !closed )
+            {
+                closed = true;
+                closeAll();
+            }
+        }
+
+        // Copied from IOUtils
```
ragadeeshu: Any specific reason that we copy this instead of call it? :)

community/primitive-collections/src/test/java/org/neo4j/collection/primitive/PrimitiveLongResourceCollectionsTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.junit.Test;
+
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.LongPredicate;
+
+import org.neo4j.graphdb.Resource;
+
+import static org.junit.Assert.assertEquals;
+
+public class PrimitiveLongResourceCollectionsTest
```
ragadeeshu, line 31: I get the impression that the iterators do not close their resources when exhausted. Is this the expected behavior?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIterator.java
```diff
@@ -143,4 +135,25 @@ private boolean nextFromCurrent()
...
+    {
+        if ( !closed )
+        {
+            try
+            {
+                cursor.close();
+            }
+            catch ( IOException e )
+            {
+                throw new UncheckedIOException( e );
```
ragadeeshu, line 150: Why unchecked IO exception? Wasn't it a normal one before?

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexReaderTest.java
```diff
@@ -96,45 +93,6 @@ public void closeIteratorMustCloseNativeAndLucene() throws Exception
         verify( luceneIter, times( 1 ) ).close();
     }
 
-    @Test
```
ragadeeshu: What happened to these tests? Is this no longer true?

ragadeeshu (commented):
@burqen I need explanations for a few points I'm unclear on. Apart from that, very nice!

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIterator.java
burqen: Yes it was. But the `Resource` interface does not allow `close` to throw.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NodeValueIterator.java
burqen: Just good practice. "I'm not gonna use this any more, feel free to garbage collect"

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/NodeValueIndexProgressor.java
burqen: Because of vararg we need to define if we are passing in a null element or a null array.

community/primitive-collections/src/test/java/org/neo4j/collection/primitive/PrimitiveLongResourceCollectionsTest.java
burqen: You are correct, they don't close when exhausted, nice catch! I've actually not thought about that for the iterators from `PrimitiveLongResourceCollections`.
The `PrimitiveLongResourceIterator` interface does not say anything about such behaviour explicitly, but what other behaviour would you realistically want when you explicitly create a resource iterator?

Pro
 - Good safety net for when you forget to close
 - Convenient to not be forced to close if you exhaust it (which you most often do)

Con
 - Allow faulty code and hide bugs.
 - More side-effects -> less control and harder to reason about.

For the general purpose of `PrimitiveLongResourceCollections` I think the cons are bigger than the pros.

Side note: Label and schema index iterators do close their resources on exhaust for legacy reasons. Up until now this was the only way to make sure they where closed at all.

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveLongResourceCollections.java
burqen: Introducing a dependency on `neo4j-io` would create a circular dependency.

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexReaderTest.java
burqen: Because they are now Resources they are not allowed to throw checked exceptions on close.

burqen:
@ragadeeshu Thank you for the feedback :D Are you happy with my comments?

burqen:
@ragadeeshu Rebased, in hope that it will solve compilation problem in `compile_docstest`.

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/NodeValueIndexProgressor.java
ragadeeshu: Cool

ragadeeshu (approved):

PR #10478, opened 2017-11-18
Use new API for create and delete node

pontusmelke:
This PR accomplishes two things.
- Adds node creation/deletion, label creation/deletion, create explicit index, remove node from explicit index
- Changes so that the new API is used for node creation and deletion is used both in Cypher and Core API.

We are not yet using the new API for label creation and deletion since in order for that to work properly,  we first need reads to be transaction-state aware and handle constraints. This is planned as the next chunk of work following this PR.


community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
```diff
@@ -282,13 +299,15 @@ public final void graphProperties( org.neo4j.internal.kernel.api.PropertyCursor
     public final void nodeExplicitIndexLookup(
             NodeExplicitIndexCursor cursor, String index, String key, Value value ) throws KernelException
     {
+        ((org.neo4j.kernel.impl.newapi.NodeExplicitIndexCursor) cursor).setRead( this );
```
burqen: Can we decide on one pattern of using qualified names for consistently `org.neo4j.kernel.impl.newapi` or `org.neo4j.internal.kernel.api` and import the other.
I prefer having qualified name in method signature (so for `org.neo4j.internal.kernel.api`), but either way pick one.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
```diff
@@ -97,7 +110,7 @@ public final void nodeIndexSeek(
             if ( j > 0 )
             {
                 filters = Arrays.copyOf( filters, j );
-                target = new NodeValueClientFilter( target, new NodeCursor( this ), new PropertyCursor( this ), filters );
+                target = new NodeValueClientFilter( target, cursors.allocateNodeCursor(), cursors.allocatePropertyCursor(), this, filters );
```
burqen, line 112: Is this the sole reason for turning the dependency around between `Read` and `Cursors`?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeWriteTestBase.java
```diff
@@ -69,25 +77,104 @@ public void shouldRollbackOnFailure() throws Exception
         }
     }
 
+    @Test
+    public void shouldRemoveNode() throws Exception
```
burqen, line 81: Can we add test `shouldRemoveNodeFromTxState`

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeWriteTestBase.java
```diff
@@ -69,25 +77,104 @@ public void shouldRollbackOnFailure() throws Exception
...
     @Test
     public void shouldAddLabelNode() throws Exception
     {
+        // Given
         long node;
         int labelId;
         final String labelName = "Town";
+        try ( org.neo4j.graphdb.Transaction ctx = graphDb.beginTx() )
+        {
+            node = graphDb.createNode( label( labelName ) ).getId();
```
burqen: I think what you meant to do here was to create a node without a label and then add the label in a separate transaction (with the new api).

community/neo4j/src/test/java/org/neo4j/locking/QueryExecutionLocksIT.java
```diff
@@ -107,6 +106,8 @@
 import static org.junit.Assert.assertTrue;
 import static org.neo4j.values.virtual.VirtualValues.EMPTY_MAP;
 
+//import org.neo4j.cypher.internal.javacompat.ExecutionEngine;
```
burqen: Commented import

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeWriteTestBase.java
```diff
@@ -69,25 +77,104 @@ public void shouldRollbackOnFailure() throws Exception
...
         {
-            assertThat(
-                    graphDb.getNodeById( node ).getLabels(),
-                    equalTo( Iterables.iterable( label( labelName ) ) ) );
+            assertThat( graphDb.getNodeById( node ).getLabels(), equalTo( Iterables.iterable( label( labelName ) ) ) );
+        }
+    }
+
+    @Test
+    public void shouldAddLabelNodeOnce() throws Exception
```
burqen, line 156: Additional tests: `shouldAddLabelOnceWithNodeInTxState` and `shouldAddLabelOnceWithLabelInTxState`

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/IndexTxStateUpdaterTest.java
```diff
@@ -0,0 +1,157 @@
...
+        when( readOps.txState() ).thenReturn( txState );
+
+        indexTxUpdater = new IndexTxStateUpdater( storeReadLayer, readOps );
+
+    }
+
+    // LABELS
+
+    @Test
+    public void shouldNotUpdateIndexesOnChangedIrrelevantLabel() throws EntityNotFoundException
```
burqen, line 113: Can we also add test for add and remove label in same tx (no index update should occur).

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/MockStore.java
```diff
@@ -62,6 +62,11 @@ public DynamicRecord nextRecord()
         }
     };
 
+    MockStore( Cursors cursors )
+    {
+        super( cursors, null );//TODO fake txstateholder?
```
burqen: Forgotten or left intentionally?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -19,6 +19,12 @@
  */
 package org.neo4j.internal.kernel.api;
 
+import java.util.Map;
+
+import org.neo4j.internal.kernel.api.exceptions.KernelException;
+
 public interface ExplicitIndexWrite
```
burqen: I would love some javadoc for this interface.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/ExplicitIndexCursorWritesTestBase.java
```diff
@@ -0,0 +1,159 @@
...
+        {
+            ExplicitIndexWrite indexWrite = tx.indexWrite();
+            indexWrite.nodeRemoveFromExplicitIndex( INDEX_NAME, nodeId );
+            tx.success();
+        }
+
+        // Then
+        try ( org.neo4j.graphdb.Transaction ctx = graphDb.beginTx() )
+        {
+            IndexHits<Node> hits = graphDb.index().forNodes( INDEX_NAME ).get( KEY, "this is it" );
```
burqen: Magic string, use static `VALUE `

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/ExplicitIndexCursorWritesTestBase.java
```diff
@@ -0,0 +1,159 @@
...
```
burqen: Magic string, use static `VALUE`

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
burqen: What's the reason for not having `nodeAddToExplicitIndex` here as well?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactions.java
```diff
@@ -142,10 +151,15 @@ public KernelTransactions( StatementLocksFactory statementLocksFactory,
...
-                new ExplicitIndexTransactionStateImpl( indexConfigStore, explicitIndexProviderLookup ) );
+        this.autoIndexing = autoIndexing;
+        this.explicitIndexStore = explicitIndexStore;
+        this.explicitIndexTxStateSupplier = () ->
+                new CachingExplicitIndexTransactionState(
+                        new ExplicitIndexTransactionStateImpl( indexConfigStore, explicitIndexProviderLookup ) );
         this.clock = clock;
         blockNewTransactions();
+        this.cursors = cursors;
+
```
burqen: Empy line

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/ExplicitIndexCursorWritesTestBase.java
```diff
@@ -0,0 +1,159 @@
...
+            IndexHits<Node> hits = graphDb.index().forNodes( INDEX_NAME ).get( KEY, VALUE );
+            assertThat( hits.next().getId(), equalTo( nodeId ) );
+            assertFalse( hits.hasNext() );
+            hits.close();
+            ctx.success();
+        }
+    }
+
+    @Test
+    public void shouldCreateExplicitIndex() throws Exception
```
burqen, line 123: Can we also have test for creating same explicit index twice?

community/kernel/src/main/java/org/neo4j/kernel/api/explicitindex/AutoIndexOperations.java
```diff
@@ -44,8 +46,11 @@ void propertyChanged( DataWriteOperations ops, long entityId, int propertyKeyId,
     void propertyRemoved( DataWriteOperations ops, long entityId, int propertyKey )
         throws AutoIndexingKernelException;
 
+    //TODO this method should be removed once fully migrated
     void entityRemoved( DataWriteOperations ops, long entityId ) throws AutoIndexingKernelException;
```
burqen, line 50: Does it make sense to deprecate to signal this intention?

burqen (changes requested):
All commits except the four latest reviewed. Will continue with the rest tomorrow.

Additional note: Scala code did not receive any review.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
pontusmelke: it is coming to a PR near you soon, stay tuned!

community/kernel/src/main/java/org/neo4j/kernel/api/explicitindex/AutoIndexOperations.java
pontusmelke: I have already done the work for cleaning this up in the next chunk of work. Trying to keep this PR:s smallish - so if you just pretend you didn't see this we'll have it fixed in the next PR :)

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeWriteTestBase.java
pontusmelke: If you mean that we should be able to see the removal you have to wait a little bit, reads are not yet tx state aware. It is coming in the next PR.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeWriteTestBase.java
pontusmelke: If you mean the same thing, it is coming in the next PR where reads are made tx state aware

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/IndexTxStateUpdaterTest.java
pontusmelke: This is controlled by the TransactionState not by IndexTxStateUpdater

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/ExplicitIndexCursorWritesTestBase.java
pontusmelke: Done

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
pontusmelke: All cursor allocation should be done via the factory

community/kernel/src/main/java/org/neo4j/kernel/api/explicitindex/AutoIndexOperations.java
burqen: Thank you that effort, of keeping the PRs as small as possible 🎉

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
burqen: Yes, absolutely! Was just wondering if this was the whole reason for turning the dependency around (which would be completely fine), or if I missed some other subtle reason.

burqen (approved):

burqen:
@pontusmelke I'm happy! But let's see what @MishaDemianenko and @chrisvest has to say as well.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
fickludd: Letting cursors be unaware of read makes the cursor objects very flexible. I'm personally thinking about what is the actual target life span of cursors (a transaction?), but we are not completely settled on this yet. Meanwhile it seemed like inverting the dependency was the cleanest way of creating all cursors from the factory. 

Would be happy to discuss cursor life span in more detail once more of the basics are in place!

fickludd:
@chrisvest @MishaDemianenko @burqen ~~We have noticed nodes are not locked on delete in this PR currently. This of course causes interesting failures under concurrent load. We are addressing that now, will grow the PR a little more.~~

Now they are locked.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
burqen: Thanks for the explanation! 😄

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/StubPropertyCursor.java
```diff
@@ -0,0 +1,179 @@
...
+    @Override
+    public Value propertyValue()
+    {
+        return values[offset];
+    }
+
+    @Override
+    public <E extends Exception> void writeTo( ValueWriter<E> target )
+    {
+
```
fickludd, line 81: We should throw `UnsupportedOperationException` here, otherwise it will bite us later.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
```diff
@@ -294,4 +292,14 @@ ArrayValue array( PropertyCursor cursor, long reference, PageCursor page )
         buffer.flip();
         return PropertyUtil.readArrayFromBuffer( buffer );
     }
+
+    public boolean nodeExists( long id )
```
fickludd, line 296: How about calling this `nodeExistsInStore` just to be clear that we are not checking tx-state?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NewKernel.java
```diff
@@ -37,20 +37,18 @@
...
     private volatile boolean isRunning;
 
-    public NewKernel( StorageEngine engine, Supplier<ExplicitIndexTransactionState> explicitIndexes, InwardKernel kernel )
+    public NewKernel( StorageEngine engine, InwardKernel kernel )
     {
         this.engine = engine;
-        this.explicitIndexes = explicitIndexes;
         this.kernel = kernel;
         this.isRunning = false;
+        this.cursors = new Cursors(  );
```
fickludd, line 51: We should be able to remove this construction. Let's construct on `start` and delete it on `stop`

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -34,20 +34,18 @@
 
 class NodeCursor extends NodeRecord implements org.neo4j.internal.kernel.api.NodeCursor
 {
-    private final Read read;
-    private final RecordCursor<DynamicRecord> labelCursor;
+    private Read read;
+    private RecordCursor<DynamicRecord> labelCursor;
```
fickludd, line 38: I'm thinking we should close/null these on cursor close, now that they are set on initialize.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeExplicitIndexCursor.java
```diff
@@ -28,16 +28,11 @@
 class NodeExplicitIndexCursor extends IndexCursor
         implements org.neo4j.internal.kernel.api.NodeExplicitIndexCursor, ExplicitClient
 {
-    private final Read read;
+    private Read read;
```
fickludd, line 31: Null on close.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeLabelIndexCursor.java
```diff
@@ -29,15 +29,10 @@
 class NodeLabelIndexCursor extends IndexCursor
         implements org.neo4j.internal.kernel.api.NodeLabelIndexCursor, NodeLabelClient
 {
-    private final Read read;
+    private Read read;
```
fickludd, line 32: Null on close.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeValueIndexCursor.java
```diff
@@ -29,16 +29,11 @@
 class NodeValueIndexCursor extends IndexCursor
         implements org.neo4j.internal.kernel.api.NodeValueIndexCursor, NodeValueClient
 {
-    private final Read read;
+    private Read read;
```
fickludd, line 32: Null on close.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
```diff
@@ -46,21 +46,20 @@
 public class PropertyCursor extends PropertyRecord implements org.neo4j.internal.kernel.api.PropertyCursor
 {
     private static final int MAX_BYTES_IN_SHORT_STRING_OR_SHORT_ARRAY = 32;
-    private final Read read;
+    private Read read;
```
fickludd, line 49: Null on close.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
```diff
@@ -146,7 +162,7 @@ private void labelScan( IndexProgressor.NodeLabelClient client, PrimitiveLongRes
     @Override
     public final void allNodesScan( org.neo4j.internal.kernel.api.NodeCursor cursor )
     {
-        ((NodeCursor) cursor).scan();
+        ((NodeCursor) cursor).scan(this);
```
fickludd, line 165: Checkstyle OMG !?!?!?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
```diff
@@ -246,57 +262,60 @@ public final void relationships(
...
                 query instanceof Value ? ((Value) query).asObject() : query ) );
     }
 
     @Override
     public final void nodeExplicitIndexQuery(
-            NodeExplicitIndexCursor cursor, String index, String key, Object query ) throws KernelException
+            org.neo4j.internal.kernel.api.NodeExplicitIndexCursor cursor, String index, String key, Object query ) throws KernelException
     {
+        ((NodeExplicitIndexCursor) cursor).setRead( this );
         explicitIndex( (org.neo4j.kernel.impl.newapi.NodeExplicitIndexCursor) cursor, explicitNodeIndex( index ).query(
```
fickludd, line 319: Having the same implementation and interface names is pretty annoying in places, here being one. Maybe we should think it over once more... 

Here I guess the `NodeExplicitIndexCursor` without package is the impl one, so we could remove the package on this line?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipCursor.java
```diff
@@ -24,12 +24,11 @@
 
 abstract class RelationshipCursor extends RelationshipRecord implements RelationshipDataAccessor
 {
-    final Read read;
+    Read read;
```
fickludd, line 27: Null on close. Also, now this is not initialized here, which makes for a slightly confusing situation were it would be easy to forget to initialize read. We could add init/close to this class too, or helpful asserts.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipExplicitIndexCursor.java
```diff
@@ -29,16 +29,11 @@
 class RelationshipExplicitIndexCursor extends IndexCursor
         implements org.neo4j.internal.kernel.api.RelationshipExplicitIndexCursor, ExplicitClient
 {
-    private final Read read;
+    private Read read;
```
fickludd, line 32: Null on close.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipGroupCursor.java
```diff
@@ -31,19 +31,18 @@
 class RelationshipGroupCursor extends RelationshipGroupRecord
         implements org.neo4j.internal.kernel.api.RelationshipGroupCursor
 {
-    private final Read read;
+    private Read read;
```
fickludd, line 34: Null on close.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipScanCursor.java
```diff
@@ -46,9 +41,10 @@ void scan( int label )
         next = 0;
         this.label = label;
         highMark = read.relationshipHighMark();
+        this.read = read;
```
fickludd, line 44: Null this on close too.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipTraversalCursor.java
```diff
@@ -90,29 +90,29 @@ boolean check( long source, long target, long origin )
...
      */
-    void buffered( long nodeReference, Record record )
+    void buffered( long nodeReference, Record record, Read read )
     {
         this.originNodeReference = nodeReference;
         this.buffer = Record.initialize( record );
         this.groupState = GroupState.NONE;
         this.filterState = FilterState.NONE;
         this.filterType = NO_ID;
+        this.read = read;
```
fickludd, line 109: Yaya, null it on close etc.

fickludd (approved):
This looks good overall, but found a bunch of nitpicks that could be polished.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeWriteTestBase.java
```diff
@@ -69,6 +70,53 @@ public void shouldRollbackOnFailure() throws Exception
...
+    public void shouldRemoveNode() throws Exception
+    {
+        long node;
+        try ( org.neo4j.graphdb.Transaction tx = graphDb.beginTx() )
+        {
+            node = graphDb.createNode().getId();
+            tx.success();
+        }
+
+        try ( Transaction tx = session.beginTransaction() )
```
chrisvest, line 90: Would be convenient if these had slightly different names. It shouldn't stop this PR, though.

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionTerminationTest.java
```diff
@@ -340,7 +343,8 @@ static CommitterAction random()
                     mock( Pool.class ), Clocks.fakeClock(), CpuClock.NOT_AVAILABLE, HeapAllocation.NOT_AVAILABLE,
                     TransactionTracer.NULL,
                     LockTracer.NONE, PageCursorTracerSupplier.NULL,
-                            mock( StorageEngine.class, RETURNS_MOCKS ), new CanWrite() );
+                            mock( StorageEngine.class, RETURNS_MOCKS ), new CanWrite(), mock(Cursors.class), AutoIndexing.UNSUPPORTED, mock(
+                            ExplicitIndexStore.class) );
```
chrisvest, line 347: Looks like something checkstyle might complain about.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Labels.java
```diff
@@ -45,4 +45,17 @@ public int label( int offset )
...
+    public boolean contains( int labelToken )
+    {
+        for ( long label : labels )
+        {
+            if ( label == labelToken )
+            {
+                return true;
+            }
+        }
+        return false;
```
chrisvest, line 59: A little surprising we're not using binary search here, since the labels should be sorted. I went digging through the history back to 2.2, but apparently it's always been a loop.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Transaction.java
```diff
@@ -61,7 +63,7 @@
     /**
      * @return The Write operations of the graph.
      */
-    Write dataWrite();
+    Write dataWrite() throws KernelException;
```
chrisvest, line 66: I would add it to the javadoc that the KernelException is thrown for instance in the case where the transaction cannot be upgraded to a write transaction, e.g. when it has already done schema modifications.

chrisvest (approved):
I had a few comments, but nothing that'll hold this PR back.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Labels.java
fickludd: If we gamble of nodes typically having say 1-4 labels, couldn't it be that the loop is actually faster? Although we could stop searching if we passed the label.

PR #10479, opened 2017-11-19
Faster ValueUtils.of

pontusmelke:
We are trying to avoid using `ValueUtils.of` in the runtime, however there are some situations where it is necessary to use it so having it be really slow doesn't help. The main speed up comes from removing using exceptions as flow control.

Benchmarks for:
```
{
  objects: ["1", "2", ..., "100", { list:  ["1", "2", ..., "100"] }, 
  float: 13.7, 
  int: 13
}
```

Before:
```
Of.of  50.845 ± 3.617  us/op
```
After:
```
Of.of 3.483 ± 0.536  us/op
```

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
@@ -172,7 +172,7 @@ public final class ValueUtils
 
     public static ListValue asListValue( List<?> collection )
     {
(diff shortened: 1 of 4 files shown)


chrisvest:
Another trick you can do, is to check super types first, and then drill into more concrete types. Effectively you'll get some sort of tree-search for the types. For instance, Node and Relationship are both Entities. Path and List are both Iterables, and Point implements Geometry. It also helps to check for the most likely types first. It roughly looks like this is already the case.

SaschaPeukert (approved):

PR #10480, opened 2017-11-20
Remove reference to runtime context and transaction from query cache

fickludd:
Because of an anonymous class `ExecutionPlan` implementation, we have been referencing whole runtime contexts from every plan in the query cache, including transaction objects. This PR removes this abstraction leak and should thus reduce memory footprint of the query cache.


enterprise/cypher/cypher-compiled-runtime-3.2/src/main/scala/org/neo4j/cypher/internal/compiled_runtime/v3_2/BuildCompiledExecutionPlan.scala
```diff
@@ -57,10 +60,42 @@ object BuildCompiledExecutionPlan extends Phase[CompiledRuntimeContext, Compilat
     }
   }
 
-  private def createExecutionPlan(context: CompiledRuntimeContext, compiled: CompiledPlan) = new ExecutionPlan {
-    private val fingerprint = context.createFingerprintReference(compiled.fingerprint)
+//  private def createExecutionPlan(context: CompiledRuntimeContext, compiled: CompiledPlan) = new ExecutionPlan {
```
systay: This should probably not be left lingering

systay (approved):
LGTM

PR #10482, opened 2017-11-20
 Support NestedPlanExpression in slotted runtime

henriknyman:
- Interleave slot allocation for expressions with slot allocation for
logical plans.
- Recursively allocate slots for NestedPlanExpressions and merge
the information into the physical plan.
- Make sure we use the expression converter that does
NestedPlanExpression -> NestedPipeExpression everywhere in
SlottedPipeBuilder.
- Add NestedPlanExpression to SlottedRewriter and pass in the
full slot configurations map so that expressions can be
correctly rewritten within nested plans.


enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -122,11 +126,16 @@ object SlotAllocation {
         case (Some(_), Some(right)) if comingFrom eq right =>
           val rhsSlots = resultStack.pop()
           val lhsSlots = resultStack.pop()
-          val result = allocate(current, nullable, lhsSlots, rhsSlots)
-          allocations += (current.assignedId -> result)
+          // NOTE: This assumes expressions on this logical plan are evaluated in the scope of the LHS
```
systay: Have you considered ValueHashJoin? It will evaluate expressions on both sides

systay: See: https://github.com/neo4j/neo4j/blob/3.4/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlottedRewriter.scala#L95

henriknyman: Good point!

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -635,13 +638,25 @@ object CypherComparisonSupport {
...
                                                                        Runtimes.ProcedureOrSchema)) +
         TestConfiguration(Versions.V2_3 -> Versions.V3_1, Planners.all, Runtimes.Default) +
         TestScenario(Versions.Default, Planners.Rule, Runtimes.Default)
 
+    /**
+      * These experimental configurations will only be executed if you explicitly specify them in the test expectation.
+      * I.e. there will be no check to see if they unexpectedly succeed on tests where they were not explicitly requested.
+      */
+    def Experimental: TestConfiguration =
+      TestConfiguration(Versions.Default, Planners.Default, Runtimes(Runtimes.Morsel))
```
fickludd: This is a good way to keep `Morsel` in a maintainable way.

enterprise/cypher/cypher/pom.xml
```diff
@@ -348,6 +348,11 @@
       <groupId>commons-codec</groupId>
       <artifactId>commons-codec</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.bitbucket.inkytonik.kiama</groupId>
+      <artifactId>kiama_2.11</artifactId>
+      <version>2.1.0</version>
```
fickludd, line 354: License problem?

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -55,17 +55,19 @@ object SlotAllocation {
...
+  def allocateSlots(lp: LogicalPlan,
+                    initialSlotsAndArgument: Option[SlotsAndArgument] = None): PhysicalPlan = {
 
     val allocations = new mutable.OpenHashMap[LogicalPlanId, SlotConfiguration]()
     val arguments = new mutable.OpenHashMap[LogicalPlanId, Size]()
 
     val planStack = new mutable.Stack[(Boolean, LogicalPlan)]()
     val resultStack = new mutable.Stack[SlotConfiguration]()
     val argumentStack = new mutable.Stack[SlotsAndArgument]()
+    initialSlotsAndArgument.foreach(argumentStack.push)
```
fickludd, line 67: We could use `SlotAllocation.NO_ARGUMENT()` to encode no argument and not need the `Option`. This is also ok though.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -97,17 +99,19 @@ object SlotAllocation {
...
 
         case (Some(_), None) =>
           val sourceSlots = resultStack.pop()
           val argument = if (argumentStack.isEmpty) NO_ARGUMENT()
                          else argumentStack.top
-          val result = allocate(current, nullable, sourceSlots, recordArgument(_, argument))
-          allocations += (current.assignedId -> result)
-          resultStack.push(result)
+          val result = allocateExpressions(current, nullable, sourceSlots, allocations, arguments)
+          val result2 = allocate(current, nullable, result, recordArgument(_, argument))
```
fickludd: How about calling them `nestedResult` and `result` instead?

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -135,6 +144,43 @@ object SlotAllocation {
...
+                                  argumentSizes: mutable.Map[LogicalPlanId, Size]): SlotConfiguration = {
+    val newSlots = lp.treeFold(slots) {
+      case p: LogicalPlan if p.assignedId != lp.assignedId =>
+        slots => (slots, None)
+
+      case e: ScopeExpression =>
+        slots => {
+          e.introducedVariables.foreach {
+            case v@Variable(name) =>
+              slots.newReference(name, true, CTAny)
```
fickludd: I'm still not convinced that we should allocate local variables among the streamed slots, but if you think it's the way to go then let's try.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -135,6 +144,43 @@ object SlotAllocation {
...
+          val slotsAndArgument = SlotsAndArgument(argumentSlotConfiguration, Size(slots.numberOfLongs, slots.numberOfReferences))
+
+          // Allocate slots for nested plan
+          val nestedPhysicalPlan = allocateSlots(e.plan, initialSlotsAndArgument = Some(slotsAndArgument))
+
+          // Update the physical plan
+          slotConfigurations ++= nestedPhysicalPlan.slotConfigurations
+          argumentSizes ++= nestedPhysicalPlan.argumentSizes
+
+          (slots, None)
```
fickludd: But this means that `slots` is unchanged by nested plans, or am I misreading?

enterprise/cypher/physical-planning/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocationTest.scala
```diff
@@ -736,4 +736,26 @@ class SlotAllocationTest extends CypherFunSuite with LogicalPlanningTestSupport2
...
+    // then
+    allocations should have size 3
+    allocations(plan.assignedId) should equal(
+      SlotConfiguration(Map("z" -> RefSlot(0, nullable = true, CTAny)), 0, 1)
+    )
+    allocations(argument.assignedId) should equal(allocations(plan.assignedId))
+    allocations(nestedPlan.assignedId) should equal(
+      SlotConfiguration(Map("x" -> LongSlot(0, nullable = false, CTNode)), 1, 0)
+    )
+  }
```
fickludd, line 760: Nested plans sound complicated to me... could we test something more to specify the slot configuration of 
 * upstream operators
 * the passing of the parent operator scope to the nested plan
 * the passing of the parent argument to the nested plan

fickludd (changes requested):
I know you are still working on this, but before we merge this I would like some more tests around the nested plan allocation to specify the behavior in some more complicated cases. 

The implementation looks good to me.

enterprise/cypher/physical-planning/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocationTest.scala
henriknyman: There are tests for this in the acceptance spec suite, like `CaseExpression.feature`

henriknyman:
@fickludd We discussed introducing a runtime acceptance level of testing, where we give logical plans, build and execute, and assert on the result. I think it will be suitable for the additional test coverage you wanted to see.

PR #10489, opened 2017-11-21
Added configurable cypher query replan threshold decay algorithms

OliviaYtterbrink:
Current neo4j.conf defaults are optimized for new databases and initial import scenarios, designed to prevent too frequent query replanning. However, for larger databases, these setting will result in query planning never happening unless you edit indexes (create or drop), or manually run the new query plan cache clearing procedure (only available in 3.4). This is because the statistics threshold of 0.75 actually implies that the database needs to quadruple in size in order to replan.

We have seen this issue in particular on clusters where instances that join the cluster after initial import will plan once, and then never again, even if the database doubles or triples in size, dramatically changing the statistics, and quite possibly needing a replan.

This PR adds a decay function to the threshold, such that the older the plan is in the case, the smaller the statistics change required to trigger a replan. The new defaults are:

* `cypher.min_replan_interval=10s` (same as before, never check the stats more frequently than once every 10s)
* `cypher.statistics_divergence_threshold=0.75` (same as before, do not replan unless relevant graph statistics change by more than 75%, which for increasing database sizes means quadruple in size)
* `unsupported.cypher.statistics_divergence_target=0.1` (slowly reduce the threshold down to 10% so that queries ultimately get replanned even with small database changes)
* `unsupported.cypher.target_replan_interval=7h` (take 7h to reduce the threshold down to 10%)
* `unsupported.cypher.replan_algorithm=inverse` (use an inverse function to model threshold decay)

changelog: Fixes issue with medium and large databases that might change relatively slowly from never having their Cypher queries replanned. This has been achieved by having the database statistics divergence threshold set by `cypher.statistics_divergence_threshold` shrink over time, by default from 75% to 10% over 7 hours. The decay is modelled using an inverse function that tends toward zero over many days, and so ultimately all queries will get replanned if even small changes occur.

### DIFF
--- a/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/CacheAccessor.scala
+++ b/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/CacheAccessor.scala
@@ -19,13 +19,15 @@
  */
 package org.neo4j.cypher.internal.compiler.v3_2
 
+import org.neo4j.cypher.internal.compiler.v3_2.executionplan.CacheCheckResult
+
 trait CacheAccessor[K <: AnyRef, T <: AnyRef] {
   def getOrElseUpdate(cache: LFUCache[K, T])(key: K, f: => T): T
   def remove(cache: LFUCache[K, T])(key: K, userKey: String, secondsSinceReplan: Int)
 }
 
 class QueryCache[K <: AnyRef, T <: AnyRef](cacheAccessor: CacheAccessor[K, T], cache: LFUCache[K, T]) {
-  def getOrElseUpdate(key: K, userKey: String, isStale: T => (Boolean, Int), produce: => T): (T, Boolean) = {
+  def getOrElseUpdate(key: K, userKey: String, isStale: T => CacheCheckResult, produce: => T): (T, Boolean) = {
     if (cache.size == 0)
       (produce, false)
     else {
@@ -37,9 +39,9 @@ class QueryCache[K <: AnyRef, T <: AnyRef](cacheAccessor: CacheAccessor[K, T], c
         })
       }.flatMap { value =>
         if (!planned) {
-          val (stale, age) = isStale(value)
-          if (stale) {
-            cacheAccessor.remove(cache)(key, userKey, age)
+          val cacheCheck = isStale(value)
+          if (cacheCheck.isStale) {
+            cacheAccessor.remove(cache)(key, userKey, cacheCheck.secondsSinceReplan)
             None
           } else {
             Some((value, planned))
--- a/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/CacheAccessor.scala
+++ b/community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/CacheAccessor.scala
@@ -21,11 +21,11 @@ package org.neo4j.cypher.internal.compiler.v3_2
 
 trait CacheAccessor[K <: AnyRef, T <: AnyRef] {
   def getOrElseUpdate(cache: LFUCache[K, T])(key: K, f: => T): T
-  def remove(cache: LFUCache[K, T])(key: K, userKey: String)
+  def remove(cache: LFUCache[K, T])(key: K, userKey: String, secondsSinceReplan: Int)
 }
 
 class QueryCache[K <: AnyRef, T <: AnyRef](cacheAccessor: CacheAccessor[K, T], cache: LFUCache[K, T]) {
-  def getOrElseUpdate(key: K, userKey: String, isStale: T => Boolean, produce: => T): (T, Boolean) = {
+  def getOrElseUpdate(key: K, userKey: String, isStale: T => (Boolean, Int), produce: => T): (T, Boolean) = {
     if (cache.size == 0)
       (produce, false)
     else {
@@ -36,9 +36,14 @@ class QueryCache[K <: AnyRef, T <: AnyRef](cacheAccessor: CacheAccessor[K, T], c
           produce
         })
       }.flatMap { value =>
-        if (!planned && isStale(value)) {
-          cacheAccessor.remove(cache)(key, userKey)
-          None
+        if (!planned) {
+          val stale = isStale(value)
+          if (stale._1) {
+            cacheAccessor.remove(cache)(key, userKey, stale._2)
+            None
+          } else {
+            Some((value, planned))
+          }
         }
         else {
           Some((value, planned))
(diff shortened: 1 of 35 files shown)


craigtaverner:
One idea I had was to simplify the three new config settings into one.

This means changing:

* `unsupported.cypher.statistics_divergence_target=0.1` (slowly reduce the threshold down to 10% so that queries ultimately get replanned even with small database changes)
* `unsupported.cypher.target_replan_interval=7h` (take 7h to reduce the threshold down to 10%)
* `unsupported.cypher.replan_algorithm=inverse` (use an inverse function to model threshold decay)

Into:

* `unsupported.cypher.replan_algorithm=inverse(0.1,7h)` (take 7h to reduce the threshold down to 10% using 'inverse' function)

The advantage is that it is only one setting, the disadvantage is that the neo4j.conf settings system cannot be used to pre-filter valid options. We currently already have numerically invalid options (like threshold increase instead of decrease) result in no decay (revert to old behaviour), but it feels like we will have even less opportunity to have neo4j.conf parsing checks with this simpler single config.

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/executionplan/ExecutionPlan.scala
```diff
@@ -28,7 +28,7 @@ abstract class ExecutionPlan {
   def run(queryContext: QueryContext, planType: ExecutionMode, params: Map[String, Any]): InternalExecutionResult
   def isPeriodicCommit: Boolean
   def plannerUsed: PlannerName
-  def isStale(lastTxId: () => Long, statistics: GraphStatistics): Boolean
+  def isStale(lastTxId: () => Long, statistics: GraphStatistics): (Boolean, Int)
```
pontusmelke: To me it is not obvious what this function returns now, the extra `int` is a bit confusing. Could we rewrite to something like:

```
def checkCache(lastTxId: () => Long, statistics: GraphStatistics): CacheData
```
where

```
case class CacheData(isStale:Boolean, ageInSeconds: Int)
```
(But better names please)

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/CacheAccessor.scala
```diff
@@ -36,9 +36,14 @@ class QueryCache[K <: AnyRef, T <: AnyRef](cacheAccessor: CacheAccessor[K, T], c
           produce
         })
       }.flatMap { value =>
-        if (!planned && isStale(value)) {
-          cacheAccessor.remove(cache)(key, userKey)
-          None
+        if (!planned) {
+          val stale = isStale(value)
```
pontusmelke: `val (isStale, age) = isStale(value)`

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/executionplan/PlanFingerprint.scala
```diff
@@ -23,23 +23,91 @@ import java.time.Clock
...
+    initialThreshold / (1.0 + decayFactor * (millisSincePreviousReplan - initialMillis))
+  }
+}
+
+case class StatsDivergenceExponentialDecayCalculator(initialThreshold: Double, targetThreshold: Double, initialMillis: Long, targetMillis: Long) extends StatsDivergenceCalculator {
+  val decayFactor: Double = (Math.log(initialThreshold) - Math.log(targetThreshold)) / (targetMillis - initialMillis)
+
+  def decay(millisSincePreviousReplan: Long): Double = {
+    val exponent = -1.0 * decayFactor * (millisSincePreviousReplan - initialMillis)
+    initialThreshold * Math.pow(Math.E, exponent)
```
pontusmelke: ~`Math.pow(Math.E, exponent)`~ -> `Math.exp(exponent)`

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/spi/InstrumentedGraphStatistics.scala
```diff
@@ -38,7 +39,7 @@ class MutableGraphStatisticsSnapshot(val map: mutable.Map[StatisticsKey, Double]
   def freeze: GraphStatisticsSnapshot = GraphStatisticsSnapshot(map.toMap)
 }
 
-case class GraphStatisticsSnapshot(statsValues: Map[StatisticsKey, Double] = Map.empty) {
+case class GraphStatisticsSnapshot(statsValues: Map[StatisticsKey, Double] = Map.empty, timestamp: Date = new Date) {
```
pontusmelke: I can't see where we use the timestamp?

community/cypher/cypher-compiler-3.2/src/main/scala/org/neo4j/cypher/internal/compiler/v3_2/executionplan/PlanFingerprint.scala
```diff
@@ -23,23 +23,91 @@ import java.time.Clock
...
+trait StatsDivergenceCalculator {
+  val initialThreshold: Double
+  val initialMillis: Long
+
+  def shouldCheck(currentTimeMillis: Long, lastCheckTimeMillis: Long): Boolean = currentTimeMillis - initialMillis >= lastCheckTimeMillis
+
+  def decay(millisSincePreviousReplan: Long): Double
+}
+
+case class StatsDivergenceInverseDecayCalculator(initialThreshold: Double, targetThreshold: Double, initialMillis: Long, targetMillis: Long) extends StatsDivergenceCalculator {
```
pontusmelke, line 66: Do we really need two different calculators? How would we tell an end-user when to use one or the other?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -191,14 +191,27 @@
...
     public static Setting<Integer> query_cache_size = setting( "dbms.query_cache_size", INTEGER, "1000", min( 0 ) );
 
-    @Description( "The threshold when a plan is considered stale. If any of the underlying" +
-                  " statistics used to create the plan has changed more than this value, " +
-                  "the plan is considered stale and will be replanned. " +
-                  "A value of 0 means always replan, and 1 means never replan." )
+    @Description( "The threshold when a plan is considered stale. If any of the underlying " +
+                  "statistics used to create the plan have changed more than this value, " +
+                  "the plan will be considered stale and will be replanned. Change is calculated as " +
+                  "abs(a-b)/max(a,b). This means that a value of 0.75 requires the database to approximately " +
```
pontusmelke, line 197: If you want this here I think you need to explain what `a` and `b` are (and they are not always cardinalities, could also be a selectivity)

pontusmelke (changes requested):
Looks good, just a couple of minor things.

pontusmelke (approved):

PR #10490, opened 2017-11-22
Bump procedure compiler test utils

MishaDemianenko:
- Bump version of compile-testing to be able to compile with java 9.
- In DuplicationAwareBaseProcessor change source version to be
     SourceVersion.latestSupported() instead of SourceVersion.RELEASE_8
- Update BadRecord age type to be long instead of int.
  Because of https://bugs.openjdk.java.net/browse/JDK-8066843 incorrect
  type of age field in BadRecord was not notices before.


fickludd (approved):
Looks good. I bet a rebase would clear the docs problem.

PR #10491, opened 2017-11-22
Update QueryLoggerIT test expectations

MishaDemianenko:
Since query logging is async process we can't rely on how reporting of
executed queries is actually done so we need at least to shutdown db
to be sure that messages actually logged.

### DIFF
--- a/enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerIT.java
+++ b/enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerIT.java
@@ -70,6 +70,7 @@ import static org.hamcrest.Matchers.endsWith;
 import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.lessThan;
+import static org.hamcrest.Matchers.lessThanOrEqualTo;
 import static org.hamcrest.Matchers.not;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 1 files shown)


enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerIT.java
```diff
@@ -350,7 +354,7 @@ public void queryLogRotation() throws Exception
                               .map( this::readAllLinesSilent )
                               .flatMap( Collection::stream )
                               .collect( Collectors.toList() );
-        assertThat( "Expected log file to have at least one log entry", loggedQueries.size(), lessThan( 200 ) );
+        assertThat( "Expected log file to have at least one log entry", loggedQueries.size(), lessThanOrEqualTo( 202 ) );
```
burqen, line 357: Why do we check `lessThan`? Is it not more interesting to check `greaterThan` if we want to make sure that some number of queries are logged?
Failure message could use an update as well.

burqen (changes requested):
Changes looks good! Just have a clarifying question.

enterprise/query-logging/src/test/java/org/neo4j/kernel/impl/query/QueryLoggerIT.java
MishaDemianenko: logging and rotation is async and you do not know have many of your queries where logs in which file and you can't rely on how much queries will be in a file that new rotation policy: "CALL dbms.setConfigValue('" + log_queries_max_archives.name() + "','1')" will keep.

chrisvest: Because this test makes an online configuration change that will reduce the number of query log files that we keep. This change must take effect for the log files that exists at that time, so we must end up with _fewer_ log files than we otherwise would.

burqen: Aah, yes! Thanks :D

burqen (approved):

PR #10495, opened 2017-11-23
Reads tx aware and more writing

pontusmelke:
The main part of this PR is about making node reading transaction state aware. This means that that both `NodeCursor` and `PropertyCursor` are now reading not only from disk but also from the transaction state. Apart from that this PR also adds:

- property writes
- now support the all the explicit index operations
- change the core api to use the new api for reads.

In order to use the new API also for writing labels and properties we are now only missing constraint validation, which we'll handle in an upcoming chunk of work. We also choose not enable node reads in cypher, in order to somewhat reduce the scope of the PR.

~This PR builds on top of https://github.com/neo4j/neo4j/pull/10478, it will be marked as NOT_READY_TO_MERGE until that PR has been merged.~


community/kernel/src/main/java/org/neo4j/storageengine/api/txstate/PropertyContainerState.java
```diff
@@ -57,10 +57,14 @@ void visitPropertyChanges( long entityId, Iterator<StorageProperty> added,
 
     boolean hasChanges();
 
+    boolean hasPropertyChanges();
```
burqen: In my mind, `hasChanges` is the same thing as `hasPropertyChanges` because of the fact that it is part of `PropertyContainerState` which only deals with properties. The need for the new `hasPropertyChanges` arises because of `hasChanges` is used as a general "this instance of whatever has seen some changes of some sort". 

Instead of _adding_ `hasPropertyChanges` I think a better way would be to clean this up by 
- Renaming `hasChanges` to `hasPropertyChanges`
- Add specific `hasRelationshipChanges` to `NodeState`.

community/kernel/src/main/java/org/neo4j/storageengine/api/txstate/ReadableTransactionState.java
```diff
@@ -142,6 +142,8 @@
 
     RelationshipState getRelationshipState( long id );
 
+    PropertyContainerState getPropertiesState( long reference );
```
burqen: Is `reference` a node reference, relationship reference or something else? The rest of this interface seem to lean agains `nodeId`, `relId` or `relationshipId`, terminology wise. Or is it a reference to a general property container. How does that even work if we have collision on relationship and node ids?

Lot's of questions. Might be an opportunity to clarify code here.

Edit: ... and now that I read test code it seems like reference is referring to a property id?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
```diff
@@ -134,4 +140,79 @@ static boolean hasDirectFlag( long groupReference )
...
+
+    /**
+     * Encode a node id as a property reference.
+     *
+     * This allows us to encode a node reference as a property reference so that we can do lookups of changes in the
+     * transaction state.
+     * @param reference the reference to encode
+     * @return The encoded reference
+     */
+    static long setNodeFlag( long reference )
```
burqen: Nice javadoc for those new methods! I would prefer if we could be explicit about parameter names and calling them `nodeReference`, `relationshipReference` and `propertyReference` for clarity. This is also the terminology that rest of interface uses.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
```diff
@@ -53,31 +57,81 @@
     private PageCursor page;
     private PageCursor stringPage;
     private PageCursor arrayPage;
+    private PropertyContainerState propertiesState;
+    private Iterator<StorageProperty> changedProperties;
+    private StorageProperty stateValue;
```
burqen: Is `stateValue` value read from tx state or some state of this cursor? If it is value from tx state, can we please call it `txStateValue`?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeSchemaMatcher.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param <EXCEPTION> The type of exception that can be thrown when taking the action
+     * @param schemaSuppliers The suppliers to match
+     * @param node The node cursor
+     * @param property The property cursor
+     * @param specialPropertyId This property id will always count as a match for the descriptor, regardless of
+     * whether the node has this property or not
+     * @param callback The action to take on match
+     * @throws EXCEPTION This exception is propagated from the action
+     */
+    <SUPPLIER extends LabelSchemaSupplier, EXCEPTION extends Exception> void onMatchingSchema(
```
burqen, line 61: Nice javadoc on this guy! Thank you 😄

community/kernel/src/main/java/org/neo4j/kernel/api/DataWriteOperations.java
```diff
@@ -21,16 +21,17 @@
...
+import org.neo4j.internal.kernel.api.exceptions.explicitindex.ExplicitIndexNotFoundKernelException;
+import org.neo4j.kernel.api.exceptions.EntityNotFoundException;
 import org.neo4j.kernel.api.exceptions.RelationshipTypeIdNotFoundKernelException;
-import org.neo4j.kernel.api.exceptions.explicitindex.AutoIndexingKernelException;
-import org.neo4j.kernel.api.exceptions.explicitindex.ExplicitIndexNotFoundKernelException;
 import org.neo4j.kernel.api.exceptions.schema.ConstraintValidationException;
 import org.neo4j.values.storable.Value;
 
-public interface DataWriteOperations
+public interface DataWriteOperations extends ExplicitIndexWrite
```
burqen, line 34: `DataWriteOperations` define all methods from `ExplicitIndexWrite` again. Can we remove this duplication an let inherency do it's job 😉 

Edit: Or wait. I see now that `DataWriteOperations` is more detailed about what exceptions it can throw in some cases. What is the reason that they do not align so that we can remove de duplication?

Edit2: So from what I understand, the binary dependency from Cypher is the reason why we can't move the exception and clean this up. What are the plans moving forward from here?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/ExplicitIndexCursorWritesTestBase.java
```diff
@@ -19,25 +19,94 @@
...
+            nodeId = tx.dataWrite().nodeCreate();
+            ExplicitIndexWrite indexWrite = tx.indexWrite();
+            indexWrite.nodeAddToExplicitIndex( INDEX_NAME, nodeId, KEY, VALUE );
+            tx.success();
+        }
+
+        // Then
+        try ( org.neo4j.graphdb.Transaction ctx = graphDb.beginTx() )
+        {
+            IndexHits<Node> hits = graphDb.index().forNodes( INDEX_NAME ).get( KEY, "this is it" );
```
burqen: Magic string "this is it"

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/ExplicitIndexCursorWritesTestBase.java
```diff
@@ -19,25 +19,94 @@
...
 
+@SuppressWarnings( "Duplicates" )
 public abstract class ExplicitIndexCursorWritesTestBase<G extends KernelAPIWriteTestSupport>
         extends KernelAPIWriteTestBase<G>
 {
 
     private static final String INDEX_NAME = "foo";
     private static final String KEY = "bar";
     private static final String VALUE = "this is it";
+    @Test
```
burqen: Missing new line

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/ExplicitIndexCursorWritesTestBase.java
```diff
@@ -174,16 +263,144 @@ public void shouldCreateExplicitIndexTwice() throws Exception
...
+        {
+            ExplicitIndexWrite indexWrite = tx.indexWrite();
+            indexWrite.relationshipRemoveFromExplicitIndex( INDEX_NAME, relId );
+            tx.success();
+        }
+
+        // Then
+        try ( org.neo4j.graphdb.Transaction ctx = graphDb.beginTx() )
+        {
+            IndexHits<Node> hits = graphDb.index().forNodes( INDEX_NAME ).get( KEY, "this is it" );
```
burqen: Magic string "this is it"

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -22,24 +22,131 @@
...
+    /**
+     * Removes a node from an explicit index
+     *
+     * @param indexName The name of the index
+     * @param node The id of the node to add
+     * @param key The key to associate with the node
+     * @param value The value to associate with the node and key
+     * @throws ExplicitIndexNotFoundKernelException If there is no explicit index with the given name
+     */
+    void nodeRemoveFromExplicitIndex( String indexName, long node, String key, Object value )
```
burqen: Descriptions of parameters look like copy paste from add.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -22,24 +22,131 @@
...
+
+    /**
+     * Removes a node from an explicit index
+     *
+     * @param indexName The name of the index
+     * @param node The id of the node to add
+     * @param key The key to associate with the node
+     * @throws ExplicitIndexNotFoundKernelException If there is no explicit index with the given name
+     */
+    void nodeRemoveFromExplicitIndex( String indexName, long node, String key ) throws
```
burqen: Descriptions of parameters look like copy paste from add.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -22,24 +22,131 @@
...
+    /**
+     * Adds relationship to explicit index.
+     *
+     * @param indexName The name of the index
+     * @param relationship The id of the relationship to add
+     * @param key The key to associate with the node
+     * @param value The value to associate with the relationship and key
+     * @throws ExplicitIndexNotFoundKernelException If there is no explicit index with the given name
+     */
+    void relationshipRemoveFromExplicitIndex( String indexName, long relationship, String key, Object value )
```
burqen: Descriptions of parameters look like copy paste from add.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -22,24 +22,131 @@
...
+    /**
+     * Adds relationship to explicit index.
+     *
+     * @param indexName The name of the index
+     * @param relationship The id of the relationship to add
+     * @param key The key to associate with the node
+     * @throws ExplicitIndexNotFoundKernelException If there is no explicit index with the given name
      */
-    void nodeRemoveFromExplicitIndex( String indexName, long node ) throws KernelException;
+    void relationshipRemoveFromExplicitIndex( String indexName, long relationship, String key )
```
burqen: Descriptions of parameters look like copy paste from add.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -22,24 +22,131 @@
...
+            throws KernelException;
+
+    /**
+     * Adds relationship to explicit index.
+     *
+     * @param indexName The name of the index
+     * @param relationship The id of the relationship to add
+     * @throws ExplicitIndexNotFoundKernelException If there is no explicit index with the given name
+     */
+    void relationshipRemoveFromExplicitIndex( String indexName, long relationship )
```
burqen: Descriptions of parameters look like copy paste from add.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -192,8 +197,14 @@ public KernelTransactionImplementation( StatementOperationParts statementOperati
         this.accessCapability = accessCapability;
         this.statistics = new Statistics( this, cpuClock, heapAllocation );
         this.userMetaData = new HashMap<>();
+        AllStoreHolder allStoreHolder =
+                new AllStoreHolder( storageEngine, storageStatement, this, cursors, explicitIndexStore,
+                        this::assertOpen );
```
burqen, line 202: Use `KernelStatement` to also check that reference count and not only termination reason. It already implements `AssertOpen` interface.

burqen (changes requested):
My review is done. Nice stuff!

Still found a few points that could be improved.

And a few things that I missed due them not being there or me looking hard enough:
- Tests for explicit index updates due to property writes?
- Tests for schema index updates due to property writes?

community/kernel/src/main/java/org/neo4j/kernel/api/DataWriteOperations.java
fickludd: I'm not completely up to date with the problem, but if we have direct dependencies on kernel exceptions in a legacy cypher version that we support, we must patch that neo4j to remove the dependency before we can move it in 3.4. We will act on this eventually.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
fickludd: In this case it's sort of confusing, because we in essence take a nodeReference and return a propertyReference. This clashes a bit with the flag concept. Suggestions are welcome!

community/kernel/src/main/java/org/neo4j/storageengine/api/txstate/ReadableTransactionState.java
fickludd: Yes, we dithered around a bit here, but in the end I think we only use it for `propertyId`, so that should be the name.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
pontusmelke: I would never do such a thing...

community/kernel/src/main/java/org/neo4j/kernel/api/DataWriteOperations.java
pontusmelke: The plan is for DataWriteOperations to die once we are done porting. The `extends ExplicitIndexWrite` so that we don't have to have two versions of every method in `AutoIndexOperations` - on for `ExplicitIndexWrite` and one for `DataWriteOperations`.  We cannot remove `DataWriteOperations` completely until we are done porting at least all the relationship stuff as well, but I promise we'll get there eventually.

Yes you are right they are not identical since we have an unfortunate dependency from cypher that makes it hard to move `EntitiyNotFoundException`, in order to move it we need to first fix it in 2.3, 3.1, and 3.2 and then wait for them to be released so we can update the binary dependency.

Ah you mean remove from interface, yes I'll do that for the methods that are exactly the same. We still haven't ported everything though - still no drop and configuration methods in new API.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
pontusmelke: Yes I did that but it has the unfortunate side effect that we need to acquire a kernel statement for all operations that does property reads since otherwise the reference count will be zero and we will think that the transaction has been closed.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
pontusmelke: it is the value read from tx state and I agree the name is bad.

community/kernel/src/main/java/org/neo4j/storageengine/api/txstate/PropertyContainerState.java
pontusmelke: Do you mean there should be no `hasChanges` at all? And all callers should do `hasPropertyChanges() || hasRelationshipChanges()`? How would we do that without first checking type? Most places in the code doesn't know if it is a NodeState, RelationshipState, or GraphState or whatever.

community/kernel/src/main/java/org/neo4j/storageengine/api/txstate/ReadableTransactionState.java
pontusmelke: It it not a property id it is a property reference right?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/TransactionStateTestBase.java
```diff
@@ -134,18 +138,328 @@ public void shouldSeeLabelChangesInTransaction() throws Exception
...
+
+    @Test
+    public void shouldDiscoverDeletedNodeInTransaction() throws Exception
+    {
+        long nodeId;
+        try ( Transaction tx = session.beginTransaction() )
+        {
+            nodeId = tx.dataWrite().nodeCreate();
+            tx.success();
+        }
```
fickludd, line 157: At some point we should create some helpers for creating a node through the backdoor in write tests. This is done everywhere. We can do it later though.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/TransactionStateTestBase.java
```diff
@@ -134,18 +138,328 @@ public void shouldSeeLabelChangesInTransaction() throws Exception
...
+
+        // When/Then
+        try ( Transaction tx = session.beginTransaction() )
+        {
+
```
fickludd: whiteline

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/TransactionStateTestBase.java
```diff
@@ -134,18 +138,328 @@ public void shouldSeeLabelChangesInTransaction() throws Exception
...
+            nodeId = tx.dataWrite().nodeCreate();
+            propToken1 = session.token().propertyKeyGetOrCreateForName( propKey1 );
+            assertEquals( tx.dataWrite().nodeSetProperty( nodeId, propToken1, stringValue( "hello" ) ), NO_VALUE );
+            tx.success();
+        }
+
+        // When/Then
+        try ( Transaction tx = session.beginTransaction() )
+        {
+
```
fickludd: whiteline (copy pasty eh?)

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -450,6 +462,15 @@ private void assertTransactionOpen()
         }
     }
 
+    private void assertOpen()
+    {
+        Optional<Status> terminationReason = getReasonIfTerminated();
+        if ( terminationReason.isPresent() )
+        {
+            throw new TransactionTerminatedException( terminationReason.get() );
+        }
```
fickludd, line 471: Should we also check `KernelTransaction::isOpen`?

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -341,40 +348,66 @@ public Object getProperty( String key, Object defaultValue )
...
+            CursorFactory cursors = transaction.cursors();
+            try ( NodeCursor nodes = cursors.allocateNodeCursor();
+                  PropertyCursor properties = cursors.allocatePropertyCursor()
+            )
+            {
+                int propertyKey = transaction.token().propertyKey( key );
+                if ( propertyKey == KeyReadOperations.NO_SUCH_PROPERTY_KEY )
+                {
+                    return defaultValue;
+                }
```
fickludd: This check could be done before allocating cursors.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -387,6 +420,12 @@ public Object getProperty( String key, Object defaultValue )
             return Collections.emptyMap();
         }
 
+        KernelTransaction transaction = actions.kernelTransaction();
+        CursorFactory cursors = transaction.cursors();
+        int itemsToReturn = keys.length;
+        Map<String,Object> properties = new HashMap<>( itemsToReturn );
+        Token token = transaction.token();
```
fickludd: This is only half an implementation?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -128,7 +158,45 @@ public long allRelationshipsReference()
...
+                }
+                else
+                {
+                    //Mark the reference so that property cursor checks both
+                    //tx state as well as disk.
+                    ref = References.setTxStateFlag( propertiesReference );
+                }
+                //stores the node state mapped to the current property
+                //reference so that property cursor is able to retrieve the state later.
+                txState.registerProperties( ref, nodeState );
```
fickludd: We only need to register the flagged property reference. For the case of a node with no properties in store, we use the regular node lookup to get the property state.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
```diff
@@ -53,31 +57,81 @@
     private PageCursor page;
     private PageCursor stringPage;
     private PageCursor arrayPage;
+    private PropertyContainerState propertiesState;
+    private Iterator<StorageProperty> changedProperties;
```
fickludd: `txStateChangedProperties` for consistency?

community/kernel/src/test/java/org/neo4j/kernel/impl/api/KernelTransactionTerminationTest.java
```diff
@@ -343,8 +343,9 @@ static CommitterAction random()
                     mock( Pool.class ), Clocks.fakeClock(), CpuClock.NOT_AVAILABLE, HeapAllocation.NOT_AVAILABLE,
                     TransactionTracer.NULL,
                     LockTracer.NONE, PageCursorTracerSupplier.NULL,
-                            mock( StorageEngine.class, RETURNS_MOCKS ), new CanWrite(), mock(Cursors.class), AutoIndexing.UNSUPPORTED, mock(
-                            ExplicitIndexStore.class) );
+                    mock( StorageEngine.class, RETURNS_MOCKS ), new CanWrite(), mock( Cursors.class ),
+                    AutoIndexing.UNSUPPORTED, mock(
+                            ExplicitIndexStore.class ) );
```
fickludd: Creative line-break!

fickludd (changes requested):
I've been involved in some part of this PR, but by no means all. Great work! I like the test coverage, which I guess comes from existing test in many cases. 

Have some things which would be good to fix before we forget, like the property reference registry. Others are not critical.

community/kernel/src/main/java/org/neo4j/storageengine/api/txstate/ReadableTransactionState.java
fickludd: Oh yes, you're right. That could become very confusing.

community/kernel/src/main/java/org/neo4j/kernel/api/DataWriteOperations.java
burqen: Sound like you are on top of this! I'm happy 🎉

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
burqen: Fair enough. Can we work around that in some other way? Maybe by including a check on `closed` and `closing`?

Some context: 
The reason we need to do this is to protect agains reading mixed state dynamic property records as a result of property id reuse and pull updates from master. Test for this is `TerminationOfSlavesDuringPullUpdatesIT`. 

Knowing that assertOpen in `AllStoreHolder` is only used for the specific case of verifying transaction not being terminated on value property reads, checking for termination is probably good enough.

However, the knowledge dependency for seeing that this is "probably good enough" is far to complex and I would still prefer if we could verify that the transaction we are acting from is actually open and not just "not terminated".

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
burqen: I'm sorry doc, but the problem is larger than expected. It wasn't only the head.The entire parameter section is infected. We might need surgery, fast!

```
* @param key The key to associate with the node
* @param value The value to associate with the node and key
```

Same for the rest below ;)

community/kernel/src/main/java/org/neo4j/storageengine/api/txstate/PropertyContainerState.java
burqen: Yes, exactly. There should be no `hasChanges` at all. From what I can see there is only five usages and they should all be able to determine if they are interested in property or relationship changes. Or both for that matter. `NodeState` will still inherit from `PropertyContainerState`.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
burqen: I'm sorry doc, but the problem is larger than expected. It wasn't only the head.The entire parameter section is infected. We might need surgery, fast!

* @param key The key to associate with the node
* @param value The value to associate with the node and key
Same for the rest below ;)

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
burqen: Well, the reference coming in is still gona be a reference to a node, etc. Then we encode some more information into that for later use by us and when we read the reference we kind of say "yeah I know this is a node reference but I'm being promised it's safe for me to use it as a property reference." So we are not turning it into something that it isn't just make promises about additional possibilities.
Is that different perspective
a) Accurate at all?
b) Helpful?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
pontusmelke: This is what we test in the old API in KernelStatement. I know for sure that it is not enough to test isOpen. Don't know if we should test both.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
pontusmelke: hmm something has gone weird in some rebase or something. It doesn't look like this in earlier commits

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
pontusmelke: ~I don't understand?~ I do understand

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
pontusmelke: I would never do such a thing...

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
pontusmelke: I am afraid I don't understand the `assertOpen` does the same thing as before except for checking the reference counting thing. It checks the `terminationReason` and throws otherwise.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
pontusmelke: updated names of parameters

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
pontusmelke: On top of that it is also a bit confusing since `KernelTransaction::isOpen` returns `true` when the transaction is in state `closing` which is when tx-hooks are run and tx-hooks must be able to run commands so if we cannot add that to check without further refactoring.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
pontusmelke: If you mean that we should add check for `closing` and `closed` we cannot - at least not for the former since currently tx hooks are run after `closing` is set to true. I would love to clean this up but I feel that this PR is not the place. As it is now we have the same behavior as before.

burqen: Alright, I'll take that 👍

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
burqen: This `reference` parameter not updated yet.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
```diff
@@ -134,4 +140,79 @@ static boolean hasDirectFlag( long groupReference )
...
+        return relationshipReference | RELATIONSHIP_MARKER | FLAG_MARKER;
+    }
+
+    /**
+     * Checks if the property reference really is a relationship reference
+     *
+     * @param reference The reference to check
+     * @return <tt>true</tt>if the reference refers to a relationship otherwise <tt>false</tt>
+     */
+    static boolean hasRelationshipFlag( long reference )
```
burqen, line 188: This `reference` parameter not updated yet.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
```diff
@@ -134,4 +140,79 @@ static boolean hasDirectFlag( long groupReference )
...
+    /**
+     * Marks the property reference as having changes in the transaction state.
+     * <p>
+     * Setting this flag tells us that the value on disk is not the only source of truth, the transaction state must
+     * also be checked.
+     *
+     * @param reference The reference to set as having tx state
+     * @return The encoded reference.
+     */
+    static long setTxStateFlag( long reference )
```
burqen, line 203: This `reference` parameter not updated yet.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
```diff
@@ -134,4 +140,79 @@ static boolean hasDirectFlag( long groupReference )
...
+    {
+        return reference | STATE_MARKER | FLAG_MARKER;
+    }
+
+    /**
+     * Checks if the property reference has been marked as having changes in the transaction state.
+     * @param reference The reference to check
+     * @return <tt>true</tt>if the reference has been marked as having changes otherwise <tt>false</tt>
+     */
+    static boolean hasTxStateFlag( long reference )
```
burqen, line 213: This `reference` parameter not updated yet.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -22,24 +22,131 @@
...
+    void nodeAddToExplicitIndex( String indexName, long node, String key, Object value )
+            throws KernelException;
+
+    /**
+     * Removes a node from an explicit index
+     *
+     * @param indexName The name of the index
+     * @param node The id of the node to remove
+     * @param key The key to associate with the node
+     * @param value The value to associate with the node and key
```
burqen: I'm sorry. I should have been more specific from the start. Third time's the charm.

@param key The key ~to associate~ associated with the node
@param value The value ~to associate~ associated with the node and key

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -22,24 +22,131 @@
...
+     */
+    void nodeRemoveFromExplicitIndex( String indexName, long node, String key, Object value )
+            throws ExplicitIndexNotFoundKernelException;
+
+    /**
+     * Removes a node from an explicit index
+     *
+     * @param indexName The name of the index
+     * @param node The id of the node to remove
+     * @param key The key to associate with the node
```
burqen: @param key The key ~to associate~ associated with the node

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -22,24 +22,131 @@
...
+    void relationshipAddToExplicitIndex( String indexName, long relationship, String key, Object value )
+            throws KernelException;
+
+    /**
+     * Removes relationship from explicit index.
+     *
+     * @param indexName The name of the index
+     * @param relationship The id of the relationship to remove
+     * @param key The key to associate with the relationship
+     * @param value The value to associate with the relationship and key
```
burqen: @Param key The key ~to associate~ associated with the relationship
@Param value The value ~to associate~ associated with the relationship and key

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/ExplicitIndexWrite.java
```diff
@@ -22,24 +22,131 @@
...
+     */
+    void relationshipRemoveFromExplicitIndex( String indexName, long relationship, String key, Object value )
+            throws KernelException;
+
+    /**
+     * Removes relationship to explicit index.
+     *
+     * @param indexName The name of the index
+     * @param relationship The id of the relationship to remove
+     * @param key The key to associate with the relationship
```
burqen: @Param key The key ~to associate~ associated with the relationship

burqen (commented):
Alright, a few last things that are still not fixed. Then I'll be happy 🎉

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
pontusmelke: but in this case it could be either a node a relationship

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
pontusmelke: In this case it can be either node or relationship, do you have a proposal for better name

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -139,18 +140,37 @@ public boolean next()
...
+            }
+            else
+            {
+                read.node( this, next++, pageCursor );
+            }
             if ( next > highMark )
             {
                 if ( highMark == NO_ID )
                 {
+                    //we are using single
```
chrisvest: It could be more clear that we are navigating with the `single` method.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Labels.java
```diff
@@ -34,6 +38,31 @@
...
+    {
+        long[] labelArray = new long[set.size()];
+        int index = 0;
+        PrimitiveIntIterator iterator = set.iterator();
+        while ( iterator.hasNext() )
+        {
+            labelArray[index++] = iterator.next();
+        }
+        return new Labels( labelArray );
+    }
```
chrisvest, line 63: These could just as well be constructor overloads, or the constructor could be private and we always use static builder methods.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -157,7 +158,45 @@ public long allRelationshipsReference()
...
+        // property reference.
+        //
+        //However if there has been changes we can have two cases:
+        //
+        //   i: The node had no prior properties, in this case we simply encode the node id in the reference
+        //      which allows the property cursor to probe tx state for properties.
+        //   ii: The node has properties, in this case we mark the actual property reference as having tx state
+        //
+        //in both cases we need to store a mapping from the computed reference to the state in the transaction state
+        //so we can retrieve it later in the property cursor.
```
chrisvest, line 172: What happens if a transaction acquires state in between getting this property reference, and dereferencing with a property cursor?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -518,4 +512,27 @@ public void nodeExplicitIndexCreateLazily( String indexName, Map<String,String>
         assertOpen();
         allStoreHolder.getOrCreateNodeIndexConfig( indexName, customConfig );
     }
+
+    private Value readProperty(long node, int propertyKey) throws EntityNotFoundException
```
chrisvest: I think this should be called `readNodeProperty`, because I have a hunch that it'll get siblings in the future.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
pontusmelke: We are splitting hairs here but in this case we don't know yet if it is a relationship reference it is the point of the function to tell us that. So in this case and for `hasNodeFlag` I prefer just `reference`

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -391,53 +413,79 @@ public Object getProperty( String key, Object defaultValue )
...
+
+        //Find ids, note we are betting on that the number of keys
+        //is small enough not to use a set here.
+        int[] propertyIds = new int[itemsToReturn];
+        for ( int i = 0; i < itemsToReturn; i++ )
+        {
+            propertyIds[i] = token.propertyKey( keys[i] );
+        }
```
chrisvest: I think it would be safe to sort the key ids and use binary search for the contains check. Any performance advantage that linear scanning would have for small N would be negligible, while the scaling to larger N would quickly favour the sorting.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -108,7 +108,7 @@ public GraphDatabaseService getGraphDatabase()
     @Override
     public void delete()
     {
-        try
+        try ( Statement ignore = actions.statement() )
```
chrisvest, line 111: Do we have to open the statement, or could we get away with a call that just does the checks and assertions down the stack?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -813,6 +814,7 @@ private void release()
             userTransactionId = 0;
             statistics.reset();
             pool.release( this );
+            operations.release();
```
chrisvest: No! This _must_ be done _before_ `pool.release( this )`, since we might otherwise call `release` on `operations` that are in use by some arbitrary other transaction.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -246,6 +246,12 @@ else if ( hasChanges && txs.nodeIsDeletedInThisTx( next ) )
                     }
                 }
             }
+            else if ( next < 0 )
+            {
+                //no more longs out there...
+                next = NO_ID;
+                return inUse();
+            }
```
chrisvest, line 256: Was this found in a test?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
```diff
@@ -262,7 +262,7 @@ void relationship( RelationshipRecord record, long reference, PageCursor pageCur
     @Override
     void property( PropertyRecord record, long reference, PageCursor pageCursor )
     {
-        properties.loadRecordByCursor( reference, record, RecordLoad.NORMAL, pageCursor );
+        properties.loadRecordByCursor( reference, record, RecordLoad.FORCE, pageCursor );
```
chrisvest: This needs a comment that convincingly argues why this is using FORCE mode.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/StoreStatement.java
```diff
@@ -465,7 +465,7 @@ private static ByteBuffer readDynamic( AbstractDynamicStore store, long referenc
         DynamicRecord record = store.newRecord();
         do
         {
-            store.getRecordByCursor( reference, record, RecordLoad.CHECK, page );
+            store.getRecordByCursor( reference, record, RecordLoad.FORCE, page );
```
chrisvest: This also needs a comment that convincingly argues why this is using FORCE mode.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
```diff
@@ -249,6 +252,14 @@ public Value propertyValue()
             return stateValue.value();
         }
 
+        Value value = readValue();
+
+        assertOpen.assertOpen();
```
chrisvest: Should we not `assertOpen` _before_ we `readValue`?

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -420,25 +418,50 @@ public Object getProperty( String key, Object defaultValue )
...
+            Token token = transaction.token();
+
+            //Find ids, note we are betting on that the number of keys
+            //is small enough not to use a set here.
+            int[] propertyIds = new int[itemsToReturn];
+            for ( int i = 0; i < itemsToReturn; i++ )
             {
-                try ( Cursor<PropertyItem> propertyCursor = statement.readOperations().nodeGetProperties( node.get() ) )
+                propertyIds[i] = token.propertyKey( keys[i] );
+            }
```
chrisvest: Perhaps my comment from earlier got lost in this back-and-forth, but I suspect it might still be profitable to sort the propertyKeyIds and use binary search for the contains check. The performance difference for small N should be negligible, while scaling much better as N grows large.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -431,7 +431,12 @@ public Object getProperty( String key, Object defaultValue )
             int[] propertyIds = new int[itemsToReturn];
             for ( int i = 0; i < itemsToReturn; i++ )
             {
-                propertyIds[i] = token.propertyKey( keys[i] );
+                String key = keys[i];
+                if ( key == null )
+                {
+                    throw new NullPointerException( String.format( "Key %d was null", i + 1 ) );
```
chrisvest: Let's keep it zero-based and say that they "key at index %d was null"

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -323,11 +323,6 @@ public CapableIndexReference index( int label, int... properties )
 
     private void assertOpen()
     {
-        if ( !ktx.isOpen() )
-        {
-            throw new NotInTransactionException( "The transaction has been closed." );
-        }
```
chrisvest, line 317: Looks like a reasonable thing to do, though.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -392,6 +391,9 @@ public void relationshipDelete( long relationship )
     @Override
     public boolean nodeAddLabel( long node, int nodeLabel ) throws EntityNotFoundException
     {
+        acquireSharedLabelLock( nodeLabel );
+        acquireExclusiveNodeLock( node );
+
         assertOpen();
```
chrisvest: Why not do the `assertOpen` call before we take any locks?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -514,16 +513,151 @@ public Value graphRemoveProperty( int propertyKey )
...
+    }
+
+    @Override
+    public void relationshipExplicitIndexCreateLazily( String indexName, Map<String,String> customConfig )
+    {
+        assertOpen();
+        allStoreHolder.getOrCreateRelationshipIndexConfig( indexName, customConfig );
+    }
+
+    private Value readProperty( long node, int propertyKey ) throws EntityNotFoundException
```
chrisvest: I think this should be called `readNodeProperty`, because I have a hunch that it'll get siblings in the future.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
```diff
@@ -177,6 +247,19 @@ private PropertyType type()
...
     public Value propertyValue()
+    {
+        if ( txStateValue != null )
+        {
+            return txStateValue.value();
+        }
+
+        Value value = readValue();
+
+        assertOpen.assertOpen();
```
chrisvest, line 259: Should we not `assertOpen` before we `readValue`?

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -387,53 +418,87 @@ public Object getProperty( String key, Object defaultValue )
...
+            //is small enough not to use a set here.
+            int[] propertyIds = new int[itemsToReturn];
+            for ( int i = 0; i < itemsToReturn; i++ )
             {
-                try ( Cursor<PropertyItem> propertyCursor = statement.readOperations().nodeGetProperties( node.get() ) )
+                String key = keys[i];
+                if ( key == null )
                 {
-                    return PropertyContainerProxyHelper.getProperties( statement, propertyCursor, keys );
+                    throw new NullPointerException( String.format( "Key %d was null", i + 1 ) );
```
chrisvest: Let's keep it zero-based and say that they "key at index %d was null".

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -387,53 +418,87 @@ public Object getProperty( String key, Object defaultValue )
...
                 }
+                propertyIds[i] = token.propertyKey( key );
```
chrisvest: Perhaps my comment from earlier got lost in this back-and-forth, but I suspect it might still be profitable to sort the propertyKeyIds and use binary search for the contains check. The performance difference for small N should be negligible, while scaling much better as N grows large.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -381,33 +388,12 @@ public void relationshipDelete( long relationship )
...
-            if ( ktx.txState().nodeIsDeletedInThisTx( node ) )
-            {
-                // already deleted, false or EntityNotFoundException
-                return false;
-            }
-        }
+        acquireSharedLabelLock( nodeLabel );
+        acquireExclusiveNodeLock( node );
 
+        assertOpen();
```
chrisvest, line 395: Why not do the `assertOpen` call before we take any locks?

chrisvest (changes requested):
I have questions, and at least one bug that I would like to see fixed.

fickludd (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
pontusmelke: No I remember, I think last time you commented was on labels but I guess same apply here. Have added a card for it. Let's add one thing at a time so it will be easier to track what is providing improvement and not.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
pontusmelke: Yes, if I thought so to but `ktx.isOpen`  is `true` when `closing` is true which is when transaction hooks run and they expect to be able to use the transaction anyway. I would love to clear up stuff around `isOpen`, `assertOpen` et. al because they all do slightly different stuff, have added a card for doing this.  The way we do here - at least to my knowledge - we do the same as we did before.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
pontusmelke: Because that is the order we used to do it, e.g. https://github.com/neo4j/neo4j/blob/3.4/community/kernel/src/main/java/org/neo4j/kernel/impl/api/LockingStatementOperations.java#L95
I assumed there was some deep underlying reason I couldn't figure out.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
pontusmelke: nah, who needs relationships :)

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
pontusmelke: I though so too, but that leads to failures in TerminationOfSlavesDuringPullUpdatesIT, we did the same before here https://github.com/neo4j/neo4j/blob/3.4/community/kernel/src/main/java/org/neo4j/kernel/impl/api/store/StorePropertyCursor.java#L106. Somehow we read inconsistent data but are "saved" by checking the transaction before actually returning data. I never really understood, I think @burqen and @tinwelint knows more.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
pontusmelke: Yes, https://github.com/neo4j/neo4j/blob/3.4/community/kernel/src/test/java/org/neo4j/kernel/impl/core/TestExceptionTypeOnInvalidIds.java#L112

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
pontusmelke: Because ConsistentPropertyReadsIT, and it is what we currently do :) 

I'll try to motivate somehow, as you see I thought `NORMAL` was the more obvious choice...

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
pontusmelke: Long term we will remove it now it is mostly convenient because it gives us the correct assertion on start and close.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
pontusmelke: Yes you are right there is a theoretical chance this might happen if we acquire the property reference first, update the node and then use the stored property reference. If we would use the API like that we would miss updates. Currently we are not using it that way but we should fix it anyhow, I have created a card for moving the property reference magic to when we write instead of when we read.

It is tricky do now since we are still using the old API for writes and we still at least have the ambition of keeping the PR sizes down :)

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/References.java
burqen: I'm happy now 🎉  I just thought you had forgotten to check them because some of them where changed and some not 😅

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
burqen: Yes. So we are traversing a dynamic property chain, reading a long key or value that spans multiple records. If the property records are being deleted and reused while we are in the middle of reading such chain we can end up reading crap. To guard for this the slave terminate all transactions that fall outside of "safezone" (are too old) when pulling updates from master. But if we don't check this termination flag before returning that does not help us and checking it before is not enough ofc because the termination and record reuse can happen after we start reading.

burqen (approved):

pontusmelke:
@chrisvest Thanks! Nicely spotted with the order of the `release` calls 👍 . 
I have fixed most of your smaller concerns and created cards for some of the slightly bigger ones.

chrisvest (approved):
Good stuff.

PR #10496, opened 2017-11-23
Fix bugs in the DelegateFileSystemAbstraction

chrisvest:
* The `copyToDirectory` method was not getting the correct `Path` implementation for the target directory, if it was anything other than the default file system.
* And the `copyToDirectory` method was also not passing the `REPLACE_EXISTING` option to `Files.copy`, which is implied by the method on the `FileSystemAbstraction` API.


chrisvest:
Build failed on flaky tests that are already being looked at. Merged to unblock failing upstream builds.

PR #10497, opened 2017-11-23
Support ScopeExpression in slotted runtime

henriknyman:
Fixes `PrimitiveExecutionContext` to work with `ScopeExpression`.

Since we already allocated a slot for the variable in a `ScopeExpression`
we implement `newWith1` on `PrimitiveExcecutionContext` to just set the
value in the existing slot instead of creating a new context like in
the `MapExecutionContext`.

For the SlottedRewriter to be able to rewrite variables in expressions
where they are declared as Variable, the corresponding runtime
variable expressions (e.g. NodeFromSlot) needs to extend Variable.
To do this we change Variable from a case class to a class, and
implement the methods that case class gave us to allow us to still
treat it as a case class (when matching, rewriting etc.).


fickludd (approved):
LGTM

PR #10498, opened 2017-11-23
Simplify backup code and add a pagecache parameter to neo4j-admin backup

chrisvest:
* Collapsed inheritance and implementation hierarchies.
* Collapsed factory-provider-builder chains.
* Remove unused code.
* Shorten names of classes.
* Hide classes that were accidentally added to the public API by making them package private.
* Avoid using service-loading for getting default instances of things. Only alternatives are service-loaded.
* Collapse the family of helper classes around configuring the online backup admin command into a single class.

changelog: [3.4] Add `--pagecache` parameter to the `neo4j-admin backup` command, with a default value of 8 MiB that now overrides the page cache setting in `neo4j.conf`.


chrisvest:
Currently marked `NOT READY FOR MERGE` because there are still more things that can be done.

enterprise/backup/src/main/java/org/neo4j/backup/BackupFlowFactory.java
```diff
@@ -19,51 +19,60 @@
...
 import org.neo4j.helpers.progress.ProgressMonitorFactory;
 import org.neo4j.io.pagecache.PageCache;
+import org.neo4j.kernel.configuration.Config;
 import org.neo4j.logging.LogProvider;
 
-/*
-Backup flows are iterate through backup strategies and make sure at least one of them is a valid backup. Handles cases when that isn't possible.
-This factory helps in the construction of them
+/**
+ * Backup flows are iterate through backup strategies and make sure at least one of them is a valid backup. Handles
```
phughk: There was a grammar error in the original, `are` can be removed.

chrisvest:
@MishaDemianenko @phughk Please take a look at this, _and_ the referenced PRs in the other repositories.

enterprise/backup/src/main/java/org/neo4j/backup/impl/BackupCopyService.java
```diff
@@ -0,0 +1,141 @@
...
+                moves.next().move( target );
+            }
+        }
+        catch ( IOException e )
+        {
+            throw new IOException( "Failed to rename backup directory from " + oldLocation + " to " + newLocation, e );
+        }
+    }
+
+    public void clearLogs( Path neo4jHome )
```
MishaDemianenko: why would we need to remove logs(i saw that according to call path's that when we had successful backups, but why)?

enterprise/backup/src/main/java/org/neo4j/backup/impl/BackupSupportingClassesFactoryProvider.java
```diff
@@ -17,53 +17,50 @@
...
+    public static Stream<BackupSupportingClassesFactoryProvider> getProvidersByPriority()
     {
-        return StreamSupport.stream( load( BackupSupportingClassesFactoryProvider.class ).spliterator(), false )
-                .sorted( ( l, r ) -> r.getPriority() - l.getPriority() )
-                .findFirst();
+        Stream<BackupSupportingClassesFactoryProvider> providers = StreamSupport.stream(
+                load( BackupSupportingClassesFactoryProvider.class ).spliterator(), false );
+        // Inject the default provider into the stream, so it participates in sorting by priority.
+        providers = Stream.concat( providers, Stream.of( new BackupSupportingClassesFactoryProvider() ) );
+        return providers.sorted( ( l, r ) -> r.getPriority() - l.getPriority() );
```
MishaDemianenko: Comparator.comparingInt( BackupSupportingClassesFactoryProvider::getPriority )

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupCommandBuilder.java
```diff
@@ -0,0 +1,358 @@
...
+            Writer fileWriter = new BufferedWriter( new FileWriter( file ) );
+            for ( Map.Entry<String,String> entry : config.getRaw().entrySet() )
+            {
+                fileWriter.write( String.format( "%s=%s\n", entry.getKey(), entry.getValue() ) );
+            }
+            fileWriter.close();
+        }
+        catch ( IOException e )
+        {
+            e.printStackTrace();
```
MishaDemianenko: Does not look like a good handling of exception? should we be able to continue if this fails?

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupCommandBuilder.java
```diff
@@ -0,0 +1,358 @@
...
+
+    /**
+     * Removes empty args and is a convenience method
+     * @param args nullable, can be empty
+     * @return cleaned command line parameters
+     */
+    private static String[] args( String... args )
+    {
+        List<String> cleanedArgs = Stream.of( args )
+                .filter( Objects::nonNull )
```
MishaDemianenko: this whole thing can be replaced with 
`Stream.of( args )
                .filter( StringUtils::isNotEmpty )
                .toArray( String[]::new )` right?

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupCommandBuilder.java
```diff
@@ -0,0 +1,358 @@
...
+                .orElse( "" );
+    }
+
+    private String argAdditionalConf( File backupTarget )
+    {
+        if ( additionalConfig == null )
+        {
+            return "";
+        }
+        File configFile = neo4jHomeFromTarget( backupTarget ).resolve( "../additional_neo4j.conf" ).toFile();
```
MishaDemianenko, line 302: what is additional_neo4j.conf and why do we need it? how its correlate with other tools behavior?

enterprise/backup/src/test/java/org/neo4j/backup/impl/BackupProtocolServiceIT.java
```diff
@@ -789,12 +794,13 @@ public void shouldHandleBackupWhenLogFilesHaveBeenDeleted() throws Exception
     private GraphDatabaseAPI deleteLogFilesAndRestart()
             throws IOException
     {
-        final FileFilter logFileFilter = pathname -> pathname.getName().contains( "logical" );
+        final Predicate<Path> logFileFilter = p -> p.getFileName().toString().contains( "logical" );
```
MishaDemianenko: would we delete any "logical" file? shouldn't it use our log files abstraction instead?

enterprise/com/src/test/java/org/neo4j/com/storecopy/FileMoveProviderTest.java
```diff
@@ -0,0 +1,246 @@
...
+        File expectedNormalCopy = new File( targetDirectory, aNormalFile.getName() );
+        assertTrue( expectedNormalCopy.exists() );
+
+        // and correct files are copied over the page cache
+        File expectedPageCacheCopy = new File( targetDirectory, aPageCacheFile.getName() );
+        assertTrue( expectedPageCacheCopy.toString(), pageCache.getCachedFileSystem().fileExists( expectedPageCacheCopy ) );
+    }
+
+    @Test
+    public void filesAreMovedBeforeDirectories() throws IOException // TODO doesnt test anything maybe
```
MishaDemianenko: TODO

MishaDemianenko (changes requested):
Only couple of small comments here and there, looks good otherwise

enterprise/backup/src/main/java/org/neo4j/backup/impl/BackupCopyService.java
chrisvest: I don't know. Can you answer, @phughk ?

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupCommandBuilder.java
chrisvest: Can you answer this one, @phughk ?

enterprise/backup/src/main/java/org/neo4j/backup/impl/BackupCopyService.java
phughk: Ah, that is an actual error on my part, sorry! It should be `clearIdFiles` and have the following implementation
https://github.com/neo4j/neo4j/blob/3.2/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java#L437

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupCommandBuilder.java
phughk: The idea behind this code is that you could supply an `Config additionalConfig` that would override defaults. This method in the builder would physically write that config to a file (that is hardcoded as additional_neo4j.conf) and add it as a parameter.
I don't think it is used anywhere at the moment, so if you think it should be thrown out then please go ahead.

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupCommandBuilder.java
phughk: It's part of a builder, so it shouldn't be throwing a checked exception but having it wrapped as unchecked should be fine. Simply printing out the trace is wrong, I admit I overlooked that.

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupCommandBuilder.java
phughk: Yes, thanks!

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupCommandBuilder.java
MishaDemianenko: while the idea to have an ability to provide additional config can be a good one or not, its important that all tools that we have should have same capabilities in terms how they can be configured, where they store their logs, etc and having something that specific for a backup can be not that good idea, but i will leave it for you guys

MishaDemianenko (approved):

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupCommandBuilder.java
chrisvest: This feature is used in the related blockdevice PR, so let's keep it.

enterprise/backup/src/main/java/org/neo4j/backup/impl/BackupCopyService.java
chrisvest: I'll have a fix for this ready soon.

PR #10501, opened 2017-11-24
Refactor contents of PBI --> ImportLogic

tinwelint:
so that the ParallelBatchImporter code is very clear on what happens
and that the different stages become eligible for use in a different implementation.

### DIFF
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -47,6 +47,7 @@ import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.configuration.Settings;
 import org.neo4j.kernel.impl.logging.LogService;
 import org.neo4j.kernel.impl.logging.StoreLogService;
+import org.neo4j.kernel.impl.store.format.RecordFormatSelector;
 import org.neo4j.kernel.impl.util.Converters;
 import org.neo4j.kernel.impl.util.OsBeanUtil;
 import org.neo4j.kernel.impl.util.Validator;
@@ -71,6 +72,7 @@ import org.neo4j.unsafe.impl.batchimport.input.csv.IdType;
 import org.neo4j.unsafe.impl.batchimport.staging.ExecutionMonitors;
 
 import static java.nio.charset.Charset.defaultCharset;
+
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.logs_directory;
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.store_internal_log_path;
 import static org.neo4j.helpers.Exceptions.launderedException;
@@ -79,6 +81,7 @@ import static org.neo4j.helpers.Strings.TAB;
 import static org.neo4j.io.ByteUnit.mebiBytes;
 import static org.neo4j.kernel.configuration.Settings.parseLongWithUnit;
 import static org.neo4j.kernel.impl.util.Converters.withDefault;
+import static org.neo4j.unsafe.impl.batchimport.AdditionalInitialIds.EMPTY;
 import static org.neo4j.unsafe.impl.batchimport.Configuration.BAD_FILE_NAME;
 import static org.neo4j.unsafe.impl.batchimport.Configuration.DEFAULT;
 import static org.neo4j.unsafe.impl.batchimport.Configuration.DEFAULT_MAX_MEMORY_PERCENT;
@@ -479,7 +482,7 @@ public class ImportTool
         }
     }
 
-    private static Long parseMaxMemory( String maxMemoryString )
+    static Long parseMaxMemory( String maxMemoryString )
     {
         if ( maxMemoryString != null )
         {
@@ -517,10 +520,13 @@ public class ImportTool
         life.start();
         BatchImporter importer = new ParallelBatchImporter( storeDir,
                 fs,
+                null, // no external page cache
                 configuration,
                 logService,
                 ExecutionMonitors.defaultVisible(),
-                dbConfig );
+                EMPTY,
+                dbConfig,
+                RecordFormatSelector.selectForConfig( dbConfig, logService.getInternalLogProvider() ) );
         printOverview( storeDir, nodesFiles, relationshipsFiles, configuration, out );
         success = false;
         try
@@ -605,7 +611,7 @@ public class ImportTool
         return file != null && file.exists() ? Config.defaults( MapUtil.load( file ) ) : Config.defaults();
     }
 
-    private static void printOverview( File storeDir, Collection<Option<File[]>> nodesFiles,
+    static void printOverview( File storeDir, Collection<Option<File[]>> nodesFiles,
             Collection<Option<File[]>> relationshipsFiles,
             org.neo4j.unsafe.impl.batchimport.Configuration configuration, PrintStream out )
     {
--- a/community/import-tool/src/test/java/org/neo4j/tooling/QuickImport.java
+++ b/community/import-tool/src/test/java/org/neo4j/tooling/QuickImport.java
@@ -22,7 +22,6 @@ package org.neo4j.tooling;
 import java.io.File;
 import java.io.IOException;
 import java.io.StringReader;
-
 import org.neo4j.csv.reader.CharSeeker;
 import org.neo4j.csv.reader.CharSeekers;
 import org.neo4j.csv.reader.Extractors;
@@ -45,6 +44,7 @@ import org.neo4j.unsafe.impl.batchimport.input.csv.Header;
 import org.neo4j.unsafe.impl.batchimport.input.csv.IdType;
 
 import static java.lang.System.currentTimeMillis;
+import static java.util.Collections.emptyList;
 
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.dense_node_threshold;
 import static org.neo4j.kernel.configuration.Settings.parseLongWithUnit;
@@ -139,7 +139,7 @@ public class QuickImport
             @Override
             public long maxMemoryUsage()
             {
-                String custom = args.get( ImportTool.Options.MAX_MEMORY.key(), null );
+                String custom = args.get( ImportTool.Options.MAX_MEMORY.key(), (String) ImportTool.Options.MAX_MEMORY.defaultValue() );
                 return custom != null ? ImportTool.parseMaxMemory( custom ) : DEFAULT.maxMemoryUsage();
             }
         };
@@ -166,6 +166,7 @@ public class QuickImport
                 consumer = new ParallelBatchImporter( dir, fileSystem, null, importConfig,
                         new SimpleLogService( sysoutLogProvider, sysoutLogProvider ), defaultVisible(), EMPTY, dbConfig,
                         RecordFormatSelector.selectForConfig( dbConfig, sysoutLogProvider ) );
+                ImportTool.printOverview( dir, emptyList(), emptyList(), importConfig, System.out );
             }
             consumer.doImport( input );
         }
--- a/community/import-tool/src/test/java/org/neo4j/tooling/QuickImport.java
+++ b/community/import-tool/src/test/java/org/neo4j/tooling/QuickImport.java
@@ -32,20 +32,25 @@ import org.neo4j.io.fs.DefaultFileSystemAbstraction;
 import org.neo4j.io.fs.FileSystemAbstraction;
 import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.impl.logging.SimpleLogService;
+import org.neo4j.kernel.impl.store.format.RecordFormatSelector;
 import org.neo4j.logging.FormattedLogProvider;
 import org.neo4j.unsafe.impl.batchimport.BatchImporter;
 import org.neo4j.unsafe.impl.batchimport.ParallelBatchImporter;
 import org.neo4j.unsafe.impl.batchimport.input.Collector;
+import org.neo4j.unsafe.impl.batchimport.input.DataGeneratorInput;
 import org.neo4j.unsafe.impl.batchimport.input.Input;
+import org.neo4j.unsafe.impl.batchimport.input.SimpleDataGenerator;
 import org.neo4j.unsafe.impl.batchimport.input.csv.Configuration;
 import org.neo4j.unsafe.impl.batchimport.input.csv.Header;
 import org.neo4j.unsafe.impl.batchimport.input.csv.IdType;
 
 import static java.lang.System.currentTimeMillis;
+
 import static org.neo4j.graphdb.factory.GraphDatabaseSettings.dense_node_threshold;
 import static org.neo4j.kernel.configuration.Settings.parseLongWithUnit;
-import static org.neo4j.tooling.DataGeneratorInput.bareboneNodeHeader;
-import static org.neo4j.tooling.DataGeneratorInput.bareboneRelationshipHeader;
+import static org.neo4j.unsafe.impl.batchimport.AdditionalInitialIds.EMPTY;
+import static org.neo4j.unsafe.impl.batchimport.input.DataGeneratorInput.bareboneNodeHeader;
+import static org.neo4j.unsafe.impl.batchimport.input.DataGeneratorInput.bareboneRelationshipHeader;
 import static org.neo4j.unsafe.impl.batchimport.input.csv.Configuration.COMMAS;
 import static org.neo4j.unsafe.impl.batchimport.input.csv.DataFactories.defaultFormatNodeFileHeader;
 import static org.neo4j.unsafe.impl.batchimport.input.csv.DataFactories.defaultFormatRelationshipFileHeader;
@@ -130,6 +135,13 @@ public class QuickImport
             {
                 return pageCacheMemory;
             }
+
+            @Override
+            public long maxMemoryUsage()
+            {
+                String custom = args.get( ImportTool.Options.MAX_MEMORY.key(), null );
+                return custom != null ? ImportTool.parseMaxMemory( custom ) : DEFAULT.maxMemoryUsage();
+            }
         };
 
         float factorBadNodeData = args.getNumber( "factor-bad-node-data", 0 ).floatValue();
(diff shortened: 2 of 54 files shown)


enterprise/neo4j-enterprise/src/test/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStoresIT.java
```diff
@@ -57,6 +57,7 @@ public void startBatchingNeoStoreWithMetricsPluginEnabled() throws Exception
                         logService, AdditionalInitialIds.EMPTY, config ) )
         {
             // empty block
```
MishaDemianenko: not that empty anymore :)

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -605,7 +611,7 @@ private static Config loadDbConfig( File file ) throws IOException
         return file != null && file.exists() ? Config.defaults( MapUtil.load( file ) ) : Config.defaults();
     }
 
-    private static void printOverview( File storeDir, Collection<Option<File[]>> nodesFiles,
+    static void printOverview( File storeDir, Collection<Option<File[]>> nodesFiles,
```
MishaDemianenko, line 614: this method still should be private

community/import-tool/src/test/java/org/neo4j/tooling/QuickImport.java
```diff
@@ -130,6 +135,13 @@ public long pageCacheMemory()
             {
                 return pageCacheMemory;
             }
+
+            @Override
+            public long maxMemoryUsage()
+            {
+                String custom = args.get( ImportTool.Options.MAX_MEMORY.key(), null );
```
MishaDemianenko: `(String) ImportTool.Options.MAX_MEMORY.defaultValue()` instead of null as default value

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
```diff
@@ -872,7 +872,10 @@ public String toString()
     public void close()
     {
         dataCache.close();
-        trackerCache.close();
+        if ( trackerCache != null )
```
MishaDemianenko, line 859: line:832 method `compareDataCache` not used anywhere

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NodeRelationshipCache.java
```diff
@@ -868,49 +859,24 @@ private boolean chunkHasChange( long nodeId, byte chunkMask )
...
-                    peekableAllTypes.next();
-                    typesToImportThisRound.add( type.getKey() );
-                    currentSetOfRelationshipsMemoryUsage += memoryUsageForThisType;
-                }
-                return typesToImportThisRound.isEmpty() ? null : typesToImportThisRound;
-            }
-        };
+    public static long calculateMaxMemoryUsage( long numberOfDenseNodes, long numberOfRelationships )
+    {
+        long maxDenseNodesForThisType = min( numberOfDenseNodes, numberOfRelationships * 2/*nodes/rel*/ );
```
MishaDemianenko, line 869: it looks like that original intention was to use `max` function instead of `min`

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/executor/DynamicTaskExecutor.java
```diff
@@ -195,7 +193,7 @@ private void parkAWhile()
         parkStrategy.park( Thread.currentThread() );
     }
 
-    private static final UncaughtExceptionHandler SILENT_UNCAUGHT_EXCEPTION_HANDLER = ( t, e ) ->
+    public static final UncaughtExceptionHandler SILENT_UNCAUGHT_EXCEPTION_HANDLER = ( t, e ) ->
```
MishaDemianenko: This field declared in the middle of the class should not be declared there.
Also, DynamicTaskExecutor does not look like handy utility class would suggest moving that field for example to `Exceptions`, if we need that kind of handler at all. What would be a scenario in which we would like to all of the uncaught exceptions instead of at least report them?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/StageExecution.java
```diff
@@ -85,6 +88,11 @@ public String getStageName()
         return stageName;
```
MishaDemianenko, line 88: `getExecutionTime` and  `getExecutionTime` is never used now

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -73,90 +77,147 @@
...
         this.ioTracer = ioTracer;
         this.externalPageCache = externalPageCache;
-        this.neoStores = newStoreFactory( DEFAULT_NAME ).openAllNeoStores( true );
-        if ( alreadyContainsData( neoStores ) )
+    }
+
+    private boolean databaseExistsAndContainsData( PageCache pageCache, File storeDir )
+    {
+        File metaDataFile = new File( storeDir, StoreType.META_DATA.getStoreFile().fileName( StoreFileType.STORE ) );
+        try ( PagedFile pagedFile = pageCache.map( metaDataFile, pageCache.pageSize(), StandardOpenOption.READ ); )
```
MishaDemianenko, line 125: it should be enough to check if META_DATA store file is present right without all this mapping dance?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -73,90 +77,147 @@
...
+
+    /**
+     * Called when expecting a previous attempt/state of a database to open, where some store files should be kept,
+     * but others deleted. All temporary stores will be deleted in this call.
+     *
+     * @param mainStoresToKeep {@link Predicate} controlling which files to keep, i.e. {@code true} means keep, {@code false} means delete.
+     * @param tempStoresToKeep {@link Predicate} controlling which files to keep, i.e. {@code true} means keep, {@code false} means delete.
+     * @throws IOException on I/O error.
+     */
+    public void pruneAndOpenExistingStore( Predicate<StoreType> mainStoresToKeep, Predicate<StoreType> tempStoresToKeep ) throws IOException
```
MishaDemianenko, line 170: this looks like a method that is quite dangerous to keep; it brings too much tempt and calling it can have really harsh consequences

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -73,90 +77,147 @@
...
         life.start();
         labelScanStore = new NativeLabelScanStore( pageCache, storeDir, FullStoreChangeStream.EMPTY, false, new Monitors(),
                 RecoveryCleanupWorkCollector.IMMEDIATE );
         life.add( labelScanStore );
     }
 
+    private void instantiateStores() throws IOException
+    {
+        neoStores = newStoreFactory( DEFAULT_NAME ).openAllNeoStores( true );
+        // TODO why do we need this counts store thing here?
```
MishaDemianenko, line 204: TODO

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
```diff
@@ -0,0 +1,528 @@
...
+    private IdGenerator idGenerator;
+    private MemoryUsageStatsProvider memoryUsageStats;
+    private InputIterable<InputNode> nodes;
+    private InputIterable<InputRelationship> relationships;
+    private InputIterable<InputNode> cachedNodes;
+    private long peakMemoryUsage;
+    private long availableMemoryForLinking;
+
+    /**
+     * Advanced usage of the parallel batch importer, for special and very specific cases. Please use
```
MishaDemianenko: comment to this constructor is a bit confusing :)

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
```diff
@@ -0,0 +1,528 @@
...
+        this.log = logService.getInternalLogProvider().getLog( getClass() );
+        this.executionMonitor = ExecutionSupervisors.withDynamicProcessorAssignment( executionMonitor, config );
+        this.maxMemory = config.maxMemoryUsage();
+
+        initialize( input );
+    }
+
+    private void initialize( Input input )
+    {
+        log.info( "Import starting" );
```
MishaDemianenko, line 150: we do not really start anything here right?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
```diff
@@ -0,0 +1,528 @@
...
+        if ( nodeLabelsCache != null )
+        {
+            nodeLabelsCache.close();
+        }
+        if ( idMapper != null )
+        {
+            idMapper.close();
+        }
+        inputCache.close();
+        // TODO close badCollector here instead of in import tool?
```
MishaDemianenko: TODO

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
```diff
@@ -116,286 +65,36 @@ public ParallelBatchImporter( File storeDir, FileSystemAbstraction fileSystem, P
...
-                    .run( max( maxMemory, peakMemoryUsage ), neoStore, highNodeId );
+            logic.importNodes();
+            logic.prepareIdMapper();
+            logic.importRelationships();
+            logic.calculateNodeDegrees();
 
-            // Count nodes per label and labels per node
-            try ( CountsAccessor.Updater countsUpdater = neoStore.getCountsStore().reset(
-                    neoStore.getLastCommittedTransactionId() ) )
+            int type = 0;
```
MishaDemianenko: I think we should hide this loop inside ImportLogic call that will try to import all the types using the same loop. Smt like: `logic.linkRelationships()` that will keep all the looping as internal detail

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/ImportLogicTest.java
```diff
@@ -0,0 +1,95 @@
...
+        int numberOfRelationships = 0;
+        for ( int i = 0; i < numberOfTypes; i++ )
+        {
+            int count = random.nextInt( 1, 100 );
+            types.add( Pair.of( "TYPE" + i, (long) count ) );
+            numberOfRelationships += count;
+        }
+        types.sort( ( t1, t2 ) -> Long.compare( t2.other(), t1.other() ) );
+        @SuppressWarnings( "unchecked" )
+        RelationshipTypeDistribution typeDistribution = new RelationshipTypeDistribution( types.stream().toArray( Pair[]::new ) );
```
MishaDemianenko: types.stream().toArray -> types.toArray

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
```diff
@@ -0,0 +1,528 @@
...
+                .run( max( maxMemory, peakMemoryUsage ), neoStore, neoStore.getNodeStore().getHighId() );
+    }
+
+    /**
+     * Builds the counts store. Requires that {@link #importNodes()} and {@link #importRelationships()} has run.
+     */
+    public void buildCountsStore()
+    {
+        // Count nodes per label and labels per node
+        try ( CountsAccessor.Updater countsUpdater = neoStore.getCountsStore().reset(
```
MishaDemianenko, line 437: we have a CountsComputer thingy that does kind of same thing already

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
```diff
@@ -0,0 +1,528 @@
...
+    }
+
+    /**
+     * Imports nodes w/ their properties and labels from {@link Input#nodes()}. This will as a side-effect populate the {@link IdMapper},
+     * to later be used for looking up ID --> nodeId in {@link #importRelationships()}. After a completed node import,
+     * {@link #prepareIdMapper()} must be called.
+     *
+     * @throws IOException on I/O error.
+     */
+    public void importNodes() throws IOException
```
MishaDemianenko, line 199: importNodes and importRelationships steps perfrom: 
neoStore.startFlushingPageCache(); 
execute... 
neoStore.stopFlushingPageCache();
why do we need those only for those 2 steps and why we can't have background flusher that flushes data during any step executuion?

MishaDemianenko (changes requested):
Only a couple of comments and questions most of them are minors.

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
tinwelint: Call from QuickImport got lost, fixing.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
tinwelint: Well, we set things up and the very next thing that happens is actual import stuff, so I think itäs a good place to have it

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
tinwelint: It turns out that the background flushing helps some stages and makes some worse. There are also some stages that don't write anything at all.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
tinwelint: Yes, kind of... it has other means of progress and configuration and what not. Do you feel it's absolutely necessary to do this change? It's also a bit unrelated to this PR

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
tinwelint: Some idea about cleaning up relationship between some classes, there's a card already for it tho.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ParallelBatchImporter.java
tinwelint: We can't hide it, it's like this because of another feature that will be built on top of it, but not right now. We could make an additional convenience method tho.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/NodeRelationshipCache.java
tinwelint: No it's `min` because there cannot be more dense nodes than there are actual dense nodes.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/executor/DynamicTaskExecutor.java
tinwelint: We want to have more control over exception printing, not that all internal threads just print stuff when they fail. In the importer typically an exception causes a panic, which spreads, making everything stop and close and then from one place print the error.

I agree about moving it out though.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/StageExecution.java
tinwelint: Perhaps you meant, for the other one, `getStateName`... it will be tough, in upcoming PR. And I believe `getExecutionTime` too.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
tinwelint: No, apparently it's not. We do something strange in store migration where some files are present at the time of starting a migration.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
tinwelint: It will be used in an upcoming feature. I know it doesn't need to be here right now... but I don't see the need to go through the work of moving it to another PR later.

Also, import and BatchingNeoStores as a whole is already dangerous and unsafe lands to be in.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
tinwelint: A mystery I haven't quite understood yet. Can it stay so that some other adventurer or future-me will find out?

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/ImportLogicTest.java
tinwelint: Tomejto - tomato, no?

enterprise/neo4j-enterprise/src/test/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStoresIT.java
tinwelint: Certainly not :)

tinwelint:
@MishaDemianenko fixed most comments.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
MishaDemianenko: it fine if we will fix it in future PR, we will need to do that anyway when we will extract importer out of kernel

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
MishaDemianenko: )) ok

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/ImportLogicTest.java
MishaDemianenko: 🍅

MishaDemianenko (approved):

PR #10502, opened 2017-11-24
Fixes deadlock issue in ForkedProcessorStep

tinwelint:
Where changing processor count frequently back and forth could potentially
result in processor threads staying alive longer than expected and could
affect the result of the processed units. This in turn could result in a
deadlock where the CompletedBatchSender could forever observe some
batches as incomplete whereas they actually were.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
@@ -19,17 +19,21 @@
  */
 package org.neo4j.unsafe.impl.batchimport.staging;
 
+import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.neo4j.unsafe.impl.batchimport.Configuration;
 import org.neo4j.unsafe.impl.batchimport.executor.ParkStrategy;
 import org.neo4j.unsafe.impl.internal.dragons.UnsafeUtil;
 
+import static org.apache.commons.lang3.SystemUtils.IS_OS_WINDOWS;
+
 import static java.lang.Integer.max;
 import static java.lang.Integer.min;
+import static java.lang.String.format;
 import static java.lang.System.nanoTime;
 import static java.util.concurrent.TimeUnit.MICROSECONDS;
-import static org.apache.commons.lang3.SystemUtils.IS_OS_WINDOWS;
+
 import static org.neo4j.unsafe.impl.internal.dragons.UnsafeUtil.getFieldOffset;
 
 /**
@@ -52,9 +56,8 @@ public abstract class ForkedProcessorStep<T> extends AbstractStep<T>
 
     private final Object[] forkedProcessors;
     private volatile int numberOfForkedProcessors;
-    private final Unit noop = new Unit( -1, null, 0 );
-    private final AtomicReference<Unit> head = new AtomicReference<>( noop );
-    private final AtomicReference<Unit> tail = new AtomicReference<>( noop );
+    private final AtomicReference<Unit> head;
+    private final AtomicReference<Unit> tail;
     private final Thread downstreamSender;
     private volatile int numberOfProcessors = 1;
     private final int maxProcessors;
@@ -67,6 +70,10 @@ public abstract class ForkedProcessorStep<T> extends AbstractStep<T>
         this.maxProcessors = config.maxNumberOfProcessors();
         this.forkedProcessors = new Object[this.maxProcessors];
 
+        Unit noop = new Unit( -1, null, 0 );
+        head = new AtomicReference<>( noop );
+        tail = new AtomicReference<>( noop );
+
         applyProcessorCount();
         downstreamSender = new CompletedBatchesSender( name + " [CompletedBatchSender]" );
         maxQueueLength = 200 + maxProcessors;
@@ -76,36 +83,38 @@ public abstract class ForkedProcessorStep<T> extends AbstractStep<T>
     {
         if ( numberOfForkedProcessors != numberOfProcessors )
         {
-            synchronized ( this )
+            awaitAllCompleted();
+
+            // Now there's nothing in the pipeline so the changes below doesn't race with other batches being queued or completed
+
+            int processors = numberOfProcessors;
+            while ( numberOfForkedProcessors < processors )
             {
-                int processors = numberOfProcessors;
-                while ( numberOfForkedProcessors < processors )
+                if ( forkedProcessors[numberOfForkedProcessors] == null )
                 {
-                    forkedProcessors[numberOfForkedProcessors] =
-                            new ForkedProcessor( numberOfForkedProcessors, head.get() );
-                    numberOfForkedProcessors++;
+                    forkedProcessors[numberOfForkedProcessors] = new ForkedProcessor( numberOfForkedProcessors, head.get() );
                 }
-                while ( numberOfForkedProcessors > processors )
-                {
-                    numberOfForkedProcessors--;
-                    // It will notice itself later, the most important thing here is that further Units
-                    // will have a lower number of processor as expected max
-                }
-                awaitEmpty();
+                numberOfForkedProcessors++;
+            }
+            while ( numberOfForkedProcessors > processors )
+            {
+                numberOfForkedProcessors--;
+                // It will notice itself later, the most important thing here is that further Units
+                // will have a lower number of processor as expected max
             }
         }
     }
 
-    private void awaitEmpty()
+    private void awaitAllCompleted()
     {
-        while ( head.get().ticket - tail.get().ticket > 0 )
+        while ( head.get() != tail.get() )
         {
             PARK.park( receiverThread = Thread.currentThread() );
         }
     }
 
     @Override
-    public synchronized int processors( int delta )
+    public int processors( int delta )
     {
         numberOfProcessors = max( 1, min( numberOfProcessors + delta, maxProcessors ) );
         return numberOfProcessors;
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
```diff
@@ -222,7 +237,6 @@ public void run()
                     {
                         sendDownstream( candidate );
                     }
-                    current.next = null;
```
chrisvest, line 225: I think this was here for preventing potential GC nepotism on the Unit chain. The `current` Unit here is the oldest instance still held live on the heap, and thus the most likely one to have been promoted to the old generation. Pointers from the old generation to objects in the young generation will keep them alive, even if the old objects are technically dead. Those old generation pointers won't be neutralised until an old-gen GC happens, and we'd like to avoid those. Nulling out this reference helps reduce this problem.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
```diff
@@ -252,6 +266,7 @@ public void run()
     {
         private final int id;
         private Unit current;
+        private volatile boolean end;
```
chrisvest: This field is never written to.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
```diff
@@ -119,11 +128,11 @@ public void start( int orderingGuarantees )
...
-    public long receive( long ticket, T batch )
+    public synchronized long receive( long ticket, T batch )
     {
         long time = nanoTime();
         applyProcessorCount();
-        while ( head.get().ticket - tail.get().ticket >= maxQueueLength )
+        while ( queuedBatches.get() >= maxQueueLength )
         {
             PARK.park( receiverThread = Thread.currentThread() );
         }
```
chrisvest, line 138: I think we could avoid synchronising this method if we pulled the top-level if-clause from `applyProcessorCount` into here, and guarded the checking with a StampedLock read/write locks. The read-lock would be used for checking if a processor change is necessary, and we'd upgrade to a write-lock if we need to apply any processor count changes. I suppose this could also be done in `applyProcessorCount` but I'd like it to be simple.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
```diff
@@ -76,36 +83,38 @@ private void applyProcessorCount()
...
                 }
-                awaitEmpty();
+                numberOfForkedProcessors++;
+            }
+            while ( numberOfForkedProcessors > processors )
+            {
+                numberOfForkedProcessors--;
+                // It will notice itself later, the most important thing here is that further Units
+                // will have a lower number of processor as expected max
             }
```
chrisvest: Why do we do this one processor at a time, in each their own while-loop?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
```diff
@@ -76,36 +83,38 @@ private void applyProcessorCount()
...
+                    forkedProcessors[numberOfForkedProcessors] = new ForkedProcessor( numberOfForkedProcessors, head.get() );
                 }
-                awaitEmpty();
+                numberOfForkedProcessors++;
+            }
+            while ( numberOfForkedProcessors > processors )
+            {
+                numberOfForkedProcessors--;
+                // It will notice itself later, the most important thing here is that further Units
+                // will have a lower number of processor as expected max
```
chrisvest: Did you mean to set the `end` field of a processor here?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
```diff
@@ -266,25 +281,19 @@ public void run()
...
                             forkedProcess( id, candidate.processors, candidate.batch );
                             candidate.processorDone( nanoTime() - time );
                         }
-                        else
-                        {
-                            // The id of this processor is greater than that of the next unit's expected max.
-                            // This means that the number of assigned processors to this step has decreased
-                            // and that this processor have reached the end of its life.
-                            break;
-                        }
```
chrisvest, line 287: Now the number of processor threads will never actually decrease. Maybe you wanted to set `end` to `true` in `applyProcessorCount`.

chrisvest (changes requested):

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
ragadeeshu: This is actually by design. There was a whole pile of races involved with actually removing the processor threads, since new threads could be created holding the same ID. We will reuse the threads if the processor count first decreases and then increases again, so we considered it to be an acceptable trade-off.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
ragadeeshu: Good catch, this is leftovers from when we shut down threads, now we only let them idle instead in order to avoid racing with the creation of new threads.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
ragadeeshu: The reason that this was here is that the 'noop' list entry was the very head of the list was kept as a field, keeping the entire list in memory. We fixed that and this call was no longer needed.

The reason that we wanted to remove this line, however, is that it could race with threads that had fallen behind, rendering them unable to catch up & causing a deadlock.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
ragadeeshu: No, we meant for the field end to no longer exist, actually ;)

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
ragadeeshu: Yeah... For the creation it makes sense, but it no longer makes sense for setting the value in the case where we should decrease, good call.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
ragadeeshu: Sounds like a very interesting idea!

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
chrisvest: You are saying it's possible for `current.isCompleted() == false && current.next.isCompleted() == true`? That means ForkedProcessors can complete their assigned work out-of-order, which is a little surprising.

tinwelint: The reason we don't null this anymore is that not all units will be processed by all processors and so some units will be completed before all processors have seen them. And to allow those processors to still follow the chain it'll have to be remain.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
tinwelint: We synchronized for simplicity reasons. Your idea is interesting, but will increase complexity for a piece of code we desperately wanted to reduce complexity in.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
tinwelint: Or maybe this can be changed since changing number of processors awaits and depletes the current queue of units... hmm

ragadeeshu: The main takeaway is, probably, since we now can have idle threads many batches can be finished while the excess threads are resting. If they are needed later and need to catch up they need this path in order to find the active part of the list.

ragadeeshu:
@chrisvest updated :)

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/ForkedProcessorStep.java
chrisvest: Alright, let's keep it as is for now. Especially since new processors now catch up from the tail.

chrisvest (approved):

chrisvest:
I suspect that performance can be improved if the forked processors avoid parking while the `stripingLock` is read- or write-locked, since we in that case can be fairly sure that something will be added to the queue pretty soon (unless we processed it already). We might even want to do additional spinning beyond that, to avoid those `park` calls.

I think we can also change the `recieverThread` field to be updated atomically, and make the waiting thread CAS the field to `null` when it stops parking. This would minimise unnecessary `unpark` calls.

The `park` and `unpark` methods involves modifying the run-queue in the OS kernel, which in turn is protected by locks that we can contend on if we call these methods too much from too many threads.

tinwelint:
@chrisvest definitely interesting to experiment further with, but for the time being there's no evidence that the overhead in ForkedProcessorStep even shows up on the radar when measuring/profiling. With that said it's probably hard to conclusively say that based on profiling since e.g. the parking and unparking could have effects that shows up in different ways.

PR #10506, opened 2017-11-26
Fully read and write buffers from/into channels

MishaDemianenko:
Update places where ByteBuffers were not fully written or read.
Most affected from change set was IdContainer that had both problems
and any of those can potentially cause id file corruption (real or "heisen")

Add additional checks into NativeLabelScanStoreMigrator and CountsMigrator
that will check that stores are in "store ok" state before performing
migration.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdContainer.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdContainer.java
@@ -137,11 +137,11 @@ public class IdContainer
     private static long readAndValidate( StoreChannel channel, File fileName ) throws IOException
     {
         ByteBuffer buffer = ByteBuffer.allocate( HEADER_SIZE );
-        int read = channel.read( buffer );
-        if ( read != HEADER_SIZE )
+        int totalBytesRead = readBuffer( channel, buffer );
+        if ( totalBytesRead != HEADER_SIZE )
         {
             throw new InvalidIdGeneratorException(
-                    "Unable to read header, bytes read: " + read );
+                    "Unable to read header, bytes read: " + totalBytesRead );
         }
         buffer.flip();
         byte storageStatus = buffer.get();
@@ -174,7 +174,7 @@ public class IdContainer
         ByteBuffer buffer = ByteBuffer.allocate( Byte.BYTES );
         buffer.put( STICKY_GENERATOR ).flip();
         fileChannel.position( 0 );
-        fileChannel.write( buffer );
+        fileChannel.writeAll( buffer );
         fileChannel.force( false );
     }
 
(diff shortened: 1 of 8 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdContainer.java
```diff
@@ -314,6 +314,22 @@ public static void createEmptyIdFile( FileSystemAbstraction fs, File file, long
         }
     }
 
+    private static int readBuffer( StoreChannel channel, ByteBuffer buffer ) throws IOException
```
burqen, line 317: Maybe let `StoreChannel` have `readAll( ByteBuffer buffer )` the same way it has `writeAll` instead of this static method?

community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/NativeLabelScanStoreMigratorTest.java
```diff
@@ -271,9 +283,9 @@ private ByteBuffer readFileContent( File nativeLabelIndex, int length ) throws I
         }
     }
 
-    private void storeFileContent( File nativeLabelIndex, ByteBuffer sourceBuffer ) throws IOException
+    private void storeFileContent( File file, ByteBuffer sourceBuffer ) throws IOException
     {
-        try ( StoreChannel storeChannel = fileSystem.create( nativeLabelIndex ) )
+        try ( StoreChannel storeChannel = fileSystem.create( file ) )
         {
             storeChannel.write( sourceBuffer );
```
burqen: `writeAll` here as well

community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/IdContainer.java
burqen, line 317: Does it make sense for `StoreChannel` to have `readAll` the same way it has `writeAll` instead of having this static method here?

community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/NativeLabelScanStoreMigratorTest.java
burqen: Can use `writeAll` here as well.

burqen (approved):
Changes look good. Found some potential improvements. Take it or leave it 🎉

burqen:
@MishaDemianenko Looks like comments where posted twice. Oh, well. Review done anyway.

community/kernel/src/test/java/org/neo4j/kernel/impl/storemigration/participant/NativeLabelScanStoreMigratorTest.java
MishaDemianenko: Updated in this PR

MishaDemianenko:
@burqen that is valid good suggestion i will do that as part of separate PR for 3.4 only together with some other things for those channels.

PR #10507, opened 2017-11-27
Core state downloader will now retry indefinitely

RagnarW:
When the raft message handler need a new snapshot it will schedule
this on the downloader service. This will pause the application process
until download is successful.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/RaftMessageHandler.java
```diff
@@ -69,7 +67,7 @@ public synchronized void handle( RaftMessages.ClusterIdAwareMessage clusterIdAwa
                 ConsensusOutcome outcome = raftMachine.handle( clusterIdAwareMessage.message() );
                 if ( outcome.needsFreshSnapshot() )
                 {
```
RagnarW, line 69: I guess we should not schedule multiple downloads if one is ongoing.

PR #10508, opened 2017-11-27
Close Kernel API IndexCursor before reuse.

fickludd:
This fixes an OOM seen in benchmarking.


SaschaPeukert (approved):
Good stuff! Except that the wrong progressor is checked and closed. Will fix that

fickludd:
@SaschaPeukert Wow, that was a last minute change that didn't work out. Called `close()` before, and then inlined it during commit review.

PR #10510, opened 2017-11-27
Disable reduce and map projections in slotted runtime

henriknyman:
Temporarily disable these expression until we have all the
compatibility implementations in PrimitiveExecutionContext.

Fixes cypher tests in docs build.

### DIFF
--- a/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlottedRewriter.scala
+++ b/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlottedRewriter.scala
@@ -21,6 +21,7 @@ package org.neo4j.cypher.internal.compatibility.v3_4.runtime
 
 import org.neo4j.cypher.internal.compatibility.v3_4.runtime.ast._
 import org.neo4j.cypher.internal.compiler.v3_4.planner.CantCompileQueryException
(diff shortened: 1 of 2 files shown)


enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlottedRewriter.scala
```diff
@@ -222,6 +223,12 @@ class SlottedRewriter(tokenContext: TokenContext) {
       case e@IsNull(Property(Variable(key), PropertyKeyName(propKey))) =>
         Not(checkIfPropertyExists(slotConfiguration, key, propKey))(e.position)
 
+      case _: ReduceExpression =>
```
systay, line 226: think we should merge these last 4 cases into a single one:
```
case _: ReduceExpression | _: DesugaredMapProjection | _: ShortestPathExpression | _: PatternExpresion =>
         throw new CantCompileQueryException(s"This expression is not yet supported in the slotted runtime")
```

henriknyman: Maybe, although you would get a less useful stack trace when you are looking at why something was not supported. But we have already removed some of these in the next PR we are currently working on. This is just a quick fix for the docs build.

PR #10511, opened 2017-11-27
add 'WebSocketFrameAggregator' handler to the websockets pipeline

ali-ince:
When browser sends out a message in split frames (a single message split across a sequence of frames), server fails to build up the original message and hangs the request indefinitely. This is caused by the fact that in `org.neo4j.bolt.transport.WebSocketFrameTranslator` we only care of `BinaryWebSocketFrame`, but not `ContinuationWebSocketFrame`.

This PR adds `WebSocketFrameAggregator` into the websocket channel pipeline, so that split frames are aggregated into a single message and processed so.

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/transport/TransportSelectionHandler.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/transport/TransportSelectionHandler.java
@@ -26,6 +26,7 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.handler.codec.ByteToMessageDecoder;
 import io.netty.handler.codec.http.HttpObjectAggregator;
 import io.netty.handler.codec.http.HttpServerCodec;
+import io.netty.handler.codec.http.websocketx.WebSocketFrameAggregator;
 import io.netty.handler.codec.http.websocketx.WebSocketServerProtocolHandler;
 import io.netty.handler.ssl.SslContext;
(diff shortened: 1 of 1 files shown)


community/bolt/src/main/java/org/neo4j/bolt/transport/TransportSelectionHandler.java
```diff
@@ -42,6 +43,7 @@
 {
     private static final String WEBSOCKET_MAGIC = "GET ";
     private static final int MAX_WEBSOCKET_HANDSHAKE_SIZE = 65536;
+    private static final int MAX_WEBSOCKET_FRAME_SIZE = 65536;
```
lutovich, line 46: Can `MAX_WEBSOCKET_FRAME_SIZE` and `MAX_WEBSOCKET_HANDSHAKE_SIZE` be made a single constant?

lutovich (approved):

community/bolt/src/main/java/org/neo4j/bolt/transport/TransportSelectionHandler.java
ali-ince: That was my first intention but later on I decided to keep those distinct because `MAX_WEBSOCKET_HANDSHAKE_SIZE` applies to the pre-websocket communication where handshake occurs, whereas `MAX_WEBSOCKET_FRAME_SIZE` applies to websocket communication.

PR #10514, opened 2017-11-27
Fix test that makes sure no files are locked by load csv

SaschaPeukert:
... by only running it on the current version. There is no need to test it in compatibility mode (which won't work in 2.3 anyway).


fickludd (approved):
Mucho nice.

PR #10516, opened 2017-11-27
Make the LargePageListIT use less native memory…

chrisvest:
…and improve the error messaging if it becomes a problem again.

### DIFF
--- a/community/unsafe/src/main/java/org/neo4j/unsafe/impl/internal/dragons/NativeMemoryAllocationRefusedError.java
+++ b/community/unsafe/src/main/java/org/neo4j/unsafe/impl/internal/dragons/NativeMemoryAllocationRefusedError.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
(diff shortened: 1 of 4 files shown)


community/unsafe/src/main/java/org/neo4j/unsafe/impl/internal/dragons/NativeMemoryAllocationRefusedError.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public NativeMemoryAllocationRefusedError( long size )
+    {
+        this.size = size;
+    }
+
+    @Override
+    public String getMessage()
+    {
+        String message = super.getMessage();
```
burqen, line 34: Can this ever be anything else than `null` if we never use the super message constructor?

burqen (approved):
Changes look good. Just curious about one thing. PR approved never the less.

community/unsafe/src/main/java/org/neo4j/unsafe/impl/internal/dragons/NativeMemoryAllocationRefusedError.java
chrisvest: Not right now. Maybe that's overly defensive?

PR #10517, opened 2017-11-27
Fix race between index population job and index map update

MishaDemianenko:
Previously IndexingService on index creation was starting index population job
inside function that was doing index map state update.
And in cases when job were started before map state update was finished
the job will miss any updates to index map that was caused by index create
request that actually created that particular population job execution.
This PR changes that behaviour and will start job population, if required,
after index state map update.


tinwelint (approved):
Good catch

PR #10520, opened 2017-11-28
Make Property rewritable by slotted runtime

henriknyman:
For the SlottedRewriter to be able to rewrite property expressions
where they are declared as Property, the corresponding runtime
expressions (e.g. NodeProperty) needs to extend Property.
To do this we change Property from a case class to a class, and
implement the methods that case class gave us to allow us to still
treat it as a case class (when matching, rewriting etc.).

- Introduces an abstract class LogicalProperty that is inherited by both Property and RuntimeProperty
- Introduces an abstract class LogicalVariable that is inherited by both Variable and RuntimeVariable
The previous solution with a class that is extended instead of a case class had problems with the copy method potentially breaking tree rewriting.

- Move calculation of needsExclusiveLock from the constructor of
the runtime SetOperation to the pipe builder, and compute it on the
AST expressions instead of the command expressions.

### DIFF
--- a/community/cypher/expressions-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/expressions/Expression.scala
+++ b/community/cypher/expressions-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/expressions/Expression.scala
@@ -30,14 +30,14 @@ object Expression {
 
   val DefaultTypeMismatchMessageGenerator = (expected: String, existing: String) => s"expected $expected but was $existing"
 
-  final case class TreeAcc[A](data: A, stack: Stack[Set[Variable]] = Stack.empty) {
+  final case class TreeAcc[A](data: A, stack: Stack[Set[LogicalVariable]] = Stack.empty) {
     def mapData(f: A => A): TreeAcc[A] = copy(data = f(data))
 
     def inScope(variable: Variable) = stack.exists(_.contains(variable))
-    def variablesInScope: Set[Variable] = stack.toSet.flatten
+    def variablesInScope: Set[LogicalVariable] = stack.toSet.flatten
 
-    def pushScope(newVariable: Variable): TreeAcc[A] = pushScope(Set(newVariable))
-    def pushScope(newVariables: Set[Variable]): TreeAcc[A] = copy(stack = stack.push(newVariables))
+    def pushScope(newVariable: LogicalVariable): TreeAcc[A] = pushScope(Set(newVariable))
+    def pushScope(newVariables: Set[LogicalVariable]): TreeAcc[A] = copy(stack = stack.push(newVariables))
     def popScope: TreeAcc[A] = copy(stack = stack.pop)
   }
 
@@ -81,8 +81,8 @@ abstract class Expression extends ASTNode {
 
   // All variables referenced from this expression or any of its children
   // that are not introduced inside this expression
-  def dependencies: Set[Variable] =
-    this.treeFold(TreeAcc[Set[Variable]](Set.empty)) {
+  def dependencies: Set[LogicalVariable] =
+    this.treeFold(TreeAcc[Set[LogicalVariable]](Set.empty)) {
       case scope: ScopeExpression =>
         acc =>
           val newAcc = acc.pushScope(scope.introducedVariables)
@@ -95,7 +95,7 @@ abstract class Expression extends ASTNode {
 
   // All (free) occurrences of variable in this expression or any of its children
   // (i.e. excluding occurrences referring to shadowing redefinitions of variable)
-  def occurrences(variable: Variable): Set[Ref[Variable]] =
+  def occurrences(variable: LogicalVariable): Set[Ref[Variable]] =
     this.treeFold(TreeAcc[Set[Ref[Variable]]](Set.empty)) {
       case scope: ScopeExpression => {
         case acc =>
@@ -108,7 +108,7 @@ abstract class Expression extends ASTNode {
       }
     }.data
 
-  def copyAndReplace(variable: Variable) = new {
+  def copyAndReplace(variable: LogicalVariable) = new {
     def by(replacement: => Expression): Expression = {
       val replacedOccurences = occurrences(variable)
       self.endoRewrite(bottomUp(Rewriter.lift {
@@ -119,8 +119,8 @@ abstract class Expression extends ASTNode {
 
   // List of child expressions together with any of its dependencies introduced
   // by any of its parent expressions (where this expression is the root of the tree)
-  def inputs: Seq[(Expression, Set[Variable])] =
-    this.treeFold(TreeAcc[Seq[(Expression, Set[Variable])]](Seq.empty)) {
+  def inputs: Seq[(Expression, Set[LogicalVariable])] =
+    this.treeFold(TreeAcc[Seq[(Expression, Set[LogicalVariable])]](Seq.empty)) {
       case scope: ScopeExpression=>
         acc =>
           val newAcc = acc.pushScope(scope.introducedVariables)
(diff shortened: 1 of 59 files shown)


henriknyman:
I am open to better suggestions on the naming of abstract base class vs. concrete case class (Variable and Property)

community/cypher/expressions-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/expressions/Expression.scala
```diff
@@ -53,6 +73,12 @@ abstract class Expression extends ASTNode {
       acc => (acc :+ e, None)
   }
 
+  // Collects all sub-expressions recursively
+  def subExpressions: Seq[Expression] = this.treeFold(List.empty[Expression]) {
+    case e: Expression if e != this =>
+      acc => (acc :+ e, Some(identity))
```
pontusmelke, line 79: Since you are appending to the list here using a `scala.List` is unnecessarily inefficient. `Vector` of `ArrayBuffer` is probably a lot faster

community/cypher/expressions-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/expressions/Property.scala
```diff
@@ -18,6 +18,35 @@ package org.neo4j.cypher.internal.v3_4.expressions
 
 import org.neo4j.cypher.internal.util.v3_4.InputPosition
 
-case class Property(map: Expression, propertyKey: PropertyKeyName)(val position: InputPosition) extends Expression {
+class Property(val map: Expression, val propertyKey: PropertyKeyName)(val position: InputPosition) extends Expression {
```
pontusmelke: What is the difference between extending a case class and extending a class where we reimplement all the "case class" methods?

Wouldn't it be nicer with `abstract class AbstractProperty`, `case class Property extends AbstractProperty` and `abstract class RuntimeProperty extends AbstractProperty`? We currently only have case classes overriding `RuntimeProperty` which means they will also override `copy`, `productElement` et. al, but if we at some point do something like

```
class MyProperty(val a: String, val b: String, val c: String, val d: String) extends Property(...)
```
we will end up with very weird `copy`, `hashCode`, `equals` etc for that class.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/SetOperation.scala
```diff
@@ -157,13 +157,12 @@ case class SetPropertyOperation(entityExpr: Expression, propertyKey: LazyPropert
       } finally ops.releaseExclusiveLock(entityId)
     }
   }
+
+  override def needsExclusiveLock = true
```
pontusmelke, line 161: This is the default right?

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlottedRewriter.scala
```diff
@@ -143,17 +143,17 @@ class SlottedRewriter(tokenContext: TokenContext) {
         val rewrittenProjection = e.projection.endoRewrite(rewriter)
         e.copy(plan = rewrittenPlan, projection = rewrittenProjection)(e.position)
 
-      case prop@Property(Variable(key), PropertyKeyName(propKey)) =>
+      case prop@Property(Variable(key), PropertyKeyName(propKey)) if !prop.isInstanceOf[RuntimeProperty] =>
```
pontusmelke: I would prefer matching on `RuntimeProperty` instead of `isInstanceOf`

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/ast/NodeProperty.scala
```diff
@@ -19,20 +19,22 @@
  */
 package org.neo4j.cypher.internal.compatibility.v3_4.runtime.ast
 
-case class NodeProperty(offset: Int, propToken: Int, name: String) extends RuntimeExpression {
+import org.neo4j.cypher.internal.v3_4.expressions.Property
+
+case class NodeProperty(offset: Int, propToken: Int, name: String)(prop: Property) extends RuntimeProperty(prop) {
```
pontusmelke, line 24: Why can't we have `..., name: String, prop: Property)`? Applies to everything that extends `RuntimeProperty`

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/ast/RuntimeProperty.scala
```diff
@@ -0,0 +1,30 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher.internal.compatibility.v3_4.runtime.ast
+
+import org.neo4j.cypher.internal.frontend.v3_4.SemanticCheck
+import org.neo4j.cypher.internal.frontend.v3_4.semantics.{SemanticCheckResult, SemanticCheckableExpression}
+import org.neo4j.cypher.internal.util.v3_4.InputPosition
+import org.neo4j.cypher.internal.v3_4.expressions.{Expression, Property}
+
+class RuntimeProperty(val prop: Property) extends Property(map = prop.map, propertyKey = prop.propertyKey)(InputPosition.NONE)
```
pontusmelke: abstract

community/cypher/expressions-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/expressions/Property.scala
```diff
@@ -18,35 +18,6 @@ package org.neo4j.cypher.internal.v3_4.expressions
 
 import org.neo4j.cypher.internal.util.v3_4.InputPosition
 
-class Property(val map: Expression, val propertyKey: PropertyKeyName)(val position: InputPosition) extends Expression {
+case class Property(map: Expression, propertyKey: PropertyKeyName)(val position: InputPosition) extends Expression {
```
pontusmelke: Argh... I shouldn't have looked commit by commit

pontusmelke (approved):
I mostly had comments on the first commit which as it turns out were all fixed in later commits.  I am not crazy about the `LogicalVariable` and `LogicalProperty` (not sure what is _logical_ about them), I don't have any super suggestions though. I don't mind calling them `AbstractX` since that is what they are.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/ast/NodeProperty.scala
henriknyman: It leads to infinite recursion in the top down tree rewriter (I tried...)

community/cypher/expressions-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/expressions/Expression.scala
henriknyman: True. It is copied from `arguments()` above it so I guess it applies to that too then. I am not sure if the order matters here, otherwise it could just add to head.

henriknyman:
@pontusmelke My first thought was also `Abstract` but @fickludd didn't like that :)

fickludd:
If you like `Abstract` we can go with that. I think I would rather name the abstract class `Variable`, and the concrete ones longer names. This makes unapplies easier to read which is preferable to easier declarations imo. Not super happy with `Logical` either. I guess I though they were logical like logical plans.

pontusmelke:
I am not saying that I love abstract, I don't see what is _logical_ about them but I do see what is _abstract_ about them. Not very important we can change name later.

PR #10523, opened 2017-11-28
Enable use of community pipes in slotted runtime

henriknyman:
Make PrimitiveExecutionContext compatible with MapExecutionContext
Add compatibility implementations of parts of the old 
ExecutionContext API used by the Community interpreted runtime pipes,
based on the assumption that slot allocation will allocate slots for
all symbols that the pipes will access.

This will enable us to fallback on using community pipes where we
lack specialized implementations in the slotted runtime.

For now we enable the remaining operators that do not do any
actual allocation during slot allocation:

- Deletes
- Set property
- Set/remove label
- LockNodes
- AssertSameNode
- SemiApplys
- SelectOrSemiApplys

Builds on https://github.com/neo4j/neo4j/pull/10520, first commit is 
"Make PrimitiveExecutionContext compatible with MapExecutionContext"

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/executionplan/DefaultExecutionResultBuilderFactory.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/executionplan/DefaultExecutionResultBuilderFactory.scala
@@ -33,18 +33,18 @@ import org.neo4j.values.virtual.MapValue
 
 import scala.collection.mutable
 
-case class DefaultExecutionResultBuilderFactory(pipeInfo: PipeInfo,
-                                                columns: List[String],
-                                                logicalPlan: LogicalPlan) extends ExecutionResultBuilderFactory {
+class DefaultExecutionResultBuilderFactory(pipeInfo: PipeInfo,
+                                           columns: List[String],
+                                           logicalPlan: LogicalPlan) extends ExecutionResultBuilderFactory {
   def create(): ExecutionResultBuilder =
-    ExecutionWorkflowBuilder()
+    new ExecutionWorkflowBuilder()
 
-  case class ExecutionWorkflowBuilder() extends ExecutionResultBuilder {
-    private val taskCloser = new TaskCloser
-    private var externalResource: ExternalCSVResource = new CSVResources(taskCloser)
-    private var maybeQueryContext: Option[QueryContext] = None
-    private var pipeDecorator: PipeDecorator = NullPipeDecorator
-    private var exceptionDecorator: CypherException => CypherException = identity
+  class ExecutionWorkflowBuilder() extends ExecutionResultBuilder {
+    protected val taskCloser = new TaskCloser
+    protected var externalResource: ExternalCSVResource = new CSVResources(taskCloser)
+    protected var maybeQueryContext: Option[QueryContext] = None
+    protected var pipeDecorator: PipeDecorator = NullPipeDecorator
+    protected var exceptionDecorator: CypherException => CypherException = identity
 
     def setQueryContext(context: QueryContext) {
       maybeQueryContext = Some(context)
@@ -64,11 +64,15 @@ case class DefaultExecutionResultBuilderFactory(pipeInfo: PipeInfo,
       exceptionDecorator = newDecorator
     }
 
+    protected def createQueryState(params: MapValue, queryId: AnyRef) = {
+      new QueryState(queryContext, externalResource, params, pipeDecorator, queryId = queryId,
+        triadicState = mutable.Map.empty, repeatableReads = mutable.Map.empty)
+    }
+
     def build(queryId: AnyRef, planType: ExecutionMode, params: MapValue,
               notificationLogger: InternalNotificationLogger, runtimeName: RuntimeName): InternalExecutionResult = {
       taskCloser.addTask(queryContext.transactionalContext.close)
-      val state = new QueryState(queryContext, externalResource, params, pipeDecorator, queryId = queryId,
-        triadicState = mutable.Map.empty, repeatableReads = mutable.Map.empty)
+      val state = createQueryState(params, queryId)
       try {
         createResults(state, planType, notificationLogger, runtimeName)
       }
@@ -110,7 +114,7 @@ case class DefaultExecutionResultBuilderFactory(pipeInfo: PipeInfo,
       }
     }
 
-    private def queryContext = maybeQueryContext.get
+    protected def queryContext = maybeQueryContext.get
 
     private def buildResultIterator(results: Iterator[ExecutionContext], isUpdating: Boolean): ResultIterator = {
       val closingIterator = new ClosingIterator(results, taskCloser, exceptionDecorator)
--- a/community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/Pipe.scala
+++ b/community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/Pipe.scala
@@ -46,12 +46,21 @@ trait Pipe {
 
   // Used by profiling to identify where to report dbhits and rows
   def id: LogicalPlanId
+
+  // TODO: Alternatively we could pass the logicalPlanId when we create contexts, and in the SlottedQueryState use the
+  // SlotConfigurations map to get the slot configuration needed for the context creation,
+  // but then we would get an extra map lookup at runtime every time we create a new context.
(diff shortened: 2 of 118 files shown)


enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/PrimitiveExecutionContext.scala
```diff
@@ -77,24 +80,29 @@ case class PrimitiveExecutionContext(slots: SlotConfiguration) extends Execution
...
   }
 
-  override def setLongAt(offset: Int, value: Long): Unit = longs(offset) = value
+  override def setLongAt(offset: Int, value: Long): Unit =
+    longs(offset) = value
 
-  override def getLongAt(offset: Int): Long = longs(offset)
+  override def getLongAt(offset: Int): Long =
+    longs(offset)
+    // Uncomment and replace with this to check for uninitialized long slots
```
pontusmelke, line 88: When are we going to do this? Maybe a TODO?

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/PrimitiveExecutionContext.scala
```diff
@@ -39,6 +41,7 @@ object PrimitiveExecutionContext {
 case class PrimitiveExecutionContext(slots: SlotConfiguration) extends ExecutionContext {
 
   override val longs = new Array[Long](slots.numberOfLongs)
+  //java.util.Arrays.fill(longs, -2L) // Uncomment this to check for uninitialized long slots
```
pontusmelke, line 44: When are we dong this? TODO?

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/PrimitiveExecutionContext.scala
```diff
@@ -110,14 +118,78 @@ case class PrimitiveExecutionContext(slots: SlotConfiguration) extends Execution
...
 
-  // This method is called from ScopeExpressions. We should already have allocated a slot for the given
-  // key, so we just set the value in the existing slot instead of creating a new context like in
-  // the MapExecutionContext.
+  //-----------------------------------------------------------------------------------------------------------
+  // Compatibility implementations of the old ExecutionContext API used by Community interpreted runtime pipes
+  //-----------------------------------------------------------------------------------------------------------
+  override def get(key: String): Option[AnyValue] = {
+    slots.get(key) match {
+      case Some(LongSlot(offset, false, CTNode)) =>
```
pontusmelke, line 126: Since this is runtime code maybe we should measure a little bit here. I think this is a case there a big instance of mess could be significantly faster. But lets measure first, this is much easier on the eye.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/PrimitiveExecutionContext.scala
```diff
@@ -168,4 +230,40 @@ case class PrimitiveExecutionContext(slots: SlotConfiguration) extends Execution
         throw new InternalException(s"Ouch, no suitable slot for key $key1 = $value1\nSlots: ${slots}")
     }
   }
+
+  def isRefInitialized(offset: Int): Boolean = {
+    refs(offset) != null
```
pontusmelke, line 237: Do we need to bounds check?

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/PrimitiveExecutionContext.scala
```diff
@@ -152,6 +152,8 @@ case class PrimitiveExecutionContext(slots: SlotConfiguration) extends Execution
   }
 
   override def +=(kv: (String, AnyValue)): this.type = {
+    // NOTE: Here we assume the slot has been allocated as a RefSlot
+    // If we aim for always using LongSlots internally we sho
```
pontusmelke: we should finish sentences.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/executionplan/DefaultExecutionResultBuilderFactory.scala
```diff
@@ -33,18 +33,18 @@ import org.neo4j.values.virtual.MapValue
 
 import scala.collection.mutable
 
-case class DefaultExecutionResultBuilderFactory(pipeInfo: PipeInfo,
-                                                columns: List[String],
-                                                logicalPlan: LogicalPlan) extends ExecutionResultBuilderFactory {
+class DefaultExecutionResultBuilderFactory(pipeInfo: PipeInfo,
```
pontusmelke, line 36: Not very important but I think I would prefer having this abstract and then concrete classes for `interpreted` and `slotted`

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/Pipe.scala
```diff
@@ -46,12 +46,21 @@ trait Pipe {
 
   // Used by profiling to identify where to report dbhits and rows
   def id: LogicalPlanId
+
+  // TODO: Alternatively we could pass the logicalPlanId when we create contexts, and in the SlottedQueryState use the
```
pontusmelke, line 50: How will this TODO be resolved? The var here is not optimal since it makes all pipes where most of them are case classes stateful.

pontusmelke (approved):
I don't like the setter on pipe, we can fix that later though.

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/PrimitiveExecutionContext.scala
henriknyman: When trying to find bugs that could be related to using uninitialized slots

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/PrimitiveExecutionContext.scala
henriknyman: Yes in a later PR I have added this comment about a suggested improvement:
```
  // TODO: As an optimization we can create a map from string key to precompiled getter/setter methods in the SlotConfiguration and use below
  // to avoid matching and unapplys
```

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/PrimitiveExecutionContext.scala
henriknyman: We shouldn't ask for out-of-bounds offsets

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/Pipe.scala
henriknyman: I guess we could implement some junk in community to pass it in to the constructor...

pontusmelke: If we are adding a TODO we should have a plan in place to make the TODO go away otherwise it'll stay there forever

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/PrimitiveExecutionContext.scala
pontusmelke: 👍

PR #10525, opened 2017-11-28
[tls] disable client-initiated renegotiation

martinfurmanski:
Client-initiated renegotiation is a possible source for DDoS.


andrewkerr9000 (approved):
LGTM

This is not a well documented setting...
https://bugs.openjdk.java.net/browse/JDK-8073643

chrisvest:
I think it would be good to have comments above those lines that explain why that particular setting is a good idea.

martinfurmanski:
@chrisvest Good enough?

chrisvest:
@martinfurmanski Yes.

PR #10526, opened 2017-11-28
Solves problem around queries with many optional matches

systay:
When multiple optional matches exist, and they do not have dependencies
between each other, the planner is free to solve them in any order. This, unfortunately, 
causes a problem where the planner thinks it has failed to solve the original query.

This PR fixes this by comparing QG in a different way when we encounter this rare case.

### DIFF
--- a/community/cypher/ir-3.2/src/main/scala/org/neo4j/cypher/internal/ir/v3_2/QueryGraph.scala
+++ b/community/cypher/ir-3.2/src/main/scala/org/neo4j/cypher/internal/ir/v3_2/QueryGraph.scala
@@ -23,8 +23,10 @@ import org.neo4j.cypher.internal.frontend.v3_2.ast._
 import org.neo4j.cypher.internal.ir.v3_2.helpers.ExpressionConverters._
 
 import scala.collection.{GenTraversableOnce, mutable}
+import scala.runtime.ScalaRunTime
 
-case class QueryGraph(patternRelationships: Set[PatternRelationship] = Set.empty,
+case class QueryGraph(// !!! If you change anything here, make sure to update the equals method at the bottom of this class !!!
+                      patternRelationships: Set[PatternRelationship] = Set.empty,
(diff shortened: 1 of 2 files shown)


OliviaYtterbrink (approved):

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
```diff
@@ -29,6 +29,39 @@ import scala.collection.mutable.ArrayBuffer
 
 class MatchAcceptanceTest extends ExecutionEngineFunSuite with QueryStatisticsTestSupport with NewPlannerTestSupport {
 
+  test("should not fail") {
```
OliviaYtterbrink: a bit more descriptive test name would be good

community/cypher/ir-3.2/src/main/scala/org/neo4j/cypher/internal/ir/v3_2/QueryGraph.scala
```diff
@@ -361,23 +367,79 @@ case class QueryGraph(patternRelationships: Set[PatternRelationship] = Set.empty
...
+
+  /**
+    * We have to do this special treatment of QG to avoid problems when checking that the produced plan actually
+    * solves what we set out to solve. In some rare circumstances, we'll get a few optional matches that are independent of each other.
+    *
+    * Given the way our planner works, it can unpredictably plan these optional matches in different orders, which leads to an exception being thrown when
+    * checking that the correct query has been solved.
+    */
+  override def equals(in: scala.Any): Boolean = in match {
+    case other: QueryGraph =>
```
OliviaYtterbrink: override canEqual as well

OliviaYtterbrink (changes requested):

OliviaYtterbrink (approved):

PR #10527, opened 2017-11-28
[tls] make version 1.2 the only default

martinfurmanski:
This is the securest default, but might break deployments already
relying on 1.0 or 1.1.


andrewkerr9000 (approved):
LGTM

PR #10529, opened 2017-11-28
Store correct last stored transaction id in transaction log file header on rotation

MishaDemianenko:
Previously on transactional log file rotation we were using last committed
transaction id as the id of the last transaction that can be found in the previous log file.
And that is not always correct since there is a window where transaction is already
stored in transaction log file, but not yet published as committed.
In case if the wrong header will be stored we will not gonna be able to find transactions
that were stored in older files since we will look for them in the incorrect place.

This PR change that behavior by using current committing transaction id
instead of last committed transaction id.
It's safe to do so since during rotation:
- we flush previous tx log file so we guarantee the durability of tx that still registered as committing;
- the operation is done under tx file lock so its impossible to race with
any ongoing commit

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/files/LogFilesBuilder.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/files/LogFilesBuilder.java
@@ -221,9 +221,10 @@ public class LogFilesBuilder
         requireNonNull( fileSystem );
         Supplier<LogVersionRepository> logVersionRepositorySupplier = getLogVersionRepositorySupplier();
         LongSupplier lastCommittedIdSupplier = lastCommittedIdSupplier();
+        LongSupplier committingTransactionIdSupplier = committingIdSupplier();
         long rotationThreshold = getRotationThreshold();
         return new TransactionLogFilesContext( rotationThreshold, logEntryReader,
-                lastCommittedIdSupplier, logFileCreationMonitor, logVersionRepositorySupplier, fileSystem );
+                lastCommittedIdSupplier, committingTransactionIdSupplier, logFileCreationMonitor, logVersionRepositorySupplier, fileSystem );
     }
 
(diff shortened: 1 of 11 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/files/LogFilesBuilder.java
```diff
@@ -309,6 +310,37 @@ private LongSupplier lastCommittedIdSupplier() throws IOException
         }
     }
 
+    private LongSupplier committingIdSupplier() throws IOException
+    {
+        if ( transactionIdStore != null )
+        {
+            return () -> transactionIdStore.committingTransactionId();
```
chrisvest: I think we should use method references instead, because that captures the target object instance, instead of capturing this builder object and referencing through the `transactionIdStore` field on every call. This also applies to the dependency resolver based lambda bellow, and in the `lastCommittedIdSupplier` method above.

chrisvest (commented):
Had one comment, and a question: Why is this PR for 3.4 instead of 3.3, where the bug looks like it was introduced?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/files/LogFilesBuilder.java
MishaDemianenko: updated to method references in mentioned cases, except where its based on dependency resolver since that is lazy case and we do not have TransactionIdStore during construction time yet

MishaDemianenko:
@chrisvest this problem exists even in earlier versions (at least form 3.0) but before merging it down I want to have "a few" test runs over different test pipeline to make sure we do not introduce any other problems

chrisvest (approved):

chrisvest:
@MishaDemianenko Okay, that's a fair reason. Approved.

PR #10535, opened 2017-11-29
Move SlottedPipeBuilderTest and remove kernel dependency

fickludd:
Builds on https://github.com/neo4j/neo4j/pull/10541
First commit is "Remove kernel dependency from planner spi module"


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/BuildSlottedExecutionPlan.scala
```diff
@@ -68,7 +68,7 @@ object BuildSlottedExecutionPlan extends Phase[EnterpriseRuntimeContext, Logical
       }
       val (logicalPlan, physicalPlan) = rewritePlan(context, from.logicalPlan)
       val converters = new ExpressionConverters(SlottedExpressionConverters, CommunityExpressionConverter)
-      val pipeBuilderFactory = EnterprisePipeBuilderFactory(physicalPlan)
+      val pipeBuilderFactory = SlottedPipeBuilder.Factory(physicalPlan)
```
henriknyman: I actually liked the more descriptive case class name better (easier to find, better `toString`, if you direct import it you still see what it is etc.), so I would suggest `SlottedPipeBuilderFactory`.

henriknyman (commented):
No more work to be done here, but since this conflicts with a series of stacked slotted runtime PRs I added the NOT READY FOR MERGE just so we can control the timing of the merge in order to maximize the throughput.

henriknyman:
Attached some fixes for comments on a previous review (https://github.com/neo4j/neo4j/pull/10523)

PR #10537, opened 2017-11-29
Store channel remove "write" and add "readAll"

MishaDemianenko:
Remove error-prone "write" method in StoreChannel and add helper readAll.


community/io/src/main/java/org/neo4j/io/fs/AbstractStoreChannel.java
```diff
@@ -32,25 +32,25 @@ public FileLock tryLock() throws IOException
     }
 
     @Override
-    public int write( ByteBuffer src, long position ) throws IOException
```
burqen: Actually looks like `AbstractStoreChannel` can be completely removed. No usages.

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstractionTest.java
```diff
@@ -39,7 +39,7 @@
 import org.neo4j.io.fs.OpenMode;
 import org.neo4j.io.fs.StoreChannel;
 
-import static java.nio.ByteBuffer.allocateDirect;
+import static java.nio.ByteBuffer.allocate;
```
burqen, line 42: What is the reason for using `allocate` instead of `allocateDirect`? Change seems fine, just curious.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/FailureStorage.java
```diff
@@ -139,8 +139,8 @@ private String readFailure( File failureFile ) throws IOException
         try ( StoreChannel channel = fs.open( failureFile, OpenMode.READ ) )
         {
             byte[] data = new byte[(int) channel.size()];
-            int readData = channel.read( ByteBuffer.wrap( data ) );
-            return readData <= 0 ? "" : UTF8.decode( withoutZeros( data ) );
+            channel.read( ByteBuffer.wrap( data ) );
```
burqen: `readAll` here as well?

burqen (changes requested):
Looks good! Found some small points of improvement.

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstractionTest.java
MishaDemianenko: When you allocate direct buffer you need to manage it yourself including de-allocating it to free that memory that was never done first of all. When you do that you want to put some data outside of heap, for example, to avoid GC to spend its time on it. And in this case, its only additional burden that does not bring any gain at all.

MishaDemianenko:
@burqen updated, thx for comments!

community/io/src/test/java/org/neo4j/graphdb/mockfs/EphemeralFileSystemAbstractionTest.java
burqen: Aah, yes of course! The de-allocation. Thanks for explaining ☺️

burqen:
Cant approve from phone apparently. But it is approved! Happy merging!

burqen (approved):

PR #10541, opened 2017-11-29
Support Foreach in slotted runtime

henriknyman:
- Add ForeachSlottedPipe
- Allocate foreach inner variable on lhs before rhs allocation
- Pass SemanticTable to slot allocation so we can use long slots
for variables of lists of entities
- Add SlottedPipeBuilderUtils with helpers to make specialized
getter/setter functions for slots

- Minor optimization of empty projections
- Fix arguments for AbstractLetSemiApply

Update succeeding acceptance tests

### DIFF
--- a/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
+++ b/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
@@ -20,6 +20,7 @@
 package org.neo4j.cypher.internal.compatibility.v3_4.runtime
 
 import org.neo4j.cypher.internal.compatibility.v3_4.runtime.SlotConfiguration.Size
+import org.neo4j.cypher.internal.frontend.v3_4.semantics.SemanticTable
 import org.neo4j.cypher.internal.ir.v3_4.IdName
 import org.neo4j.cypher.internal.util.v3_4.InternalException
 import org.neo4j.cypher.internal.util.v3_4.symbols._
@@ -56,6 +57,7 @@ object SlotAllocation {
     * @return the slot configurations of every operator.
     */
   def allocateSlots(lp: LogicalPlan,
+                    semanticTable: SemanticTable,
                     initialSlotsAndArgument: Option[SlotsAndArgument] = None): PhysicalPlan = {
 
     val allocations = new mutable.OpenHashMap[LogicalPlanId, SlotConfiguration]()
@@ -99,7 +101,7 @@ object SlotAllocation {
           val argument = if (argumentStack.isEmpty) NO_ARGUMENT()
                          else argumentStack.top
           recordArgument(current, argument)
-          val slotsIncludingExpressions = allocateExpressions(current, nullable, argument.slotConfiguration.copy(), allocations, arguments)
+          val slotsIncludingExpressions = allocateExpressions(current, nullable, argument.slotConfiguration.copy(), allocations, arguments)(semanticTable)
           val result = allocate(current, nullable, slotsIncludingExpressions)
           allocations += (current.assignedId -> result)
           resultStack.push(result)
@@ -108,7 +110,7 @@ object SlotAllocation {
           val sourceSlots = resultStack.pop()
           val argument = if (argumentStack.isEmpty) NO_ARGUMENT()
                          else argumentStack.top
-          val slotsIncludingExpressions = allocateExpressions(current, nullable, sourceSlots, allocations, arguments)
+          val slotsIncludingExpressions = allocateExpressions(current, nullable, sourceSlots, allocations, arguments)(semanticTable)
           val result = allocate(current, nullable, slotsIncludingExpressions, recordArgument(_, argument))
           allocations += (current.assignedId -> result)
           resultStack.push(result)
@@ -116,6 +118,7 @@ object SlotAllocation {
         case (Some(left), Some(right)) if (comingFrom eq left) && isAnApplyPlan(current) =>
           planStack.push((nullable, current))
           val argumentSlots = resultStack.top
+          allocateLhsOfApply(current, nullable, argumentSlots)(semanticTable) // This never copies the slot configuration
           argumentStack.push(SlotsAndArgument(argumentSlots.copy(), argumentSlots.size()))
           populate(right, nullable)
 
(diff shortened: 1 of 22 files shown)


enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/ForeachSlottedPipe.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.neo4j.cypher.internal.runtime.interpreted.{ExecutionContext, ListSupport}
+import org.neo4j.cypher.internal.v3_4.logical.plans.LogicalPlanId
+
+import scala.collection.JavaConverters._
+
+case class ForeachSlottedPipe(lhs: Pipe, rhs: Pipe, innerVariableSlot: Slot, expression: Expression)
+                             (val id: LogicalPlanId = LogicalPlanId.DEFAULT)
+  extends PipeWithSource(lhs) with Pipe with ListSupport {
+
+  val setVariableFun = makeSetValueInSlotFunctionFor(innerVariableSlot)
```
systay: I think it's nicer if this stuff is done by the PipeBuilder and not the pipes themself. Makes it easier to see what code is runtime code and which is compile time

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/ProjectionSlottedPipe.scala
```diff
@@ -39,11 +39,16 @@ case class ProjectionSlottedPipe(source: Pipe, introducedExpressions: Map[Int, E
         val result = expression(ctx, state)
         ctx.setRefAt(offset, result)
   }
+
   protected def internalCreateResults(input: Iterator[ExecutionContext], state: QueryState): Iterator[ExecutionContext] = {
-    input.map {
-      ctx =>
-        projectionFunctions.foreach(_(ctx, state))
-        ctx
+    if (projectionFunctions.isEmpty)
```
systay, line 44: This is weird. Why would we end up with a projection operator that does not project? I much prefer if we simply got rid of the plan instead.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -518,9 +549,11 @@ object SlotAllocation {
          _: AbstractSemiApply |
          _: AbstractSelectOrSemiApply |
          _: AbstractLetSelectOrSemiApply |
+         _: AbstractLetSemiApply |
```
systay, line 552: Just noticed we have very similar check here: https://github.com/neo4j/neo4j/blob/3.4/community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/CardinalityCostModel.scala#L156

Maybe it's time to add a trait to the plans instead of duplicating this code in at least two places?

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -501,6 +508,30 @@ object SlotAllocation {
         throw new SlotAllocationFailed(s"Don't know how to handle $p")
     }
 
+  private def allocateLhsOfApply(plan: LogicalPlan,
```
systay, line 511: I don't see a single new unit test in slot allocation around this new code.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -469,6 +473,9 @@ object SlotAllocation {
         lhs.newReference(collectionName.name, nullable, CTList(CTAny))
         lhs
 
+      case _: ForeachApply =>
+        lhs
```
systay, line 477: It's not clear to me why slot allocation for ForeachApply is not done here. It's done on line 516 instead.

systay (changes requested):

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
henriknyman: Ah, that would be good

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/ProjectionSlottedPipe.scala
henriknyman: Or not build the pipe. This is a quick fix for the observation that we get a lot of empty projections from the rewriters around ORDER BY and RETURN. We should solve that, but do you prefer to have this problem until then?

enterprise/cypher/slotted-runtime/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/pipes/ForeachSlottedPipe.scala
henriknyman: I think there is a trade-off. It also makes the Pipe tree harder to read and understand when debugging if you do not see the slots but just functions everywhere (and we will spend a lot of time debugging).
Since we currently have mixed approaches I think we should first decide on one approach and then unify this across all the pipes.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
henriknyman: The slot for the inner variable has to be included in the argument for the RHS. The way slot allocation currently works the whole RHS slot configuration have already been allocated when we come here.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
henriknyman: We thought we could spend more time on runtime acceptance tests instead. But I added a test now.

systay (approved):

PR #10542, opened 2017-11-29
Use new API for reading nodes in interpreted runtime

pontusmelke:
This PR switches to use the new API in `QueryContex`t for `all`, `allPrimitive` and label and property reads. The PR also contains a refactoring of `QueryContext` so that operations return `NodeValue` and `EdgeValue` instead of `Node` and `Relationship`.

~Builds on top of https://github.com/neo4j/neo4j/pull/10495~


community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -161,11 +161,17 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
...
-  override def isLabelSetOnNode(label: Int, node: Long): Boolean =
-    transactionalContext.statement.readOperations().nodeHasLabel(node, label)
+  override def isLabelSetOnNode(label: Int, node: Long): Boolean = {
+    var nodeCursor: NodeCursor = null;
+    try {
+      nodeCursor = cursors.allocateNodeCursor()
+      reads().singleNode(node, nodeCursor)
+      if (!nodeCursor.next()) false
+      else nodeCursor.labels().contains(label)
+    }
```
fickludd: Forgot to close the cursor. If we go this route, we will not some utility for try-with-resources in scala.

pontusmelke: we already have that https://github.com/neo4j/neo4j/blob/3.4/community/cypher/frontend-3.4/src/main/scala/org/neo4j/cypher/internal/frontend/v3_4/helpers/package.scala#L27. However requires lambda, thought we'll do it without allocation here but you are right that I forgot to close it here.

fickludd: 1) Cool, I didn't know we had that... should be moved out of the frontend module though.
2) Ok, I like less allocation, so let's to that.

fickludd (approved):

fickludd:
I'm a little surprised by the added-nodes snapshot in `NodeCursor`. Did we do this before? I though it was only needed to freeze the `hasTxStateWithChanges()` check on init... I guess that breaks on `CREATE ... MATCH () CREATE ()`. Mystified!

pontusmelke:
@fickludd No we just thought/hoped we could do without it but we ended up with endless loops for stuff like

```
try(Transaction tx = db.beginTx())
{
   db.execute("CREATE ()");
   db.execute("MATCH() CREATE()");
}
```
In this case we will have `hasTxStateWithChanges == true` and then we will be endlessly matching on changes in the tx state and create new nodes until we run out of memory.

PR #10548, opened 2017-11-30
Core state downloader will now retry indefinitely

RagnarW:
When the raft message handler need a new snapshot it will schedule
this on the downloader service. This will pause the application process
until download is successful.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
@@ -236,5 +236,6 @@ public class CoreServerModule
         life.add( raftServer ); // must start before core state so that it can trigger snapshot downloads when necessary
         life.add( coreLife );
         life.add( catchupServer ); // must start last and stop first, since it handles external requests
+        life.add( downloadService );
     }
 }
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
@@ -55,6 +55,7 @@ import org.neo4j.causalclustering.core.state.RaftLogPruner;
 import org.neo4j.causalclustering.core.state.RaftMessageHandler;
 import org.neo4j.causalclustering.core.state.machines.CoreStateMachinesModule;
 import org.neo4j.causalclustering.core.state.snapshot.CoreStateDownloader;
+import org.neo4j.causalclustering.core.state.snapshot.CoreStateDownloaderService;
 import org.neo4j.causalclustering.core.state.storage.DurableStateStorage;
 import org.neo4j.causalclustering.core.state.storage.StateStorage;
 import org.neo4j.causalclustering.discovery.TopologyService;
@@ -177,10 +178,13 @@ public class CoreServerModule
 
         CoreStateDownloader downloader = new CoreStateDownloader( localDatabase, servicesToStopOnStoreCopy,
                 remoteStore, catchUpClient, logProvider, storeCopyProcess, coreStateMachinesModule.coreStateMachines,
-                snapshotService, commandApplicationProcess, topologyService );
+                snapshotService, topologyService );
+
+        CoreStateDownloaderService downloadService = new CoreStateDownloaderService( platformModule
+                .jobScheduler, downloader, commandApplicationProcess, logProvider );
 
         RaftMessageHandler messageHandler = new RaftMessageHandler( localDatabase, logProvider,
-                consensusModule.raftMachine(), downloader, commandApplicationProcess );
+                consensusModule.raftMachine(), downloadService, commandApplicationProcess );
 
         int queueSize = config.get( CausalClusteringSettings.raft_in_queue_size );
         int maxBatch = config.get( CausalClusteringSettings.raft_in_queue_max_batch );
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
@@ -39,8 +39,8 @@ import org.neo4j.causalclustering.core.CausalClusteringSettings;
 import org.neo4j.causalclustering.core.IdentityModule;
 import org.neo4j.causalclustering.core.consensus.ConsensusModule;
 import org.neo4j.causalclustering.core.consensus.ContinuousJob;
-import org.neo4j.causalclustering.core.consensus.RaftMessages;
 import org.neo4j.causalclustering.core.consensus.LeaderAvailabilityHandler;
+import org.neo4j.causalclustering.core.consensus.RaftMessages;
 import org.neo4j.causalclustering.core.consensus.RaftServer;
 import org.neo4j.causalclustering.core.consensus.log.pruning.PruningScheduler;
 import org.neo4j.causalclustering.core.consensus.membership.MembershipWaiter;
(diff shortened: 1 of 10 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/CoreServerModule.java
```diff
@@ -177,10 +178,13 @@ private OnlineBackupKernelExtension pickBackupExtension( NeoStoreDataSource data
 
         CoreStateDownloader downloader = new CoreStateDownloader( localDatabase, servicesToStopOnStoreCopy,
                 remoteStore, catchUpClient, logProvider, storeCopyProcess, coreStateMachinesModule.coreStateMachines,
-                snapshotService, commandApplicationProcess, topologyService );
+                snapshotService, topologyService );
+
+        CoreStateDownloaderService downloadService = new CoreStateDownloaderService( platformModule
```
martinfurmanski, line 185: Make this a LifeCycle and add it to platformModule.life at some appropriate place. Downloads should not be allowed when stopped and stopping could interrupt an on-going download.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloaderService.java
```diff
@@ -0,0 +1,66 @@
...
+        this.applicationProcess = applicationProcess;
+        this.log = logProvider.getLog( getClass() );
+        this.downloadSnapshotGroup = new JobScheduler.Group( "download snapshot", POOLED );
+    }
+
+    public void scheduleDownload( LeaderLocator leaderLocator )
+    {
+        if ( currentJob == null || currentJob.hasCompleted() )
+        {
+            synchronized ( this )
```
martinfurmanski: Just make the entire function synchronized. This double-checking trick is both unsafe and unnecessary.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/PersistentSnapshotDownloader.java
```diff
@@ -0,0 +1,115 @@
...
+        COMPLETED
+    }
+
+    @Override
+    public void run()
+    {
+        state = State.RUNNING;
+        try
+        {
+            applicationProcess.pauseApplier( CoreStateDownloaderService.OPERATION_NAME );
```
martinfurmanski: This will not work correctly on subsequent attempts after failure because you will pause several times, but only resume once on the last successful attempt. The pause/resume functionality is reference counting so you will be left with a reference count > 0. You need an object with a lifetime of all the attempts to pause and resume, once each.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/PersistentSnapshotDownloader.java
```diff
@@ -0,0 +1,115 @@
...
+            while ( true )
+            {
+                if ( Thread.interrupted() )
+                {
+                    break;
```
martinfurmanski: Log a warning when interrupted.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/PersistentSnapshotDownloader.java
```diff
@@ -0,0 +1,115 @@
...
+                }
+                catch ( StoreCopyFailedException e )
+                {
+                    log.error( "Failed to download snapshot. Retrying in {} ms.", timeout.getMillis(), e );
+                }
+                catch ( NoLeaderFoundException e )
+                {
+                    log.warn( "No leader found. Retrying in {} ms.", timeout.getMillis() );
+                }
+                LockSupport.parkNanos( TimeUnit.MILLISECONDS.toNanos( timeout.getMillis() ) );
```
martinfurmanski: Parking cannot be interrupted.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloaderService.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public CoreStateDownloaderService( JobScheduler jobScheduler, CoreStateDownloader downloader,
+            CommandApplicationProcess applicationProcess,
+            LogProvider logProvider )
+    {
+        this.jobScheduler = jobScheduler;
+        this.downloader = downloader;
+        this.applicationProcess = applicationProcess;
+        this.log = logProvider.getLog( getClass() );
+        this.downloadSnapshotGroup = new JobScheduler.Group( "download snapshot", POOLED );
```
martinfurmanski: I think you should just add this group in the long list in JobScheduler.Groups and reference that.

martinfurmanski (changes requested):

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloaderService.java
```diff
@@ -0,0 +1,66 @@
...
+    {
+        if ( currentJob == null || currentJob.hasCompleted() )
+        {
+            currentJob = new PersistentSnapshotDownloader( leaderLocator, applicationProcess, downloader, log );
+            jobScheduler.schedule( downloadSnapshot, currentJob );
+        }
+    }
+
+    @Override
+    public void stop() throws Throwable
```
martinfurmanski: synchronize this also

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloaderService.java
```diff
@@ -0,0 +1,66 @@
...
+
+import static org.neo4j.kernel.impl.util.JobScheduler.Groups.downloadSnapshot;
+
+public class CoreStateDownloaderService extends LifecycleAdapter
+{
+    private final JobScheduler jobScheduler;
+    private final CoreStateDownloader downloader;
+    private final CommandApplicationProcess applicationProcess;
+    private final Log log;
+    private PersistentSnapshotDownloader currentJob = null;
```
martinfurmanski: null is already the default

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/PersistentSnapshotDownloader.java
```diff
@@ -0,0 +1,157 @@
...
+    @Override
+    public void run()
+    {
+        if ( !initialStateOk() )
+        {
+            return;
+        }
+        try
+        {
+            state = State.RUNNING;
```
martinfurmanski: This variable state is accessed from two threads so you need some type of synchronization mechanism. Might be enough to make it volatile. I think it can be good to split it up into two variables, state and keepRunning, because right now you have two threads updating the same variable and their interactions are hard to follow. The variable state can be updated by PersistentSnaptshotDownloader#run and keepRunning can be updated from #stop. The while-loop then checks keepRunning.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/PersistentSnapshotDownloader.java
```diff
@@ -0,0 +1,157 @@
...
+            log.error( "Persistent snapshot downloader was interrupted" );
+        }
+        finally
+        {
+            applicationProcess.resumeApplier( OPERATION_NAME );
+            state = State.COMPLETED;
+        }
+    }
+
+    private boolean initialStateOk()
```
martinfurmanski: I don't think you should need this method at all. The state transitions should be obvious and simple. Do you really need INITIATED at all? Can't it just start out in RUNNING-state? Perhaps all you really need is a simple boolean which says isComplete and similarly the method isRunning(), which is only used from tests, can perhaps be eliminated?

martinfurmanski (changes requested):

martinfurmanski (approved):

PR #10549, opened 2017-11-30
Better progress monitor in importer

tinwelint:
Importer now shows a more understandable progress. This is split up into:

- Importing nodes
- Importing relationships
- Linking data together
- Post processing

Each of those has printing of dots from from 0..100%. Also estimates of memory and disk usage are provided for each of those stages.


community/csv/src/main/java/org/neo4j/csv/reader/MultiReadable.java
```diff
@@ -19,38 +19,74 @@
...
+    public MultiReadable( RawIterator<CharReadable,IOException> readers )
+    {
+        this.actual = readers;
+    }
 
-    public MultiReadable( RawIterator<Reader,IOException> actual ) throws IOException
+    @Override
+    public void close() throws IOException
+    {
+        closeCurrent();
```
MishaDemianenko, line 51: shouldn't we try to close all readers here and not only current one?

community/csv/src/test/java/org/neo4j/csv/reader/BufferedCharSeekerTest.java
```diff
@@ -854,10 +854,12 @@ public void closeSeeker() throws IOException
     {
         private final StringReader reader;
         private final int maxBytesPerRead;
+        private final String data;
         private int position;
```
MishaDemianenko: field is never assigned

community/kernel/src/main/java/org/neo4j/helpers/Format.java
```diff
@@ -35,10 +35,12 @@
      * sources is an easier task.
      */
     public static final TimeZone DEFAULT_TIME_ZONE = TimeZone.getTimeZone( "UTC" );
-    public static final String DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSSZ";
-    public static final String TIME_FORMAT = "HH:mm:ss.SSS";
 
     private static final String[] BYTE_SIZES = { "B", "kB", "MB", "GB" };
+    private static final String[] COUNT_SIZES = { "", "k", "M", "bn", "tr" };
```
MishaDemianenko: At least case should be aligned between different count sizes

community/kernel/src/main/java/org/neo4j/helpers/Format.java
```diff
@@ -106,14 +108,24 @@ public static String time( Date date, TimeZone timeZone )
...
     public static String bytes( long bytes )
     {
-        double size = bytes;
-        for ( String suffix : BYTE_SIZES )
+        return suffixCount( bytes, BYTE_SIZES, KB );
+    }
+
+    public static String count( long count )
+    {
+        return suffixCount( count, COUNT_SIZES, 1_000 );
```
MishaDemianenko, line 116: do not think its an idea that someone will ever see TB as count suffix

community/kernel/src/main/java/org/neo4j/kernel/impl/cache/MeasureDoNothing.java
```diff
@@ -23,20 +23,69 @@
...
+    }
+
+    public static Monitor loggingMonitor( Log log )
+    {
+        return new Monitor()
+        {
+            @Override
+            public void started()
+            {
+                log.debug( "GC Monitor started. " );
```
MishaDemianenko, line 46: while this change improves existing code we should stop at some point calling this thing a GC monitor, because its not

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInput.java
```diff
@@ -237,4 +251,77 @@ public Collector badCollector()
...
+        long[] estimates = new long[4]; // [entity count, property count, property size, labels (for nodes only)]
+        for ( DataFactory<E> dataFactory : dataFactories ) // one input group
+        {
+            // One group of input files
+            Header header = null;
+            RawIterator<CharReadable,IOException> dataItems = dataFactory.create( config ).stream();
+            while ( dataItems.hasNext() )
+            {
+                CharReadable stream = dataItems.next();
+                // TODO for now only a single 1MB chunk from the start of each file is sampled.
```
MishaDemianenko: this todo should be fixed or converted into the card

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/HumanUnderstandableExecutionMonitor.java
```diff
@@ -0,0 +1,466 @@
...
+        {
+            stashedProgress += progress;
+            progress = 0;
+        }
+    }
+
+    private void endPrevious()
+    {
+        updateProgress( goal ); // previous ended
+        // TODO print some end stats for this stage?
```
MishaDemianenko, line 229: TODOs here and below

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/HumanUnderstandableExecutionMonitor.java
```diff
@@ -0,0 +1,466 @@
...
+import static org.neo4j.helpers.Format.date;
+import static org.neo4j.helpers.Format.duration;
+import static org.neo4j.helpers.collection.Iterables.last;
+import static org.neo4j.unsafe.impl.batchimport.cache.GatheringMemoryStatsVisitor.totalMemoryUsageOf;
+
+/**
+ * Prints progress you can actually understand, with capabilities to on demand print completely incomprehensible
+ * details only understandable to a select few.
+ */
+public class HumanUnderstandableExecutionMonitor implements ExecutionMonitor
```
MishaDemianenko, line 59: would be nice to have tests for this class to make sure we report progress correctly

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/OnDemandDetailsExecutionMonitor.java
```diff
@@ -0,0 +1,279 @@
...
+            details.get( details.size() - 1 ).print( out );
+        }
+    }
+
+    private static void printIndented( PrintStream out, String string )
+    {
+        out.println( "\t" + string );
+    }
+
+    private void triggerGC()
```
MishaDemianenko: this is useless action right since we disable explicit GC everywhere?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/OnDemandDetailsExecutionMonitor.java
```diff
@@ -0,0 +1,279 @@
...
+ *     <th>Command</th>
+ *     <th>Description</th>
+ *   </tr>
+ *   <tr>
+ *     <th>i</th>
+ *     <th>Print {@link SpectrumExecutionMonitor compact information} about each executed stage up to this point</th>
+ *   </tr>
+ * </table>
+ */
+public class OnDemandDetailsExecutionMonitor implements ExecutionMonitor
```
MishaDemianenko, line 67: can't find tests for this guy as well

MishaDemianenko (commented):
Couple of mostly minor comments here and there

community/csv/src/main/java/org/neo4j/csv/reader/MultiReadable.java
tinwelint: Only one reader at a time is open

community/kernel/src/main/java/org/neo4j/helpers/Format.java
tinwelint: https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes as you can see they aren't all the same case. Perhaps we should just conform to that instead, which means it should be:

`"", "k", "M", "G", "T"`

community/kernel/src/main/java/org/neo4j/helpers/Format.java
tinwelint: There will be some day. Although I see no point in removing it to save a few characters in the code, do you?

community/kernel/src/main/java/org/neo4j/kernel/impl/cache/MeasureDoNothing.java
tinwelint: It's something of a rough indicator of amount of time spent in Full GC

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/HumanUnderstandableExecutionMonitor.java
tinwelint: That will be a bit difficult, but can try.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInput.java
tinwelint: I think it's fine as it is... will remove it.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/OnDemandDetailsExecutionMonitor.java
tinwelint: `disable explicit GC everywhere` is that a strict policy implemented somewhere, if so where? With that said it has been useful while testing and may be useful for others, but it's likely not for most cases.

community/csv/src/test/java/org/neo4j/csv/reader/BufferedCharSeekerTest.java
tinwelint: It's assigned in the constructor and used in `length()` isn't it?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/OnDemandDetailsExecutionMonitor.java
tinwelint: There's a refactoring ahead of this PR in the making, which pulls out the functionality about reacting to user input... which really most of this class is about. Testing could be added to that instead IMHO.

community/csv/src/main/java/org/neo4j/csv/reader/MultiReadable.java
MishaDemianenko: ok if we rely on the fact that those readable that we pass into constructor do not need to be closed then ok

community/csv/src/test/java/org/neo4j/csv/reader/BufferedCharSeekerTest.java
MishaDemianenko: position filed? no, we only return it from  `public long position()` method right?

community/kernel/src/main/java/org/neo4j/helpers/Format.java
MishaDemianenko: ok sounds good

community/kernel/src/main/java/org/neo4j/helpers/Format.java
MishaDemianenko: 'TB' is not a valid count size prefix right? (and inside suffixCount you can see that it will be used for case when number is big)

community/kernel/src/main/java/org/neo4j/kernel/impl/cache/MeasureDoNothing.java
MishaDemianenko: pauses of those threads can have nothing to do with GC in general and with full GC in particular but lets keep it away from current pr

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/OnDemandDetailsExecutionMonitor.java
MishaDemianenko: ok

MishaDemianenko:
@tinwelint 2 small comments about position field and `TB` suffix

community/csv/src/main/java/org/neo4j/csv/reader/MultiReadable.java
tinwelint: The thing passed into the constructor provides them and opens each in `next()`, so yes

community/kernel/src/main/java/org/neo4j/helpers/Format.java
tinwelint: `k`, `M`, `G`, `T`... so you mean `T` don't you? And why is that not OK?

community/csv/src/test/java/org/neo4j/csv/reader/BufferedCharSeekerTest.java
tinwelint: Now I get it... you _meant_ to click the line containing `position`, whereas you clicked the line containing `data`. True, `position` can be removed.

community/kernel/src/main/java/org/neo4j/helpers/Format.java
MishaDemianenko: no no, 'T' is fine, but if you will go into `suffixCount` method and if `value` is big enough  you will unconditionally use `TB` suffix (line 130)

MishaDemianenko (approved):

PR #10550, opened 2017-11-30
Introduce volumetric check-pointing, where the check-point frequency is adjusted to avoid the transaction log retention policy falling too far behind.

chrisvest:


enterprise/kernel/src/main/resources/META-INF/services/org.neo4j.kernel.impl.transaction.log.checkpoint.CheckPointThresholdPolicy
```diff
@@ -0,0 +1,2 @@
+org.neo4j.kernel.impl.enterprise.transaction.log.checkpoint.VolumetricCheckPointPolicy
+org.neo4j.kernel.impl.enterprise.transaction.log.checkpoint.ContinuousThresholdPolicy
```
MishaDemianenko, line 2: VolumetricCheckPointPolicy is missing here, either all of them should be added into services file, either none

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -341,6 +341,17 @@
...
+    @Description( "Configures the general policy for when check-points should occur. The default policy is to " +
+                  "check-point 'periodically', as specified by the 'dbms.checkpoint.interval.tx' and " +
+                  "'dbms.checkpoint.interval.time' settings. " +
+                  "The Neo4j Enterprise Edition provides two alternative policies: " +
+                  "The first is to check-point 'continuously', which will ignore those settings and run the " +
+                  "check-point process all the time. " +
+                  "The second is the 'volumetric' policy, which makes a best-effort at check-pointing often enough " +
+                  "so that the database doesn't get too far behind on deleting old transaction logs in accordance " +
+                  "with the 'dbms.tx_log.rotation.retention_policy' setting." )
+    public static final Setting<String> check_point_policy = setting( "dbms.checkpoint", STRING, "periodically" );
```
MishaDemianenko: it looks like options should be 'volumetric' and 'periodic' instead

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointThreshold.java
```diff
@@ -45,10 +56,92 @@
...
+     * {@link #isCheckPointingNeeded(long, Consumer)}, if this {@link CheckPointThreshold} instance has any opinion on
+     * the matter, or return {@link LongStream#empty()} is fine with some default checking frequency.
+     * <p>
+     * This is returned as an {@link LongStream} because a threshold might be composed of multiple other thresholds.
+     * It is up to the caller to figure out how to best schedule this threshold, if the stream contains more than one
+     * frequency. One way could be to use the lowest frequency, e.g. with {@link LongStream#min()}.
+     *
+     * @return A stream desired scheduling frequencies, if any specific ones are desired by this threshold.
+     */
+    LongStream checkFrequencyMillis();
```
MishaDemianenko: why result of this call in a LongStream and not just long? we do not use this as a source of frequencies now at all
and not quite clear why this should be more complex than it's already is

community/kernel/src/main/resources/META-INF/services/org.neo4j.kernel.impl.transaction.log.checkpoint.CheckPointThresholdPolicy
```diff
@@ -0,0 +1 @@
+org.neo4j.kernel.impl.transaction.log.checkpoint.PeriodicThresholdPolicy
```
MishaDemianenko, line 1: same here with *Policy services either all of them should be here either none

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruningImpl.java
```diff
@@ -31,15 +35,41 @@
...
+    private LongStream getLogVersionsToDelete( long upToVersion )
+    {
+        // We synchronise on the pruneStrategy because it's stateful and not thread safe, and even though we have the
+        // pruneLock, we don't want mightHaveLogsToPrune to use it because it could make pruneLogs think that pruning
+        // is taking place, when really we were just checking.
+        // The pruneLock is used to guard against actual pruning happening concurrently, while synchronising on the
+        // strategy is done to protect the internal state of the strategy.
+        synchronized ( pruneStrategy )
+        {
+            return pruneStrategy.findLogVersionsToDelete( upToVersion );
```
MishaDemianenko: Instead of this long comment why not create abstract class of pruneStretagy with synchronized findLogVersionsToDelete that will invoke user-provided implementation of actual find method?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruningImpl.java
```diff
@@ -31,15 +35,41 @@
...
+                           PhysicalLogFiles logFiles,
+                           LogProvider logProvider )
     {
+        this.fs = fs;
         this.pruneStrategy = pruneStrategy;
+        this.logFiles = logFiles;
         this.msgLog = logProvider.getLog( getClass() );
     }
 
+    private LongStream getLogVersionsToDelete( long upToVersion )
```
MishaDemianenko: what is the reason to return long stream here that was created based on transaction log files and convert them back into a log file in the next call?
It looks like this whole machinery can be simplified a lot if we will pass "LogVersionVisitor" into PhysicalLogFiles that will simply deletion if satisfied without any intermediate states

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruning.java
```diff
@@ -24,7 +24,31 @@
...
     void pruneLogs( long currentVersion );
+
+    /**
+     * Check if there might be a desire to prune logs. This could be used as a hint to schedule some log pruning soon,
+     * and/or increase the check pointing frequency.
+     *
+     * @return {@code true} if calling {@link #pruneLogs(long)} now <em>might</em> cause log files to be deleted.
+     * Otherwise {@code false} if we are pretty sure that we don't need to prune any logs right now.
+     */
+    boolean mightHaveLogsToPrune();
```
MishaDemianenko, line 39: not quite sure why do we need this method here. It looks like it was introduced for VolumetricCheckPointThreshold and should be able to decide if its threshold reached or not just based on "PhysicalLogFiles" and not clear why LogPruningImpl have that suggestion method since that is what out thresholds should be doing

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/PeriodicThresholdPolicy.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public CheckPointThreshold createThreshold( Config config, Clock clock, LogProvider logProvider )
+    {
+        int txThreshold = config.get( GraphDatabaseSettings.check_point_interval_tx );
+        final CountCommittedTransactionThreshold countCommittedTransactionThreshold =
+                new CountCommittedTransactionThreshold( txThreshold );
+
+        long timeMillisThreshold = config.get( GraphDatabaseSettings.check_point_interval_time ).toMillis();
```
tinwelint, line 52: Could you perhaps use the `CheckPointThreshold.createTreshold` method introduced in one of your previous commits here instead?

tinwelint (approved):
Looks dandy to me, just one small comment about potential duplicated code.

MishaDemianenko (changes requested):

enterprise/kernel/src/main/resources/META-INF/services/org.neo4j.kernel.impl.transaction.log.checkpoint.CheckPointThresholdPolicy
chrisvest: You mean the periodic one is missing?

community/kernel/src/main/resources/META-INF/services/org.neo4j.kernel.impl.transaction.log.checkpoint.CheckPointThresholdPolicy
chrisvest: This file is in community, which only has the periodic policy.

enterprise/kernel/src/main/resources/META-INF/services/org.neo4j.kernel.impl.transaction.log.checkpoint.CheckPointThresholdPolicy
MishaDemianenko: Yes correct, both comments about service files are invalid and you can ignore them

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
chrisvest: They are aliases.

MishaDemianenko: they are aliases in threshold classes but they should be aligned here in a property description - since those are the values that user will gonna use. You need to be insane to go to threshold classes to get all the possible values that you can use

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/PeriodicThresholdPolicy.java
chrisvest: The `CheckPointThreshold.createTreshold` method delegates to the `createThreshold` methods on the policy objects.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointThreshold.java
chrisvest: At first I was using `OptionalLong`, but that API has no convenient way of reducing the multiple options that become available inside `CheckPointThreshold.or`.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruning.java
chrisvest: The responsibility of how and when to prune logs lies with `LogPruning`. It's the only thing that knows how to figure out if we are getting behind on log pruning or not. The volumetric check point threshold relies on it for this information. It is also the owner of the pruning threshold. The alternative is to instantiate our own `LogPruneStrategy`, and then maintain its state in parallel with the instance in `LogPruning`.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruningImpl.java
chrisvest: Because this was a smaller change to the `LogPruneStrategy` implementations. The log files have to be visited in oldest-to-newest order, but the log version brackets and the threshold computation is currently done in newest-to-oldest order. So to do this, we'll have to rewrite `ThresholdBasedPruneStrategy`.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruningImpl.java
chrisvest: I think the `ThresholdBasedPruneStrategy` is the only implementation that matters. We can just synchronise on the method there, and then document on the interface that the strategy has to be thread-safe.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CheckPointThreshold.java
MishaDemianenko: why even optional long instead of just long?
now you have reduction phase on a later step where you take min instead you can move your min function inside 'CheckPointThreshold.or' and use plain long in all the places right?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruning.java
MishaDemianenko: It's not about the responsibility of `how` it's about `when`. LogPruning aware only how to delete log files. All `when` responsibilities are hidden in different thresholds. How LogPrunning can know anything about external thresholds and can suggest something to them? It up to thresholds to answer those questions. And to be able to answer those questions they shouldn' require thing that prune files

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruningImpl.java
MishaDemianenko: hm..don't know how bit that change is especially taking into account that ThresholdBasedPruneStrategy was already changed to return stream instead. But can live with this stream ok.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruningImpl.java
MishaDemianenko: ok let's do that only thing synchronised

MishaDemianenko (approved):

PR #10552, opened 2017-11-30
Forbidding unsupported planner/version/runtime options

sherfert:
Specifying 3.3 rule / 3.4 rule / slotted rule / compiled rule
will fail now.


Lojjs (approved):
I think we should add a test for the 3.3 - rule combination once this and "33434" is merged.
Otherwise it looks fine!

PR #10553, opened 2017-11-30
Refactor reused ids arrays allocation

MishaDemianenko:
Avoid multiple arrays creation during id allocations.
Allow IdContainer and FreeIdKeeper to allocate and return precise array
with reusable ids.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/id/FreeIdKeeper.java
@@ -25,10 +25,12 @@ import java.nio.ByteBuffer;
 import java.util.ArrayDeque;
 import java.util.Deque;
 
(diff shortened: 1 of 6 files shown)


tinwelint (commented):
Whereas you're reducing allocation from 2 to 1 long[] while getting a batch, which is not bad... I believe the even bigger litterer is the `Deque<Long>` in `FreeIdKeepper` isn't it? Isn't your change completely dwarfed by that?

tinwelint:
Also there are some new failing tests, all problems rotating the counts store and we all know that such errors usually are second failures hiding some real problem. You think your changes could have resulted in those failures?

MishaDemianenko:
@tinwelint those test failures are not only this PR failures those tests fail everywhere now and they related to latest CC work

MishaDemianenko:
@tinwelint also on a note of dequeuing primitives: yes you right that deq will be replaced with specialized primitive version i just wanted to keep that work separate from this PR

tinwelint (approved):
Wonderful

PR #10558, opened 2017-12-01
Forbidding unsupported planner/version/runtime options

OliviaYtterbrink:
Specifying 3.2 rule / 3.3 rule / slotted rule / compiled rule will fail now. Back port of #10552

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
@@ -524,12 +524,14 @@ object CypherComparisonSupport {
 
     def Interpreted: TestConfiguration =
       TestConfiguration(Versions.Default, Planners.Default, Runtimes(Runtimes.Interpreted, Runtimes.Slotted)) +
-        TestConfiguration(Versions.V2_3 -> Versions.V3_2, Planners.all, Runtimes.Default) +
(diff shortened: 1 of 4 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/HelpfulErrorMessagesTest.scala
```diff
@@ -52,6 +52,18 @@ class HelpfulErrorMessagesTest extends ExecutionEngineFunSuite with CypherCompar
       "The given query is not currently supported in the selected cost-based planner"))
   }
 
+  test("should provide sensible error message for 3.3 rule planner") {
+    intercept[Exception](graph.execute("CYPHER 3.3 planner=rule RETURN 1")).getMessage should be("Unsupported PLANNER - VERSION combination: rule - 3.3")
+  }
+
```
Lojjs, line 58: @OliviaYtterbrink  I think we should have a test for the combination rule-3.2 here as well

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -524,12 +524,14 @@ object CypherComparisonSupport {
```
Lojjs, line 524: Nice work here with all the configurations! Looks correct as far as I can see :)

Lojjs (changes requested):
I miss a test, otherwise good work

Lojjs (approved):

PR #10560, opened 2017-12-01
Don't use offheap collections in compiled runtime

fickludd:
Stop compiled runtime from using off heap data structures. This was a remnant from the compiled runtime prototyping that should have been removed, now it is.

Also added an on-heap primitive long-long map, which the lack of was probably the reason for the off-heaping.


chrisvest:
Is this not also a problem in 3.3?

fickludd:
@chrisvest yes, moving this over to `3.2` actually.

PR #10561, opened 2017-12-01
Configuration parameter for db.index.explicit.forNodes

sherfert:
db.index.explicit.forNodes now accepts a second paramater "configuration", that can be used to configure the index. This new parameter can be omitted.


sarmbruster:
LGTM

Lojjs (approved):
Looks good and now builds locally!

PR #10562, opened 2017-12-01
No offheap collection in birk

fickludd:
Stop compiled runtime from using off heap data structures. This was a remnant from the compiled runtime prototyping that should have been removed, now it is.

Also added an on-heap primitive long-long map, which the lack of was probably the reason for the off-heaping.

### DIFF
--- a/community/primitive-collections/src/test/java/org/neo4j/collection/primitive/hopscotch/BasicTableTest.java
+++ b/community/primitive-collections/src/test/java/org/neo4j/collection/primitive/hopscotch/BasicTableTest.java
@@ -153,6 +153,26 @@ public class BasicTableTest
             @Override
             public Table newTable( int capacity )
             {
+                return new LongKeyLongValueTable( capacity );
+            }
+
+            @Override
+            public boolean supportsLongs()
+            {
+                return true;
+            }
+
+            @Override
+            public Object sampleValue()
+            {
+                return new long[] {Math.abs( random.nextLong() )};
+            }
+        } } );
+        result.add( new Object[] { new TableFactory()
+        {
+            @Override
+            public Table newTable( int capacity )
+            {
                 return new LongKeyObjectValueTable( capacity );
             }
 
--- a/community/primitive-collections/src/test/java/org/neo4j/collection/primitive/hopscotch/PrimitiveLongMapTest.java
+++ b/community/primitive-collections/src/test/java/org/neo4j/collection/primitive/hopscotch/PrimitiveLongMapTest.java
@@ -939,7 +939,7 @@ public class PrimitiveLongMapTest
     {
         // GIVEN
         PrimitiveLongLongVisitor<RuntimeException> visitor;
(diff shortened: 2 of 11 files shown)


community/primitive-collections/src/test/java/org/neo4j/collection/primitive/hopscotch/PrimitiveLongMapTest.java
```diff
@@ -939,7 +939,7 @@ public void longLongEntryVisitorShouldSeeAllEntriesIfItDoesNotBreakOut()
     {
         // GIVEN
         PrimitiveLongLongVisitor<RuntimeException> visitor;
-        try ( PrimitiveLongLongMap map = Primitive.offHeapLongLongMap() )
+        try ( PrimitiveLongLongMap map = Primitive.longLongMap() )
```
pontusmelke, line 942: Maybe we should have two separate tests instead of just changing the old one?

pontusmelke (approved):
👍 Had one small comment, do with it what you wish!

community/primitive-collections/src/test/java/org/neo4j/collection/primitive/hopscotch/BasicTableTest.java
```diff
@@ -149,6 +149,26 @@ public Object sampleValue()
...
+            @Override
+            public boolean supportsLongs()
+            {
+                return true;
+            }
+
+            @Override
+            public Object sampleValue()
+            {
+                return new long[] {Math.abs( random.nextLong() )};
```
chrisvest, line 168: Why the `Math.abs`? The `LongKeyIntValueTable` above doesn't do this. I guess you were following the steps of the off-heap table?

chrisvest (approved):
Good stuff. One question popped up, but I suppose it's not your fault, so it doesn't matter.

community/primitive-collections/src/test/java/org/neo4j/collection/primitive/hopscotch/PrimitiveLongMapTest.java
fickludd: I have added duplicates below, it's just the git diff that looks wonky.

community/primitive-collections/src/test/java/org/neo4j/collection/primitive/hopscotch/BasicTableTest.java
fickludd: I also though that was strange, but actually the `LongKeyIntValueTable` also does this, because of the `random.nextInt( Integer.MAX_VALUE )`. Not allowing negative integers is a bit surprising. 

Don't know if it's related to the magical unicorn `LongKeyIntValueTable#NULL = -1`?

chrisvest: That's probably it.

PR #10563, opened 2017-12-01
Adding transaction timeout support for explicit transactions in bolt

zhenlineo:


community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
```diff
@@ -47,6 +49,7 @@
     private static final Pattern BEGIN = Pattern.compile("(?i)^\\s*BEGIN\\s*;?\\s*$");
     private static final Pattern COMMIT = Pattern.compile("(?i)^\\s*COMMIT\\s*;?\\s*$");
     private static final Pattern ROLLBACK = Pattern.compile("(?i)^\\s*ROLLBACK\\s*;?\\s*$");
+    private static final String TX_TIMEOUT_MILLIS = "txTimeout";
```
lutovich, line 53: Maybe "timeout" instead of "txTimeout"? We already have "bookmark" and "bookmarks" without "tx" prefix.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
```diff
@@ -196,7 +199,16 @@ State run( MutableTransactionState ctx, SPI spi, String statement,
                     {
                         if ( BEGIN.matcher( statement ).matches() )
                         {
-                            ctx.currentTransaction = spi.beginTransaction( ctx.securityContext );
+                            AnyValue txTimeoutMills = params.get( TX_TIMEOUT_MILLIS );
+                            if ( txTimeoutMills instanceof LongValue )
```
lutovich: I think it is better to do `instanceof NumberValue` and then use `NumberValue#longValue()` method. JS only has a single `Number` type which we pack as float.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachine.java
```diff
@@ -196,7 +199,16 @@ State run( MutableTransactionState ctx, SPI spi, String statement,
...
+                            {
+                                ctx.currentTransaction = spi.beginTransaction( ctx.securityContext,
+                                        ((LongValue) txTimeoutMills).value(), TimeUnit.MILLISECONDS );
+                            }
+                            else
+                            {
+                                ctx.currentTransaction = spi.beginTransaction( ctx.securityContext );
```
lutovich: I think we should only go here if `txTimeoutMills == Values.NO_VALUE`. Exception should be thrown if it's not a `NO_VALUE` and not an instance of `NumberValue`

lutovich (commented):
@zhenlineo review completed, made couple comments.

zhenlineo:
Feature withdrawn

PR #10564, opened 2017-12-01
Minor optimizations of Kernel API implementation

fickludd:
Improves performance of most basic Kernel API reads by some cheap profile driven optimizations:

 - Fetch stores once in `StoreStatement`
 - Reuse the `LabelCursor` in `NodeCursor` between initializations
 - Faster happy path transaction termination checks (without `Optional`).

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
@@ -62,7 +62,10 @@ class NodeCursor extends NodeRecord implements org.neo4j.internal.kernel.api.Nod
         this.next = 0;
         this.highMark = read.nodeHighMark();
         this.read = read;
-        this.labelCursor = read.labelCursor();
+        if ( labelCursor != null )
+        {
+            labelCursor = read.labelCursor();
+        }
     }
 
     void single( long reference, Read read )
@@ -79,7 +82,10 @@ class NodeCursor extends NodeRecord implements org.neo4j.internal.kernel.api.Nod
         //This marks the cursor as a "single cursor"
         this.highMark = NO_ID;
         this.read = read;
-        this.labelCursor = read.labelCursor();
+        if ( labelCursor != null )
+        {
+            labelCursor = read.labelCursor();
+        }
     }
 
     @Override
@@ -274,8 +280,11 @@ class NodeCursor extends NodeRecord implements org.neo4j.internal.kernel.api.Nod
             pageCursor = null;
         }
         read = null;
-        labelCursor = null;
-
+        if ( labelCursor != null )
+        {
+            labelCursor.close();
+            labelCursor = null;
+        }
         reset();
     }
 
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -58,7 +58,10 @@ void scan( Read read )
         this.next = 0;
         this.highMark = read.nodeHighMark();
         this.read = read;
-        this.labelCursor = read.labelCursor();
+        if ( labelCursor != null )
```
pontusmelke: I might be missing something here but shouldn't it be `labelCursor == null`

pontusmelke (changes requested):
Nice, I may be wrong but I had one comment that I at least would like to understand?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
tinwelint: Sure looks like it, yeah

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/StoreStatement.java
```diff
@@ -331,123 +338,119 @@ public Properties properties()
...
         {
-            return neoStores.getPropertyStore().getStringStore().openPageCursorForReading( reference );
+            return propertyStore.getStringStore().openPageCursorForReading( reference );
         }
 
         @Override
         public PageCursor openArrayPageCursor( long reference )
         {
-            return neoStores.getPropertyStore().getArrayStore().openPageCursorForReading( reference );
+            return propertyStore.getArrayStore().openPageCursorForReading( reference );
```
tinwelint: It'd sort of make sense to keep `propertyStore.getArrayStore()` as a field too.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/StoreStatement.java
```diff
@@ -331,123 +338,119 @@ public Properties properties()
...
         }
 
         @Override
         public ByteBuffer loadString( long reference, ByteBuffer buffer, PageCursor page )
         {
-            return readDynamic( neoStores.getPropertyStore().getStringStore(), reference, buffer, page );
+            return readDynamic( propertyStore.getStringStore(), reference, buffer, page );
```
tinwelint, line 453: It'd sort of make sense to keep `propertyStore.getStringStore()` as a field too.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/StoreStatement.java
fickludd: I didn't put the array and string stores in fields because

1) they're accessed less often, and when they are accessed the access will be costly, so the relative cost of getting the store will be small.
2) The getter methods are just access methods to final fields and should be extremely cheap. The `getNodeStore()` seems to be harder for the compiler to optimize away, don't ask me why. 

Do you think I should fieldify anyway?

pontusmelke (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/StoreStatement.java
tinwelint: I think you should, if nothing else then at least for consistency?

fickludd: 👍

PR #10565, opened 2017-12-03
Refactorings around memory allocation in the page cache.

chrisvest:
* The memory allocated for the page list is now accounted as part of `dbms.memory.pagecache.size`.
* The page cache memory setting is now a string that `ByteUnit` is able to parse. Making it a string allows more versatile configurations in the future.
* The ability to specify a cache page size is more restricted in APIs, but still available because it is still used, e.g. in many GBPTree tests.


community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
```diff
@@ -72,6 +72,18 @@ public synchronized long sumUsedMemory()
         return sum;
     }
 
+    @Override
+    public synchronized long availableMemory()
```
ragadeeshu: In regards to usedMemory and availableMemory, it feels like one could use the other, keeping the calculation that is simplest? If this isn't true, i.e. if the available and used memory has no such relationship, I feel like this should be elaborated upon in the javadoc.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/StandalonePageCacheFactory.java
```diff
@@ -43,21 +43,16 @@ private StandalonePageCacheFactory()
...
-
-    public static PageCache createPageCache( FileSystemAbstraction fileSystem, Integer pageSize,
+    public static PageCache createPageCache( FileSystemAbstraction fileSystem,
             PageCacheTracer tracer, PageCursorTracerSupplier cursorTracerSupplier )
     {
         SingleFilePageSwapperFactory factory = new SingleFilePageSwapperFactory();
         factory.open( fileSystem, Configuration.EMPTY );
 
-        int cachePageSize = pageSize != null ? pageSize : factory.getCachePageSizeHint();
+        int cachePageSize = factory.getCachePageSizeHint();
```
ragadeeshu: 👍

community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
```diff
@@ -79,25 +72,29 @@ public synchronized long availableMemory()
...
         {
-            availableInCurrentGrab = grab.limit - grab.nextAlignedPointer;
+            availableInCurrentGrab = grab.limit - grab.nextPointer;
         }
         return Math.max( memoryReserve, 0L ) + availableInCurrentGrab;
     }
 
     @Override
-    public synchronized long allocateAligned( long bytes )
+    public synchronized long allocateAligned( long bytes, long alignment )
```
ragadeeshu: I'm a bit curious about an application when you need this feature :) (allocation-wise alignment)

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCache.java
```diff
@@ -214,14 +222,14 @@ public MuninnPageCache(
...
-            int maxPages,
+            MemoryAllocator memoryAllocator,
             int cachePageSize,
             PageCacheTracer pageCacheTracer,
             PageCursorTracerSupplier pageCursorTracerSupplier )
     {
         verifyHacks();
         verifyCachePageSizeIsPowerOfTwo( cachePageSize );
-        verifyMinimumPageCount( maxPages, cachePageSize );
+        int maxPages = calculatePageCount( memoryAllocator, cachePageSize );
```
ragadeeshu, line 259: awesome

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCache.java
```diff
@@ -257,15 +261,22 @@ private static void verifyCachePageSizeIsPowerOfTwo( int cachePageSize )
...
         if ( maxPages < minimumPageCount )
         {
             throw new IllegalArgumentException( String.format(
                     "Page cache must have at least %s pages (%s bytes of memory), but was given %s pages.",
-                    minimumPageCount, minimumPageCount * cachePageSize, maxPages ) );
+                    minimumPageCount, minimumPageCount * memoryPerPage, maxPages ) );
+        }
+        if ( maxPages > Integer.MAX_VALUE )
+        {
+            return Integer.MAX_VALUE;
```
ragadeeshu: Maybe this value should be fetched from the page cache meta data?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -525,9 +524,8 @@
                   "JVM enough heap to hold all your transaction state and query context, and then leave the rest for " +
                   "the page cache. If no page cache memory is configured, then a heuristic setting is computed based " +
                   "on available system resources." )
-    public static final Setting<Long> pagecache_memory =
-            buildSetting( "dbms.memory.pagecache.size", BYTES, null)
-                    .constraint( min( PageCache.PAGE_SIZE * 30L ) ).build();
+    public static final Setting<String> pagecache_memory =
+            buildSetting( "dbms.memory.pagecache.size", STRING, null ).build();
```
ragadeeshu: Is it ok to change property surface 3.3 > 3.4?

ragadeeshu (commented):
@chrisvest I love these changes. Some minor points where I would like clarifications :)

ragadeeshu:
It seems like a lot of stuff was listed as outdated. Please do take a look despite that :)

community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
chrisvest: IO buffers might require different alignment than other allocations.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
chrisvest: I'm not sure. The only public API usage of Setting that I can think of is `GraphDatabaseBuilder.setConfig`, where the field is just passed through. The alternative is to have two settings that does the same thing but differ in type. I don't know if they can both be called `dbms.memory.pagecache.size` in that case. If they can't, then I suspect it would cause more confusion.

community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
chrisvest: Available memory counts the reserve (if any is left) plus whatever is left in the current grab, while used memory is the sum of memory allocated. So they don't really have a relationship like that.

chrisvest:
@ragadeeshu I have addressed your review comments.

community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
ragadeeshu: Then, is it really safe to verify your tests with `assertThat( mman.usedMemory() + mman.availableMemory(), is( 8192L ) );` ?

ragadeeshu (approved):
Then, is it really safe to verify your tests with assertThat( mman.usedMemory() + mman.availableMemory(), is( 8192L ) ); ?

community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
chrisvest: Hmm… 🤔

chrisvest:
@ragadeeshu Well spotted. I have addressed your comment.

ragadeeshu:
@chrisvest LGTM

chrisvest:
Rebased to see if we can get a greener build.

PR #10567, opened 2017-12-04
Test hardening

chrisvest:
Impermanent test graph databases no longer uses the "global" store locker.


community/consistency-check/src/test/java/org/neo4j/consistency/ConsistencyCheckServiceIntegrationTest.java
```diff
@@ -89,7 +89,7 @@ protected void generateInitialData( GraphDatabaseService graphDb )
         }
     };
 
-    private final TestDirectory testDirectory = TestDirectory.testDirectory();
+    private final TestDirectory testDirectory = TestDirectory.testDirectory( ConsistencyCheckServiceIntegrationTest.class );
```
MishaDemianenko: Why do you need to pass class explicitly in all those tests? TestDirectory rule will gonna get it from description object that JUnit runner will supply?

chrisvest: Oh, right. It's a completely different reason that collusions on directories were possible.

MishaDemianenko (approved):

PR #10570, opened 2017-12-04
Fix selection regression in compiled runtime

pontusmelke:
We have seen a regression in the compiled runtime for selections since https://github.com/neo4j/neo4j/pull/10411. Turns out it is due to the extra `instance_of` checks introduced in that PR. This PR fixes that regression, plus removes some compiler warnings from the cypher codebase.


fickludd (approved):
Very nice.

PR #10571, opened 2017-12-04
Adding more useful toString() to Cursors

SaschaPeukert:


pontusmelke:
@SaschaPeukert needs rebase

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/KernelAPIReadTestBase.java
```diff
@@ -44,6 +48,27 @@
 @SuppressWarnings( "WeakerAccess" )
 public abstract class KernelAPIReadTestBase<ReadSupport extends KernelAPIReadTestSupport>
 {
+    public class ConditionalTeardown implements TestRule
+    {
+        public Statement apply( Statement base, Description description )
+        {
+            return statement( base, description );
```
pontusmelke: inline?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/KernelAPIReadTestBase.java
```diff
@@ -44,6 +48,27 @@
 @SuppressWarnings( "WeakerAccess" )
 public abstract class KernelAPIReadTestBase<ReadSupport extends KernelAPIReadTestSupport>
 {
+    public class ConditionalTeardown implements TestRule
```
pontusmelke: How about calling this class something like `CursorsClosedPostCondition`? Since that is what it does.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/KernelAPIReadTestBase.java
pontusmelke: This can be inlined

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/KernelAPIWriteTestBase.java
```diff
@@ -43,12 +47,36 @@
 @SuppressWarnings( "WeakerAccess" )
 public abstract class KernelAPIWriteTestBase<WriteSupport extends KernelAPIWriteTestSupport>
 {
+    public class ConditionalTeardown implements TestRule
```
pontusmelke: This is an exact copy from the other test case. We really should extract it to its own utility (see other comment for the name of that utility).

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeExplicitIndexCursor.java
```diff
@@ -96,4 +96,18 @@ public boolean isClosed()
...
+    public String toString()
+    {
+        if ( isClosed() )
+        {
+            return "NodeExplicitIndexCursor[closed state]";
+        }
+        else
+        {
+            return "NodeExplicitIndexCursor[node=" + node + ", expectedSize=" + expectedSize + ", score=" + score +
+                    " ,underlying record=" + super.toString() + " ]";
```
pontusmelke: Tiny cosmetics: space after comma.

pontusmelke (changes requested):
Just some small things.

pontusmelke (approved):

pontusmelke:
@SaschaPeukert something missing in the tests that made a lot of test fail with a `NullPointerException`

PR #10574, opened 2017-12-04
Monitoring of Raft message processing performance

andrewkerr9000:
Delay is difference between unmarshalling of message and start of
processing.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/RaftServerModule.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/RaftServerModule.java
@@ -110,7 +110,6 @@ public class RaftServerModule
         return clusterBindingHandler
                 .compose( leaderAvailabilityHandler )
                 .compose( batchingMessageHandler )
-                .compose( clusterBindingHandler )
                 .compose( monitoringHandler )
                 .apply( messageApplier );
     }
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/RaftServerModule.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/RaftServerModule.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.causalclustering.core;
+
+import java.util.function.Function;
+
+import org.neo4j.causalclustering.catchup.storecopy.LocalDatabase;
+import org.neo4j.causalclustering.core.consensus.ConsensusModule;
+import org.neo4j.causalclustering.core.consensus.ContinuousJob;
+import org.neo4j.causalclustering.core.consensus.LeaderAvailabilityHandler;
+import org.neo4j.causalclustering.core.consensus.RaftMessageMonitoringHandler;
+import org.neo4j.causalclustering.core.consensus.RaftMessages.ReceivedInstantClusterIdAwareMessage;
+import org.neo4j.causalclustering.core.consensus.RaftServer;
+import org.neo4j.causalclustering.core.server.CoreServerModule;
+import org.neo4j.causalclustering.core.state.RaftMessageApplier;
+import org.neo4j.causalclustering.identity.MemberId;
+import org.neo4j.causalclustering.logging.MessageLogger;
+import org.neo4j.causalclustering.messaging.ComposableMessageHandler;
+import org.neo4j.causalclustering.messaging.CoreReplicatedContentMarshal;
+import org.neo4j.causalclustering.messaging.LifecycleMessageHandler;
+import org.neo4j.causalclustering.messaging.LoggingInbound;
+import org.neo4j.kernel.impl.factory.PlatformModule;
+import org.neo4j.kernel.impl.util.JobScheduler;
+import org.neo4j.kernel.monitoring.Monitors;
+import org.neo4j.logging.LogProvider;
+import org.neo4j.ssl.SslPolicy;
+
+import static org.neo4j.kernel.impl.util.JobScheduler.SchedulingStrategy.NEW_THREAD;
+
+public class RaftServerModule
+{
+    private final PlatformModule platformModule;
+    private final ConsensusModule consensusModule;
+    private final IdentityModule identityModule;
+    private final LocalDatabase localDatabase;
+    private final SslPolicy sslPolicy;
+    private final Monitors monitors;
+    private final MessageLogger<MemberId> messageLogger;
+    private final LogProvider logProvider;
+
+    RaftServerModule( PlatformModule platformModule, ConsensusModule consensusModule, IdentityModule identityModule, LocalDatabase localDatabase,
+            SslPolicy sslPolicy, Monitors monitors, MessageLogger<MemberId> messageLogger )
+    {
+        this.platformModule = platformModule;
+        this.consensusModule = consensusModule;
+        this.identityModule = identityModule;
+        this.localDatabase = localDatabase;
+        this.sslPolicy = sslPolicy;
+        this.monitors = monitors;
+        this.messageLogger = messageLogger;
+        this.logProvider = platformModule.logging.getInternalLogProvider();
+    }
+
+    public void create( CoreServerModule coreServerModule )
+    {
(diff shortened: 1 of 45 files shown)


andrewkerr9000:
Diff is long but simple.

I experimented with a message wrapper (similar to `ClusterIdAwareMessage`) to include the `receivedAt` Instant, but that was looking uglier.

Passing `Clock` instances around is not strictly necessary.

andrewkerr9000:
There may be some room for improvement with timer, e.g. a different named timer for each message type.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
```diff
@@ -1017,13 +1032,17 @@ public int hashCode()
 
     abstract class BaseRaftMessage implements RaftMessage
     {
-        protected MemberId from;
-        private Type type;
+        protected final MemberId from;
+        private final Type type;
+        private final Clock clock;
```
martinfurmanski: I might be able to live with the recievedAt, but not the clock. Perhaps you can add another handler in the Netty-pipeline which fills in recievedAt?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMachine.java
```diff
@@ -280,6 +308,24 @@ public synchronized ConsensusOutcome handle( RaftMessages.RaftMessage incomingMe
...
+    {
+        long delay = Duration.between( Instant.now( clock ), message.receivedAt() ).toMillis();
+
+        raftMessageDelayMonitor.setDelay( delay );
+
+        long heartbeatInterval = leaderAvailabilityTimers.getHeartbeatInterval();
+        long electionTimeout = leaderAvailabilityTimers.getElectionTimeout();
+        if ( delay >= electionTimeout )
+        {
+            log.warn( "Message %s processing delay %s greater than election timeout interval %s", message, delay, electionTimeout );
```
martinfurmanski: I am skeptical to this warning, which could have a great spamming and unnecessary support-inducing potential. Having the metrics available should perhaps be a sufficient solution?

martinfurmanski (changes requested):
I am slightly skeptical to this overall because it sort of messes with the layering of code and measures something very deep in the stack. The worst offender here though with the clock I believe can be fixed.

I think it has potential to be useful for our internal debugging of the system going forward though, and perhaps even generally for customers in a little longer perspective.

I would like to see this in action in one of our more stressful workloads.

I am also missing tests showing that the various timers work and also Java docs describing them.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
andrewkerr9000: Keeping a reference to the clock makes it easier to reuse for creating messages in response. Alternatives aren't great:

* Don't inject a Clock, hard code instead
* Allow a null Instant, or make Optional
* Modify message handle methods to take a clock
* Make role handlers not singletons and inject clock
* Wrap messages in a timing aware wrapper similar to ClusterIdAwareMessage. Tried this, all of plain message / cluster ID message / timing message / cluster ID and timing message occur, it got messy. Could be the best solution, would be nicer in Scala :smiling_imp:

andrewkerr9000:
I am a little concerned that the monitoring, in particular the Timers, could appreciably impact performance or memory footprint.

There is (currently) one Timer per message type, plus one for all types. There could be one per role / message type combination, which may be excessive.

andrewkerr9000:
I'm thinking of inserting a new MessageHandler in front of the RaftMachine to be responsible for monitoring RaftMachine.

It may then be worth introducing a class to be responsible for pipelining message handlers to make it clearer how they are wired up.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
martinfurmanski: I would be in favour of "Allow a null Instant, or make Optional" because this field is not really part of the message proper and should not be serialized. It is merely an additional tag on the message that gets added when it passes through the processing layers. Wrapping it would also be ok, but just having a special field is perhaps easier to work with and is also low overhead.

andrewkerr9000:
Added a received-at aware wrapper for Raft messages and a monitoring-dedicated handler. There are a couple of further refactorings I want to look at too.

andrewkerr9000:
Introduced a module for creating chained raft message handler, with dependencies. Refactored handler chaining to be functional composition. Refactored life cycle handling so that head of message handler chain is added to core life, and delegates onwards - this now includes the job scheduler for BatchingMessageHandler. Also split out cluster binding logic into a dedicated handler, with two instances in handler chain - the second in on the processing thread but may be redundant.

Will squash.

Will also fire up soak tests, both to observe monitoring and to check I haven't broken anything - PR is quite invasive now.

May also improve unit test coverage on RaftMessageHandler/RaftMessageApplier

andrewkerr9000:
@martinfurmanski just about finished messing about with this now. Care to take another look? Hopefully clear to review commit by commit.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/core/consensus/CatchUpTest.java
```diff
@@ -95,9 +95,9 @@ public void newMemberWithNoLogShouldCatchUpFromPeers() throws Throwable
         net.disconnect( sleepyId );
 
         // when
+        fixture.members().withId( leaderId ).raftInstance().handle( new Request( leader, valueOf( 30 ) ) );
```
martinfurmanski: Was this an intended move of this line?

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/core/consensus/roles/PreVotingVoterTest.java
```diff
@@ -52,7 +52,8 @@ public void shouldRespondPositivelyIfWouldVoteForCandidate() throws Exception
         RaftState raftState = initialState();
 
         // when
-        Outcome outcome = new Follower().handle( new RaftMessages.PreVote.Request( member1, 0, member1, 0, 0 ), raftState, log() );
+        Outcome outcome = new Follower().handle( new RaftMessages.PreVote.Request( member1, 0, member1, 0, 0 ),
```
martinfurmanski: This file seems unnecessarily reformatted.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/RaftServerModule.java
```diff
@@ -0,0 +1,117 @@
...
+
+        ComposableMessageHandler leaderAvailabilityHandler =
+                LeaderAvailabilityHandler.composable( consensusModule.getLeaderAvailabilityTimers(), consensusModule.raftMachine()::term );
+
+        ComposableMessageHandler clusterBindingHandler = ClusterBindingHandler.composable( logProvider );
+
+        return clusterBindingHandler
+                .compose( leaderAvailabilityHandler )
+                .compose( batchingMessageHandler )
+                .compose( clusterBindingHandler )
```
martinfurmanski: Should the cluster-binding really be twice in the stack here?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
```diff
@@ -979,38 +1006,14 @@ public String toString()
...
-
-        @Override
-        public <T, E extends Exception> T dispatch( Handler<T,E> visitor ) throws E
-        {
-            return message().dispatch( visitor );
-        }
     }
 
-    class ReceivedInstantAwareMessageImpl implements ReceivedInstantAwareMessage
+    class ReceivedInstantAwareMessageImpl<RM extends RaftMessage> implements ReceivedInstantAwareMessage<RM>
```
martinfurmanski: I wonder if it would be possible and beneficial to flatten the type-hierarchy of raft messages, to just have a single base RaftMessage type with all the different behaviours in it; ReceivedInstant, default implementations of dispatch, ClusterId, and so on.

martinfurmanski (approved):
Minor comments. Checking approve, but fix what makes sense.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/RaftServerModule.java
andrewkerr9000: It's the existing logic. I expect it's unnecessary.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMessages.java
andrewkerr9000: Not all RaftMessage instances have (or need) a ReceivedInstant, or a ClusterId, or either. This way keeps all the complexity in one place.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/core/consensus/CatchUpTest.java
andrewkerr9000: Doubtful. Will revert.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/core/consensus/roles/PreVotingVoterTest.java
andrewkerr9000: Well spotted. This would have been a left over from the abandoned approach of including an Instant in all RaftMessages.

PR #10576, opened 2017-12-04
Unflake WorksyncTest#mustCombineWork

ragadeeshu:
By making sure that we apply at least one unit of work before starting the
other appliers.


chrisvest (approved):
👌

PR #10578, opened 2017-12-05
Use specialised primitive long queue in FreeIdKeeper

MishaDemianenko:
Use primitive long queue instead of Deque<Long> in FreeIdKeeper
to avoid conversion between primitive long and object wrapper
during id reuse.


tinwelint (approved):
Nice!

PR #10580, opened 2017-12-05
Improved timer service

martinfurmanski:
This replaces the very custom polling timer service with a
simple abstraction working on top of the job scheduler.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/schedule/Delay.java
```diff
@@ -0,0 +1,45 @@
...
+package org.neo4j.causalclustering.core.consensus.schedule;
+
+import java.util.concurrent.TimeUnit;
+
+public class Delay
+{
+    private final long amount;
+    private final TimeUnit unit;
+
+    Delay( long amount, TimeUnit unit )
```
andrewkerr9000, line 29: Does this have advantages over using a `java.time.Duration`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/CatchupPollingProcess.java
```diff
@@ -62,7 +65,7 @@
  */
 public class CatchupPollingProcess extends LifecycleAdapter
 {
-    enum Timeouts implements TimeoutName
+    enum Timeouts implements TimerName
```
phughk: If `TimeoutName` was changed to `TimerName` shouldn't `Timeouts` be renamed to `Timers`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMachine.java
```diff
@@ -124,28 +129,30 @@ public synchronized void startTimers()
...
+            electionTimer.cancel( true, true );
         }
         if ( heartbeatTimer != null )
         {
-            heartbeatTimer.cancel();
+            heartbeatTimer.cancel( true, true );
         }
     }
 
     private TimeoutHandler renewing( ThrowingAction<Exception> action )
```
phughk, line 153: Could also be renamed to `TimerHandler`

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/schedule/Timer.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Deactivates the timer and cancels a currently running job.
+     * <p>
+     * Be careful to not have both a timeout handler executing in parallel
+     * with a timer, because this will just cancel the timer. If you for example
+     * {@link #reset()} in the timeout handler, but keep executing the handler,
+     * then a subsequent cancel will not ensure that the first execution of the
+     * handler.
```
andrewkerr9000: Incomplete sentence?

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/tx/CatchupPollingProcessTest.java
```diff
@@ -96,7 +102,8 @@ public void shouldSendPullRequestOnTick() throws Throwable
         when( txApplier.lastQueuedTxId() ).thenReturn( lastAppliedTxId );
 
         // when
-        timeoutService.invokeTimeout( TX_PULLER_TIMEOUT );
+        timerService.invoke( TX_PULLER_TIMEOUT );
+        scheduler.runJob();
```
andrewkerr9000: These lines are repeated - may be worth a method?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/shipping/RaftLogShipper.java
```diff
@@ -350,22 +330,21 @@ private boolean timedOut()
 
     private void scheduleTimeout( long deltaMillis )
     {
-        // TODO: This cancel/create dance is a bit inefficient... consider something better.
-
         timeoutAbsoluteMillis = clock.millis() + deltaMillis;
 
-        if ( timeout != null )
+        if ( timer == null )
```
phughk, line 335: Is this the same logic? The former timer doesn't get cancelled I think

phughk (changes requested):

andrewkerr9000 (approved):
LGTM, but there's some Javadoc that looks incomplete. I'd also be inclined to use a `java.time.Duration` rather than introduce the `Delay` type unless there's a strong reason not to.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/RaftMachine.java
martinfurmanski: I think here the semantics are correct. A timer is like an egg timer and when it rings a timeout has occurred and thus you handle it with a timeout handler.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/CatchupPollingProcess.java
martinfurmanski: I agree, but didn't care at the time because this was outside the timer-framework. Will fix though!

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/schedule/Delay.java
martinfurmanski: I had some impedance mismatch trying to use Duration, because it was too flexible compared to what I would be able to support. Perhaps it was just my lack of skill dealing properly with time though.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/schedule/Timer.java
martinfurmanski: Yes, will fix!

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/tx/CatchupPollingProcessTest.java
martinfurmanski: Ah, I actually wrote these tests before I introduced the FakeClockJobScheduler which I think can be used here. Will attempt!

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/shipping/RaftLogShipper.java
martinfurmanski: Interesting. Will investigate what is up here.

martinfurmanski: Yes, so this works a bit differently now. What you are looking at here is the idempotent operation of creating the timer (once). There is no need to cancel() the Timer with this new TimerService when you set() or reset() it.

martinfurmanski:
@phughk @andrewkerr9000 Addressed addressable comments. PTAL!

phughk (approved):

PR #10582, opened 2017-12-05
Add eagerness before updating procedures

### LINKED ISSUE #9055: User procedure call fails to delete node: EntityNotFoundException/QueryExecutionException
**Version:** community-3.1.1 
**Operating System:** arch-linux  
**API:** user procedures/cypher

The following example procedure tries to delete a node (`endNode`) related to the input node. The procedure call returns, but the transaction is rolled back with the error `EntityNotFoundException` (HTTP-API) or `QueryExecutionException` (neo4j-shell).

The error occurs depending on how the input nodes are matched in the cypher query.

### Example procedure
```
package example;
import org.neo4j.graphdb.*;
import org.neo4j.procedure.*;

public class example {

    @Procedure(mode = Mode.WRITE)
    @Description("example.testdelete(node, relType) deletes node and related (relType) nodes")
    public void testdelete(@Name("node") Node node, @Name("relType") String relType) {

        Iterable<Relationship> rels = node.getRelationships(RelationshipType.withName(relType), Direction.OUTGOING);
        for (Relationship rel: rels) {
            Node endNode = rel.getEndNode();
            rel.delete();
            endNode.delete(); // this succeeds but transaction with procedure call fails
        }
        node.delete();
    }
}
```

### Steps to reproduce
create a node with related node that should be deleted via the procedure call.
```
create (n1:Label {name: "test"})-[:REL]->(n2) return n1,n2;
+--------------------------------------+
| n1                     | n2          |
+--------------------------------------+
| Node[1]{name:"test"}   | Node[2]{}   |
+--------------------------------------+
```

### Fails
The procedure call fails when accessing node attributes while selecting the nodes to input into the call.
```
match (n) where n.name = "test" call example.testdelete(n, "REL") return null;
QueryExecutionException: Node with id 2 has been deleted in this transaction <- neo4j-shell
{ "code": "Neo.ClientError.Statement.EntityNotFound",
  "message": "Node with id 2 has been deleted in this transaction"  } <- HTTP-API
```
Splitting match and call like this also fails.
```
match (n) where n.name = "test" with n as node call example.testdelete(node, "REL") return null;
````

### Works
With input nodes matched only via patterns that exclude the to be deleted node
```
match (n) where id(n)=1 call example.testdelete(n, "REL") return null;
match (n)-[:REL]->() call example.testdelete(n, "REL") return null;
match (n:Label) where n.name = "test" call example.testdelete(n, "REL") return null;
```

The error seems to be related to the cypher execution order in matching attributes/filtering of a node list and the procedure call.

### Workaround
Collecting all candiate nodes prior to calling `testdelete` works.
```
match (n) where n.name = "test" with collect(n) as ln unwind ln as l call example.testdelete(l, "REL") return null;
```

### Question: bug or limitation

Should this be considered a bug or a limitiation in mixing/interaction of cypher with procedure calls?
Perhaps there is a better/correct way to do this?

Thanks

pontusmelke:
Whenever a procedure is annotated with `@WRITE` we need to make the plan
eager.

changelog: Fixes #9055, user procedures fails to delete node.

### DIFF
--- a/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/EagerizationAcceptanceTest.scala
+++ b/community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/EagerizationAcceptanceTest.scala
@@ -172,7 +172,7 @@ class EagerizationAcceptanceTest
     assertNumberOfEagerness(query, 1)
   }
 
-  test("should not introduce extra eagerness after CALL of writing procedure") {
+  test("should introduce extra eagerness after CALL of writing procedure") {
     // Given
     registerProcedure("user.mkRel") { builder =>
       builder.in("x", Neo4jTypes.NTNode)
@@ -203,11 +203,10 @@ class EagerizationAcceptanceTest
     val result = executeWithCostPlannerOnly(query)
     result.size should equal(4)
 
-    // Correct! Eagerization happens as part of query context operation
-    assertNumberOfEagerness(query, 0)
+    assertNumberOfEagerness(query, 1)
   }
 
-  test("should not introduce extra eagerness after CALL of writing void procedure") {
+  test("should introduce extra eagerness after CALL of writing void procedure") {
     // Given
     var counter = Counter(0)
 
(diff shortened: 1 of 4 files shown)


pontusmelke:
I added `NOT_READY_FOR_MERGE` because I want to check that this doesn't cause any issues with apoc.

PR #10583, opened 2017-12-05
Don't depend on EntityNotFoundException in compiler module

pontusmelke:
We have a binary dependency on `EntityNotFoundException` from
`cypher-compiler-2.3` which makes it impossible to move the exception
in 3.4. We need to fix this binary dependency in all versions - or at
least 2.3 and 3.1 - and then do new releases and finally update the
dependencies in 3.4.


fickludd (approved):
What an annoying dependency leak. Good fix.

PR #10589, opened 2017-12-05
Release the storecopy mutex after the store copy

martinfurmanski:
The store copy request handler just queues up requests
on the netty pipeline and the release of the mutex must
be scheduled as a response to the completion of the actual
handling. This is done by registering on the future supplied
by netty.

Also did a general clean up of the code here with a better flow,
cleaner paragraphs, refactoring of methods and a few comments.


phughk (changes requested):
Needs testing

tinwelint (approved):
Generally looks like a sane change.

phughk (approved):

martinfurmanski:
@phughk I have refactored a lot of things around this now. Feel free to take another look.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreResource.java
```diff
@@ -0,0 +1,115 @@
...
+        this.path = relativePath;
+        this.recordSize = recordSize;
+        this.pageCache = pageCache;
+        this.fs = fs;
+    }
+
+    ReadableByteChannel open() throws IOException
+    {
+        Optional<PagedFile> existingMapping = pageCache.getExistingMapping( file );
+        if ( existingMapping.isPresent() )
```
phughk, line 55: I think this can be changed to
`return existingMapping.map(pFile -> pFile.openReadableBytChannel()).orElse(fs.open(file, "r"));`

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreResource.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public boolean equals( Object o )
+    {
+        if ( this == o )
+        {
+            return true;
+        }
+        if ( o == null || getClass() != o.getClass() )
```
phughk, line 96: Pedantic comment about `class!=subclass<=>true` for subclasses.

phughk (changes requested):
Minor comments

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreResource.java
martinfurmanski: There is a trw in there as well, which does the close().

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreResource.java
martinfurmanski: Is that really standard? I just let IntelliJ generate this.

phughk (approved):

PR #10551, opened 2017-11-30
3.4 support for 3.3

sherfert:

PR #10593, opened 2017-12-06
Reporting PlannerVersion argument and testing it in CypherComparisonSupport

sherfert:
This adds a new PlanDescription Argument `PlannerVersion`, so that we can distinguish PlannerVersion and RuntimeVersion and assert both are correct when running older versions.

This should wait for https://github.com/neo4j/neo4j/pull/10551 to go in before being merged.


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/Compatibility.scala
```diff
@@ -163,7 +163,7 @@ extends LatestRuntimeVariablePlannerCompatibility[CONTEXT3_4, T, StatementV3_3](
           val logicalPlanStateV3_3 = compiler.planPreparedQuery(preparedQuery, contextV3_3)
           val logicalPlanStateV3_4 = helpers.as3_4(logicalPlanStateV3_3)
           // Here we switch from 3.3 to 3.4
-          val result = createExecPlan.transform(logicalPlanStateV3_4, contextV3_4)
+          val result: CompilationState = createExecPlan.transform(logicalPlanStateV3_4, contextV3_4)
```
craigtaverner: Do we need explicit type here?

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/helpers.scala
```diff
@@ -146,7 +146,8 @@ object helpers {
 
   def as3_4(logicalPlan: LogicalPlanStateV3_3) : LogicalPlanState = {
     val startPosition = logicalPlan.startPosition.map(as3_4)
-    val plannerName = as3_4(logicalPlan.plannerName)
+    // Wrap the planner name to correctly report version 3.3.
+    val plannerName = PlannerNameWithVersion(as3_4(logicalPlan.plannerName), "3.3")
```
craigtaverner: Is "3.3" not perhaps defined as a constant somewhere already?

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/RewindableExecutionResult.scala
```diff
@@ -289,9 +289,10 @@ object RewindableExecutionResult {
     override def javaColumnAs[T](column: String): ResourceIterator[T] = inner.javaColumnAs(column)
 
     override def executionPlanDescription(): InternalPlanDescription = {
-      var description = lift(inner.executionPlanDescription())
-      if (!description.arguments.exists(_.isInstanceOf[Arguments.Version])) {
-        description = description.addArgument(planDescription.InternalPlanDescription.Arguments.Version("CYPHER 3.1"))
+      var description = lift(inner.executionPlanDescription()).addArgument(Arguments.PlannerVersion("3.1"))
+      if (!description.arguments.exists(_.isInstanceOf[Arguments.RuntimeVersion])) {
+        // FIXME do we need this
+        description = description.addArgument(Arguments.RuntimeVersion("CYPHER 3.1"))
```
craigtaverner: I wonder if this works around bugs in some older version of the compiler for 3.1, and if that bug is fixed, then this workaround could be removed. Perhaps it is as simple as removing the workaround and running all compatibility tests?

community/cypher/planner-spi-3.4/src/main/scala/org/neo4j/cypher/internal/planner/v3_4/spi/PlannerName.scala
```diff
@@ -68,3 +70,11 @@ object PlannerNameFor {
...
   }
 }
+
+/**
+ * When running older planner from current version
+ */
+case class PlannerNameWithVersion(inner: PlannerName, version: String) extends PlannerName {
+  override def name = inner.name
+  override def toTextOutput = inner.toTextOutput
+}
```
craigtaverner, line 80: Missing newline

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/InternalPlanDescription.scala
```diff
@@ -276,7 +281,7 @@ final case class PlanDescriptionImpl(id: LogicalPlanId,
 
   override def toString = {
     val version = arguments.collectFirst {
-      case Version(v) => s"Compiler $v$NL"
+      case RuntimeVersion(v) => s"Compiler $v$NL"
```
craigtaverner: I wonder if `Compiler` should be changed to `Runtime`. The word 'compiler' before meant both planner and runtime, but now we have a clearer separation between planner and runtime.

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/LogicalPlan2PlanDescription.scala
```diff
@@ -33,9 +33,10 @@ object LogicalPlan2PlanDescription extends ((LogicalPlan, PlannerName) => Intern
                      plannerName: PlannerName): InternalPlanDescription = {
     val readOnly = input.solved.readOnly
     new LogicalPlan2PlanDescription(readOnly).create(input)
-      .addArgument(Version("CYPHER 3.4"))
+      .addArgument(RuntimeVersion("CYPHER 3.4"))
```
craigtaverner: Before the Version related to both planner and runtime, and the text `Cypher x.y` would imply an `x.y` version of the Cypher language (parsing, planning, executing). Now we are talking here only about the runtime, but in fact from the users perspective, when they see `Cypher x.y` they are thinking about the version of the language spec, which is tied to the parser and planner much more than the runtime.

I think this text is also going to come out in the query plan description, and we will see the wrong values here. For example, if the user runs `PROFILE CYPHER 3.3 RETURN 1`, they are going to see a plan description with `CYPHER 3.4` printed, instead of the version they expected.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -492,53 +495,55 @@ object CypherComparisonSupport {
...
-      val rulePlannerFallback = Planners.Rule.acceptedPlannerNames.contains(reportedPlanner) && Versions.V3_1.acceptedVersionNames.contains(reportedVersion)
-
+      val (reportedRuntime: String, reportedPlanner: String, reportedVersion: String, reportedPlannerVersion: String) = extractConfiguration(internalExecutionResult)
       if (!runtime.acceptedRuntimeNames.contains(reportedRuntime))
         fail(s"did not use ${runtime.acceptedRuntimeNames} runtime - instead $reportedRuntime was used. Scenario $name")
       if (!planner.acceptedPlannerNames.contains(reportedPlanner))
         fail(s"did not use ${planner.acceptedPlannerNames} planner - instead $reportedPlanner was used. Scenario $name")
-      if (!(version.acceptedVersionNames.contains(reportedVersion) || rulePlannerFallback))
-        fail(s"did not use ${version.acceptedVersionNames} version - instead $reportedVersion was used. Scenario $name")
+      if (!version.acceptedRuntimeVersionNames.contains(reportedVersion))
```
craigtaverner, line 503: Have we removed support for testing the RULE planner now?

craigtaverner (changes requested):
Previously the single version would look like `CYPHER 3.x` and would be printed in the plan description for `EXPLAIN` and `PROFILE` queries. This same text is now used to describe the runtime version, and a separate field for plannerVersion (containing only the number `3.x`) is added to the arguments, so we will not see it in the execution plan description, but only see the runtime version, which will confuse users.

Before this PR the Version field related to both planner and runtime, and the text `Cypher x.y` would imply an `x.y` version of the Cypher language (parsing, planning, executing). Now we are talking here only about the runtime, but in fact from the users perspective, when they see `Cypher x.y` they are thinking about the version of the language spec, which is tied to the parser and planner much more than the runtime.

I think this text is also going to come out in the query plan description, and we will see the wrong values here. For example, if the user runs `PROFILE CYPHER 3.3 RETURN 1`, they are going to see a plan description with `CYPHER 3.4` printed, instead of the version they expected.

Perhaps remove the prefix `CYPHER ` from the RuntimeVersion, and in the printout of the plan description print s"CYPHER $plannerVersion" instead of RuntimeVersion.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
Lojjs: No, but we have removed rule fallback i.e. you can not write something like Cypher 3.3 planner=rule anymore so rule will not be a special case

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/LogicalPlan2PlanDescription.scala
sherfert: I will change the toString of PlanDescription prefix to account for both plannerVersion and runtimeVersion. This will account for both this and the above comment.

community/cypher/cypher/src/test/scala/org/neo4j/cypher/internal/RewindableExecutionResult.scala
sherfert: Since the "remove duplicate tx wrappers" PR changes that code already (and removes the unnecessary parts), I am not changing it here, to avoid conflicts.

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/LogicalPlan2PlanDescription.scala
sherfert: As of the latest commit, the output of the planDescription of `CYPHER 3.3 EXPLAIN MATCH (n) RETURN n` is

```
Planner COST 3.3

Runtime INTERPRETED CYPHER 3.4

+-----------------+----------------+-----------+
| Operator        | Estimated Rows | Variables |
+-----------------+----------------+-----------+
| +ProduceResults |              1 | n         |
| |               +----------------+-----------+
| +AllNodesScan   |              1 | n         |
+-----------------+----------------+-----------+
```

I agree that having "CYPHER" in the runtime version vs. not having it in the planner version is kinda weird.

sherfert: Also, with that latest commit, the browser still displays "Cypher version: CYPHER 3.4, planner: COST, runtime: INTERPRETED" for the above EXPLAIN query...
Probably we need to change the interaction between the two here.

craigtaverner (approved):

PR #10595, opened 2017-12-06
Remove scala warnings

pontusmelke:
Warnings are annoying, lets have fewer of them.


sherfert:
Looks nice! I am just wondering if there is a particular reason to do this in 3.3 instead of 3.4?

pontusmelke:
@sherfert I think it makes us look bad.

craigtaverner (approved):

PR #10596, opened 2017-12-06
Fix regressions in Core API

pontusmelke:
After merging https://github.com/neo4j/neo4j/pull/10495 we have seen a couple of regressions in the Core API benchmarks. This PR addresses these regression. Fixes include:

- Don't be stupid in `NodeProxy::getProperties(String[])`
- Add `nodeExists` to API to get more efficient `nodeGetById`
- ~Reuse node cursors and property cursors.~
- Load label cursor lazily
- Make `Read` monomorphic

### DIFF
--- a/community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
+++ b/community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
@@ -41,7 +41,7 @@ import org.neo4j.graphalgo.impl.path.ShortestPath.ShortestPathPredicate
 import org.neo4j.graphdb._
 import org.neo4j.graphdb.security.URLAccessValidationError
 import org.neo4j.graphdb.traversal.{Evaluators, TraversalDescription, Uniqueness}
-import org.neo4j.internal.kernel.api.{IndexQuery, NodeCursor, PropertyCursor}
+import org.neo4j.internal.kernel.api.{IndexQuery, NodeCursor}
 import org.neo4j.kernel.GraphDatabaseQueryService
 import org.neo4j.kernel.api.exceptions.ProcedureException
 import org.neo4j.kernel.api.exceptions.schema.{AlreadyConstrainedException, AlreadyIndexedException}
@@ -103,6 +103,8 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
   private def writes() = transactionalContext.kernelTransaction.dataWrite()
   private def reads() = transactionalContext.kernelTransaction.dataRead()
   private def cursors = transactionalContext.kernelTransaction.cursors()
+  private val nodeCursor = allocateAndTraceNodeCursor()
+  private val propertyCursor = allocateAndTracePropertyCursor()
   private def tokenRead = transactionalContext.kernelTransaction.tokenRead()
   private def tokenWrite = transactionalContext.kernelTransaction.tokenWrite()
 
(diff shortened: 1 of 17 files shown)


community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Read.java
```diff
@@ -73,6 +73,13 @@ void nodeIndexSeek( IndexReference index, NodeValueIndexCursor cursor, IndexOrde
      */
     void singleNode( long reference, NodeCursor cursor );
 
+    /**
+     * Checks if a node exists in the database
+     * @param id The id of the node to check
+     * @return <tt>true</tt> if the node exists, otherwise <tt>false</tt>
+     */
+    boolean nodeExists( long id );
```
fickludd, line 81: A little test of this would be nice, as it lies in the API.

pontusmelke: Added tests

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -134,47 +135,34 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
...
-      }
-      VirtualValues.list(labelArray: _*)
-    } finally {
-      if (nodeCursor != null) {
-        nodeCursor.close()
-      }
+  override def getLabelsForNode(node: Long): ListValue = {
+    val nodeCursor = acquireNodeCursor
+    reads().singleNode(node, nodeCursor)
+    if (!acquireNodeCursor.next()) {
```
fickludd: this should be `nodeCursor` i guess, even though they amount to the same thing. Would actually prefer to define `acquireNodeCursor()` with an empty parameter-list to clarify that it is a call and not a value.

community/kernel/src/main/java/org/neo4j/kernel/api/KernelTransaction.java
```diff
@@ -96,6 +98,10 @@
      */
     Statement acquireStatement();
 
+    NodeCursor nodeCursor();
+
+    PropertyCursor propertyCursor();
```
fickludd: This itches a little... it feels like a hack to accommodate Core API. I guess it's safe as long as Core API transactions are thread bound and this cursors are not embedded in returned iterators or similar.

WDYT?

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -579,6 +569,17 @@ public boolean hasProperty( String key )
         }
     }
 
+    private KernelTransaction safeAcquireTransaction()
+    {
+        KernelTransaction transaction = actions.kernelTransaction();
+        Optional<Status> terminationReason = transaction.getReasonIfTerminated();
```
fickludd: Checking `transaction.isTerminated` first, and also getting the reason if true is faster in the happy path.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -814,7 +814,12 @@ protected boolean fetchNext()
 
     private void assertTransactionOpen()
     {
-        Optional<Status> terminationReason = spi.currentTransaction().getReasonIfTerminated();
+        assertTransactionOpen( spi.currentTransaction() );
+    }
+
+    private void assertTransactionOpen( KernelTransaction transaction )
+    {
+        Optional<Status> terminationReason = transaction.getReasonIfTerminated();
```
fickludd: Same `transaction.isTerminated` optimization could be used here.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -73,11 +73,6 @@ void scan( Read read )
         this.next = 0;
         this.highMark = read.nodeHighMark();
         this.read = read;
-        if ( labelCursor == null )
-        {
-            labelCursor = read.labelCursor();
-        }
-        this.labelCursor = read.labelCursor();
```
fickludd, line 80: Wow, this previous _solution_ looks pretty stupid. Did I code that?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -347,6 +338,15 @@ private void reset()
         setId( NO_ID );
     }
 
+    private RecordCursor<DynamicRecord> labelCursor()
```
fickludd: This is much better. In most dbs we will not need this cursor.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
```diff
@@ -90,6 +91,24 @@ public AllStoreHolder( StorageEngine engine,
         this.explicitIndexStore = explicitIndexStore;
     }
 
+    @Override
+    public boolean nodeExists( long id )
+    {
+        if ( hasTxStateWithChanges() )
```
fickludd: Should there be an `assertOpen()` here somewhere?

fickludd (changes requested):
All of this is awesome. Got some minor things. Happy with the added tests.

pontusmelke:
@fickludd I think you are right, got a little carried away shaving percents out of the benchmarks. Reverted to use allocation for each call.

fickludd (approved):

PR #10597, opened 2017-12-06
Do not override UDC settings with default values on startup

MishaDemianenko:
Do not override UDC settings that specified in neo4j instance config
with default values.
Cleanup re-invented default property values loading.
Use settings default values instead.

### DIFF
--- a/community/udc/src/main/filtered-resources/org/neo4j/ext/udc/udc.properties
+++ b/community/udc/src/main/filtered-resources/org/neo4j/ext/udc/udc.properties
@@ -1,3 +1,0 @@
-unsupported.dbms.udc.source=${neo4j-source-distribution}
-unsupported.dbms.udc.reg=unreg
-
--- a/integrationtests/src/test/java/org/neo4j/ext/udc/impl/UdcExtensionImplTest.java
+++ b/integrationtests/src/test/java/org/neo4j/ext/udc/impl/UdcExtensionImplTest.java
@@ -48,7 +48,6 @@ import org.neo4j.ext.udc.UdcConstants;
 import org.neo4j.ext.udc.UdcSettings;
 import org.neo4j.graphdb.GraphDatabaseService;
 import org.neo4j.graphdb.factory.GraphDatabaseBuilder;
-import org.neo4j.helpers.collection.MapUtil;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
 import org.neo4j.test.TestEnterpriseGraphDatabaseFactory;
 import org.neo4j.test.TestGraphDatabaseFactory;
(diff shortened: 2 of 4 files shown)


community/udc/src/main/filtered-resources/org/neo4j/ext/udc/udc.properties
```diff
@@ -1,3 +0,0 @@
-unsupported.dbms.udc.source=${neo4j-source-distribution}
```
chrisvest, line 1: Should we not also add this injection to neo4j.conf? Otherwise we won't be collecting distribution information at all.

davidegrohmann: If we remove this file, these two https://github.com/neo4j/neo4j/blob/3.3/community/udc/pom.xml#L68 and https://github.com/neo4j/neo4j/blob/3.3/community/udc/pom.xml#L89 are useless, no?

integrationtests/src/test/java/org/neo4j/ext/udc/impl/UdcExtensionImplTest.java
```diff
@@ -243,20 +244,6 @@ private void blockUntilServerAvailable( final URL url ) throws Exception
...
-        config.put( UdcSettings.udc_registration_key.name(), "marketoid" );
-
-        // When
-        graphdb = createDatabase( config );
-
-        // Then
-        assertGotSuccessWithRetry( IS_GREATER_THAN_ZERO );
-        assertEquals( "test-reg", handler.getQueryMap().get( REGISTRATION ) );
-    }
-
```
davidegrohmann, line 259: Maybe add:

```
    @Test
    public void shouldUseTheCustomConfiguration() throws Exception
    {
        // Given
        config.put( UdcSettings.udc_source.name(), "my_source" );
        config.put( UdcSettings.udc_registration_key.name(), "my_key" );

        // When
        graphdb = createDatabase( config );

        // Then
        Config config = ((GraphDatabaseAPI) graphdb).getDependencyResolver().resolveDependency(Config.class);

        assertEquals("my_source",config.get(UdcSettings.udc_source));
        assertEquals("my_key",config.get(UdcSettings.udc_registration_key));
    }
```

community/udc/src/main/filtered-resources/org/neo4j/ext/udc/udc.properties
MishaDemianenko: @chrisvest we have an injection of this property in neo4j.conf when we do packaging:
dbms.jvm.additional=-Dunsupported.dbms.udc.source=#{unsupported.dbms.udc.source}
and "unsupported.dbms.udc.source" is replaced with tarball or zip when we do standard packaging for example

MishaDemianenko: @davidegrohmann yes you are right those 2 profiles are useless and always were and they already removed in 3.4 development branch

integrationtests/src/test/java/org/neo4j/ext/udc/impl/UdcExtensionImplTest.java
MishaDemianenko: I would say its useless since we already have "shouldOverrideSourceWithSystemProperty" that already checks that custom value propagation works, but if adding new test will make you even a bit happier reviewer - why not

klaren (approved):
There are still unused leftovers, but I believe I got rid of most of them in 3.4. Maybe you do a quick check when you forward merge this? Don't do it now since that would produce way too many conflicts.

PR #10599, opened 2017-12-07
Delete browser code and instead pull Browser in as a dependency

oskarhane:


oskarhane:
Failing because the new browser dependency isn't at maven central yet.

PR #10600, opened 2017-12-07
Use new API for reading labels in Core API

pontusmelke:
Forgot about these last time around.

~Builds on top of https://github.com/neo4j/neo4j/pull/10596~

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
@@ -718,12 +718,16 @@ public class NodeProxy implements Node
     @Override
     public boolean hasLabel( Label label )
     {
-        try ( Statement statement = actions.statement() )
+        KernelTransaction transaction = safeAcquireTransaction();
+        try ( Statement ignore = transaction.acquireStatement();
+              NodeCursor nodes = transaction.cursors().allocateNodeCursor() )
         {
-            int labelId = statement.readOperations().labelGetForName( label.name() );
-            return statement.readOperations().nodeHasLabel( getId(), labelId );
+            int labelId = transaction.tokenRead().labelGetForName( label.name() );
+            transaction.dataRead().singleNode( nodeId, nodes );
+            return nodes.next() && nodes.labels().contains( labelId );
+
         }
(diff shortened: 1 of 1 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -732,23 +736,23 @@ public boolean hasLabel( Label label )
...
+                throw new NotFoundException( "Node not found" );
+            }
+            LabelSet labelSet = nodes.labels();
+            TokenRead tokenRead = transaction.tokenRead();
+            ArrayList<Label> list = new ArrayList<>( labelSet.numberOfLabels() );
+            for ( int i = 0; i < labelSet.numberOfLabels(); i++ )
+            {
+                list.add( label( tokenRead.labelGetName( labelSet.label( i ) ) ) );
+            }
+            return list;
```
fickludd, line 755: Maybe we should make the LabelSet iterable instead, if we are anyway creating that object?

fickludd (changes requested):
This looks fine. Would like to change the labels to to create a new buffer, but let LabelSet implement iterable instead.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
pontusmelke: That would mean an extra object though? We would need to instantiate the iterator, would probably not be much worse since escape analysis should put it on the stack but still...

fickludd: Eh, not sure I follow... I'm was thinking to return `nodes.labels()` here, which means we allocate the LabelSet, and the iterator in the caller. With the current impl we allocate LabelSet + ArrayList here, and iterator in the caller. I realise that that wasn't really obvious from my previous statement.

fickludd (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
fickludd: Oops, forgot about converting to strings... nevermind!

PR #10604, opened 2017-12-07
Refactoring around Cypher query cache

systay:


sherfert:
I think this misses our latest commit from today

SaschaPeukert (approved):
Looks good to me. Nice improvement in readability

PR #10606, opened 2017-12-08
Removes accidental limitation of number of reltypes to 64K for high_limit format

tinwelint:
The format itself already supports 3B, i.e. ~16M relationship types, but the format
was accidentally limited to 2B from outside upon its construction.
This removes that limit and will not incur any format change.


MishaDemianenko (approved):

MishaDemianenko:
looks good...but please wait with merging it one more test I want to do..

MishaDemianenko:
Ok so just reading works fine in old format, but as soon as any update operations will encounter higher then supported id (for example on recovery) it will fail

MishaDemianenko:
Lets wait with merging this for now, please contact me offline for reasoning about it

tinwelint:
Succeeded by #10674

PR #10608, opened 2017-12-08
Import co-locates double record units for relationships

tinwelint:
If it sees that the number of relationships will likely be big enough that many of the relationships will require secondary record units.

This PR removes some home brew id sequences and generators that the importer previously had in favor of the ones in the stores. The importer looks at the estimates and decides to allocate two ids per relationships if number of relationships are large enough and the format is the high limit format. This to reduce contention in the import and faster reads from the db after import.


community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/store/SecondaryUnitPrepareIdSequenceTest.java
```diff
@@ -0,0 +1,81 @@
...
+        // when
+        long recordId = 10;
+        IdSequence prepared = idSequence.apply( actual ).apply( recordId );
+        long nextRecordId = prepared.nextId();
+        assertEquals( 10 + 1, nextRecordId );
+        verifyNoMoreInteractions( actual );
+        try
+        {
+            prepared.nextId();
+            fail( "Shoul've failed" );
```
MishaDemianenko: `Shoud`

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
```diff
@@ -157,16 +157,35 @@ public void initialize( Input input ) throws IOException
...
         idGenerator = input.idGenerator();
         nodeRelationshipCache = new NodeRelationshipCache( numberArrayFactory, config.denseNodeThreshold() );
-        memoryUsageStats = new MemoryUsageStatsProvider( nodeRelationshipCache, idMapper, neoStore );
         nodes = input.nodes();
         relationships = input.relationships();
         cachedNodes = cachedForSure( nodes, inputCache.nodes( MAIN, true ) );
-        dependencies.satisfyDependencies( input.calculateEstimates( neoStore.getPropertyStore().newValueEncodedSizeCalculator() ),
-                idMapper, neoStore, nodeRelationshipCache );
+        Estimates inputEstimates = input.calculateEstimates( neoStore.getPropertyStore().newValueEncodedSizeCalculator() );
+        sanityCheckEstimatesWithRecordFormat( inputEstimates );
```
MishaDemianenko, line 166: would really like to see tests for those sanity checks

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -77,6 +80,8 @@
  */
 public class BatchingNeoStores implements AutoCloseable, MemoryStatsVisitor.Visitable
 {
+    // Empirical and slightly defensive threshold where relationship records seem to start requiring double record units.
+    private static final long DOUBLE_RELATIONSHIP_RECORD_UNIT_THRESHOLD = 1L << 33;
```
MishaDemianenko: while this empirical metric can be good enough, i think we need to add a bit of description how it was built (based on what assumptions) or direct future us into "requiredDataLength" from record formats etc.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -418,4 +425,18 @@ public void success()
     {
         successful = true;
     }
+
+    public boolean determineDoubleRelationshipRecordUnits( Estimates inputEstimates )
+    {
+        doubleRelationshipRecordUnits =
+                // TODO figure out in a good way if this is the high limit format
+                recordFormats.relationship().getMaxId() > 2L << 36 &&
```
MishaDemianenko: 2L << 36 should be extracted as constant

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/store/BatchingNeoStores.java
```diff
@@ -418,4 +425,18 @@ public void success()
     {
         successful = true;
     }
+
+    public boolean determineDoubleRelationshipRecordUnits( Estimates inputEstimates )
+    {
+        doubleRelationshipRecordUnits =
+                // TODO figure out in a good way if this is the high limit format
```
MishaDemianenko: and probably TODO can be removed

MishaDemianenko (changes requested):
Couple of small comments, good otherwise

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
tinwelint: Pulled it out and unit tested that, looks better?

tinwelint:
@MishaDemianenko updated

MishaDemianenko (approved):

PR #10612, opened 2017-12-10
Add additional case to parse in inconsistency report

jakewins:
The parse assumes the output comes from a timestamped log; however,
this is (apparently, because I got one which wasn't) not always the
case.

This broadens the parse just-so to parse reports without timestamps.


tinwelint (approved):

jakewins:
Merged manually

PR #10613, opened 2017-12-11
Prometheus endpoint for metrics

klaren:
[Prometheus](https://prometheus.io/) is monitoring and alerting toolkit maintained by the [Cloud Native Computing Foundation](https://www.cncf.io/).

To enable the Prometheus HTTP endpoint you have to set `metrics.prometheus.enabled=true` in your `neo4j.conf`. The setting `metrics.prometheus.endpoint=<ip:port>` can be used to change the listening ip and port, default one is `localhost:2004`.

For example, adding the following to your configuration file:
```
metrics.prometheus.enabled=true
metrics.prometheus.endpoint=localhost:2004
```
will start a Prometheus endpoint at port 2004 accepting connections from the local machine.

Security wise, Prometheus states the following:
>It is presumed that untrusted users have access to the Prometheus HTTP endpoint and logs. They have access to all time series information contained in the database, plus a variety of operational/debugging information.

and should thus be managed at a firewall/OS level. If authentication is a requirement, a reverse proxy can be used on the local machine to expose the information.

### DIFF
--- a/enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
+++ b/enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
@@ -153,4 +153,12 @@ public class MetricsSettings implements LoadableConfig
 
     @Description( "The reporting interval for Graphite. That is, how often to send updated metrics to Graphite." )
     public static Setting<Duration> graphiteInterval = setting( "metrics.graphite.interval", DURATION, "3s" );
+
+    // Prometheus settings
+    @Description( "Set to `true` to enable the Prometheus endpoint" )
+    public static Setting<Boolean> prometheusEnabled = setting( "metrics.prometheus.enabled", BOOLEAN, FALSE );
+
+    @Description( "The hostname and port to use as Prometheus endpoint" )
+    public static Setting<HostnamePort> prometheusEndpoint =
+            setting( "metrics.prometheus.endpoint", HOSTNAME_PORT, "localhost:8080" );
 }
--- a/enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
+++ b/enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
@@ -160,5 +160,5 @@ public class MetricsSettings implements LoadableConfig
 
     @Description( "The hostname and port to use as Prometheus endpoint" )
     public static Setting<HostnamePort> prometheusEndpoint =
-            setting( "metrics.prometheus.endpoint", HOSTNAME_PORT, "localhost:8080" );
+            setting( "metrics.prometheus.endpoint", HOSTNAME_PORT, "localhost:2004" );
 }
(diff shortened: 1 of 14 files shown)


enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
```diff
@@ -153,4 +153,12 @@
...
     @Description( "The reporting interval for Graphite. That is, how often to send updated metrics to Graphite." )
     public static Setting<Duration> graphiteInterval = setting( "metrics.graphite.interval", DURATION, "3s" );
+
+    // Prometheus settings
+    @Description( "Set to `true` to enable the Prometheus endpoint" )
+    public static Setting<Boolean> prometheusEnabled = setting( "metrics.prometheus.enabled", BOOLEAN, FALSE );
+
+    @Description( "The hostname and port to use as Prometheus endpoint" )
+    public static Setting<HostnamePort> prometheusEndpoint =
+            setting( "metrics.prometheus.endpoint", HOSTNAME_PORT, "localhost:8080" );
```
MishaDemianenko: wouldn't say that 8080 is a good default port value, why not to use next port after default graphite server (2004)?

enterprise/metrics/src/main/java/org/neo4j/metrics/output/PrometheusOutput.java
```diff
@@ -0,0 +1,106 @@
...
+    {
+        this.stop();
+    }
+
+    @Override
+    public void report( SortedMap<String,Gauge> gauges, SortedMap<String,Counter> counters,
+            SortedMap<String,Histogram> histograms, SortedMap<String,Meter> meters, SortedMap<String,Timer> timers )
+    {
+        // Prometheus does not support events, just output the latest event that occurred
+        String metricKey = gauges.firstKey();
```
MishaDemianenko, line 103: this looks strange: we already registered export of registry in the init and here again during reporting we can potentially register some other set of events? how is that possible?

MishaDemianenko (commented):
couple of comments/questions, good otherwise

enterprise/metrics/src/main/java/org/neo4j/metrics/MetricsSettings.java
klaren: It's an HTTP server, they typically reside in the 8000 range. 8080 was taken from Prometheus example configurations, 2004 sound equally good.

enterprise/metrics/src/main/java/org/neo4j/metrics/output/PrometheusOutput.java
klaren: This is for the cases where we do not register, `CHECK_POINT_DURATION` and `LOG_ROTATION_DURATION`. But I did not test how this affects logging if both csv and Prometheus is activated at the same time.

klaren:
@MishaDemianenko New commit available for review.

MishaDemianenko (approved):

PR #10614, opened 2017-12-11
Ensure that the page cache events from the catchup client are published

chrisvest:
The catchup client is used by the Read Replicas to keep up to date with
the core replicas.

This is basically the same change as PR #10347, but for the Read
Replicas.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/BatchingTxApplier.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/BatchingTxApplier.java
@@ -21,6 +21,7 @@ package org.neo4j.causalclustering.catchup.tx;
 
 import java.util.function.Supplier;
 
+import org.neo4j.io.pagecache.tracing.cursor.PageCursorTracerSupplier;
 import org.neo4j.kernel.impl.api.TransactionCommitProcess;
 import org.neo4j.kernel.impl.api.TransactionQueue;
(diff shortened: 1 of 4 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/BatchingTxApplier.java
```diff
@@ -53,12 +55,13 @@
     private volatile boolean stopped;
 
     public BatchingTxApplier( int maxBatchSize, Supplier<TransactionIdStore> txIdStoreSupplier,
-            Supplier<TransactionCommitProcess> commitProcessSupplier,
-            Monitors monitors, LogProvider logProvider )
+                              Supplier<TransactionCommitProcess> commitProcessSupplier, Monitors monitors,
+                              PageCursorTracerSupplier pageCursorTracerSupplier, LogProvider logProvider )
```
martinfurmanski, line 59: This indentation looks non-standard (http://neo4j.github.io/).

martinfurmanski (approved):
LGTM

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/BatchingTxApplier.java
chrisvest: Looks pretty standard according to your linked style.

PR #10615, opened 2017-12-11
Unflake RelationshipIdReuseStressIT

tinwelint:
By ensuring that the removal thread executes a threshold amount of removals.
The assertions and the test as a whole hinges on there being removals,
but previously the test was simply given an empirically decided amount of time
which wasn't enough on some machines.

### DIFF
--- a/enterprise/kernel/src/test/java/org/neo4j/graphdb/store/id/RelationshipIdReuseStressIT.java
+++ b/enterprise/kernel/src/test/java/org/neo4j/graphdb/store/id/RelationshipIdReuseStressIT.java
@@ -48,6 +48,7 @@ import org.neo4j.helpers.collection.Iterables;
 import org.neo4j.helpers.collection.Iterators;
 import org.neo4j.kernel.DeadlockDetectedException;
 import org.neo4j.kernel.impl.enterprise.configuration.EnterpriseEditionSettings;
+import org.neo4j.kernel.impl.storageengine.impl.recordstorage.id.IdController;
 import org.neo4j.kernel.impl.store.id.IdGeneratorFactory;
 import org.neo4j.kernel.impl.store.id.IdType;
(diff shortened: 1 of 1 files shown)


enterprise/kernel/src/test/java/org/neo4j/graphdb/store/id/RelationshipIdReuseStressIT.java
```diff
@@ -83,18 +86,25 @@ public void relationshipIdReused() throws Exception
...
         AtomicBoolean stopFlag = new AtomicBoolean( false );
         RelationshipsCreator relationshipsCreator = new RelationshipsCreator( stopFlag, bandLabel, cityLabel );
+        RelationshipRemover relationshipRemover = new RelationshipRemover( bandLabel, cityLabel, stopFlag );
+        IdController idController = embeddedDatabase.getDependencyResolver().resolveDependency( IdController.class );
 
-        List<Future> futures = new ArrayList<>();
-        futures.add( startRelationshipRemoval( bandLabel, cityLabel, stopFlag ) );
-        futures.add( startRelationshipsCreator( relationshipsCreator ) );
+        List<Future<?>> futures = new ArrayList<>();
+        futures.add( executorService.submit( relationshipRemover ) );
```
klaren, line 93: It looks a bit strange you only inlined half of the methods. Otherwise looks good!

klaren (approved):

enterprise/kernel/src/test/java/org/neo4j/graphdb/store/id/RelationshipIdReuseStressIT.java
tinwelint: Inlined the ones that were needed only (where the instances was created before calling the method).

PR #10621, opened 2017-12-11
Unflake HostnamePortTest

tinwelint:
By accepting the fact that some machines will have host names that are
unparsable by URI.create


MishaDemianenko:
please see: https://github.com/neo4j/neo4j/pull/10619 instead

PR #10622, opened 2017-12-11
Check constraints in new API

pontusmelke:
This PR adds constraints checking to the new API:

- check so that unique constraints are upheld on `nodeAddLabel` and `nodeSetProperty`
- use new API for label and property writes in Core API
- use new API for label and property writes in interpreted runtime
- index operations now respect transaction state


community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
```diff
@@ -74,15 +76,14 @@
 
     public AllStoreHolder( StorageEngine engine,
             StorageStatement statement,
-            TxStateHolder txStateHolder,
+            KernelTransactionImplementation ktx,
```
fickludd, line 88: Hmm, was the `KernelTransactionImplementation` introduced just because of convenience? From a architectural point of view I think it makes these classes unnecessarily bounded, and it makes it easier to further tie them together. Would prefer `TxStateHolder + AssertOpen`, and also because it's easier to understand what this class needs.

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/OperationsLockTest.java
```diff
@@ -238,4 +238,14 @@ public void shouldAcquireSchemaReadLockBeforeGettingConstraintsByLabelAndPropert
         // THEN
         verify( locks ).acquireShared( LockTracer.NONE, ResourceTypes.LABEL, descriptor.getLabelId() );
     }
+
+    @Test
+    public void shouldAcquireSchemaReadLockBeforeGettingConstraintsByLabel() throws Exception
```
fickludd: These tests are good.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/ConstraintTestBase.java
```diff
@@ -134,4 +132,26 @@ public void shouldBeAbleCheckExistenceOfConstraints() throws Exception
...
+
+    @Test
+    public void shouldFindAllConstraints() throws Exception
+    {
+        // GIVEN
+        try ( org.neo4j.graphdb.Transaction tx = graphDb.beginTx() )
+        {
+            graphDb.schema().constraintFor( Label.label( "FOO" ) ).assertPropertyIsUnique( "prop1" ).create();
+            graphDb.schema().constraintFor( Label.label( "BAR" ) ).assertPropertyIsUnique( "prop2" ).create();
+            graphDb.schema().constraintFor( Label.label( "BAZ" ) ).assertPropertyIsUnique( "prop3" ).create();
```
fickludd: Can we add some `existence` and `node key` constraints here as well, or is there some edition problem?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -421,4 +537,6 @@ public Read dataRead()
     {
         return allStoreHolder;
     }
+
+
```
fickludd: whitelines

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -152,12 +173,107 @@ public boolean nodeAddLabel( long node, int nodeLabel ) throws EntityNotFoundExc
...
+    }
+
+    /**
+     * Check so that there is not an existing node with the exact match of label and property
+     */
+    private void validateNoExistingNodeWithExactValues( IndexBackedConstraintDescriptor constraint,
+            IndexQuery.ExactPredicate[] propertyValues, long modifiedNode
+    ) throws UniquePropertyValueValidationException, UnableToValidateConstraintException
+    {
+        try (NodeValueIndexCursor valueCursor = cursors.allocateNodeValueIndexCursor())
```
fickludd: checkstyle

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -198,7 +198,7 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
   override def indexSeek(index: IndexDescriptor, values: Seq[Any]): Iterator[NodeValue] = {
     indexSearchMonitor.indexSeek(index, values)
     val predicates = index.properties.zip(values).map(p => IndexQuery.exact(p._1, p._2))
-    val indexResult = transactionalContext.statement.readOperations().indexQuery(index, predicates: _*)
+    val indexResult = transactionalContext.statement.readOperations().indexQuery(cypherToKernel(index), predicates: _*)
```
fickludd, line 200: In this class we could use Kernel `IndexDescriptor`s directly, or whatever version of those that ends up in the Kernel API in the end. Just a note, don't do it now.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/TxStateIndexReader.java
```diff
@@ -0,0 +1,202 @@
...
+import org.neo4j.values.storable.Value;
+
+import static org.neo4j.collection.primitive.PrimitiveLongCollections.resourceIterator;
+import static org.neo4j.kernel.impl.api.StateHandlingStatementOperations.assertOnlyExactPredicates;
+
+/**
+ * An index reader that decorate the {@link #query(IndexQuery...)} method
+ * with data from the transaction state.
+ */
+class TxStateIndexReader implements IndexReader
```
fickludd: When we paired briefly, you said this was going away. Forgot?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -480,6 +481,86 @@ public void shouldGetNoIndex() throws Exception
...
+              NodeValueIndexCursor node = cursors.allocateNodeValueIndexCursor() )
+        {
+            // when
+            tx.dataWrite().nodeSetProperty( strOne, prop, stringValue("ett") );
+            tx.dataRead().nodeIndexSeek( index, node, IndexOrder.NONE, IndexQuery.exact( prop, "ett" ) );
+
+            // then
+            assertTrue( node.next() );
+        }
+    }
```
fickludd: I don't remember what other tx-state tests we have for index reading. So I'm going to list some other things I think we might be interested in testing if they are not tested in other places:

 * Deleted nodes and range/prefix scans
 * Some composite index queries
 * Added/removed label with various query types
 * Swapping a indexed node in a tx (remove prop from (a:L), add to (b:L))

fickludd (changes requested):
This is great as always. 

I have some thought around the testing the tx-state index queries, and putting the `KernelTransactionImplementation` in AllStoreHolder itches a little. Otherwise great progress!

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/TxStateIndexReader.java
pontusmelke: yes

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
pontusmelke: The abstract `Read` needs to be able to do locking. `Operations` also holds a `KernelTransactionImplementation` and they feel like siblings so I didn't think this made things worse.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
pontusmelke: Ok, I wont :)

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
fickludd: Ok, let's leave and act later if it becomes a problem.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -166,6 +183,27 @@ public void shouldPerformExactLookup() throws Exception
...
+        int firstName = token.propertyKey( "firstname" );
+        int surname = token.propertyKey( "surname" );
+        CapableIndexReference index = schemaRead.index( label, firstName, surname );
+        try ( NodeValueIndexCursor node = cursors.allocateNodeValueIndexCursor();
+              PrimitiveLongSet uniqueIds = Primitive.longSet() )
+        {
+            // when
+            IndexValueCapability valueCapability = index.valueCapability( ValueGroup.TEXT );
+            read.nodeIndexSeek( index, node, IndexOrder.NONE, IndexQuery.exact( firstName, "Joe" ),
+                    IndexQuery.exact( firstName, "Dalton" ) );
```
fickludd: `firstName` -> `surname`... it feels unsettling that this does not cause an assertion error somewhere.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/NodeValueIndexCursorTestBase.java
```diff
@@ -553,11 +613,354 @@ public void shouldFindUpdatedNodeInIndexSeek() throws Exception
...
+        int prop = token.propertyKey( "prop" );
+        CapableIndexReference index = schemaRead.index( label, prop );
+        try ( org.neo4j.internal.kernel.api.Transaction tx = session.beginTransaction();
+              NodeValueIndexCursor node = cursors.allocateNodeValueIndexCursor() )
+        {
+            // when
+            tx.dataWrite().nodeRemoveLabel( strOne, label );
+            tx.dataWrite().nodeAddLabel( strOneNoLabel, label );
+            tx.dataRead().nodeIndexSeek( index, node, IndexOrder.NONE,
+                    IndexQuery.range( prop, "ett", true, "tre", true ) );
```
fickludd: This test might be working by accident... `strOneNoLabel` has `prop = "one"`, but this test is scanning `[ett,tre]`. I think `strThree1`, `strThree2` and `strThree1` are also in that range, so if we were to change the order of returned results, this test could start failing.

fickludd (changes requested):
Now the test-coverage is really nice!

Two details, then I'm happy.

pontusmelke:
@fickludd That thing with firstname/surname mixup is somewhat unsettling but it looks like the implementation of composite seeks doesn't use the property tokens at all in the actual seek - just the values. So you can actually use whatever there, not sure how that works but it seems to have been like that for some time.

fickludd (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
```diff
@@ -471,18 +476,23 @@ public final void futureRelationshipPropertyReferenceRead( long reference )
...
     }
 
     @Override
     public boolean hasTxStateWithChanges()
     {
-        return txStateHolder.hasTxStateWithChanges();
+        return ktx.hasTxStateWithChanges();
+    }
+
+    protected void sharedOptimisticLock( ResourceType resource, long resourceId )
```
burqen: Can be package private

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -152,12 +181,119 @@ public boolean nodeAddLabel( long node, int nodeLabel ) throws EntityNotFoundExc
             return false;
         }
 
+        //Check so that that we are not breaking uniqueness constraints
+        //We do this by looking if there is an existing node in the index that
+        //with the same label and property combination.
```
burqen: //Check so that ~that~ we are not breaking uniqueness constraints
        //We do this by ~looking~checking if there is an existing node in the index ~that~
        //with the same label and property combination.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -680,22 +692,28 @@ public void addLabel( Label label )
...
         catch ( EntityNotFoundException e )
         {
             throw new NotFoundException( "No node with id " + getId() + " found.", e );
         }
-        catch ( InvalidTransactionTypeKernelException e )
+        catch ( KernelException e )
         {
             throw new ConstraintViolationException( e.getMessage(), e );
         }
+
```
burqen: Empty line

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
```diff
@@ -363,4 +470,10 @@ private String checkValidTokenName( String name ) throws IllegalTokenNameExcepti
         }
         return name;
     }
+
+    String indexGetFailure( IndexDescriptor descriptor ) throws IndexNotFoundKernelException
+    {
+        return storeReadLayer.indexGetFailure( descriptor.schema() );
+    }
+
```
burqen: Empty line

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeValueIndexCursor.java
```diff
@@ -122,8 +193,77 @@ public String toString()
         }
         else
         {
-            return "NodeValueIndexCursor[node=" + node + ", open state with: keys=" + Arrays.toString( keys ) + ", values=" + Arrays.toString( values ) +
-                    ", underlying record=" + super.toString() + " ]";
+            return "NodeValueIndexCursor[node=" + node + ", open state with: keys=" + Arrays.toString( query ) +
```
burqen: Printing the query as keys will look something like: `... keys=[IndexQuery.ExistsPredicate[propertyKeyId=0]]`. I think it is worth extracting the actual keys from the queries and only print those.

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/IndexReader.java
```diff
@@ -56,24 +57,10 @@
      * @param client the client which will control the progression though query results.
      * @param query the query so serve.
      */
-    default void query(
+    void query(
```
burqen, line 60: Really nice to get rid of this! 👍

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/IndexReader.java
```diff
@@ -117,4 +111,5 @@ public boolean hasFullNumberPrecision( IndexQuery... predicates )
             return true;
         }
     };
+
```
burqen: Empty line

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/SchemaRead.java
```diff
@@ -32,4 +37,31 @@
      * @return the IndexReference, or {@link CapableIndexReference#NO_INDEX} if such an index does not exist.
      */
     CapableIndexReference index( int label, int... properties );
+
+    /**
+     * Finds all constraints for the given schema
+     * @param descriptor The descriptor of the schema
+     * @return All constraints for the given schema
+     */
+    Iterator<ConstraintDescriptor> constraintsGetForSchema( SchemaDescriptor descriptor );
```
burqen, line 46: My understanding has been that we intend to use `IndexReference` as key for the index mapping and leave `SchemaDescriptor` as an internal thing. That's the reason why `index( int label, int... properties )` return `CapableIndexReference` and why `interface Read` use `IndexReference` everywhere.

Are we abandoning this plan? If so, `IndexReference` should be cleaned up. If not, we should be consistent about on what layer we switch from `IndexReference` to `IndexDescriptor`.

What are your thoughts about this?

burqen (changes requested):
This looks great 👍 

I have one concern about the mixed use of `SchemaDescriptor` and `IndexReference`. See comment.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/SchemaRead.java
pontusmelke: I wasn't aware of that plan. I am actually terribly confused over the distinction between `IndexDescriptor` versus `IndexReference`. This API is actually not needed for this PR we could have managed with just internal API. Could we fix this API later in a separate PR?

burqen: Yeah, I'm confused as well 😅  Let's clean this up in separate PR. I would be happy to help!

pontusmelke: Thanks, we should maybe have a `ConstraintReference` as well

burqen: At least!

burqen (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -467,10 +467,9 @@ private void assertTransactionOpen()
 
     public void assertOpen()
     {
-        Optional<Status> terminationReason = getReasonIfTerminated();
-        if ( terminationReason.isPresent() )
+        if ( terminationReason != null )
         {
-            throw new TransactionTerminatedException( terminationReason.get() );
+            throw new TransactionTerminatedException( terminationReason );
```
chrisvest: You should grab and store the `terminationReason` in a local variable, since in general a volatile field could change from one read to another.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/LabelSet.java
```diff
@@ -51,5 +53,11 @@ public boolean contains( int labelToken )
         {
             return false;
         }
+
+        @Override
+        public long[] all()
+        {
+            return new long[0];
```
chrisvest: This value could be a field in this anonymous class.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/LabelSet.java
```diff
@@ -32,6 +32,8 @@
 
     boolean contains( int labelToken );
 
+    long[] all();
```
chrisvest, line 38: Javadocs should say whether modifications are reflected in the LabelSet, and if there's any sort of order to be expected.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -223,6 +231,17 @@ public boolean nodeAddLabel( long node, int nodeLabel ) throws EntityNotFoundExc
             }
         }
 
+        //This is true if we are adding a property
+        if ( changedPropertyKeyId != NO_SUCH_PROPERTY_KEY )
```
chrisvest: Following the comment, should the parameter be called `addedPropertyKeyId`? (likewise for the `Value`)

chrisvest: Also, your static import seems to not have taken effect elsewhere in the file.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeValueClientFilter.java
```diff
@@ -177,10 +178,7 @@ else if ( property.propertyKey() < filter.propertyKeyId() )
                 }
             }
         }
-        if ( accepted < filters.length )
-        {
-            return false; // not all filters were matched
-        }
-        return target.acceptNode( reference, values );
+        // not all filters were matched
+        return accepted >= filters.length && target.acceptNode( reference, values );
```
chrisvest: The comment ends up divorced from its context, here.

community/server/src/test/java/org/neo4j/server/rest/RetrieveNodeIT.java
```diff
@@ -173,7 +173,7 @@ public void shouldGet404WhenRetrievingNonExistentNode() throws Exception
     {
         gen.get()
                 .expectedStatus( 404 )
-                .get( nodeUri + "00000" );
+                .get( nodeUri + "600000" );
```
chrisvest: This looks like it came out of nowhere?

chrisvest (approved):
I had some small comments, but it looks pretty good overall.

community/server/src/test/java/org/neo4j/server/rest/RetrieveNodeIT.java
pontusmelke: I don't know why this hit us now but this test started to be flaky. If this test happens to run when there have been no nodes in the database  the `nodeUri` will be `http://localhost:51280/db/data/node/0` so adding "00000" will not do much good as "000000" will resolve to node id 0 which is an existing node and we will not get a 404. 
So if you run only this test or if it runs first the test will fail.

The 60000 is of course completely random but it is a large number and it is unlikely we will have that many nodes in the database in this test. I updated to the less random number 12345 and added a comment for future confused readers.

PR #10627, opened 2017-12-12
Fix optional arguments in planShortestPaths

fickludd:
This bug has not been observed yet - fixing because it's easy.


pontusmelke (approved):
👍

PR #10629, opened 2017-12-12
Lower memory requirement of recovery

tinwelint:
By keeping only a single transaction in memory at any point in time. There was no performance benefit batching multiple transactions, mainly since the main benefactor of this batching would be indexing, but updates to indexes during recovery are batched in another way anyway.


tinwelint:
Note, this change actually also applies to 3.3, even if recovery there do index updates. The reason for that is that internal index updaters avoid refresh until the end of recovery anyway.

MishaDemianenko (approved):

tinwelint:
Null-merged the batching into 3.3 and forward, since there's a slightly bigger change required to not mess up performance. On the positive side w/ that extra work the performance is even better than before!

PR #10633, opened 2017-12-12
Improve server shutdown behaviour

RagnarW:
Use same shutdown procedure in RaftServer and CatchupServer.

ServerShutdown now handles errors and throws after both channel and
event loop group have been closed. It will also expose possible failure
when shutting down event loop group.

Also fix logging when unable to shutdwon event loop group within given
time


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/ServerShutdown.java
```diff
@@ -0,0 +1,110 @@
...
+                    {
+                        log.error( "Exception when shutting down event loop group", cause );
+                        throw cause;
+                    }
+                } );
+            }
+        }
+    }
+
+    private interface ErrorRunner
```
martinfurmanski: Perhaps just use Runnable?

RagnarW: Runnable will not work since it needs to throw.

andrewkerr9000: We have `ThrowingAction` which does the same but has a type parameter for the throwable type (although it extends Exception)

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/CatchupServer.java
```diff
@@ -208,28 +209,15 @@ private ChannelInboundHandler decoders( CatchupServerProtocol protocol )
...
-        {
-            return;
-        }
-
         log.info( "CatchupServer stopping and unbinding from " + listenAddress );
         try
         {
-            channel.close().sync();
-            channel = null;
+            new ServerShutdown( log, workerGroup, channel ).shutdown();
```
andrewkerr9000: Maybe use a static method, e.g. `ServerShutdown.shutdown( log, workerGroup, channel );`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/ServerShutdown.java
RagnarW: I would have to wrap the cause of the future since it is a throwable. But if this is to be a private sub interface I don't really see the point.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/CatchupServer.java
RagnarW: I agree

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/ServerShutdown.java
```diff
@@ -0,0 +1,110 @@
...
+        this.log = log;
+        this.eventLoopGroup = eventLoopGroup;
+        this.channel = channel;
+    }
+
+    public void shutdown() throws Throwable
+    {
+        try ( ErrorHandler errorHandler = new ErrorHandler() )
+        {
+            if ( channel != null )
```
lutovich: Why do we need to both close channel and shutdown the event loop group? Isn't it enough to just do the later? Or channel comes from a different group/bootstrap?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/ServerShutdown.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+            catch ( Throwable t )
+            {
+                if ( throwable == null )
+                {
+                    throwable = t;
+                }
+                else
+                {
+                    throwable.addSuppressed( t );
```
lutovich: Super edge case - this will throw if `t == throwable`

lutovich (commented):
Maybe also add couple tests for `ServerShutdown`.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/ServerShutdown.java
RagnarW: Not sure. I just kept the old logic. It should be enough according to https://netty.io/wiki/user-guide-for-4.x.html#shutting-down-your-application

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/server/ServerShutdown.java
andrewkerr9000: :+1:

PR #10635, opened 2017-12-13
More stringent shutdown

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloaderService.java
```diff
@@ -52,20 +56,38 @@ public CoreStateDownloaderService( JobScheduler jobScheduler, CoreStateDownloade
...
-        if (currentJob != null)
+        stopped = true;
+
+        if ( currentJob != null )
         {
             currentJob.stop();
         }
+        if ( jobHandle != null )
+        {
+            jobHandle.cancel( false );
```
RagnarW: The scheduler is designed to only run one task at the time so this should have no effect. Since it is either currently running or completed.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloaderService.java
```diff
@@ -52,20 +56,38 @@ public CoreStateDownloaderService( JobScheduler jobScheduler, CoreStateDownloade
...
+            try
+            {
+                jobHandle.waitTermination();
```
RagnarW: We could also wait on isComplete.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloaderService.java
```diff
@@ -52,20 +56,38 @@ public CoreStateDownloaderService( JobScheduler jobScheduler, CoreStateDownloade
...
-            jobScheduler.schedule( downloadSnapshot, currentJob );
+            jobHandle = jobScheduler.schedule( downloadSnapshot, currentJob );
         }
     }
 
     @Override
     public synchronized void stop() throws Throwable
     {
-        if (currentJob != null)
+        stopped = true;
```
RagnarW, line 72: Do we need to implement start which sets stopped=false? Or will this entire object be recreated?

RagnarW (changes requested):

martinfurmanski:
@RagnarW PTAL

RagnarW (approved):

RagnarW (changes requested):
Should perhaps make sure the new stop logic in DownloaderService is covered in test

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloaderService.java
martinfurmanski: I think that is flexible, contract-wise. We do not expect this particular object to be restartable. The service has he same lifespan as the entire database process.

RagnarW (approved):

PR #10636, opened 2017-12-13
Remove various warnings

systay:


community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/LogicalPlanConverter.scala
```diff
@@ -203,6 +203,10 @@ object LogicalPlanConverter {
           s"Failed trying to rewrite $classNameV3_3 - 3.4 class not found ($classNameV3_4)", e)
         case Failure(e: NoSuchElementException) => throw new InternalException(
           s"Failed trying to rewrite $classNameV3_3 - this class does not have a constructor", e)
+        case Failure(e: Exception) => throw new InternalException(
+          s"Failed trying to rewrite $classNameV3_3", e)
+        case Failure(e: Throwable) => throw new InternalException(
```
sherfert: I am not sure it makes sense to catch all Throwables. A Throwable that is not an Exception is usually something very bad like OutOfMemoryError, which you wouldn't like to hide in another Exception.

sherfert (commented):
I really like that you take the time to reduce the amount of warnings we have! I have only one little comment I would like you to consider.

PR #10637, opened 2017-12-13
Truncate log file non readable end

MishaDemianenko:
It can be the case that on recovery log entry file reader will be not able
to read log file till the end because LogEntry was not written completely,
for example. In those cases, we need to trim those bytes because it can happen
that if we will start appending after that to a log file and will try to
re-read this half stored record again - we will be able to read it (since
bytes will be there), but the content of that record will be absolute gibberish.
To prevent that from happening we need to make sure that our log files are
always readable from a to z and trim anything that we can't read.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/recovery/LogTailScanner.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/recovery/LogTailScanner.java
@@ -90,9 +90,11 @@ public class LogTailScanner
             oldestVersionFound = version;
             CheckPoint latestCheckPoint = null;
             try ( LogVersionedStoreChannel channel = logFiles.openForVersion( version );
-                  LogEntryCursor cursor = new LogEntryCursor( logEntryReader, new ReadAheadLogChannel( channel ) ) )
+                  ReadAheadLogChannel readAheadLogChannel = new ReadAheadLogChannel( channel );
+                  LogEntryCursor cursor = new LogEntryCursor( logEntryReader, readAheadLogChannel ) )
             {
                 LogEntry entry;
+                long maxEntryReadPosition = 0;
                 while ( cursor.next() )
                 {
                     entry = cursor.get();
@@ -124,6 +126,11 @@ public class LogTailScanner
                     {
                         latestLogEntryVersion = entry.getVersion();
                     }
+                    maxEntryReadPosition = readAheadLogChannel.position();
+                }
+                if ( hasUnreadableBytes( channel, maxEntryReadPosition ) )
+                {
+                    corruptedTransactionLogs = true;
                 }
             }
             catch ( Throwable t )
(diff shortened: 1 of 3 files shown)


ragadeeshu (commented):
I was expecting to see a test case testing the specific case described? :)

community/kernel/src/main/java/org/neo4j/kernel/recovery/LogTailScanner.java
```diff
@@ -155,6 +162,11 @@ else if ( entry instanceof LogEntryStart )
                 oldestStartEntryTransaction, oldestVersionFound, highestLogVersion, latestLogEntryVersion );
     }
 
+    private boolean hasUnreadableBytes( LogVersionedStoreChannel channel, long maxEntryReadEndPosition ) throws IOException
+    {
+        return channel.position() > maxEntryReadEndPosition;
```
tinwelint, line 167: So this is based on the assumption that the `ReadAheadLogChannel` reads ahead in chunks and if stops before reading all the bytes that it read ahead from the underlying channel, i.e. some bad entry was encountered, right?

Will that always be safe to assume? What if the entry cursor reads something bad right at the end of a chunk, before it reads ahead to the next chunk? Wouldn't we fail to recognize that if only doing this comparison?

MishaDemianenko:
@ragadeeshu well this work is result of flaky test investigation so you have test out of the box, but since that was only flaky, maybe will be able to produce more accurate scenario for this one

community/kernel/src/main/java/org/neo4j/kernel/recovery/LogTailScanner.java
MishaDemianenko: we will check for the difference between position that channel had after last successful entry read and position after read ahead cursor was exhausted. That should show the difference only when we have partial record in the end of the file only and not between chunks inside read ahead cursor.

MishaDemianenko:
@ragadeeshu additional test added

community/kernel/src/test/java/org/neo4j/kernel/RecoveryCorruptedTransactionLogIT.java
```diff
@@ -395,7 +395,31 @@ public void repetitiveRecoveryIfCorruptedLogsWithCheckpoints() throws IOExceptio
 
         while ( logFiles.getHighestLogVersion() > 0 )
         {
-            truncateBytesFromLastLogFile( 1 + random.nextInt( 100 ) );
+            int bytesToTrim = 1 + random.nextInt( 100 );
+            System.out.println("Trim: " + bytesToTrim);
```
ragadeeshu: Is this sysout here on purpose? If yes, maybe we should add a rule to suppress output?

ragadeeshu (commented):
@MishaDemianenko only 1 small question

MishaDemianenko:
@ragadeeshu never been there :), updated

ragadeeshu (approved):

PR #10640, opened 2017-12-14
Fix cancellation of Timer

martinfurmanski:
JobHandle is a wrapper on top of a future. If you cancel a future
you cannot wait for it, no matter what. Thus the abstraction now
disallows cancelling if waiting is required.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/schedule/Timer.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/schedule/Timer.java
@@ -19,8 +19,6 @@
  */
 package org.neo4j.causalclustering.core.consensus.schedule;
 
-import java.util.concurrent.CancellationException;
-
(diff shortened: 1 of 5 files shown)


phughk (approved):

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/schedule/Timer.java
```diff
@@ -124,7 +124,9 @@ public synchronized void reset()
      * then a subsequent cancel will not ensure that the first execution of the
      * handler was cancelled.
      *
-     * @param mayInterruptIfRunning Interrupt any currently running job.
+     * N.B: The job will not be interrupted when await is used!
```
andrewkerr9000: So the pair of booleans select from three possible behaviours. Maybe replace them both with a three valued Enum?

andrewkerr9000 (commented):
LGTM, but maybe the API can be improved?

martinfurmanski:
@andrewkerr9000 PTAL

PR #10643, opened 2017-12-14
Exception suppressor helper

martinfurmanski:
@RagnarW @lutovich PTAL - suggested replacement for #10633 

PR is 3.4 now, but suggest basing and integrating with the shutdown changes in #10633 if we go with this, on 3.1

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/scenarios/ExceptionSuppressor.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/scenarios/ExceptionSuppressor.java
@@ -1,0 +1,77 @@
+package org.neo4j.causalclustering.scenarios;
+
+import org.neo4j.function.ThrowingAction;
+
+/**
+ * Allows clean error handling when closing down multiple resources which are independent.
+ * One can differentiate exceptions to be caught and suppressed from those which should cause
+ * the current chain of suppressed exceptions to be thrown.
+ * <p>
+ * On {@link #close()} the chain of suppressed exceptions is thrown.
+ */
+class ExceptionSuppressor implements AutoCloseable
+{
+    private final Class defaultSuppressedClass;
+    private Throwable first = null;
+
+    public ExceptionSuppressor()
+    {
+        this( Exception.class );
+    }
+
+    public ExceptionSuppressor( Class defaultSuppressedClass )
+    {
+        this.defaultSuppressedClass = defaultSuppressedClass;
+    }
+
+    public <E extends Throwable> void execute( ThrowingAction<E> action, Class suppressedClass ) throws E
+    {
+        try
+        {
+            action.apply();
+        }
+        catch ( Throwable ex )
+        {
+            if ( first == null )
+            {
+                first = ex;
+            }
+            else
+            {
+                first.addSuppressed( ex );
+            }
+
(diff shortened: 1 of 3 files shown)


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/scenarios/ExceptionSuppressor.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Allows clean error handling when closing down multiple resources which are independent.
+ * One can differentiate exceptions to be caught and suppressed from those which should cause
+ * the current chain of suppressed exceptions to be thrown.
+ * <p>
+ * On {@link #close()} the chain of suppressed exceptions is thrown.
+ */
+class ExceptionSuppressor implements AutoCloseable
+{
+    private final Class defaultSuppressedClass;
```
lutovich, line 14: Maybe make type `Class<? extends Throwable>`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/scenarios/ExceptionSuppressor.java
```diff
@@ -0,0 +1,77 @@
...
+    {
+        this( Exception.class );
+    }
+
+    public ExceptionSuppressor( Class defaultSuppressedClass )
+    {
+        this.defaultSuppressedClass = defaultSuppressedClass;
+    }
+
+    public <E extends Throwable> void execute( ThrowingAction<E> action, Class suppressedClass ) throws E
```
lutovich, line 27: Can second param be of type `Class<E>`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/scenarios/ExceptionSuppressor.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        catch ( Throwable ex )
+        {
+            if ( first == null )
+            {
+                first = ex;
+            }
+            else
+            {
+                first.addSuppressed( ex );
```
lutovich, line 41: I'd suggest adding `first != ex` check because this is a generic utility class that does not know which actions it execute. Same exception might be thrown by different actions and self-suppressions are not allowed.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/scenarios/ExceptionSuppressor.java
```diff
@@ -0,0 +1,77 @@
...
+    {
+        execute( action, defaultSuppressedClass );
+    }
+
+    @Override
+    public void close() throws Exception
+    {
+        if ( first != null )
+        {
+            if ( first instanceof Error )
```
lutovich, line 62: Maybe always throw error wrapped in an `ExecutionException` or `CompletionException`? Basically like `Future#get()` or `CompletableFuture#getNow()` do?

lutovich (commented):
@martinfurmanski looks good to me. Added couple comments.

martinfurmanski:
@lutovich Thanks for your review, I agree with most of your points. The modules that were supposed to use this already got simplified in another way, so we have no use for it right now. Will reopen if it becomes relevant again.

PR #10645, opened 2017-12-14
Fix various problem in mixing slotted and community pipes

fickludd:
Fixes the remaining blacklisted acceptance test with a simple patch, and some problems around copying `ExecutionContext`s when using community Pipes from slotted. 

Also rename `PrimitiveExecutionContext` => `SlottedExecutionContext` and properly name the set and copyWith operations. These behave identically in interpreted, but need to be correctly distinguished for mixing in interpreted pipes in the slotted runtime. Note the apart from renaming them, some incorrect `set` calls were changed to `copyWith`.


systay (approved):
Looks really good. 

Nice change to classes and method names - easier to understand this way.

PR #10646, opened 2017-12-15
Adds more configurability to import surface

tinwelint:
Allows importer to be configured for storing empty arrays as property values, instead of ignoring them. Also adds `trim-strings` configuration option to `neo4j-admin import`.

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
@@ -40,6 +40,7 @@ import org.neo4j.helpers.collection.MapUtil;
 import org.neo4j.kernel.configuration.Config;
 import org.neo4j.kernel.impl.util.Validators;
 import org.neo4j.server.configuration.ConfigLoader;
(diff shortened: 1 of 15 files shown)


community/dbms/src/main/java/org/neo4j/commandline/dbms/ImportCommand.java
```diff
@@ -101,7 +101,9 @@ public String usage()
             .withArgument( new OptionalBooleanArg( "ignore-missing-nodes", false,
                     "If relationships referring to missing nodes should be ignored during the import." ) )
             .withArgument( new OptionalBooleanArg( "ignore-empty-arrays", Configuration.COMMAS.emptyArraysAsNull(),
-                    "If empty array properties should be ignored, or if 'false', stores as empty arrays." ) );
+                    "If empty array properties should be ignored, or if 'false', stores as empty arrays." ) )
```
praveenag: We last decided to remove this functionality altogether. Before we merge this PR I'd like to go over the decision. And see whether the reasoning still stands true.

praveenag (changes requested):

PR #10647, opened 2017-12-15
Reworked transaction batching during recovery

tinwelint:
Applies all transactions immediately instead of building up 100 at a time and apply
as batch. The upside of doing that was to do index refreshing less often,
one refresh per batch.

By avoiding refresh completely during recovery and instead do one refresh after recovery
completes avoids the need to do this batching (which can require a lot of heap for
large transactions) and also avoids refreshing entirely. This change results in
even faster recovery than before while at the same time using less memory.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/OnlineIndexProxy.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/OnlineIndexProxy.java
@@ -98,11 +98,23 @@ public class OnlineIndexProxy implements IndexProxy
     @Override
     public IndexUpdater newUpdater( final IndexUpdateMode mode )
     {
-        // If this proxy is flagged with taking extra care about idempotency then escalate ONLINE to ONLINE_IDEMPOTENT.
-        // At the time of writing this there's no other mode that needs to be escalated.
-        IndexUpdater actual = accessor.newUpdater(
-                mode == IndexUpdateMode.ONLINE && forcedIdempotentMode ? IndexUpdateMode.ONLINE_IDEMPOTENT : mode );
+        IndexUpdater actual = accessor.newUpdater( escalateModeIfNecessary( mode ) );
         return started ? updateCountingUpdater( actual ) : actual;
+    }
+
+    private IndexUpdateMode escalateModeIfNecessary( IndexUpdateMode mode )
+    {
+        if ( forcedIdempotentMode )
+        {
+            // If this proxy is flagged with taking extra care about idempotency then escalate ONLINE to ONLINE_IDEMPOTENT.
+            if ( mode != IndexUpdateMode.ONLINE )
+            {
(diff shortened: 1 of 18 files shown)


tinwelint:
Continuation of #10629

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/OnlineIndexProxy.java
```diff
@@ -97,7 +98,10 @@ public void start()
     @Override
     public IndexUpdater newUpdater( final IndexUpdateMode mode )
     {
-        IndexUpdater actual = accessor.newUpdater( forcedIdempotentMode ? IndexUpdateMode.RECOVERY : mode );
+        // If this proxy is flagged with taking extra care about idempotency then escalate ONLINE to ONLINE_IDEMPOTENT.
+        // At the time of writing this there's no other mode that needs to be escalated.
+        IndexUpdater actual = accessor.newUpdater(
```
MishaDemianenko: i think this condition moving us in the wrong direction and its a step that puts into the valley of assumptions based bugs even deeper then it was before. Just looking into what i can observe from this method signature and code around it its an accessor who should be able to create appropriate updater, based on correctly supplier accessor factory

MishaDemianenko (changes requested):
A single comment. also how big is perf improvement?

tinwelint:
I could observe a 20% improvement on a decently sized recovery w/ thousands of transactions being recovered and containing 5 indexes. What's important is that it's _not slower_ imho, given the lowered memory usage.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/OnlineIndexProxy.java
tinwelint: I don't think it makes things worse, it effectively does the exact same thing it used to do with the added explicitness around more index update modes. I can see if it'd be possible to makes this even better tho.

tinwelint:
@MishaDemianenko updates. Makes it more clear imho.

MishaDemianenko (approved):

tinwelint:
@MishaDemianenko To be able to drag this along over this long period of time and branches back and forth the only option was to reduce number of commits. Also the experimental nature of the work lead to extremely hand-wavy and random commits. The end state is good, but the way there is not of any use in commit history. Also to split this up afterwards would have been a big amount of work too.

Regarding performance: node import we've seen be 2-4x faster, relationship import, which is often dominated by `EncodingIdMapper#binarySearch` anyway we've seen between 1 and 1,5x improvements. Garbage-wise I don't have exact measurements, but previously import could easily create a couple of gigabytes a second... annoying for the garbage collector. This has been reduced to only really creating String objects for the property values... which can still be a lot given that it goes this much faster. There's an upcoming change to avoid Strings too, but that's out of this context.

PR #10648, opened 2017-12-15
Reworked importer CSV parsing

tinwelint:
For higher performance, higher parallelism and less garbage.

Instead of going through the stages w/ the queuing and passing batches,
importing is done by fully executing the importing threads in parallel.
All data structures and access patterns have been made to allow for
concurrent access, mostly by sharing nothing.

Garbage was a big concern previously and so this new importing doesn't
create objects for the entities, but instead allow visitors to observe
the different fields, directly from an underlying buffer which has been
read from the input source.


tinwelint:
Changes in this branch which makes CSV --> record ids non-deterministic may break algorithms in development for future features, so putting this on ice a bit.

community/import-tool/src/test/java/org/neo4j/tooling/ImportToolTest.java
```diff
@@ -1281,6 +1315,7 @@ private void shouldPrintReferenceLinkAsPartOfErrorMessage( List<String> nodeIds,
         }
     }
 
+    @Ignore
```
chrisvest: Should this still be ignored?

chrisvest: The test seems to pass.

community/import-tool/src/test/java/org/neo4j/tooling/ImportToolTest.java
```diff
@@ -1566,6 +1603,7 @@ public void shouldBeEquivalentToUseRawAsciiOrCharacterAsQuoteConfiguration1() th
...
```
chrisvest: This test is not passing, but maybe it just need to be updated to the new error format?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/BatchingIdGetter.java
```diff
@@ -0,0 +1,72 @@
...
+import org.neo4j.kernel.impl.store.id.RenewableBatchIdSequences;
+import org.neo4j.kernel.impl.store.id.validation.IdValidator;
+import org.neo4j.kernel.impl.store.record.AbstractBaseRecord;
+
+/**
+ * Exposes batches of ids from a {@link RecordStore} as a {@link PrimitiveLongIterator}.
+ * It makes use of {@link IdSequence#nextIdBatch(int)} (with default batch size the number of records per page)
+ * and caches that batch, exhausting it in {@link #next()} before getting next batch.
+ *
+ * TODO use the {@link RenewableBatchIdSequences} instead.
```
chrisvest: TODO?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/DataStatistics.java
```diff
@@ -19,56 +19,184 @@
...
+    public class Client implements AutoCloseable
+    {
+        private long[] counts = new long[8]; // index is relationship type id
+        private int highestTypeId;
+
+        public void increment( int typeId )
+        {
+            if ( typeId >= counts.length )
+            {
+                counts = Arrays.copyOf( counts, max( counts.length * 2, typeId ) );
```
chrisvest: This needs to be `typeId + 1`, since the parameter here is the _length_, while `typeId` is elsewhere used as an _index_.

chrisvest: And this should be tested, in case there are other places in the code that aren't fully aligned on this.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/DataImporter.java
```diff
@@ -0,0 +1,268 @@
...
+ * <ol>
+ * <li>Get {@link InputChunk chunk} of data and for every entity in it:</li>
+ * <li>Parse its data, filling current record with data using {@link InputEntityVisitor} callback from parsing</li>
+ * <li>Write record(s)</li>
+ * <li>Repeat until no more chunks from input.</li>
+ * </ol>
+ */
+public class DataImporter
+{
+    public static final String ID_PROPERTY = "__id";
```
chrisvest: `ID_PROPERTY` is never used.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
```diff
@@ -150,18 +141,13 @@ public void initialize( Input input ) throws IOException
...
-        writeMonitor = new IoMonitor( neoStore.getIoTracer() );
         idMapper = input.idMapper( numberArrayFactory );
-        idGenerator = input.idGenerator();
         nodeRelationshipCache = new NodeRelationshipCache( numberArrayFactory, config.denseNodeThreshold() );
-        nodes = input.nodes();
-        relationships = input.relationships();
-        cachedNodes = cachedForSure( nodes, inputCache.nodes( MAIN, true ) );
         Estimates inputEstimates = input.calculateEstimates( neoStore.getPropertyStore().newValueEncodedSizeCalculator() );
         sanityCheckEstimatesWithRecordFormat( inputEstimates );
         dependencies.satisfyDependencies( inputEstimates, idMapper, neoStore, nodeRelationshipCache );
```
chrisvest, line 183: There are some `System.out` and `System.err` print-outs in this class. What's that about?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/EntityImporter.java
```diff
@@ -0,0 +1,173 @@
...
+            {
+                propertyBlocks[i] = new PropertyBlock();
+            }
+        }
+        return propertyBlocks[propertyBlocksCursor++];
+    }
+
+    private void encodeProperty( PropertyBlock block, int key, Object value )
+    {
+        // TODO: dynamic record ids, batching of those
```
chrisvest: TODO?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/DataImporter.java
```diff
@@ -0,0 +1,268 @@
...
+            throws IOException
+    {
+        LongAdder roughEntityCountProgress = new LongAdder();
+        ExecutorService pool = Executors.newFixedThreadPool( numRunners,
+                new NamedThreadFactory( title + "Importer" ) );
+        IoMonitor writeMonitor = new IoMonitor( stores.getIoTracer() );
+        ControllableStep step = new ControllableStep( title, roughEntityCountProgress, Configuration.DEFAULT,
+                writeMonitor, memoryStatsProvider );
+        StageExecution execution = new StageExecution( title, null, Configuration.DEFAULT, Collections.singletonList( step ), 0 );
+        InputIterator dataIterator = data.iterator();
```
chrisvest: The `dataIterator` will be closed by each of the `ExhaustingEntityImporterRunnable`s. So it will be closed `numRunners` times. I think we should instead close it at this level, with a `try-with-resources` that spans until the executor has been shut down.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/DataImporter.java
```diff
@@ -0,0 +1,268 @@
...
+
+        return roughEntityCountProgress.sum();
+    }
+
+    public static void importNodes( int numRunners, Input input, BatchingNeoStores stores, IdMapper idMapper,
+            NodeRelationshipCache nodeRelationshipCache, ExecutionMonitor executionMonitor, Monitor monitor )
+            throws IOException
+    {
+        importData( NODE_IMPORT_NAME, numRunners, input.nodes(), stores, () ->
+            new NodeImporter( stores, idMapper, monitor ), executionMonitor, new MemoryUsageStatsProvider( stores, idMapper ) );
```
chrisvest: How about having methods for creating these lambdas? The `NodeImporter` one is not so bad, but the `RelationshipImporter` involves a lot of fields.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/LabelIndexWriterStep.java
```diff
@@ -62,7 +60,7 @@ protected void process( Batch<InputNode,NodeRecord> batch, BatchSender sender )
     @Override
     public void close() throws Exception
     {
-        super.close();
         writer.close();
+        super.close();
```
chrisvest: The `super.close()` method will close the executor, after which the task queues will be empty and all workers will have terminated. If we close the `writer` before we close the executor, we can risk racing with on-going work, can we not?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
```diff
@@ -748,15 +682,15 @@ private long dataValue( long index )
         return clearCollision( dataCache.get( trackerCache.get( index ) ) );
     }
 
-    private long findIndex( LongArray array, long value )
+    private long findCollisionIndex( long value )
     {
         // can't be done on unsorted data
-        long low = 0;
-        long high = highestSetIndex;
+        long low = 0 + 0;
```
chrisvest: Just `0` will do.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
```diff
@@ -37,26 +39,28 @@
...
 import static org.neo4j.helpers.Numbers.safeCastLongToInt;
+import static org.neo4j.helpers.Numbers.safeCastLongToShort;
 import static org.neo4j.unsafe.impl.batchimport.Utils.unsignedCompare;
 import static org.neo4j.unsafe.impl.batchimport.Utils.unsignedDifference;
 import static org.neo4j.unsafe.impl.batchimport.cache.idmapping.string.ParallelSort.DEFAULT;
-import static org.neo4j.unsafe.impl.batchimport.cache.idmapping.string.SourceInformation.encodeSourceInformation;
 
 /**
  * Maps arbitrary values to long ids. The values can be {@link #put(Object, long, Group) added} in any order,
  * but {@link #needsPreparation() needs} {@link #prepare(InputIterable, Collector, ProgressListener) preparation}
```
chrisvest: This part of the javadocs needs to be updated as well.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
```diff
@@ -344,13 +315,15 @@ private static boolean isCollision( long eId )
 
         private int numberOfCollisions;
         private int localProgress;
+        private final LongFunction<Object> inputIdLookup;
```
chrisvest: This field is not used.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
```diff
@@ -609,60 +567,50 @@ public long dataValue( long nodeId )
...
+            }
 
-                previousEid = eid;
-                previousGroupId = groupId;
+            // Potential duplicate
+            // We cast the collision index to an int here. This means that we can't support > int-range
+            // number of collisions. But that's probably alright since the data structures and
+            // actual collisions values for all these collisions wouldn't fit in a heap anyway.
+            Object inputId = collisionValues.get( safeCastLongToInt( collisionIndex ) );
+            int detectorIndex = detector.add( inputId );
```
chrisvest, line 585: The `add` method loops through its `inputIdArray`, and we are calling it in a loop over the collisions. So we have a quadratic algorithm here. How likely is that to be a problem?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/StringEncoder.java
```diff
@@ -127,7 +127,13 @@ private void reMap( byte[] bytes, int inputLength )
...
-                reMap[bytes[i]] = (byte) (numChars++ % 256);
+                synchronized ( this )
+                {
+                    if ( reMap[bytes[i]] == -1 )
+                    {
+                        reMap[bytes[i]] = (byte) (numChars++ % 256);
+                    }
+                }
             }
             bytes[i] = reMap[bytes[i]];
```
chrisvest, line 138: This double-check locking will not work, because the `reMap[bytes[i]]` loads are not volatile. There is also no coordination between the load in the if-condition, and the load that assigns to `bytes[i]`. The value at that array index could be changed by another thread in between the two loads.

Documentation and testing of this new multi-threaded behaviour is also missing.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/Tracker.java
```diff
@@ -30,7 +30,15 @@
...
  * any order and so in the end when all ids have been added the index goes through a
+<<<<<<< HEAD
  * {@link IdMapper#prepare(InputIterable, Collector, ProgressListener) prepare phase} where these ids are sorted
+=======
+<<<<<<< HEAD
+ * {@link IdMapper#prepare(LongFunction, Collector, ProgressListener) prepare phase} where these ids are sorted
+=======
+ * {@link IdMapper#prepare(java.util.function.LongFunction, Collector, ProgressListener) prepare phase} where these ids are sorted
+>>>>>>> Rewrite of how data from Input gets written into records
+>>>>>>> c82c373... Rewrite of how data from Input gets written into records
```
chrisvest: Your merge conflicts are showing.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/ByteBufferReadableChannel.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+import org.neo4j.kernel.impl.transaction.log.ReadableClosableChannel;
+
+/**
+ * TODO a bit overkill to wrap a byte[] -> {@link ByteBuffer} -> {@link ReadableClosableChannel}?
+ */
+public class ByteBufferReadableChannel implements ReadableClosableChannel
```
chrisvest: Judging by usages, it seems that ValueType is working (by proxy) on ByteBuffers anyway, rather than real channels. Maybe that should be surfaced in the parameter lists to the read and write methods on ValueType? Then you wouldn't need these classes, but you'd have to fix that (as far as I have found) one place where a "real" channel is being worked on.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/Groups.java
```diff
@@ -28,9 +28,11 @@
  */
 public class Groups
 {
+    public static final int MAX_NUMBER_OF_GROUPS = 0x10000;
```
chrisvest: `MAX_NUMBER_OF_GROUPS` isn't used anywhere.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputCache.java
```diff
@@ -250,4 +233,20 @@ public void close() throws IOException
             fs.deleteFile( file( RELATIONSHIPS_HEADER, subType ) );
         }
     }
+
+    static ByteBuffer newChunkHeaderBuffer()
+    {
+        return ByteBuffer.allocate( Integer.BYTES );
+    }
+
+    static String sample( ByteBuffer buffer )
```
chrisvest: The `sample` method is debugging leftovers?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputCache.java
```diff
@@ -128,47 +128,33 @@
     private final FileSystemAbstraction fs;
     private final File cacheDirectory;
     private final RecordFormats recordFormats;
-    private final Configuration config;
```
chrisvest, line 131: The `NO_ENTITIES` and `END_OF_CACHE` constants are no longer used.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityCacher.java
```diff
@@ -44,212 +41,263 @@
...
-            channel.putInt( (Integer) key );
+            int position = buffer.position();
+            if ( position + withSufficientSpaceFor >= buffer.capacity() )
+            {
+                array = Arrays.copyOf( array, array.length * 2 ); // double in size
+                buffer = ByteBuffer.wrap( array );
+                buffer.position( position );
+                bufferAsChannel = new ByteBufferFlushableChannel( buffer );
+            }
+            return buffer;
```
chrisvest: This will work because the chunk size is so large to begin with, and this method is only called with small constants. But there is bug-potential here, where `position + withSufficientSpaceFor` could in principle be greater than twice the buffer capacity. Also, `withSufficientSpaceFor` could be called `requiredSpace`, which is shorter.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityCacher.java
```diff
@@ -44,212 +41,263 @@
...
+            int length = type.length( value );
+            buffer( 1 + length ).put( type.id() );
+            try
+            {
+                type.write( value, bufferAsChannel );
+            }
+            catch ( Exception e )
+            {
+                throw e;
+            }
```
chrisvest: Catch and immediately re-throw?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityCacher.java
```diff
@@ -44,212 +41,263 @@
...
+        }
 
-        header.put( END_OF_HEADER );
-        // This is a special value denoting the end of the stream. This is done like this since
-        // properties are the first thing read for every entity.
-        channel.putShort( END_OF_ENTITIES );
+        private String stringify( Object value )
+        {
+            return value.getClass().isArray() ? Arrays.toString( (Object[]) value ) : value.toString();
+        }
```
chrisvest: Debugging leftovers?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityCacher.java
```diff
@@ -44,212 +41,263 @@
...
+import static org.neo4j.unsafe.impl.batchimport.input.InputCache.newChunkHeaderBuffer;
 
 /**
- * Abstract class for caching {@link InputEntity} or derivative to disk using a binary format.
- * Currently each token type limited to have as maximum {#link Integer.MAX_VALUE} items.
+ * Abstract class for caching entities or derivative to disk using a binary format.
+ * Currently each token type is limited to have a maximum of {#link Integer.MAX_VALUE} items.
  */
-abstract class InputEntityCacher<ENTITY extends InputEntity> implements Receiver<ENTITY[],IOException>
+abstract class InputEntityCacher implements InputCacher
```
chrisvest: Please add more documentation of what is going on here, and how it's thread-safe. This class is hard to follow.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
```diff
@@ -19,204 +19,291 @@
...
     }
 
-    private int collectiveNumberOfKeys( Object[] properties, Object[] otherProperties )
+    @Override
+    public boolean labels( String[] labels )
     {
-        int collidingKeys = 0;
-        for ( int i = 0; i < properties.length; i += 2 )
+        checkClear();
+        for ( String label : labels )
```
chrisvest: `Collections.addAll`

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
```diff
@@ -19,204 +19,291 @@
...
+        endIdGroup = Group.GLOBAL;
+        hasIntType = false;
+        intType = -1;
+        stringType = null;
+    }
+
+    @Override
+    public InputEntity clone() throws CloneNotSupportedException
+    {
+        InputEntity clone = new InputEntity();
```
chrisvest: The clone instance should always be acquired from calling `super.clone()`. Once you have that, you only need replace the array list instances with their clones in turn. Take a look at the `Object.clone()` javadocs.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
```diff
@@ -19,204 +19,291 @@
...
+        clone.objectStartId = objectStartId;
+        clone.startIdGroup = startIdGroup;
+        clone.hasLongEndId = hasLongEndId;
+        clone.longEndId = longEndId;
+        clone.objectEndId = objectEndId;
+        clone.endIdGroup = endIdGroup;
+        clone.hasIntType = hasIntType;
+        clone.intType = intType;
+        clone.stringType = stringType;
+        return clone;
```
chrisvest: This implementation of `clone` does not retain the delegate field. Is that intended?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityCacher.java
```diff
@@ -44,212 +41,263 @@
...
+        protected ByteBuffer buffer;
+        private FlushableChannel bufferAsChannel;
+        private final int[] previousGroupIds = new int[2];
+
+        SerializingInputEntityVisitor( InputEntityVisitor actual, int chunkSize )
         {
-            channel.put( SAME_GROUP );
+            super( actual );
+            this.lengthThreshold = chunkSize;
+            this.array = new byte[chunkSize + chunkSize / 10]; // some wiggle room
```
chrisvest: Same as `chunkSize / 5`?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityReader.java
```diff
@@ -161,159 +124,114 @@ private void readHeader( StoreChannel header ) throws IOException
...
-            // Read next entity
-            Object properties = readProperties( channel );
-            if ( properties == null )
-            {
-                return null;
-            }
-
-            return readNextOrNull( properties, state );
+            channel.close();
+            closeAction.run();
```
chrisvest: Hmm… The `StandardOpenOption.DELETE_ON_CLOSE` would have been a perfect alternative, had our `FileSystemAbstraction` supported them.

chrisvest: Except for the maintenance of that `subtype` set, whatever that is.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityReader.java
```diff
@@ -161,159 +124,114 @@ private void readHeader( StoreChannel header ) throws IOException
...
+                break;
+            case 0:
+                break;
+            default:
+                for ( int i = 0; i < count; i++ )
+                {
+                    Object token = readToken( PROPERTY_KEY_TOKEN );
+                    Object value = readValue( );
+                    // TODO fix this some other way
+                    if ( token instanceof Integer )
```
chrisvest: What are the plans for this TODO?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityReader.java
```diff
@@ -161,159 +124,114 @@ private void readHeader( StoreChannel header ) throws IOException
...
+            {
+            // This is a special value denoting the end of the stream. This is done like this since
+            // properties are the first thing read for every entity.
+            case END_OF_ENTITIES:
+                return false;
+            case HAS_FIRST_PROPERTY_ID:
+                visitor.propertyId( channel.getLong() );
+                break;
+            case 0:
+                break;
```
chrisvest: The `default` case seems to handle `0` just as well. That brings us down to 3 cases, at which point a switch-statement might be overkill on the ceremony.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityReader.java
```diff
@@ -161,159 +124,114 @@ private void readHeader( StoreChannel header ) throws IOException
...
         {
-            throw new InputException( "Couldn't get position from cached input data", e );
-        }
-    }
+            int id = channel.getInt();
+            if ( id == -1 )
+            {
+                // This is a real token id
+                return channel.getInt();
+            }
```
chrisvest: When does this happen?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityReader.java
```diff
@@ -51,104 +40,78 @@
...
 
 /**
  * Abstract class for reading cached entities previously stored using {@link InputEntityCacher} or derivative.
  * Entity data is read in batches, each handed off to one ore more processors which interprets the bytes
- * into {@link InputEntity} instances. From the outside this is simply an {@link InputIterator},
+ * into input data. From the outside this is simply an {@link InputIterator},
  * the parallelization happens inside.
  */
-abstract class InputEntityReader<ENTITY extends InputEntity> extends InputIterator.Adapter<ENTITY>
+abstract class InputEntityReader implements InputIterator
```
chrisvest: This class could be named `InputEntityCacheReader`, and the `InputEntityCacher` could be called `InputEntityCacheWriter`, to indicate that they are siblings around this cache format.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInputChunk.java
```diff
@@ -0,0 +1,297 @@
...
+        this.seeker = seeker;
+        this.header = header;
+        this.entries = header.entries();
+        this.decorator = decorator;
+        this.visitor = null;
+        if ( header.entries().length == 0 )
+        {
+            return false;
+        }
+        return true;
```
chrisvest: The return could be simplified to the expression.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/Decorator.java
```diff
@@ -22,9 +22,9 @@
 import java.util.function.Function;
 
 import org.neo4j.graphdb.Resource;
-import org.neo4j.unsafe.impl.batchimport.input.InputEntity;
+import org.neo4j.unsafe.impl.batchimport.input.InputEntityVisitor;
 
-public interface Decorator<ENTITY extends InputEntity> extends Function<ENTITY,ENTITY>, Resource
+public interface Decorator extends Function<InputEntityVisitor,InputEntityVisitor>, Resource
```
chrisvest, line 27: This implements `Resource` to get a `close` method, which we then provide a default empty implementation of. But we don't actually use this feature anywhere. Let's just not make it extend `Resource`.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/Header.java
```diff
@@ -37,8 +39,12 @@
          * to read at the very top of it.
          * @param configuration {@link Configuration} specific to the format of the data.
          * @param idType type of values we expect the ids to be.
+         * @param groups {@link Groups} to register groups in.
+         * @return the created {@link Header}.
          */
-        Header create( CharSeeker dataSeeker, Configuration configuration, IdType idType );
+        Header create( CharSeeker dataSeeker, Configuration configuration, IdType idType, Groups groups );
+
+        boolean isDefined();
```
chrisvest: What does this mean?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/ByteBufferFlushableChannel.java
```diff
@@ -27,7 +27,7 @@
 import org.neo4j.kernel.impl.transaction.log.ReadableClosableChannel;
 
 /**
- * TODO a bit overkill to wrap a byte[] -> {@link ByteBuffer} -> {@link ReadableClosableChannel}?
+ * Wraps a byte[] -> {@link ByteBuffer} -> {@link ReadableClosableChannel}?
```
chrisvest: These two still have question marks at the end.

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/input/RandomEntityDataGenerator.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private static String randomRelationshipType( Random random, Distribution<String> relationshipTypes )
+    {
+        return relationshipTypes.random( random );
+    }
+
+    private static Object randomProperty( Entry entry, Randoms random )
+    {
+        // TODO crude way of determining value type
+        String type = entry.extractor().name();
```
chrisvest: Yeah, perhaps an enum would be better.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInputIterator.java
```diff
@@ -0,0 +1,135 @@
...
+    CsvInputIterator( CharReadable stream, Decorator decorator, Header.Factory headerFactory, IdType idType,
+            Configuration config, Groups groups, Collector badCollector, Extractors extractors ) throws IOException
+    {
+        this( stream, decorator, extractHeader( stream, headerFactory, idType, config, groups ), config, idType, badCollector, extractors );
+    }
+
+    static Header extractHeader( CharReadable stream, Header.Factory headerFactory, IdType idType,
+            Configuration config, Groups groups ) throws IOException
+    {
+        if ( !headerFactory.isDefined() )
```
chrisvest, line 86: Looks like this is only ever `false` in testing. Is it necessary to have this?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/RecordDataAssembler.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public boolean append( RecordCursor<RECORD> cursor, RECORD[] array, long id, int index )
+    {
+        RECORD record = array[index];
+        if ( !cursor.next( id, record, RecordLoad.CHECK ) || (filter != null && !filter.test( record )) )
+        {
+            return false;
+        }
+        return true;
```
chrisvest: This can be simplified to return an expression. And the null-check could be avoided by initialising the `filter` field to `Predicates.alwaysTrue` if no filter is otherwise given in the constructor.

community/csv/src/main/java/org/neo4j/csv/reader/Chunker.java
```diff
@@ -0,0 +1,57 @@
...
+ * processors, each processing chunks of data.
+ */
+public interface Chunker extends Closeable
+{
+    /**
+     * @return a new allocated {@link Chunk} which is to be later passed into {@link #nextChunk(Chunk)}
+     * to fill it with data. When a {@link Chunk} has been fully processed then it can be passed into
+     * {@link #nextChunk(Chunk)} again to get more data.
+     */
+    Chunk newChunk();
```
chrisvest, line 40: Looks like `newChunk` is only really used in tests, and `EagerParserChunker` doesn't support it anyway?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInputIterator.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public boolean next( InputChunk chunk ) throws IOException
+    {
+        if ( config.multilineFields() )
+        {
+            EagerlyReadInputChunk csvChunk = (EagerlyReadInputChunk) chunk;
+            return chunker.nextChunk( csvChunk );
+        }
+
+        CsvInputChunk csvChunk = (CsvInputChunk) chunk;
```
chrisvest: These casts are quite smelly. They seem to indicate that this chunk abstraction doesn't really work here.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/EagerlyReadInputChunk.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Override
+    public void close() throws IOException
+    {
+    }
+
+    @Override
+    public char[] data()
+    {
+        throw new UnsupportedOperationException();
```
chrisvest: All the methods from `Source.Chunk` are unsupported. Why is it implementing this interface? 🤔

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
```diff
@@ -19,204 +19,373 @@
...
+    public Object id()
     {
-        Collection<Pair<String,?>> fields = new ArrayList<>();
-        toStringFields( fields );
+        return hasLongId ? longId : objectId;
+    }
+
+    public Object endId()
+    {
+        return hasLongEndId ? longEndId : objectEndId;
```
chrisvest, line 220: Could the auto-box be avoided here? Or will exposing getters for each of the `has*`, `long*` and `object*` fields just move the same auto-boxing further up the stack?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityCacher.java
chrisvest: If possible, I'd recommend separating domain logic and multi-threading concerns in the code. Things are always more complicated when the two are entangled.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/ByteBufferReadableChannel.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+import org.neo4j.kernel.impl.transaction.log.ReadableClosableChannel;
+
+/**
+ * Wraps a byte[] -> {@link ByteBuffer} -> {@link ReadableClosableChannel}?
+ */
+public class ByteBufferReadableChannel implements ReadableClosableChannel
```
chrisvest: Judging by usages, it seems that ValueType is working (by proxy) on ByteBuffers anyway, rather than real channels. Maybe that should be surfaced in the parameter lists to the read and write methods on ValueType? Then you wouldn't need these classes, but you'd have to fix that (as far as I have found) one place where a "real" channel is being worked on.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/DataImporter.java
```diff
@@ -0,0 +1,291 @@
...
+
+        return roughEntityCountProgress.sum();
+    }
+
+    public static void importNodes( int numRunners, Input input, BatchingNeoStores stores, IdMapper idMapper,
+            ExecutionMonitor executionMonitor, Monitor monitor )
+                    throws IOException
+    {
+        importData( NODE_IMPORT_NAME, numRunners, input.nodes(), stores, () ->
+            new NodeImporter( stores, idMapper, monitor ), executionMonitor, new MemoryUsageStatsProvider( stores, idMapper ) );
```
chrisvest: How about having methods for creating these lambdas? The NodeImporter one is not so bad, but the RelationshipImporter involves a lot of fields.

chrisvest (changes requested):
I had a couple of comments. Please address them in _new_ commits.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/BatchingIdGetter.java
tinwelint: Removed TODO note, not really applicable.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/DataImporter.java
tinwelint: You mean introduce variable? Sure

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
tinwelint: The import tool generally prints things to `out`, `err`, although this is a bit down the stack so it could be done using a monitor instead I'd guess.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/LabelIndexWriterStep.java
tinwelint: In a happy scenario (non-panic) this will not be a problem since `close()` will be called after all work has completed. I don't mind changing this order anyway.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/DataImporter.java
chrisvest: Variable or method. Either works.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/ImportLogic.java
chrisvest: At least the PrintStreams should be passed in, instead of referencing `System.out/err` directly.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
tinwelint: In normal cases there are, say 0 to a hand full collisions for some of the values. I wouldn't consider that a problem for 99.99% of all use cases.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/StringEncoder.java
tinwelint: It's somewhat unrelated to this PR though, I can create a card for it

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/ByteBufferReadableChannel.java
tinwelint: I know I struggled a bit with that back and forth... I'd be happy to let it rest, but the wrapping is a bit annoying, even though it's just a very peripheral thing in the outer corner of the import code.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
tinwelint: And create a new object? Perhaps it'll not be an actual instance after escape analysis, or?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
tinwelint: I think it would unfortunately. This isn't part of the main paths anyway.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
tinwelint: Cool, didn't know that.

tinwelint: Although that will force those lists to be non-final, won't it?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
tinwelint: Hmm, don't think it's used actually. The clone() method that is.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityCacher.java
tinwelint: More like `chunkSize` plus `chunkSize/10`

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityCacher.java
tinwelint: Some old remnant I believe.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityReader.java
tinwelint: Turns out that this is old unused stuff anyway

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityReader.java
tinwelint: None, really.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntityReader.java
tinwelint: When the cached input already contains real token ids, they are allowed to.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInputIterator.java
tinwelint: I do think that there's one external use of this in the ldbc importer as well tho.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInputIterator.java
tinwelint: This will be slightly rewritten, but as with an API which has explicit allocation like:

````
MyInterface
{
    MyThing allocateAThing();

    void fillMyThingWithSomething( MyThing thing );
}
````

This cast comes by design, doesn't it?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/Decorator.java
tinwelint: Again this is for an external use case in the ldbc importer which I'm not sure we can live without.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/EagerlyReadInputChunk.java
tinwelint: It's this weird thing where there's the normal lazy chunk/parser and then we have to, against our will, support the multi-line fields feature and it's conflation of newline with record delimiters... this changes how parsing can be done significantly, for that case only... but it needs to fit in with the rest at the same time.

I'd guess this is a turd from that behind. I'm sure it can be done better, I just don't see it clearly a.t.m.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/csv/Header.java
tinwelint: Added javadoc

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/input/RandomEntityDataGenerator.java
tinwelint: Dunno if it's worth adding that, since it's only test code caring about it a.t.m. OK to just remove the TODO?

tinwelint:
@chrisvest thanks, addressed most comments

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
chrisvest: `Collections.addAll` is a static method that iterates an array and adds all the array elements to a given collection. The same as what this code does.

chrisvest (approved):

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/input/InputEntity.java
tinwelint: Cheers, I'll use that one

PR #10650, opened 2017-12-15
Data snapshot query execution

MishaDemianenko:
Introduce possibility to execute query based on data snapshot.
Any data modification will mark corresponding pages with a version that
will be equal to transaction id in which that particular change was introduced.
Also allow count store to track transactions in which changes where introduced.

To be able to guarantee that query result is based on a data that was
present on query execution start and was not modified while it was running,
the engine will verify that all paged that were accessed during query
execution have the version that is less or equal that was last closed
on a moment when query started.

Support for snapshot query execution in CC and HA

Snapshot query execution is enabled by 'unsupported.dbms.query.snapshot'
setting and number of retries that engine will do while trying to get
stable snapshot is controlled by 'unsupported.dbms.query.snapshot.retries'.


community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPage.java
```diff
@@ -47,6 +50,10 @@
 
     private long pointer;
 
+    // max transaction id that updated this page
+    @SuppressWarnings( "unused" )
+    private volatile long lastModifiedTxId = UNBOUND_LAST_MODIFIED_TX_ID;
```
tinwelint: Since you're changing this field in some cases with `Unsafe`, could you take that all the way and always change/read it with `Unsafe` and remove `volatile`?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPage.java
```diff
@@ -84,6 +91,32 @@ public long address()
         return pointer;
     }
 
+    public long getLastModifiedTxId()
+    {
+        return lastModifiedTxId;
+    }
+
+    public long resetLastModifiedTxId()
```
tinwelint: Would be nice if the name mentioned that it returned the txId, like `getAndResetLastModifiedTxId`

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCache.java
```diff
@@ -196,6 +197,7 @@
     // 'true' (the default) if we should print any exceptions we get when unmapping a file.
     private boolean printExceptionsOnClose;
     private PageCursorTracerSupplier pageCursorTracerSupplier;
+    private VersionContextSupplier versionContextSupplier;
```
tinwelint: Can be final?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -65,6 +66,7 @@
     private long pointer;
     private int pageSize;
     private int filePageSize;
+    protected VersionContext versionContext;
```
tinwelint: Can be final?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPagedFile.java
```diff
@@ -609,6 +610,25 @@ MuninnPage evictPage( long filePageId )
         return (MuninnPage) element;
     }
 
+    void setHighestEvictedTransactionId( long modifiedTransactionId )
```
tinwelint: Looks like a very similar method as `MuninnPage#setLastModifiedTxId`, could be made a utility with nice javadoc?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPagedFile.java
```diff
@@ -74,6 +75,10 @@
     // Used to trace the causes of any exceptions from getLastPageId.
     private volatile Exception closeStackTrace;
 
+    // max modifier transaction id among evicted pages for this file
+    private static final long evictedTransactionIdOffset = UnsafeUtil.getFieldOffset( MuninnPagedFile.class, "highestEvictedTransactionId" );
+    private volatile long highestEvictedTransactionId;
```
tinwelint: Also same note about `Unsafe` vs `volatile` as in `MuninnPage`

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnReadPageCursor.java
```diff
@@ -60,10 +62,24 @@ public boolean next() throws IOException
...
         return true;
     }
 
+    private void verifyContext()
+    {
+        if ( versionContext.lastClosedTransactionId() == Long.MAX_VALUE )
+        {
+            return;
+        }
+        if ( page.getLastModifiedTxId() > versionContext.lastClosedTransactionId() ||
```
tinwelint: Would be very nice with human wording about what these conditions actually mean and how they play out.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnReadPageCursor.java
```diff
@@ -60,10 +62,24 @@ public boolean next() throws IOException
...
+             pagedFile.getHighestEvictedTransactionId() > versionContext.lastClosedTransactionId() )
```
tinwelint: Also, read `versionContext.lastClosedTransactionId()` once instead of once per condition?

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/context/EmptyVersionContext.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public void initRead()
+    {
+    }
+
+    @Override
+    public void initWrite( long committingTransactionId )
+    {
+
```
tinwelint: OCD newline

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/context/EmptyVersionContext.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public long lastClosedTransactionId()
+    {
+        return Long.MAX_VALUE;
+    }
+
+    @Override
+    public void markAsDirty()
+    {
+
```
tinwelint: OCD newline

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCacheTest.java
```diff
@@ -349,4 +565,82 @@ public void mustThrowIfMappingFileWouldOverflowReferenceCount() throws Exception
...
+        @Override
+        public boolean isDirty()
+        {
+            return dirty;
+        }
+
+        @Override
+        public void clear()
+        {
+
```
tinwelint: OCD newline

tinwelint (changes requested):

community/cypher/cypher/src/test/java/org/neo4j/cypher/internal/javacompat/EagerResultITTest.java
```diff
@@ -0,0 +1,345 @@
...
+import org.neo4j.test.TestGraphDatabaseFactory;
+import org.neo4j.test.rule.TestDirectory;
+
+import static org.hamcrest.Matchers.hasSize;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+
+public class EagerResultITTest
```
chrisvest: We have both IT and Test in the name. Probably one will be enough?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPage.java
chrisvest: And if a delegate for `getAndSetLong` is added to `UnsafeUtil`, then this method could be implemented in terms of it.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnReadPageCursor.java
chrisvest: If `MuninnPageCursor.pageFault` were to set the `lastModifiedTxId` to that of `pagedFile.getHighestEvictedTransactionId`, then we could omit the "highest evicted" check here, could we not? If so, then that would be one less volatile read.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageEvictionCallback.java
```diff
@@ -38,5 +38,6 @@ public void onEvict( long filePageId, Page page )
         assert removed == page :
                 "Removed unexpected page when cleaning up translation table for filePageId " + filePageId + ". " +
                         "Evicted " + page + " but removed " + removed + " from the translation table.";
+        file.setHighestEvictedTransactionId( removed.resetLastModifiedTxId() );
```
chrisvest: I think we should do this before we `evictPage`, so there is no gap where "highest evicted" is lower than it should be.

chrisvest: Oh, wait. We need the removed page object. In that case, I think this logic should be pushed down into `file.evictPage` so that we can capture this value before we remove the page from the translation table.

chrisvest: I think this is important because while we are evicting the page, it is exclusively locked, so no one else can touch it, which in turn means that the "last modified tx id" will not change while we are evicting. The exclusive lock in turn means that as long as it remains in the translation table, no thread will initiate a page fault of that translation table entry, and therefor no thread will need the "last modified tx id" stored in the page object. So, by moving the tx id from the page object to the paged file, just before we remove the translation table entry, will ensure that no thread will miss out on any latest tx id.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -285,41 +287,20 @@ boolean applicable( DiagnosticsPhase phase )
...
-            AvailabilityGuard availabilityGuard,
-            SystemNanoClock clock,
-            AccessCapability accessCapability,
-            StoreCopyCheckPointMutex storeCopyCheckPointMutex,
-            IdController idController )
+            StartupStatisticsProvider startupStatistics, Guard guard, CommitProcessFactory commitProcessFactory,
+            AutoIndexing autoIndexing, PageCache pageCache, ConstraintSemantics constraintSemantics, Monitors monitors,
+            Tracers tracers, Procedures procedures, IOLimiter ioLimiter, AvailabilityGuard availabilityGuard,
+            SystemNanoClock clock, AccessCapability accessCapability, StoreCopyCheckPointMutex storeCopyCheckPointMutex,
+            IdController idController, VersionContextSupplier versionContextSupplier )
```
chrisvest: It becomes harder to see what's added or removed, when long parameter lists are collapsed like this.

community/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/ConfiguringPageCacheFactory.java
```diff
@@ -54,12 +56,15 @@
      * @param config page swapper configuration
      * @param pageCacheTracer global page cache tracer
      * @param pageCursorTracerSupplier supplier of thread local (transaction local) page cursor tracer that will provide
-     * thread local page cache statistics
+* thread local page cache statistics
```
chrisvest: Accidental removal of indentation.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/counts/CountsTracker.java
```diff
@@ -84,46 +85,16 @@
...
-            @Override
-            public void rotationFailed( File source, File target, Headers headers, Exception e )
-            {
-                log.error( format( "Failed to rotate counts store at transaction %d to [%s], from [%s].",
-                        headers.get( FileVersion.FILE_VERSION ).txId, target, source ), e );
-            }
-        }, new RotationTimerFactory( clock,
-                config.get( counts_store_rotation_timeout ).toMillis() ), 16, 16, HEADER_FIELDS );
+        super( fs, pages, baseFile, new CountsTrackerRotationMonitor( logProvider ), new RotationTimerFactory( clock,
+                config.get( counts_store_rotation_timeout ).toMillis() ), versionContextSupplier,16, 16, HEADER_FIELDS );
```
chrisvest: Missing a space in the parameter list, but otherwise 👍

chrisvest (changes requested):
Looks good, except I'd like you take a look at my suggestions around eviction and faulting.

chrisvest:
A note on forward merging: with the PageList being off-heap, we can choose to only spend the space on the version field, if versioning is enabled. Alternatively we could have a separate VersionList, which is only initialised when needed. The memory usage is slightly changed either way, however, which might be a problem for some tests in 3.4.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
MishaDemianenko: well, the problem with that formatting is when you do not really introduce new parameter manually it's reformatted, but will return it to its previous form here since 3.2 and things...

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPage.java
chrisvest: I like keeping the `volatile` modifier as a reminder that this is the mode of access used on this field.

MishaDemianenko: i would like (and actually did that), to keep it as volatile for exactly the same reason :)

community/unsafe/src/main/java/org/neo4j/unsafe/impl/internal/dragons/UnsafeUtil.java
```diff
@@ -307,6 +307,32 @@ public static Object getAndSetObject( Object obj, long offset, Object newValue )
...
+
+    /**
+     * Atomically set field or array element to a maximum between current value and provided <code>newValue</code>
+     */
+    public static void compareAndSetMaxLong( Object object, long fieldOffset, long newValue )
+    {
+        long currentValue;
+        do
+        {
+            currentValue = UnsafeUtil.getLong( object, fieldOffset );
```
chrisvest: I think this needs to be a `getLongVolatile`.

MishaDemianenko: yes, thx that's a mistake it should be

chrisvest (approved):

MishaDemianenko:
@chrisvest @tinwelint now everything should be good and ready for merge.
Note: will be null merged into 3.3 and 3.4. Separate PR will be created since some additional testing (for 3.3) and rewrite for 3.4 required.

tinwelint (approved):

PR #10652, opened 2017-12-15
Remove unused classes

MishaDemianenko:


pom.xml
```diff
@@ -246,6 +246,12 @@
                   <ignoreMissingAnnotations>false</ignoreMissingAnnotations>
                 </missing-classes>
                 <filter>
+X                  <classes>
```
klaren: `X` marks the spot? :smile:

klaren (approved):
Chrismas cleaning :tada:

Amazing that the pom file still works. Fix it and it's approved.

pom.xml
MishaDemianenko: thats strange it worked, but now i know where my X went)

PR #10654, opened 2017-12-18
CSV parser do not trim quoted strings

tinwelint:
Trimming strings can help correct simple general mistakes in CSV data files and header,
but strings that are quoted are typically put there explicitly and should not be
altered by the parser.

The trimming logic has been moved into BufferedCharSeeker Which means it can make
better decisions about trimming when quotes are involved.
Basically trimming now means that whitespace around delimiters are removed
and quotes will be removed if (after potential whitespace have been trimmed)
value starts and ends with quotes. This results in a more predictable trimming,
especially in combination with quoting.

Examples of w/ or w/o trim strings:

- w/o:
  `"a", " b " , c , "d "," e "` -> `|a| " b " | c | "d "| e |`
  here the string is simply chopped up at each delimiter and if the quotes aren't
  precisely the first and last character of the value then they will be treated
  as quote characters.
- w/ BEFORE this commit:
  `"a", " b " , c , "d "," e "` -> `|a|" b "|c|"d "|e|`
  here the quoted values which begin with whitespace before quotes will only
  have those whitespaces trimmed, but keep the quote characters. It will also
  trim a quoted string after removing its quotes if the quotes were precisely
  the first and last characters of the value.
- w/ AFTER this commit:
  `"a", " b " , c , "d "," e "` -> `|a| b |c|d | e |`
  here the whitespace around delimiters are trimmed and quotes are correctly
  recognized and therefore also removed if they are the first and last characters
  after initial whitespace around delimiters have been removed.

### DIFF
--- a/community/csv/src/main/java/org/neo4j/csv/reader/BufferedCharSeeker.java
+++ b/community/csv/src/main/java/org/neo4j/csv/reader/BufferedCharSeeker.java
@@ -37,6 +37,7 @@ public class BufferedCharSeeker implements CharSeeker
     private static final char EOL_CHAR_2 = '\r';
     private static final char EOF_CHAR = (char) -1;
     private static final char BACK_SLASH = '\\';
+    private static final char WHITESPACE = ' ';
 
     private char[] buffer;
     private int dataLength;
@@ -63,6 +64,7 @@ public class BufferedCharSeeker implements CharSeeker
     private final boolean legacyStyleQuoting;
     private final Source source;
     private Chunk currentChunk;
+    private final boolean trim;
 
     public BufferedCharSeeker( Source source, Configuration config )
     {
@@ -71,6 +73,7 @@ public class BufferedCharSeeker implements CharSeeker
         this.lineStartPos = this.bufferPos;
         this.multilineFields = config.multilineFields();
         this.legacyStyleQuoting = config.legacyStyleQuoting();
+        this.trim = config.trimStrings();
     }
 
     @Override
@@ -96,12 +99,20 @@ public class BufferedCharSeeker implements CharSeeker
             ch = nextChar( skippedChars );
             if ( quoteDepth == 0 )
             {   // In normal mode, i.e. not within quotes
-                if ( ch == quoteChar && seekStartPos == bufferPos - 1/* -1 since we just advanced one */ )
+                if ( isWhitespace( ch ) && trim )
+                {
+                    if ( seekStartPos == bufferPos - 1/* -1 since we just advanced one */ )
+                    {
+                        // We found a whitespace, which was the first of the value and we've been told to trim that off
+                        seekStartPos++;
+                    }
+                }
+                else if ( ch == quoteChar && seekStartPos == bufferPos - 1/* -1 since we just advanced one */ )
                 {   // We found a quote, which was the first of the value, skip it and switch mode
                     quoteDepth++;
+                    isQuoted = true;
                     seekStartPos++;
                     quoteStartLine = lineNumber;
-                    continue;
                 }
                 else if ( isNewLine( ch ) )
                 {   // Encountered newline, done for now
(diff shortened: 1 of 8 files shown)


MishaDemianenko (approved):

praveenag (approved):
LGTM

fickludd (approved):
Nice.

So does this mean

``` "a",  "  b "  , c , "d "   =>    'a', ' b ', 'c', 'd ' ```?

**Comment by tinwelint** [comment]
@fickludd you've accidentally hit a topic of conversation I had w/ @jexp ... i.e. whether or not whitespace around delimiter counts as part of the field or not, with regards to quoting... where things start to become a bit hairy. Quote characters will be regarded simply as characters like any other if a field doesn't start with a quote. To have a CSV file where there's arbitrary whitespace between delimiter and where the field actually starts is confusing for both the humans and the parser... both asking the question: do the field start with a quote or not?

Parsing a line w/ trim strings will currently (including this PR), quite unexpectedly, result in:
`"a", " b " , c , "d "," e "` -> `a|" b "|c|"d "| e |`
Notice the quoting remaining around 'b' and 'd'. This is due to the logic in `BufferedCharSeeker` where it tries to reason about what quote characters mean in different contexts. The trimming logic sits outside currently. To improve on this the knowledge about trimming strings have to move inside `BufferedCharSeeker`. At the end of writing this I'm now leaning towards making this change is the correct way to go about it, i.e. so that your line will become what you suggest. I.e.

1. w/o trim strings (i.e. simply chopping at delimiter chars):
`"a", " b " , c , "d "," e "` -> `|a| " b " | c | "d "| e |`

2. w/ trim strings, as before this PR, i.e. how 3.2 is currently right now:
`"a", " b " , c , "d "," e "` -> `|a|" b "|c|"d "|e|`

3. w/ trim strings, as suggested by this PR before pulling in trim logic into `BufferedCharSeeker`:
`"a", " b " , c , "d "," e "` -> `|a|" b "|c|"d "| e |`

4. w/ trim strings, w/ suggested trim logic inside `BufferedCharSeeker`:
`"a", " b " , c , "d "," e "` -> `|a| b |c|d | e |`

In this light (2) and (3) don't look very sane, do they? Making it like (4), i.e. your suggestions I do think makes sense.

**Comment by fickludd** [comment]
@tinwelint oops, seems like I stepped in it! 

The example I typed is just what I would expect from a blackbox csv parser. I would argue that the quotes are part of the csv data structure, and not the actual data. Therefore they should be gone after parsing. With regard to putting whitespace between the delimiter and the quote, I like to support it just because sometimes people without OCD handwrite csv files, and this functionality will probably make their experience easier. 

So I think 4) is the best option.

**Comment by Dcanzano** [comment]
how would i define a field that i want to include double quotes in the value, for example if i want the result to be

"Bob" Robert Williams

my understanding is that the data in the CSV would need to appear as

"""Bob"" Robert Williams"

with whatever proposed change would that still correctly import.

Also from above #4 seems most sensible

**Comment by tinwelint** [comment]
@Dcanzano yes, the Robert Williams example still holds true, so all good there!

File: community/csv/src/main/java/org/neo4j/csv/reader/BufferedCharSeeker.java [hunk 1]
```diff
@@ -96,12 +99,20 @@ public boolean seek( Mark mark, int untilChar ) throws IOException
             ch = nextChar( skippedChars );
             if ( quoteDepth == 0 )
             {   // In normal mode, i.e. not within quotes
-                if ( ch == quoteChar && seekStartPos == bufferPos - 1/* -1 since we just advanced one */ )
+                if ( isWhitespace( ch ) && trim )
```
MishaDemianenko, line 102: Hm.. shouldn't we use `Character.isWhitespace` here instead, since `whitespace`most of the time means more than singe whitespace char?

MishaDemianenko (commented):
One small question

community/csv/src/main/java/org/neo4j/csv/reader/BufferedCharSeeker.java
tinwelint: Don't think that will work since `Character.isWhitespace` includes tabs and much more... tab can be a delimiter.

MishaDemianenko (approved):

community/csv/src/main/java/org/neo4j/csv/reader/BufferedCharSeeker.java
```diff
@@ -250,6 +252,19 @@ private boolean eof( Mark mark )
...
 
+    private boolean getTrimStringIgnoreErrors( Configuration config, boolean defaultValue )
+    {
+        try
+        {
+            return config.trimStrings();
+        }
+        catch ( Throwable t )
+        {
+            System.err.println( "Error extracting 'trimStrings' setting from configuration: " + config );
```
tinwelint: Very surrealistic, but I understand that we can't do much about it. Perhaps I'll just remove the System.err though. Cheers

PR #10655, opened 2017-12-18
Use new API from compiled code

pontusmelke:
This PR changes to use the new API for:
- `AllNodesScan`
- `nodeGetProperty`
- `nodeHasLabel`

from the compiled runtime.

### DIFF
--- a/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledCursorUtils.java
+++ b/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledCursorUtils.java
@@ -29,7 +29,7 @@ import org.neo4j.values.storable.Value;
 import org.neo4j.values.storable.Values;
 
 /**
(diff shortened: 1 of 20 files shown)


thobe:
The test failures are very relevant. It looks like the generated code is trying to call a method called `getOrLoadCursors()`, but no such method exists in the generated code.

Not sure about all the `NullPointerException`s. Needs further investigation.

pontusmelke:
@thobe That was almost all test things, we have to add that method explicitly in a test as well as in the main code. The null pointers came from missing mocking. There was one actual error as well for `nodeHasLabel` where the wrong signature was used which borked the bytecode generation

thobe:
Ok... that makes me think that the test has too much "real code" in it - i.e. that it is responsible for things that are not really about testing the system, or too much setup. But I'm going to look the other way and tell myself that we are going to replace this subsystem in the glorious future when all the children are happy.

All in all this looks like an improvement, let's get it in.

PR #10663, opened 2017-12-19
Only count the locks acquired by the current query

thobe:
Previously the active lock count from `dbms.listQueries()` would return the lock count of the current transaction. This makes it so that it only returns the lock count from the current query.

It might, however, not count locks that have been re-entered from a previous query in the same transaction.


MishaDemianenko (approved):
Approved, only one small note: it's labeled as `3.2` but PR targeting `3.4`

thobe:
@MishaDemianenko sorry about that. I picked the wrong target in the GitHub UI, the branch was based on 3.2, should be fixed now.

PR #10664, opened 2017-12-19
Remove extra layer of abstraction

pontusmelke:
Having an extra layer between the store and AllStoreHolder caused a
significant regression probably due to the call sites being megamorhphic
preventing inlining to take place.


chrisvest (approved):
I like that, as an apparent side-effect, the code around `loadString` and `loadArray` found a better home as well. 👍

PR #10665, opened 2017-12-19
Fix usage of sun.misc.Unsafe that fails on ExcelsiorJET.

chrisvest:
See https://twitter.com/FOODzzzEATamA/status/918707468082954241 for details.


MishaDemianenko:
@chrisvest from when ExcelsiorJET is a supported VM for us especially in 3.2?

MishaDemianenko (commented):
I do not think we need to introduce additional code for JDK that we do not support out of the blue

PR #10668, opened 2017-12-20
Properly handle STARTS WITH null

pontusmelke:
Queries like `MATCH (n:N) WHERE n.prop STARTS WITH null RETURN n`
should return empty and not throw a `NullPointerException`

changelog: `STARTS WITH null` should not throw an error


thobe (approved):
Looks good to me.

PR #10673, opened 2017-12-20
Primitive long list

MishaDemianenko:
Introduce primitive long list and use it in places where
List\<Long\> was used before.

### DIFF
--- a/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveLongList.java
+++ b/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveLongList.java
@@ -28,7 +28,7 @@ public class PrimitiveLongList implements PrimitiveLongCollection
 {
     private static final int DEFAULT_SIZE = 8;
 
-    private long[] elements = PrimitiveLongCollections.EMPTY_LONG_ARRAY;
+    private long[] elements;
     private int size;
 
     PrimitiveLongList()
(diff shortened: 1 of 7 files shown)


community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveLongList.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.Arrays;
+
+/**
+ * List implementation that holds primitive longs in array that grows on demand.
+ */
+public class PrimitiveLongList implements PrimitiveLongCollection
+{
+    private static final int DEFAULT_SIZE = 8;
+
+    private long[] elements = PrimitiveLongCollections.EMPTY_LONG_ARRAY;
```
tinwelint: Unnecessary assignment, no?

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/PrimitiveLongList.java
```diff
@@ -0,0 +1,130 @@
...
+    {
+        int newCapacity = elements.length << 1;
+        if ( newCapacity < 0 )
+        {
+            throw new IllegalStateException( "Fail to increase list capacity." );
+        }
+        elements = Arrays.copyOf( elements, newCapacity );
+    }
+
+    private class PrimitiveLongListIterator implements PrimitiveLongIterator
```
tinwelint: Doesn't quite honor the contract. Try instead doing:

```
private class PrimitiveLongListIterator extends PrimitiveLongCollections.PrimitiveLongBaseIterator`
{
    private int cursor;

    @Override
    boolean fetchNext()
    {
        return cursor < size ? next( elements[cursor++] ) : false;
    }
}
```

tinwelint (changes requested):
Nice, some small comments

MishaDemianenko:
@tinwelint updated

tinwelint (approved):

PR #10675, opened 2017-12-21
Checking parameter list explicitly

### LINKED ISSUE #10607: Community version doesn't check parameters were provided
Community version doesn't check parameters are provided when there is no match.

- Neo4j version: 3.3.0
- Operating system: Xubuntu 16.04
- API/Driver: Bolt and browser
- **Steps to reproduce**

I expect following code to fail on empty database (or a database with no `Person` nodes) because the parameter name is not provided:
```
        try (Driver driver = GraphDatabase.driver("bolt://localhost")) {

            try (Session session = driver.session()) {

                try (Transaction tx = session.beginTransaction()) {

                    StatementResult run = tx.run("MATCH (n:Person {name:{name}}) RETURN n");
                    run.consume();

                    tx.success();

                }
            }
        }
```

- Expected behavior
Fail with 
```
Exception in thread "main" org.neo4j.driver.v1.exceptions.ClientException: Expected a parameter named name
```
This fails as expected on enterprise version or on community when there are some nodes with `Person` label.
- Actual behavior
Just returns no results.

Lojjs:
If a parameters was missing in the parameter list, an error was only thrown if that parameter was used in the execution. This PR introduces to always check the parameter list during planning.

changelog: Fixes #10607 by explicitly checking that all parameters present in the query are also provided in the parameter list

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/ExecutionEngine.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/ExecutionEngine.scala
@@ -25,6 +25,7 @@ import org.neo4j.cypher._
 import org.neo4j.cypher.internal.compatibility.v3_3._
 import org.neo4j.cypher.internal.compatibility.v3_3.runtime.helpers.{RuntimeJavaValueConverter, RuntimeScalaValueConverter, ValueConversion}
 import org.neo4j.cypher.internal.compiler.v3_3.prettifier.Prettifier
+import org.neo4j.cypher.internal.frontend.v3_3.ParameterNotFoundException
 import org.neo4j.cypher.internal.frontend.v3_3.phases.CompilationPhaseTracer
 import org.neo4j.cypher.internal.spi.v3_3.TransactionalContextWrapper
 import org.neo4j.cypher.internal.tracing.{CompilationTracer, TimingCompilationTracer}
@@ -73,7 +74,7 @@ class ExecutionEngine(val queryService: GraphDatabaseQueryService,
 
   private val executionMonitor = kernelMonitors.newMonitor(classOf[QueryExecutionMonitor])
 
-  private val cacheAccessor = new MonitoringCacheAccessor[String, (ExecutionPlan, Map[String, Any])](cacheMonitor)
+  private val cacheAccessor = new MonitoringCacheAccessor[String, (ExecutionPlan, Map[String, Any], Seq[String])](cacheMonitor)
 
   private val preParsedQueries = new LFUCache[String, PreParsedQuery](getPlanCacheSize)
   private val parsedQueries = new LFUCache[String, ParsedQuery](getPlanCacheSize)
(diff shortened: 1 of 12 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NodeIndexSeekByRangeAcceptanceTest.scala
```diff
@@ -1051,7 +1051,7 @@ class NodeIndexSeekByRangeAcceptanceTest extends ExecutionEngineFunSuite with Cy
       planComparisonStrategy = ComparePlansWithAssertion((plan) => {
         //THEN
         plan should useOperators(IndexSeekByRange.name)
-      }, Configs.AllRulePlanners))
+      }, Configs.AllRulePlanners), params = Map("param" -> Array[Int](1, 2, 3)))
```
Lojjs, line 1054: In my solution, a correct parameter list is also required for EXPLAIN queries, not sure if that is the right thing todo.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/ExecutionEngine.scala
```diff
@@ -216,6 +218,17 @@ class ExecutionEngine(val queryService: GraphDatabaseQueryService,
     throw new IllegalStateException("Could not execute query due to insanely frequent schema changes")
   }
 
+  @throws(classOf[ParameterNotFoundException])
+  private def checkParameters(queryParams: Seq[String], givenParams: MapValue, extractedParams: Map[String, Any]) {
+    exceptionHandler.runSafely {
+      queryParams.foreach( key => {
+        if (!givenParams.containsKey(key) && !extractedParams.contains(key)) {
+          throw new ParameterNotFoundException("Expected a parameter named " + key)
```
craigtaverner: Would it not be even more useful to collect all missing parameters and throw an exception listing them all, instead of throwing one naming only the first missing parameter?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NodeIndexSeekByRangeAcceptanceTest.scala
craigtaverner: I wonder if this could be a problem? I suspect there are those that can use the browser to explain queries with nice graphical query plan output that will now no longer be able to do so because the browser does not support query parameters.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ParameterValuesAcceptanceTest.scala
```diff
@@ -119,4 +119,28 @@ class ParameterValuesAcceptanceTest extends ExecutionEngineFunSuite with CypherC
...
+    // all versions of 3.3
+    val config = Configs.Version3_3 + Configs.Procs - Configs.AllRulePlanners
+
+    failWithError(config, "MATCH (n:Person {name:{name}}) RETURN n", Seq("Expected a parameter named name"), params = "nam" -> "Neo")
+  }
+
+  test("match with misspelled parameter should return error for non-empty db") {
+    val config = Configs.AbsolutelyAll - Configs.Compiled - Configs.Cost2_3
+    failWithError(config, "CREATE (n:Person) WITH n MATCH (n:Person {name:{name}}) RETURN n", Seq("Expected a parameter named name"), params = "nam" -> "Neo")
+  }
```
craigtaverner: The solution tests both giveParams and extractedParams, is there a way to write a test that demonstrates this?

craigtaverner (changes requested):

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/NodeIndexSeekByRangeAcceptanceTest.scala
Lojjs: Ah, that can certainly be a problem. `EXPLAIN` queries should probably be excluded from the check.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/ExecutionEngine.scala
Lojjs: That is a good idea, will fix it.

craigtaverner (approved):

PR #10676, opened 2017-12-21
Compiled runtime should return String[] not Object[] for properties

pontusmelke:
In embedded we need to preserve types so that when we return for

```
CREATE (p:Person {names:['adsf', 'adf' ]})
```

we should return a `String[]`

```
MATCH (n) RETURN n.names
```

changelog: Do not return Object[] when querying for array property in Cypher.


fickludd (approved):
Looks good!

PR #10682, opened 2017-12-22
Disable channel auto-read when bolt messages are queued up

ali-ince:
When bolt messages are queued up in Bolt Worker Thread's job queue by continuous submission of messages (in an intent of batch queries or not waiting previous time-consuming queries to complete), job queue may start blocking IO threads which in turn starts blocking other client connections that are served by that specific IO thread.

This PR changes blocking job queue to a non-blocking one and sets channel's auto-read property based on job queue size. When job queue size gets over a pre-defined high water mark (`300`), channel's auto-read property is disabled and when it gets below a pre-defined low water mark (`100`) it is enabled back.

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltChannelAutoReadLimiter.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltChannelAutoReadLimiter.java
@@ -34,9 +34,6 @@ public class BoltChannelAutoReadLimiter implements BoltWorkerQueueMonitor
     protected static final String LOW_WATERMARK_NAME = "low_watermark";
     protected static final String HIGH_WATERMARK_NAME = "high_watermark";
 
-    private static final int defaultLowWatermark = FeatureToggles.getInteger( BoltChannelAutoReadLimiter.class, LOW_WATERMARK_NAME, 100 );
-    private static final int defaultHighWatermark = FeatureToggles.getInteger( BoltChannelAutoReadLimiter.class, HIGH_WATERMARK_NAME, 300 );
-
     private final AtomicInteger queueSize = new AtomicInteger( 0 );
     private final Channel channel;
     private final Log log;
@@ -45,7 +42,8 @@ public class BoltChannelAutoReadLimiter implements BoltWorkerQueueMonitor
 
     public BoltChannelAutoReadLimiter( Channel channel, Log log )
     {
-        this( channel, log, defaultLowWatermark, defaultHighWatermark );
+        this( channel, log, FeatureToggles.getInteger( BoltChannelAutoReadLimiter.class, LOW_WATERMARK_NAME, 100 ),
+                FeatureToggles.getInteger( BoltChannelAutoReadLimiter.class, HIGH_WATERMARK_NAME, 300 ) );
     }
 
     public BoltChannelAutoReadLimiter( Channel channel, Log log, int lowWatermark, int highWatermark)
--- a/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltChannelAutoReadLimiter.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltChannelAutoReadLimiter.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2002-2017 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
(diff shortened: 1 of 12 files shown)


community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltChannelAutoReadLimiter.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Collection;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.neo4j.logging.Log;
+
+import static java.util.Objects.requireNonNull;
+
+public class BoltChannelAutoReadLimiter implements BoltWorkerQueueMonitor
+{
+    private static final int defaultLowWatermark = Integer.getInteger( "org.neo4j.bolt.autoread.loWatermark", 100 );
```
lutovich: Maybe use `FeatureToggles` for these system props?

community/bolt/src/test/java/org/neo4j/bolt/v1/transport/integration/BoltChannelAutoReadLimiterIT.java
```diff
@@ -0,0 +1,209 @@
...
+import static org.neo4j.bolt.v1.messaging.message.InitMessage.init;
+import static org.neo4j.bolt.v1.messaging.message.RunMessage.run;
+import static org.neo4j.bolt.v1.messaging.util.MessageMatchers.msgSuccess;
+import static org.neo4j.bolt.v1.transport.integration.TransportTestUtil.eventuallyReceives;
+import static org.neo4j.kernel.api.proc.ProcedureSignature.procedureSignature;
+
+@RunWith( Parameterized.class )
+public class BoltChannelAutoReadLimiterIT
+{
+    protected EphemeralFileSystemRule fsRule = new EphemeralFileSystemRule();
```
lutovich: Fields can be private

community/bolt/src/test/java/org/neo4j/bolt/v1/transport/integration/BoltChannelAutoReadLimiterIT.java
```diff
@@ -0,0 +1,209 @@
...
+    public AssertableLogProvider logProvider;
+
+    @Parameterized.Parameter( 0 )
+    public Factory<TransportConnection> cf;
+
+    @Parameterized.Parameter( 1 )
+    public HostnamePort address;
+
+    protected TransportConnection client;
+    private final String version = "Neo4j/" + Version.getNeo4jVersion();
```
lutovich: Is this field used?

lutovich (commented):
@ali-ince changes look fine to me. Made couple small comments.

One more question. What do you think about making `BoltWorkerQueueMonitor` a slightly smaller interface:
```
interface BoltWorkerQueueMonitor {
    void queueSizeChanged(int newSize);
}
```
Implementation is currently only concerned about queue size but receives `Job` instances, which it does not really use.

ali-ince:
@lutovich I originally thought that we may make use of `BoltWorkerQueueMonitor` for logging / monitoring purposes that may include information about `Job` instances queued / dequeued. Does that make sense, wdyt?

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltChannelAutoReadLimiter.java
```diff
@@ -0,0 +1,123 @@
...
+
+public class BoltChannelAutoReadLimiter implements BoltWorkerQueueMonitor
+{
+    protected static final String LOW_WATERMARK_NAME = "low_watermark";
+    protected static final String HIGH_WATERMARK_NAME = "high_watermark";
+
+    private static final int defaultLowWatermark = FeatureToggles.getInteger( BoltChannelAutoReadLimiter.class, LOW_WATERMARK_NAME, 100 );
+    private static final int defaultHighWatermark = FeatureToggles.getInteger( BoltChannelAutoReadLimiter.class, HIGH_WATERMARK_NAME, 300 );
+
+    private final AtomicInteger queueSize = new AtomicInteger( 0 );
```
lutovich: Can this field have type `int`? I think it is only accessed from a `RunnableBoltWorker` thread

lutovich:
@ali-ince yeah, makes sense. Added one other small comment.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/BoltChannelAutoReadLimiter.java
ali-ince: `RunnableBoltWorker#queue` is invoked by netty IO threads, so I think it would be better to keep it as an `AtomicInteger`. Am I missing some point?

lutovich: You are right, never mind me :)

lutovich (approved):

PR #10684, opened 2017-12-26
Do not reconnect LDAP context after TLS negotiation

MishaDemianenko:
Reconnecting after TLS negotiation in java 9 will create
separate connection/session that will not have TLS
enabled.
Updated behavior also matches to javadoc described usage pattern.


MishaDemianenko:
@fickludd / @henriknyman guys can you please take a look, thx!

fickludd (approved):
I don't remember why there is a reconnect, but I know we have rich testing of the ldap permutations. If this can (and should) be removed then I'm 👍. We should also monitor the security smoke tests to see that we do not break with some other LDAP server. Maybe henrik will have a glance when he's back, but let's merge meanwhile.

PR #10685, opened 2017-12-27
Define automatic module names for java 9 modules

MishaDemianenko:
Use unnamed modules for now; introduce separate property for
scala vm version since current version does not support have support
for -target:jvm-9
Use unnamed modules, export missed packages to unnamed modules.
Separate warm hugs to 'javax.xml.bind.DatatypeConverter' lovers.

### DIFF
--- a/community/server/src/test/java/org/neo4j/server/rest/transactional/ConcurrentTransactionAccessTest.java
+++ b/community/server/src/test/java/org/neo4j/server/rest/transactional/ConcurrentTransactionAccessTest.java
@@ -33,7 +33,7 @@ import org.neo4j.server.rest.transactional.error.InvalidConcurrentTransactionAcc
 import org.neo4j.server.rest.web.TransactionUriScheme;
 import org.neo4j.test.DoubleLatch;
 
-import static javax.xml.bind.DatatypeConverter.parseLong;
+import static java.lang.Long.parseLong;
 import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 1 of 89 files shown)


MishaDemianenko:
Jersey so far does not have support for java 9. This PR, if it will be merged, should fix that:  https://github.com/jersey/jersey-1.x/pull/25

klaren:
Looks good. This might break user code that relies on the fact we bundle `jersey-multipart` though. Also, I believe that revapi still poses a problem with the execution time.

MishaDemianenko:
@klaren i do not think our jersey update will be resolved anytime soon, so i just switched back versions to 1.8 even for java 9 profile, but at least we will have automatic module names for what we have now

community/server/src/test/java/org/neo4j/server/rest/transactional/ConcurrentTransactionAccessTest.java
```diff
@@ -33,7 +33,7 @@
 import org.neo4j.server.rest.web.TransactionUriScheme;
 import org.neo4j.test.DoubleLatch;
 
-import static javax.xml.bind.DatatypeConverter.parseLong;
```
klaren, line 36: :egg:

klaren (changes requested):
It looks a bit incinsistent with both
```
org.neo4j.internal.cypher.*
```
and
```
org.neo4j.cypher.internal.*
```
as module names. I would prefer the first one to align it with `org.neo4j.internal.kernel.api`.

klaren (approved):
It matches root packages so it's fine.

PR #10686, opened 2017-12-27
Remove some minor inefficiencies

pontusmelke:
Some profiling of property access in the new API showed that we can do a little better by doing minor tweaks.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Labels.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Labels.java
@@ -19,11 +19,11 @@
  */
 package org.neo4j.kernel.impl.newapi;
 
-import java.util.Collection;
 import java.util.Arrays;
+import java.util.Collection;
+
 import org.neo4j.collection.primitive.PrimitiveIntIterator;
 import org.neo4j.collection.primitive.PrimitiveIntSet;
-
 import org.neo4j.internal.kernel.api.LabelSet;
 
 public class Labels implements LabelSet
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Labels.java
```diff
@@ -78,6 +78,9 @@ public int label( int offset )
     @Override
     public boolean contains( int labelToken )
     {
+        //It may look tempting to use binary search
+        //however doing a linear search is actually faster for reasonable
+        //label sizes (≤100 labels)
```
chrisvest, line 83: 👍

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
```diff
@@ -132,32 +134,49 @@ public boolean next()
...
+                    long current = currentBlock();
+                    PropertyType type = PropertyType.getPropertyTypeOrNull( current );
+                    if ( type == null )
+                    {
+                        break;
+                    }
+                    block += type.calculateNumberOfBlocksUsed( current );
+                }
+                //nothing left, need to read a new record
+                if ( block >= numberOfBlocks || type() == null )
```
Lojjs, line 160: Use `type` instead of `type()` here as well

Lojjs (changes requested):
Overall it looks like a good improvement, just one minor comment.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/PropertyCursor.java
pontusmelke: This is slightly confusing but we cannot reuse type here since after we have done `block +=...` we might be at another property with a different type.

Lojjs (approved):

PR #10689, opened 2017-12-27
Do not rely on Thread.sleep in begin__execute__rollback_concurrently in TransactionIT

MishaDemianenko:
Instead of relying on Thread.sleep detect ongoing statement execution
by inspecting active ongoing transactions.

### DIFF
--- a/community/server/src/test/java/org/neo4j/server/rest/transactional/integration/TransactionIT.java
+++ b/community/server/src/test/java/org/neo4j/server/rest/transactional/integration/TransactionIT.java
@@ -20,6 +20,8 @@
 package org.neo4j.server.rest.transactional.integration;
 
 import org.codehaus.jackson.JsonNode;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


community/server/src/test/java/org/neo4j/server/rest/transactional/integration/TransactionIT.java
```diff
@@ -585,17 +600,10 @@ public void begin__execute__rollback_concurrently() throws Exception
...
-            try
-            {
-                latch.await();
-                Thread.sleep( 100 );
-            }
-            catch ( InterruptedException ignored )
-            {
-                // go
-            }
+            waitForStatementExecution( statement );
```
chrisvest, line 616: Looks like we are still racing with the completion of the `statement` query.

I think it would be more robust to have a third transaction lock a node, make the statement try to modify a property of that node where it will get stuck on the entity lock. Then we should be able to wait for the statement execution without worrying that it might have passed us by already.

This would incidentally expand the scope of the test to include verifying that we can terminate transactions even when they are waiting in the lock manager.

chrisvest (commented):
I had a comment with a possible improvement.

MishaDemianenko:
@chrisvest why not, updated

chrisvest (approved):

PR #10692, opened 2017-12-27
Cleanup junit.framework and org.mockito.Matchers deprecations

MishaDemianenko:


burqen (approved):
Cleanup cleanup cleanup! Nice 😄 

Needs rebase though and test failure looks relevant but it also looks like you already fixed it here.

PR #10697, opened 2017-12-28
Keep Labels, Nodes, Relationship _STATE's in primitive maps in TxState

MishaDemianenko:
Update LABEL_STATE, NODE_STATE, RELATIONSHIP_STATE to use primitive maps
and maps that keep underlying state.

### DIFF
--- a/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/hopscotch/PrimitiveLongObjectHashMap.java
+++ b/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/hopscotch/PrimitiveLongObjectHashMap.java
@@ -19,6 +19,9 @@
  */
 package org.neo4j.collection.primitive.hopscotch;
 
+import java.util.Iterator;
+
+import org.neo4j.collection.primitive.PrimitiveLongIterator;
 import org.neo4j.collection.primitive.PrimitiveLongObjectMap;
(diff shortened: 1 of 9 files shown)


community/primitive-collections/src/main/java/org/neo4j/collection/primitive/hopscotch/PrimitiveLongObjectHashMap.java
```diff
@@ -165,4 +174,35 @@ public boolean equals( Object o )
             return hash == that.hash;
         }
     }
+
+    private class ValueIterable implements Iterable<VALUE>
+    {
+        private final PrimitiveLongObjectHashMap<VALUE> map;
+        private final PrimitiveLongIterator iterator;
+
+        ValueIterable( PrimitiveLongObjectHashMap<VALUE> map )
```
tinwelint: It looks like this Iterable will only be able to produce one good iterator. Perhaps it's not super important, but the `private final PrimitiveLongIterator iterator` could might as well be created and kept inside the anonymous `Iterable` created in `iterator()`, couldn't it?

tinwelint (commented):
Looks like good and straight forward changes. Had one comment about the `ValueIterable` though

MishaDemianenko:
@tinwelint sure, updated

tinwelint (approved):

PR #10698, opened 2017-12-28
Lazy propertiesMap in TxState

MishaDemianenko:


community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/TxState.java
```diff
@@ -152,7 +152,7 @@ void setMap( TxState state, Map<Long, RelationshipStateImpl> map )
     // Tracks added and removed relationships, not modified relationships
     private RelationshipDiffSets<Long> relationships;
 
-    private PrimitiveLongObjectMap<PropertyContainerState> propertiesMap = Primitive.longObjectMap();
+    private PrimitiveLongObjectMap<PropertyContainerState> propertiesMap;
```
tinwelint, line 155: I think the use of `propertiesMap` looks very strange, as if `TxState` was used to pass state from `NodeCursor` to `PropertyCursor` somehow... I'd like to instead remove this map of accidental state altogether and change the cursor code. Lemme try and do that in a PR, before you merge this OK?

tinwelint (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/TxState.java
tinwelint: Something like #10708

MishaDemianenko: I didn't want to mess with work that is ongoing for new API so did minimal change around tx state, but if guys will be happy with changes #10708 then we can close this PR.

MishaDemianenko:
Closing in favor of https://github.com/neo4j/neo4j/pull/10708

PR #10700, opened 2017-12-28
Primitive long based collection for added/removed nodes and relationships

MishaDemianenko:
Introduce specialized primitive long version of diffset.
Switch added and removed relationships, nodes in transaction state
to use the new specialized collection.


tinwelint (approved):
Looks good. Also I see that you've converted some int -> long for keys. I like it, it's more unified... perhaps it's about time to rid ourselves of the SuperReadableDiffSets and SuperDiffSets too?

MishaDemianenko:
this will not gonna go anywhere in current state, closing

PR #10708, opened 2018-01-02
Simplifies property access in new kernel API

tinwelint:
Especially w/ regards to transaction state. Read interface has methods for initializing
node/relationship property cursors. Previously those methods didn't have the entity reference
as argument. Because of this limitation the interaction between
the NodeCursor/RelationshipCursor and PropertyCursor was made complicated by encoding
different behaviour w/ regards to transaction state into the property reference.
On top of that TxState was used as an additional helper to pass information about tx state
from NodeCursor/RelationshipCursor to PropertyCursor, resulting in more memory usage in TxState
as well as going via a map.

This is simplified by letting Read methods accept nodeReference for its nodeProperties call
and relationshipReference for its relationshipProperties call. Now TxState can be left
out of this interaction and its propertiesMap removed. This also results in simpler cursor code.

### DIFF
--- a/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Read.java
+++ b/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Read.java
@@ -119,20 +119,24 @@ public interface Read
     void relationships( long nodeReference, long reference, RelationshipTraversalCursor cursor );
 
     /**
+     * @param nodeReference
+     *         the owner of the properties.
      * @param reference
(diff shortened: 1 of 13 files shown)


pontusmelke (changes requested):
Nice, much better.

- needs rebase
- I think you can completely remove the `txFlags` methods from `References` now.

tinwelint:
@pontusmelke cheers, updated

pontusmelke (approved):

PR #10709, opened 2018-01-02
Allow null realm in ShiroAuthToken

lutovich:
This PR makes `ShiroAuthToken` handle `null` realm values. It is possible for the provided auth token map to contain `null` as explicit value.


craigtaverner (approved):

fickludd (approved):
LGTM, like the drive-by clean-ups as well.

PR #10711, opened 2018-01-02
Additional testing for half-applied constraints

ragadeeshu:


enterprise/neo4j-enterprise/src/test/java/org/neo4j/HalfAppliedConstraintRecoveryIT.java
```diff
@@ -17,20 +17,25 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-package org.neo4j.kernel.impl.api.constraints;
+package org.neo4j;
```
chrisvest: This needs to be AGPL, now that you've moved the test to Enterprise.

chrisvest (approved):
Looks good, apart from the license header.

PR #10714, opened 2018-01-03
Less indirections in iterators

pontusmelke:
There were recently added a bunch of extra layers when `PrimitiveLongIterator` were
exchanged for `PrimitiveLongResourceIterator`. Basically we ended up with three layers
of primitive iterators which caused a regression. This PR reduces the number of layers.


burqen (approved):
This looks good! Nice to get rid of some layers 🎉

PR #10715, opened 2018-01-03
 Make SlotConfigurations and ArgumentSizes Attributes

sherfert:
Also add some convenience methods to Attribute. I ask especially for
a review of the `iterator` method, that I am quite unsure about.

### DIFF
--- a/community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/attribution/Attribute.scala
+++ b/community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/attribution/Attribute.scala
@@ -23,7 +23,7 @@ import scala.collection.mutable.ArrayBuffer
 trait Attribute[T] {
 
   // TODO crappy perf?
-  val array:ArrayBuffer[Unchangeable[T]] = new ArrayBuffer[Unchangeable[T]]()
+  private val array:ArrayBuffer[Unchangeable[T]] = new ArrayBuffer[Unchangeable[T]]()
 
   def set(id:Id, t:T): Unit = {
     val requiredSize = id.x + 1
@@ -40,7 +40,7 @@ trait Attribute[T] {
     array.size > id.x && array(id.x).hasValue
   }
 
-  def getOrElse(id: Id, other: T): T = {
+  def getOrElse(id: Id, other: => T): T = {
     if(isDefinedAt(id)) get(id) else other
   }
 
--- a/community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/attribution/Attribute.scala
+++ b/community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/attribution/Attribute.scala
@@ -22,17 +22,16 @@ import scala.collection.mutable.ArrayBuffer
 
 trait Attribute[T] {
 
-  // TODO crappy perf?
-  private val array:ArrayBuffer[Unchangeable[T]] = new ArrayBuffer[Unchangeable[T]]()
+  private val array: ArrayBuffer[Unchangeable[T]] = new ArrayBuffer[Unchangeable[T]]()
 
-  def set(id:Id, t:T): Unit = {
+  def set(id: Id, t: T): Unit = {
     val requiredSize = id.x + 1
     if (array.size < requiredSize)
       resizeArray(requiredSize)
     array(id.x).value = t
   }
 
-  def get(id:Id): T = {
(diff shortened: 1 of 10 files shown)


community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/attribution/Attribute.scala
```diff
@@ -36,6 +36,54 @@ trait Attribute[T] {
...
+      else  {
+        if(nextTup == null) {
+          fetchNext()
+        }
+        nextTup != null
+      }
+    }
+
+    override def next() = {
+      if(nextTup == null) {
```
systay: I think it leads to simpler code an less bugs if `.next()` just checks if `.hasNext()` first. This will advance the pointer for us, and we know whether to return something or not. At the moment, I think we would return null instead of throwing if you first check `.hasNext`, and ignore a false it returns and still do a `.next()`, if that makes sense.

community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/attribution/Attribute.scala
```diff
@@ -36,6 +36,54 @@ trait Attribute[T] {
...
+        fetchNext()
+      }
+      val res = nextTup
+      nextTup = null
+      res
+    }
+  }
+
+  def size = iterator.size
```
systay: won't this just lead to an endless loop?

community/cypher/util-3.4/src/test/scala/org/neo4j/cypher/internal/util/v3_4/attribution/AttributeTest.scala
```diff
@@ -0,0 +1,59 @@
...
+    val attr = new TestAttribute
+    attr.set(Id(2), 33)
+    attr.set(Id(4), 27)
+
+    val iter = attr.iterator
+    iter.hasNext should be(true)
+    iter.next() should be((Id(2),33))
+    iter.hasNext should be(true)
+    iter.next() should be((Id(4),27))
+    iter.hasNext should be(false)
```
systay, line 56: After this line, you could add a `intercept[NoSuchElementException](iter.next()` to check that correct error is produced.

systay (changes requested):
I really like how the code ends up looking. Much prefer it over the maps we keep sending around.

Minor comments on the iterator implementation.

community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/attribution/Attribute.scala
sherfert: No. Iterator.size drains the iterator until `hasNext == false`. I will add a test, though.

community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/attribution/Attribute.scala
```diff
@@ -22,34 +22,82 @@ import scala.collection.mutable.ArrayBuffer
...
+        val res = nextTup
+        nextTup = null
+        res
+      } else {
+        throw new NoSuchElementException
+      }
+    }
+  }
+
+  def size = iterator.size
```
fickludd, line 83: If we ever think this is slow, we could just keep a private size field which is incremented on `set` if there was no previous value.

community/cypher/util-3.4/src/test/scala/org/neo4j/cypher/internal/util/v3_4/attribution/AttributeTest.scala
```diff
@@ -0,0 +1,83 @@
...
+  test("test isDefinedAt") {
+    val attr = new TestAttribute
+    attr.set(Id(2), 33)
+    attr.set(Id(4), 27)
+
+    attr.isDefinedAt(Id(1)) should be(false)
+    attr.isDefinedAt(Id(2)) should be(true)
+    attr.isDefinedAt(Id(3)) should be(false)
+    attr.isDefinedAt(Id(4)) should be(true)
+
```
fickludd, line 43: Adding `attr.isDefinedAt(Id(5)) should be(true)` would increase coverage to out-of-bounds in the underlying array.

fickludd (approved):
I agree w. Andrés, the direction this is going is looking very promising.

systay (approved):

PR #10716, opened 2018-01-03
Separate initial cluster size from runtime raft member set

jimwebber:
Splitting out minimum_consensus_group_size from expected_core_cluster_size to address the different lifecycle concerns.

The latter is about when we will start up (N/2+1 instances ready), the
former is how small we will allow the Raft group to shrink.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
@@ -34,6 +34,7 @@ import org.neo4j.causalclustering.core.consensus.log.cache.InFlightCacheFactory;
 import org.neo4j.configuration.Description;
 import org.neo4j.configuration.Internal;
 import org.neo4j.configuration.LoadableConfig;
+import org.neo4j.configuration.ReplacedBy;
 import org.neo4j.graphdb.config.BaseSetting;
 import org.neo4j.graphdb.config.InvalidSettingException;
 import org.neo4j.graphdb.config.Setting;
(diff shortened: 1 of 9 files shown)


jjaderberg (approved):
LGTM

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
```diff
@@ -101,10 +102,21 @@
     public static final Setting<Integer> raft_in_queue_max_batch =
             setting( "causal_clustering.raft_in_queue_max_batch", INTEGER, "64" );
 
-    @Description( "Expected number of Core machines in the cluster" )
+    @Description( "Expected number of Core machines in the cluster before startup" )
```
martinfurmanski, line 105: I think we can do a better job documenting the behaviour here. The text here will show up in the reference manual. I am specifically missing a short description, which captures the gist of it. Longer descriptions go in the manual. This can be used as inspiration:

https://stackoverflow.com/questions/43103105/monitoring-a-causal-clusters-redundancy

Comment applies to all three settings.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
```diff
@@ -101,10 +102,21 @@
...
+    @Deprecated
+    @ReplacedBy( "minimum_core_cluster_size_at_startup and minimum_core_cluster_size_at_runtime" )
     public static final Setting<Integer> expected_core_cluster_size =
             setting( "causal_clustering.expected_core_cluster_size", INTEGER, "3" );
 
+    @Description( "Minimum number of Core machines in the cluster at startup" )
```
martinfurmanski: Perhaps formation instead of startup? In actuality it is for formation of the underlying hazelcast cluster. The majority of the members must be available to select a hazelcast master. I think minimum doesn't perhaps describe this majority quorum behaviour sufficiently.

martinfurmanski (changes requested):
You have to update neo4j.conf as well.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
jimwebber: Done.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/CausalClusteringSettings.java
jimwebber: Done.

martinfurmanski:
@jimwebber Still missing updates to neo4j.conf.

jimwebber:
@martinfurmanski addressed the config changes, different repo.

martinfurmanski (approved):

PR #10718, opened 2018-01-03
Use new api for index operations

pontusmelke:


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ProfilerAcceptanceTest.scala
```diff
@@ -420,7 +420,7 @@ class ProfilerAcceptanceTest extends ExecutionEngineFunSuite with CreateTempFile
 
         //THEN
         assertDbHits(7)(result)("Expand(All)")
-        assertDbHits(2)(result)("NodeUniqueIndexSeek")
+        assertDbHits(3)(result)("NodeUniqueIndexSeek")
```
SaschaPeukert, line 423: We do we now have an additional DBHit?

SaschaPeukert (commented):
Looks awesome. Just one question

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ProfilerAcceptanceTest.scala
pontusmelke: We have an extra dbhit on the first call because we load the index reference from the database on first load. Once cached we will go back to only having 2 dbhits.

PR #10723, opened 2018-01-04
Much reduced memory usage of collision values in importer

tinwelint:
- Reduced big tracker cache from 6 --> 5 bytes, i.e. an overall ~8% lowered memory usage for large imports
- Moved keeping of collision values off-heap and reduced down to 1/6th of the memory it used to occupy in heap
- Removed array for keeping duplicate node ids entirely

These changes are targeted to improve scenarios where there are lots of duplicate node IDs in the input files. Previously the management of those duplicates required a lot of heap memory. Now this has changed to require only a fraction of that, and also moved off-heap.

There is now only remnants of the page cache meta data left in the heap that needs to be considered when chosing heap size for an import, nothing else by design requires a heap larger than a couple of gigabytes, tops, for an import regardless of data size and its condition.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
@@ -742,7 +742,6 @@ public class EncodingIdMapper implements IdMapper
         return clearCollision( dataCache.get( trackerCache.get( index ) ) );
     }
 
-//<<<<<<< HEAD
     private long findCollisionIndex( long value )
     {
         // can't be done on unsorted data
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
@@ -21,16 +21,15 @@ package org.neo4j.unsafe.impl.batchimport.cache.idmapping.string;
 
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.List;
-import java.util.function.IntFunction;
 import java.util.function.LongFunction;
 
+import org.neo4j.collection.primitive.PrimitiveLongCollections;
+import org.neo4j.collection.primitive.PrimitiveLongIterator;
 import org.neo4j.function.Factory;
 import org.neo4j.helpers.progress.ProgressListener;
 import org.neo4j.unsafe.impl.batchimport.HighestId;
 import org.neo4j.unsafe.impl.batchimport.Utils.CompareType;
 import org.neo4j.unsafe.impl.batchimport.cache.ByteArray;
-import org.neo4j.unsafe.impl.batchimport.cache.IntArray;
 import org.neo4j.unsafe.impl.batchimport.cache.LongArray;
 import org.neo4j.unsafe.impl.batchimport.cache.LongBitsManipulator;
 import org.neo4j.unsafe.impl.batchimport.cache.MemoryStatsVisitor;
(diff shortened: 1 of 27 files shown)


tinwelint:
Builds on top of #10648

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/EncodingIdMapper.java
```diff
@@ -678,6 +732,7 @@ private long dataValue( long index )
         return clearCollision( dataCache.get( trackerCache.get( index ) ) );
     }
 
+//<<<<<<< HEAD
```
chrisvest: Merge left-overs.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/StringCollisionValues.java
```diff
@@ -0,0 +1,101 @@
...
+            throw new IllegalArgumentException( string );
+        }
+
+        long endOffset = offset + 2 + bytes.length;
+        ByteArray end = cache.at( endOffset );
+        if ( end != current )
+        {
+            offset = endOffset;
+            current = end;
+        }
```
chrisvest: Are we sure that the value will always fit within a single chunk?

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/StringCollisionValues.java
```diff
@@ -0,0 +1,101 @@
...
+
+        long startOffset = offset;
+        current.setByte( offset++, 0, (byte) (length & 0xFF) );
+        current.setByte( offset++, 0, (byte) ((length >>> 8) & 0xFF) );
```
chrisvest: Why not getShort and setShort?

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/StringCollisionValuesTest.java
```diff
@@ -0,0 +1,80 @@
...
+    {
+        // given
+        try ( StringCollisionValues values = new StringCollisionValues( factory, 100_000 ); )
+        {
+            // when
+            long[] offsets = new long[100];
+            String[] strings = new String[offsets.length];
+            for ( int i = 0; i < offsets.length; i++ )
+            {
+                String string = random.string();
```
chrisvest, line 76: By default, the maximum length of string that this will produce is 20 characters. I think that misses a lot of important edge cases.

chrisvest (commented):
Looks okay. Just a few comments. Please take a look.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/idmapping/string/StringCollisionValues.java
tinwelint: This is because we can cross a dynamic byte array boundary underneath. It's a different use of ByteArray, one where it's truly used as a byte[]. Doing it this way will work around issues with this use case.

tinwelint: Oh wait, we work-around this right above this. I think this a remnant from when that wasn't done. I'll change it.

chrisvest: Yeah, it's a bit hard to follow when you need to be weary of the boundary or not.

tinwelint:
Added testing for larger keys, much larger and your concerns came true. Updated the implementation to handle all that.

chrisvest (approved):

PR #10724, opened 2018-01-04
Refactor bolt server to easier allow multiple protocol versions

lutovich:
PR makes it easier to introduce new bolt protocol versions. It adds a factory for protocol handlers and extracts interface for V1 format so that it can be easier extended with new types.


community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/Neo4jPackV1.java
```diff
@@ -0,0 +1,609 @@
...
+        {
+            error = new Error( Status.Request.Invalid,
+                    "Point is not yet supported as a return type in Bolt" );
+            packNull();
+        }
+
+        @Override
+        public void writeNull() throws IOException
+        {
+            packNull();
```
pontusmelke, line 299: Maybe we could inline all these method calls, not sure if there is a benefit of having both `pack( X value )` and `writeX( X value )`?

community/bolt/src/test/java/org/neo4j/bolt/v1/messaging/BoltRequestMessageWriter.java
```diff
@@ -58,7 +58,7 @@ public void onInit( String clientName, Map<String,Object> credentials ) throws I
     {
         packer.packStructHeader( 1, INIT.signature() );
```
pontusmelke: This is actually wrong but it is an error that has propagated to all drivers. The init struct has size 2 and not size 1. We are saved by ignoring the struct size when we read the init message.

community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/Neo4jPackV1.java
lutovich: You are right, having both `#pack()` and `#write()` methods is confusing. Former come from `PackStream.Packer` which this class extends. So this class is a mix of high level methods like `#pack(AnyValue)` and low level packstream methods like `#packStructHeader()`.

I think simple inlining will make things even more confusing. This class will have to deal with `PackOutput` directly and take on some packstream functionality.

community/bolt/src/test/java/org/neo4j/bolt/v1/messaging/BoltRequestMessageWriter.java
lutovich: True :)
We do not really use struct size for predefined messages like INIT or RUN. It is kind of ignored [here](https://github.com/neo4j/neo4j/blob/3.4.0-alpha04/community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/BoltRequestMessageReader.java#L58). Code decides how to unpack based on the struct signature. Worth fixing anyway, I'll create cards.

community/bolt/src/main/java/org/neo4j/bolt/v1/messaging/Neo4jPackV1.java
zhenlineo: @pontusmelke Is it correct that`Neo4jPackV1.Packer` is actually a bolt impl of `AnyValueWriter` and `PackStream.Packer` is a bolt impl of `ValueWriter`?

Then maybe the long term cleanup plan would be make PackStream.Packer impl `ValueWriter` and `Neo4jPackV1.Packer` impl `AnyValueWriter`?

zhenlineo (commented):
Vote for keeping as it is and we will try to clean in later versions.

PR #10726, opened 2018-01-04
Use primitive long diff set for index state diffs

MishaDemianenko:
Introduce specialized primitive long version of diffsets.
Use primitive long diff set for index state diffs.


tinwelint (approved):
LGTM

PR #10729, opened 2018-01-05
Adds means of more efficiently calling NumberArray#swap

tinwelint:
By introducing NumberArray#duplicate(), which can create intermediary data structure
for swapping (and potentially for other things too) once instead of for every
call to the swap method.

Those who have heavy parallel load and do lots of swap calls should consider grabbing
a duplicate of their number array and swapping on that instead of directly on the
shared number array.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/DynamicNumberArray.java
+++ b/community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/cache/DynamicNumberArray.java
@@ -111,4 +111,10 @@ abstract class DynamicNumberArray<N extends NumberArray<N>> implements NumberArr
             chunk.close();
         }
     }
+
+    @Override
+    public N duplicate()
+    {
+        throw new UnsupportedOperationException( "Not supported for dynamic arrays" );
+    }
 }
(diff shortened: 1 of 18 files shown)


tinwelint:
@chrisvest we explored the xor approach a little, but didn't get to a running thing and so couldn't compare the two approaches. Perhaps it may still be interesting, but for now this _should_ be enough.

chrisvest:
I'm not a fan of this approach. Not supporting `DynamicNumberArray` violates Liskov's Substitution Principle. Insisting on copying whole elements means you're forced to allocate, while not giving you any atomicity anyway since the inner single-element-copy loops are internally piecewise anyway.

Why not copy piecewise directly, keeping intermediate state in local variables? Then you could copy in a single iteration if you work byte-wise, or two loops if you try long-wise first as an possible optimisation.

tinwelint:
You're right. This perhaps makes things unnecessarily complicated and leaves dynamic arrays behind a bit (this approach is a bit difficult to support for dynamic arrays, something I didn't become aware of until actually going ahead and implementing this).

I'll give your idea a shot instead.

tinwelint:
See #10734 instead

PR #10731, opened 2018-01-05
No unnecessary copies when rewriting.

sherfert:
We created a lot of unnecessary copies of case classes that have Seq's as children , if the
used implementation was a ListBuffer. We don't do that any more.

### DIFF
--- a/community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/Rewritable.scala
+++ b/community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/Rewritable.scala
@@ -34,7 +34,11 @@ object Rewriter {
 
 object RewriterWithArgs {
   def lift(f: PartialFunction[(AnyRef, Seq[AnyRef]), AnyRef]): RewriterWithArgs =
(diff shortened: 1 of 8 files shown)


community/cypher/util-3.4/src/main/scala/org/neo4j/cypher/internal/util/v3_4/Rewritable.scala
```diff
@@ -34,7 +34,11 @@ object Rewriter {
 
 object RewriterWithArgs {
   def lift(f: PartialFunction[(AnyRef, Seq[AnyRef]), AnyRef]): RewriterWithArgs =
-    f.orElse(PartialFunction({ case (a: AnyRef, _) => a }))
+    f.orElse(PartialFunction({
+      // We need to dup anything not matched by f given the children
+      case (p: Product, children) => new DuplicatableProduct(p).dup(children).asInstanceOf[AnyRef]
+      case (a: AnyRef, children) => new DuplicatableAny(a).dup(children)
```
fickludd: Arguably we no longer need implicit classes here, but could just call functions instead with the instance as argument?

fickludd (approved):
Nice find! Could speed up stuff quite a bit in big trees I think.

PR #10733, opened 2018-01-05
AFS to only gunk up part of byte[] it has access to

jakewins:
Before this commit, AdversarialFileSystem will inject random data
into the complete array given to `getBytes(..)`; meaning a call
like `getBytes(bigBuffer, 1000, 10)` could lead to the entire
buffer being messed up, rather than just the range the user has
asked to read into.

This changes the AFS to only "mess up" the section of the array
that would actually be written to by Muninn.

This also resolves a bug in EFS, reading buffers larger than the EFS 
channel buffer.

### DIFF
--- a/community/io/src/test/java/org/neo4j/adversaries/pagecache/AdversarialReadPageCursorTest.java
+++ b/community/io/src/test/java/org/neo4j/adversaries/pagecache/AdversarialReadPageCursorTest.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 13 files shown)


chrisvest (approved):
👍

community/io/src/test/java/org/neo4j/adversaries/pagecache/AdversarialReadPageCursorTest.java
```diff
@@ -0,0 +1,35 @@
+package org.neo4j.adversaries.pagecache;
```
chrisvest, line 20: Needs a license header.

community/io/src/test/java/org/neo4j/test/impl/ChannelOutputStreamTest.java
```diff
@@ -0,0 +1,41 @@
+package org.neo4j.test.impl;
```
chrisvest, line 20: One more license header.

community/io/src/test/java/org/neo4j/adversaries/pagecache/AdversarialReadPageCursorTest.java
MishaDemianenko: mvn chris:checkLicensePlease :)

jakewins:
Merged manually

PR #10734, opened 2018-01-05
Improved ByteArray#swap

tinwelint:
by avoiding any local memory allocation, instead copying the data byte-wise,
or if possible piece-wise (long,int,short).


tinwelint:
@chrisvest better?

chrisvest (approved):

PR #10735, opened 2018-01-05
Add support for PF_NO_FAULT to the MuninnPageCache.

chrisvest:
This will be useful for future features that either want to extract a profile about what store data is, and is not, in memory, or for features that desire to off-load their IO to other threads, etc.

### DIFF
--- a/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnReadPageCursor.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnReadPageCursor.java
@@ -44,7 +44,7 @@ final class MuninnReadPageCursor extends MuninnPageCursor
             pinEvent.done();
         }
         lockStamp = 0; // make sure not to accidentally keep a lock state around
-        clearPageState();
+        clearPageCursorState();
     }
 
     @Override
@@ -56,9 +56,9 @@ final class MuninnReadPageCursor extends MuninnPageCursor
         {
             return false;
         }
-        pin( nextPageId, false );
         currentPageId = nextPageId;
         nextPageId++;
+        pin( currentPageId, false );
         return true;
     }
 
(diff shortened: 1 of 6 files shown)


community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
```diff
@@ -5210,4 +5211,244 @@ public void eagerFlushMustWriteToFileOnUnpin() throws Exception
...
+            {
+                writer.close();
+            }
+            // If the page is evicted, then our read must have been inconsistent.
+            assertTrue( nofault.shouldRetry() );
+            // However, we are no longer in memory, because the page we had earlier got evicted.
+            verifyNoFaultReadIsNotInMemory( nofault );
+        }
+    }
+    // todo no fault read cursor must allow eviction of page
```
MishaDemianenko: todo

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnReadPageCursor.java
```diff
@@ -56,9 +56,9 @@ public boolean next() throws IOException
         {
             return false;
         }
-        pin( nextPageId, false );
         currentPageId = nextPageId;
         nextPageId++;
+        pin( currentPageId, false );
```
MishaDemianenko, line 61: why do we move page pin below assigning current page id?

MishaDemianenko (commented):
One small question and one todo found, otherwise looks good

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnReadPageCursor.java
chrisvest: Because `pin` will set `currentPageId = UNMAPPED_PAGE_ID` in `uncommonPin` when `PF_NO_FAULT` is enabled.

MishaDemianenko (approved):

community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
chrisvest: 👍 Removed it. It was a leftover reminder about the test just above.

PR #10736, opened 2018-01-05
Use cursor directly instead of intermediate iterator

pontusmelke:
By having direct access to the cursor we can get a significant speedup
instead of delegating via an iterator when doing label scans.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIndexProgressor.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIndexProgressor.java
@@ -47,16 +47,18 @@ class LabelScanValueIndexProgressor extends LabelScanValueIndexAccessor implemen
     }
 
     /**
-     * @return next node id in the current {@link LabelScanValue} or, if current value exhausted,
-     * goes to next {@link LabelScanValue} from {@link RawCursor}. Returns {@code true} if next node id
-     * was found, otherwise {@code false}.
+     *  Progress through the index until the next accepted entry.
+     *
+     *  Progress the cursor to the current {@link LabelScanValue}, if this is not accepted by the client or if current
+     *  value is exhausted it continues to the next {@link LabelScanValue}  from {@link RawCursor}.
+     * @return <code>true</code> if an accepted entry was found, <code>false</code> otherwise
      */
     @Override
     public boolean next()
     {
-        while ( true )
+        for ( ; ; )
         {
-            if ( bits != 0 )
+            while ( bits != 0 )
             {
                 int delta = Long.numberOfTrailingZeros( bits );
                 bits &= bits - 1;
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIndexProgressor.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIndexProgressor.java
@@ -22,53 +22,28 @@ package org.neo4j.kernel.impl.index.labelscan;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.Collection;
-import java.util.NoSuchElementException;
 
-import org.neo4j.collection.primitive.PrimitiveLongCollections;
-import org.neo4j.collection.primitive.PrimitiveLongIterator;
-import org.neo4j.collection.primitive.PrimitiveLongResourceIterator;
 import org.neo4j.cursor.RawCursor;
+import org.neo4j.graphdb.Resource;
 import org.neo4j.index.internal.gbptree.Hit;
+import org.neo4j.storageengine.api.schema.IndexProgressor;
 
 /**
- * {@link PrimitiveLongIterator} which iterate over multiple {@link LabelScanValue} and for each
+ * {@link IndexProgressor} which steps over multiple {@link LabelScanValue} and for each
  * iterate over each set bit, returning actual node ids, i.e. {@code nodeIdRange+bitOffset}.
  *
- * The provided {@link RawCursor} is managed externally, e.g. {@link NativeLabelScanReader},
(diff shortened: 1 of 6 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
```diff
@@ -32,6 +32,8 @@
 import org.neo4j.graphdb.index.Index;
 import org.neo4j.index.internal.gbptree.GBPTree;
 import org.neo4j.index.internal.gbptree.Hit;
+import org.neo4j.kernel.impl.index.schema.NumberHitIndexProgressor;
```
burqen: Unused import

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanReader.java
```diff
@@ -108,6 +110,23 @@ public PrimitiveLongResourceIterator nodesWithAllLabels( int... labelIds )
         return new CompositeLabelScanValueIterator( iterators, true );
     }
 
+    public LabelScanValueIndexProgressor nodesWithLabelIndex( IndexProgressor.NodeLabelClient client,
```
burqen: Missing `@Override`

community/kernel/src/main/java/org/neo4j/storageengine/api/schema/LabelScanReader.java
```diff
@@ -33,6 +33,15 @@
...
     PrimitiveLongResourceIterator nodesWithLabel( int labelId );
 
+    /**
+     * Returns the index progressor over the label id
+     *
+     * @param client the client to communicate with
+     * @param labelId label token id
+     * @return index progressor with the given {@code labelId}
+     */
+    IndexProgressor nodesWithLabelIndex( IndexProgressor.NodeLabelClient client, int labelId );
```
burqen: Got some comments about this guy.

1. This is the same pattern that we use for `IndexReader`. In that case, the full responsibility for connecting progressor and client lies with implementation and the progressor is never visible from from the outside. Here we instead return the progressor and leave the responsibility of initializing the client with the progressor to the caller. I would prefer that we use the same usage pattern in both cases. Unless there is some reason for the different approaches that I don't see.

2. I think adding "index" to this method name adds more confusion than it removes. I would prefer that we use the same method name as above , `nodesWithLabel`.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIndexProgressor.java
```diff
@@ -0,0 +1,89 @@
...
+        this.client = client;
+    }
+
+    /**
+     * @return next node id in the current {@link LabelScanValue} or, if current value exhausted,
+     * goes to next {@link LabelScanValue} from {@link RawCursor}. Returns {@code true} if next node id
+     * was found, otherwise {@code false}.
+     */
+    @Override
+    public boolean next()
```
burqen: This javadoc describe the iterator version of `fetchNext`.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIndexProgressor.java
```diff
@@ -0,0 +1,89 @@
...
+    {
+        while ( true )
+        {
+            if ( bits != 0 )
```
burqen: I think this should be a while loop. So that we exhaust those bits if client does not accept, before returning to next label bitset.

burqen (changes requested):
Sweet! I think it is a very good idea to use the progressor-client pattern here. I found some points of improvement and I got some suggestions / questions.

1. Do we have any benchmark results?

2. May I also suggest that we use the same pattern for
`Read.nodeLabelUnionScan` and `Read.nodeLabelIntersectionScan`? (Not necessarily in this PR)

3. Just as for `IndexReader`, we create duplication of methods in `LabelScanReader` interface where we do/don't use progressor-client pattern. This duplication could be removed with `OperationsFacade` later on, right?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/LabelScanValueIndexProgressor.java
chrisvest: Note that `while ( true ) …` can also be written as `for (;;) …` where the jump is unconditional.

pontusmelke:
@burqen 
1. Yes, about 30-40 percent faster. I'll send you a link.
2. I agree, I made one attempt but it gets a little tricky since we cannot compose the "progessors" as we do with iterators because of how they interact with the client. If we "compose" RawCursors  it also gets a little tricky since we must keep track of both the `LabelScanKey` and `LabelScanValue` in order to compute the next long - then we're almost back at having some sort of `PrimitiveLongIterator`. I am sure there is a nice solution out there - so let's talk about it and we'll fix it in another PR.
3. Yes

burqen:
@pontusmelke Nice! I'm almost happy. What's your thoughts about my comment on 

```
public interface LabelScanReader extends Resource
{
    IndexProgressor nodesWithLabelIndex( IndexProgressor.NodeLabelClient client, int labelId );
}
```
?

"1. This is the same pattern that we use for IndexReader. In that case, the full responsibility for connecting progressor and client lies with implementation and the progressor is never visible from from the outside. Here we instead return the progressor and leave the responsibility of initializing the client with the progressor to the caller. I would prefer that we use the same usage pattern in both cases. Unless there is some reason for the different approaches that I don't see."

pontusmelke:
@burqen Something like this?

burqen:
@pontusmelke Yep! 👍

burqen (approved):

PR #10741, opened 2018-01-07
Lucene PageCache backend

jakewins:
As discussed. This is split into three commits: the first  is just a micro version of https://github.com/neo4j/neo4j/pull/10733 to get this green, one is the actual implementation, and the last is plugging the implementation in. 

In this PR, the default DirectoryFactory is set to use this implementation, to exercise it in the tests.

---

Note: This is an unsanctioned proposal, if you are an outside
reader please do not expect this to be merged.

Kernel: As discussed. Benchmarks forthcoming, opening PR to discuss
feasibility of this. Suggested end-game, if we decide to do this, would
be to introduce this behind an unsanctioned toggle, pickle in
soak testing for a few months and then evaluate.

Overview & Safety
--

This provides an implementation of the Lucene Directory API that uses
`FileSystemAbstraction` for writes and `PageCache` for reads.

It is modelled after, and is intended to inherit the safety aspects of,
the `MMapDirectory` that ships with Lucene. In fact, it is precisely the
same as `MMapDirectory`, except is uses the `PageCache` as its buffer
backend for index inputs.

Since Lucene writes to immutable files, and since it does not read
from those files before they are fully written, this implementation
has the same safety properties as MMapDirectory: Safety comes from
proper fsyncing of writes, and this implementation shares the exact
same write behavior as MMapDirectory.

Hence: Review of this should, outside of regular crash fuzzing, verify
that the above actually holds true: The write-side implementation should
be equivalent to that of MMapDirectory.

Background
--

Lucene provides multiple storage backends:

- RAMDirectory
    in-memory, used by kernel in Ephemeral mode

- NIOFSDirectory
    persistent, loads fixed chunks of indexes into a buffer
    as they are read; meaning lots of IO. Used by kernel if
    -Dorg.neo4j.kernel.api.impl.index.storage.DirectoryFactory.default_directory_factory=true

- MMapDirectory
    persistent, memory maps whole indexes as they are read,
    used by kernel by default.

RAMDirectory is not persistent. NIOFSDirectory is slow, so MMapDirectory
is used by default.

However, there are issues with MMapDirectory as well:

*1. Swapping*

Because MMapDirectory memory maps entire indexes, any non-trivial
deployment will be forced to swap the Neo4j process. This introduces at
least two issues. 1, there is no way to operate Neo4j in an environment
without swap space and, 2, memory management of Neo4j is expanded to
require operators to consider their virtual memory limits on day one.

Since standard AWS images do not have swap enabled by default, this
poses a substantial UX pain: Most users will experience this by the
Neo4j process simply consuming indefinite amounts of memory, with no
indication that turning on swap would solve the problem.

There may also be problems with this swapping then inducing competition
between hot indexes and runtime data structures in the database itself,
although this isn't entirely clear.

*2. Java MappedByteBuffer*

As we are acutely aware, the Java memory mapping APIs have some
undesireable characteristics. Lucene tries to alleviate this by
using the Cleaner API.
While this somewhat reduces GC pain, the BBs objects are still stuck
on ReferenceQueues by the Cleaner API, increasing GC work during
safepoints. It also means we're implicitly calling unsafe.freeMemory(address),
and memory mapping, over and over as index readers are opened and closed.

Rationale
---

This implementation thus provides two benefits:

1) It allows Neo4j to run in environments with no or limited Swap
2) It gives Neo4j predictable and configurable memory use
3) It removes substantial numbers of `mmap/dealloc` from hot paths


jakewins:
@chrisvest updated as discussed; added the NOT_READY_FOR_MERGE back on, remembering we need to set the default back to `niofs` before this goes anywhere, and the other PR should go in first.

jakewins:
@chrisvest updated again; one of the benchmark suites came across a bug. `getFilePointer` is supposed to return the actual position in the underlying actual file, no matter if the input is a clone of a clone of slice of a slice of a clone. 

Added a fix and some tests for that, which made the benchmark happy. 

Also added an optimized implementation of `skipBytes`; the default the parent class simply skips by reading into a throw-away buffer; expanded the tests in `PagedIndexInputTest` to test every read operation with both `seek` and `skipBytes`

jakewins:
Just kidding; apparently `getFilePointer` was supposed to work as I'd originally implemented it; the problem was that this test failed:

    input.seek(1337);
    assert input.getFilePointer() == input.clone().getFilePointer()

Because

    assert input.clone().getFilePointer() == 0

PR has been modified to correctly set the "current" position of a clone to be == the position of the source you're cloning from.

I guess the other change made my repro of the benchmark issue pass by random chance(?) - fortunately it caused like a billion regular tests to fail.

Assuming tests go green from this, this is ready for a proper review methinks.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
```diff
@@ -0,0 +1,219 @@
...
+ * <p>
+ * The root input has an instance of {@link RootInputResources}, and the clones
+ * all share an instance of {@link CloneInputResources}.
+ * <p>
+ * Each input has it's own {@link PageCursor}, which needs closing, and all
+ * inputs from the same root
+ * share a {@link PagedFile}. Cursors are closed if/when the input associated
+ * with the cursor is closed,
+ * and everything still open is closed when the root input is closed.
+ * <p>
```
chrisvest: Good comment 👍

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedDirectory.java
```diff
@@ -0,0 +1,138 @@
...
+        ensureOpen();
+        fs.renameFile(
+                path.resolve( source ).toFile(),
+                path.resolve( dest ).toFile(),
+                ATOMIC_MOVE );
+        fs.forceDirectory( path );
+    }
+
+    @Override
+    public synchronized void close()
```
chrisvest: I don't see any other method that this synchronisation would avoid overlapping with.

chrisvest: The `isOpen` field is volatile anyway.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
```diff
@@ -0,0 +1,456 @@
...
+        if ( pageSize - pageOffset < 2 )
+        {
+            long originalPageId = currentPageId;
+            int originalPageOffset = currentPageOffset;
+            try
+            {
+                currentPageId = pageId;
+                currentPageOffset = pageOffset;
+                // Delegate to parent, which reads byte-by-byte
+                return super.readShort();
```
chrisvest: This can produce torn reads because the shouldRetry only protects the individual bytes, and not the word. Hmm… maybe that's technically fine, since the mapped files are read-only, so you're only duking out with eviction?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
```diff
@@ -0,0 +1,456 @@
...
+     * @return the logical position of this slice, 0 being start of slice
+     */
+    @Override
+    public long getFilePointer()
+    {
+        try
+        {
+            return currentPageId * pageSize + currentPageOffset - startPosition;
+        }
+        catch ( NullPointerException npe )
```
chrisvest: What can be null here?

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/storage/paged/FSALockFactoryTest.java
```diff
@@ -0,0 +1,169 @@
...
+{
+    private static final int N_THREADS = 7;
+    private static final int SECONDS_TO_RUN = 5;
+
+    @Rule
+    public EphemeralFileSystemRule fs = new EphemeralFileSystemRule();
+    @Rule
+    public PageCacheRule pageCacheRule = new PageCacheRule();
+    @Rule
+    public TemporaryFolder tmpFolder = new TemporaryFolder();
```
chrisvest: `RuleChain` these guys to ensure they're stacked in a predictable order.

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/storage/paged/FSALockFactoryTest.java
```diff
@@ -0,0 +1,169 @@
...
+                String assertionError =
+                        result.get( SECONDS_TO_RUN * 100, TimeUnit.SECONDS );
+                if ( assertionError != null )
+                {
+                    fail( assertionError );
+                }
+            }
+            catch ( InterruptedException | ExecutionException | TimeoutException e )
+            {
+                throw Exceptions.launderedException( e );
```
chrisvest: `Exceptions.launderedException` is being removed, so avoid calling it here. Just wrap and throw `RuntimeException`.

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/storage/paged/FSALockFactoryTest.java
chrisvest: Also, why not `TestDirectory`?

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInputTest.java
```diff
@@ -0,0 +1,444 @@
...
+public class PagedIndexInputTest
+{
+    @Rule
+    public TemporaryFolder tmpDir = new TemporaryFolder();
+
+    @Rule
+    public FileSystemRule fs = new EphemeralFileSystemRule();
+
+    @Rule
+    public PageCacheRule pageCache = new PageCacheRule();
```
chrisvest: I'd `RuleChain` and `TestDirectory` these guys.

community/io/src/main/java/org/neo4j/io/fs/FileSystemAbstraction.java
```diff
@@ -88,22 +86,12 @@
...
 
     void deleteFileOrThrow( File file ) throws IOException;
 
-    interface ThirdPartyFileSystem extends Closeable
-    {
-        @Override
-        void close();
-
-        void dumpToZip( ZipOutputStream zip, byte[] scratchPad ) throws IOException;
-    }
```
chrisvest, line 105: 🎉

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
```diff
@@ -52,84 +51,140 @@
...
-        private final boolean USE_DEFAULT_DIRECTORY_FACTORY =
-                FeatureToggles.flag( DirectoryFactory.class, "default_directory_factory", true );
+        boolean isEphemeral = config.get( GraphDatabaseFacadeFactory.Configuration.ephemeral );
+        boolean isUsingCustomIOConfig = CustomIOConfigValidator.customIOConfigUsed( config );
+
+        if ( isEphemeral )
+        {
+            // If the database is in ephemeral mode, we need to use the PagedDirectory, since that
+            // is the only directory mode that works through the FileSystemAbstraction.
+            assert !isUsingCustomIOConfig : "Can't both be in ephemeral mode and use block storage, that makes no sense.";
```
chrisvest: I don't think it's the responsibility to check this here.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
```diff
@@ -52,84 +51,140 @@
...
+        case "paged":
+            if ( !isUsingCustomIOConfig )
+            {
+                factory = new PagedDirectoryFactory( pageCache );
+                break;
+            }
+            System.err.println( "WARN: Paged indexes are not available when custom IO is in use. " +
+                    "Most likely this is because you're running on special hardware. Falling back to memory-mapped indexes." );
+        case "niofs":
+            factory = new MemoryMappedDirectoryFactory();
```
chrisvest: Should "niofs" not be the same as heap? The MemoryMappedDirectoryFactory should respond to "mmap" and "default", I think.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
```diff
@@ -22,27 +22,26 @@
...
 import org.neo4j.unsafe.impl.internal.dragons.FeatureToggles;
 
-import static java.lang.Math.min;
-
-public interface DirectoryFactory extends FileSystemAbstraction.ThirdPartyFileSystem
+public interface DirectoryFactory extends AutoCloseable
 {
+    int MAX_MERGE_SIZE_MB = FeatureToggles.getInteger( DirectoryFactory.class, "max_merge_size_mb", 5 );
+    int MAX_CACHED_MB = FeatureToggles.getInteger( DirectoryFactory.class, "max_cached_mb", 50 );
+    String DEFAULT_FACTORY = FeatureToggles.getString( DirectoryFactory.class, "factory", "paged" );
```
chrisvest: I don't think this should be the default just yet.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
```diff
@@ -52,84 +51,140 @@
...
+                factory = new PagedDirectoryFactory( pageCache );
+                break;
+            }
+            System.err.println( "WARN: Paged indexes are not available when custom IO is in use. " +
+                    "Most likely this is because you're running on special hardware. Falling back to memory-mapped indexes." );
+        case "niofs":
+            factory = new MemoryMappedDirectoryFactory();
+            break;
+        default:
+            System.err.printf( "WARN: Unknown index IO implementation '%s', using default.\n", DEFAULT_FACTORY );
```
chrisvest: Would it be possible to get some logging in here, instead of `System.err.println`s?

community/lucene-index/src/test/java/org/neo4j/index/lucene/ConstraintIndexFailureIT.java
```diff
@@ -52,6 +55,8 @@
     public final TestDirectory storeDir = TestDirectory.testDirectory();
     @Rule
     public final DefaultFileSystemRule fileSystemRule = new DefaultFileSystemRule();
+    @Rule
+    public final PageCacheRule pageCacheRule = new PageCacheRule();
```
chrisvest: A good time to `RuleChain`, I think.

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/LuceneSchemaIndexPopulationIT.java
```diff
@@ -61,6 +63,8 @@
     public TestDirectory testDir = TestDirectory.testDirectory();
     @Rule
     public final DefaultFileSystemRule fileSystemRule = new DefaultFileSystemRule();
+    @Rule
+    public final PageCacheRule pageCacheRule = new PageCacheRule();
```
chrisvest: Again.

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/LuceneSchemaIndexUniquenessVerificationIT.java
```diff
@@ -71,6 +72,8 @@
     public TestDirectory testDir = TestDirectory.testDirectory();
     @Rule
     public final DefaultFileSystemRule fileSystemRule = new DefaultFileSystemRule();
+    @Rule
+    public final PageCacheRule pageCacheRule = new PageCacheRule();
```
chrisvest: I guess we have a lot of these.

community/lucene-index/src/test/java/org/neo4j/index/PagedDirectoryRule.java
```diff
@@ -0,0 +1,78 @@
...
+import org.neo4j.test.rule.ExternalResource;
+import org.neo4j.test.rule.PageCacheRule;
+import org.neo4j.test.rule.fs.EphemeralFileSystemRule;
+
+public class PagedDirectoryRule extends ExternalResource implements DirectoryFactory
+{
+    @Rule
+    public EphemeralFileSystemRule fileSystemRule = new EphemeralFileSystemRule();
+    @Rule
+    public PageCacheRule pageCacheRule = new PageCacheRule();
```
chrisvest: Can you really nest rules like this? If so, then you need to rule chain them to ensure their order.

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/schema/ReadOnlyLuceneSchemaIndexTest.java
```diff
@@ -44,19 +45,23 @@
     private final TestDirectory testDirectory = TestDirectory.testDirectory();
     private final ExpectedException expectedException = ExpectedException.none();
     private final DefaultFileSystemRule fileSystemRule = new DefaultFileSystemRule();
+    private final PageCacheRule pageCacheRule = new PageCacheRule();
 
     @Rule
     public RuleChain ruleChain = RuleChain.outerRule( testDirectory )
-            .around( expectedException ).around( fileSystemRule );
+            .around( expectedException ).around( fileSystemRule ).around( pageCacheRule );
```
chrisvest, line 53: 👍

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/schema/populator/UniqueDatabaseIndexPopulatorTest.java
```diff
@@ -99,7 +99,7 @@ public void setUp() throws Exception
     {
         File folder = testDir.directory( "folder" );
         indexStorage = new PartitionedIndexStorage( directoryFactory, fileSystemRule.get(), folder, false );
-        index = LuceneSchemaIndexBuilder.create( descriptor, Config.defaults() )
+        index = LuceneSchemaIndexBuilder.create( descriptor, Config.defaults(), directoryFactory)
```
chrisvest: Checkstyle; missing space at the end there.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProviderImpl.java
```diff
@@ -65,27 +67,26 @@
     /**
      * Creates a provider of fulltext indices for the given database. This is the entry point for all fulltext index
      * operations.
-     *
-     * @param db Database that this provider should work with.
+     *  @param db Database that this provider should work with.
```
chrisvest: Stray space.

chrisvest (changes requested):
I had some comments. One thing that I didn't mention everywhere, is that I'd like to see the use of RuleChain be spread to the tests that are touched here.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
jakewins: ack

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
jakewins: +1

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
jakewins: +1

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedDirectory.java
jakewins: +1, I think this came from the Directory impl I was using as a template for implementing this

jakewins: Double checked this; this came from FSDirectory; but it likewise doesn't have anything else that's synchronized, so I guess that's an artifact of some code that no longer exists in the Lucene code base.. anyway, removing.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
jakewins: Ah, forgot to remove this

community/lucene-index/src/test/java/org/neo4j/index/lucene/ConstraintIndexFailureIT.java
jakewins: +1

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/LuceneSchemaIndexPopulationIT.java
jakewins: +1

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/LuceneSchemaIndexUniquenessVerificationIT.java
jakewins: :)

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/storage/paged/FSALockFactoryTest.java
jakewins: I've not yet figured out the benefit of it vs. the `TemporaryFolder` that's built into junit.. and the junit one doesn't live in `test-jar` dependency, so it reduces issues from transient dependencies not working.

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/storage/paged/FSALockFactoryTest.java
jakewins: +1

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInputTest.java
jakewins: +1

community/lucene-index/src/test/java/org/neo4j/kernel/api/impl/schema/populator/UniqueDatabaseIndexPopulatorTest.java
jakewins: ack

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/FulltextProviderImpl.java
jakewins: +1

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
jakewins: Hmm.. this is interesting, yeah.. I guess if we actually had concurrent writes, the only way to do this would be to write-lock the two pages? There's currently no way to do `shouldRetry` across two pages right?

That said, yes, this should be safe since we're guaranteed read-only

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
```diff
@@ -52,84 +51,140 @@
...
+     * larger ones, but the point remains that lots of tiny files are part of the basic operation of the index.
+     * <p>
+     * This class limits the actual physical file creation by caching the tiny files in heap buffers, attempting
+     * to merge them together into larger files, that in turn actually get written to disk.
+     */
+    final class NRTCachingDirectoryFactory implements DirectoryFactory
+    {
+        // Its unclear to me that this actually benefits us - we refresh readers constantly, wouldn't
+        // that force this to flush all the time? This adds substantial heap load by allocating
+        // buffers for every index write on the heap - does it bring a clear benefit?
```
MishaDemianenko: yes it is, using NRT for us gives a lot of perf benefits

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
MishaDemianenko: I'm sure it should not be the default one for old branches at least :)
Please keep current page directory default setup as soon as you done with your testing

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
```diff
@@ -231,6 +220,39 @@ public void deleteFileOrThrow( File file ) throws IOException
         return StreamFilesRecursive.streamFilesRecursive( directory, this );
     }
 
+    @Override
+    public void forceDirectory( Path dir )
```
MishaDemianenko: Our file system abstraction always operates on top of files and would be good if those newly introduced methods will follow that.

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
```diff
@@ -231,6 +220,39 @@ public void deleteFileOrThrow( File file ) throws IOException
...
+            // exceptions if assertions are disabled or the platform is not *nix. I'm leaving this code as a faithful
+            // implementation of their algorithm, but am noting we may want to convert this, at the very least, to a log
+            // message to track platforms this fails on?
+            assert !(SystemUtils.IS_OS_LINUX || SystemUtils.IS_OS_MAC) : "On Linux and MacOSX fsyncing a directory should not throw IOException, " +
+                    "we just don't want to rely on that in production (undocumented). Got: " + ioe;
+        }
+    }
+
+    @Override
+    public void force( Path path ) throws IOException
```
MishaDemianenko: Since Path (File) can represent both directory and file from FSA interface perspective it feels wrong that we have 2 different methods to sync file and directory separately. That's something that FSA should take care of instead of the user

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
```diff
@@ -231,6 +220,39 @@ public void deleteFileOrThrow( File file ) throws IOException
         return StreamFilesRecursive.streamFilesRecursive( directory, this );
     }
 
+    @Override
+    public void forceDirectory( Path dir )
+    {
+        // Lucene relies on flushing directory metadata to disk for commit safety; this code is from their way of doing that.
```
MishaDemianenko: it's a DFSA, do not think its a good place for comments about lucene even if this code is basically their IOUtil.fsync.
Can you please instead add a java doc for a method that will include some descri[ption together with openjdk mail chain, https://bugs.openjdk.java.net/browse/JDK-8080235 and blog post link instead, would be really nice if we can avoid this kind of things at all, also because we trying to add recent jdk version to a list of our lovely jdk's 

But all of that mentioned above is actually only in case we actually need those methods here. Since its a page cache directory and those are page cache managed files shouldn't we ask page cache to flush files instead?
Also since its that custom factory do we really need to create tons of those small files and rename those and as result sync directory at all?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
chrisvest: You can do shoudRetry across multiple pages by opening linked page cursors.

community/io/src/main/java/org/neo4j/io/fs/DelegateFileSystemAbstraction.java
```diff
@@ -204,10 +197,8 @@ public void renameFile( File from, File to, CopyOption... copyOptions ) throws I
     {
         try ( Stream<Path> listing = Files.list( path( directory ) ) )
         {
-            return listing
-                    .filter( entry -> filter.accept( entry.getParent().toFile(), entry.getFileName().toString() ) )
-                    .map( Path::toFile )
-                    .toArray( File[]::new );
+            return listing.filter( entry -> filter.accept( entry.getParent().toFile(), entry.getFileName().toString() ) ).map( Path::toFile ).toArray(
```
MishaDemianenko: now this line is a bit longer than my laptop's preferable size(

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/builder/AbstractLuceneIndexBuilder.java
```diff
@@ -54,6 +55,10 @@ public AbstractLuceneIndexBuilder( Config config )
      */
     public T withIndexStorage( PartitionedIndexStorage indexStorage )
     {
+        // TODO: I don't understand this; indexStorage has a directoryFactory,
```
MishaDemianenko: please either react to this todo fix it or remove since who else should react to it then. As I can see index storage you can pass index storage from the outside for a specific index by its id when you try to locate index and that looks like only one use case for this.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/builder/AbstractLuceneIndexBuilder.java
```diff
@@ -37,13 +37,14 @@
  */
 public abstract class AbstractLuceneIndexBuilder<T extends AbstractLuceneIndexBuilder<T>>
 {
-    protected LuceneIndexStorageBuilder storageBuilder = LuceneIndexStorageBuilder.create();
+    protected LuceneIndexStorageBuilder storageBuilder;
     private final Config config;
     private OperationalMode operationalMode = OperationalMode.single;
 
-    public AbstractLuceneIndexBuilder( Config config )
+    public AbstractLuceneIndexBuilder( Config config, DirectoryFactory defaultDirectoryFactory )
```
MishaDemianenko, line 44: also since now you make directoryFactory mandatory constructor parameter looks like 'withDirectoryFactory' method should be removed

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/FSALockFactory.java
```diff
@@ -0,0 +1,232 @@
...
+import org.neo4j.io.fs.FileSystemAbstraction;
+import org.neo4j.io.fs.OpenMode;
+import org.neo4j.io.fs.StoreChannel;
+
+/**
+ * Clone of {@link NativeFSLockFactory} for {@link FileSystemAbstraction}.
+ */
+public class FSALockFactory extends LockFactory
+{
+    // Lucene implements concurrency across multiple Directory instances
```
MishaDemianenko, line 47: Not sure that this is true in fact since they refer to a writer and not to a process; also couple of times because of the problems in a code when we  by mistake create multiple writers this locks saved us and failed fast, so i'm not sure that we need to remove them completely as locks, but we can avoid taking them as file locks.
Also since its a locks for page cache based indexes shouldn't we just skip taking any locks as file locks completely instead of mimic their NativeFSLockFactory?
Also, do not think its a good idea to disable this check at all for any of the indexes at least not completely

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
```diff
@@ -0,0 +1,205 @@
...
+public interface InputResources
+{
+    /**
+     * Resource tracking used by Root Inputs - inputs created by {@link Directory#openInput(String, IOContext)}.
+     */
+    class RootInputResources implements InputResources
+    {
+        private final PagedFile pagedFile;
+        private final CloneInputResources cloneResources = new CloneInputResources( this );
+        private LinkedList<PagedIndexInput> clones = new LinkedList<>();
```
MishaDemianenko: why the linked list? why? 🌵 😿

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
```diff
@@ -0,0 +1,205 @@
...
+            this.pagedFile = pagedFile;
+        }
+
+        @Override
+        public synchronized PageCursor openCursor( long pageId, PagedIndexInput owner ) throws IOException
+        {
+            // Called concurrently with #close(..)
+            assert owner.cursor == null : String.format( "Multiple cursors opened for the same input: %s", owner );
+
+            if ( clones == null )
```
MishaDemianenko: please introduce a boolean flag that for all future readers of this class will clearly explain that its a closed flag and do not use side effects to check a state, then you can be sure list of your cursors is always not null while you working with this input

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        @Override
+        public synchronized void close( PagedIndexInput input ) throws IOException
+        {
+            // Note that this gets called concurrently with `openCursor` if the index is closed while under load!
+            assert input.resources == this :
+                    String.format( "Closing input with unrelated resource tracker: " + "%s closed by %s", input, this );
+
+            if ( clones == null )
```
MishaDemianenko: same as above

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
```diff
@@ -0,0 +1,205 @@
...
+            {
+                // Already closed
+                return;
+            }
+
+            LinkedList<PagedIndexInput> localClones = this.clones;
```
MishaDemianenko: using a proper closed flag you will avoid doing this

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
```diff
@@ -0,0 +1,205 @@
...
+            this.clones = null;
+
+            // Close all clones
+            IOException errorOnClose = null;
+            for ( PagedIndexInput clone : localClones )
```
MishaDemianenko: IoUtils.closeAll

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedDirectory.java
```diff
@@ -0,0 +1,132 @@
...
+public class PagedDirectory extends BaseDirectory
+{
+    private final PageCache pageCache;
+    private final FileSystemAbstraction fs;
+
+    private final Path path; // The underlying filesystem path
+
+    public PagedDirectory( Path path, PageCache pageCache )
+    {
+        // TODO: We can probably use SingleInstanceLockFactory here
```
MishaDemianenko, line 48: i think we should not most probably for the same reason as described above

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
```diff
@@ -0,0 +1,441 @@
...
+import org.neo4j.io.pagecache.PagedFile;
+
+public class PagedIndexInput extends IndexInput implements RandomAccessInput
+{
+    /**
+     * Tracks open resources across cloned inputs, see {@link InputResources}.
+     */
+    final InputResources resources;
+
+    /** Star of the show */
```
MishaDemianenko, line 50: hmmm which one?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
```diff
@@ -0,0 +1,441 @@
...
+    public final void close() throws IOException
+    {
+        resources.close( this );
+    }
+
+    private void boundsCheck() throws EOFException
+    {
+        if ( cursor.checkAndClearBoundsFlag() )
+        {
+            throw new EOFException( "read past EOF: " + this );
```
MishaDemianenko: this exception message is a bit misleading, would be good to include positions, file name etc that   caused this as well

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
```diff
@@ -0,0 +1,441 @@
...
+            currentPageOffset -= pageSize;
+            assert currentPageOffset <= pageSize : "Should never read past two page boundaries";
+        }
+    }
+
+    private void moveCursor( long pageId, int offsetInPage ) throws IOException
+    {
+        if ( !cursor.next( pageId ) || isEOF( pageId, offsetInPage ) )
+        {
+            throw new EOFException( "read past EOF: " + this );
```
MishaDemianenko: same as above

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexOutput.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.apache.lucene.store.OutputStreamIndexOutput;
+
+import java.io.FilterOutputStream;
+import java.io.IOException;
+import java.nio.file.Path;
+
+import org.neo4j.io.fs.FileSystemAbstraction;
+
+// This should be replaced by writes through the page cache once the page
```
MishaDemianenko: block device support for this most probably will never come so why not just do it now? what are those features?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexOutput.java
```diff
@@ -0,0 +1,60 @@
...
+// cache has the feature set needed for this. Mainly this is around the
+// block device integration missing a few small features.
+class PagedIndexOutput extends OutputStreamIndexOutput
```
MishaDemianenko: please reference 'inspiration' of this class: FSIndexOutput
also would be nice to get https://issues.apache.org/jira/browse/LUCENE-5164 and 
http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/b1fec09a9588/src/share/native/java/io/io_util.c
in javadoc where people can find out of what air this constant comes from

MishaDemianenko (changes requested):
Have couple of comments here and there, sorry :)
Also would be nice to get a description how the layout of indexes looks like. Do we still create tons of files? and do we really need to(we never need more than 2 snapshots in the worst case, i think)? how page cache reacts to that huge amount of files over time? how this handles a huge number of indexes? more questions to come. And how it looks like from benchmarks perspective so far (microbenchmarks/soaks/ldbc)?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
jakewins: Ack - I'm keeping the "DO NOT MERGE" label on as a reminder to change the default back before merging :)

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/DirectoryFactory.java
jakewins: got it, removing this comment

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
jakewins: +1

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
jakewins: +1

community/io/src/main/java/org/neo4j/io/fs/DelegateFileSystemAbstraction.java
jakewins: Doh. I'd missed setting my line length properly, configured it to 120 now.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/builder/AbstractLuceneIndexBuilder.java
jakewins: Ack

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/FSALockFactory.java
jakewins: Agree with this; doing the only-in-process locking is what `SingleInstanceLockFactory` does. However, it felt a bit scary to make more than one major change at a time with Lucene - ie. I opted to keep the locking the same as we currently have (file locking via NativeFSLockFactory), so that if we hit issues with the changes this patch introduces there are fewer moving parts.

All to say: I agree, but think we should wait to make the change to in-process locking separately from this PR.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
jakewins: Haha, I guess because I just needed `append` and `iterate`, so it seemed like an efficient option.. are you thinking b/c of the heap overhead? I guess we do end up creating a hella lot of these index input objects.. 

The MMap implementation from Lucene uses `WeakIdentityMap` for this (see `ByteBufferIndexInput`), so I don't think we're adding overhead vs. current status quo.. 

WDYT of switching to have `PagedIndexInput.next`? Or something different?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
jakewins: Ack

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
jakewins: +1

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
jakewins: :)

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
jakewins: nice

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
jakewins: +1

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexOutput.java
jakewins: Things needed for block device support:

- expose atomic file renaming through page cache (or something similar)
- handle > 512 open files

Things needed for standard file system support:

- Ability to write a file where `length % page size != 0`

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexOutput.java
jakewins: +1

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/builder/AbstractLuceneIndexBuilder.java
jakewins: ack; I'd rather not expand the set of changes in this PR anymore; removing the TODO

community/io/src/main/java/org/neo4j/io/fs/DefaultFileSystemAbstraction.java
jakewins: Agree this is kind of.. broken. Clearly agree Lucene shouldn't be mentioned here - but at the same time, hard for future readers to really see context without talking about it, since it's the only reason it's here :/ I'm modifying to link to the JDK ticket and not talk directly about Lucene.

On need for this: As you noticed further down, we only read via the Page Cache; reasons for why added to that comment below. That means that, at least for now, we can't flush via the page cache. 

I guess this answers your earlier question about files and file formats: This uses (for now, at least) the *exact* same disk format as regular Lucene, and the *exact* same set of files. It is the same write path, just moved to happen via FSA. The only difference is that we do reads via the Page Cache instead of memory mapping.

jakewins:
@MishaDemianenko addressed! See response inline re: format. Benchmarks for LDBC so far show no difference from this change - a bit of a bummer, it'd be nice to get performance gains on top of the memory benefit, but I guess we can't get everything we want... 

I'm still waiting on the bulk of the micro benchmarks to come back, will update once I've got data from there.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
MishaDemianenko: lets do `PagedIndexInput.next`:)

MishaDemianenko:
@jakewins ok thx

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/InputResources.java
jakewins: Done! 

Note that I extended `IOUtils` to offer an option for `closeAll` that doesn't allocate arrays; otherwise removing the linked list here kind of is undone by having to convert the `PagedInput.next`-based list into something `closeAll` can consume. Now the `PagedInput.close` is GC-free. 

Might be worth looking at keeping `PagedIndexInput` on a freelist in future diffs, when I was testing for Lucenes `close` behavior I noted it opened several hundred inputs for a simple read query..

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
```diff
@@ -0,0 +1,531 @@
...
+
+    private void moveCursorTo( long pageId, int offsetInPage ) throws IOException
+    {
+        if ( isEOF( pageId, offsetInPage ) )
+        {
+            throw newEOFException( String.format( "EOF at { pageId: %d, offsetInPage: %d } %s", pageId, offsetInPage,
+                    describeCurrentPosition() ) );
+        }
+
+        if ( cursor.getCurrentPageId() != pageId && !cursor.next( pageId ) )
```
chrisvest: `cursor.next( pageId )` already does this current page id check.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
```diff
@@ -0,0 +1,531 @@
...
+    {
+        // Cross-page read?
+        int bytesInCurentPage = pageSize - pageOffset;
+        if ( bytesInCurentPage < 4 )
+        {
+            return readPageCrossingInt( pageId, pageOffset, bytesInCurentPage );
+        }
+
+        // Regular read
+        moveCursorTo( pageId, pageOffset );
```
chrisvest: This method seems to be pretty hot. I wonder if you could optimistically combine `pageId`, `pageOffset` and `moveCursorTo` into a single method call, and then go straight to the int-read loop. And then rely on `checkAndClearBoundsFlag` to indicate if you should throw away that result and try a cross-page instead instead, which in turn would figure out if we're actually really EOF or not.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/storage/paged/PagedIndexInput.java
```diff
@@ -0,0 +1,531 @@
...
+    public void seek( long pos ) throws IOException
+    {
+        long newPageId = pageId( pos );
+        int newOffset = pageOffset( pos );
+
+        if ( isEOF( newPageId, newOffset ) )
+        {
+            throw newEOFException(
+                    String.format( "seek EOF check failed for { pos: %d, pageId: %d, offsetInPage: %d }", pos,
+                            newPageId, newOffset ) );
```
chrisvest: With all these exceptions, try extracting all of the exception construction to a separate method. For instance in this case you could extract a `newSeekEofException( pos, newPageId, newOffset )`. That would reduce the amount of bytecode in the method, and give the JIT inliner more material to work with.

chrisvest (commented):
Just a couple of comments based on what shows up in the flamegraphs.

jakewins:
@chrisvest good notes; implemented the two simpler ones, and a variant on the optimistic readInt; I like the idea of it, and there's a variant in the branch now that moves the slow path to happen behind an optimistic fast path like you suggest. 

However, I didn't combine the `pageId,pageOffset,moveCursor` into one call, mainly because I'm not sure I understand the benefit of it - wouldn't we expect all that stuff to get inlined anyway? They are very tiny methods.. The drawback of doing it is it makes it a bit trickier to share the optimized code between `readInt(pos)` and `readInt()`. Or is what you're proposing to drop that in favor of duplication, since that'd get rid of the extra stack frame between `readInt(pos)` and `readInt(page, offset)`?

jakewins:
This is showing big regressions in LDBC, and I'm out of bandwidth to explore the root cause. Closing this for now, putting it to sleep in my fork..

PR #10744, opened 2018-01-07
Upgrade metrics version from 3.1.2 to 4.0.2

MishaDemianenko:


klaren (changes requested):
Changes looks good. Do note that `GraphiteOutput` is never tested so I think we should add one or do a simple manual test before merging if it's hard to automate.

klaren (approved):
Manually tested and works

PR #10745, opened 2018-01-08
Use new API for label scans

pontusmelke:
This PR accomplishes the following things:

1. Makes label scans transaction-state-aware in new API
2. Use the new API in interpreted runtime(s)
3. Use the new API in compiled runtime
4. Use the new API in Core API


community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/CantCompileQueryException.scala
```diff
@@ -25,5 +25,6 @@ import org.neo4j.cypher.internal.util.v3_4.spi.MapToPublicExceptions
 class CantCompileQueryException(message: String = "Internal error - should have used fall back to execute query, but something went horribly wrong", cause:Throwable=null)
   extends CypherException(message, cause) {
 
+  if (cause != null) cause.printStackTrace()
```
fickludd: Whoops

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/TransactionStateTestBase.java
```diff
@@ -518,6 +514,194 @@ public void shouldNotSeeDeletedNode() throws Exception
...
+    public void shouldFindUpdatedNodeInInLabelScan() throws Exception
+    {
+        // Given
+        Node node = createNode( "label" );
+
+        try ( org.neo4j.internal.kernel.api.Transaction tx = session.beginTransaction();
+              NodeLabelIndexCursor cursor = cursors.allocateNodeLabelIndexCursor() )
+        {
+            // when
+            tx.dataWrite().nodeAddLabel( node.node, node.labels[0] );
```
fickludd: Did you mean to re-assign the same label to the node, or should it be created with no label?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/TransactionStateTestBase.java
```diff
@@ -518,6 +514,194 @@ public void shouldNotSeeDeletedNode() throws Exception
...
+    {
+        // Given
+        Node node = createNode( "label1", "label2" );
+
+        try ( org.neo4j.internal.kernel.api.Transaction tx = session.beginTransaction();
+              NodeLabelIndexCursor cursor = cursors.allocateNodeLabelIndexCursor() )
+        {
+            // when
+            tx.dataWrite().nodeRemoveLabel( node.node, node.labels[1] );
+            tx.dataRead().nodeLabelUnionScan( cursor, node.labels );
```
fickludd, line 624: Maybe I'm confused. I though the Union scan should return all nodes with a least one of the provided labels? In that we should still find node as it retains `:label1`.

fickludd (changes requested):
I'm slightly unsure about the `nodeLabelUnionScan` so would like to discuss that before merging. It seems to behave the same as `nodeLabelIntersectScan` to me, but maybe I'm misreading somehow. I guess we don't really have any way to benefit from scanning over multiple layers yet anyway, as the label scan store is on single label? So we could just slam `UnsupportedOperationException` on these methods until the kernel can provide a more effective implementation or someone want to use them in the runtime.

Otherwise great test coverage and nice implementation.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/TransactionStateTestBase.java
pontusmelke: No the confusion is on me. I disabled `nodeIntersectionScan` for now since it is tricky to combine with tx state and currently not used.

fickludd (approved):

PR #10748, opened 2018-01-08
Support for dynamic sized keys and values in GBPTree

burqen:
## What?
Add support for dynamic sized keys and values to GBPTree. Size of a key+value combination can not be larger than half a page. Support for this will come later.

## Why?
Dynamic sized keys and values are needed for string index.

## How?
Every node in the tree has an offset array to the actual keys. This offset array is sorted, while the actual keys are added from right to left in an append only fashion.
```
LEAF:
P1:[HEADER][K1*|K2*|K3*].......[K1///|V1//][K3/|V3////][K2|V2]
                     ->         <-
INTERNAL:
P1:[HEADER][C0|K1*|C1|K2*|C2].......[K1///][K2/]
                           ->       <-
```
Remove is done by removing the pointer from the offset array and marking the actual key with a tombstone. Those will later be cleaned up when needed. We call this to defragment a node.

All logic for moving keys between nodes has moved from `InternalTreeLogic` to `TreeNode` implementation. This include merge, split and rebalance.


community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
```diff
@@ -0,0 +1,111 @@
...
+import static org.neo4j.index.internal.gbptree.PageCursorUtil.putUnsignedShort;
+
+/**
+ * keySize and valueSize indicate size of key and value in number of bytes.
+ *
+ * In LEAF:
+ * [keySize 2B|valueSize 2B|actualKey|actualValue]
+ *
+ * In INTERNAL:
+ * [keySize 2B|actualKey]
```
tinwelint: Is this outdated? INTERNAL has Child and key in each offset slow, right?

community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Check read key size for tombstone.
+     * @return True if read key size has tombstone.
+     */
+    static boolean hasTombstone( int readKeySize )
+    {
+        return (readKeySize & FLAG_TOMBSTONE) != 0;
+    }
+
+    static int readValueSize( PageCursor cursor )
```
tinwelint: Would be nice to explain the dependency between `readKeySize` `readValueSize` and `readKeyOffset`, reading them in order... a bit like how GSP(P)s are read.

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNodeDynamicSize.java
```diff
@@ -19,44 +19,188 @@
...
+import static org.neo4j.index.internal.gbptree.DynamicSizeUtil.readKeyOffset;
+import static org.neo4j.index.internal.gbptree.DynamicSizeUtil.readKeySize;
+import static org.neo4j.index.internal.gbptree.DynamicSizeUtil.readValueSize;
+import static org.neo4j.index.internal.gbptree.GenerationSafePointerPair.read;
+import static org.neo4j.index.internal.gbptree.TreeNode.Type.INTERNAL;
+import static org.neo4j.index.internal.gbptree.TreeNode.Type.LEAF;
+
 public class TreeNodeDynamicSize<KEY, VALUE> extends TreeNode<KEY,VALUE>
 {
+    private static final int BYTE_POS_ALLOCSPACE = BASE_HEADER_LENGTH;
```
tinwelint: Would be nice to document what ALLOCSPACE and other new concepts is, perhaps right here.

community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
```diff
@@ -0,0 +1,111 @@
...
+ * In INTERNAL:
+ * [keySize 2B|actualKey]
+ *
+ * First bit in keySize is used as a tombstone, set to 1 if key is dead.
+ * This leaves 15 bits for actual size -> max possible key size is 32768 bytes
+ * which is more than page size and therefore be enough.
+ */
+class DynamicSizeUtil
+{
+    static final int BYTE_SIZE_OFFSET = 2;
```
tinwelint: The naming scheme in TreeNode is otherwise `BYTE_POS...` for byte offsets in the page and `SIZE_...` for size (in bytes) of certain thing. These constants are named `BYTE_SIZE...` and so hurts the eye when reading. Would be great to have one scheme to stick to.

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNodeDynamicSize.java
```diff
@@ -19,44 +19,188 @@
...
+        int keyPosOffset = keyPosOffset( pos, type );
+        cursor.setOffset( keyPosOffset );
+
+        // Read actual offset to key
+        int keyOffset = readKeyOffset( cursor );
+
+        // Verify offset is reasonable
+        if ( keyOffset > pageSize )
+        {
+            cursor.setCursorException( "Tried to read key on offset " + keyOffset + ". Page size is " + pageSize );
```
tinwelint: PageCursor checks bounds, later propagated in `checkAndClearBoundsFlag()` doesn't it? Do we need to do this specifically here anyway?

community/index/src/test/java/org/neo4j/index/internal/gbptree/SimpleByteArrayLayout.java
```diff
@@ -145,4 +148,29 @@ private int byteArrayCompare( byte[] a, byte[] b )
...
+        RawBytes key = newKey();
+        key.bytes = ByteBuffer.allocate( 16 ).putLong( seed ).putLong( seed ).array();
+        return key;
+    }
+
+    @Override
+    public RawBytes value( long seed )
+    {
+        RawBytes value = newValue();
+        value.bytes = ByteBuffer.allocate( 17 ).putLong( seed ).putLong( seed ).array();
```
tinwelint: Why 17, is there some magic where value needs to be one byte bigger or something, or is it just for fun?

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNode.java
```diff
@@ -329,15 +341,14 @@ static void goTo( PageCursor cursor, String messageOnError, long nodeId )
...
      *
      * Keys and values from left are divide between left and right and the new key and value is inserted where it belongs.
      *
      * Key count is updated.
      */
-    abstract void doSplitLeaf( PageCursor leftCursor, int leftKeyCount, PageCursor rightCursor, int rightKeyCount, int insertPos,
-            KEY newKey,
-            VALUE newValue, int middlePos );
+    abstract void doSplitLeaf( PageCursor leftCursor, int leftKeyCount, PageCursor rightCursor, int insertPos, KEY newKey, VALUE newValue,
+            StructurePropagation<KEY> structurePropagation );
```
tinwelint: It looks like only the `rightKey` from `StructurePropagation` is used in the implementations of this method. Perhaps it would be nice to not pass in `StructurePropagation`, to limit it to basically `InternalTreeLogic`?

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNodeDynamicSize.java
```diff
@@ -562,6 +572,11 @@ private int middle( PageCursor leftCursor, int insertPos, KEY newKey, VALUE newV
         return middle;
     }
 
+    private int halfSpace()
+    {
+        return (pageSize - HEADER_LENGTH_DYNAMIC) / 2;
```
tinwelint: Looks like this could be calculated once and put as a field, although perhaps the compiler does this anyway...

community/index/src/main/java/org/neo4j/index/internal/gbptree/SeekCursor.java
```diff
@@ -513,7 +513,7 @@ private boolean internalNext() throws IOException
                 if ( verifyExpectedFirstAfterGoToNext )
                 {
                     pos = seekForward ? 0 : keyCount - 1;
-                    bTreeNode.keyAt( cursor,firstKeyInNode, pos, INTERNAL );
```
tinwelint: Hah, was thinking about the method name `internalNext()`, that clashes with LEAF,INTERNAL, doesn't it? Stumbled upon it when reading this change and thinking why the heck you'd want to read a LEAF key when you're doing a next for an INTERNAL node... before it struck me :)

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -424,7 +424,8 @@ public GBPTree( PageCache pageCache, File indexFile, Layout<KEY,VALUE> layout, i
             this.pagedFile = openOrCreate( pageCache, indexFile, tentativePageSize, layout );
             this.pageSize = pagedFile.pageSize();
             closed = false;
-            this.bTreeNode = new TreeNodeFixedSize<>( pageSize, layout );
+            this.bTreeNode = layout.fixedSize() ? new TreeNodeFixedSize<>( pageSize, layout )
```
tinwelint: I was thinking about selection of `TreeNode` and the version header. Nowhere will the version somehow contain which impl of `TreeNode` that the tree is using, will it? The selection will be on `Layout#fixedSize()` which theoretically can change between neo4j versions (perhaps not practically though). Is this something we should bring into the version header as well? Maybe mentioning selection of `TreeNode` in the version header will have future benefits too. What are your thoughts?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -1046,14 +1046,35 @@ void printTree() throws IOException
...
+    // Utility method
+    /**
+     * Print node with given id to System.out, if node with id exists.
+     * @param id the page id of node to print
+     */
+    void printNode( int id ) throws IOException
+    {
+        if ( id < freeList.lastId() )
+        {
+            // Use write lock to avoid adversary interference
```
tinwelint, line 1067: So grabbing a write cursor because some tests have adversarial cursors? The things that adversarial cursors do also happen in real life. Imho instead either change the comment or do this with a read cursor.

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNodeDynamicSize.java
```diff
@@ -0,0 +1,1262 @@
...
+
+    @Override
+    void defragmentInternal( PageCursor cursor )
+    {
+        doDefragment( cursor, INTERNAL );
+    }
+
+    private void doDefragment( PageCursor cursor, Type type )
+    {
+        /*
```
tinwelint, line 409: Another gorgeous doc/comment, awesome!

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNode.java
```diff
@@ -30,42 +30,22 @@
...
+    {
+        LEAF,
+        INTERNAL
+    }
+
+    enum Overflow
+    {
+        YES,
+        NO,
+        NEED_DEFRAG
```
tinwelint: I was thinking... `NEED_DEFRAG` means `YES`, but with a condition that it must be defragged first, right? Could it be called `YES_AFTER_DEFRAG` instead?

tinwelint (changes requested):
Great work. I generally agree with the structural changes of the code and looks like a nice approach. 

Some fairly small comments here and there for this first round of review.

EDIT: I see that some comments are outdated, due to commenting on individual commits, but perhaps some provide value anyway.

community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
burqen: The javadoc try to explain how the actual key look, not the offset array slot. Will try to make this more clear.

community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
burqen: Right! I'm a little bit unsure about where to put that though as `DynamicSizeUtil` is not responsible for where the stuff is placed relative to each other.

Hmm, putting it into `DynamicSizeUtil` for now anyway.

burqen: Put the explanation in class doc. Thoughts?

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: Yes, that part was forgotten / ignored for now. But we should definitely make TreeNode version part of complete version in some way!

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
burqen: The thinking was that `printNode`, just as `printTree` should never be used in production but is just a utility method for testing. If you still think we should use read cursor here anyway, I'm ok with that.

community/index/src/main/java/org/neo4j/index/internal/gbptree/SeekCursor.java
burqen: Haha, yeah that's an unfortunate clash

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNode.java
burqen: Aaah, yes! Nice catch

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNode.java
burqen: `NEED_DEFRAG` mean `NO_BUT_YOU_NEED_TO_DO_SOME_CLEANUP_FIRST`. They all answer the question "will I get an OVERFLOW if I try to insert this item?". Will make naming more clear.

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNodeDynamicSize.java
burqen: I think this was put in during development when there was actually bad keySizes around because of bad overwrites. And if bad keySizes are actually around we don't necessarily notice those (as long as they are non negative), because we are not necessarily out of bounds.

community/index/src/test/java/org/neo4j/index/internal/gbptree/SimpleByteArrayLayout.java
burqen: Yes, to create some "not everything align with 16", but now we vary size a bit more as you probably noticed :D (which actually chased out quite a few bugs)

burqen:
@tinwelint Comments addressed :D

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
tinwelint: Well OK, tree printing is a special little thing, sure. I'm OK with that.

tinwelint (approved):

PR #10749, opened 2018-01-08
Fix slot allocation of cartesian product to honor slot order

### LINKED ISSUE #10611: 3.3.1-enterprise - Returning results from relationships that don't exist
I posted this in slack. I ended up downgrading to 3.2.8 and everything works as expected. I don't have a good/short way to reproduce. I would be happy to set everything back up to 3.3.1 and show someone the error if it helped, but I mostly wanted to make this issue in case anyone else experiencing something similar there is a larger consensus.

My query - ran either direclty in the neo4j browser, or with the js driver.
```
rows = await _db.run(`MATCH (u:user {uuid: $uid})-[:allowed_to]->(c:company {uuid: $cid})
MATCH (f:file {uuid: $uuid})
MATCH (f)-[:uploaded_for]->(t) // get targets
OPTIONAL MATCH (tg:tag)-[:tags]->(f) // get tags
RETURN {
  uuid: f.uuid,
  targets: collect(DISTINCT t),
  tags: collect(DISTINCT tg)
}
`, {
uid: req.user.uid,
cid: req.user.cid,
uuid: fid,
})
```
What seems to be happening is that it's throwing things into `tags` that don't have the relationship of `tags`.
For example, a partial output of the results is (everything else is as expected):
```
...
"tags": [{
  "identity": {
  "low": 3132,
  "high": 0
},
"labels": [
  "note"
],
"properties": {
  "note": "What is this?",
  "updated_at": 1512661314571,
  "editable": false,
  "created_at": 1512661314571,
  "details": "Is this working right?",
  "label": "dooda",
  "category": "mycat",
  "uuid": "2bca4acb-0241-4a4a-ae37-ba34f92b3acd"
}}, {
"identity": {
  "low": 20,
  "high": 0
},
"labels": [
 "company"
],
"properties": {
  "name": "Allebrum",
  "uuid": "45a8404a-073a-49c3-8705-63142e31ff5d",
  "email": ""
}}]}
```
but if I look at the `tags` relationship, all I have is one relationship that has nothing to do with the ones specified: 
screen shot 2017-12-07 at 12 15 09 pm

As mentioned, during development we used 3.2.5. We are using docker, so the installs on the production server we clean installs. At one point I had upgraded from 3.3.0 to 3.3.1 to fix an issue we were having that was resolved in 3.3.1. Thinking this was a bad upgrade, we deleted everything and started again...the same result happened.

- Neo4j version: 3.3.1-enterprise
- Operating system: Neo4j Docker 3.3.1-enterprise
- API/Driver: Cypher and JS neo4j driver produced same results
- **Steps to reproduce**

- Expected behavior
A single result.
- Actual behavior
Two results, with unreleated nodes placed in tags array

fickludd:
Slot allocation for `CartesianProduct` was not maintaining the slot order from the rhs, which could lead to scrambled variable values upstream. This is already fixed in `3.4`, so simply forward-merging the test is enough.

Fixes #10611.

### DIFF
--- a/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/PipelineInformation.scala
+++ b/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/PipelineInformation.scala
@@ -181,8 +181,8 @@ class PipelineInformation(private var slots: Map[String, Slot],
     case _ => throw new InternalException("Uh oh... There was no slot for `$name`")
   }
 
-  def foreachSlot[U](f: ((String,Slot)) => U): Unit =
-    slots.foreach(f)
+  def foreachSlotOrdered[U](f: ((String, Slot)) => U): Unit =
+    slots.toSeq.sortBy(_._2)(SlotOrdering).foreach(f)
 
   def canEqual(other: Any): Boolean = other.isInstanceOf[PipelineInformation]
 
@@ -205,4 +205,15 @@ class PipelineInformation(private var slots: Map[String, Slot],
   private def checkNotAlreadyTaken(key: String) =
     if (slots.contains(key))
       throw new InternalException("Tried overwriting already taken variable name")
+
+  object SlotOrdering extends Ordering[Slot] {
+    def compare(x: Slot, y: Slot): Int = (x, y) match {
+      case (_: LongSlot, _: RefSlot) =>
+        -1
+      case (_: RefSlot, _: LongSlot) =>
+        1
+      case _ =>
+        x.offset - y.offset
+    }
+  }
 }
--- a/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/PipelineInformation.scala
+++ b/enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/PipelineInformation.scala
@@ -182,7 +182,7 @@ class PipelineInformation(private var slots: Map[String, Slot],
   }
 
   def foreachSlotOrdered[U](f: ((String, Slot)) => U): Unit =
-    slots.toSeq.sortBy(_._2)(SlotOrdering).foreach(f)
+    slots.toIndexedSeq.sortBy(_._2)(SlotOrdering).foreach(f)
 
   def canEqual(other: Any): Boolean = other.isInstanceOf[PipelineInformation]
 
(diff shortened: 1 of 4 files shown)


systay:
FYI - It's preferable to have the "Fixes..." comments on the commit and not the PR description. By doing that, the issue is automatically closed once the commit is merged, and we can find it easy by looking at `git log`

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/PipelineInformation.scala
```diff
@@ -181,8 +181,8 @@ class PipelineInformation(private var slots: Map[String, Slot],
     case _ => throw new InternalException("Uh oh... There was no slot for `$name`")
   }
 
-  def foreachSlot[U](f: ((String,Slot)) => U): Unit =
-    slots.foreach(f)
+  def foreachSlotOrdered[U](f: ((String, Slot)) => U): Unit =
+    slots.toSeq.sortBy(_._2)(SlotOrdering).foreach(f)
```
pontusmelke: I think you should use `toIndexedSeq` or `toVector` here, sorting a `scala.List` is really slow.

enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/PipelineInformation.scala
```diff
@@ -205,4 +205,15 @@ class PipelineInformation(private var slots: Map[String, Slot],
   private def checkNotAlreadyTaken(key: String) =
     if (slots.contains(key))
       throw new InternalException("Tried overwriting already taken variable name")
+
+  object SlotOrdering extends Ordering[Slot] {
```
pontusmelke: Maybe a comment why this slot ordering makes sense?

enterprise/cypher/physical-planning/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/SlotAllocationTest.scala
```diff
@@ -28,6 +28,7 @@ import org.neo4j.cypher.internal.ir.v3_3.{CardinalityEstimation, IdName, Planner
 import org.neo4j.cypher.internal.v3_3.logical.plans.{Ascending, _}
 import org.neo4j.cypher.internal.v3_3.logical.{plans => logicalPlans}
 
+//noinspection NameBooleanParameters
```
pontusmelke, line 31: Can't we name the boolean parameters instead?

pontusmelke (changes requested):
Nice, just some tiny tiny things

enterprise/cypher/physical-planning/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/SlotAllocationTest.scala
fickludd: We can, but I think it makes for less readable code in this case.

pontusmelke (approved):
👍

enterprise/cypher/physical-planning/src/test/scala/org/neo4j/cypher/internal/compatibility/v3_3/runtime/SlotAllocationTest.scala
pontusmelke: ok, you win

PR #10751, opened 2018-01-08
Unify pruning var implementations

fickludd:
This change is already done in `3.3` (#10666), but it was only null-merged into `3.4` because the forward merge had some unexpected complications. After this PR there is only one pruning var implementation in production.


pontusmelke (approved):
This looks as good as it did in 3.3, however needs a rebase and there seem to be one test failure

pontusmelke:
@fickludd sorry but needs more rebasing

fickludd:
@pontusmelke puh, easy rebase luckily.

PR #10753, opened 2018-01-09
Report tool now check free disk space before starting

klaren:
Reports can be quite large if both transaction files and heap dump is included.

This PR makes sure that the available disk space is large enough before starting to generate a report to avoid running out of disk space. The estimation of the final size is a pure guess and pessimistically chosen.

It also introduces a `--force` argument to the tool that can be used to ignore the warning about the free disk space.

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/dbms/diagnostics/jmx/JmxDump.java
+++ b/community/dbms/src/main/java/org/neo4j/dbms/diagnostics/jmx/JmxDump.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
 import java.lang.management.ManagementFactory;
+import java.lang.management.MemoryMXBean;
 import java.lang.management.MonitorInfo;
 import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
@@ -44,7 +45,7 @@ import javax.management.remote.JMXServiceURL;
 
 import org.neo4j.diagnostics.DiagnosticsReportSource;
 import org.neo4j.diagnostics.DiagnosticsReportSources;
-import org.neo4j.diagnostics.DiagnosticsReporterProgressCallback;
+import org.neo4j.diagnostics.DiagnosticsReporterProgressInteractions;
 import org.neo4j.diagnostics.ProgressAwareInputStream;
 
 /**
@@ -111,8 +112,7 @@ public class JmxDump
         ThreadMXBean threadMxBean;
         try
         {
-            threadMxBean =
-                    ManagementFactory.newPlatformMXBeanProxy( mBeanServer, "java.lang:type=Threading", ThreadMXBean.class );
+            threadMxBean = ManagementFactory.getPlatformMXBean( mBeanServer, ThreadMXBean.class );
         }
         catch ( IOException e )
         {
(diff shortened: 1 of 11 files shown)


community/dbms/src/main/java/org/neo4j/dbms/diagnostics/jmx/JmxDump.java
```diff
@@ -181,26 +181,44 @@ public String destinationPath()
...
+                {
+                    MemoryMXBean bean = ManagementFactory.getPlatformMXBean( mBeanServer, MemoryMXBean.class );
+                    long totalMemory = bean.getHeapMemoryUsage().getCommitted() + bean.getNonHeapMemoryUsage().getCommitted();
+
+                    // We first write raw to disk then write to archive, 5x compression is a reasonable worst case estimation
+                    return (long) (totalMemory * 1.2);
+                }
+                catch ( IOException e )
+                {
+                    progress.error( "Unable to determine size of neo4j instance memory", e );
```
MishaDemianenko: note: throw exception instead

community/kernel/src/main/java/org/neo4j/diagnostics/DiagnosticsReporter.java
```diff
@@ -75,7 +75,23 @@ public void dump( Set<String> classifiers, Path destination, DiagnosticsReporter
...
+        // Estimate an upper bound of the final size and make sure it will fit
+        long estimatedFinalSize = sources.stream().mapToLong(
+                diagnosticsReportSource -> diagnosticsReportSource.estimatedSize( progress ) ).sum();
+        long freeSpace = destinationFolder.toFile().getFreeSpace();
+        if ( estimatedFinalSize > freeSpace )
+        {
+            String message =
+                    String.format( "WARNING: Free available disk space for %s is %s, worst case estimate is %s",
+                            destination.getFileName(), Format.bytes( freeSpace ), Format.bytes( estimatedFinalSize ) );
+            if ( !progress.shouldIgnorePotentialFullDisk( message ) )
```
MishaDemianenko: let's change this condition from `!shouldIgnore` to `continueOnFull` to remove negation

community/kernel/src/main/java/org/neo4j/diagnostics/DiagnosticsReporter.java
```diff
@@ -75,7 +75,23 @@ public void dump( Set<String> classifiers, Path destination, DiagnosticsReporter
         }
 
         // Make sure target directory exists
-        Files.createDirectories( destination.getParent() );
+        Path destinationFolder = destination.getParent();
+        Files.createDirectories( destinationFolder );
+
+        // Estimate an upper bound of the final size and make sure it will fit
+        long estimatedFinalSize = sources.stream().mapToLong(
```
MishaDemianenko: lets extract this space check into separate helper method instead

community/kernel/src/main/java/org/neo4j/diagnostics/DiagnosticsReporterProgressInteractions.java
```diff
@@ -48,6 +48,14 @@
      */
     void error( String msg, Throwable throwable );
 
+    /**
+     * A check whether to ignore the free available disk space when generating the archive.
+     *
+     * @param message with information about the disk space.
+     * @return true if the execution should continue even if risking to fill up the disk.
+     */
+    boolean shouldIgnorePotentialFullDisk( String message );
```
MishaDemianenko: it feels a bit wrong that its a progress reporter who decide should or shouldn't we ignore potential OOD problems.

community/kernel/src/main/java/org/neo4j/diagnostics/InteractiveProgress.java
```diff
@@ -102,6 +104,33 @@ public void error( String msg, Throwable throwable )
...
+    {
+        if ( force )
+        {
+            return true;
+        }
+        out.println( message );
+        while ( true )
+        {
+            out.print( "Ignore available disk space warning? [yN]: " );
+            String answer = System.console().readLine();
```
MishaDemianenko: `answer` can be null just use `StringUtils.isEmpty` instead

MishaDemianenko (changes requested):
Nice in general, couple of comments here and there

community/kernel/src/main/java/org/neo4j/diagnostics/DiagnosticsReporter.java
klaren: Still a negation though :sweat_smile: 
Choosing `stopOnPotentialFullDisk`

klaren:
@MishaDemianenko Addressed comments, please have another look :)

MishaDemianenko (approved):

PR #10754, opened 2018-01-09
Guarantee release cleanerLock in GBPTree

burqen:
In CleanupJob, processing the job and releasing the cleaner lock
is separated into run and close such that a RecoveryCleanupWorkCollector
can close a CleanupJob and thus release the cleaner lock without
actually running the cleanup job. This makes it possible to open a dirty
GBPTree and write a new header without starting cleanup jobs and
not risking a deadlock between checkpoint (needed for writing header)
and cleanup job.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/CleanupJob.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/CleanupJob.java
@@ -43,7 +43,7 @@ public interface CleanupJob extends Runnable
     Exception getCause();
 
     /**
-     * Mark this job and cleanup all it's resources.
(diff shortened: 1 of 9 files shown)


community/index/src/main/java/org/neo4j/index/internal/gbptree/CleanupJob.java
```diff
@@ -42,6 +42,11 @@
      */
     Exception getCause();
 
+    /**
+     * Mark this job and cleanup all it's resources.
```
tinwelint: `Mark this job....` as closed perhaps? No biggie, I do believe that anyone reading that kind of gets what it wants to convey though

tinwelint (approved):

PR #10756, opened 2018-01-09
Better error message on lock acquisition failure

jakewins:
We may want to explore swapping to use TransactionTerminated in 3.4, unless there is some reason I'm not seeing why a user would want to differentiate between this error and TransactionTerminated. For 3.2 and 3.3 it seems better to simply improve the error message to limit API changes.

### DIFF
--- a/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
+++ b/community/common/src/main/java/org/neo4j/kernel/api/exceptions/Status.java
@@ -184,7 +184,7 @@ public interface Status
                 "transaction was active. Transaction may succeed if retried." ),
         LockClientStopped( TransientError,
                 "The transaction has been terminated, so no more locks can be acquired. This can occur because the " +
-                "transaction ran longer than the configured transaction timeout, or because a human operator manually " + 
+                "transaction ran longer than the configured transaction timeout, or because a human operator manually " +
                 "terminated the transaction, or because the database is shutting down."),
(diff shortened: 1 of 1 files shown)


lutovich (commented):
I like the new error message but do not think we should fix it in 3.2. Imho it is better to improve message of `TransactionTerminated` in 3.4 and use it everywhere instead of `LockClientStopped`. `Status.Transaction.Terminated` should probably also be removed.

MishaDemianenko:
Praise to human operators, but I agree with @lutovich what is the reason to change it back in 3.2? that does not look like the critical thing to fix, so why not update that only in the latest version?

jakewins:
Agree on @lutovich's suggested 3.4 change; but what's the drawback to improving the message in 3.2? The benefit would be that people on maintenance versions hitting this problem understand the error, which they currently are struggling with. 

Eg. merging this minimal change will remove one little pathway to support questions, and introduces no risk, and then in 3.4 we do what @lutovich says, or is there an argument for why merging this could cause problems?

MishaDemianenko:
Or their handling of exception with the specific message will fail on next minor version release.
Maintenance versions should contain as minimal changes as possible to minimise any for foreseeable or non-foreseeable impact.

jakewins:
Hmm, the intention of Status was that the API compatibility lived on the status code type, but I'll agree to disagree, this is your call :P Closing this then.

PR #10759, opened 2018-01-09
Read Replicas updates can be started and stopped via a procedure.

jimwebber:
Use case: user wants to do large analytical job on some read replica
and doesn't want updates while the job us running.

CALL dbms.cluster.freezeReadReplica()
pauses catchup

CALL dbms.cluster.unfreezeReadReplica()
resumes catchup

This only works on read replicas.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/EnterpriseReadReplicaEditionModule.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/readreplica/EnterpriseReadReplicaEditionModule.java
@@ -121,10 +121,14 @@ import static org.neo4j.causalclustering.discovery.ResolutionResolverFactory.cho
  */
 public class EnterpriseReadReplicaEditionModule extends EditionModule
 {
+
+    private final CatchupPollingProcess catchupPollingProcess;
+    private final LogService logService;
+
     EnterpriseReadReplicaEditionModule( final PlatformModule platformModule,
                                         final DiscoveryServiceFactory discoveryServiceFactory, MemberId myself )
     {
-        LogService logging = platformModule.logging;
+        logService = platformModule.logging;
 
         ioLimiter = new ConfigurableIOLimiter( platformModule.config );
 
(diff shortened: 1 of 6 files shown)


jimwebber:
I'd like some feedback on whether the freeze/unfreeze procedures need to be secured.

davidegrohmann:
@jimwebber how would this interact with bolt+routing and bookmarks? I believe you could still get queries routed to that machine and some might get stuck by waiting forever on the frozen replicate for a bookmark.

 Also note that the integration with LDAP might still interfere with your long running analytic query.

I think a better approach to your use case would be to remove the replica completely from the cluster, in particular, from the membership set.  Basically to use the read-replica as a single instance for some time before readding it to the cluster.

chrisvest:
Could unfreezing happen automatically when the transaction commits? I'm thinking as if freeze takes a special shared lock (to allow multiple concurrent analytical transactions to do this) that coordinates with an exclusive lock that is taken for catchup.

martinfurmanski:
@jimwebber @davidegrohmann I guess you could use the server_groups to create something functional even in a routing environment, but it would not be a particularly well-polished design overall. Unless there is a strong motivator behind this work, I would suggest that we perhaps relegate this back to a future design discussion?

@davidegrohmann Can you expand on the LDAP-integration issue? You mean the analytical query could be aborted midways for security reasons? Could this be connectivity issues alone or does it have to be updates to security policies?

davidegrohmann:
@martinfurmanski it is not an issue by per design when security changes happen, you should be aware that it might cause your query to abort. How to work improve there should be discussed with the security team.

jerrinot:
what if a client dies before it has a chance to resume the catch-up process?

jimwebber:
@jerrinot Not sure what you mean. The "client" in this case is a read replica. If it dies, it can be recovered. Post-recovery the operator can unfreeze the read replica again.

jerrinot:
@jimwebber: thanks! This is the part I was missing: _"The "client" in this case is a read replica." _

I thought an arbitrary client could trigger the freeze, send a bunch of analytics commands and then unfreeze.

jimwebber:
@jerrinot ah, I see. So many things that play the "client" role!

It could also be a "user" client, but that code does not - yet - exist. But we have some options to address it:

1. Do nothing. Frozen replicas have log entries, leave it to the operator.
2. Have freezing associated with a transaction and unfreeze when transaction completes.
3. Report frozen/unfrozen analytics instances along with the reader/writer/router metadata to the drivers.
4. Probably others.

I'll have real-time chat with @martinfurmanski and figure out what might be best.

jerrinot:
@jimwebber: once again thanks for the insight, it's much appreciated!

Dcanzano:
@jimwebber   i presume when authentication is enabled the only user who will be able to call the stored procedure is a user with ADMIN role?  or will an ARCHITECT also be allowed?

jimwebber:
@Dcanzano same permissions as other built in procedures.

PR #10761, opened 2018-01-10
Allow changing types of properties

pontusmelke:
Before applying a property update to the transaction state we do a check
to see if the value has actually been changed. However since by our
equality semantics a change of type not necessarily implies a change of
value, e.g. `intValue == tofloat(intvalue)` evaluates to `true`
(at least for integers small enough to fit the mantissa of the
float), we run into problems when users want to force a type change via
for example,

```
MATCH (n) SET n.prop = tofloat(n.prop)
```

in Cypher or via

```
n.setProperty( "prop", (Double) n.getProperty("prop") );
```

in Core API.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
@@ -969,10 +969,7 @@ public class StateHandlingStatementOperations implements
             // This because there are established ways of rebuilding auto-indexes which involves this function.
             autoIndexProperty( nodeId, property, ops, existingProperty, autoIndexing.nodes() );
 
-            //It is not enough to check equality here since by our equality semantics `int == tofloat(int)` is `true`
-            //so by only checking for equality users cannot change type of property without also "changing" the value.
-            //Hence the extra type check here.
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
```diff
@@ -969,7 +969,10 @@ public Property nodeSetProperty( KernelStatement state, long nodeId, DefinedProp
             // This because there are established ways of rebuilding auto-indexes which involves this function.
             autoIndexProperty( nodeId, property, ops, existingProperty, autoIndexing.nodes() );
 
-            if ( !property.equals( existingProperty ) )
+            //It is not enough to check equality here since by our equality semantics `int == tofloat(int)` is `true`
+            //so by only checking for equality users cannot change type of property without also "changing" the value.
+            //Hence the extra type check here.
+            if ( property.getClass() != existingProperty.getClass() || !property.equals( existingProperty ) )
```
MishaDemianenko: instead of copypasting comment and update check in 2 places please extract helper method instead

MishaDemianenko (changes requested):
One small thing from me

pontusmelke:
@MishaDemianenko fixed

MishaDemianenko (approved):

PR #10762, opened 2018-01-10
Remove deprecation annotation for legacy ssl settings

martinfurmanski:
The annotation is used by the configuration validator to issue
warnings, leading to unnecessary warnings, since the setting
is still valid though not preferred. This was an unintended
side-effect and most easily resolved by not using the deprecation
annotation here.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
@@ -31,27 +31,24 @@ import static org.neo4j.kernel.configuration.Settings.derivedSetting;
 import static org.neo4j.kernel.configuration.Settings.pathSetting;
 
 /**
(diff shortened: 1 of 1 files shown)


phughk:
Even if it is giving unnecessary warnings, the fact that these settings are deprecated signify that they will be removed in future versions (which they will), yes?

martinfurmanski:
@phughk Yes, removed breakingly in 4.0, and users will have to migrate to the new settings.

andrewkerr9000:
Surely if a setting is not preferred and will be removed in a future version then a deprecation warning is entirely appropriate?

martinfurmanski:
@andrewkerr9000 I guess the main problem is that the deprecated setting is still shipped as the default. There is really nothing special most users have to do because of this, so a warning is unnecessary. I did not know that the Deprecation annotation would induce a warning, as it was merely an internal code annotation for me

I think it would be better if the configuration framework had its own set of annotations instead of trying to overload the usage of standard ones, e.g. @WarnOnUsage, but that is a different matter.

PR #10766, opened 2018-01-10
Refactor native index

OliviaYtterbrink:
Separate generic parts to prepare for String and Spatial indexing.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexReader.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexReader.java
@@ -40,15 +40,15 @@ import org.neo4j.storageengine.api.schema.IndexReader;
 import org.neo4j.storageengine.api.schema.IndexSampler;
 import org.neo4j.values.storable.Value;
 
-public abstract class NativeSchemaIndexReader<KEY extends NativeSchemaKey, VALUE extends NativeSchemaValue>
+abstract class NativeSchemaIndexReader<KEY extends NativeSchemaKey, VALUE extends NativeSchemaValue>
         implements IndexReader
 {
-    protected final GBPTree<KEY,VALUE> tree;
-    protected final Layout<KEY,VALUE> layout;
-    protected final IndexSamplingConfig samplingConfig;
+    private final GBPTree<KEY,VALUE> tree;
+    private final Layout<KEY,VALUE> layout;
+    private final IndexSamplingConfig samplingConfig;
 
     private final Set<RawCursor<Hit<KEY,VALUE>,IOException>> openSeekers;
-    protected final IndexDescriptor descriptor;
+    private final IndexDescriptor descriptor;
 
     NativeSchemaIndexReader( GBPTree<KEY,VALUE> tree, Layout<KEY,VALUE> layout,
             IndexSamplingConfig samplingConfig,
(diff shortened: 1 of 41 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexReader.java
```diff
@@ -123,96 +115,12 @@ public PrimitiveLongResourceIterator query( IndexQuery... predicates ) throws In
...
-        {
-            treeKeyFrom.initAsLowest();
-        }
-        else
-        {
-            treeKeyFrom.from( rangePredicate.fromInclusive() ? Long.MIN_VALUE : Long.MAX_VALUE, fromValue );
-            treeKeyFrom.entityIdIsSpecialTieBreaker = true;
-        }
-    }
+    public abstract void query( IndexProgressor.NodeValueClient cursor, IndexOrder indexOrder, IndexQuery... predicates );
```
burqen: Instead of extracting all of `query`, could we do something like this:

```
query( ... )
{
    validateQuery( ... );
    KEY treeKeyFrom = layout.newKey();
    KEY treeKeyTo = layout.newKey();
    initializeRangeForQuery( treeKeyFrom, treeKeyTo, predicate );
    startSeekForInitializedRange( cursor, treeKeyFrom, treeKeyTo, predicates );
}

abstract validateQuery( ... );

abstract initializeKeysForQuery( treeKeyFrom, treeKeyTo, predicate );
```

So that we keep the shared logic of what is called when but and only extract the "how".

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NativeSchemaKey.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.neo4j.index.internal.gbptree.GBPTree;
+import org.neo4j.values.storable.Value;
+
+/**
+ * Includes value and entity id (to be able to handle non-unique values).
+ * A value can be any {@link Number} and is represented as a {@code long} to store the raw bits and a type
+ * to say if it's a long, double or float.
+ */
+public interface NativeSchemaKey
```
burqen: This javadoc is now pretty outdated 😅  Maybe something along those lines:

This is the abstraction of what NativeSchemaIndex with friends need from a schema key.
Note that is says nothing about how keys are compared, serialized, read, written, etc. This is the job of Layout.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberSchemaKey.java
```diff
@@ -111,20 +132,20 @@ void initAsHighest()
...
 
     /**
-     * Extracts raw bits and type from a {@link NumberValue} and store as state of this {@link SchemaNumberKey} instance.
+     * Extracts raw bits and type from a {@link Value} and store as state of this {@link NativeSchemaKey} instance.
      *
-     * @param value actual {@link NumberValue} value.
+     * @param value actual {@link Value} value.
      */
-    private void extractRawBitsAndType( NumberValue value )
+    private void extractRawBitsAndType( Value value )
```
burqen: This little fellow can take `NumberValue` again and remember the javadoc 😄

burqen (changes requested):
This is great! Found a few things that could be improved, see comments.

Also, there are quite a few methods and classes that have become either public or protected. I think that most of them can be package-private but I'm not sure what visibility you actually need for spatial so I'll leave it be and we can clean that up later.

burqen (approved):

burqen:
@OliviaYtterbrink Excellent! Will just wait for the tests to go green and then I'll merge.

PR #10770, opened 2018-01-10
Fix online backup problems.

chrisvest:
* Fixes a service loading problem on Java9: To be able to service load class implementations across implicit module boundaries, the classes involved must be public.
* Fixes a directory file descriptor leak in `BackupProtocolServiceIT`, that caused a test to fail on Windows.


burqen (approved):
Looking good!

PR #10771, opened 2018-01-10
Fix that MERGE did not enforce relationship uniqueness.

sherfert:
Add predicates to MERGE clauses, so that our addUniquenessPredicates rewriter is able to
add a uniqueness predicate.


community/cypher/cypher-compiler-3.1/src/test/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/rewriters/NameMatchPatternElementTest.scala
```diff
@@ -30,23 +30,23 @@ class NameMatchPatternElementTest extends CypherFunSuite {
     val original = parser.parse("MATCH (n)-[r:Foo]->() RETURN n")
     val expected = parser.parse("MATCH (n)-[r:Foo]->(`  UNNAMED20`) RETURN n")
 
-    val result = original.rewrite(nameMatchPatternElements)
+    val result = original.rewrite(nameMatchAndMergePatternElements)
```
craigtaverner: This test class could also test MERGE clauses

community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MergeNodeAcceptanceTest.scala
```diff
@@ -30,4 +30,10 @@ class MergeNodeAcceptanceTest extends ExecutionEngineFunSuite with QueryStatisti
       assertStats(result, nodesCreated = 1, propertiesWritten = 1, labelsAdded = 1)
     }
   }
+
+  test("Merging with self loop and relationship uniqueness") {
+    graph.execute("CREATE (a) CREATE (a)-[:X]->(a)")
+    val result = updateWithBothPlannersAndCompatibilityMode("MERGE (a)-[:X]->(b)-[:X]->(c) RETURN 42")
+    assertStats(result, relationshipsCreated = 2, nodesCreated = 3)
+  }
```
craigtaverner, line 38: Since the fix seems relevant to all predicates, perhaps we should have more tests:

* Test for relationship uniqueness without self-relationships
* Test for other predicates than relationship uniqueness

craigtaverner (changes requested):

community/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MergeNodeAcceptanceTest.scala
craigtaverner: I've notice that the second case is already covered by the fact that Cypher does not allow explicit WHERE clauses for MERGE, and the implicit WHERE from in-lined Label and Property checks was already checked for separately, so the only predicate not covered was relationship uniqueness, which is good. I the meantime I have written a test for the first case.

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ast/rewriters/nameMatchAndMergePatternElements.scala
```diff
@@ -22,14 +22,17 @@ package org.neo4j.cypher.internal.compiler.v3_1.ast.rewriters
 import org.neo4j.cypher.internal.frontend.v3_1.ast._
 import org.neo4j.cypher.internal.frontend.v3_1.{Rewriter, bottomUp}
 
-case object nameMatchPatternElements extends Rewriter {
+case object nameMatchAndMergePatternElements extends Rewriter {
```
craigtaverner: There is already a nameUpdatingClause Rewriter that does both Match and Merge. I think we can get rid of this one entirely.

community/cypher/cypher-compiler-3.1/src/main/scala/org/neo4j/cypher/internal/compiler/v3_1/ASTRewriter.scala
```diff
@@ -47,7 +47,7 @@ class ASTRewriter(rewriterSequencer: (String) => RewriterStepSequencer, shouldEx
       enableCondition(containsNoReturnAll),
       foldConstants,
       ApplyRewriter("extractParameters", extractParameters),
-      nameMatchPatternElements,
+      nameMatchAndMergePatternElements,
```
craigtaverner: There is already a nameUpdatingClauses Rewriter. That could be used here instead of changing nameMatchPatternElements

craigtaverner (approved):

PR #10772, opened 2018-01-10
Do not rebind node cursor on property set to the same node twice.

MishaDemianenko:
Do not rebind node cursor to the same node on property read.
Introduce helper methods to bind cursors to single node.
Use getId directly instead of nodeReference.
Use 'Long.compare' instead of home made analog.


MishaDemianenko:
@pontusmelke can you please take a look, thx!

tinwelint (approved):
Looks like a nice cleanup to me

PR #10773, opened 2018-01-10
Fix the periodic checkpoint policy scheduling frequency.

chrisvest:
This fixes a mistake that snuck in with the volumetric check pointing
policy.

In order to have a somewhat accurate check pointing frequency based on
the transaction counts, we need to check those counts pretty regularly.
However, the scheduling frequency for the time based check point
threshold was changed to match that of the checkpoint interval, while
the frequency for the transaction count was changed to be once every
day. The check frequency is now the lesser of the two instead of being
hard coded to 10 seconds, which means that the transaction check
frequency would end up being dictated by the time based threshold.
Effectively rendering the transaction count based threshold useless.

To fix this, the transaction count threshold has changed its checking
frequency to 10 seconds, so _it_ now dictates the checking frequency,
and thus we will notice with much greater timeliness if we should do
a check point based on the transaction counts.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CountCommittedTransactionThreshold.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/checkpoint/CountCommittedTransactionThreshold.java
@@ -19,8 +19,6 @@
  */
 package org.neo4j.kernel.impl.transaction.log.checkpoint;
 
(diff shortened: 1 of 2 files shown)


MishaDemianenko (approved):
Be careful since it looks like you want this change in 3.2 but for now you target 3.4

chrisvest:
Well spotted. Changed the target without problem.

PR #10780, opened 2018-01-11
Porting of space filling curves from "Neo4j Spatial"

OliviaYtterbrink:
This is a separate module so that it can be used for future versions of the Neo4j Spatial library from which it was derived, so we don't maintain two copies.
The included Envelope and EnvelopTest are trimmed down from the Neo4j Spatial versions, so we might want to refactor that later, either include all, or have Neo4j Spatial extend this one.

### DIFF
--- a/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/Envelope.java
+++ b/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/Envelope.java
@@ -23,6 +23,8 @@ import java.util.Arrays;
 
 public class Envelope
 {
+    private double[] min;
+    private double[] max;
 
     /**
      * Copy constructor
@@ -103,7 +105,7 @@ public class Envelope
 
     public int getDimension()
     {
-        return isValid() ? min.length : 0;
+        return min.length;
     }
 
     /**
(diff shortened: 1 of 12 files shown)


community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/Envelope.java
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    public boolean intersects( Envelope other )
+    {
+        if ( isValid() && other.isValid() && getDimension() == other.getDimension() )
+        {
+            boolean result = true;
+            for ( int i = 0; i < min.length; i++ )
+            {
+                result = result && other.min[i] <= max[i] && other.max[i] >= min[i];
```
sherfert: I see we have a shortcut in `covers`. If `ans` becomes false, we return it immediately. Can't we do the same here?

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/Envelope.java
```diff
@@ -0,0 +1,389 @@
...
+            return result ? new Envelope( i_min, i_max ) : null;
+        }
+        else
+        {
+            throw new IllegalArgumentException(
+                    "Cannot calculate intersection of Envelopes with different dimensions: " + this.getDimension() + " != " + other.getDimension() );
+        }
+    }
+
+    // Attributes
```
sherfert: This seems to be the only class (of this PR) that has Attributes at the bottom. I don't like that personally, but we should try to make it consistent at least.

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/Envelope.java
```diff
@@ -0,0 +1,389 @@
...
+        if ( !isValid() )
+        {
+            min = other.min.clone();
+            max = other.max.clone();
+        }
+        else
+        {
+            for ( int i = 0; i < min.length; i++ )
+            {
+                if ( other.min[i] < min[i] )
```
sherfert, line 150: I think this will hit an ArrayIndexOutOfBoundsException if `other.dimension != this.dimension`. In other methods, this is checked beforehand. Any reason for not doing it here?

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/Envelope.java
```diff
@@ -0,0 +1,389 @@
...
+    /**
+     * General constructor for the n-dimensional case
+     * @param min
+     * @param max
+     */
+    public Envelope( double[] min, double[] max )
+    {
+        this.min = min.clone();
+        this.max = max.clone();
+        if ( !isValid() )
```
sherfert, line 46: `isValid` should be final, since it is called from a Constructor.

As we're doing this here and throwing an Exception if not valid, we will never have invalid Envelopes. This means we could removed the isValid checks in all the other methods. There are only two mutating methods, `expandToInclude` and `translate` and they both seem safe w.r.t `isValid`.

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
```diff
@@ -0,0 +1,483 @@
...
+        final int[] npointValues;
+
+        CurveRule( int dimension, int[] npointValues )
+        {
+            this.dimension = dimension;
+            this.npointValues = npointValues;
+            assert npointValues.length == length();
+        }
+
+        int length()
```
sherfert: Should be final.

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/HilbertSpaceFillingCurve3D.java
```diff
@@ -0,0 +1,248 @@
...
+
+        private char direction( int start, int end )
+        {
+            end -= start;
+            switch ( end )
+            {
+            case 1:
+                return 'F'; // move up      000->001
+            case 2:
+                return 'U'; // move right   000->010
```
sherfert: This and the comment above look wrong. Front and up?

sherfert (changes requested):
I have just minor comments, but still some that are worth going over. For the main part of the logic I will probably need to do another pass, since it is very complex.

sherfert (approved):

PR #10782, opened 2018-01-11
Rename setting dbms.logs.timezone to dbms.db.timezone

klaren:
The setting `dbms.logs.timezone` is renamed to `dbms.db.timezone` and the setting now affects results returned by the `dbms.listQueries` and `dbms.listTransactions` procedures as well.

The setting  `dbms.logs.timezone` will continue to work in 3.4 and be synonymous with `dbms.db.timezone`, before being removed in 4.0

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -366,7 +366,13 @@ public class GraphDatabaseSettings implements LoadableConfig
     public static final Setting<Level> store_internal_log_level = setting( "dbms.logs.debug.level",
             options( Level.class ), "INFO" );
 
+    @Description( "Database timezone." )
+    public static final Setting<LogTimeZone> db_timezone =
+            setting( "dbms.db.timezone", options( LogTimeZone.class ), LogTimeZone.UTC.name() );
+
     @Description( "Database logs timezone." )
+    @Deprecated
+    @ReplacedBy( "dbms.db.timezone" )
     public static final Setting<LogTimeZone> log_timezone =
             setting( "dbms.logs.timezone", options( LogTimeZone.class ), LogTimeZone.UTC.name() );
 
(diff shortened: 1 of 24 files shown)


community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -366,7 +366,13 @@
     public static final Setting<Level> store_internal_log_level = setting( "dbms.logs.debug.level",
             options( Level.class ), "INFO" );
 
+    @Description( "Database timezone." )
+    public static final Setting<LogTimeZone> db_timezone =
+            setting( "dbms.db.timezone", options( LogTimeZone.class ), LogTimeZone.UTC.name() );
+
     @Description( "Database logs timezone." )
+    @Deprecated
+    @ReplacedBy( "dbms.db.timezone" )
```
burqen, line 375: So folk who are using this setting write now will see their logs with a different time zone after migration until updating to new setting?
Are we allowed to break that?

Edit: But we migrate the setting, so that should be fine 👍

burqen (changes requested):
Looks good 👍  got two thoughts.

1. Would be sweet with a similar test for `QueryStatusResult` as you added for `TransactionStatusResult`.

2. Can we please have a more descriptive changelog text? It should mention deprecation of `dbms.log.timezone`, automatic migration and timezone now working for procedures `listTransactions` and `listQueries`.

klaren:
1. That would including writing a test for all of `QueryStatusResult` since it's currently missing.
2. It's not automatic migration, it's runtime migration with a warning in the startup log, the config is not modified.

klaren:
@burqen Updated description.

burqen:
@klaren 
1. Yeah, that would ofc be nice, but not as part of this PR.
2. I like this description!
3. Are we allowed to "break" this behaviour?

klaren:
@burqen 
3. Ideally not, but `dbms.logs.timezone` was introduced last patch release and therefore has minimal spread. And the ones we know use it requested this behavior.

burqen (approved):

PR #10787, opened 2018-01-11
Allow missing JAVA_HOME

klaren:


community/dbms/src/main/java/org/neo4j/dbms/diagnostics/jmx/JMXDumper.java
```diff
@@ -71,9 +71,21 @@ public JMXDumper( Path homeDir, FileSystemAbstraction fs, PrintStream out, Print
...
+                }
+                else if ( e instanceof java.lang.NoClassDefFoundError ) // Attach API is not available
+                {
+                    printError( "Unable to attach to process. Reason: JDK is not available, please point " +
+                            "environment variable JAVA_HOME to a valid JDK location.", e);
+                }
+                else
+                {
+                    throw new RuntimeException( e );
+                }
```
benbc: Why `instanceof` instead of multiple catch blocks?

klaren: Compiler did not allow me to put `ClassNotFoundException` the citing dead code. But after I switched that to `ClassDefNotfound` that works just fine. Nice catch!

PR #10788, opened 2018-01-11
Log constraint index creation to find rare race condition

ragadeeshu:


community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/ConstraintIndexCreator.java
```diff
@@ -121,15 +137,18 @@ public long createUniquenessConstraintIndex(
             // At this point the integrity of the constraint to be created was checked
             // while holding the lock and the index rule backing the soon-to-be-created constraint
             // has been created. Now it's just the population left, which can take a long time
+            monitor.event( Thread.currentThread().getId(), Monitor.RELEASED_LOCK );
```
tinwelint: The log could be in unexpected order if this event is registered _before_ actually releasing the label lock, couldn't it?

tinwelint (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/ConstraintIndexCreator.java
ragadeeshu: @tinwelint Maybe, yes, but my thinking was that the opposite could potentially cause another thread to report grabbing the lock before this thread reports releasing the lock. Thoughts?

tinwelint: An event before and another, different one, after?

tinwelint (approved):

PR #10793, opened 2018-01-12
Remove not used TransactionTerminated state in favor of `Status.Transaction.Terminated`

MishaDemianenko:


klaren (approved):
Nice canonicalization!

PR #10795, opened 2018-01-12
Avoid producing tons of state maps for single node cursor

MishaDemianenko:
Remember state of already defined nodes only for case of nodes iteration
and do not do that if cursor created to go over single node.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
@@ -34,6 +34,7 @@ import org.neo4j.kernel.impl.store.RecordCursor;
 import org.neo4j.kernel.impl.store.record.DynamicRecord;
 import org.neo4j.kernel.impl.store.record.NodeRecord;
 
+import static java.util.Collections.emptySet;
 import static org.neo4j.kernel.impl.newapi.References.setDirectFlag;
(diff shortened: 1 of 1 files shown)


MishaDemianenko:
@tinwelint @pontusmelke guys, can you please take a look, it resolves super annoying slowness problem

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
```diff
@@ -91,7 +94,8 @@ void single( long reference, Read read )
         this.highMark = NO_ID;
         this.read = read;
         this.hasChanges = HasChanges.MAYBE;
-        this.addedNodes = null;
+        this.addedNodes = emptySet();
+        this.single = true;
```
pontusmelke: I think you can use `this.highMark == NO_ID` to check for `single` without introducing a new field. There is already a method `isSingle` method you can use instead.

tinwelint (approved):
Makes it slightly more complicated, but nice improvement imho

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeCursor.java
MishaDemianenko: Thx! was blind or something. Switch to use that method instead.

PR #10796, opened 2018-01-12
Add score to fulltext results

ragadeeshu:
This means that you can get the lucene scoring of the results returned by the fulltext addon.

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/ValuesIterator.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/ValuesIterator.java
@@ -49,6 +49,12 @@ public abstract class ValuesIterator extends PrimitiveLongCollections.PrimitiveL
         }
 
         @Override
+        public float currentScore()
+        {
+            return 0;
+        }
+
+        @Override
         public long getValue( String field )
         {
             return 0;
@@ -57,6 +63,7 @@ public abstract class ValuesIterator extends PrimitiveLongCollections.PrimitiveL
 
     protected final int size;
     protected int index;
+    public abstract float currentScore();
 
     ValuesIterator( int size )
     {
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/ValuesIterator.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/ValuesIterator.java
@@ -63,6 +63,11 @@ public abstract class ValuesIterator extends PrimitiveLongCollections.PrimitiveL
 
     protected final int size;
     protected int index;
+
+    /**
+     * Gets the score for the current iterator position.
+     * @return The score of the value, or 0 if scoring is not kept or applicable.
+     */
     public abstract float currentScore();
 
     ValuesIterator( int size )
--- a/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/ScoreEntityIterator.java
+++ b/enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/ScoreEntityIterator.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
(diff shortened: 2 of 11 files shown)


enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomProcedures.java
```diff
@@ -50,6 +53,7 @@
 {
     @Context
     public FulltextProvider provider;
+    public static final Function<Pair<Long,Float>,EntityOutput> QUERY_RESULT_MAPPER = result -> new EntityOutput( result.first(), result.other() );
```
MishaDemianenko: it looks like this mapper can be private

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/ValuesIterator.java
```diff
@@ -57,6 +63,7 @@ public long getValue( String field )
 
     protected final int size;
     protected int index;
+    public abstract float currentScore();
```
MishaDemianenko: java doc is missing

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/ScoreEntityIterator.java
```diff
@@ -0,0 +1,139 @@
...
+    {
+        return iterator.hasNext();
+    }
+
+    @Override
+    public Pair<Long,Float> next()
+    {
+        if ( hasNext() )
+        {
+            return Pair.of( currentIterator().next(), currentIterator().currentScore() );
```
MishaDemianenko: It looks like instead of using Pair it's better to create specific ScoreEntity POJO with two primitive fields: entity/value and score; 
so uses can understand directly understand what first and the second value represents.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/ScoreEntityIterator.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.NoSuchElementException;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+import org.neo4j.helpers.collection.Pair;
+import org.neo4j.kernel.api.impl.index.collector.ValuesIterator;
+
+public class ScoreEntityIterator implements Iterator<Pair<Long,Float>>
```
MishaDemianenko: Except for high level integration test that shows how it works, in most basic scenario, there are no tests for more advanced scenario, for example, iterators combine case

MishaDemianenko (changes requested):
Also would be nice to have some description in PR describing new possibility for possible users

ragadeeshu:
@MishaDemianenko thanks, updated!

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/ScoreEntityIterator.java
```diff
@@ -105,35 +108,95 @@ public static ScoreEntityIterator emptyIterator()
...
+            }
+        }
+    }
+
+    /**
+     * A ScoreEntry consists of an entity id together with its score.
+     */
+    public static class ScoreEntry
+    {
+        private final long entityid;
```
MishaDemianenko: looks like `i` in `id` should be in upper case

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/ScoreEntityIterator.java
```diff
@@ -105,35 +108,95 @@ public static ScoreEntityIterator emptyIterator()
 
     private static class ConcatenatingScoreEntityIterator extends ScoreEntityIterator
     {
-        private final Iterator<? extends ScoreEntityIterator> iterators;
+        private final List<? extends ScoreEntityIterator> iterators;
+        private final ScoreEntry[] buffer;
         private ScoreEntityIterator currentIterator;
```
MishaDemianenko: this field is not used anymore and should be deleted

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/ScoreEntityIterator.java
```diff
@@ -105,35 +108,95 @@ public static ScoreEntityIterator emptyIterator()
...
+            {
+                fetch();
+            }
+            return nextHead != null;
+        }
+
+        private void fetch()
+        {
+            int candidateHead = -1;
+            for ( int i = 0; i < iterators.size(); i++ )
```
MishaDemianenko, line 135: do not think its that important in this case since we will have one iterator (ok, ok - two) in mostly in all the cases, but if in future or you still think it's important to make this part faster we can use binary heap here instead.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/ScoreEntityIterator.java
```diff
@@ -105,35 +108,95 @@ public static ScoreEntityIterator emptyIterator()
...
+        {
+            return entityid;
+        }
+
+        public float score()
+        {
+            return score;
+        }
+
+        public ScoreEntry( long entityid, float score )
```
MishaDemianenko: this constructor can be private

MishaDemianenko (approved):
I still have couple minor comments, but will approve PR now since fixing those should be super simple

PR #10799, opened 2018-01-12
Pluggable batch importer

tinwelint:
Loaded using service loading. Since importer logic moved into ImportLogic and
ParallelBatchImporter is a thin layer using it, this service loading enables
other implementations using that same base

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
@@ -88,10 +88,10 @@ import org.neo4j.kernel.impl.util.monitoring.SilentProgressReporter;
 import org.neo4j.logging.NullLogProvider;
 import org.neo4j.unsafe.impl.batchimport.AdditionalInitialIds;
 import org.neo4j.unsafe.impl.batchimport.BatchImporter;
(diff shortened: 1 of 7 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
```diff
@@ -371,6 +371,8 @@ public boolean parallelRecordReadsWhenWriting()
                     readAdditionalIds( lastTxId, lastTxChecksum, lastTxLogVersion, lastTxLogByteOffset );
 
             // We have to make sure to keep the token ids if we're migrating properties/labels
+            // Use the standard importer, instead of loading from BatchImporterFactory because an aborted store migration
+            // will be restarted from scratch anyway
```
MishaDemianenko: That's true that there is no feature driven reason to use edition based importer in this place right now, but still, maybe it makes sense to do a switch now so any future edition based improvements can be used automatically?

MishaDemianenko (approved):

community/kernel/src/main/java/org/neo4j/kernel/impl/storemigration/participant/StoreMigrator.java
tinwelint: Perhaps you're right, I'm OK with doing this change.

MishaDemianenko (approved):

PR #10800, opened 2018-01-12
Reduce the number of allocations for big plans

pontusmelke:
We do a lot of allocation in planning, so much so that for big plans the situation gets completely out of hand. For big queries of the type, 
```
MATCH (a) CREATE (:A1)-[:R1]->(:B1), (:A2)-[:R2]->(:B2) , ...
```
where we get big deeply nested logical plans we make it very hard for the JVM to get through planning because of the many allocations. This PR tackles the worse offenders for these types of queries.


fickludd (approved):
This looks like a bunch of good changes!

kfreytag:
Would be great to get this into 3.3

jakewins:
Heads up that I tested this patch with the hiccuping instance we had, and this, together with the two PRs that were merged already, made the sad query happy. Thanks for jumping on this @pontusmelke!

PR #10801, opened 2018-01-13
Removes some unnecessary indirections from embedded impl

tinwelint:
NodeProxy/RelationshipProxy needs some sort of SPI, acting as their relationship
to the database instance. There were a couple of unnecessary indirections and
round trips for the execution paths right below them, which were apparently
not really required and only made code more complicated and weird.
This is a cleanup and simplification of that part of the code.


chrisvest (approved):
Nice bit of simplification. I suspect the build might go green with a rebase. Please make sure it's green before merging.

PR #10802, opened 2018-01-14
Collections primitivization

MishaDemianenko:
Use primitive collections where its possible
in all the modules


klaren (approved):
Changes looks good. :+1:  

Not sure about the performance implications though, would be nice to get some benchmarking on all these primitivization changes.

PR #10805, opened 2018-01-15
Update compiler dependency to match neo4j release 3.2.9

SaschaPeukert:


pontusmelke (changes requested):
There is a build failure, something has changed in 3.2.9 that requires an update here.

pontusmelke (approved):

PR #10809, opened 2018-01-15
Better reaction to half-initialized GBPTree index file

tinwelint:
If creation of a GBPTree index file is interrupted mid-way, e.g. by process
crashing or something similar then the index file will exist, but its
meta pages empty or half-initialized. Previously only a missing index file
was interpreted correctly by throwing IOException, such that the layer
immediately above GBPTree would recognized that and rebuild that index or similar,
whereas the other cases of half-initialized meta pages would throw other
types of internal exceptions and not be properly handled by above layer.

This commit improves on this situation to interpret one or more missing or
all-zero meta pages the same way, i.e. by throwing proper IOException.

### DIFF
--- a/community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
+++ b/community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
@@ -56,6 +56,7 @@ import org.neo4j.collection.primitive.Primitive;
 import org.neo4j.collection.primitive.PrimitiveLongCollections;
 import org.neo4j.collection.primitive.PrimitiveLongSet;
 import org.neo4j.cursor.RawCursor;
+import org.neo4j.function.ThrowingConsumer;
 import org.neo4j.graphdb.mockfs.EphemeralFileSystemAbstraction;
 import org.neo4j.index.internal.gbptree.GBPTree.Monitor;
 import org.neo4j.io.pagecache.DelegatingPageCache;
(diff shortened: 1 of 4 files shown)


community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
```diff
@@ -684,6 +686,75 @@ public void readHeaderMustThrowIfFileDoesNotExist() throws Exception
         }
     }
 
+    @Test
+    public void readHeaderMustThrowIOExceptionIfFileIsEmpty() throws Exception
```
burqen: Really sweet testing!
Can we also verify that constructor behave the same way as `readHeader` here?

burqen (changes requested):
This is sweet 👍  I think we should unite

burqen (changes requested):
This is sweet! 👍 
I think we should also align constructor to behave in the same way and add testing for that. Constructor does not use `readHeader`.

Would also be neat if we could verify intended behaviour in some test. The population restart that is.

community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
tinwelint: Oh of course, good call. I'll duplicate these tests for the constructor!

tinwelint:
Right about the testing... the test seeing this also saw all these three issues, so they spawned from that and that test could see them, but of course if that test changes then that accidental testing goes away. I can add specific ITs for this.

burqen:
@tinwelint That would be awesome!

tinwelint:
@burqen updated

burqen (approved):

PR #10811, opened 2018-01-15
Fix a few cases where a static initialiser references a subclass.

chrisvest:
These situations can lead to classloader deadlocks, if the super-class and the sub-class ends up being loaded concurrently in two different threads.
Not all cases are fixed: some a locked in place by being public API, and another was left be due to unpredictable consequences for forward merging.


burqen (changes requested):
Nice catch! Since this is not a real problem for us right now I think we can direct this cleanup to latest dev branch instead.

PR #10815, opened 2018-01-15
Cleanup accidental Path exposure as a Resource.

MishaDemianenko:
Even taking into account that this change already have been already
leaked to public, cleaning this up as soon as possible to minimize
future impact.
The current effect should be minimal since path instances were not
treated as resources anywhere, except tests.
Any user code, if that exists, should be fixed by directly implementing
Resource interface if required.


thobe:
Looks like we did the same thing: #10816

thobe (approved):
Looks good.

PR #10816, opened 2018-01-16
Path should not extend Resource

thobe:
First of all, it makes it much trickier to use paths.
Secondly the implementation of close() did nothing.


thobe:
Duplicate of #10815

PR #10818, opened 2018-01-16
Handle large plans

pontusmelke:
In 3.4 we have introduced a few new places where we can overflow the stack and a few new inefficiencies which hurts us for big plans.

- `hashCode` for logical plan was overflowing the stack
- we did a full `toString` for the full logical plan when  the "schema" runtime and the compiled runtime decided it couldn't handle the plan. Nice for debugging but it means basically all queries need to generate big string descriptions for each query twice.
- We were keeping a hashMap with logical plans as key that was never used. Computing `equals` on logical plans is quite expensive so we should not do that if we don't really have to.
- We do a lot of hashing of QueryGraph, by memoizing the value we save a lot of time.

### DIFF
--- a/community/cypher/cypher-logical-plans-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/logical/plans/LogicalPlan.scala
+++ b/community/cypher/cypher-logical-plans-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/logical/plans/LogicalPlan.scala
@@ -21,14 +21,19 @@ package org.neo4j.cypher.internal.v3_4.logical.plans
 
 import java.lang.reflect.Method
 
-import org.neo4j.cypher.internal.util.v3_4.Foldable._
-import org.neo4j.cypher.internal.util.v3_4.{Foldable, InternalException, Rewritable, Unchangeable}
-import org.neo4j.cypher.internal.util.v3_4.Rewritable._
 import org.neo4j.cypher.internal.ir.v3_4.{PlannerQuery, Strictness}
-import org.neo4j.cypher.internal.util.v3_4.attribution.{IdGen, SameId}
+import org.neo4j.cypher.internal.util.v3_4.Foldable._
+import org.neo4j.cypher.internal.util.v3_4.Rewritable._
+import org.neo4j.cypher.internal.util.v3_4.attribution.{Id, IdGen, SameId}
+import org.neo4j.cypher.internal.util.v3_4.{Foldable, InternalException, Rewritable, Unchangeable}
 import org.neo4j.cypher.internal.v3_4.expressions.Expression
 
+import scala.util.hashing.MurmurHash3
 import scala.collection.mutable
+
+object LogicalPlan {
+  val LOWEST_TX_LAYER = 0
+}
 
 /*
 A LogicalPlan is an algebraic query, which is represented by a query tree whose leaves are database relations and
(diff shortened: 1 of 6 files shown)


enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlottedRewriter.scala
```diff
@@ -333,17 +326,15 @@ class SlottedRewriter(tokenContext: TokenContext) {
       propExpression
   }
 
-  private def stopAtOtherLogicalPlans(thisPlan: LogicalPlan): (AnyRef) => Boolean = { lp =>
-    lp match {
-      case _: LogicalPlan =>
-        lp != thisPlan
+  private def stopAtOtherLogicalPlans(thisPlan: LogicalPlan): (AnyRef) => Boolean = {
+    case lp@(_: LogicalPlan) =>
+      lp.id != thisPlan.id
```
sherfert, line 331: Just a question: Does this assume that we have no collisions in hashCode or does it tolerate that we stop traversing if we have a hash collision?

pontusmelke: Not sure I follow, what are we hashing?

sherfert: Nevermind. I confused `id` with `debugId`.

pontusmelke: 👍

community/cypher/cypher-logical-plans-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/logical/plans/LogicalPlan.scala
```diff
@@ -21,15 +21,20 @@ package org.neo4j.cypher.internal.v3_4.logical.plans
...
+import org.neo4j.cypher.internal.util.v3_4.attribution.{Id, IdGen, SameId}
+import org.neo4j.cypher.internal.util.v3_4.{Foldable, InternalException, Rewritable, Unchangeable}
 import org.neo4j.cypher.internal.v3_4.expressions.Expression
 
+import scala.util.hashing.MurmurHash3
 import scala.collection.mutable
 
+object LogicalPlan {
+  val LOWEST_TX_LAYER = 0
+}
```
fickludd, line 36: I think this was added by mistake. I'm going to merge anyway if this get green, and remove that in my next PR.

fickludd (approved):
Awesome clean-ups and optimizations. Making these mistakes is very subtle indeed.

community/cypher/cypher-logical-plans-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/logical/plans/LogicalPlan.scala
pontusmelke: yes it might have snuck in via a rebase conflict

PR #10819, opened 2018-01-16
Make `dbms.tx_log.rotation.*` settings dynamic

klaren:
`dbms.tx_log.rotation.retention_policy` and `dbms.tx_log.rotation.size` is now makred as dynamic and can be changed at runtime.

If the retention policy is changed while a logical log pruning is happening, it will take effect the next time the pruning is evaluated, typically after a checkpoint has occurred.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -47,7 +47,6 @@ import org.neo4j.kernel.impl.cache.MonitorGc;
 import org.neo4j.logging.Level;
 import org.neo4j.logging.LogTimeZone;
 
-import static org.neo4j.kernel.configuration.Settings.ANY;
 import static org.neo4j.kernel.configuration.Settings.BOOLEAN;
 import static org.neo4j.kernel.configuration.Settings.BYTES;
 import static org.neo4j.kernel.configuration.Settings.DEFAULT;
@@ -507,13 +506,19 @@ public class GraphDatabaseSettings implements LoadableConfig
             "Can be used for specifying the threshold to prune logical logs after. For example \"10 days\" will " +
             "prune logical logs that only contains transactions older than 10 days from the current time, " +
             "or \"100k txs\" will keep the 100k latest transactions and prune any older transactions." )
-    public static final Setting<String> keep_logical_logs = buildSetting( "dbms.tx_log.rotation.retention_policy",
-            STRING, "7 days" ).constraint( illegalValueMessage( "must be `true`/`false` or " +
-                    "of format '<number><optional unit> <type>' for example `100M size` for " +
-                    "limiting logical log space on disk to 100Mb," +
-                    " or `200k txs` for limiting the number of transactions to keep to 200 000", matches(ANY) ) ).build();
+    @Dynamic
+    public static final Setting<String> keep_logical_logs =
+            buildSetting( "dbms.tx_log.rotation.retention_policy", STRING, "7 days" ).constraint( illegalValueMessage(
+                    "must be `true`, `false` or of format `<number><optional unit> <type>`. " +
+                            "Valid units are `k`, `M` and `G`. " +
+                            "Valid types are `files`, `size`, `txs`, `entries`, `hours` and `days`. " +
+                            "For example, `100M size` will limiting logical log space on disk to 100Mb," +
+                            " or `200k txs` will limiting the number of transactions to keep to 200 000", matches(
+                            "^(true|keep_all|false|keep_none|(\\d+[KkMmGg]?( (files|size|txs|entries|hours|days))))$" ) ) )
+                    .build();
 
     @Description( "Specifies at which file size the logical log will auto-rotate. Minimum accepted value is 1M. " )
+    @Dynamic
     public static final Setting<Long> logical_log_rotation_threshold =
             buildSetting( "dbms.tx_log.rotation.size", BYTES, "250M" ).constraint( min( ByteUnit.mebiBytes( 1 ) ) ).build();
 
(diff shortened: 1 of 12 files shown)


community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -507,13 +506,19 @@
...
-                    " or `200k txs` for limiting the number of transactions to keep to 200 000", matches(ANY) ) ).build();
+    @Dynamic
+    public static final Setting<String> keep_logical_logs =
+            buildSetting( "dbms.tx_log.rotation.retention_policy", STRING, "7 days" ).constraint( illegalValueMessage(
+                    "must be `true`, `false` or of format `<number><optional unit> <type>`. " +
+                            "Valid units are `k`, `M` and `G`. " +
+                            "Valid types are `files`, `size`, `txs`, `entries`, `hours` and `days`. " +
+                            "For example, `100M size` will limiting logical log space on disk to 100Mb," +
+                            " or `200k txs` will limiting the number of transactions to keep to 200 000", matches(
+                            "^(true|keep_all|false|keep_none|(\\d+[KkMmGg]?( (files|size|txs|entries|hours|days))))$" ) ) )
```
MishaDemianenko, line 517: i do not think we need to have this regexp here

community/kernel/src/test/java/org/neo4j/graphdb/factory/GraphDatabaseSettingsTest.java
```diff
@@ -416,4 +417,31 @@ public void shouldStillSupportCustomNameForHttpsConnector() throws Exception
...
+    {
+        String[] validSet =
+                new String[]{"true", "keep_all", "false", "keep_none", "10 files", "10k files", "10K size", "10m txs",
+                        "10M entries", "10g hours", "10G days"};
+
+        String[] invalidSet = new String[]{"invalid", "all", "10", "10k", "10k a"};
+
+        for ( String valid : validSet )
+        {
+            Config.defaults( keep_logical_logs, valid );
```
MishaDemianenko: i think we need to add a bit more assertions here

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruningIT.java
```diff
@@ -0,0 +1,113 @@
...
+import org.neo4j.test.rule.EmbeddedDatabaseRule;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.neo4j.graphdb.factory.GraphDatabaseSettings.keep_logical_logs;
+
+public class LogPruningIT
+{
+    @Rule
+    public DatabaseRule db = new EmbeddedDatabaseRule().withSetting( keep_logical_logs, "true" );
```
MishaDemianenko: 'final' is missing

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/log/pruning/LogPruningIT.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private final SimpleTriggerInfo triggerInfo = new SimpleTriggerInfo( "forced trigger" );
+
+    @Test
+    public void pruningStrategyShouldBeDynamic() throws IOException
+    {
+        CheckPointer checkPointer = db.getDependencyResolver().resolveDependency( CheckPointer.class );
```
MishaDemianenko: extract a helper method here to get T from 'DependencyResolver' that should help with readability and characters

MishaDemianenko (changes requested):
Couple of minor comments otherwise looks good

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
klaren: We do, otherwise, validation will not work for dynamic setting updates.

community/kernel/src/test/java/org/neo4j/graphdb/factory/GraphDatabaseSettingsTest.java
klaren: Assert on what? It throws if it's invalid.
Edit: but I can add a comment clarifying that.

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
MishaDemianenko: ah, stupid me; missed \" quote and was reading it as part of description string

community/kernel/src/test/java/org/neo4j/graphdb/factory/GraphDatabaseSettingsTest.java
MishaDemianenko: i'm thinking about asserts that Config.defaults(....).get(..property) equals to expected value

klaren: It's an identity setting which returns exactly the same as you put in. But it does not hurt to have an assert there :)

klaren:
@MishaDemianenko New stuff to dig into!

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
MishaDemianenko: that strange feeling when you see a thumbs up emoji under `stupid me` comment... 🙉

MishaDemianenko (approved):

PR #10820, opened 2018-01-16
Fix bug in NodeOuterHashJoin slot allocation and row creation

fickludd:
Rows need to be created using the `ExecutionContextFactory`, otherwise they do not get the correct slots when mixed in from the slotted runtime.

### DIFF
--- a/community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/NodeOuterHashJoinPipe.scala
+++ b/community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/NodeOuterHashJoinPipe.scala
@@ -26,11 +26,12 @@ import org.neo4j.values.storable.Values
 import org.neo4j.values.virtual.VirtualNodeValue
 
 import scala.collection.mutable
+import scala.collection.mutable.MutableList
 import scala.collection.mutable.ListBuffer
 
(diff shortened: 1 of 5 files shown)


community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/NodeOuterHashJoinPipe.scala
```diff
@@ -59,7 +60,7 @@ case class NodeOuterHashJoinPipe(nodeVariables: Set[String], source: Pipe, inner
     rowsWithNullAsJoinKey ++ joinedRows ++ rowsWithoutRhsMatch
   }
 
-  private def addNulls(in: ExecutionContext): ExecutionContext = in.set(nullColumns)
+  private def addNulls(in: ExecutionContext): ExecutionContext = executionContextFactory.copyWith(in).set(nullColumns)
```
pontusmelke, line 63: This is the actual error right?

pontusmelke (approved):
Nice!

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/pipes/NodeOuterHashJoinPipe.scala
fickludd: Yes. That and line 51.

PR #10821, opened 2018-01-16
Added a `neo4j-admin memrec` command, which can print recommendations for the Neo4j heap and pagecache memory settings.

chrisvest:
This is an initial stab at having automated recommendations for the Neo4j memory settings.

It looks like this:

```
$ bin/neo4j-admin help
usage: neo4j-admin <command>

Manage your Neo4j instance.

environment variables:
    HEAP_SIZE     Set size of JVM heap during command execution.
                  Takes a number and a unit, for example 512m.

available commands:

General
    check-consistency
    help
        This help text, or help for the command specified in <command>.
    import
    memrec
        Print Neo4j heap and pagecache memory settings recommendations.
    store-info
        Prints information about a Neo4j database store.

Authentication
    set-default-admin
    set-initial-password

Offline backup
    dump
    load

```

And

```
$ bin/neo4j-admin help memrec
usage: neo4j-admin memrec [--memory=<16g>]

environment variables:
    HEAP_SIZE     Set size of JVM heap during command execution.
                  Takes a number and a unit, for example 512m.

Print heuristic memory setting recommendations for the Neo4j JVM heap and
pagecache. The heuristic is based on the total memory of the system the command
is running on, or on the amount of memory specified with the --memory argument.
The heuristic assumes that the system is dedicated to running Neo4j. If this is
not the case, then use the --memory argument to specify how much memory can be
expected to be dedicated to Neo4j.

The output is formatted such that it can be copy-posted into the neo4j.conf
file.

options:
  --memory=<16g>   Recommend memory settings with respect to the given amount of
                  memory, instead of the total memory of the system running the
                  command. [default:16g]
```

And

```
$ bin/neo4j-admin memrec --memory 12g
# Memory settings recommendation from neo4j-admin memrec:
#
# Assuming the system is dedicated to running Neo4j and has 12g of memory,
# we recommend a heap size of around 4600m, and a page cache of around 4g,
# and that about 3600m is left for the operating system, and the native memory
# needed by Lucene and Netty.
#
# Tip: If the indexing storage use is high, e.g. there are many indexes or most
# data indexed, then it might advantageous to leave more memory for the
# operating system.
#
# Tip: The more concurrent transactions your workload has and the more updates
# they do, the more heap memory you will need. However, don't allocate more
# than 31g of heap, since this will disable pointer compression, also known as
# "compressed oops", in the JVM and make less effective use of the heap.
#
# Tip: Setting the initial and the max heap size to the same value means the
# JVM will never need to change the heap size. Changing the heap size otherwise
# involves a full GC, which is desirable to avoid.
#
# Based on the above, the following memory settings are recommended:
dbms.memory.heap.initial_size=4600m
dbms.memory.heap.max_size=4600m
dbms.memory.pagecache.size=4g
```

### DIFF
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
@@ -1,0 +1,249 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.commandline.dbms;
+
+import java.util.Locale;
+import java.util.function.ToDoubleFunction;
+
+import org.neo4j.commandline.admin.AdminCommand;
+import org.neo4j.commandline.admin.IncorrectUsage;
+import org.neo4j.commandline.admin.OutsideWorld;
+import org.neo4j.commandline.arguments.Arguments;
+import org.neo4j.commandline.arguments.OptionalNamedArg;
+import org.neo4j.kernel.impl.util.OsBeanUtil;
+
+import static java.lang.String.format;
+import static org.neo4j.configuration.ExternalSettings.initialHeapSize;
+import static org.neo4j.configuration.ExternalSettings.maxHeapSize;
+import static org.neo4j.graphdb.factory.GraphDatabaseSettings.pagecache_memory;
+import static org.neo4j.io.ByteUnit.ONE_GIBI_BYTE;
+import static org.neo4j.io.ByteUnit.ONE_KIBI_BYTE;
+import static org.neo4j.io.ByteUnit.ONE_MEBI_BYTE;
+import static org.neo4j.io.ByteUnit.gibiBytes;
+import static org.neo4j.io.ByteUnit.mebiBytes;
+import static org.neo4j.io.ByteUnit.tebiBytes;
+import static org.neo4j.kernel.configuration.Settings.BYTES;
+import static org.neo4j.kernel.configuration.Settings.buildSetting;
+
+public class MemoryRecommendationsCommand implements AdminCommand
+{
+    // Fields: {System Memory in GiBs; OS memory reserve in GiBs; JVM Heap memory in GiBs}.
+    // And the page cache gets what's left, though always at least 100 MiB.
+    // Heap never goes beyond 31 GiBs, which is the upper limit for compressed oops.
+    // More details on heap sizing wrt. compressed oops:
+    // https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html
+    private static final Bracket[] datapoints = {
+            new Bracket( 0.01, 0.005, 0.005 ),
+            new Bracket( 1.0, 0.5, 0.5 ),
+            new Bracket( 2.0, 1, 0.5 ),
+            new Bracket( 4.0, 1.5, 2 ),
+            new Bracket( 6.0, 2, 3 ),
+            new Bracket( 8.0, 2.5, 3.5 ),
+            new Bracket( 10.0, 3, 4 ),
+            new Bracket( 12.0, 3.5, 4.5 ),
+            new Bracket( 16.0, 4, 5 ),
+            new Bracket( 24.0, 6, 8 ),
+            new Bracket( 32.0, 8, 12 ),
+            new Bracket( 64.0, 12, 24 ),
+            new Bracket( 128.0, 16, 31 ),
+            new Bracket( 256.0, 20, 31 ),
+            new Bracket( 512.0, 24, 31 ),
+            new Bracket( 1024.0, 30, 31 ),
+    };
+    private static final String ARG_MEMORY = "memory";
+    private final OutsideWorld outsideWorld;
+
+    static long recommendOsMemory( long totalMemoryBytes )
+    {
+        Brackets brackets = findMemoryBrackets( totalMemoryBytes );
+        return brackets.recommend( Bracket::osMemory );
+    }
+
+    static long recommendHeapMemory( long totalMemoryBytes )
+    {
+        Brackets brackets = findMemoryBrackets( totalMemoryBytes );
+        return brackets.recommend( Bracket::heapMemory );
+    }
+
+    static long recommendPageCacheMemory( long totalMemoryBytes )
+    {
+        long osMemory = recommendOsMemory( totalMemoryBytes );
+        long heapMemory = recommendHeapMemory( totalMemoryBytes );
+        long recommendation = totalMemoryBytes - osMemory - heapMemory;
+        recommendation = Math.max( mebiBytes( 100 ), recommendation );
+        recommendation = Math.min( tebiBytes( 16 ), recommendation );
+        return recommendation;
+    }
+
+    private static Brackets findMemoryBrackets( long totalMemoryBytes )
+    {
+        double totalMemoryGB = ((double) totalMemoryBytes) / ((double) gibiBytes( 1 ));
+        Bracket lower = null;
+        Bracket upper = null;
+        for ( int i = 1; i < datapoints.length; i++ )
+        {
+            if ( totalMemoryGB < datapoints[i].totalMemory )
+            {
+                lower = datapoints[i - 1];
+                upper = datapoints[i];
+                break;
+            }
+        }
+        if ( lower == null )
+        {
+            lower = datapoints[datapoints.length - 1];
+            upper = datapoints[datapoints.length - 1];
+        }
+        return new Brackets( totalMemoryGB, lower, upper );
+    }
+
+    public static Arguments buildArgs()
+    {
+        long totalPhysicalMemory = OsBeanUtil.getTotalPhysicalMemory();
+        return new Arguments()
+                .withArgument( new OptionalNamedArg( ARG_MEMORY, "8g", bytesToString( totalPhysicalMemory ),
+                        "Recommend memory settings with respect to the given amount of memory, " +
+                        "instead of the total memory of the system running the command." ) );
+    }
+
+    static String bytesToString( long bytes )
+    {
+        if ( bytes > ONE_GIBI_BYTE )
+        {
+            return format( Locale.ROOT, "%.1fg", bytes / (double) ONE_GIBI_BYTE );
+        }
+        else if ( bytes > ONE_MEBI_BYTE )
+        {
+            return format( Locale.ROOT, "%.1fm", bytes / (double) ONE_MEBI_BYTE );
+        }
+        else if ( bytes > ONE_KIBI_BYTE )
+        {
+            return format( Locale.ROOT, "%.0fk", bytes / (double) ONE_KIBI_BYTE );
+        }
+        else
+        {
+            return String.valueOf( bytes );
+        }
+    }
+
+    MemoryRecommendationsCommand( OutsideWorld outsideWorld )
+    {
+        this.outsideWorld = outsideWorld;
+    }
+
+    @Override
+    public void execute( String[] args ) throws IncorrectUsage
+    {
+        Arguments arguments = buildArgs().parse( args );
+
+        String mem = arguments.get( ARG_MEMORY );
+        long memory = buildSetting( ARG_MEMORY, BYTES ).build().apply( arguments::get );
+        String os = bytesToString( recommendOsMemory( memory ) );
+        String heap = bytesToString( recommendHeapMemory( memory ) );
+        String pagecache = bytesToString( recommendPageCacheMemory( memory ) );
+
+        print( "# Memory settings recommendation from neo4j-admin memrec:" );
+        print( "#" );
+        print( "# Assuming the system is dedicated to running Neo4j and has " + mem + " of memory," );
+        print( "# we recommend a heap size of around " + heap + ", and a page cache of around " + pagecache + "," );
+        print( "# and that about " + os + " is left for the operating system, and the native memory" );
+        print( "# needed by Lucene and Netty." );
+        print( "#" );
+        print( "# Tip: If the indexing storage use is high, e.g. there are many indexes or most" );
+        print( "# data indexed, then it might advantageous to leave more memory for the" );
+        print( "# operating system." );
+        print( "#" );
+        print( "# Tip: The more concurrent transactions your workload has and the more updates" );
+        print( "# they do, the more heap memory you will need. However, don't allocate more" );
+        print( "# than 31g of heap, since this will disable pointer compression, also known as" );
+        print( "# \"compressed oops\", in the JVM and make less effective use of the heap." );
+        print( "#" );
+        print( "# Tip: Setting the initial and the max heap size to the same value means the" );
+        print( "# JVM will never need to change the heap size. Changing the heap size otherwise" );
+        print( "# involves a full GC, which is desirable to avoid." );
+        print( "#" );
+        print( "# Based on the above, the following memory settings are recommended:" );
+        print( initialHeapSize.name() + "=" + heap );
+        print( maxHeapSize.name() + "=" + heap );
+        print( pagecache_memory.name() + "=" + pagecache );
+    }
+
+    private void print( String text )
+    {
+        outsideWorld.stdOutLine( text );
+    }
+
+    private static final class Bracket
+    {
+        private final double totalMemory;
+        private final double osMemory;
+        private final double heapMemory;
+
+        private Bracket( double totalMemory, double osMemory, double heapMemory )
+        {
+            this.totalMemory = totalMemory;
+            this.osMemory = osMemory;
+            this.heapMemory = heapMemory;
+        }
+
+        double osMemory()
+        {
+            return osMemory;
+        }
+
+        double heapMemory()
+        {
+            return heapMemory;
+        }
+    }
+
+    private static final class Brackets
+    {
+        private final double totalMemoryGB;
+        private final Bracket lower;
+        private final Bracket upper;
+
+        private Brackets( double totalMemoryGB, Bracket lower, Bracket upper )
+        {
+            this.totalMemoryGB = totalMemoryGB;
+            this.lower = lower;
+            this.upper = upper;
+        }
+
+        private double differenceFactor()
+        {
+            if ( lower == upper )
+            {
+                return 0;
+            }
+            return (totalMemoryGB - lower.totalMemory) / (upper.totalMemory - lower.totalMemory);
+        }
+
+        public long recommend( ToDoubleFunction<Bracket> parameter )
+        {
+            double factor = differenceFactor();
+            double lowerParam = parameter.applyAsDouble( lower );
+            double upperParam = parameter.applyAsDouble( upper );
+            double diff = upperParam - lowerParam;
+            double recommend = lowerParam + (diff * factor);
+            return mebiBytes( (long) (recommend * 1024.0) );
+        }
+    }
+}
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
@@ -123,19 +123,37 @@ public class MemoryRecommendationsCommand implements AdminCommand
                         "instead of the total memory of the system running the command." ) );
     }
 
-    static String bytesToString( long bytes )
+    static String bytesToString( double bytes )
     {
-        if ( bytes > ONE_GIBI_BYTE )
+        double gibi1 = ONE_GIBI_BYTE;
+        double mebi1 = ONE_MEBI_BYTE;
+        double mebi100 = 100 * mebi1;
+        long kibi100 = 100 * ONE_KIBI_BYTE;
+        if ( bytes >= gibi1 )
         {
-            return format( Locale.ROOT, "%.1fg", bytes / (double) ONE_GIBI_BYTE );
+            double gibibytes = bytes / gibi1;
+            double modMebi = bytes % gibi1;
+            if ( modMebi >= mebi100 )
+            {
+                return format( Locale.ROOT, "%dm", Math.round( bytes / mebi100 ) * 100 );
+            }
+            else
+            {
+                return format( Locale.ROOT, "%.0fg", gibibytes );
+            }
         }
-        else if ( bytes > ONE_MEBI_BYTE )
+        else if ( bytes >= mebi1 )
         {
-            return format( Locale.ROOT, "%.1fm", bytes / (double) ONE_MEBI_BYTE );
-        }
-        else if ( bytes > ONE_KIBI_BYTE )
-        {
-            return format( Locale.ROOT, "%.0fk", bytes / (double) ONE_KIBI_BYTE );
+            double mebibytes = bytes / mebi1;
+            double modKibi = bytes % mebi1;
+            if ( modKibi >= kibi100 )
+            {
+                return format( Locale.ROOT, "%dk", Math.round( bytes / kibi100 ) * 100 );
+            }
+            else
+            {
+                return format( Locale.ROOT, "%.0fm", mebibytes );
+            }
         }
         else
         {
--- a/community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
+++ b/community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
@@ -46,9 +46,7 @@ public class MemoryRecommendationsCommand implements AdminCommand
 {
     // Fields: {System Memory in GiBs; OS memory reserve in GiBs; JVM Heap memory in GiBs}.
     // And the page cache gets what's left, though always at least 100 MiB.
-    // Heap never goes beyond 31 GiBs, which is the upper limit for compressed oops.
-    // More details on heap sizing wrt. compressed oops:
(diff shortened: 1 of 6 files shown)


community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+import static org.neo4j.io.ByteUnit.mebiBytes;
+import static org.neo4j.io.ByteUnit.tebiBytes;
+import static org.neo4j.kernel.configuration.Settings.BYTES;
+import static org.neo4j.kernel.configuration.Settings.buildSetting;
+
+public class MemoryRecommendationsCommand implements AdminCommand
+{
+    // Fields: {System Memory in GiBs; OS memory reserve in GiBs; JVM Heap memory in GiBs}.
+    // And the page cache gets what's left, though always at least 100 MiB.
+    // Heap never goes beyond 32 GiBs, which is the upper limit for compressed oops..
```
spacecowboy: You fail to mention Lucene and Netty off-heap memory needs. How are those taken into account here?
And what do you mean "oops"?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+                .withArgument( new OptionalBooleanArg( ARG_PRINT_CONF, false,
+                        "Print what an updated neo4j.conf file would look like to standard out. Does not change " +
+                        "existing settings unless --overwrite is specified." ) )
+                .withArgument( new OptionalBooleanArg( ARG_UPDATE_CONF, false,
+                        "Update the neo4j.conf file, or the given --conf-file, in-place by adding the recommended " +
+                        "memory settings, unless they are already specified. " +
+                        "Specify --overwrite to also update existing settings." ) )
+                .withArgument( new OptionalBooleanArg( ARG_OVERWRITE, false,
+                        "Causes --print-conf and --update-conf to overwrite existing settings, if set." ) )
+                .withArgument( new OptionalNamedArg( ARG_CONF_FILE, "conf" + File.separator + "neo4j.conf", "",
```
spacecowboy: This default path is incorrect for everything except the tarball. I think it would be easier for everybody if it just didn't have a default value.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+
+    public static Arguments buildArgs()
+    {
+        return new Arguments()
+                .withArgument( new OptionalNamedArg( ARG_MEMORY, "8g", OsBeanUtil.getTotalPhysicalMemory() + "",
+                        "Recommend memory settings with respect to the given amount of memory, " +
+                        "instead of the total memory of the system running the command." ) )
+                .withArgument( new OptionalBooleanArg( ARG_PRINT_CONF, false,
+                        "Print what an updated neo4j.conf file would look like to standard out. Does not change " +
+                        "existing settings unless --overwrite is specified." ) )
```
spacecowboy: I wonder if you should refer the the variable name here instead of hardcoding `--overwrite`

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
spacecowboy: One reason for me saying so is that on Linux systems the user running the command will typically not have the same environment set as the Neo4j service does. So there will be no way to figure out where the config is.

A user will be invoking it like this

    neo4j-admin memrec .... /etc/neo4j/neo4j.conf

that command will fail because the user doesn't have write access to `/etc`. So second time around the user sudos:

    sudo neo4j-admin memrec .... /etc/neo4j/neo4j.conf

In neither case does the user have `NEO4J_CONF` set in their environment because that is set in the neo4j system service. So there is no way for the tool to figure out where to put the config in any reasonable automatic sense.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+            // The existing setting is enabled, but we are not allowed to overwrite it, so we return the line
+            // unmodified.
+            return line;
+        }
+        // We also need to make sure that what we have found, is not just a mention of the setting in some comment.
+        // We recognise commented out settings (as opposed to mentions of the setting in a comment) by the equals sign
+        // that follows the setting name, and the absence of any prose text before the setting name.
+        int endOfSettingIndex = settingIndex + name.length();
+        int equalsSign = line.indexOf( '=', endOfSettingIndex );
+        if ( equalsSign == -1 || containsProse( line, endOfSettingIndex, equalsSign ) ||
```
spacecowboy: Why do you differentiate between different kinds of comments?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+        int settingIndex = line.indexOf( name );
+        if ( settingIndex == -1 )
+        {
+            // The setting we're looking for is not present in this line, so we just return it unmodified.
+            return line;
+        }
+        // Otherwise, we need to modify this line, such that it reflects the new value.
+        // Except, if 'overwrite' is false, then we don't modify the line if the setting is not commented out.
+        int firstComment = line.indexOf( '#' );
+        if ( !overwrite && (settingIndex < firstComment || firstComment == -1 || alreadyDefined.test( name )) )
```
spacecowboy: I'm confused about `alreadyDefined` in this statement. Why is it needed? And if it is needed, then why is it `OR`d with `!overwrite`?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommandProvider.java
```diff
@@ -0,0 +1,87 @@
...
+                "can be expected to be dedicated to Neo4j.%n" +
+                "%n" +
+                "By default, the command will print its recommendations to the console in a human readable form. " +
+                "The command can alternatively print what the neo4j.conf file would look like with the new settings " +
+                "by specifying --print-conf, and those updated settings can be written back to the conf file " +
+                "(updated in-place) by specifying --update-conf. Any existing settings already enabled in the file " +
+                "will not be updated unless --overwrite is specified.%n" +
+                "%n" +
+                "If --print-conf or --update-conf is specified, then the command will by default look for, and work " +
+                "on, the standard neo4j.conf file. A different file can be specified with the --conf-file argument." );
```
spacecowboy: See objection to "standard neo4j.conf file" above

community/dbms/src/test/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommandTest.java
```diff
@@ -0,0 +1,433 @@
...
+                "# A comment.",
+                initialHeapSize.name() + "=" + heap + " # commented out setting",
+                maxHeapSize.name() + "=" + heap + " # commented out setting without value",
+                pagecache_memory.name() + "=" + pagecache
+        ) );
+        verify( outsideWorld ).stdErrLine( "File updated: " + confFile );
+    }
+
+    @Test
+    public void mustNotDefineSettingsMoreThanOnce() throws Exception
```
spacecowboy: Plus 19 on this test 👍

community/dbms/src/test/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommandTest.java
```diff
@@ -0,0 +1,433 @@
...
+        assertThat( recommendHeapMemory( gibiBytes( 1920 ) ), between( gibiBytes( 31 ), gibiBytes( 33 ) ) );
+    }
+
+    @Test
+    public void mustRecommendPageCacheMemory()
+    {
+        assertThat( recommendPageCacheMemory( mebiBytes( 100 ) ), between( mebiBytes( 95 ), mebiBytes( 105 ) ) );
+        assertThat( recommendPageCacheMemory( gibiBytes( 1 ) ), between( mebiBytes( 95 ), mebiBytes( 105 ) ) );
+        assertThat( recommendPageCacheMemory( gibiBytes( 3 ) ), between( mebiBytes( 470 ), mebiBytes( 530 ) ) );
+        assertThat( recommendPageCacheMemory( gibiBytes( 6 ) ), between( mebiBytes( 980 ), mebiBytes( 1048 ) ) );
```
spacecowboy, line 83: How did you come up with the particular numbers? (I ask because no 2 people seem to recommend the same thing)

spacecowboy (changes requested):
Excellent amount of tests.

How does it actually look when you run it? Could you copy paste the output of 

`neo4j-admin help`
`neo4j-admin help memrec`
`neo4j-admin memrec`

spacecowboy:
I've changed the labels on the PR so it ends up in the correct place in the change log

community/dbms/src/test/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommandTest.java
chrisvest: I meditated on it for a bit, and then wrote down what felt reasonable.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
chrisvest: In `neo4j.conf` we have a passage that look like this:

```# The legacy policy is a special policy which is not defined in
# the policy configuration section, but rather derives from
# dbms.directories.certificates and associated files
# (by default: neo4j.key and neo4j.cert). Its use will be deprecated.
```

If a similar passage gets introduced by us or by an admin, for the settings that we are interested in here, then we should avoid clobbering it.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
chrisvest: Lucene and Netty do their own native memory stuff, so since that is neither heap nor pagecache, I account them in the OS bucket.

"oops" is "ordinary object pointer", and "compressed oops" is a JVM feature where object references are represented with 32 bits instead of 64 bits, when the heap is bellow a certain threshold.

Let me see if I can add some references for that.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
spacecowboy: well sure, but we should just avoid modifying comments period. So I don't see the reason why we need to differentiate between different kinds of comments. Just ignore all lines where first non-whitespace character is `#`.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
spacecowboy: OK. Then I just think you should mention that in the comment or somewhere else reasonable. It's one of the biggest surprises about Neo4js memory management and entirely non-obvious for 99% of users/devs.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+                        "memory settings, unless they are already specified. " +
+                        "Specify --overwrite to also update existing settings." ) )
+                .withArgument( new OptionalBooleanArg( ARG_OVERWRITE, false,
+                        "Causes --print-conf and --update-conf to overwrite existing settings, if set." ) )
+                .withArgument( new OptionalNamedArg( ARG_CONF_FILE, "conf" + File.separator + "neo4j.conf", "",
+                        "An alternative path to the configuration file, for use with --print-conf and --update-conf. " +
+                        "If left unspecified, the default neo4j.conf file location will be used." ) );
+    }
+
+    static String bytesToString( long bytes )
```
MishaDemianenko: we already have an almost identical function in ByteUnit, any chances of using that instead of introducing new?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+
+    @Override
+    public void execute( String[] args ) throws IncorrectUsage, CommandFailed
+    {
+        Arguments arguments = buildArgs().parse( args );
+
+        Path confFile = arguments.getOptionalPath( ARG_CONF_FILE )
+                                 .orElse( configDir.resolve( Config.DEFAULT_CONFIG_FILE_NAME ) );
+        String memoryArgument = arguments.get( ARG_MEMORY );
+        long memory = buildSetting( ARG_MEMORY, BYTES, null ).build().apply( arguments::get );
```
MishaDemianenko: we have a `buildSetting` function that does not require ugly null parameter

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+        {
+            out.accept( setting + "=" + value );
+        }
+    }
+
+    private void printRecommendation( String memoryArgument, String heap, String pagecache )
+    {
+        outsideWorld.stdOutLine(
+                "Based on the available memory, " + memoryArgument + ", and assuming the system is" +
+                " dedicated to running Neo4j,\nthe following memory settings are recommended:" );
```
MishaDemianenko: please use `System.lineSeparator()` instead of `\n` if you want to add line separator here

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+import static org.neo4j.io.ByteUnit.mebiBytes;
+import static org.neo4j.io.ByteUnit.tebiBytes;
+import static org.neo4j.kernel.configuration.Settings.BYTES;
+import static org.neo4j.kernel.configuration.Settings.buildSetting;
+
+public class MemoryRecommendationsCommand implements AdminCommand
+{
+    // Fields: {System Memory in GiBs; OS memory reserve in GiBs; JVM Heap memory in GiBs}.
+    // And the page cache gets what's left, though always at least 100 MiB.
+    // Heap never goes beyond 32 GiBs, which is the upper limit for compressed oops..
```
MishaDemianenko: We talked already earlier today about heaps that are bigger than 32g, i think we need to mention somewhere that our recommendations are for heaps lover than 32g. 
Also based on previously observed db's config values would be good to validate user defined 
values for common misconfigurations (like 33g heaps, for example).

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
MishaDemianenko: I feel the pain here, do you think it will gonna improve the situation if we will gonna mention that in comments for memory related settings in default config template file?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+            printRecommendation( memoryArgument, heap, pagecache );
+        }
+
+        if ( updateConf )
+        {
+            ArrayList<String> updatedLines = new ArrayList<>();
+            outputUpdatedConf( confFile, heap, pagecache, overwrite, updatedLines::add );
+            try
+            {
+                Files.write( confFile, updatedLines, StandardCharsets.UTF_8 );
```
MishaDemianenko: shouldn't we keep a backup of the original config file first and not simply override?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+            defineIfMissing( out, redefined, maxHeapSize.name(), heap );
+            defineIfMissing( out, redefined, pagecache_memory.name(), pagecache );
+        }
+        catch ( Exception e ) // Also catch runtime exceptions due to use of Streams.
+        {
+            throw new CommandFailed( "Failed to read configuration file: " + confFile, e );
+        }
+    }
+
+    private String tryUpdateSetting( String line, Predicate<String> alreadyDefined, Set<String> redefined,
```
MishaDemianenko: this method and related trying to do a bunch of work by parsing and trying to understand user config file, in the same time it will fail to do so in some of the cases that we support during normal config loading since the method does not cover all possible format possibilities.
For example, multi-line lines or `:` as a separator just to name a few, so maybe instead of doing this attempt we will perform additional validation of already configured values or generate a suggested property values, so users can update config file if they think its relevant for them.
What do you guys think?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,385 @@
...
+        return brackets.recommend( Bracket::osMemory );
+    }
+
+    static long recommendHeapMemory( long totalMemoryBytes )
+    {
+        Brackets brackets = findMemoryBrackets( totalMemoryBytes );
+        return brackets.recommend( Bracket::heapMemory );
+    }
+
+    static long recommendPageCacheMemory( long totalMemoryBytes )
```
MishaDemianenko, line 83: We have a heuristic currently that tries to determine correct page cache size in case if that was not configured explicitly. We need to merge those two things and have only one "recommender"

MishaDemianenko (commented):
Have some of the comments, sorry if those were discussed already

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
spacecowboy: We should mention it everywhere we talk about different buckets of memory

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
spacecowboy: Personally I'm all for it. I think it would be sensible to do 

```
sudo neo4j-admin memrec >> /etc/neo4j/neo4j.conf
```

but I know support will really like an in-place update feature.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
chrisvest: To be clear, the "it" we want to mention is the fact that Lucene and Netty have their own native memory usage, right?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
chrisvest: It uses the wrong unit names. I would have merged them had this been in 3.4.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
chrisvest: Or it should be two separate calls to `stdOutLine`.

MishaDemianenko: absolutely, any of those that till make new line correct on different platforms

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
spacecowboy: yes.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
chrisvest: I think we should rely on pipes on the command line instead of touching files at all.

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
chrisvest: And then instead just append the missing settings to the end of the file?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
MishaDemianenko: maybe we can have a different variant of that function there something like `bytesToString( long bytes, Function<Unit,String> unitNameMapper) `? (also its used on in one internal place in 3.3 so I think we more than safe to update it in one way or another in 3.3 as well)

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
MishaDemianenko: Fantastic, all my files related comments closed in this case then 🥋

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
chrisvest: Taking your example literally, I think we can all find a solution that can satisfy us all:

```
sudo neo4j-admin memrec >> /etc/neo4j/neo4j.conf
```

All this needs to do, really, is just produce a _section_ of configuration file text, that can be appended to the end of an existing (or not) configuration file. There is no need parse any existing file. We just have to produce output in a format that works in a configuration file.

spacecowboy: my original example was incorrect. I naturally meant

    neo4j-admin memrec | sudo tee -a /etc/neo4j/neo4j.conf

But yes. The way the config *should* work is that the last defined value is used in case of duplicate values, so there should not be any need to change anything in the file.

That is of course how it *should* be. I know Java does that, I am less certain about our wrapper scripts parsing of the config. That's a separate issue though.

chrisvest:
@spacecowboy @MishaDemianenko I believe your comments have been addressed. Please take another look.

spacecowboy (approved):
I like the output a lot! 👍 

```
$ bin/neo4j-admin memrec --memory=5G
# Memory settings recommendation from neo4j-admin memrec:
#
# Assuming the system is dedicated to running Neo4j and has 5G of memory,
# we recommend a heap size of around 2600m, and a page cache of around 768m,
# and that about 1800m is left for the operating system, and the native memory
# needed by Lucene and Netty.
#
# Tip: If the indexing storage use is high, e.g. there are many indexes or most
# data indexed, then it might advantageous to leave more memory for the
# operating system.
#
# Tip: The more concurrent transactions your workload has and the more updates
# they do, the more heap memory you will need. However, don't allocate more
# than 31g of heap, since this will disable pointer compression, also known as
# "compressed oops", in the JVM and make less effective use of the heap.
#
# Tip: Setting the initial and the max heap size to the same value means the
# JVM will never need to change the heap size. Changing the heap size otherwise
# involves a full GC, which is desirable to avoid.
#
# Based on the above, the following memory settings are recommended:
dbms.memory.heap.initial_size=2600m
dbms.memory.heap.max_size=2600m
dbms.memory.pagecache.size=768m
```

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,267 @@
...
+import static org.neo4j.kernel.configuration.Settings.BYTES;
+import static org.neo4j.kernel.configuration.Settings.buildSetting;
+
+public class MemoryRecommendationsCommand implements AdminCommand
+{
+    // Fields: {System Memory in GiBs; OS memory reserve in GiBs; JVM Heap memory in GiBs}.
+    // And the page cache gets what's left, though always at least 100 MiB.
+    // Heap never goes beyond 31 GiBs, which is the upper limit for compressed oops.
+    // More details on heap sizing wrt. compressed oops:
+    // https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html
```
MishaDemianenko: no 🙈 , let's talk with docs team about our own page about that kind of problems instead?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
MishaDemianenko: As I can see one part of the recommendation is gone, but heuristic that calculates size still there. Shouldn't our recommended value and heuristic evaluated be the same and use the same code for evaluation?

community/dbms/src/main/java/org/neo4j/commandline/dbms/MemoryRecommendationsCommand.java
```diff
@@ -0,0 +1,265 @@
...
+import static org.neo4j.io.ByteUnit.ONE_MEBI_BYTE;
+import static org.neo4j.io.ByteUnit.gibiBytes;
+import static org.neo4j.io.ByteUnit.mebiBytes;
+import static org.neo4j.io.ByteUnit.tebiBytes;
+import static org.neo4j.kernel.configuration.Settings.BYTES;
+import static org.neo4j.kernel.configuration.Settings.buildSetting;
+
+public class MemoryRecommendationsCommand implements AdminCommand
+{
+    // Fields: {System Memory in GiBs; OS memory reserve in GiBs; JVM Heap memory in GiBs}.
```
jakewins, line 47: Note that Netty allocates off-heap RAM as well:

- there is no actual upper bound, it will allocate until malloc fails
- it is lazy, it allocates 16MiB chunks in "arenas:, by default there are `numCpu * 2` arenas, all with one chunk each

TL;DR: Assume Netty allocates 16 * numCPU * 2 MiB at bootstrap, and then unbound additional amounts until we modify the Causal Cluster and Bolt input code to use fixed size I/O buffers.

See https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java

PR #10824, opened 2018-01-16
Adding reference counting and tracing to Cursors

SaschaPeukert:


community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Cursor.java
```diff
@@ -40,7 +40,7 @@
  * }
  * </pre></code>
  */
-interface Cursor extends AutoCloseable
+public interface Cursor extends AutoCloseable
```
pontusmelke, line 43: There is a comment above here explaining why the interface is package private, if we change it to public we should also update the comment.

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/TraceCursorTest.java
```diff
@@ -0,0 +1,502 @@
...
+import org.neo4j.internal.kernel.api.exceptions.KernelException;
+import org.neo4j.internal.kernel.api.security.SecurityContext;
+import org.neo4j.values.storable.Values;
+
+import static org.junit.Assert.fail;
+
+public class TraceCursorTest
+{
+    private ReadTestSupport readTestSupport = new ReadTestSupport();
+    private static final TemporaryFolder folder = new TemporaryFolder();
```
pontusmelke: `@Rule`

community/kernel/src/test/java/org/neo4j/kernel/impl/newapi/TraceCursorTest.java
```diff
@@ -0,0 +1,502 @@
...
+    private Kernel kernel;
+    private Session session;
+
+    @Before
+    public void setup()
+    {
+        try
+        {
+            folder.create();
```
pontusmelke: Can't we just rely on `@Rule`

pontusmelke (changes requested):
Nice implementation, nice test coverage!

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
```diff
@@ -49,36 +57,73 @@
...
 
     @Override
-    public final void nodeIndexSeek(
-            org.neo4j.internal.kernel.api.IndexReference index,
+    public void acquireCursor( Cursor cursor )
+    {
+        if ( TRACK_CURSORS )
+        {
+            StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
+            cursorsOpenCloseCalls.put( cursor, Arrays.copyOfRange(stackTrace, 2, stackTrace.length) );
```
chrisvest: Looks like setting `RECORD_CURSOR_TRACES` to `false` will increase the overhead of recording cursor traces, since the traces are captured regardless, and then stored in a cache shared among all threads, and the cache has to evict on all but the first put.

chrisvest: Also, setting `TRACK_CURSORS` to `true` but leaving `RECORD_CURSOR_TRACES` at `false` will only keep a single element of history, which can help with finding bugs but not so much with finding fixes.

SaschaPeukert: So you propose we set the lruCache size in either way to `CURSORS_TRACK_HISTORY_MAX_SIZE`?

chrisvest: Yeah, and then you only need a single feature toggle for the lot.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Read.java
```diff
@@ -49,36 +57,73 @@
...
+        org.neo4j.internal.kernel.api.SchemaRead,
+        CursorTracker
 {
     private final Cursors cursors;
     final KernelTransactionImplementation ktx;
 
+    private long cursorsInUse;
+
+    private static final boolean TRACK_CURSORS = flag( Read.class, "trackCursors", false );
+    private static final boolean RECORD_CURSORS_TRACES = flag( KernelStatement.class, "recordCursorTraces", false );
```
fickludd: This flag is on `KernelStatement` =)

pontusmelke (approved):

pontusmelke:
@SaschaPeukert something is wrong, basically all tests fail?

fickludd:
This functionality did not work out as wanted, so we are closing this PR. We might later come back to this branch and salvage the code for automatically closing all cursors on transaction close.

PR #10825, opened 2018-01-17
Solve nested unwind bug in compiled runtime and add some feature tests.

Lojjs:
To be forward merged to 3.3 and 3.4

### DIFF
--- a/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
+++ b/community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
@@ -35,6 +35,7 @@ import java.util.stream.Collectors;
 import java.util.stream.DoubleStream;
 import java.util.stream.IntStream;
 import java.util.stream.LongStream;
+import java.util.stream.Stream;
 
 import org.neo4j.cypher.internal.compiler.v3_2.spi.NodeIdWrapper;
 import org.neo4j.cypher.internal.compiler.v3_2.spi.RelationshipIdWrapper;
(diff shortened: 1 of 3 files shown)


community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
```diff
@@ -416,30 +428,15 @@ else if ( iterable == null )
         {
             return Collections.emptyIterator();
         }
+        else if ( iterable instanceof String )
+        {
+            String[] singleStringArray = {(String) iterable};
```
sherfert: You can use Stream.of( iterable ).iterator()

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/UnwindAcceptance.feature
```diff
@@ -228,3 +228,57 @@ Feature: UnwindAcceptance
...
+      | 3.5 |
+      | 4.5 |
+      | 5.5 |
+    And no side effects
+
+  Scenario: Nested unwind with strings
+    Given an empty graph
+    When executing query:
+      """
+      WITH [['a', 'b'], ['c', 'd'], 'e'] AS nested
```
craigtaverner, line 271: Should also work with mixed types

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
```diff
@@ -400,18 +401,35 @@ else if ( iterable instanceof PrimitiveEntityStream )
...
         }
+        else if ( iterable instanceof Integer)
+        {
+            return (IntStream.of((Integer) iterable)).iterator();
+        }
         else if ( iterable instanceof IntStream )
         {
             return ((IntStream) iterable).iterator();
         }
+        else if ( iterable instanceof String )
```
craigtaverner: Is it necessary to have type specific checks here? Perhaps just `Stream.of(iterable)` in the `else` clause would be simpler? What does the interpreted runtime do in this case?

craigtaverner (changes requested):

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
sherfert: Usually the primitive streams are more performant. But then again, a stream with just one element won't be a bottleneck...

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/UnwindAcceptance.feature
Lojjs: Good point, I have added a new test for that.

craigtaverner (approved):

PR #10829, opened 2018-01-17
Cleanup java-driver deprecated usage

MishaDemianenko:
Switch bookmarks usage to be session based instead of transaction based


MishaDemianenko:
@lutovich can you please take a 👀 , thx!

lutovich (approved):
Looks good!

PR #10830, opened 2018-01-17
Resumable store copy

phughk:
This PR introduces resumable store copy in the sense that failed store file download will be retried. Previously all store files would be streamed to the client in one response and if any issue would occur we would have to redo the entire process. The new strategy is as follows:

CLIENT -> Prepare store request
SERVER -> Grabs checkpoint lock and streams all "atomic-files" (count store, legacy index and additional providers). It then returns a response that includes last committed transaction id, a list of store file names, a list of lucene index descriptors.
CLIENT -> for each store file: send get store file request
SERVER -> stream requested file
CLIENT -> for each descriptor: send get lucene snapshot request
SERVER -> stream lucene snapshot files.

If prepare store request fails then store copy fails. Each get file/snapshot request may be retried as long as the configured termination condition allows.

Each file/snapshot request include the "last committed tx id" given by the prepare store copy respons so we can try to download from any instance in the cluster. In this PR all requests are to leader.

As of yet, nothing has been done to "pull tx" which is called after the store files has been copied. This will be investigated in a different PR.

KERNEL:
NeoStoreFileListing have been modified to provide builders for listing files. IndexService has been modified to expose descriptors and to allow getting specific snapshot for descriptor.

PrepareStoreCopyRequestHandler, GetStoreFileRequestHandler and GetLuceneSnapshotRequestHandler all use NeoStoreFileListing and would be good looking into.


phughk:
Team-kernel tag added as it would be good to have confirmation that locks are good and all kernel components are used correctly.

chrisvest:
Just letting you know that I've started reviewing this. It may take some time before I've made my way through it.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/CatchUpResponseAdaptor.java
```diff
@@ -35,43 +36,54 @@
...
+        fail( signal, response );
+    }
+
+    @Override
+    public void onStoreListingResponse( CompletableFuture<T> signal, PrepareStoreCopyResponse response )
+    {
+        fail( signal, response );
+    }
+
+    private <U> void fail( CompletableFuture<T> signal, U response )
```
chrisvest: Maybe call this something else to avoid confusing it with the JUnit `fail` method.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -808,9 +809,36 @@ public InwardKernel getKernel()
         return kernelModule.kernelAPI();
     }
 
-    public ResourceIterator<StoreFileMetadata> listStoreFiles( boolean includeLogs ) throws IOException
+    public ResourceIterator<StoreFileMetadata> listAllStoreFiles( boolean includeLogs ) throws IOException
```
chrisvest: I'm not convinced about the value of this rename.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -808,9 +809,36 @@ public InwardKernel getKernel()
...
+        }
+        else
+        {
+            return listStoreFilesExclusively().excludeLogFiles().build();
+        }
+    }
+
+    public NeoStoreFileListing.ExclusiveFileLister listStoreFilesExclusively()
+    {
+        return getNeoStoreFileListing().exclusiveBuilder();
```
chrisvest: Do we really need two builders?

chrisvest: You could add `includeAll` / `excludeAll` methods to the builder.

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
```diff
@@ -808,9 +809,36 @@ public InwardKernel getKernel()
...
+    {
+        return getNeoStoreFileListing().inclusiveBuilder();
+    }
+
+    private NeoStoreFileListing getNeoStoreFileListing()
+    {
+        return kernelModule.fileListing();
+    }
+
+    public NeoStoreFileIndexListing getNeoStoreIndexListing()
```
chrisvest: Why are the indexes separate in this way?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+        else if ( transactionTooFarBehind )
+        {
+            storeFileStreamingProtocol.end( ctx, StoreCopyFinishedResponse.Status.E_TOO_FAR_BEHIND );
+        }
+        else
+        {
+            File storeDir = dataSource.get().getStoreDir();
+            ResourceIterator<StoreFileMetadata> resourceIterator =
+                    dataSource.get().getNeoStoreIndexListing().getSnapshot( snapshotRequest.descriptor() );
```
chrisvest: Contrary to the name of this class, this will also get native indexes.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -634,16 +637,36 @@ private void closeAllIndexes()
         } );
     }
 
-    public ResourceIterator<File> snapshotStoreFiles() throws IOException
+    public Set<Descriptor> getIndexDescriptors()
```
chrisvest: These are not index descriptors, but schema index provider descriptors. Why do we need the schema index provider descriptors? Do you want to resume per provider?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -634,16 +637,36 @@ private void closeAllIndexes()
...
+        Set<SchemaIndexProvider.Descriptor> fromProviders = new HashSet<>();
+        for ( IndexProxy indexProxy : indexMapRef.getAllIndexProxies() )
+        {
+            Descriptor providerDescriptor = indexProxy.getProviderDescriptor();
+            fromProviders.add( providerDescriptor );
+        }
+        return fromProviders;
+    }
+
+    public ResourceIterator<File> snapshotStoreFiles( Predicate<Descriptor> filter ) throws IOException
```
chrisvest: I'd say that these are index files rather than store files.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/MultiResource.java
```diff
@@ -0,0 +1,55 @@
...
+            }
+            catch ( RuntimeException e )
+            {
+                exception = e;
+            }
+        }
+        if ( exception != null )
+        {
+            throw exception;
+        }
```
chrisvest: `IOUtils.close`

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/MultiResource.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package org.neo4j.kernel.impl.transaction.state;
+
+import java.util.Collection;
+
+import org.neo4j.graphdb.Resource;
+
+public final class MultiResource implements Resource
+{
+    private final Collection<? extends Resource> snapshots;
```
chrisvest: These are not snapshots, but resources.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/MultiResource.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.transaction.state;
```
chrisvest: I think `org.neo4j.kernel.impl.util` is a better home. This thing is pretty generic.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -144,6 +142,185 @@ private void gatherNonRecordStores( Collection<StoreFileMetadata> files, boolean
...
+    {
+        ExclusiveFileLister excludeLogFiles();
+        ExclusiveFileLister excludeNonRecordStoreFiles();
+        ExclusiveFileLister excludeNeoStoreFiles();
+        ExclusiveFileLister excludeLabelScanStoreFiles();
+        ExclusiveFileLister excludeSchemaIndexStoreFiles();
+        ExclusiveFileLister excludeExplicitIndexStoreFiles();
+        ExclusiveFileLister excludeAdditionalProviders();
+        ResourceIterator<StoreFileMetadata> build() throws IOException;
+    }
```
chrisvest: Can't we get away with one builder, and `excludeAll` / `includeAll` methods? I also think there's no need for the interfaces, when you can expose a `final` class in this case.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -144,6 +142,185 @@ private void gatherNonRecordStores( Collection<StoreFileMetadata> files, boolean
...
+            {
+                for ( StoreFileProvider additionalProvider : additionalProviders )
+                {
+                    resources.add( additionalProvider.addFilesTo( files ) );
+                }
+            }
+
+            placeMetaDataStoreLast( files );
+
+            return resourceIterator( files.iterator(), new MultiResource( resources ) );
```
chrisvest: You have two identical `MultiResource` classes. The one embedded in this class, and the one you copied for the indexes. Please remove one.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -634,16 +637,36 @@ private void closeAllIndexes()
...
     {
         Collection<ResourceIterator<File>> snapshots = new ArrayList<>();
         for ( IndexProxy indexProxy : indexMapRef.getAllIndexProxies() )
         {
-            snapshots.add( indexProxy.snapshotFiles() );
+            Descriptor providerDescriptor = indexProxy.getProviderDescriptor();
+            if ( filter.test( providerDescriptor ) )
+            {
+                snapshots.add(  indexProxy.snapshotFiles() );
```
chrisvest: Formatting.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileIndexListing.java
```diff
@@ -0,0 +1,104 @@
...
+
+import static org.neo4j.helpers.collection.Iterators.resourceIterator;
+
+public class NeoStoreFileIndexListing
+{
+    private final LabelScanStore labelScanStore;
+    private final IndexingService indexingService;
+    private final ExplicitIndexProviderLookup explicitIndexProviders;
+
+    private final Function<File,StoreFileMetadata> toNotAStoreTypeFile = file -> new StoreFileMetadata( file, RecordFormat.NO_RECORD_SIZE );
```
chrisvest: This can be `static`. I think the `NeoStoreFileIndexingListing` gives enough context that we know these are not store files. The field could instead say what it does rather than what it does not, e.g. `toStoreFileMetadata`.

chrisvest: Isn't this line too long for our code style?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileIndexListing.java
```diff
@@ -0,0 +1,104 @@
...
+            files.addAll( getSnapshotFilesMetadata( snapshot ) );
+        }
+        // Intentionally don't close the snapshot here, return it for closing by the consumer of
+        // the targetFiles list.
+        return new MultiResource( snapshots );
+    }
+
+    private List<StoreFileMetadata> getSnapshotFilesMetadata( ResourceIterator<File> snapshot )
+    {
+        return snapshot.stream().map( toNotAStoreTypeFile ).collect( Collectors.toList() );
```
chrisvest: You could also take the target `Collection<StoreFileMetadata>` as a parameter and use `forEach( target::add )` to avoid copying via the `toList` collector.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
chrisvest: Also, we never need the information that this is a set. It's turned into a Collection by the caller.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/CloseablesListener.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.neo4j.helpers.Exceptions;
+
+class CloseablesListener implements AutoCloseable, GenericFutureListener<Future<Void>>
+{
+    private final List<AutoCloseable> closeables = new ArrayList<>();
+
+    <T extends AutoCloseable> T sample( T closeable )
```
chrisvest: Why is this called `sample` and not `add`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/CloseablesListener.java
```diff
@@ -0,0 +1,77 @@
...
+                if ( exception != e )
+                {
+                    exception = Exceptions.combine( exception, e );
+                }
+            }
+        }
+        if ( exception != null )
+        {
+            throw exception;
+        }
```
chrisvest: `IOUtils.close`

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
```diff
@@ -0,0 +1,134 @@
...
+                    StoreFileMetadata storeFileMetadata = resourceIterator.next();
+                    storeFileStreamingProtocol.stream( ctx,
+                            new StoreResource( storeFileMetadata.file(), relativePath( storeDir, storeFileMetadata.file() ), storeFileMetadata.recordSize(),
+                                    pageCache, fs ) );
+                }
+                storeFileStreamingProtocol.end( ctx, StoreCopyFinishedResponse.Status.SUCCESS ).addListener( closeablesListener );
+            }
+            catch ( Throwable e )
+            {
+                storeFileStreamingProtocol.end( ctx, StoreCopyFinishedResponse.Status.E_UNKNOWN ).addListener( closeablesListener );
```
chrisvest: You are dropping the `Throwable` on the floor, here. How about keeping the `StoreCopyFinishedResponse.Status` in a local variable, initialised to `E_UNKNOWN`, and then move `storeFileStreamingProtocol.end(…)` to the `finally` clause?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
```diff
@@ -0,0 +1,134 @@
...
+import org.neo4j.kernel.NeoStoreDataSource;
+import org.neo4j.kernel.impl.transaction.log.checkpoint.CheckPointer;
+import org.neo4j.kernel.impl.transaction.log.checkpoint.SimpleTriggerInfo;
+import org.neo4j.logging.Log;
+import org.neo4j.logging.LogProvider;
+import org.neo4j.storageengine.api.StoreFileMetadata;
+
+import static org.neo4j.io.fs.FileUtils.relativePath;
+
+public class GetLuceneSnapshotRequestHandler extends SimpleChannelInboundHandler<GetLuceneSnapshotRequest>
```
chrisvest: This should be `GetIndexSnapshotRequestHandler`, and likewise for the `*Request` etc.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
```diff
@@ -0,0 +1,134 @@
...
+        this.storeFileStreamingProtocol = storeFileStreamingProtocol;
+        this.pageCache = pageCache;
+        this.fs = fs;
+        this.log = logProvider.getLog( GetLuceneSnapshotRequestHandler.class );
+    }
+
+    @Override
+    protected void channelRead0( ChannelHandlerContext ctx, GetLuceneSnapshotRequest snapshotRequest ) throws Exception
+    {
+        log.debug( "Requesting Lucene snapshot %s", snapshotRequest );
```
chrisvest: _index_ snapshot.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
```diff
@@ -0,0 +1,134 @@
...
+                    dataSource.get().getNeoStoreIndexListing().getSnapshot( snapshotRequest.descriptor() );
+            try
+            {
+                closeablesListener.sample( resourceIterator );
+                while ( resourceIterator.hasNext() )
+                {
+                    StoreFileMetadata storeFileMetadata = resourceIterator.next();
+                    storeFileStreamingProtocol.stream( ctx,
+                            new StoreResource( storeFileMetadata.file(), relativePath( storeDir, storeFileMetadata.file() ), storeFileMetadata.recordSize(),
+                                    pageCache, fs ) );
```
chrisvest: I think you could extract some local variables to shorten this line.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/CatchupAddressProvider.java
```diff
@@ -0,0 +1,114 @@
...
+import org.neo4j.causalclustering.core.consensus.RaftMachine;
+import org.neo4j.causalclustering.discovery.TopologyService;
+import org.neo4j.causalclustering.identity.MemberId;
+import org.neo4j.helpers.AdvertisedSocketAddress;
+
+public interface CatchupAddressProvider
+{
+    AdvertisedSocketAddress primary() throws CatchupAddressResolutionException;
+
+    AdvertisedSocketAddress secondary() throws CatchupAddressResolutionException;
```
chrisvest: Some javadocs on the interface and methods would be helpful, I think.

chrisvest: Meanwhile, I don't find any usages of the `secondary()` method. Can it be removed (making this a functional interface in the process) or is there some functionality that isn't implemented yet? Or is in the commercial repo?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/CatchupClientProtocol.java
```diff
@@ -35,6 +35,8 @@ public CatchupClientProtocol()
         STORE_COPY_FINISHED,
         TX_STREAM_FINISHED,
         FILE_HEADER,
+        PREPARE_STORE_COPY_RESPONSE,
+        LUCENE_SNAPSHOT_RESPONSE,
```
chrisvest: `s/LUCENE/INDEX/`

Is adding/changing these enums a protocol change? Does it involve some versioning?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/CatchupServerProtocol.java
```diff
@@ -28,6 +28,12 @@ public CatchupServerProtocol()
 
     public enum State
     {
-        MESSAGE_TYPE, GET_STORE, GET_STORE_ID, GET_CORE_SNAPSHOT, TX_PULL
+        MESSAGE_TYPE,
+        PREPARE_STORE_COPY,
+        GET_STORE_ID,
+        GET_CORE_SNAPSHOT,
+        GET_STORE_FILE,
+        GET_LUCENE_SNAPSHOT,
```
chrisvest: `s/LUCENE/INDEX/`

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/RequestMessageType.java
```diff
@@ -26,9 +26,11 @@
 public enum RequestMessageType implements Message
 {
     TX_PULL_REQUEST( (byte) 1 ),
-    STORE( (byte) 2 ),
+    PREPARE_STORE_COPY( (byte) 2 ),
```
chrisvest: This looks like a protocol change? I don't know if that matters.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/RequestMessageType.java
```diff
@@ -26,9 +26,11 @@
...
     CORE_SNAPSHOT( (byte) 3 ),
     STORE_ID( (byte) 4 ),
+    STORE_FILE( (byte) 6 ),
+    LUCENE_SNAPSHOT( (byte) 7 ),
```
chrisvest: `s/LUCENE/INDEX/`

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/ResponseMessageType.java
```diff
@@ -29,7 +29,9 @@
     STORE_COPY_FINISHED( (byte) 4 ),
     CORE_SNAPSHOT( (byte) 5 ),
     TX_STREAM_FINISHED( (byte) 6 ),
-    UNKNOWN( (byte) 200 ),;
+    PREPARE_STORE_COPY_RESPONSE( (byte) 7 ),
+    LUCENE_SNAPSHOT_RESPONSE( (byte) 8 ),
```
chrisvest: `s/LUCENE/INDEX/`

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequest.java
```diff
@@ -0,0 +1,135 @@
...
+            putString( channel, key );
+
+            String version = getLuceneSnapshotRequest.descriptor().getVersion();
+            putString( channel, version );
+        }
+
+        private void putString( WritableChannel channel, String version ) throws IOException
+        {
+            channel.putInt( version.length() );
+            channel.put( version.getBytes(), version.length() );
```
chrisvest: Maybe call it `value` instead of `version`. Also, you are letting the platform environment dictate the string encoding, here. I'd pin that down to UTF-8, so both ends know what to expect.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequest.java
```diff
@@ -0,0 +1,135 @@
...
+    static class LuceneSnapshotRequestMarshall extends SafeChannelMarshal<GetLuceneSnapshotRequest>
+    {
+        @Override
+        protected GetLuceneSnapshotRequest unmarshal0( ReadableChannel channel ) throws IOException, EndOfStreamException
+        {
+            StoreId storeId = StoreIdMarshal.INSTANCE.unmarshal( channel );
+            long requiredTransactionId = channel.getLong();
+            byte[] descriptorKey = getBytes( channel );
+            byte[] descriptorVersion = getBytes( channel );
+            return new GetLuceneSnapshotRequest( storeId, new SchemaIndexProvider.Descriptor( new String( descriptorKey ), new String( descriptorVersion ) ),
```
chrisvest: You are letting the platform environment dictate the encoding in your `new String`s here.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/PrepareStoreCopyResponse.java
```diff
@@ -0,0 +1,230 @@
...
+        public void marshal( PrepareStoreCopyResponse prepareStoreCopyResponse, WritableChannel buffer ) throws IOException
+        {
+            buffer.putInt( prepareStoreCopyResponse.status.ordinal() );
+            buffer.putLong( prepareStoreCopyResponse.transactionId );
+            marshalFiles( buffer, prepareStoreCopyResponse.files );
+            marshalIndexDescriptors( buffer, prepareStoreCopyResponse.indexDescriptors );
+        }
+
+        @Override
+        protected PrepareStoreCopyResponse unmarshal0( ReadableChannel channel ) throws IOException, EndOfStreamException
```
chrisvest: `EndOfStreamException` is never thrown in this method, as far as I can see.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/PrepareStoreCopyResponse.java
```diff
@@ -0,0 +1,230 @@
...
+            SchemaIndexProvider.Descriptor[] indexDescriptors = unmarshalIndexDescriptors( channel );
+            return new PrepareStoreCopyResponse( files, indexDescriptors, transactionId, status );
+        }
+
+        private static void marshalFiles( WritableChannel buffer, File[] files ) throws IOException
+        {
+            buffer.putInt( files.length );
+            for ( File file : files )
+            {
+                byte[] bytes = file.getName().getBytes();
```
chrisvest: Give `getBytes` a `Charset`, for instance `StandardCharsets.UTF_8`, to avoid the platform encoding dictating what this produces.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/PrepareStoreCopyResponse.java
```diff
@@ -0,0 +1,230 @@
...
+        private void marshalIndexDescriptors( WritableChannel buffer, SchemaIndexProvider.Descriptor[] descriptors ) throws IOException
+        {
+            buffer.putInt( descriptors.length );
+            for ( SchemaIndexProvider.Descriptor descriptor : descriptors )
+            {
+                byte[] ketBytes = descriptor.getKey().getBytes();
+                buffer.putInt( ketBytes.length );
+                buffer.put( ketBytes, ketBytes.length );
+
+                byte[] versionBytes = descriptor.getVersion().getBytes();
```
chrisvest: Same encoding issue for key and version.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetStoreFileRequest.java
```diff
@@ -0,0 +1,119 @@
...
+    {
+        @Override
+        protected GetStoreFileRequest unmarshal0( ReadableChannel channel ) throws IOException, EndOfStreamException
+        {
+            StoreId storeId = StoreIdMarshal.INSTANCE.unmarshal( channel );
+            long requiredTransactionId = channel.getLong();
+            int fileNameLength = channel.getInt();
+            byte[] fileNameBytes = new byte[fileNameLength];
+            channel.get( fileNameBytes, fileNameLength );
+            return new GetStoreFileRequest( storeId, new File( new String( fileNameBytes ) ), requiredTransactionId );
```
chrisvest: `fileNameBytes` is interpreted according to platform encoding. Give it a `StandardCharsets.UTF_8` to pin it down.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetStoreFileRequest.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        @Override
+        public void marshal( GetStoreFileRequest getStoreFileRequest, WritableChannel channel ) throws IOException
+        {
+            StoreIdMarshal.INSTANCE.marshal( getStoreFileRequest.expectedStoreId(), channel );
+            channel.putLong( getStoreFileRequest.requiredTransactionId() );
+            String name = getStoreFileRequest.file().getName();
+            channel.putInt( name.length() );
+            channel.put( name.getBytes(), name.length() );
```
chrisvest: Platform encoding again, on `name.getBytes()`.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/PrepareStoreCopyResponse.java
```diff
@@ -0,0 +1,230 @@
...
+            }
+            return files;
+        }
+
+        private static File unmarshalFile( ReadableChannel channel ) throws IOException
+        {
+            int characters = channel.getInt();
+            byte[] name = new byte[characters];
+            channel.get( name, characters );
+            return new File( new String( name ) );
```
chrisvest: Again with the platform encoding.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/PrepareStoreCopyResponse.java
```diff
@@ -0,0 +1,230 @@
...
+        private SchemaIndexProvider.Descriptor unmarshalIndexDescriptor( ReadableChannel channel ) throws IOException
+        {
+            int keyCharacters = channel.getInt();
+            byte[] key = new byte[keyCharacters];
+            channel.get( key, keyCharacters );
+
+            int versionCharacters = channel.getInt();
+            byte[] version = new byte[versionCharacters];
+            channel.get( version, versionCharacters );
+            return new SchemaIndexProvider.Descriptor( new String( key ), new String( version ) );
```
chrisvest: Again with the platform encoding.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/PrepareStoreCopyResponse.java
```diff
@@ -0,0 +1,230 @@
...
+
+        private static File unmarshalFile( ReadableChannel channel ) throws IOException
+        {
+            int characters = channel.getInt();
+            byte[] name = new byte[characters];
+            channel.get( name, characters );
+            return new File( new String( name ) );
+        }
+
+        private SchemaIndexProvider.Descriptor[] unmarshalIndexDescriptors( ReadableChannel channel ) throws IOException
```
chrisvest: I'd call these index provider descriptors.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClient.java
```diff
@@ -19,63 +19,132 @@
...
     public StoreCopyClient( CatchUpClient catchUpClient, LogProvider logProvider )
     {
         this.catchUpClient = catchUpClient;
         log = logProvider.getLog( getClass() );
+        this.logProvider = logProvider;
     }
 
-    public long copyStoreFiles( AdvertisedSocketAddress from, StoreId expectedStoreId, StoreFileStreams storeFileStreams ) throws StoreCopyFailedException
+    public long copyStoreFiles( CatchupAddressProvider catchupAddressProvider, StoreId expectedStoreId, StoreFileStreams storeFileStreams,
+            Supplier<TerminationCondition> requestWiseTerminationCondition )
```
chrisvest: The `StoreCopyClient` is always created together with `RemoteStore`, which is the only class that calls this method and passes  a `requestWiseTerminationCondition`. I wonder if this argument could be made a field instead, since this argument list is already quite long. And perhaps `RemoteStore` could be put in charge of creating the `StoreCopyClient`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/PrepareStoreCopyResponseAdaptor.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Used client side for event handling of a store listing request
+ */
+public class PrepareStoreCopyResponseAdaptor extends CatchUpResponseAdaptor<PrepareStoreCopyResponse>
+{
+    private final StoreFileStreams storeFileStreams;
+    private final Log log;
+    private String destination;
+    private int requiredAlignment;
+    private Set<String> debugCopiedFiles = new HashSet<>();
```
chrisvest: I don't think we need this field.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/TerminationCondition.java
```diff
@@ -0,0 +1,34 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.causalclustering.catchup.storecopy;
+
+interface TerminationCondition
+{
+    void assertContinue() throws StoreCopyFailedException;
```
chrisvest: This can be annotated as a functional interface. It would be good to add some javadocs.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/TerminationCondition.java
```diff
@@ -0,0 +1,34 @@
...
+
+    class ContinueIndefinitely implements TerminationCondition
```
chrisvest: This class is only used in tests. It could be inlined as lambdas there, or you could have a static method in the interface, that returns a do-nothing lambda implementation. That way you won't need to name and expose a concrete type for this.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -144,6 +142,185 @@ private void gatherNonRecordStores( Collection<StoreFileMetadata> files, boolean
...
+        }
+    }
+
     private Resource gatherExplicitIndexFiles( Collection<StoreFileMetadata> files ) throws IOException
```
chrisvest: This and the following two private methods are no longer used.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -144,6 +142,185 @@ private void gatherNonRecordStores( Collection<StoreFileMetadata> files, boolean
...
+        public InclusiveFileLister includeAdditionalProviders()
+        {
+            excludeAdditionalProviders = false;
+            return this;
+        }
+
+        public ResourceIterator<StoreFileMetadata> build() throws IOException
+        {
+            List<StoreFileMetadata> files = new ArrayList<>();
+            List<Resource> resources = new ArrayList<>();
```
chrisvest, line 258: We should make sure that the resources we've collected, are closed if that `IOException` is thrown mid-way through.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/StoreListingResponseHandler.java
```diff
@@ -0,0 +1,46 @@
...
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.SimpleChannelInboundHandler;
+
+import org.neo4j.causalclustering.catchup.storecopy.PrepareStoreCopyResponse;
+
+public class StoreListingResponseHandler extends SimpleChannelInboundHandler<PrepareStoreCopyResponse>
+{
+    private final CatchupClientProtocol protocol;
+    private CatchUpResponseHandler handler;
```
chrisvest: Any reason `handler` is not also `final`?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
```diff
@@ -0,0 +1,134 @@
...
+                storeFileStreamingProtocol.end( ctx, StoreCopyFinishedResponse.Status.E_UNKNOWN ).addListener( closeablesListener );
+            }
+            finally
+            {
+                protocol.expect( CatchupServerProtocol.State.MESSAGE_TYPE );
+            }
+        }
+    }
+
+    private boolean isTransactionWithinReach( long atLeast )
```
chrisvest: Perhaps `requiredTxId` is more descriptive of what this is, than `atLeast`.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+            catch ( IOException e )
+            {
+                log.error( "Unable to acquire lock while preparing to serve catchup, consider the the transaction to be `behind`" );
+                return false;
+            }
+        }
+    }
+
+    private boolean isEqualOrGreater( long atLeast )
```
chrisvest: This method could be called `isWithinLastCheckPoint`, to make it sound more domain relevant.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
```diff
@@ -0,0 +1,134 @@
...
+        else
+        {
+            try
+            {
+                checkpointerSupplier.get().tryCheckPoint( new SimpleTriggerInfo( "Store file copy" ) );
+                return isEqualOrGreater( atLeast );
+            }
+            catch ( IOException e )
+            {
+                log.error( "Unable to acquire lock while preparing to serve catchup, consider the the transaction to be `behind`" );
```
chrisvest: Is this really an error, or can we log it as a warning instead? The fallback is that we'll do a full copy instead of an incremental one, right? Logging with `error` could potentially mean that an ops person gets woken up at night.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetStoreFileRequestHandler.java
```diff
@@ -0,0 +1,149 @@
...
+                        if ( storeFileMetadata.isEmpty() )
+                        {
+                            throw new IllegalArgumentException( "No such file: " + getFileName( fileRequest ) );
+                        }
+                        else if ( storeFileMetadata.size() > 1 )
+                        {
+                            throw new IllegalStateException( "Found multiple files of: " + getFileName( fileRequest ) );
+                        }
+                        return storeFileMetadata.get( 0 );
+                    } ) );
```
chrisvest: I think it reads better if you do these checks outside of the collector.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetStoreFileRequestHandler.java
```diff
@@ -0,0 +1,149 @@
...
+                log.error( "Unable to acquire lock while preparing to serve catchup, consider the the transaction to be `behind`" );
+                return false;
+            }
+        }
+    }
+
+    private boolean isEqualOrGreater( long atLeast )
+    {
+        return checkpointerSupplier.get().lastCheckPointedTransactionId() >= atLeast;
+    }
```
chrisvest: The `isTransactionWithinReach` and `isEqualOrGreater` methods are also found in the `GetLuceneSnapshotRequestHandler` class. Could we find a way to only have one copy?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
```diff
@@ -0,0 +1,134 @@
...
+                return false;
+            }
+        }
+    }
+
+    private boolean isEqualOrGreater( long atLeast )
+    {
+        return checkpointerSupplier.get().lastCheckPointedTransactionId() >= atLeast;
```
chrisvest: This is true if the most recent checkpoint transaction at the time of the "prepare store copy" is still the most recent one at this time, or if a more recent checkpoint exists. It will be false if the most recent checkpoint is now older than the most recent checkpoint observed at the "prepare store copy" time.

What's the logic behind this check? The `isTransactionWithinReach` sounds like we are asserting that the transaction log is still containing the transaction pointed to by the latest checkpoint at "prepare store copy" time, but I'm not sure that this check accomplishes this.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/MaximumTotalRetries.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        if ( tries >= maxRetries )
+        {
+            throw new StoreCopyFailedException( "Maximum allowed retries exceeded: " + maxRetries );
+        }
+    }
+
+    private boolean timeHasExpired( long previousCheck, long currentTime )
+    {
+        return (currentTime - previousCheck) > allowedInBetweenTimeMillis;
```
chrisvest, line 62: This is most likely always true on the _first_ call to `assertContinue`. I don't know if that's intentional.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/PrepareStoreCopyFiles.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public StoreResource[] getAtomicFilesSnapshot() throws IOException
+    {
+        Stream<StoreFileMetadata> countStoreStream =
+                neoStoreDataSource.listStoreFilesInclusively().includeNeoStoreFiles().build().stream().filter( isCountFile() );
+        Stream<StoreFileMetadata> explicitIndexStream =
+                neoStoreDataSource.listStoreFilesInclusively().includeExplicitIndexStoreStoreFiles().includeAdditionalProviders().build().stream();
+        StoreFileMetadata[] metaDataStores =
+                closeablesListener.sample( Stream.concat( countStoreStream, explicitIndexStream ) ).toArray( StoreFileMetadata[]::new );
```
chrisvest: I think you should add the resource iterators to the `closeablesListener` one at a time, so if the second `build()` fails, you don't forget to close the first one. If you add the resource iterators directly, you can then later build the streams, and then use them to map to StoreResources, and then return the final array.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/PrepareStoreCopyRequestHandler.java
```diff
@@ -0,0 +1,119 @@
...
+                streamingProtocol.stream( channelHandlerContext, storeResource );
+            }
+            channelHandlerContext.write( ResponseMessageType.PREPARE_STORE_COPY_RESPONSE );
+            PrepareStoreCopyResponse success = createSuccessfulResponse( checkPointer, prepareStoreCopyFiles );
+            channelHandlerContext.writeAndFlush( success ).addListener( closeablesListener );
+        }
+        catch ( Throwable e )
+        {
+            channelHandlerContext.writeAndFlush( PrepareStoreCopyResponse.error( PrepareStoreCopyResponse.Status.E_LISTING_STORE ) ).addListener(
+                    closeablesListener );
```
chrisvest: You're not propagating the `Throwable` here.

chrisvest: You could move `channelHandlerContext.writeAndFlush( success ).addListener( closeablesListener );` to the `finally` clause, and instead just assign `success` in the catch clause, or whatever you want to call that variable.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/RemoteStore.java
```diff
@@ -53,6 +56,7 @@
  */
 public class RemoteStore
 {
+    public static final Supplier<TerminationCondition> DEFAULT_TERMINATION_CONDITIONS = () -> new MaximumTotalRetries( 10, 10 );
```
chrisvest: This can be private.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClient.java
```diff
@@ -19,63 +19,132 @@
...
+                }
+                if ( !successful )
                 {
-                    storeFileStreams.write( destination, requiredAlignment, fileChunk.bytes() );
-                    return fileChunk.isLast();
+                    log.error( "Failed to download file '%s'", file );
+                    terminationCondition.assertContinue();
                 }
+            }
+            while ( !successful );
```
chrisvest, line 98: If we fail because the `lastTransactionId` is too old, then there's no point in retrying, right? Could we have a case to handle that? And I suppose in that case we'd also have to throw away anything we've already copied thus far, right?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClient.java
```diff
@@ -97,4 +166,61 @@ public void onGetStoreIdResponse( CompletableFuture<StoreId> signal, GetStoreIdR
...
         }
     }
+
+    private boolean successfulFileDownload( StoreCopyFinishedResponse response ) throws StoreCopyFailedException
+    {
+        StoreCopyFinishedResponse.Status responseStatus = response.status();
+        log.debug( "Request for individual file resulted in response type: %s", response.status() );
+        if ( responseStatus == StoreCopyFinishedResponse.Status.E_TOO_FAR_BEHIND )
+        {
+            return false;
```
chrisvest, line 176: Regarding my comment above, I guess we'd want to throw a `StoreCopyFailedException` here instead.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloader.java
```diff
@@ -103,32 +99,30 @@ void downloadSnapshot( MemberId source ) throws StoreCopyFailedException
...
             {
                 StoreId localStoreId = localDatabase.storeId();
-                CatchupResult catchupResult = remoteStore.tryCatchingUp( fromAddress, localStoreId, localDatabase
-                        .storeDir(), false );
+                CatchupResult catchupResult = remoteStore.tryCatchingUp( primary, localStoreId, localDatabase.storeDir(), false );
 
                 if ( catchupResult == E_TRANSACTION_PRUNED )
                 {
-                    log.info( format( "Failed to pull transactions from %s (%s). They may have been pruned away", source, fromAddress ) );
+                    log.info( format( "Failed to pull transactions from %s (%s). They may have been pruned away", addressProvider, primary ) );
```
chrisvest: Will the address provider have a useful toString in this case?

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/storecopy/FakeCatchupServer.java
```diff
@@ -0,0 +1,217 @@
...
+    @Override
+    protected ChannelHandler getStoreFileRequestHandler( CatchupServerProtocol protocol )
+    {
+        return new SimpleChannelInboundHandler<GetStoreFileRequest>()
+        {
+            @Override
+            protected void channelRead0( ChannelHandlerContext channelHandlerContext, GetStoreFileRequest getStoreFileRequest ) throws Exception
+            {
+//                log.debug( "Received request for file %s", getStoreFileRequest.file().getName() );
+                log.info( "Received request for file %s", getStoreFileRequest.file().getName() );
```
chrisvest: Debugging leftovers?

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/storecopy/MaximumTotalRetriesTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        maximumTotalRetries.assertContinue();
+        maximumTotalRetries.assertContinue();
+        maximumTotalRetries.assertContinue();
+        expectedException.expect( StoreCopyFailedException.class );
+        maximumTotalRetries.assertContinue();
+    }
+
+    @Test
+    public void shouldContinueIfAlloedInBetweenTimeIsMet() throws Exception
```
chrisvest: `s/Alloed/Allowed/`

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/storecopy/SimpleCatchupClient.java
```diff
@@ -0,0 +1,124 @@
...
+                lastCheckPointedTransactionId ), new StoreCopyClient.StoreFileCopyResponseAdaptor( streamToDisk, log ) );
+    }
+
+    @Override
+    public void close() throws Exception
+    {
+        CloseablesListener closeablesListener = new CloseablesListener();
+        closeablesListener.sample( streamToDisk );
+        closeablesListener.sample( clientPageCache );
+        closeablesListener.close();
```
chrisvest: You could also `IOUtils.closeAll` these.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClientIT.java
```diff
@@ -0,0 +1,262 @@
...
+
+    private FakeFile luceneFileA = new FakeFile( "lucene", "Lucene 123" );
+
+    private static void writeContents( FileSystemAbstraction fileSystemAbstraction, File file, String contents )
+    {
+        byte[] bytes = contents.getBytes();
+        try ( StoreChannel storeChannel = fileSystemAbstraction.create( file ) )
+        {
+            storeChannel.write( ByteBuffer.wrap( bytes ) );
+            storeChannel.close();
```
chrisvest: The try-with-resources takes care of the closing, right?

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClientIT.java
```diff
@@ -0,0 +1,262 @@
...
+                return throwableSupplier.get();
+            }
+            catch ( Throwable throwable )
+            {
+                throw new RuntimeException( throwable );
+            }
+        };
+    }
+
+    private interface ThrowableSupplier<T>
```
chrisvest: We also have `org.neo4j.function.ThrowingSupplier`.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/storecopy/TerminationConditionMaximumRetriesTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public void exceedingNumberOfRetriesCausesError() throws StoreCopyFailedException, InterruptedException
+    {
+        // given a setup used for verifying number of retries
+        TerminationCondition subject = new MaximumTotalRetries( MAX_RETRIES, DURATION_BETWEEN_RETRIES_SHORT );
+
+        // when we exceed the nubmer of retries
+        for ( int i = 0; i < MAX_RETRIES - 1; i++ )
+        {
+            subject.assertContinue();
+            Thread.sleep( 2 * DURATION_BETWEEN_RETRIES_SHORT );
```
chrisvest: ಠ_ಠ

Can you somehow use `FakeClock` instead?

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/storecopy/TestPageCache.java
```diff
@@ -0,0 +1,58 @@
...
+import org.neo4j.io.pagecache.impl.SingleFilePageSwapperFactory;
+import org.neo4j.io.pagecache.impl.muninn.MuninnPageCache;
+import org.neo4j.io.pagecache.tracing.DefaultPageCacheTracer;
+import org.neo4j.io.pagecache.tracing.PageCacheTracer;
+import org.neo4j.io.pagecache.tracing.cursor.DefaultPageCursorTracerSupplier;
+import org.neo4j.io.pagecache.tracing.cursor.PageCursorTracerSupplier;
+import org.neo4j.io.pagecache.tracing.cursor.context.EmptyVersionContext;
+import org.neo4j.io.pagecache.tracing.cursor.context.EmptyVersionContextSupplier;
+
+public class TestPageCache
```
chrisvest: Can't you use `PageCacheRule` or `StandalonePageCacheFactory` instead?

chrisvest (changes requested):
Don't be alarmed by the number of comments. Most of them are small things.

I think, however, that you are grabbing the wrong kind of descriptors. The schema provider descriptors are for the _schema index providers_, where Lucene is one, the native indexes are one, etc. I think what you want to grab are the _index descriptors_, such that you are grabbing a snapshot, so to speak, per index, rather than for all indexes for a given schema index provider. For instance, if the Lucene schema index provider is servicing all your indexing needs in a particular database, then all of your indexes would be stream as a single snapshot, if I'm reading the code correctly.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/CatchupAddressProvider.java
RagnarW: Yes. Currently we do all request to the leader. We plan to expand on this so that we only the the prepare request to the leader and then the interface will allow us to route the file/snapshot requests to other instances.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/CatchupClientProtocol.java
RagnarW: It is. AFAIK we will resolve this together with the rolling upgrades feature that Martin and Andrew are working on before 3.4 is released.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/RequestMessageType.java
RagnarW: It is. AFAIK we will resolve this together with the rolling upgrades feature that Martin and Andrew are working on before 3.4 is released.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
RagnarW: If this is false the request will fail and the client will have to retry somewhere else. We do not try to achieve that this is true its only a condition for being able to continue.

Currently all requests are to the leader so this should always be true. But we want to be able to send requests to other instances, possible we will have to force checkpointing to happen to avoid this condition constantly failing?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/MaximumTotalRetries.java
RagnarW: I think this is correct. The in between time is to avoid increment if there is some  temporary unavailability on the client side.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClient.java
RagnarW: Currently that is true. But we plan to download files from multiple instances and some of them might be falling behind. In that case we want to retry the request against another instance.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClient.java
RagnarW: Same response as above :)

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/snapshot/CoreStateDownloader.java
RagnarW: It should!

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileIndexListing.java
RagnarW: Quite recently I got the news that we use 160 characters in our code style. I had been using 120

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/GetLuceneSnapshotRequestHandler.java
RagnarW: Fallback is that client will retry somewhere else if it is allowed.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/catchup/storecopy/TerminationConditionMaximumRetriesTest.java
RagnarW: This looks to be overlapping MaximumTotalRetriesTest which uses a fake clock. I'll just remove it.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/MaximumTotalRetries.java
```diff
@@ -49,8 +49,8 @@ public void assertContinue() throws StoreCopyFailedException
         if ( timeHasExpired( previousCheck, currentTime ) )
         {
             tries++;
-            previousCheck = currentTime;
         }
+        previousCheck = currentTime;
```
RagnarW: This was mistakenly moved

community/kernel/src/main/java/org/neo4j/kernel/NeoStoreDataSource.java
RagnarW: As in, why not only have NeoStoreFileListing getter and then get it from there?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/IndexDescriptorSerializer.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.IOException;
+
+import org.neo4j.kernel.api.schema.LabelSchemaDescriptor;
+import org.neo4j.kernel.api.schema.index.IndexDescriptor;
+import org.neo4j.storageengine.api.ReadableChannel;
+import org.neo4j.storageengine.api.WritableChannel;
+
+class IndexDescriptorSerializer
+{
+    private static final byte LABEL_SCHEMA_DESCRIPTOR = 1;
```
chrisvest, line 31: Good future-proving 👍

chrisvest (approved):

PR #10833, opened 2018-01-17
Tx-state aware RelationshipScanCursor and relationship writes

fickludd:
Add the following functionality to the Kernel API implementation

 * Create relationship
 * Delete relationship
 * Tx-state aware single relationship lookups

Probably relationship scans are also tx-state aware, but haven't added tests yet. Could are here or in separate PR - indicate how we should do it in review.

### DIFF
--- a/community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
+++ b/community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
@@ -37,15 +37,16 @@ public abstract class RelationshipTransactionStateTestBase<G extends KernelAPIWr
         {
             n1 = tx.dataWrite().nodeCreate();
             n2 = tx.dataWrite().nodeCreate();
+
+            // setup extra relationship to challenge the implementation
             long decoyNode = tx.dataWrite().nodeCreate();
-            label = tx.tokenWrite().relationshipTypeGetOrCreateForName( "R" ); // to have >1 relationship in the db
+            label = tx.tokenWrite().relationshipTypeGetOrCreateForName( "R" );
             tx.dataWrite().relationshipCreate( n2, label, decoyNode );
             tx.success();
         }
 
         try ( Transaction tx = session.beginTransaction() )
         {
-            label = tx.tokenWrite().relationshipTypeGetOrCreateForName( "R" );
             long r = tx.dataWrite().relationshipCreate( n1, label, n2 );
             try ( RelationshipScanCursor relationship = cursors.allocateRelationshipScanCursor() )
             {
--- a/community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
+++ b/community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.internal.kernel.api;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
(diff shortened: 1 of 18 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipCursor.java
```diff
@@ -84,4 +98,52 @@ public long propertiesReference()
...
+        }
+    }
+
+    protected void loadFromTxState( long reference )
+    {
+        read.txState().relationshipVisit( reference, this );
+    }
+
+    @Override
+    public void visit( long relationshipId, int typeId, long startNodeId, long endNodeId ) throws RuntimeException
```
tinwelint: Unecessary `RuntimeException`

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipCursor.java
```diff
@@ -84,4 +98,52 @@ public long propertiesReference()
...
+    {
+        setId( relationshipId );
```
tinwelint: What about all the other record fields? Would a call to `initialize` be better?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -149,10 +149,15 @@ public boolean nodeDelete( long node ) throws AutoIndexingKernelException
     }
 
     @Override
-    public long relationshipCreate( long sourceNode, int relationshipLabel, long targetNode )
+    public long relationshipCreate( long sourceNode, int relationshipLabel, long targetNode ) throws EntityNotFoundException
     {
         ktx.assertOpen();
-        throw new UnsupportedOperationException();
+        nodeExists( sourceNode );
```
tinwelint: How is locking managed in this new implementation? Are the nodes already locked at this point?

thobe: I would like it if in due time that was the case - that locking be handled through specific locking API calls at this level. However that is not the case right now, and I think it is a bit much to add such an API change before having completed the transition to the use of this API. Therefore I think we should make sure to handle locking in this implementation for now.

fickludd: Riiiight, maybe I possibly forgot to add that locking code while sweating over tx-state complications.... Added now, good catch!

fickludd: ...we do lock here for `nodeCreate` for example, so continued the pattern for now.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
```diff
@@ -0,0 +1,95 @@
...
+            n2 = tx.dataWrite().nodeCreate();
+            long decoyNode = tx.dataWrite().nodeCreate();
+            label = tx.tokenWrite().relationshipTypeGetOrCreateForName( "R" ); // to have >1 relationship in the db
+            tx.dataWrite().relationshipCreate( n2, label, decoyNode );
+            tx.success();
+        }
+
+        try ( Transaction tx = session.beginTransaction() )
+        {
+            label = tx.tokenWrite().relationshipTypeGetOrCreateForName( "R" );
```
SaschaPeukert: You shouldn't need to do this again, right?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
```diff
@@ -0,0 +1,95 @@
...
+        long n1, n2;
+        try ( Transaction tx = session.beginTransaction() )
+        {
+            n1 = tx.dataWrite().nodeCreate();
+            n2 = tx.dataWrite().nodeCreate();
+            long decoyNode = tx.dataWrite().nodeCreate();
+            label = tx.tokenWrite().relationshipTypeGetOrCreateForName( "R" ); // to have >1 relationship in the db
+            tx.dataWrite().relationshipCreate( n2, label, decoyNode );
+            tx.success();
+        }
```
SaschaPeukert: It might be helpful to add a comment explaining that this first transaction is just setting up a decoy relationship?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -563,11 +592,34 @@ private void acquireExclusiveNodeLock( long node )
         }
     }
 
+    private void acquireExclusiveRelationshipLock( long relationshipId )
+    {
+        if ( !ktx.hasTxStateWithChanges() || !ktx.txState().relationshipIsAddedInThisTx( relationshipId ) )
```
SaschaPeukert, line 597: I might be wrong here, but what about a transaction like this:

CREATE ()-[a:R]->()
CREATE  ()-[b:R]->()
DELETE a
DELETE b

Wouldn't b be unable to get the exclusive lock here?

SaschaPeukert (commented):
All in all very nice! I have just a few questions

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
fickludd: As both a and b are created in the transaction, no locks are needed or acquired for this case. The locks guard against concurrent modification of the nodes by other transactions, but as a and b are created in the current transaction, no other transaction will know about them until post commit, and therefor we do not need to lock them.

tinwelint (approved):

PR #10834, opened 2018-01-18
Fix regression for label scans

pontusmelke:
After switching to the new API there was a regression for scanning nodes by label in the Core API. This PR addresses that PR by:
- Less indirection in the cursor, or rather the cursor knows about the type of the progressor and thus avoiding a megamorphic call site
- More efficient cursor-to-iterator in `GraphDatabaseFacade`

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
@@ -119,6 +119,7 @@ import static org.neo4j.kernel.impl.api.explicitindex.InternalAutoIndexing.NODE_
 import static org.neo4j.kernel.impl.api.explicitindex.InternalAutoIndexing.RELATIONSHIP_AUTO_INDEX;
 import static org.neo4j.kernel.impl.api.operations.KeyReadOperations.NO_SUCH_LABEL;
 import static org.neo4j.kernel.impl.api.operations.KeyReadOperations.NO_SUCH_PROPERTY_KEY;
+import static org.neo4j.kernel.impl.store.record.AbstractBaseRecord.NO_ID;
 
 /**
  * Implementation of the GraphDatabaseService/GraphDatabaseService interfaces - the "Core API". Given an {@link SPI}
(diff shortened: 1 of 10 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -922,4 +895,69 @@ public GraphPropertiesProxy newGraphPropertiesProxy()
...
+        private final NodeLabelIndexCursor cursor;
+        private final Statement statement;
+        private long next;
+        private boolean closed;
+
+        NodeCursorResourceIterator( NodeLabelIndexCursor cursor, Statement statement )
+        {
+            this.cursor = cursor;
+            this.statement = statement;
+            next = -2L;
```
craigtaverner: Should be UNINITIALIZED

craigtaverner: Or perhaps the first fetchNext could be called in the constructor to remove the need for this special value?

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -922,4 +895,69 @@ public GraphPropertiesProxy newGraphPropertiesProxy()
     {
         return new GraphPropertiesProxy( this );
     }
+
+    private final class NodeCursorResourceIterator implements ResourceIterator<Node>
+    {
+        private static final long UNINITIALIZED = -2L;
+        private static final long NO_ID = -1L;
```
craigtaverner: Could use org.neo4j.kernel.impl.store.record.AbstractBaseRecord.NO_ID

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeLabelIndexCursor.java
```diff
@@ -109,7 +109,8 @@ public boolean next()
         }
         else
         {
-            return innerNext();
+            LabelScanValueIndexProgressor p = progressor;
```
craigtaverner: This change does not seem to do anything? The same code exists in `innerNext`, right?

craigtaverner (changes requested):

craigtaverner (approved):

PR #10835, opened 2018-01-18
Prefix cursor implementations with `Default`

fickludd:
Adding a prefix to the cursor implementation names makes it easier to talk about them, and reduced confusion in some places. On the other hand it makes the names longer and harder to read. 

This renaming is an option. Vote +1 or -1. On majority we make the change?

Other alternative prefixes
 * `Plain`
 * `A`
 * `Unextraordinary`
 * `Stock`
 * `Vanilla`
 * `Pcc` (Pessimistic Concurrency Control)
 * `Record`


thobe:
Since these cursor implementations use `Record` objects as their abstraction for reading the store, how about we use the prefix `RecordBased` instead of `Default`, since it actually says something useful, and hints at a future where the `Record` objects finally die in favour of an abstraction that allows us to implement more efficient store formats.

fickludd:
I'm also not very passionate about `Default`. Please remember thought the a non-record-based kernel might not be the next one. If the next kernel implementation uses a off-heap tx-state but still with record store access, naming this implementation `Record` is not that natural in differentiating the two. Likewise if the next one only changes concurrency control for example. That's why I've been looking a non-descriptive names.

chrisvest:
+1 on `Default`, until a second product implementation exists and we know how they're different.

burqen:
+1 on `Default` for me as well.

fickludd:
Hmm ok. Anyone want to keep them having the same name?

I'm just slightly annoyed that `Default` is long to type and think, but it seems like I'm the only one concerned about that? Would prefer shorter prefix... `BlandNodeCursor`?

burqen:
@fickludd I agree that `Default` is more complicated and longer than what I would prefer but I can't think of any other name that also "fits". I don't like `Bland` because the word in it self doesn't fit my code vocabulary. It feels like it is taken from the outside in some way. This is a pretty vague feeling from my subjective perspective and like I said, I don't like `Default` that much better 😅

thobe:
@fickludd but will that next implementation (that is also record based) co-exist with the current implementation? If not, I'd argue it is an evolution of the same implementation, rather than a new implementation.

fickludd:
Ok, merging this because different name seems to be better that same name.

PR #10836, opened 2018-01-18
Introduce temporal types to Cypher

thobe:
This introduces temporal `Value` types and the beginning of functions for creating an manipulating them through Cypher.

These functions are defined as built in functions in the database, and not "special cases" in the parser. In order to further support this, the `CallableUserFunction` interface has been updated to work with `Value` objects, moving the conversion to embedded types to the reflective function implementation (that conversion happened on a higher level before, so this means that we can avoid the conversion for non-reflective functions).

### DIFF
--- a/community/cypher/runtime-util/src/test/scala/org/neo4j/cypher/FakeClock.scala
+++ b/community/cypher/runtime-util/src/test/scala/org/neo4j/cypher/FakeClock.scala
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher
+
+import java.time.Instant
+
+import org.neo4j.time.Clocks
+
+trait FakeClock {
+  val clock: org.neo4j.time.FakeClock = Clocks.fakeClock(Instant.now())
+}
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/core/PathProxy.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/core/PathProxy.java
@@ -1,0 +1,284 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.core;
+
+import java.util.Arrays;
+import java.util.Iterator;
+
+import org.neo4j.graphdb.Node;
+import org.neo4j.graphdb.Path;
+import org.neo4j.graphdb.PropertyContainer;
+import org.neo4j.graphdb.Relationship;
+import org.neo4j.graphdb.traversal.Paths;
+
+import static org.neo4j.helpers.collection.Iterators.iteratorsEqual;
+
+public class PathProxy implements Path
+{
+    private final EmbeddedProxySPI proxySPI;
+    private final long[] nodes;
+    private final long[] relationships;
+    private final int[] directedTypes;
+
+    /**
+     * @param proxySPI
+     *         the API into the kernel.
+     * @param nodes
+     *         the ids of the nodes in the path, in order.
+     * @param relationships
+     *         the ids of the relationships in the path, in order.
+     * @param directedTypes
+     *         an encoding of the types and directions of the relationships.
+     *         An entry at position {@code i} of this array should be {@code typeId} if the relationship at {@code i}
+     *         has its start node at {@code i} and its end node at {@code i + 1}, and should be {@code ~typeId} if the
+     *         relationship at {@code i} has its start node at {@code i + 1} and its end node at {@code i}.
+     */
+    public PathProxy( EmbeddedProxySPI proxySPI, long[] nodes, long[] relationships, int[] directedTypes )
+    {
+        assert nodes.length == relationships.length + 1;
+        assert relationships.length == directedTypes.length;
+        this.proxySPI = proxySPI;
+        this.nodes = nodes;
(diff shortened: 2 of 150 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/TemporalFunctionsAcceptanceTest.scala
```diff
@@ -0,0 +1,51 @@
...
+  test("should get current default datetime") {
+    val result = executeWith(supported, "RETURN datetime() as now")
+
+    val now = single(result.columnAs[DateTimeValue]("now"))
+
+    now shouldBe a[DateTimeValue]
+  }
+
+  test("should get current 'now' datetime") {
+    val result = executeWith(supported, "RETURN datetime({when:'now'}) as now")
```
systay: This looks weird to me. I get using maps for mult-argument procedures, to avoid using positional arguments. For single argument procs, I think it's unnecessary to be inconsistent with most other procedures and functions we have and use maps.

systay:
This makes all kinds of giddy. It's been a long time coming. :)

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/TemporalFunctionsAcceptanceTest.scala
thobe: Please see [CIP2015-08-06](https://github.com/thobe/openCypher/blob/date-time/cip/1.accepted/CIP2015-08-06-date-time.adoc), the map is needed here to distinguish from the different overloads of the function. It looks weird because the other tests have not been written yet.

thobe: But your comment got me thinking. So we discussed it and came up with what might be a better way to do this particular thing: https://github.com/opencypher/openCypher/pull/15/commits/5bb14895b95e40a0e3769c9120c9e3b161787227

thobe:
I have simplified the content of this branch (in all the commits, to facilitate reviewing), some additional features are moved out into https://github.com/thobe/neo4j/tree/3.4-temporal-experiments

These simplifications are in line with simplifications that have been done to [the specification](https://github.com/thobe/openCypher/blob/date-time/cip/1.accepted/CIP2015-08-06-date-time.adoc). The additional features in that branch are no longer in line with the specification.

thobe:
I'm not sure the commit entitled `CastSupport for temporal values` is actually needed, it just follows a pattern that existed in that file.

thobe:
Remaining work:

- [ ] Lots of testing needs to be added.
- [ ] There are quite a few blanks left to fill in for the implementation of the temporal functions.
- [ ] Storage of temporal values. _After which we can:_
- [ ] Add tests to the Core API for storing and retrieving `java.time.{ZonedDateTime,LocalDateTime,LocalDate,OffsetTime,LocalTime}`
- [ ] BOLT serialisation of temporal values.
- [ ] Support for reading temporal values from CSV files in the importer (by using the static `parse(...)` methods of the respective types)

community/kernel/src/main/java/org/neo4j/kernel/impl/core/PathProxy.java
```diff
@@ -0,0 +1,284 @@
...
+            if ( inTx )
+            {
+                try
+                {
+                    String name = proxySPI.getRelationshipTypeById( type < 0 ? ~type : type ).name();
+                    string.append( ':' ).append( name );
+                }
+                catch ( Exception e )
+                {
+                    inTx = false;
```
sherfert, line 83: Is there another way to figure out, if we are in a trnasaction currently?

thobe: Through quite a bit of refactoring.

thobe: There is a bit of redemption though, which makes this solution not-so-bad, and makes that refactoring not-really-worth-it, and that is that typically the `getRelationshipTypeById(...)`-method works even outside of a transaction - although it is not guaranteed to... (and would definitely fail if the database has been shut down)

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
```diff
@@ -0,0 +1,805 @@
...
+        String s = matcher.group( sec );
+        String n = matcher.group( sub );
+        long seconds = optLong( s );
+        if ( strict )
+        {
+            if ( h > 24 )
+            {
+                throw new IllegalArgumentException( "hours out of range: " + h );
+            }
+            if ( m > 60 )
```
sherfert, line 366: I could only find the  valid ranges for minute, second, etc. for the `Time` types. If the caps are at 59 for minutes and 60 for seconds _there_, why are they both at 60 for durations?

thobe: good point, should have been `>=`

thobe: Classic kind of thing that should be covered by a test case :)

thobe:
@sherfert I actually think that writing tests for some code is a great way of reviewing it, so in a way the low test coverage is a feature of the pull request...

community/values/src/main/java/org/neo4j/values/storable/PointValue.java
```diff
@@ -168,9 +169,9 @@ else if ( this.coordinate.length < other.coordinate.length )
     }
 
     @Override
-    public Object asObjectCopy()
+    public Point asObjectCopy()
     {
-        return new PointValue( this.crs, this.coordinate );
+        return this;
```
sherfert, line 174: I am unsure if it is OK not to create a new object (as done with `clone` in other implementations of `asObjectCopy`).

thobe: If you read the rest of the code, you will find that `PointValue` is never mutated, and that the `coordinate` field is not made public to anything that can mutate it.

Would applying this diff make you more comfortable?

```
--- a/community/values/src/main/java/org/neo4j/values/storable/PointValue.java
+++ b/community/values/src/main/java/org/neo4j/values/storable/PointValue.java
@@ -32,8 +32,8 @@ import static java.util.Collections.singletonList;
 
 public class PointValue extends ScalarValue implements Comparable<PointValue>, Point
 {
-    private CoordinateReferenceSystem crs;
-    private double[] coordinate;
+    private final CoordinateReferenceSystem crs;
+    private final double[] coordinate;
 
     PointValue( CoordinateReferenceSystem crs, double... coordinate )
     {
```

community/cypher/runtime-util/src/test/scala/org/neo4j/cypher/FakeClock.scala
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher
+
+import java.time.Instant
+
+import org.neo4j.time.Clocks
+
+trait FakeClock {
```
sherfert, line 26: Would it make sense to add this here?
```
self: GraphDatabaseTestSupport  =>
```

thobe: Maybe. It wasn't needed, so I didn't bother with the extra typing.

community/values/src/main/java/org/neo4j/values/storable/PointValue.java
thobe: Looking further at the usages of the `coordinate()` method through which the `double[] coordinate` field might leak, I think it would make a lot of sense to instead expose these two methods:

```
public int numberOfCoordinates()
{
    return coordinate.length;
}
public double coordinate( int i )
{
    return coordinate[i];
}
```

sherfert (approved):

henriknyman (approved):
Looks good so far. 
I guess we want to also consider changing procedure call in a similar way as function call?
Have you benchmarked any performance impact?

thobe:
@henriknyman yes, I agree that we would want to change procedures similarly to how functions are changed here, but I wanted to limit the scope of this particular changeset.

I have not yet measured the performance impact, but I have added it to the benchmarking queue, so it should be done soon.

PR #10837, opened 2018-01-18
Use tmp byte buffer on insert and remove slot

burqen:
Previously slot was moved one slot at the time using cursor.copyTo.
This turned out to be very expensive especially when a large number
items needed to be moved, when keys are small.

Instead we now use a byte array that we temporarily write to
before writing it down again. This gives us the possibility to
operate on bigger chunks.

Benchmarking will decide how big this byte array should be.

NOT READY FOR MERGE - Waiting for benchmark results.


burqen:
Update! Fullstack benchmarks show no difference. Waiting for gbptree specific benchmarks.

tinwelint:
So this guy is null and void, due to up and coming #10852 right?

burqen:
@tinwelint Yes! Closing this bad boy.

PR #10839, opened 2018-01-18
More fine grained versioning of GBPTree format

tinwelint:
Previously there was this global FORMAT_VERSION that controlled a single format version
of the whole GBPTree (and index component in general). Now with the introduction of
dynamic-size format in additional to the fixed-size this required a change because
there's now two active formats concurrently.

This change goes a bit further than that in that it splits up the version int
in four pieces, four different versions of a GBPTree. Two of them are used now,
which are format identifier and format version. Only the format identifier is used
now, sort of, but soon perhaps the format version will also come into play.
The two remaining versions are left as unused for now. Format identifier for
fixed-size format is kept at 2, which was the old FORMAT_VERSION and its
version set to 0. This keeps backwards compatibility with existing indexes.

Any existing dynamic-size indexes created before this commit will have to be recreated,
but in practice this will not be a problem because it shouldn't really be used anywhere.

Logic for open/create in GBPTree constructor has had some slight face lift to be
able to handle this new format diversity, introducing Meta class for reading/writing
and also verifying meta page.


tinwelint:
@burqen updated

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNodeSelector.java
```diff
@@ -0,0 +1,147 @@
...
+         * @param layout {@link Layout} that will be used in this format.
+         * @return the instantiated {@link TreeNode}.
+         */
+        <KEY,VALUE> TreeNode<KEY,VALUE> create( int pageSize, Layout<KEY,VALUE> layout );
+
+        /**
+         * Specifies the format identifier of the physical layout of tree nodes.
+         * A format identifier must be unique among all possible existing format identifiers.
+         * It's used to differentiate between different types of formats.
+         * On top of this a specific format identifier may be {@link #formatVersion()}.
```
burqen: I don't understand what is meant by this last sentence. Wanna help? 😅

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNodeSelector.java
```diff
@@ -0,0 +1,147 @@
...
+ */
+package org.neo4j.index.internal.gbptree;
+
+import static java.lang.String.format;
+
+/**
+ * Able to select implementation of {@link TreeNode} to use in different scenarios, should be used in favor of directly
+ * instantiating {@link TreeNode} instances.
+ */
+class TreeNodeSelector
```
burqen, line 28: Nice javadoc as always! I like this class 😄

community/index/src/main/java/org/neo4j/index/internal/gbptree/Meta.java
```diff
@@ -0,0 +1,243 @@
...
+ * For compatibility reasons this int has been kept but used differently, i.e. split up into four individual versions,
+ * one {@code byte} each. These are:
+ *
+ * <pre>
+ *     <------- int ------>
+ * msb [ 3 ][ 2 ][ 1 ][ 0 ] lsb
+ *       ▲    ▲    ▲    ▲
+ *       │    │    │    │
+ *       │    │    │    └──────────── {@link TreeNode#formatIdentifier()}
+ *       │    │    └───────────────── {@link TreeNode#formatVersion()}
```
burqen: `TreeNode#formatIdentifier` and `TreeNode#formatVersion` can not be resolved

community/index/src/main/java/org/neo4j/index/internal/gbptree/Meta.java
```diff
@@ -0,0 +1,243 @@
...
+ *       │    └────────────────────── {@link #CURRENT_STATE_VERSION}
+ *       └─────────────────────────── {@link #CURRENT_GBPTREE_VERSION}
```
burqen: Do we need to keep those two bytes reserved for state- and gbptree-version or can we just say that they are unused for now?

community/index/src/main/java/org/neo4j/index/internal/gbptree/Meta.java
```diff
@@ -0,0 +1,243 @@
...
+                    "Tried to open using layout not compatible with " +
+                    "what the index was created with. Created with: layoutIdentifier=%d,majorVersion=%d,minorVersion=%d. " +
+                    "Opened with layoutIdentifier=%d,majorVersion=%d,minorVersion=%d",
+                    layoutIdentifier, layoutMajorVersion, layoutMinorVersion,
+                    layout.identifier(), layout.majorVersion(), layout.minorVersion() );
+        }
+
+        Factory formatByLayout = TreeNodeSelector.selectByLayout( layout );
+        if ( formatByLayout.formatIdentifier() != formatIdentifier ||
+             formatByLayout.formatVersion() != formatVersion )
```
burqen, line 168: Here we assume that each `Layout` will only ever be OK with one format which is not necessarily the case. A layout may demand a Dynamic layout for example but maybe don't care about which version.
A better way IMHO would be to create the Format from the read identifier and version and then check compatibility between layout and format, i.e. does the format fulfil the demands of the layout?

community/index/src/main/java/org/neo4j/index/internal/gbptree/Meta.java
```diff
@@ -0,0 +1,243 @@
...
+        layout.writeMetaData( cursor );
+        checkOutOfBounds( cursor );
+    }
+
+    private static byte extractIndividualVersion( int format, int shift )
+    {
+        return (byte) ((format >>> shift) & MASK_BYTE);
+    }
+
+    int allVersionsCombined()
```
burqen: Can be private

community/index/src/main/java/org/neo4j/index/internal/gbptree/Meta.java
```diff
@@ -0,0 +1,243 @@
...
+    {
+        return pageSize;
+    }
+
+    byte getFormatIdentifier()
+    {
+        return formatIdentifier;
+    }
+
+    public byte getFormatVersion()
```
burqen: Can be package-private

community/index/src/main/java/org/neo4j/index/internal/gbptree/Meta.java
```diff
@@ -0,0 +1,243 @@
...
+    {
+        return (byte) ((format >>> shift) & MASK_BYTE);
+    }
+
+    int allVersionsCombined()
+    {
+        return formatIdentifier >>> SHIFT_FORMAT_IDENTIFIER | formatVersion >>> SHIFT_FORMAT_VERSION;
+    }
+
+    public int getPageSize()
```
burqen: Can be package-private

community/index/src/main/java/org/neo4j/index/internal/gbptree/Meta.java
```diff
@@ -0,0 +1,243 @@
...
+ *       │    │    │    └──────────── {@link TreeNode#formatIdentifier()}
+ *       │    │    └───────────────── {@link TreeNode#formatVersion()}
+ *       │    └────────────────────── {@link #CURRENT_STATE_VERSION}
+ *       └─────────────────────────── {@link #CURRENT_GBPTREE_VERSION}
+ * </pre>
+ *
+ * {@link #CURRENT_STATE_VERSION} and {@link #CURRENT_GBPTREE_VERSION} aren't used yet because they have
+ * never needed to be versioned yet, but remain reserved for future use. The are fixed at 0 a.t.m.
+ */
+class Meta
```
burqen, line 49: I really like that we are breaking this part out from `GBPTree`! Feels like "Oh God, why did we not do this before?!"

community/index/src/test/java/org/neo4j/index/internal/gbptree/SimpleLongLayout.java
```diff
@@ -29,21 +29,75 @@
 {
     private final int keyPadding;
     private String customNameAsMetaData;
+    private final boolean fixedSize;
+    private final int identifier;
+    private final int majorVersion;
+    private final int minorVersion;
 
-    SimpleLongLayout( int keyPadding, String customNameAsMetaData )
+    static class Builder
```
burqen, line 37: Yes! This makes it so much easier to use. Just what a test utility class needs :D

community/index/src/test/java/org/neo4j/index/internal/gbptree/SimpleLongLayout.java
```diff
@@ -29,21 +29,75 @@
...
+        }
+
+        SimpleLongLayout build()
+        {
+            return new SimpleLongLayout( keyPadding, customNameAsMetaData, fixedSize, identifier, majorVersion, minorVersion );
+        }
     }
 
-    SimpleLongLayout( int keyPadding )
+    static Builder layout()
```
burqen: Assuming static import, `layout()` reads a bit ill-defined. Maybe call method `longLayout()`?

community/index/src/test/java/org/neo4j/index/internal/gbptree/FormatCompatibilityTest.java
```diff
@@ -135,13 +156,12 @@ public void shouldDetectFormatChange() throws Throwable
 
     private void tellDeveloperToCommitThisFormatVersion()
     {
-        fail( format( "This is merely a notification to developer. Format has changed, %s.FORMAT_VERSION has also " +
+        fail( format( "This is merely a notification to developer. Format has changed and its version has also " +
```
burqen, line 159: I still find this test fascinating 😄

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberSchemaIndexProvider.java
```diff
@@ -98,6 +97,14 @@ public IndexAccessor getOnlineAccessor(
             long indexId, IndexDescriptor descriptor, IndexSamplingConfig samplingConfig ) throws IOException
     {
         File storeFile = nativeIndexFileFromIndexId( indexId );
+        NumberLayout layout = layout( descriptor );
+        return new NumberSchemaIndexAccessor<>(
+                pageCache, fs, storeFile, layout, recoveryCleanupWorkCollector, monitor, descriptor, indexId,
+                samplingConfig );
```
burqen: Funky line breaks

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberSchemaIndexProvider.java
```diff
@@ -147,7 +151,7 @@ public InternalIndexState getInitialState( long indexId, IndexDescriptor descrip
         try
         {
             NativeSchemaIndexHeaderReader headerReader = new NativeSchemaIndexHeaderReader();
-            GBPTree.readHeader( pageCache, nativeIndexFileFromIndexId( indexId ), new ReadOnlyMetaNumberLayout(),
+            GBPTree.readHeader( pageCache, nativeIndexFileFromIndexId( indexId ), layout( descriptor ),
                     headerReader );
```
burqen: Funky line break

community/index/src/main/java/org/neo4j/index/internal/gbptree/Layout.java
```diff
@@ -221,96 +221,4 @@ public boolean compatibleWith( long layoutIdentifier, int majorVersion, int mino
             return layoutIdentifier == identifier() && majorVersion == majorVersion() && minorVersion == minorVersion();
         }
     }
-
-    /**
-     * Abstract {@link Layout} for read-only scenarios. Mainly used for reading header off of a closed
-     * {@link GBPTree}.
-     */
-    @SuppressWarnings( "rawtypes" )
-    abstract class ReadOnlyMetaLayout implements Layout
```
burqen, line 230: Very good riddance!

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -800,7 +750,7 @@ private static PagedFile mapWithCorrectPageSize( PageCache pageCache, File index
         // This index was created with another page size, re-open with that actual page size
         if ( pageSize != pageCache.pageSize() )
         {
-            if ( pageSize > pageCache.pageSize() )
+            if ( pageSize > pageCache.pageSize() || pageSize < 0 )
```
burqen, line 753: Nice catch

burqen (changes requested):
I really like PR! Found some small polish points, otherwise it mostly praise 🎉

community/index/src/main/java/org/neo4j/index/internal/gbptree/Meta.java
tinwelint: Indeed!

community/index/src/main/java/org/neo4j/index/internal/gbptree/Meta.java
tinwelint: True, however this isn't a problem that we have right now. Imho right now what's important is that we store the identifier and version and then we can adapt the checking around it anyway we like when it gets to that. Guessing how it will work in the future will most likely end up in something that needs to be changed anyway.

community/index/src/test/java/org/neo4j/index/internal/gbptree/FormatCompatibilityTest.java
tinwelint: It's teh shit

PR #10840, opened 2018-01-18
Bolt V2 with point type support

lutovich:
Introduced an extension of Bolt V1 with one additional type - point. New protocol version is exactly the same as the previous one but is able to serialize/deserialize points. 2D and 3D points are supported.

Made protocol test utilities able to work with different protocol versions.


ali-ince (approved):
I think changes related to `Point` support looks good.

I had one point about `TransportTestUtil` or `Neo4jPack` usage in existing tests. I thought that it would be better if we inject both `Neo4jPackV1` and `Neo4jPackV2` as parameters into the test runs to ensure we do not break  V1 over V2 (at least for V2). WDYT?

lutovich:
@ali-ince made IT's in bolt module parameterized with `Neo4jPackV1` and `Neo4jPackV2`

ali-ince (approved):
:+1:

PR #10842, opened 2018-01-18
Optional match or pattern comprehension with writeRead conflicts

Lojjs:
Two bug fixes in one PR!

Should be forward merged to 3.3 and 3.4

changelog: Fixes bug with missing results when a relationship created in the same transaction was not found by optional match or pattern comprehension

### DIFF
--- a/community/cypher/ir-3.2/src/main/scala/org/neo4j/cypher/internal/ir/v3_2/UpdateGraph.scala
+++ b/community/cypher/ir-3.2/src/main/scala/org/neo4j/cypher/internal/ir/v3_2/UpdateGraph.scala
@@ -228,7 +228,7 @@ trait UpdateGraph {
     val allRelPatternsRead = qg.allPatternRelationshipsRead
 
     //CREATE () MATCH ()-->()
-    (allRelPatternsWritten.nonEmpty && qg.patternRelationships.nonEmpty) && allRelPatternsRead.exists(r => {
+    (allRelPatternsWritten.nonEmpty && allRelPatternsRead.nonEmpty) && allRelPatternsRead.exists(r => {
       val readProps = qg.allKnownPropertiesOnIdentifier(r.name).map(_.propertyKey)
       // CREATE ()-[]->() MATCH ()-[]-()?
       r.types.isEmpty && readProps.isEmpty ||
--- a/community/cypher/ir-3.2/src/main/scala/org/neo4j/cypher/internal/ir/v3_2/UpdateGraph.scala
+++ b/community/cypher/ir-3.2/src/main/scala/org/neo4j/cypher/internal/ir/v3_2/UpdateGraph.scala
@@ -232,8 +232,12 @@ trait UpdateGraph {
   def createRelationshipOverlapHorizon(allRelPatternsRead: Set[RelationshipPattern]): Boolean = {
     //CREATE () MATCH ()-->()
     (allRelPatternsWrittenNonEmpty && allRelPatternsRead.nonEmpty) && allRelPatternsRead.exists(r => {
-      val readProps = r.properties.map(_.asInstanceOf[MapExpression].items.map(_._1).toSet).get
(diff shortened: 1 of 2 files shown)


systay:
Nice stuff! Now we just need a changelog message for this PR.

community/cypher/ir-3.2/src/main/scala/org/neo4j/cypher/internal/ir/v3_2/UpdateGraph.scala
```diff
@@ -217,24 +222,37 @@ trait UpdateGraph {
...
+    (allRelPatternsWrittenNonEmpty && qg.allPatternRelationshipsRead.nonEmpty) && qg.allPatternRelationshipsRead.exists(r => {
+      val readProps = qg.allKnownPropertiesOnIdentifier(r.name).map(_.propertyKey)
+      relationshipOverlap(r.types.toSet, readProps)
+    })
+  }
+
+  def createRelationshipOverlapHorizon(allRelPatternsRead: Set[RelationshipPattern]): Boolean = {
+    //CREATE () MATCH ()-->()
+    (allRelPatternsWrittenNonEmpty && allRelPatternsRead.nonEmpty) && allRelPatternsRead.exists(r => {
+      val readProps = r.properties.map(_.asInstanceOf[MapExpression].items.map(_._1).toSet).get
```
sherfert: Is this a safe cast? Do all `RelationshipPattern` have a `MapExpression` as `properties`?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/EagerizationAcceptanceTest.scala
```diff
@@ -2714,6 +2714,41 @@ class EagerizationAcceptanceTest
...
+        |MERGE (tag:Tag { value: tags.value})
+        |MERGE (node)-[:HAS_TAG]->(tag)
+        |WITH DISTINCT node
+        |OPTIONAL MATCH (node)-[:HAS_TAG]->(t:Tag)
+        |RETURN COLLECT(t.value) as tags
+      """.stripMargin
+
+    val result = executeWithCostPlannerAndInterpretedRuntimeOnly(query)
+    result.toList should equal(List(Map("tags" -> Vector("apples", "oranges"))))
+    assertNumberOfEagerness(query, 2)
```
sherfert, line 2749: I am wondering why this query has an eager between the two MERGEs and the one above has not. I just checked and this was not introduced with your changes. It had an eager there before while the other one didn't. Any idea as to why this is the case? Is one query missing an eager or is this one producing an eager to much?

sherfert (approved):
Cool stuff! Some questions that need clarification and then this is good to go.

community/cypher/ir-3.2/src/main/scala/org/neo4j/cypher/internal/ir/v3_2/UpdateGraph.scala
Lojjs: Hmm, by definition it always have a `properties` that is an `Expression`, but not necessarily a  `MapExpression`. I will change this to a safer cast in an if-statement that checks that it actually is a `MapExpression`

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/EagerizationAcceptanceTest.scala
Lojjs: Interesting, no idea why that would differ, will try to look into it.

Lojjs: Have created a card about it

PR #10843, opened 2018-01-18
Fix a bug where Causal Cluster store-copy or online backup would occasionally get stuck in an infinite loop during recovery of the temporary store. Also makes store copy faster in HA, CC and online backup.

chrisvest:
The lifecycle of the StreamToDisk and the resources it opens, such as PagedFiles and PagedWritableByteChannels, are managed separately from the threads that are writing file contents through it.

Previously, the StreamToDisk implementation was oblivious to this concurrency, and would freely allow the write threads to race with the closing of the StreamToDisk instance.

This could ultimately lead to write page cursors leaking, as in, not being closed when the StreamToDisk instance was closed.

When a write page cursor was leaked, it would cause a page in the page cache (usually the last page of a file) to be permanently write locked, and this in turn would cause all future read page cursors that would access that page, to get stuck in infinite shouldRetry-loops.

Since this would always happen to the last page in a file, the database would get stuck in these infinite loops on startup, as part of the highId scanning, or the GBPTree crash recovery scanning.

This bug is fixed by making the opening and closing of files in StreamToDisk atomic and mutually exclusive. Furthermore, the writing to a paged file, and the closing of the paged file, are made mutually exclusive. Both of these pairs of operations are protected by each their own monitor locks per file, so the IO parallelism is largely preserved.

Additionally, the fields in the TrackingResponseHandler that are updated when a store copy is initiated, have been made volatile to ensure that their updates are visible to the IO threads that will be doing the writing. This ensures that the network IO ends up with the correct handlers, and thus the current StreamToDisk instance, in case multiple instances end up being juggled at runtime.

### DIFF
--- a/community/io/src/main/java/org/neo4j/io/pagecache/impl/PageCacheFlusher.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/impl/PageCacheFlusher.java
@@ -21,6 +21,7 @@ package org.neo4j.io.pagecache.impl;
 
 import java.io.Flushable;
 import java.io.IOException;
+import java.util.Objects;
 
 import org.neo4j.concurrent.BinaryLatch;
 import org.neo4j.io.pagecache.IOLimiter;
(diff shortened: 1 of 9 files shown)


ragadeeshu (approved):
If you verified that this makes it faster, awesome, go for it.

chrisvest:
@RagnarW @ragadeeshu The PR changed a fair bit after the first review. Can I have some updated opinions, please?

RagnarW (approved):

PR #10844, opened 2018-01-18
Dev-Nullifying debug.log messages from backup client

jimwebber:
Backup client brings up an embedded neo4j instance which, if run
locally with the production instance will spam its debug.log with
starting and stopping messages. These look scary to an operator.

This commit changes behavior so that when neo4j-admin backup is run,
that all debug.log output from its embedded instance goes to dev/null
or equivalent on Windows.


martinfurmanski:
Will there be a way for enabling debug logs? They are useful when debugging.

digitalstain:
Superseded by https://github.com/neo4j/neo4j/pull/11104

PR #10845, opened 2018-01-18
Refine bolt channel throttling with auto-read limiter and connection halting

ali-ince:
This is a port of #10682 into 3.4 branches. 

Having both auto-read limiter and write throttling in place can result in a situation where client tries to send bolt messages where server stops reading incoming messages (by disabling auto-read) and also since client is not reading any generated server response stops processing messages, i.e. the bolt worker thread is blocked by the transport write throttle and cannot process more messages, where since client still tries to send it cannot move on to the phase where server responses are read.

Given that above condition is a result of wrong usage (sending in messages but not reading the responses), this PR also adds a client connection termination policy where a write throttle is held more than a specified duration (`15m` by default).

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/transport/TransportWriteThrottle.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/transport/TransportWriteThrottle.java
@@ -76,6 +76,8 @@ public class TransportWriteThrottle implements TransportThrottle
     @Override
     public void acquire( Channel channel ) throws TransportThrottleException
     {
+        // if this channel's max lock duration is already exceeded, we'll allow the protocol to
+        // (at least) try to communicate the error to the client before aborting the connection
         if ( !isDurationAlreadyExceeded( channel ) )
         {
             ThrottleLock lock = channel.attr( LOCK_KEY ).get();
@@ -110,6 +112,7 @@ public class TransportWriteThrottle implements TransportThrottle
                 catch ( InterruptedException ex )
                 {
                     Thread.currentThread().interrupt();
+                    throw new RuntimeException( ex );
                 }
             }
         }
--- a/community/bolt/src/main/java/org/neo4j/bolt/transport/TransportWriteThrottle.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/transport/TransportWriteThrottle.java
@@ -26,7 +26,10 @@ import io.netty.channel.ChannelInboundHandler;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.WriteBufferWaterMark;
 import io.netty.util.AttributeKey;
+import org.apache.commons.lang3.time.DurationFormatUtils;
 
+import java.time.Clock;
+import java.time.Duration;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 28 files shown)


community/bolt/src/main/java/org/neo4j/bolt/transport/TransportWriteThrottle.java
```diff
@@ -66,19 +74,43 @@ public void install( Channel channel )
     }
 
     @Override
-    public void acquire( Channel channel )
+    public void acquire( Channel channel ) throws TransportThrottleException
     {
-        ThrottleLock lock = channel.attr( LOCK_KEY ).get();
-
-        while ( channel.isOpen() && !channel.isWritable() )
+        if ( !isDurationAlreadyExceeded( channel ) )
```
lutovich: It would be nice to throw when `#acquire()` is invoked with a channel that has expired duration. Callers should close channel after `TransportThrottleException` and not call `#acquire()` again.

community/bolt/src/main/java/org/neo4j/bolt/transport/TransportWriteThrottle.java
```diff
@@ -66,19 +74,43 @@ public void install( Channel channel )
...
+                    }
+                }
+
+                try
+                {
+                    lock.lock( channel, 1000 );
+                }
+                catch ( InterruptedException ex )
+                {
+                    Thread.currentThread().interrupt();
```
lutovich, line 114: Maybe re-throw some exception here? Otherwise callers might think that `#acquire()` call succeeded and try to write to the channel.

community/bolt/src/test/java/org/neo4j/bolt/v1/transport/integration/BoltThrottleMaxDurationIT.java
```diff
@@ -0,0 +1,181 @@
...
+    private HostnamePort address;
+
+    protected TestGraphDatabaseFactory getTestGraphDatabaseFactory()
+    {
+        TestGraphDatabaseFactory factory = new TestGraphDatabaseFactory();
+
+        logProvider = new AssertableLogProvider();
+
+        factory.setInternalLogProvider( logProvider );
+        //factory.setUserLogProvider( logProvider );
```
lutovich: Line can probably be removed

community/bolt/src/test/java/org/neo4j/bolt/v1/transport/integration/BoltThrottleMaxDurationIT.java
```diff
@@ -0,0 +1,181 @@
...
+                        pullAll()
+                ) );
+            }
+
+            return null;
+        } );
+
+        try
+        {
+            otherThread.get().awaitFuture( sender );
```
lutovich, line 159: Should there be an `Assert#fail()` call after this line?

community/bolt/src/main/java/org/neo4j/bolt/transport/TransportWriteThrottle.java
ali-ince: @lutovich I did not changed this block based on our discussion.

lutovich (approved):

PR #10846, opened 2018-01-18
Added test to validate that usage is displayed by admin tool for backup

phughk:


enterprise/backup/src/test/java/org/neo4j/commandline/admin/BackupUsageTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    public void outputMatchesExpectedForMissingBackupDir() throws UnsupportedEncodingException
+    {
+        // when
+        String output = runBackup();
+
+        // then
+        String reason = "Missing argument 'backup-dir'\n\n";
+        assertEquals( reason + BackupHelpOutput.BACKUP_OUTPUT, output );
```
andrewkerr9000: These are going to be very brittle. Maybe use Matchers.containsString to sanity check the output, without breaking on every whitespace change.

andrewkerr9000 (commented):
I suggest leaving the tests less brittle.

enterprise/backup/src/test/java/org/neo4j/commandline/admin/BackupUsageTest.java
praveenag: I disagree. This is probably the only time we test usage output and we should avoid random whitespaces anyways. So if this test breaks on whitespace, it's a good thing -> we can catch it, fix it.

praveenag (approved):

enterprise/backup/src/test/java/org/neo4j/commandline/admin/BackupUsageTest.java
lutovich: I agree with @andrewkerr9000. There is now need to be this strict. It's a responsibility of the command line framework to print help correctly. Here we are just interested in presence of "Missing argument..." string. Class `BackupHelpOutput` is also not very pretty and imho should be removed.

enterprise/backup/src/test/java/org/neo4j/commandline/admin/BackupUsageTest.java
```diff
@@ -0,0 +1,109 @@
...
+        // then
+        String reason = "Missing argument 'name'\n\n";
+        assertEquals( reason + BackupHelpOutput.BACKUP_OUTPUT, output );
+    }
+
+    @Test
+    public void incorrectBackupDirectory() throws IOException
+    {
+        // when
+        Path backupDirectoryResolved = new File( "." ).toPath().toRealPath().resolve( "non_existing_dir" );
```
lutovich: Why `new File( "." ).toPath()` and not `Paths.get(".")`?
Same in `#runBackup()` method.

enterprise/backup/src/test/java/org/neo4j/commandline/admin/BackupUsageTest.java
```diff
@@ -0,0 +1,109 @@
...
+        Path configDir = new File( "." ).toPath();
+        List<String> params = new ArrayList();
+        params.add( "backup" );
+        params.addAll( Arrays.asList( args ) );
+        String[] argArray = params.toArray( new String[params.size()] );
+        subject.execute( homeDir, configDir, argArray );
+        return baos.toString();
+    }
+
+    private ParametrisedOutsideWorld capturableOutputOutsideWorld( PrintStream printStream )
```
lutovich: There exists a JUnit rule that captures `System.out` - `SuppressOutput`. Can it be used instead?

enterprise/backup/src/test/java/org/neo4j/commandline/admin/BackupUsageTest.java
praveenag: I have specifically used test like this when I was refactoring in the past. The best part about the test was it was brittle -> led me to fix things and move on. W/o this test the only way I can figure out changes to the help text after a refactoring was by building neo4j. Please dont change it to contains check.

phughk: I will split this up into 2 tests; 1 that is checking for missing argument that doesnt include the other description and another that is checking for contains on full help.

enterprise/backup/src/main/java/org/neo4j/backup/impl/OnlineBackupContextBuilder.java
```diff
@@ -191,12 +191,25 @@ public OnlineBackupContext createContext( String... args ) throws IncorrectUsage
...
+    private Path getFolder( Arguments arguments ) throws CommandFailed
+    {
+        Path path = arguments.getMandatoryPath( ARG_NAME_BACKUP_DIRECTORY );
+        try
+        {
+            return path.toRealPath();
+        }
+        catch ( IOException e )
+        {
+            throw new CommandFailed( String.format( "Directory '%s' does not exist.", path.toFile().toString() ) );
```
chrisvest: The `toString` from `Path` should do the same, and will be implicitly called on any object processed by `%s`.

enterprise/backup/src/main/java/org/neo4j/backup/impl/ParametrisedOutsideWorld.java
```diff
@@ -32,7 +32,7 @@
 /**
  * An outside world where you can pick and choose which input/output are dummies.
  */
-class ParametrisedOutsideWorld implements OutsideWorld
+public class ParametrisedOutsideWorld implements OutsideWorld
```
chrisvest: Should this be in test sources?

enterprise/backup/src/test/java/org/neo4j/backup/impl/BackupHelpOutput.java
```diff
@@ -0,0 +1,95 @@
...
+
+public class BackupHelpOutput
+{
+    public static final String BACKUP_OUTPUT = backupOutput();
+
+    private static String backupOutput()
+    {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        stringBuilder.append( "usage: neo4j-admin backup --backup-dir=<backup-path> --name=<graph.db-backup>\n" );
```
chrisvest: All these `\n`, don't they make this platform dependent?

enterprise/backup/src/test/java/org/neo4j/commandline/admin/BackupUsageTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    public void incorrectBackupDirectory() throws IOException
+    {
+        // when
+        Path backupDirectoryResolved = new File( "." ).toPath().toRealPath().resolve( "non_existing_dir" );
+        String output = runBackup( "--backup-dir=non_existing_dir", "--name=mybackup" );
+
+        // then
+        String reason = String.format( "command failed: Directory '%s' does not exist.\n", backupDirectoryResolved.toFile().toString() );
```
chrisvest: Again, I don't think there's a need for the `.toFile().toString()` bit.

enterprise/backup/src/test/java/org/neo4j/commandline/admin/AugmentedBlockerLocator.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+class AugmentedBlockerLocator
+{
+    static BlockerLocator fromList( AdminCommand.Blocker... blockers )
+    {
+        return commandName -> Arrays.stream( blockers ).filter( blocker -> blocker.commands().contains( commandName ) ).collect( Collectors.toList() );
+    }
```
chrisvest: This is only used in one place, where we don't even give it any arguments. I don't think we need this class. The test can just pass `ArrayList::new` or something like that.

enterprise/backup/src/test/java/org/neo4j/commandline/admin/AugmentedCommandLocator.java
```diff
@@ -0,0 +1,53 @@
...
+
+class AugmentedCommandLocator
+{
+    /**
+     * Useful for tests where you know the list of providers but don't want to use the service locator
+     *
+     * @param providers list of command providers that the command locator will use
+     * @return a service locator based command locator that uses provided list of commands
+     */
+    static CommandLocator fromFixedArray( AdminCommand.Provider... providers )
```
chrisvest: If meant for this to be used by other tests (do you have any in mind?) then maybe the `enterprise/backup` module isn't the right home for it.

enterprise/backup/src/main/java/org/neo4j/backup/impl/ParametrisedOutsideWorld.java
phughk: It should definitely have a `test` scope. It is only used in backup module tests at the moment.

enterprise/backup/src/test/java/org/neo4j/commandline/admin/AugmentedCommandLocator.java
phughk: I don't have any in mind. @praveenag also suggested that I should use the service loader instead o this, should I remove this method and use that instead?

enterprise/backup/src/main/java/org/neo4j/backup/impl/ParameterisedOutsideWorld.java
```diff
@@ -41,7 +41,7 @@
     private final InputStream stdin;
     private final FileSystemAbstraction fileSystemAbstraction;
 
-    ParametrisedOutsideWorld( Console console, OutputStream stdout, OutputStream stderr, InputStream stdin, FileSystemAbstraction fileSystemAbstraction )
+    public ParameterisedOutsideWorld( Console console, OutputStream stdout, OutputStream stderr, InputStream stdin, FileSystemAbstraction fileSystemAbstraction )
```
chrisvest: Looks like this line is now too long for our checkstyle :)

enterprise/backup/src/main/java/org/neo4j/backup/impl/ParametrisedOutsideWorld.java
phughk: Resovled -> moved to `command-line` module `src/main` and kept `public`

PR #10847, opened 2018-01-18
Composite transaction state

MishaDemianenko:
Introduce possibility to split transactions state into multiple
to allow stable view of the state in some moment of a transaction lifespan.
Introduce version of Read API that performs read over stable version.
Introduce API to control split/combine of transaction state.
Ability to create composable states is disabled by default
and can be allowed by TransactionStatesContainer multiState feature flag.


MishaDemianenko:
After talk with @fickludd and @henriknyman closing this PR and fallback to version map approach

PR #10848, opened 2018-01-18
Fix bugs with nested pattern comprehensions

henriknyman:
- The legacy pattern matcher relied on determining bound entities at
runtime by inspecting the execution context in a way not compatible
with SlottedExecutionContext. This introduces a new method
`boundEntities` on ExecutionContext, with an
implementation that also works for SlottedExecutionContext.

- We now also do slot allocation for the projection expression in
a NestedPlanExpression, since it can contain more nested
plan expressions.


enterprise/cypher/physical-planning/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/SlotAllocation.scala
```diff
@@ -558,19 +575,6 @@ object SlotAllocation {
...
         slotConfig
 
-      case OuterHashJoin(nodes, _, _) =>
-        // A new pipeline is not strictly needed here unless we have batching/vectorization
-        recordArgument(lp)
-        val result = lhs.copy()
-        rhs.foreachSlotOrdered {
-          case (k, slot) if !nodes(k) =>
-            result.add(k, slot)
-          // If the column is one of the join columns there is no need to add it again
```
fickludd, line 568: Hehe, I know who forgot to remove this `case` after adding `NodeJoin`...

fickludd (approved):
Taking the opportunity for some optimization while bug-fixing? Nice.

I find it a little bit strange to have this getter/setter maps live in `SlotConfiguration` which is owned by the physical planner, and then be populated in the slotted runtime. But let's keep it until proven to be a problem.

PR #10849, opened 2018-01-19
Spike for faster recovery 

digitalstain:


jimwebber:
@digitalstain some JavaDoc on `Monitors#newMonitor` would be helpful guidance for future API users not to fall into the same unwary trap.

PR #10850, opened 2018-01-19
 Backup support for the fulltext addon

ragadeeshu:


community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/NeoStoreFileListing.java
```diff
@@ -67,21 +69,31 @@ public NeoStoreFileListing( File storeDir, LogFiles logFiles,
...
+        }
 
         placeMetaDataStoreLast( files );
 
-        return resourceIterator( files.iterator(),
-                new MultiResource( asList( labelScanStoreSnapshot, schemaIndexSnapshots, explicitIndexSnapshots ) ) );
+        return resourceIterator( files.iterator(), new MultiResource( additionalResources ) );
+    }
+
+    public void registerStoreFileProvider( ThrowingFunction<Collection<StoreFileMetadata>,Resource, IOException> provider )
```
chrisvest: I think this should have a dedicated interface, where it, for instance, can be documented what it means for a provider to throw when called.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomKernelExtension.java
```diff
@@ -85,6 +84,7 @@ public void start() throws IOException, KernelException
             provider.openIndex( BLOOM_NODES, NODES );
             provider.openIndex( BLOOM_RELATIONSHIPS, RELATIONSHIPS );
             provider.registerTransactionEventHandler();
+            provider.registerFileListing(fileListing);
```
chrisvest: I think we should unwrap the `fileListing` at this level, like we do with the `transactionIdStore`, so the rest of the code doesn't need to worry about suppliers.

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomBackupIT.java
```diff
@@ -0,0 +1,154 @@
...
+    @BeforeClass
+    public static void portSetup()
+    {
+        backupPort = allocatePort();
+    }
+
+    @Rule
+    public DatabaseRule db =
+            new EmbeddedDatabaseRule().withSetting( OnlineBackupSettings.online_backup_enabled, "true" ).withSetting( OnlineBackupSettings.online_backup_server,
+                    "127.0.0.1:" + backupPort ).withSetting( bloom_enabled, "true" );
```
chrisvest: Please tighten up the indentation.

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomBackupIT.java
```diff
@@ -0,0 +1,154 @@
...
+import static org.neo4j.ports.allocation.PortAuthority.allocatePort;
+
+public class BloomBackupIT
+{
+    private static int backupPort;
+
+    @BeforeClass
+    public static void portSetup()
+    {
+        backupPort = allocatePort();
```
chrisvest: Give each individual test case their own port. This means you'll have to make your database rule `startLazily` as well.

chrisvest (changes requested):
I had a couple of comments.

chrisvest (approved):

PR #10851, opened 2018-01-19
remove initial heapsize settings for neo4j scripts

praveenag:
Remove setting initial heapsize using HEAP_SIZE environment variable.
Note: It does **not** affect the initial heapsize set by neo4j config during neo4j startup.


spacecowboy (approved):
For anyone else who might view this PR, note that it does *not* affect the initial heapsize set by neo4j config during neo4j startup. That is still in place (and should remain so)

jimwebber:
@praveenag does this need to be forward merged given the backup differences in 3.4

praveenag:
@jimwebber yes this needs to be forward merged into 3.4. But that is a straightforward merge compared to the logging changes.
@spacecowboy Thanks for the note. I will update the PR to say the same.

chrisvest (approved):

praveenag:
@chrisvest and @spacecowboy There;s a new commit in this pr indicating for the help text. Please take that into consideration as well.

spacecowboy (approved):

PR #10852, opened 2018-01-19
PageCursor shiftBytes

burqen:
An efficient way to shift bytes within one page.

co-author: @chrisvest

### DIFF
--- a/community/io/src/main/java/org/neo4j/io/pagecache/impl/CompositePageCursor.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/impl/CompositePageCursor.java
@@ -451,7 +451,7 @@ public class CompositePageCursor extends PageCursor
     @Override
     public void shiftBytes( int sourceOffset, int length, int shift )
     {
-        throw new UnsupportedOperationException( "Composite cursor does not support copyTo functionality." );
+        throw new UnsupportedOperationException( "Composite cursor does not support shiftBytes functionality... yet." );
     }
 
     @Override
--- a/community/io/src/main/java/org/neo4j/io/pagecache/impl/CompositePageCursor.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/impl/CompositePageCursor.java
@@ -449,6 +449,12 @@ public class CompositePageCursor extends PageCursor
     }
 
     @Override
+    public void shiftBytes( int sourceOffset, int length, int shift )
+    {
+        throw new UnsupportedOperationException( "Composite cursor does not support copyTo functionality." );
+    }
+
+    @Override
     public boolean checkAndClearBoundsFlag()
     {
         boolean bounds = outOfBounds | first.checkAndClearBoundsFlag() | second.checkAndClearBoundsFlag();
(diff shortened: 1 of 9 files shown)


community/io/src/main/java/org/neo4j/io/pagecache/impl/CompositePageCursor.java
```diff
@@ -448,6 +448,12 @@ public int copyTo( int sourceOffset, PageCursor targetCursor, int targetOffset,
         throw new UnsupportedOperationException( "Composite cursor does not support copyTo functionality." );
     }
 
+    @Override
+    public void shiftBytes( int sourceOffset, int length, int shift )
+    {
+        throw new UnsupportedOperationException( "Composite cursor does not support copyTo functionality." );
```
tinwelint: Why not? Or it just doesn't support it _yet_ since there's no need?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -789,6 +789,74 @@ public int copyTo( int sourceOffset, PageCursor targetCursor, int targetOffset,
...
+        }
+        else
+        {
+            unsafeShiftRight( sourceEnd, sourceStart, length, shift );
+        }
+    }
+
+    private void unsafeShiftLeft( int fromPos, int toPos, int length, int shift )
+    {
+        int longSteps = length >> 3;
```
tinwelint, line 820: Looks cool, but a bit unnecessary since the compiler will convert a `length / Long.BYTES` into just that anyway, won't it?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -789,6 +789,74 @@ public int copyTo( int sourceOffset, PageCursor targetCursor, int targetOffset,
...
+        {
+            byte b = UnsafeUtil.getByte( pointer + fromPos );
+            UnsafeUtil.putByte( pointer + fromPos + shift, b );
+            fromPos++;
+        }
+    }
+
+    private void unsafeShiftRight( int fromPos, int toPos, int length, int shift )
+    {
+        int longSteps = length >> 3;
```
tinwelint, line 841: Looks cool, but a bit unnecessary since the compiler will convert a `length / Long.BYTES` into just that anyway, won't it?

tinwelint (commented):
Cool stuff. Have you benchmarked GBPTree on it?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
burqen: One would hope so! I suggest keeping it anyway, just to be sure.

community/io/src/main/java/org/neo4j/io/pagecache/impl/CompositePageCursor.java
burqen: Exactly! Updating copied error message.

tinwelint (approved):

PR #10853, opened 2018-01-19
Launder clothes, not exceptions

klaren:
Removes `Exceptions.launderedExceptions()`
* It hides important logic, very hard to derive what exceptions are actually thrown.
* It facilitates catching `Throwable`, which should be avoided and painful if you have to do it.
* More furrow description of why this pattern is bad: [Why we deprecated Throwables.propagate](https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate)

Removes `Exceptions.withSuppressed()`
* `.addSuppressed()` can be used instead.

Removes `Exception.withCause()`
* `.initCause()` can be used instead.

Removes `Exceptions.combine()`
* Hard to derive the actual outcome of the resulting message and cause chain.
* `.addSuppressed()` or `.initCause()` is most likely more appropriate.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/helpers/RunCarefully.java
+++ b/community/kernel/src/main/java/org/neo4j/helpers/RunCarefully.java
@@ -19,7 +19,9 @@
  */
 package org.neo4j.helpers;
 
-import static org.neo4j.helpers.Exceptions.combine;
+import java.util.LinkedList;
+import java.util.List;
+
 import static org.neo4j.helpers.collection.Iterables.asArray;
 
 public class RunCarefully
@@ -38,7 +40,7 @@ public class RunCarefully
 
     public void run()
     {
-        Throwable error = null;
+        List<Throwable> errors = new LinkedList<>();
 
         for ( Runnable o : operations )
         {
@@ -48,13 +50,15 @@ public class RunCarefully
             }
             catch ( RuntimeException e )
             {
-                error = combine( error, e );
+                errors.add( e );
             }
         }
 
-        if ( error != null )
+        if ( !errors.isEmpty() )
         {
-            throw new RuntimeException( error );
+            RuntimeException exception = new RuntimeException();
+            errors.forEach( exception::addSuppressed );
+            throw exception;
         }
     }
 }
--- a/community/kernel/src/main/java/org/neo4j/helpers/RunCarefully.java
+++ b/community/kernel/src/main/java/org/neo4j/helpers/RunCarefully.java
@@ -19,7 +19,7 @@
  */
 package org.neo4j.helpers;
 
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
 
 import static org.neo4j.helpers.collection.Iterables.asArray;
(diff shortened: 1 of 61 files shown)


enterprise/ha/src/test/java/org/neo4j/kernel/impl/ha/ParallelLifecycle.java
```diff
@@ -118,7 +115,7 @@ private void perform( Action action ) throws Exception
             }
             catch ( InterruptedException | ExecutionException e )
             {
-                exception = Exceptions.combine( exception, e );
+                exception = combine( exception, e );
```
MishaDemianenko: a bit of NPE's will happen inside of combine on the first exception

community/kernel/src/main/java/org/neo4j/helpers/RunCarefully.java
```diff
@@ -38,7 +40,7 @@ public RunCarefully( Iterable<Runnable> operations )
 
     public void run()
     {
-        Throwable error = null;
+        List<Throwable> errors = new LinkedList<>();
```
MishaDemianenko: please just use arraylist

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
```diff
@@ -547,7 +548,7 @@ private void checkIdScanCursorBounds( PageCursor cursor )
     /**
      * Marks this store as "not ok".
      */
-    void setStoreNotOk( Throwable cause )
+    void setStoreNotOk( RuntimeException cause )
```
MishaDemianenko, line 551: not sure we need to restring this to only runtime ex, should we?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/VersionAwareLogEntryReader.java
```diff
@@ -94,6 +94,10 @@ public LogEntry readLogEntry( SOURCE channel ) throws IOException
                         skipped = 0;
                     }
                 }
+                catch ( ReadPastEndException e )
+                {   // Make these exceptions slip by straight out to the outer handler
+                    throw e;
```
MishaDemianenko, line 99: should we just return here instead of outer catch block then?

MishaDemianenko (changes requested):
Couple of things here and there great otherwise

community/kernel/src/main/java/org/neo4j/kernel/impl/store/CommonAbstractStore.java
klaren: If anyone needs that it's easy to refactor.

klaren:
@MishaDemianenko New commits :tada:

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/entry/VersionAwareLogEntryReader.java
klaren: I think we leave it like this to minimize the number of exit points.

MishaDemianenko (approved):

PR #10854, opened 2018-01-19
Make tests a bit more robust

OliviaYtterbrink:
Sometimes we would get a broken pipe on ibm-jdk when doing login
with faulty credentials, this should make these tests less
likely to fail.


Lojjs (approved):
Looks good to me!

PR #10855, opened 2018-01-19
Solve flaky RaftMessageProcessingMetricTest

RagnarW:
Value is time dependant and so its possible that it goes above 5
if there is for example a GC pause.


andrewkerr9000:
I suggest making this fix against 3.2, where this test was introduced. 3.3 builds have been reported flaky too.

andrewkerr9000:
They delta is I think needed due to floating point arithmetic precision loss. Nothing is actually timed. It may be prudent to add a delta to line 52 too.

andrewkerr9000:
Closing this PR: I'll open another against 3.2.

PR #10856, opened 2018-01-19
Use new API for create node

pontusmelke:
Turns out we missed porting `createNode` only did `createNodeId` this PR fixes that.

I run some of the `CreateX` benchmarks and this seem to be slightly faster than existing implementation.


SaschaPeukert (approved):
Looks good! Nice find!

PR #10857, opened 2018-01-20
Use new API for index operations in Core API

pontusmelke:


community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -611,34 +616,63 @@ private InternalTransaction beginTransactionInternal( KernelTransaction.Type typ
...
+                // weird at this point but ignore and fallback to a label scan
             }
         }
-        catch ( KernelException e )
-        {
-            // weird at this point but ignore and fallback to a label scan
-        }
+
 
         return getNodesByLabelAndPropertyWithoutIndex( propertyId, value, statement, labelId );
```
henriknyman: It looks like this statement is not being released

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeValueIndexCursor.java
```diff
@@ -91,8 +96,7 @@ public void initialize( IndexDescriptor descriptor, IndexProgressor progressor,
 
     private boolean isRemoved( long reference )
     {
-        return (changes != null && changes.isRemoved( reference )) ||
-               (read.hasTxStateWithChanges() && read.txState().addedAndRemovedNodes().isRemoved( reference ));
+        return removed.contains( reference );
```
henriknyman, line 99: Is it safe to cache this only at query-time?

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -611,34 +616,63 @@ private InternalTransaction beginTransactionInternal( KernelTransaction.Type typ
...
-        try
-        {
-            if ( null != descriptor )
+                    @Override
+                    public void close()
+                    {
+                        if ( !closed )
+                        {
+                            statement.close();
+                            cursor.close();
```
chrisvest: Use `IOUtil.closeAll` (or maybe introduce a `closeBoth` if you don't want to allocate the variadic array) or something like that, to ensure that the cursor is closed, even if `statement.close` throws an exception.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -930,16 +904,19 @@ public GraphPropertiesProxy newGraphPropertiesProxy()
...
-        private final NodeLabelIndexCursor cursor;
+
+        private final CURSOR cursor;
         private final Statement statement;
         private long next;
         private boolean closed;
         private static final long NOT_INITIALIZED = -2L;
 
-        NodeCursorResourceIterator( NodeLabelIndexCursor cursor, Statement statement )
+        NodeCursorResourceIterator( CURSOR cursor, Statement statement )
```
chrisvest, line 895: There's a problem in the `close` method of this class. Please make it use `IOUtil.closeAll` (or maybe introduce a `closeBoth` if you don't want to allocate the variadic array) or something like that, to ensure that the cursor is closed, even if `statement.close` throws an exception.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/NodeValueIndexCursor.java
fickludd: Hmm, not sure I know what you mean... could you explain why it would/wouldn't be safe?

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
fickludd: Fixed using `IOUtil.closeAll`

PR #10841, opened 2018-01-18
Tx-state aware RelationshipTraversalCursor

fickludd:
Add tx-state awareness for relationship traversal code-paths starting with `NodeCursor.allRelationships` and `NodeCursor.allRelationshipsReference`. Also refactors traversal tests for reuse between tx-state tests and pure store tests.

Builds on #10833, first commit is `efd5695`.

PR #10859, opened 2018-01-20
Tx-state aware grouped traversal

fickludd:
Complete transaction state awareness in Kernel API traversal. This PR add support for tx-state awareness when traversing via `NodeCursor.relationships( RelationshipGroupCursor )` and `NodeCursor.relationshipGroupReference`. Note that the `RelationshipGroupCursor.*count()` methods on are still not tx-state aware. We have a card for this.

Builds on #10841. First commit is `1184f3a`. The ivory tower goes higher.


community/kernel/src/main/java/org/neo4j/storageengine/api/txstate/NodeState.java
```diff
@@ -59,5 +60,9 @@ void visitLabelChanges( long nodeId, Set<Integer> added, Set<Integer> removed )
 
     PrimitiveLongIterator getAddedRelationships( Direction direction, int[] relTypes );
 
+    PrimitiveLongIterator getAddedRawRelationships();
```
tinwelint: What is "raw relationships"? Please take the opportunity to describe that in javadocs.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipDirection.java
```diff
@@ -0,0 +1,47 @@
...
+ *  a.getRelationships( INCOMING ) => r2
+ *  a.getRelationships( LOOP ) => r3
+ *  a.getRelationships( ALL ) => r1, r2, r3
+ *
+ * Note that this contrasts with /** @see #org.neo4j.graphdb.Direction(long), where
+ *  a.getRelationships( org.neo4j.graphdb.Direction.OUTGOING ) => r1, r3
+ *  a.getRelationships( org.neo4j.graphdb.Direction.INCOMING ) => r2, r3
+ *  a.getRelationships( org.neo4j.graphdb.Direction.BOTH ) => r1, r2, r3
+ */
+public enum RelationshipDirection
```
tinwelint, line 40: Yey, our third direction enum! I'm guessing the `storageengine` direction will go away soon though, or hopefully at least.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipDirection.java
```diff
@@ -31,7 +31,6 @@
  *  a.getRelationships( OUTGOING ) => r1
  *  a.getRelationships( INCOMING ) => r2
  *  a.getRelationships( LOOP ) => r3
- *  a.getRelationships( ALL ) => r1, r2, r3
```
tinwelint: Would be great w/ explanation of what `ERROR` is here

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
```diff
@@ -299,4 +280,113 @@ private void traverseWithoutGroups( RelationshipTestSupport.StartNode start, boo
...
+                tx.dataWrite().relationshipCreate( start.id, label, start.id );
+                break;
+            default:
+                throw new IllegalStateException( "Oh ye be cursed, foul checkstyle!" );
+            }
+            tx.dataWrite().relationshipDelete( head.id );
+            expectedCounts.put( kv.getKey(), rs.size() + 1 );
+        }
+
+        String newLabelName = "NEW";
```
tinwelint: I know I'll probably push your buttons here, but can we try to limit `label` for relationship type to a minimum? It looks especially weird since the methods has the term `relationshipType`. A variable called something... label primes the mind that it is a node label and you'll have to look at usages to figure out whether it's a node label or relationship type.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipTraversalCursor.java
```diff
@@ -82,6 +82,30 @@ boolean check( long source, long target, long origin )
...
+            case OUTGOING:
+                return FilterState.OUTGOING;
+            case INCOMING:
+                return FilterState.INCOMING;
+            case LOOP:
+                return FilterState.LOOP;
+            case ERROR:
+                throw new IllegalArgumentException( "There has been a RelationshipDirection.ERROR" );
+            default:
+                throw new IllegalStateException( "Still poking my eye, dear checkstyle..." );
```
tinwelint: Cute, but something more informative, like including `direction` too?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
```diff
@@ -193,30 +200,6 @@ public void shouldSeeRelationshipInTransactionBeforeCursorInitialization() throw
         }
     }
 
-//    @Test
-//    public void shouldTraverseSparseNodeViaGroups() throws Exception
-//    {
```
tinwelint, line 199: Some left-overs

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTestSupport.java
```diff
@@ -25,7 +25,11 @@
 import java.util.Map;
 import java.util.function.Function;
 
-import org.neo4j.graphdb.*;
+import org.neo4j.graphdb.Direction;
+import org.neo4j.graphdb.GraphDatabaseService;
+import org.neo4j.graphdb.Node;
```
tinwelint: Time to left IntelliJ know about this rule?

tinwelint (changes requested):

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTestSupport.java
fickludd: Strange thing is I never got that before, and fixed it with `optimize imports`, so I think IntelliJ knows. Dunno if I somehow managed to switch to default code style for a period when that happened...

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipDirection.java
fickludd: I don't really see the point of the `storageengine` one, as it's a copy of the `graphdb` direction. This one at least has different (and described) semantics. But yeah, more enums can't be bad, right?

tinwelint (approved):
You still have some commented tests in `RelationshipTransactionStateTestBase.java`, but I'll approve anyway

PR #10860, opened 2018-01-22
Creates a new operator NodeRightOuterHashJoin

sherfert:
The motivation to have this operator is that a LeftOuterJoin is pretty expensive when the LHS has a much higher cardinality than the RHS (because building the probe table is more expensive than probing it). Having both LeftOuterJoin and RightOuterJoin enables planning both and let the cost planning estimate which one is cheaper.

This implements both planning of the operator and the interpreted pipe. Slotted will use a fallback pipe
and compiled is not supporting outer joins at all.

It also changes the name of previously `NodeOuterHashJoin` to `NodeLeftOuterHashJoin`, which is
what the operator actually is.

The pipe builds the probe table from the LHS (without nulls) and probes each row from the RHS. Null rows and rows without matches in the probe table are passed through.


community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/steps/solveOptionalMatches.scala
```diff
@@ -54,7 +54,30 @@ case object outerHashJoin extends OptionalSolver {
...
-      Some(context.logicalPlanProducer.planOuterHashJoin(joinNodes, lhs, rhs, solvedHints, context))
+      Some(context.logicalPlanProducer.planLeftOuterHashJoin(joinNodes, lhs, rhs, solvedHints, context))
+    } else {
+      None
+    }
+  }
+}
+
+case object rightOuterHashJoin extends OptionalSolver {
+  override def apply(optionalQg: QueryGraph, rhs: LogicalPlan, context: LogicalPlanningContext, solveds: Solveds, cardinalities: Cardinalities) = {
```
systay: The code here looks very similar to the code above in the same file. Are there not opportunities for sharing it?

community/cypher/cypher-planner-3.4/src/test/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/OptionalMatchPlanningIntegrationTest.scala
```diff
@@ -43,10 +44,89 @@ class OptionalMatchPlanningIntegrationTest extends CypherFunSuite with LogicalPl
...
+  }
+
+  test("should choose left outer join if lhs has small cardinality") {
+    (new given {
+      labelCardinality = Map("X" -> 1.0, "Y" -> 10.0)
+      statistics = new DelegatingGraphStatistics(parent.graphStatistics) {
+        override def cardinalityByLabelsAndRelationshipType(fromLabel: Option[LabelId], relTypeId: Option[RelTypeId], toLabel: Option[LabelId]): Cardinality = {
+          // TODO proper lookup from semantic table somehow
+          // X = 0, Y = 1
+          if (fromLabel.exists(_.id == 0) && relTypeId.isEmpty && toLabel.isEmpty) /*toLabel.exists(_.id == 2)*/ {
```
systay: Looks like some comments should have been removed from this test class, right?

systay (approved):
Nice stuff! Would be interesting to work on the cost model and make sure these operators are used when they should.

PR #10861, opened 2018-01-22
Add "Page cache usage" metric

klaren:
Add monitoring of "page cache usage" that will help with tuning of the page cache size.

- This adds the metric `neo4j.page_cache.usage_ratio` that returns a number between `0.0` and `1.0`. For example, a value of `0.5` means that half of the pages are used and the rest are available for data without evictions. 
- It's also exposed the same metric through the JMX bean "Page cache" with the method `getUsageRatio()`.

### DIFF
--- a/community/io/src/test/java/org/neo4j/io/pagecache/tracing/DefaultPageCacheTracerTest.java
+++ b/community/io/src/test/java/org/neo4j/io/pagecache/tracing/DefaultPageCacheTracerTest.java
@@ -124,6 +124,7 @@ public class DefaultPageCacheTracerTest
     @Test
     public void usageRatio()
     {
+        assertThat( tracer.usageRatio(), is( Double.NaN ) );
         tracer.maxPages( 10 );
         assertThat( tracer.usageRatio(), closeTo( 0d, 0.0001 ) );
         tracer.faults( 5 );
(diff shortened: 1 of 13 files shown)


community/io/src/test/java/org/neo4j/io/pagecache/tracing/DefaultPageCacheTracerTest.java
```diff
@@ -121,6 +121,20 @@ public void shouldCalculateHitRatio() throws Exception
         assertThat( "hitRation", tracer.hitRatio(), closeTo( 3.0 / 10, 0.0001 ) );
     }
 
+    @Test
+    public void usageRatio()
+    {
+        tracer.maxPages( 10 );
```
chrisvest: How about an isNaN test for when maxPages has not yet been set?

enterprise/management/src/main/java/org/neo4j/management/PageCache.java
```diff
@@ -57,4 +57,7 @@
                   "This number should be zero, or at least not growing, in a healthy database. " +
                   "Otherwise it could indicate drive failure, storage space, or permission problems." )
     long getEvictionExceptions();
+
+    @Description( "The percentage of used pages." )
+    double getUsageRatio();
```
chrisvest: Can you get around the revapi complaint by giving this a default implementation? I think the NaN possibility should be mentioned here as well.

chrisvest (commented):
Looks good overall.

enterprise/management/src/main/java/org/neo4j/management/PageCache.java
klaren: I could probably get around it by providing a default implementation.

klaren:
@chrisvest A new commit addressing your comments.

chrisvest (approved):
If the build goes green, then I'm happy with this.

PR #10863, opened 2018-01-22
Allow Remove property with case expression

sherfert:
changelog:
Bugfix so that the cost planner can now plan the following:
REMOVE CASE WHEN ... THEN a ELSE b END.prop


Lojjs (approved):
Looks good!

PR #10864, opened 2018-01-22
Update LdbcAcceptanceTest queries

henriknyman:
Update with the latest versions of the LDBC Cypher queries


henriknyman:
@alexaverbuch *poke*

PR #10865, opened 2018-01-22
SimpleIdProvider zap page before releasing page

burqen:
This is something that the real freelist already does and
we rely on this behaviour.


tinwelint (approved):
LGTM

PR #10867, opened 2018-01-22
Fix in normalizeReturnClauses with same variable names

sherfert:
changelog:
Fixes bug with incorrect ordering when aliasing.
Queries such as `match (n),(m) return m as m2, n as m order by m` were rewritten incorrectly to sort by the wrong column `m as m2`. Now, the result is correctly ordered by `n as m`.


Lojjs (approved):
Nice fix, good with comments and tests.

PR #10868, opened 2018-01-22
Small import nuggets

tinwelint:
Found to improve various small things while working on importing really large data sets.

### DIFF
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -83,7 +83,6 @@ import static org.neo4j.helpers.Exceptions.launderedException;
 import static org.neo4j.helpers.Format.bytes;
 import static org.neo4j.helpers.Strings.TAB;
 import static org.neo4j.helpers.TextUtil.tokenizeStringWithQuotes;
-import static org.neo4j.io.ByteUnit.gibiBytes;
 import static org.neo4j.io.ByteUnit.mebiBytes;
 import static org.neo4j.io.fs.FileUtils.readTextFile;
 import static org.neo4j.kernel.configuration.Settings.parseLongWithUnit;
@@ -111,12 +110,6 @@ import static org.neo4j.unsafe.impl.batchimport.input.csv.DataFactories.defaultF
  */
 public class ImportTool
 {
-    /**
-     * A high heap size that no import should need to go beyond. Also it's small enough such that object pointers
-     * can be compressed to 4 bytes.
-     */
-    private static final long RECOMMENDED_MAX_HEAP_SIZE = gibiBytes( 28 );
-
     private static final String INPUT_FILES_DESCRIPTION =
             "Multiple files will be logically seen as one big file " +
             "from the perspective of the importer. " +
@@ -392,11 +385,6 @@ public class ImportTool
     {
         System.err.println( format( "WARNING: neo4j-import is deprecated and support for it will be removed in a future%n" +
                 "version of Neo4j; please use neo4j-admin import instead." ) );
-        if ( Runtime.getRuntime().maxMemory() > RECOMMENDED_MAX_HEAP_SIZE )
-        {
-            System.err.println( format( "WARNING: heap size is unnecessarily big, potentially causing garbage collection issues and%n" +
-                    "less room for importer off-heap caches. A heap of maximum %s is recommended", bytes( RECOMMENDED_MAX_HEAP_SIZE ) ) );
-        }
 
         PrintStream out = System.out;
         PrintStream err = System.err;
--- a/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
+++ b/community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
@@ -73,6 +73,7 @@ import org.neo4j.unsafe.impl.batchimport.input.csv.Decorator;
 import org.neo4j.unsafe.impl.batchimport.input.csv.IdType;
 import org.neo4j.unsafe.impl.batchimport.staging.ExecutionMonitors;
 
+import static java.lang.String.format;
 import static java.nio.charset.Charset.defaultCharset;
 import static java.util.Arrays.asList;
 
@@ -82,6 +83,7 @@ import static org.neo4j.helpers.Exceptions.launderedException;
 import static org.neo4j.helpers.Format.bytes;
 import static org.neo4j.helpers.Strings.TAB;
 import static org.neo4j.helpers.TextUtil.tokenizeStringWithQuotes;
+import static org.neo4j.io.ByteUnit.gibiBytes;
 import static org.neo4j.io.ByteUnit.mebiBytes;
 import static org.neo4j.io.fs.FileUtils.readTextFile;
 import static org.neo4j.kernel.configuration.Settings.parseLongWithUnit;
@@ -109,6 +111,12 @@ import static org.neo4j.unsafe.impl.batchimport.input.csv.DataFactories.defaultF
  */
 public class ImportTool
 {
+    /**
+     * A high heap size that no import should need to go beyond. Also it's small enough such that object pointers
+     * can be compressed to 4 bytes.
+     */
+    private static final long RECOMMENDED_MAX_HEAP_SIZE = gibiBytes( 28 );
+
     private static final String INPUT_FILES_DESCRIPTION =
             "Multiple files will be logically seen as one big file " +
             "from the perspective of the importer. " +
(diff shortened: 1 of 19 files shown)


community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -109,6 +111,12 @@
  */
 public class ImportTool
 {
+    /**
+     * A high heap size that no import should need to go beyond. Also it's small enough such that object pointers
+     * can be compressed to 4 bytes.
+     */
+    private static final long RECOMMENDED_MAX_HEAP_SIZE = gibiBytes( 28 );
```
MishaDemianenko: why exactly 28 and not like 29 for example?

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -382,8 +390,14 @@ public static void main( String[] incomingArguments ) throws IOException
      */
     public static void main( String[] incomingArguments, boolean defaultSettingsSuitableForTests ) throws IOException
     {
-        System.err.println("WARNING: neo4j-import is deprecated and support for it will be removed in a future\n" +
-                "version of Neo4j; please use neo4j-admin import instead.\n");
+        System.err.println( format( "WARNING: neo4j-import is deprecated and support for it will be removed in a future%n" +
+                "version of Neo4j; please use neo4j-admin import instead." ) );
+        if ( Runtime.getRuntime().maxMemory() > RECOMMENDED_MAX_HEAP_SIZE )
```
MishaDemianenko: those recommendations should be correlated with the recommendation that memrec provides

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
```diff
@@ -382,8 +390,14 @@ public static void main( String[] incomingArguments ) throws IOException
...
+        {
+            System.err.println( format( "WARNING: heap size is unnecessarily big, potentially causing garbage collection issues and%n" +
```
MishaDemianenko: its not clear about what gc issues we are talking about here?

community/import-tool/src/test/java/org/neo4j/tooling/QuickImport.java
```diff
@@ -161,6 +161,7 @@ public long maxMemoryUsage()
             }
             else
             {
+                System.out.println( "Seed " + randomSeed );
```
MishaDemianenko, line 164: looks like debug output

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/DynamicProcessorAssigner.java
```diff
@@ -51,7 +51,7 @@
 
     public DynamicProcessorAssigner( Configuration config )
     {
-        super( 100, MILLISECONDS );
+        super( 1, SECONDS );
```
MishaDemianenko, line 54: magic?

tools/src/main/java/org/neo4j/tools/dump/DumpStore.java
```diff
@@ -58,6 +60,29 @@
  */
 public class DumpStore<RECORD extends AbstractBaseRecord, STORE extends RecordStore<RECORD>>
 {
+    private static class IdRange
```
MishaDemianenko, line 63: would be nice to get java doc that explains format in which idRange should be supplied

MishaDemianenko (changes requested):

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
tinwelint: You mean some official memory recommendations? Where are those? Also, this is different, import needs virtually no heap at all... typically only max a couple of gigs. For very large stores the page cache meta data will take up a big chunk of the heap tho so therefore this is a bit higher. The limit is meant to be set to 32GiB, but being on the safe side such that object pointers can still fit in 4 bytes. If that limit is optimally 29GiB, so be it, I can change it, but I don't think it matters much. 28-29 GiB is already too high for an import, unless it's in the several hundred billions.

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
tinwelint: Long GC pauses if running into full GC. Perhaps it could be framed differently, without mentioning vague GC issues.

community/import-tool/src/test/java/org/neo4j/tooling/QuickImport.java
tinwelint: QuickImport is a test utility, lives in src/test. It now prints the seed it uses to generate its data, which is very helpful for debugging.

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/DynamicProcessorAssigner.java
tinwelint: As magic as 100 ms. It was just checked unnecessarily often.

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
MishaDemianenko: i would assume you referring to https://wiki.openjdk.java.net/display/HotSpot/CompressedOops and you want to keep oops to be 32 bits and never cross that boundary right? exactly with that limitations/suggestions in mind our recent PR https://github.com/neo4j/neo4j/pull/10821 will provide some recommended heap settings. And memrec that I'm referring to in the above comment is exactly that tool.
The problem here is different recommendations that popping up from different tools that misguide and confuse users. Would be really nice to avoid that.

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
MishaDemianenko: Maybe here we can warn instead about crossing compact heap representation boundary and a general recommendation to keep heap as small as possible if you not sure that you need that much?

community/import-tool/src/test/java/org/neo4j/tooling/QuickImport.java
MishaDemianenko: oh, please nevermind then

community/import-tool/src/main/java/org/neo4j/tooling/ImportTool.java
tinwelint: Yes, that's what I'm referring to.

`memrec` doesn't apply here because the use case is so different from running a database. E.g. in importer the page cache has an optimal size of 500MiB - 1GiB and heap around a couple of gigabytes for most datasets, virtually regardless of your machine spec. For really huge datasets the page cache still has the same recommended size and heap a little higher.

tinwelint:
Ended up reworking the memory calculation and warnings and that

community/import-tool/src/test/java/org/neo4j/tooling/PrintingImportLogicMonitorTest.java
```diff
@@ -0,0 +1,145 @@
...
+    private final ByteArrayOutputStream outBuffer = new ByteArrayOutputStream();
+    private final PrintStream out = new PrintStream( outBuffer );
+    private final ByteArrayOutputStream errBuffer = new ByteArrayOutputStream();
+    private final PrintStream err = new PrintStream( errBuffer );
+    private final ImportLogic.Monitor monitor = new PrintingImportLogicMonitor( out, err );
+
+    @After
+    public void after()
+    {
+        System.out.println( "out " + outBuffer );
```
MishaDemianenko: we do not need to print those buffers after tests

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/HeapSizeSanityChecker.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static org.neo4j.kernel.impl.util.OsBeanUtil.VALUE_UNAVAILABLE;
+import static org.neo4j.unsafe.impl.batchimport.ImportMemoryCalculator.estimatedCacheSize;
+import static org.neo4j.unsafe.impl.batchimport.ImportMemoryCalculator.optimalMinimalHeapSize;
+
+/**
+ * Sanity checking of {@link org.neo4j.unsafe.impl.batchimport.input.Input.Estimates} against heap size and free memory.
+ * Registers warnings onto a {@link ImportLogic.Monitor}.
+ */
+public class HeapSizeSanityChecker
```
MishaDemianenko: class and methods in this class can be package private

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/HeapSizeSanityChecker.java
```diff
@@ -0,0 +1,86 @@
...
+
+        // Check if the heap is big enough to handle the import
+        if ( actualHeapSize < optimalMinimalHeapSize )
+        {
+            monitor.insufficientHeapSize( optimalMinimalHeapSize, actualHeapSize );
+            return; // user have been warned about heap size issue
+        }
+
+        // Check if heap size could be tweaked
+        if ( ((freeMemoryIsKnown && freeMemory < estimatedCacheSize) || !freeMemoryIsKnown) && actualHeapSize > optimalMinimalHeapSize * 1.2 )
```
MishaDemianenko: expression can be simplified to `( (!freeMemoryIsKnown || freeMemory < estimatedCacheSize) && (actualHeapSize > (optimalMinimalHeapSize * 1.2)) )`

MishaDemianenko (changes requested):
couple of minor things

MishaDemianenko (approved):

PR #10869, opened 2018-01-23
Use StringBuilder instead of StringBuffer

MishaDemianenko:


andreikoval (approved):

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/Envelope.java
```diff
@@ -286,22 +284,22 @@ public void translate( double[] offset )
...
     public String toString()
     {
         return "Envelope: min=" + makeString(min) + ", max=" + makeString(max);
     }
 
-    // Private method
     private static String makeString( double[] vals )
     {
-        StringBuffer sb = new StringBuffer();
+        StringBuilder sb = new StringBuilder();
```
lutovich, line 295: This is probably a good candidate for stream & joining collector

community/shell/src/main/java/org/neo4j/shell/apps/Man.java
```diff
@@ -198,10 +198,9 @@ public static void printHelpString( Output out, ShellServer server, boolean list
         return availableCommands;
     }
 
-    private static synchronized String availableCommandsAsString(
-        ShellServer server )
+    private static synchronized String availableCommandsAsString( ShellServer server )
     {
-        StringBuffer commands = new StringBuffer();
+        StringBuilder commands = new StringBuilder();
```
lutovich, line 203: Maybe use stream & joining collector instead?

MishaDemianenko: talked offline about it

PR #10872, opened 2018-01-23
Fix pattern comprehension match error

henriknyman:
Missing a fallback case in SlottedExecutionContext `boundEntities`.


fickludd (approved):
Good find, LGTM.

PR #10877, opened 2018-01-23
Extend Kernel API with stable/active transaction state

fickludd:
The transaction now has two Read entry points: dataRead() and stableDataRead(). The
instance returned by stableDataRead() will not see any in-transaction writes which
have happened since the last Transaction.markAsStable(). To allow Cypher to code
against this functionality before the implementation exists, a feature toggle is
added to the Kernel API in Kernel.Modes.twoLayerTransactionState().

This commit also adds feature toggle dependent test of some of the expected
functionality in TwoLayerTxStateTestBase.

### DIFF
--- a/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Modes.java
+++ b/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Modes.java
@@ -19,16 +19,15 @@
  */
 package org.neo4j.internal.kernel.api;
 
-import org.neo4j.internal.kernel.api.security.SecurityContext;
-
 /**
- * The Kernel.
+ * Define Kernel modes.
  */
-public interface Kernel
+public interface Modes
 {
-    CursorFactory cursors();
-
-    Session beginSession( SecurityContext securityContext );
-
-    Modes modes();
+    /**
+     * True if the kernel supports the two-layer transaction state. If false, both {@link Transaction#dataRead()}
+     * and {@link Transaction#stableDataRead()} will refer to the same active instance, and
+     * {@link Transaction#markAsStable()} will be ignored.
+     */
+    boolean twoLayerTransactionState();
 }
--- a/community/kernel/src/test/java/org/neo4j/kernel/builtinprocs/StubKernelTransaction.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/builtinprocs/StubKernelTransaction.java
@@ -74,6 +74,18 @@ public class StubKernelTransaction implements KernelTransaction
     }
 
     @Override
+    public Read stableDataRead()
+    {
+        return null;
+    }
+
+    @Override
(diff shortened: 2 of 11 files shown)


community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Kernel.java
```diff
@@ -29,4 +29,6 @@
     CursorFactory cursors();
 
     Session beginSession( SecurityContext securityContext );
+
+    Modes modes();
```
thobe, line 33: What are we anticipating that we are going to do based on this `Modes`?
Some documentation about its _impact_ might be good.

community/kernel/src/test/java/org/neo4j/kernel/builtinprocs/StubKernelTransaction.java
```diff
@@ -73,6 +73,18 @@ public Read dataRead()
         throw new UnsupportedOperationException( "not implemented" );
     }
 
+    @Override
+    public Read stableDataRead()
+    {
+        return null;
```
thobe, line 79: If we expect this to not be invoked, prefer throwing an exception.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Modes.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.internal.kernel.api;
+
+/**
+ * Define Kernel modes.
+ */
+public interface Modes
```
thobe, line 25: Perhaps this should be an enum?

thobe (approved):
Looks like a simple improvement, and a step in the right direction.

MishaDemianenko (approved):
No additional comments from my side

community/kernel/src/test/java/org/neo4j/kernel/builtinprocs/StubKernelTransaction.java
fickludd: I agree, missed that here.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Modes.java
fickludd: Maybe? Let's change it when there's a need.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/Kernel.java
fickludd: There's some documentation on the actual Modes method, not enough? Can write some more on the next drive by.

PR #10878, opened 2018-01-23
Removes NativeSchemaIndexReader indirection between new and old kernel API

tinwelint:
Seems to have caused a performance regression of ~15% due to one added
layer of indirection in each call to iterator.next().

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberHitIterator.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberHitIterator.java
@@ -31,8 +31,6 @@ import org.neo4j.index.internal.gbptree.Hit;
 
 /**
  * Wraps number key/value results in a {@link PrimitiveLongIterator}.
- * The {@link RawCursor seeker} which gets passed in will have to be closed somewhere else because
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberHitIterator.java
```diff
@@ -0,0 +1,94 @@
...
+import org.neo4j.collection.primitive.PrimitiveLongCollections;
+import org.neo4j.collection.primitive.PrimitiveLongIterator;
+import org.neo4j.collection.primitive.PrimitiveLongResourceIterator;
+import org.neo4j.cursor.RawCursor;
+import org.neo4j.index.internal.gbptree.Hit;
+
+/**
+ * Wraps number key/value results in a {@link PrimitiveLongIterator}.
+ * The {@link RawCursor seeker} which gets passed in will have to be closed somewhere else because
+ * the {@link PrimitiveLongIterator} is just a plain iterator, no resource.
```
burqen: This is not true.

burqen (approved):
Cool! Just update the javadoc and I'm happy :D

PR #10884, opened 2018-01-24
Unflake ImportToolTest#shouldSkipDuplicateNodesIfToldTo

tinwelint:
Which is failing on ibm and zulu JDKs only. Could be some issue about ordering
assumptions. Made the test less hung up on specific order of nodes.


chrisvest (approved):
👍

PR #10885, opened 2018-01-24
Fix slotted shortest path predicate planning

henriknyman:
Fixes shortest path with predicates that depends on the full path

Computing which predicates has variable dependencies on the full path
of a shortest path expression was done on runtime expressions,
and did not work for slotted runtime. 
This moves the computation to be done on the logical plan expression at pipe building time.

- Update neo4j-cypher-compiler-3.3 compatibility version to 3.3.2


fickludd (approved):
Looking good.

PR #10887, opened 2018-01-24
Property level security

OliviaYtterbrink:
Added ability to blacklist property reading for specified roles.

Currently only properties on nodes are filtered due to relationship not using the `PropertyCursor` yet.

Also clears up intent of `SecurityContext` by splitting it into `LoginContext` which handles authentication and a `SecurityContext` that holds the authorization of a user.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
@@ -34,6 +34,7 @@ import org.neo4j.collection.primitive.PrimitiveLongResourceCollections;
 import org.neo4j.collection.primitive.PrimitiveLongResourceIterator;
 import org.neo4j.cursor.Cursor;
 import org.neo4j.graphdb.Direction;
+import org.neo4j.helpers.collection.Iterators;
 import org.neo4j.internal.kernel.api.IndexQuery;
 import org.neo4j.internal.kernel.api.InternalIndexState;
 import org.neo4j.internal.kernel.api.exceptions.InvalidTransactionTypeKernelException;
@@ -690,7 +691,9 @@ public class OperationsFacade
     public Iterator<Token> propertyKeyGetAllTokens()
     {
         statement.assertOpen();
-        return tokenRead().propertyKeyGetAllTokens( statement );
+        AccessMode mode = tx.securityContext().mode();
+        return Iterators.stream( tokenRead().propertyKeyGetAllTokens( statement ) ).
+                filter( propKey -> mode.allowsPropertyReads( propKey.id() ) ).iterator();
     }
 
     @Override
--- a/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
+++ b/enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
@@ -25,8 +25,10 @@ import org.apache.shiro.realm.Realm;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -171,7 +173,8 @@ public class EnterpriseSecurityModule extends SecurityModule
         }
 
         return new MultiRealmAuthManager( internalRealm, orderedActiveRealms, createCacheManager( config ),
-                securityLog, config.get( SecuritySettings.security_log_successful_authentication ) );
+                securityLog, config.get( SecuritySettings.security_log_successful_authentication ),
+                securityConfig.propertyAuthorization, securityConfig.propertyBlacklist );
     }
 
     private static List<Realm> selectOrderedActiveRealms( List<String> configuredRealms, List<Realm> availableRealms )
@@ -318,12 +321,12 @@ public class EnterpriseSecurityModule extends SecurityModule
         return new FileUserRepository( fileSystem, getDefaultAdminRepositoryFile( config ), logProvider );
     }
 
-    public static File getRoleRepositoryFile( Config config )
+    private static File getRoleRepositoryFile( Config config )
     {
         return new File( config.get( DatabaseManagementSystemSettings.auth_store_directory ), ROLE_STORE_FILENAME );
     }
 
-    public static File getDefaultAdminRepositoryFile( Config config )
+    private static File getDefaultAdminRepositoryFile( Config config )
     {
         return new File( config.get( DatabaseManagementSystemSettings.auth_store_directory ),
                 DEFAULT_ADMIN_STORE_FILENAME );
@@ -334,7 +337,7 @@ public class EnterpriseSecurityModule extends SecurityModule
         return new IllegalArgumentException( "Illegal configuration: " + message );
     }
 
-    class SecurityConfig
+    static class SecurityConfig
     {
         final List<String> authProviders;
         final boolean hasNativeProvider;
@@ -346,6 +349,9 @@ public class EnterpriseSecurityModule extends SecurityModule
         final boolean ldapAuthorization;
         final boolean pluginAuthentication;
         final boolean pluginAuthorization;
+        final boolean propertyAuthorization;
+        private final String propertyAuthMapping;
+        final Map<String,List<String>> propertyBlacklist = new HashMap<>();
 
         SecurityConfig( Config config )
         {
@@ -362,6 +368,8 @@ public class EnterpriseSecurityModule extends SecurityModule
             ldapAuthorization = config.get( SecuritySettings.ldap_authorization_enabled );
             pluginAuthentication = config.get( SecuritySettings.plugin_authentication_enabled );
             pluginAuthorization = config.get( SecuritySettings.plugin_authorization_enabled );
+            propertyAuthorization = config.get( SecuritySettings.property_level_authorization_enabled );
+            propertyAuthMapping = config.get( SecuritySettings.property_level_authorization_permissions );
         }
 
         void validate()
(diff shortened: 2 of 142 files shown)


community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -456,7 +482,13 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
         val buffer = ArrayBuffer[Int]()
         nodeCursor.properties(propertyCursor)
         while (propertyCursor.next()) {
-          buffer.append(propertyCursor.propertyKey())
+          val propertyKeyId = propertyCursor.propertyKey()
+          try {
+            transactionalContext.readOperations.propertyKeyGetName(propertyKeyId)
+            buffer.append(propertyKeyId)
+          } catch {
```
henriknyman: Using exceptions for this seems a bit expensive. Since this need to be handled during operation in normal conditions it would be great if we could use an augmented cursor or at least use a return value to communicate the 'not found' status (we could add a new API method for this if necessary).

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -1011,7 +1049,11 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
     private var _next: NodeValue = fetchNext()
 
     private def fetchNext() = {
-      if (nodeCursor.next()) fromNodeProxy(entityAccessor.newNodeProxy(nodeCursor.nodeReference()))
+      if (nodeCursor.next()) {
+        val nodeProxy = entityAccessor.newNodeProxy(nodeCursor.nodeReference())
+        val secureNodeProxy = entityAccessor.newControlledNodeProxy(nodeProxy, transactionalContext.securityContext.mode())
```
henriknyman: Since this is executed a lot (on all node scan) I think using inheritance instead of composition for ControlledNodeProxy may be warranted for performance reasons.

henriknyman (commented):
Some comments from a performance perspective

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -196,9 +197,30 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
...
-    indexSearchMonitor.indexSeek(index, values)
-    val predicates = index.properties.zip(values).map(p => IndexQuery.exact(p._1, p._2))
-    seek(index, predicates:_*)
+    val filteredIndex = filterIndexReference(index)
+    if (filteredIndex.properties().isEmpty) {
+      return Iterator.empty
+    }
+    indexSearchMonitor.indexSeek(filteredIndex, values)
+    val predicates = filteredIndex.properties.zip(values).map(p => IndexQuery.exact(p._1, p._2))
+    seek(filteredIndex, predicates:_*)
```
fickludd: Did you think about putting the access control in the kernel instead of here? To me that would seem like the most smooth place, and it would solve both Cypher and Core API access in one place.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
fickludd: +1 on not using exceptions for this. Modifying the PropertyCursor would solve it :).

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -298,7 +301,27 @@ public Value nodeGetProperty( long nodeId, int propertyKeyId ) throws EntityNotF
...
+    private boolean propertyIsAllowed( int propertyKeyId )
+    {
+        AccessMode mode = tx.securityContext().mode();
+        try
+        {
+            return mode.allowsPropertyReads( tokenRead().propertyKeyGetName( statement, propertyKeyId ) );
+        }
+        catch ( PropertyKeyIdNotFoundKernelException e )
+        {
+            return false;
```
fickludd: Again, let's avoid exceptions here. We could add a `tokenRead().propertyKeyAllowed( ... )` or `mode.allowsPropertyKey( ... )`.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OperationsFacade.java
```diff
@@ -298,7 +301,27 @@ public Value nodeGetProperty( long nodeId, int propertyKeyId ) throws EntityNotF
         }
         try ( Cursor<NodeItem> node = dataRead().nodeCursorById( statement, nodeId ) )
         {
-            return dataRead().nodeGetProperty( statement, node.get(), propertyKeyId );
+            if ( propertyIsAllowed( propertyKeyId ) )
```
fickludd: This check could be made before getting the node cursor. No need to spend any more time in `OperationsFacade` though... it's gone before 3.4 comes out.

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/EnterpriseSecurityModule.java
```diff
@@ -334,7 +337,7 @@ private static IllegalArgumentException illegalConfiguration( String message )
         return new IllegalArgumentException( "Illegal configuration: " + message );
     }
 
-    class SecurityConfig
+    static class SecurityConfig
```
fickludd, line 340: Ooh, clean-ups!

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
```diff
@@ -297,4 +304,24 @@ public void clearAuthCache()
...
+                    blackListed.addAll( roleToPropertyBlacklist.get( role ) );
+                }
+            }
+            return property -> !blackListed.contains( property );
+        }
+        else
+        {
+            return property -> true;
+        }
+    }
```
fickludd, line 343: I like the concept of pre-computing the permission check function! As it happens at runtime though, I would suggest mapping from `propertyKeyId -> boolean` instead. Also assuming that the number of unique property keys is quite small, this could be a `boolean[]` where the check is just to read the offset. Or a bitmap is we were ambitious, but let's wait with that. These things are not necessary for a first draft of this functionality, but maybe having a card for optimizing this would be nice?

fickludd (changes requested):
I like many things here, but I suggest we have a go at moving the enforcement to the new Kernel API cursors. If you catch me a day next week, we can sketch out how that could work together.

Also let's avoid using exceptions as internal signalling constructs because it's so slow.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/plugin/PropertyLevelSecurityIT.java
```diff
@@ -0,0 +1,514 @@
...
+    public void shouldNotShowRestrictedTokensForRestrictedUser() throws Throwable
+    {
+        Result result = execute( neo, "CREATE (n {name: 'Andersson', alias: 'neo'}) ", Collections.emptyMap() );
+        assertThat( result.getQueryStatistics().getNodesCreated(), equalTo( 1 ) );
+        assertThat( result.getQueryStatistics().getPropertiesSet(), equalTo( 2 ) );
+        result.close();
+        execute( smith, "MATCH (n) WHERE n.name = 'Andersson' RETURN n, n.alias as alias", Collections.emptyMap(), r ->
+        {
+            assertThat( r.hasNext(), equalTo( true ) );
+            assertThat( r.next().get( "alias" ), equalTo( null ) );
```
fickludd, line 100: Here we should also check that `n` doesn't contain `alias`!

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/plugin/PropertyLevelSecurityIT.java
```diff
@@ -0,0 +1,514 @@
...
+    public void shouldShowRestrictedTokensForUnrestrictedUser() throws Throwable
+    {
+        Result result = execute( neo, "CREATE (n {name: 'Andersson', alias: 'neo'}) ", Collections.emptyMap() );
+        assertThat( result.getQueryStatistics().getNodesCreated(), equalTo( 1 ) );
+        assertThat( result.getQueryStatistics().getPropertiesSet(), equalTo( 2 ) );
+        result.close();
+        execute( morpheus, "MATCH (n) WHERE n.name = 'Andersson' RETURN n, n.alias as alias", Collections.emptyMap(), r ->
+        {
+            assertThat( r.hasNext(), equalTo( true ) );
+            assertThat( r.next().get( "alias" ), equalTo( "neo" ) );
```
fickludd, line 114: ...and here that that `n` does contain `alias`.

enterprise/security/src/test/java/org/neo4j/server/security/enterprise/auth/plugin/PropertyLevelSecurityIT.java
```diff
@@ -0,0 +1,514 @@
...
+            assertThat( r.hasNext(), equalTo( false ) );
+        } );
+
+        execute( smith, query, Collections.emptyMap(), r -> assertThat( r.hasNext(), equalTo( false ) ) );
+    }
+
+    @Test
+    public void shouldBehaveLikeDataIsMissingForRange() throws Throwable
+    {
+        execute( neo, "CREATE (n:Person {name: 'Andersson'}) ", Collections.emptyMap() ).close();
```
fickludd, line 282: For all of this setup it would be nice to make a helper method, e.g. `createNode( "Person", "name", "Andersson" )`, to make it less boilerplaty.

fickludd (commented):
I really like the rich test coverage in `PropertyLevelSecurityIT`!

enterprise/security/src/main/java/org/neo4j/server/security/enterprise/auth/MultiRealmAuthManager.java
```diff
@@ -305,16 +308,29 @@ public void clearAuthCache()
         return infoList;
     }
 
-    Function<String,Boolean> getPropertyPermissions( Set<String> roles )
+    Function<Integer,Boolean> getPropertyPermissions( Set<String> roles, Token token )
```
fickludd: Here we could use a `IntPredicate` instead, to remove some boxing.

fickludd (approved):
Looking really good now. Minor suggestion of function specialization, which could potentially happen during that rebase =).

PR #10894, opened 2018-01-25
Add ability to switch between on- and off-heap tx state allocation

andreikoval:


community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -820,6 +820,18 @@
...
             "dbms.transaction.bookmark_ready_timeout", DURATION, "30s" ).constraint( min( Duration.ofSeconds( 1 ) ) ).build();
 
+    public enum TransactionStateMemoryAllocation
+    {
+        ON_HEAP,
+        OFF_HEAP
+    }
+
+    @Description( "Defines where to allocate memory for storing transaction state. Valid values are `on_heap` and `off_heap`." )
+    public static final Setting<TransactionStateMemoryAllocation> tx_state_memory_allocation = buildSetting(
```
MishaDemianenko: lets make this property internal and lets change its namespace to unsupported for now since general public should be use it yet. Also description should be updated to highlight that its experimental feature

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -820,6 +820,18 @@
     public static final Setting<Duration> bookmark_ready_timeout = buildSetting(
             "dbms.transaction.bookmark_ready_timeout", DURATION, "30s" ).constraint( min( Duration.ofSeconds( 1 ) ) ).build();
 
+    public enum TransactionStateMemoryAllocation
+    {
+        ON_HEAP,
+        OFF_HEAP
+    }
+
+    @Description( "Defines where to allocate memory for storing transaction state. Valid values are `on_heap` and `off_heap`." )
```
MishaDemianenko: lets not hardcode constant names in a description, this gives to future maintainer huge possibility to not update this description in case if options will be updated

community/kernel/src/main/java/org/neo4j/kernel/impl/api/CollectionsFactory.java
```diff
@@ -0,0 +1,86 @@
...
+import org.neo4j.collection.primitive.Primitive;
+import org.neo4j.collection.primitive.PrimitiveIntObjectMap;
+import org.neo4j.collection.primitive.PrimitiveLongObjectMap;
+import org.neo4j.collection.primitive.PrimitiveLongSet;
+import org.neo4j.kernel.impl.util.diffsets.PrimitiveLongDiffSets;
+import org.neo4j.memory.MemoryTracker;
+
+import static org.neo4j.collection.primitive.PrimitiveLongCollections.emptySet;
+
+public interface CollectionsFactory
```
MishaDemianenko: class should be moved to the different package;
javadoc missing

community/kernel/src/main/java/org/neo4j/kernel/impl/api/CollectionsFactory.java
```diff
@@ -0,0 +1,86 @@
...
+    PrimitiveLongDiffSets newLongDiffSets();
+
+    MemoryTracker getMemoryTracker();
+
+    /**
+     * Release underlying native memory if applicable.
+     */
+    void release();
+
+    CollectionsFactory ON_HEAP = new CollectionsFactory()
```
MishaDemianenko: lets extract this into its own class

community/kernel/src/main/java/org/neo4j/kernel/impl/api/CollectionsFactorySupplier.java
```diff
@@ -0,0 +1,31 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api;
+
+import org.neo4j.memory.LocalMemoryTracker;
+
+public interface CollectionsFactorySupplier
```
MishaDemianenko: same thing, different package, javadoc

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -1001,11 +1005,16 @@ protected void init( long threadId, PageCursorTracer pageCursorTracer )
          * Returns number of allocated bytes by current transaction.
          * @return number of allocated bytes by the thread.
          */
-        long heapAllocateBytes()
+        long heapAllocatedBytes()
         {
             return heapAllocation.allocatedBytes( transactionThreadId ) - heapAllocatedBytesWhenQueryStarted;
         }
 
+        long directAllocatedBytes()
```
MishaDemianenko: javadoc is missing

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OffHeapCollectionsFactoryImpl.java
```diff
@@ -0,0 +1,89 @@
...
+import org.neo4j.collection.primitive.PrimitiveLongObjectMap;
+import org.neo4j.collection.primitive.PrimitiveLongSet;
+import org.neo4j.kernel.impl.util.diffsets.PrimitiveLongDiffSets;
+import org.neo4j.memory.MemoryAllocationTracker;
+import org.neo4j.memory.MemoryTracker;
+
+import static java.util.Objects.requireNonNull;
+import static org.neo4j.collection.primitive.PrimitiveLongCollections.emptySet;
+
+class OffHeapCollectionsFactoryImpl implements CollectionsFactory
```
MishaDemianenko: package, javadoc

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OffHeapCollectionsFactoryImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public void release()
+    {
+        allocated.forEach( PrimitiveCollection::close );
+        allocated.clear();
+    }
+
+    private <T extends PrimitiveCollection> T track( T collection )
+    {
+        allocated.add( collection );
```
MishaDemianenko: not quite sure we need to track all the allocated collections here; It would be useful for tests to see what collections were allocated but do not think its useful for production

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OffHeapCollectionsFactoryImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public PrimitiveLongSet newLongSet()
+    {
+        return track( Primitive.offHeapLongSet( memoryTracker ) );
+    }
+
+    @Override
+    public <V> PrimitiveLongObjectMap<V> newLongObjectMap()
+    {
+        throw new UnsupportedOperationException();
```
MishaDemianenko: we should just return on heap collections for cases where we do not have anything instead yet instead.

MishaDemianenko (changes requested):
Test for different kind of collection factories are not there, we need to make sure factories can create whatever collection they requested to create.
All tests that not updated to run with on heap state only should be updated to be parametrized tests and run with both states.
PR description should be updated to be more verbose on whats going on here :)

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
andreikoval: ✅

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
andreikoval: ✅

community/kernel/src/main/java/org/neo4j/kernel/impl/api/CollectionsFactory.java
andreikoval: ✅

community/kernel/src/main/java/org/neo4j/kernel/impl/api/CollectionsFactory.java
andreikoval: ✅

community/kernel/src/main/java/org/neo4j/kernel/impl/api/CollectionsFactorySupplier.java
andreikoval: I thinks this piece of code is trivial and doesn't need javadoc.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
andreikoval: ✅

community/kernel/src/main/java/org/neo4j/kernel/impl/api/OffHeapCollectionsFactoryImpl.java
andreikoval: ✅

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/TxState.java
```diff
@@ -1334,6 +1345,45 @@ public boolean hasDataChanges()
         return hasDataChanges;
     }
 
+    /**
+     * Release all underlying resources. The instance must not be used after calling this method .
+     */
+    public void release()
```
MishaDemianenko, line 1351: invoking combinations of close and releases here looks strange

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -851,6 +856,10 @@ private void release()
             securityContext = null;
             transactionEvent = null;
             explicitIndexTransactionState = null;
+            if ( txState != null )
+            {
+                txState.release();
+            }
             txState = null;
```
MishaDemianenko: looks like this should be inside of if statement

MishaDemianenko (commented):
really nice, only two small minor notes

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/TxState.java
andreikoval: Because `close()` is a really bad name for a method releasing resources, but I didn't want to rename it everywhere. Should we do it now?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
andreikoval: Indeed.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/TxState.java
MishaDemianenko: let's have a quick talk about this release call

MishaDemianenko (approved):

PR #10899, opened 2018-01-26
Semi-automatic code cleanup, part 1

klaren:
Semi-automatic code cleanup
* Unused imports
* Redundant semicolons
* Redundant boxing/unboxing
* If statement simplification
* Foreach instead of for
* Javadoc @link errors
* Redundant escapes in regex
* Optional with null cleanup
* Never-failing assertions
* !.anyMatch to .noneMatch
* catch immediate rethrow
* .indexOf to .contains


community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -167,7 +167,10 @@
 
         /**
          * Begin a new kernel transaction with specified timeout in milliseconds.
+<<<<<<< HEAD
```
MishaDemianenko: 🦄

MishaDemianenko (approved):
One small unicorn

MishaDemianenko (approved):

PR #10900, opened 2018-01-26
Migrate Core API relationship traversal to kernel api

fickludd:
Move `NodeProxy.getRelationships` and `.hasRelationships` implementations to use Kernel API.

Also add two new concrete helper cursors in the Kernel API, which use RelationshipTraversalCursor and RelationshipGroupCursor to iterate over the relationships of a node, and filter for type and direction. Also add common interface so node-density is possible to abstract away if wanted to.


community/kernel/src/main/java/org/neo4j/kernel/impl/core/RelationshipConversion.java
```diff
@@ -23,38 +23,38 @@
...
+                            cursor.relationshipReference(),
+                            cursor.sourceNodeReference(),
+                            cursor.label(),
+                            cursor.targetNodeReference() );
+                    return true;
+            }
             close();
         }
-        return hasNext;
+        return next != null;
```
henriknyman: Here we return true if it is closed and next is not null. Is that intentional?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/RelationshipDenseSelectionCursor.java
```diff
@@ -0,0 +1,333 @@
...
+    private RelationshipTraversalCursor relationshipCursor;
+    private int[] types;
+    private Dir[] directions;
+    private int currentDirection;
+    private int nDirections;
+    private boolean onRelationship;
+    private boolean onGroup;
+    private int foundTypes;
+
+    private enum Dir
```
thobe: How is this different from `org.neo4j.kernel.impl.newapi.RelationshipDirection`?

thobe (commented):
Since the `RelationshipSelectionCursor`s are only ever used for implementing an iterator through the use of `RelationshipConversion`, how about we avoid the complication of having to support a "richer" cursor interface, and save ourselves a layer of indirection by implementing the iterator interface directly?

That would either mean moving the `RelationshipSelectionCursor` interface and implementations (renamed to `RelationshipSelectionIterator` or something like that) to the module that implements the Core API, or make it generic and oblivious to the type it actually produces, by adding an interface like this `RelationshipFactory`:

```
public abstract class RelationshipSelectionIterator<R> implements ResourceIterator<R>
{
    public interface RelationshipFactory<R>
    {
        R relationship( long id, long startNodeId, int typeId, long endNodeId );
    }
}
```

`RelationshipFactory` can then be implemented by `spi::newRelationshipProxy`.

Although moving it to the module that implements the Core API seems simpler, in which case we would just pass the `EmbeddedProxySPI` to it.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/RelationshipDenseSelectionCursor.java
fickludd: It's 
 a) using `OUT`, `IN`, `BOTH` semantics instead of `OUT`, `IN`, `LOOP` (meaning that `Dir.OUT` matches both relationships stored as outgoing and as loops, while `RelationshipDirection.OUTGOING` only matches relationships stores as outgoing)
 b) it lives in the Kernel API instead of in the Kernel implemenation.

thobe: So `OUT` here is _outgoing (including loops)_, `IN` is _incoming (including loops)_ and `LOOP` is _loops only_? That is neither the semantics of `org.neo4j.graphdb.Direction`, nor `org.neo4j.kernel.impl.newapi.RelationshipDirection`, and should thus be thoroughly documented.

fickludd: Wait I wasn't looking properly. It has the same semantics as `RelationshipDirection` of course, nothing else. I guess we could move `RelationshipDirection` to the API module if we want to use it here, but I don't really see the point. This enum is just an implementation detail of the dense node selection, the semantics of the class are describes in the initialization javadocs.

fickludd:
@thobe As we wanted to use the selection code from Cypher as well, we opted for keeping the logic in the API module, and add a factory for creating relationship object. However, when looking around traversals in the TransactionBoundQueryContext, the iterator+factory model was hard to work with.

Then we came up with the current design, where the traversal logic can be used either as an iterator or a cursor. We intend to use the cursor variant to serve Cypher traversals which currently get the RelationshipIterator in an upcoming PR.

thobe:
I'm getting a bit of a this-is-more-complex-than-it-should-need-to-be feeling. I'll play around with the code for a bit and see if that feeling goes away.

If we are indeed doing what you describe, where the logic is being moved to a shared place, why is there so much code in `NodeProxy. getRelationshipSelectionIterator()`?

In the meantime, I'm thinking that the test failure might very well be caused by the changes in this PR, since I think it is about what happens to a relationship in a transaction where it is deleted.

pontusmelke:
@thobe the test failure is unrelated, just very flaky

thobe:
@pontusmelke ugh, that sucks. let's fix it!

thobe:
Poking through the code deeper, the previous version was simpler - where there was only a cursor, and the iterator built on top of the cursor. If we need the cursor, I think it might just be clutter to split the hierarchy in this way in order to get both cursors and iterators - simpler then to get an iterator by delegating to a cursor.

I would really like to see the use case for the cursor in order to warrant its existence though, since it is currently only used by tests... I'm not a fan of unused code.

thobe:
In particular what seems bad in this code is how big `NodeProxy.getRelationshipSelectionIterator(...)` is, that should be possible to abstract away inside some cursor/iterator factory, such that NodeProxy does not have to be bothered with whether the node is dense or not. The code in that method is also fairly repetitive, a large portion of it is the same for both the dense and the non-dense case, which makes me think that the separation of concerns is wrong.

This becomes even more apparent when looking at #10915 where this code is duplicated yet again.

pontusmelke:
Replaced with https://github.com/neo4j/neo4j/pull/10915

PR #10901, opened 2018-01-26
Optimize Cypher ResultIterator

henriknyman:
Eliminate conversion to immutable map that rebuilds a new map


SaschaPeukert (approved):
Very nice find!

thobe (approved):

PR #10903, opened 2018-01-27
 Data snapshot query execution

MishaDemianenko:
Introduce possibility to execute query based on data snapshot.
Any data modification will mark corresponding pages with a version that
will be equal to transaction id in which that particular change was introduced.
Also allow count store to track transactions in which changes where introduced.

To be able to guarantee that query result is based on a data that was
present on query execution start and was not modified while it was running,
the engine will verify that all paged that were accessed during query
execution have the version that is less or equal that was last closed
on a moment when query started.

Support for snapshot query execution in CC and HA

Snapshot query execution is enabled by 'unsupported.dbms.query.snapshot'
setting and number of retries that engine will do while trying to get
stable snapshot is controlled by 'unsupported.dbms.query.snapshot.retries'.


community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/context/EmptyVersionContextSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.function.LongSupplier;
+
+/**
+ * {@link VersionContextSupplier} version that supply same {@link EmptyVersionContext} each time for cases
+ * where version context mechanics is not required
+ */
+public class EmptyVersionContextSupplier implements VersionContextSupplier
+{
+    public static final VersionContextSupplier INSTANCE = new EmptyVersionContextSupplier();
```
chrisvest: If this was an `EMPTY` or `NULL` constant on the `VersionContextSupplier` it'd be easier to find and this class wouldn't have to be public. Small thing.

chrisvest: Similar with the `EmptyVersionContext`.

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/context/VersionContext.java
```diff
@@ -0,0 +1,72 @@
...
+package org.neo4j.io.pagecache.tracing.cursor.context;
+
+/**
+ * Context that contains state of ongoing versioned data read or write.
+ *
+ * Read context performing data read with a version that is equal to last closed transaction.
+ * Write context data modification with have version that is equal to transaction that is currently committing.
+ *
+ * As soon reader that associated with a context will observe data version that it should not see, context will be
+ * marked as dirty.
```
chrisvest: This comment is not easy to follow. Could you tighten up the wording, and add something about the life cycle of these instances, and the temporal coupling of their methods?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/counts/CountsTracker.java
```diff
@@ -293,6 +264,43 @@ protected void writeFormatSpecifier( WritableBuffer formatSpecifier )
         formatSpecifier.put( 0, FORMAT );
     }
 
+    private static class CountsTrackerRotationMonitor implements RotationMonitor
```
chrisvest, line 268: 👍

enterprise/causal-clustering/src/test/java/org/neo4j/io/pagecache/impl/muninn/VersionContextTrackingIT.java
```diff
@@ -0,0 +1,214 @@
...
+            for ( int i = 0; i <= lastPageId; i++ )
+            {
+//                try ( CursorPageAccessor pageCursor = new CursorPageAccessor(
+//                        (MuninnPageCursor) pageFile.io( i, PagedFile.PF_SHARED_READ_LOCK ) ) )
+//                {
+//                    if ( pageCursor.next() )
+//                    {
+//                        maxTransactionId = Math.max( maxTransactionId, pageCursor.lastTxModifierId() );
+//                    }
+//                }
```
chrisvest: This looks important?

enterprise/causal-clustering/src/test/java/org/neo4j/io/pagecache/impl/muninn/VersionContextTrackingIT.java
```diff
@@ -0,0 +1,214 @@
...
+        {
+            cluster.coreTx( ( coreGraphDatabase, transaction ) ->
+            {
+                coreGraphDatabase.createNode();
+                transaction.success();
+            } );
+        }
+    }
+
+//    private class CursorPageAccessor extends MuninnPageCursor
```
chrisvest: Missing cleanup?

enterprise/ha/src/test/java/org/neo4j/kernel/ha/UpdatePullerTriggersPageTransactionTrackingIT.java
```diff
@@ -0,0 +1,224 @@
...
+        try ( Transaction ignored = slave.beginTx() )
+        {
+            slave.execute( "match (n) return n" );
+        }
+        catch ( QueryExecutionException executionException )
+        {
+            assertEquals( "Unable to get clean data snapshot for query 'match (n) return n' after 5 attempts.", executionException.getMessage());
+        }
+        byzantineIdSupplier.useCorrectTxId();
+        slave.execute( "match (n) return n" ).close();
```
chrisvest, line 101: Where does the update puller get involved in this test?

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -119,12 +120,27 @@ public final boolean next( long pageId ) throws IOException
...
             return true;
         }
         nextPageId = pageId;
         return next();
     }
 
+    void verifyContext()
+    {
+        VersionContext versionContext = versionContextSupplier.getVersionContext();
+        if ( versionContext.lastClosedTransactionId() == Long.MAX_VALUE )
```
chrisvest: The interface should document that this value is special, then.

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCacheTest.java
```diff
@@ -408,7 +413,9 @@ public void markContextAsDirtyWhenAnyEvictedPageHaveModificationTransactionHighe
             assertTrue( cursor.next() );
             cursor.putLong( 4 );
         }
-        pageCache.evictPages( 1, 1, EvictionRunEvent.NULL );
+
+        pageCache.pages.tryEvict( pageCache.grabFreeAndExclusivelyLockedPage( PageFaultEvent.NULL ),
+                EvictionRunEvent.NULL );
```
chrisvest: It doesn't make sense to try to evict a free page. The `tryEvict` will always return `false` because the given page is already exclusively locked. The `grabFreeAndExclusivelyLockedPage` method might do a cooperative eviction, but the evicted page will be chosen at random. Can we make this test deterministic?

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCacheTest.java
```diff
@@ -441,7 +448,9 @@ public void doNotMarkContextAsDirtyWhenAnyEvictedPageHaveModificationTransaction
...
```
chrisvest: Same as above.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
```diff
@@ -363,12 +362,22 @@ public boolean decrementUsage( long pageRef )
 
     public long getFilePageId( long pageRef )
     {
-        return UnsafeUtil.getLong( offFilePageId( pageRef ) );
+        int highByte = UnsafeUtil.getByte( offFilePageId( pageRef ) ) & UNSIGNED_BYTE_MASK;
+        long lowInt = UnsafeUtil.getInt( offFilePageId( pageRef ) + Byte.BYTES ) & UNSIGNED_INT_MASK;
+        long filePageId = (((long) highByte) << Integer.SIZE) | lowInt;
```
chrisvest, line 367: Can't you replace this with `UnsafeUtil.getLong( offFilePageId( pageRef ) ) >>> 24`? This method is called as part of `isBoundTo`, which in turn is called on every pin.

Speaking of `isBoundTo`, the file page id and the swapper id are now part of the same aligned word. So could check the binding with a single compare by masking/shifting out the usage stamp, and comparing it with a reconstructed binding word. A single compare for both fields.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
```diff
@@ -42,28 +42,31 @@
...
 class PageList
 {
     private static final boolean forceSlowMemoryClear = flag( PageList.class, "forceSlowMemoryClear", false );
 
     private static final int UNBOUND_LAST_MODIFIED_TX_ID = -1;
+    private static final int UNSIGNED_BYTE_MASK = 0xFF;
+    private static final long UNSIGNED_INT_MASK = 0xFFFFFFFFL;
+
+    // 40 bits for file page id
+    private static final long MAX_FILE_PAGE_ID = 0b11111111_11111111_11111111_11111111_11111111L;
```
chrisvest, line 60: 👍 8 pebibytes should be enough for anyone.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/SwapperSet.java
```diff
@@ -98,8 +101,7 @@ synchronized short allocate( PageSwapper swapper )
         {
             if ( !free.isEmpty() )
             {
-                //TODO safe cast
-                short id = (short) free.iterator().next();
+                short id = Numbers.safeCastIntToShort( free.iterator().next() );
```
chrisvest: You already got this in a static import.

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -116,12 +120,41 @@ public final boolean next( long pageId ) throws IOException
...
         }
         nextPageId = pageId;
         return next();
     }
 
+    void verifyContext()
+    {
+        VersionContext versionContext = versionContextSupplier.getVersionContext();
+        long lastClosedTransactionId = versionContext.lastClosedTransactionId();
+        if ( lastClosedTransactionId == Long.MAX_VALUE )
```
chrisvest, line 134: Resurfacing comment: The interface should document that this value is special, then.

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCacheTest.java
```diff
@@ -214,6 +220,247 @@ public void mustFlushDirtyPagesOnEvictingAllPages() throws Exception
...
+
+        cursorContext.initWrite( 13 );
+        try ( PageCursor cursor = pagedFile.io( 1, PF_SHARED_WRITE_LOCK ) )
+        {
+            assertTrue( cursor.next() );
+            cursor.putLong( 4 );
+        }
+
+        pageCache.pages.tryEvict( pageCache.grabFreeAndExclusivelyLockedPage( PageFaultEvent.NULL ),
+                EvictionRunEvent.NULL );
```
chrisvest: Resurfacing comment: It doesn't make sense to try to evict a free page. The `tryEvict` will always return `false` because the given page is already exclusively locked. The `grabFreeAndExclusivelyLockedPage` method might do a cooperative eviction, but the evicted page will be chosen at random. Can we make this test deterministic?

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCacheTest.java
```diff
@@ -214,6 +220,247 @@ public void mustFlushDirtyPagesOnEvictingAllPages() throws Exception
...
```
chrisvest: Resurfacing comment: Same as above.

chrisvest (changes requested):
Looks good overall, but there's still some cleanup and test fixing left to do.

enterprise/ha/src/test/java/org/neo4j/kernel/ha/UpdatePullerTriggersPageTransactionTrackingIT.java
MishaDemianenko: here we check that update pulling together with starting with tx on slave works as expected:
we check first that in case if our incorrect tx id provider will supply wrong tx id we will not gonna be able to execute query and in this line we check that as soon as id provider supply correct transaction id query execution works as expected

community/io/src/main/java/org/neo4j/io/pagecache/tracing/cursor/context/VersionContext.java
```diff
@@ -22,11 +22,24 @@
...
+ *     context will be marked as dirty.
+ *     <br/>
+ *     For example, when context is initialised with last closed transaction id and if
+ *     at the end of reading operation context is not marked as dirty its guarantee that context did not
+ *     encounter any data from more recent transaction.</li>
+ *     <li>Write context: context that performs data modifications. Any modifications will be tagged with
+ *     some version that write context was initialised with.
+ *     <br/>
+ *     For example, commit will start write context that with a version that is equal to current
+ *     committing transaction id./li>
```
chrisvest: Missed a `<` on the li-end tag, but otherwise 👍

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/PageList.java
MishaDemianenko: talked offline about it, we will look on it separately to see if it make sense to do it

MishaDemianenko:
@chrisvest now it should be it, updated

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/SwapperSet.java
```diff
@@ -107,7 +110,7 @@ synchronized int allocate( PageSwapper swapper )
         }
 
         // No free slot was found above, so we extend the array to make room for a new slot.
-        int id = swapperMappings.length;
+        short id = safeCastLongToShort( swapperMappings.length );
```
chrisvest: I guess this should say `safeCastIntToShort` instead of `Long`?

chrisvest (approved):

PR #10911, opened 2018-01-29
Instantiates PageAwareByteArrayCursor in @Before

tinwelint:
instead of in field initialization, because it would have the potential to
leave old data affecting other tests, where pointers would be observed as BROKEN


burqen (approved):
LGTM!

PR #10702, opened 2017-12-29
Spatial index functionality

OliviaYtterbrink:
Builds on #10766 and needs #10780 

changelog: Index-backed support for spatial point equality and range searches from Cypher.

For example, the following query will return all points in the specified range:

    WITH point({latitude: 56.5, longitude: 12.5}) as bottomLeft, point({latitude: 57.5, longitude: 13.5}) as topRight
    MATCH (c:City) WHERE c.location > bottomLeft AND c.location < topRight

This also includes support for storing points and arrays of points as node properties.

PR #10912, opened 2018-01-29
Improve performace of SpaceFillingCurve queries.

sherfert:
This is based on #10702. Only commits including and after b68857562c92aa4ab3a1d82a85de666cc403cd87 should be reviewed here.

This restricts the depth an index query in allowed to go into the SpaceFillingCurve, depending on the ratio of area covered by the indey and area included in the search query. It also includes different configurations to twist that behavior. These are currently used for micro-benchmarks. We should proceed to use them for fullstack-benchmarks as soon is this is merged.

### DIFF
--- a/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
+++ b/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
@@ -26,9 +26,9 @@ import org.neo4j.gis.spatial.index.Envelope;
  */
 class SearchEnvelope
 {
-    long[] min; // inclusive lower bounds
-    long[] max; // exclusive upper bounds
-    int nbrDim;
+    private long[] min; // inclusive lower bounds
+    private long[] max; // exclusive upper bounds
+    private int nbrDim;
 
     SearchEnvelope( SpaceFillingCurve curve, Envelope referenceEnvelope )
     {
@@ -37,6 +37,7 @@ class SearchEnvelope
         this.nbrDim = referenceEnvelope.getDimension();
         for ( int i = 0; i < nbrDim; i++ )
         {
+            // getNormalizedCoord gives inclusive bounds. Need to increment to make the upper exclusive.
             this.max[i] += 1;
         }
     }
@@ -99,12 +100,16 @@ class SearchEnvelope
         return true;
     }
 
-    /** Only valid for intersecting envelopes
+    /**
+     * Calculates the faction of the overlapping area between {@code this} and {@code} other compared
+     * to the area of {@code this}.
+     *
+     * Must only be called for intersecting envelopes
      */
     double fractionOf( SearchEnvelope other )
     {
         double fraction = 1.0;
-        for ( int i = 0; i < min.length; i++ )
+        for ( int i = 0; i < nbrDim; i++ )
         {
             long min = Math.max( this.min[i], other.min[i] );
             long max = Math.min( this.max[i], other.max[i] );
@@ -120,7 +125,7 @@ class SearchEnvelope
     public long getArea()
     {
         long area = 1;
-        for ( int i = 0; i < min.length; i++ )
+        for ( int i = 0; i < nbrDim; i++ )
         {
             area *= max[i] - min[i];
         }
--- a/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
+++ b/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.gis.spatial.index.curves;
+
+import org.neo4j.gis.spatial.index.Envelope;
+
+/**
+ * N-dimensional searchEnvelope
+ */
+class SearchEnvelope
+{
+    long[] min; // inclusive lower bounds
+    long[] max; // exclusive upper bounds
+    int nbrDim;
+
+    SearchEnvelope( SpaceFillingCurve curve, Envelope referenceEnvelope )
+    {
+        this.min = curve.getNormalizedCoord( referenceEnvelope.getMin() );
+        this.max = curve.getNormalizedCoord( referenceEnvelope.getMax() );
+        this.nbrDim = referenceEnvelope.getDimension();
+        for ( int i = 0; i < nbrDim; i++ )
+        {
+            this.max[i] += 1;
+        }
+    }
+
+    private SearchEnvelope( long[] min, long[] max )
+    {
+        this.min = min;
+        this.max = max;
+        this.nbrDim = min.length;
+    }
+
+    SearchEnvelope( long min, long max, int nbrDim )
+    {
+        this.nbrDim = nbrDim;
+        this.min = new long[nbrDim];
+        this.max = new long[nbrDim];
+
+        for ( int dim = 0; dim < nbrDim; dim++ )
+        {
+            this.min[dim] = min;
+            this.max[dim] = max;
+        }
+    }
+
+    SearchEnvelope quadrant( int[] quadNbrs )
+    {
+        long[] newMin = new long[nbrDim];
+        long[] newMax = new long[nbrDim];
+
+        for ( int dim = 0; dim < nbrDim; dim++ )
+        {
+            long extent = (max[dim] - min[dim]) / 2;
+            newMin[dim] = this.min[dim] + quadNbrs[dim] * extent;
+            newMax[dim] = this.min[dim] + (quadNbrs[dim] + 1) * extent;
+        }
+        return new SearchEnvelope( newMin, newMax );
+    }
+
+    boolean contains( long[] coord )
+    {
+        for ( int dim = 0; dim < nbrDim; dim++ )
+        {
+            if ( coord[dim] < min[dim] || coord[dim] >= max[dim] )
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    boolean intersects( SearchEnvelope other )
+    {
+        for ( int dim = 0; dim < nbrDim; dim++ )
+        {
+            if ( this.max[dim] <= other.min[dim] || other.max[dim] <= this.min[dim] )
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    double fractionOf( SearchEnvelope other )
+    {
+        double fraction = 1.0;
+        for ( int i = 0; i < min.length; i++ )
+        {
+            long min = Math.max( this.min[i], other.min[i] );
+            long max = Math.min( this.max[i], other.max[i] );
+            final double innerFraction = (double) (max - min) / (double) (other.max[i] - other.min[i]);
+            fraction *= innerFraction;
+        }
+        return fraction;
+    }
+}
--- a/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
+++ b/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
@@ -99,6 +99,8 @@ class SearchEnvelope
         return true;
     }
 
+    /** Only valid for intersecting envelopes
+     */
     double fractionOf( SearchEnvelope other )
     {
         double fraction = 1.0;
--- a/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
+++ b/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
@@ -113,4 +113,17 @@ class SearchEnvelope
         }
         return fraction;
     }
+
+    /**
+     * The smallest possible envelope has unit area 1
+     */
+    public long getArea()
+    {
+        long area = 1;
+        for ( int i = 0; i < min.length; i++ )
+        {
+            area *= max[i] - min[i];
+        }
+        return area;
+    }
 }
--- a/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
+++ b/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
@@ -269,6 +269,11 @@ public abstract class SpaceFillingCurve
         SearchEnvelope wholeExtent = new SearchEnvelope( 0, this.getWidth(), nbrDim );
         ArrayList<LongRange> results = new ArrayList<>(1000);
 
+        if ( monitor != null )
+        {
+            monitor.registerSearchArea( search.getArea() );
+        }
+
         addTilesIntersectingEnvelopeAt( config, monitor, 0, config.maxDepth( referenceEnvelope, this.range, nbrDim, maxLevel ), search,
                 wholeExtent, rootCurve(), 0, this.getValueWidth(), results );
         return results;
@@ -295,10 +300,11 @@ public abstract class SpaceFillingCurve
                     current = new LongRange( left );
                     results.add( current );
                 }
-            }
-            if(monitor != null)
-            {
-                monitor.addRangeAtDepth( depth );
+                if ( monitor != null )
+                {
+                    monitor.addRangeAtDepth( depth );
+                    monitor.addToCoveredArea( currentExtent.getArea() );
+                }
             }
         }
         else if ( search.intersects( currentExtent ) )
@@ -323,6 +329,7 @@ public abstract class SpaceFillingCurve
                 if(monitor != null)
                 {
                     monitor.addRangeAtDepth( depth );
+                    monitor.addToCoveredArea( currentExtent.getArea() );
                 }
             }
             else
@@ -367,7 +374,7 @@ public abstract class SpaceFillingCurve
             double value = clamp( coord[dim], range.getMin( dim ), range.getMax( dim ) );
             if ( value == range.getMax( dim ) )
             {
-                normalizedCoord[dim] = valueWidth - 1;
+                normalizedCoord[dim] = width - 1;
             }
             else
             {
--- a/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
+++ b/community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
@@ -265,7 +265,7 @@ public abstract class SpaceFillingCurve
      */
     List<LongRange> getTilesIntersectingEnvelope( Envelope referenceEnvelope, RecursionStats stats )
     {
-        SearchEnvelope search = new SearchEnvelope( referenceEnvelope );
+        SearchEnvelope search = new SearchEnvelope( this, referenceEnvelope );
         ArrayList<LongRange> results = new ArrayList<>(1000);
 
         addTilesIntersectingEnvelopeAt( stats, 0, search, new SearchEnvelope( 0, this.getWidth(), nbrDim ), rootCurve(), 0, this.getValueWidth(), results );
@@ -311,7 +311,7 @@ public abstract class SpaceFillingCurve
         else if ( search.intersects( currentExtent ) )
         {
             double overlap = search.fractionOf( currentExtent );
-            if ( overlap >= 0.50 )
+            if ( overlap >= 0.99 )
             {
                 // Note that LongRange upper bound is inclusive, hence the '-1' in several places
                 LongRange current = (results.size() > 0) ? results.get( results.size() - 1 ) : null;
@@ -358,7 +358,7 @@ public abstract class SpaceFillingCurve
     /**
      * Given a coordinate, find the corresponding normalized coordinate
      */
-    private long[] getNormalizedCoord( double[] coord )
+    long[] getNormalizedCoord( double[] coord )
     {
         long[] normalizedCoord = new long[nbrDim];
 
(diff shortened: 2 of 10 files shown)


community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
```diff
@@ -586,7 +586,7 @@ public SearchEnvelope intersection( SearchEnvelope other )
                 long[] i_min = new long[this.min.length];
                 long[] i_max = new long[this.min.length];
                 boolean result = true;
-                for ( int i = 0; i < min.length; i++ )
+                for ( int i = 0; i < min.length && result; i++ )
```
Lojjs: I think it is cleaner to put a `break` statement in the else case than checking `&& result` here

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurveConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Decides whether to stop at this depth or recurse deeper.
+     *
+     * @param overlap the overlap between search space and the current extent
+     * @param depth the current recursion depth
+     * @param maxDepth the maximum depth that was calculated to recurse to,
+     * @return if the algorithm should recurse deeper, returns {@code false}; if the algorithm
+     * should stop at this depth, returns {@code true}
+     */
+    boolean stopAtThisDepth( double overlap, int depth, int maxDepth );
```
Lojjs, line 35: A well-documented interface, I like 😃

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/HistogramMonitor.java
```diff
@@ -0,0 +1,75 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.gis.spatial.index.curves;
+
+public class HistogramMonitor implements SpaceFillingCurveMonitor
```
Lojjs, line 22: It would be helpful to use `inheritdoc` in this class as well

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/TraverseToBottomConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.gis.spatial.index.curves;
+
+import org.neo4j.gis.spatial.index.Envelope;
+
+public class TraverseToBottomConfiguration implements SpaceFillingCurveConfiguration
```
Lojjs, line 24: This class could also benefit from `inheritdoc`

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
```diff
@@ -0,0 +1,129 @@
...
+    int nbrDim;
+
+    SearchEnvelope( SpaceFillingCurve curve, Envelope referenceEnvelope )
+    {
+        this.min = curve.getNormalizedCoord( referenceEnvelope.getMin() );
+        this.max = curve.getNormalizedCoord( referenceEnvelope.getMax() );
+        this.nbrDim = referenceEnvelope.getDimension();
+        for ( int i = 0; i < nbrDim; i++ )
+        {
+            this.max[i] += 1;
```
Lojjs: Why do we do `+=1` here? Is the coordinate we get from `curve.getNormalizedCoord( referenceEnvelope.getMax() );` inclusive instead of exclusive?

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
```diff
@@ -0,0 +1,129 @@
...
+    SearchEnvelope quadrant( int[] quadNbrs )
+    {
+        long[] newMin = new long[nbrDim];
+        long[] newMax = new long[nbrDim];
+
+        for ( int dim = 0; dim < nbrDim; dim++ )
+        {
+            long extent = (max[dim] - min[dim]) / 2;
+            newMin[dim] = this.min[dim] + quadNbrs[dim] * extent;
+            newMax[dim] = this.min[dim] + (quadNbrs[dim] + 1) * extent;
```
Lojjs, line 74: I think this is not working properly with odd numbers. Let's say that we have an envelope from (0,0) to (5,5) Then `extent = (5 - 0)/ 2 = 2` and the bigger `newMax = 0 + (1 + 1) * 2 = 4`  So this will create quadrants with sides `(0 - 2]` and `(2-4]` i.e. the upper point is on `(4,4)` instead of `(5,5)`

If this is an bug we have not discovered previously, we may need to increase the test coverage to include such test Envelopes also.

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+        return true;
+    }
+
+    /** Only valid for intersecting envelopes
+     */
+    double fractionOf( SearchEnvelope other )
+    {
+        double fraction = 1.0;
+        for ( int i = 0; i < min.length; i++ )
```
Lojjs: Use `nbrDim` instead of `min.length` here

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
```diff
@@ -0,0 +1,129 @@
...
+        return fraction;
+    }
+
+    /**
+     * The smallest possible envelope has unit area 1
+     */
+    public long getArea()
+    {
+        long area = 1;
+        for ( int i = 0; i < min.length; i++ )
```
Lojjs: Use `nbrDim` instead of `min.length` here

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
```diff
@@ -0,0 +1,129 @@
...
+    /** Only valid for intersecting envelopes
+     */
+    double fractionOf( SearchEnvelope other )
+    {
+        double fraction = 1.0;
+        for ( int i = 0; i < min.length; i++ )
+        {
+            long min = Math.max( this.min[i], other.min[i] );
+            long max = Math.min( this.max[i], other.max[i] );
+            final double innerFraction = (double) (max - min) / (double) (other.max[i] - other.min[i]);
```
Lojjs: Is there some particular reason that this is marked as final? It is of course final since a new one is created each iteration, but the same is true for `min` and `max` as well

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
```diff
@@ -0,0 +1,129 @@
...
+        {
+            if ( this.max[dim] <= other.min[dim] || other.max[dim] <= this.min[dim] )
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    /** Only valid for intersecting envelopes
```
Lojjs: Can you add a clearer description here alt. a clearer name of the function? I think the other methods are kind of self explanatory, but at a first glance I didn't realise that this was the **area of the overlap** as fraction of the other envelope's area.

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/StandardConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.gis.spatial.index.curves;
+
+import org.neo4j.gis.spatial.index.Envelope;
+
+public class StandardConfiguration implements SpaceFillingCurveConfiguration
+{
+    public static final int DEFAULT_EXTRA_LEVELS = 1;
+    protected int extraLevels;
```
Lojjs: I don't understand the concept of `extraLevels`, can you add an informative comment about it here?

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/StandardConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+    public boolean stopAtThisDepth( double overlap, int depth, int maxDepth )
+    {
+        return overlap >= 0.99 || depth >= maxDepth;
+    }
+
+    /**
+     * If the search area is exactly one of the finest grained tiles (tile at maxLevel), then
+     * we want the search to traverse to maxLevel, however, for each area that is 4x larger, we would
+     * traverse one level shallower. This is achieved by a log of the ration of areas,
+     * to the base of 4 (for 2D).
```
Lojjs: Misspelled **ration** and a bit hard to understand what `a log of the ratio of areas to the base of 4` means. I've never heard `to the base of` as referring to division by a log

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/StandardConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+     * <p>
+     * {@inheritDoc}
+     */
+    @Override
+    public int maxDepth( Envelope referenceEnvelope, Envelope range, int nbrDim, int maxLevel )
+    {
+        double searchRatio = range.getArea() / referenceEnvelope.getArea();
+        final int i = (int) (Math.log( searchRatio ) / Math.log( Math.pow( 2, nbrDim ) )) + extraLevels;
```
Lojjs: I'm not sure, but do we need to use the natural logarithmic here or are you just after logarithmic behaviour in general? In the latter case, it would be more efficient to use `log2`, I suppose. Then you could write this as `(int) (Math.log2(searchRatio) / nbrDim) + extraLevels`

Lojjs: Also is the `final` used for performance reasons?

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/PartialOverlapConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+        super( extraLevels );
+        this.topThreshold = topThreshold;
+        this.bottomThreshold = bottomThreshold;
+    }
+
+    /**
+     * This simply stops at the maxDepth calculated in the maxDepth() function, or
+     * if the overlap is over by some fraction which is 99% at the top levels, bu reduces
+     * linearly to 0.5 when we get to maxDepth.
+     * <p>
```
Lojjs: Misspelled comment **bu**. Also 0.5 and 99% is only true if constructed with the default values

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/PartialOverlapConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean stopAtThisDepth( double overlap, int depth, int maxDepth )
+    {
+        double slope = (bottomThreshold - topThreshold) / maxDepth;
+        double threshold = slope * depth + topThreshold;
```
Lojjs, line 52: If it's not bad performance-wise, I think it would be clearer to have `slope = (topThreshold - bottomThreshold) / maxDepth` and `threshold = topThreshold - slope * depth` instead of a negative slope

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
```diff
@@ -318,16 +362,17 @@ else if ( search.intersects( currentExtent ) )
...
+    long[] getNormalizedCoord( double[] coord )
     {
         long[] normalizedCoord = new long[nbrDim];
 
         for ( int dim = 0; dim < nbrDim; dim++ )
         {
             double value = clamp( coord[dim], range.getMin( dim ), range.getMax( dim ) );
-            if ( value == range.getMax( dim ) )
+            // Avoiding awkward rounding errors
+            if ( value - range.getMin( dim ) == range.getMax( dim ) - range.getMin( dim ) )
```
Lojjs, line 370: This feels like an un-secure way of getting rid of double rounding errors (and I don't understand how it helps). I think you should not check that doubles are equal with == at all, but rather do `Math.abs(value - range.getMax( dim)) < epsilon`

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
```diff
@@ -260,15 +260,30 @@ private Long derivedValueFor( long[] normalizedValues, int level )
...
+    }
+
+    /**
+     * Given an envelope, find a collection of LongRange of tiles intersecting it on maxLevel and merge adjacent ones
+     */
+    List<LongRange> getTilesIntersectingEnvelope( Envelope referenceEnvelope, SpaceFillingCurveConfiguration config, SpaceFillingCurveMonitor monitor )
+    {
+        SearchEnvelope search = new SearchEnvelope( this, referenceEnvelope );
+        SearchEnvelope wholeExtent = new SearchEnvelope( 0, this.getWidth(), nbrDim );
+        ArrayList<LongRange> results = new ArrayList<>(1000);
```
Lojjs: Is 1000 an arbitrary trade-of between memory and not having to extend the size of the array so many times or is it some more reasoning behind the choice of the value? Perhaps something to benchmark later.

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurveTest.java
```diff
@@ -30,19 +30,39 @@
...
+import static org.hamcrest.Matchers.lessThanOrEqualTo;
 import static org.neo4j.gis.spatial.index.curves.HilbertSpaceFillingCurve3D.BinaryCoordinateRotationUtils3D.rotateNPointLeft;
 import static org.neo4j.gis.spatial.index.curves.HilbertSpaceFillingCurve3D.BinaryCoordinateRotationUtils3D.rotateNPointRight;
 
 public class SpaceFillingCurveTest
 {
+    private static final boolean verbose = false;
+    private static final boolean debug = false;
+
+    private static void log( String line )
```
Lojjs, line 50: Good improvement instead of all System.out.println 😄

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurveTest.java
```diff
@@ -338,39 +356,209 @@ public void shouldGet2DHilbertSearchTilesForManyLevels()
...
-            double halfTile = curve.getTileWidth( 0, level ) / 2.0;
-            long start = System.currentTimeMillis();
+            final int delta = 1;
+            final int xhalf = xmin + (xmax - xmin) / 2;
+            final int yhalf = ymin + (ymax - ymin) / 2;
+            Envelope q1 = new Envelope( xmin, xhalf - delta, ymin, yhalf - delta );
+            Envelope q2 = new Envelope( xmin, xhalf - delta, yhalf, ymax );
+            Envelope q3 = new Envelope( xhalf, xmax, yhalf, ymax );
+            Envelope q4 = new Envelope( xhalf, xmax, ymin, yhalf - delta );
+
```
Lojjs, line 383: I'm thinking that this code could be useful for more tests. Maybe we should extract it to a private method?

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurveTest.java
```diff
@@ -338,39 +356,209 @@ public void shouldGet2DHilbertSearchTilesForManyLevels()
...
+            count++;
+        }
+
+        double avg()
+        {
+            return (double) total / (double) count;
+        }
+    }
+
+    private static class MonitorDoubleStats
```
Lojjs, line 420: Instead of having one class for long and one identical for doubles, they could probably be merged to one class by using type parameterisation with AnyVal or the Numeric trait in Scala. However, that is maybe a bit overkill here...

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurveTest.java
```diff
@@ -338,39 +356,209 @@ public void shouldGet2DHilbertSearchTilesForManyLevels()
...
+    @Test
+    public void shouldHaveReasonableCoveredArea()
+    {
+        final double minExtent = 0.000001;
+        final double maxAspect = 100.0;
+        final int xmin = -100;
+        final int xmax = 100;
+        final int ymin = -100;
+        final int ymax = 100;
+        final double rectangleStepsPerDimension = 9.789;
```
Lojjs: Where does this number come from? It looks a bit too arbitrary without any comment for the curious reader. If it should be "random" for good test coverage it would be good with a comment saying so.

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurveTest.java
```diff
@@ -338,39 +356,209 @@ public void shouldGet2DHilbertSearchTilesForManyLevels()
...
+                log( String.format( formatBody, level, config.toString(),
+                        areaStats.avg(), areaStats.min, areaStats.max,
+                        rangeStats.avg(), rangeStats.min, rangeStats.max,
+                        maxDepthStats.avg(), maxDepthStats.min, maxDepthStats.max ) );
+            }
+        }
+    }
+
+    @Ignore
+    public void debugSingle()
```
Lojjs: Why is this test ignored? A TODO comment about when it should be un-ignored would be good

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurveTest.java
```diff
@@ -338,39 +356,209 @@ public void shouldGet2DHilbertSearchTilesForManyLevels()
...
+                MonitorStats maxDepthStats = new MonitorStats();
+                HilbertSpaceFillingCurve2D curve = new HilbertSpaceFillingCurve2D( envelope, level );
+
+                // For differently shaped rectangles
+                for ( double xExtent = minExtent; xExtent <= xmax; xExtent *= extensionFactor )
+                {
+                    for ( double yExtent = minExtent; yExtent <= ymax; yExtent *= extensionFactor )
+                    {
+                        // Filter out very thin rectangles
+                        double aspect = xExtent > yExtent ? xExtent / yExtent : yExtent / xExtent;
```
Lojjs: This line I find quite hard to read, maybe some parentheses or a proper if-else-statement would make it more readable. Or you could change it to something like `max(xExtent,yExtent) / min(xExtent,yExtent)`

Lojjs (changes requested):
On the actual optimization work, I mostly have minor comments about readability and simplifications etc. But I think I have found a bug regarding to SearchEnvelopes and quadrants with odd numbers. This is in old code that has been moved to a new class, so either I have thought wrong or we have not good enough test coverage.

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SearchEnvelope.java
sherfert: Exactly. Adding a comment.

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
sherfert: It's kind of empirical after what we saw in debugging. I'll add it to the config so we can benchmark it.

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurve.java
sherfert: It would be very hard to choose an epsilon, since we allow arbitrary small search ranges. 

This does actually help, since we are using exactly `value - range.getMin( dim )` in the else block *and* in the if statement now. The error I had before was, e.g., that `value = 1.99999999` and `max = 2` and `min = 1`, so we would choose the else block. But actually `1.99999999 - 1` would evaluate to `1`, just as `2 - 1`, so we should have chosen the if-block. If we end up with a value too big in the else block, we get off-by-one errors.

community/spatial-index/src/main/java/org/neo4j/gis/spatial/index/curves/StandardConfiguration.java
sherfert: Unfortunately, Math does not offer a `log2`. If it did, we could simplify as you said.

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/HistogramMonitor.java
sherfert: JavaDoc is inherited by default, if you don't add JavaDoc in the extending class. You only need that annotation, if you want to use the superclass JavaDoc and extend it in the subclass.

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/PartialOverlapConfiguration.java
sherfert: This is viewing it as a function with mx+b, where `topThreshold == b` and `slope ==m`, so I think a negative slope makes sense.

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/SpaceFillingCurveTest.java
sherfert: Right now this pattern only exists there...

community/spatial-index/src/test/java/org/neo4j/gis/spatial/index/curves/HistogramMonitor.java
Lojjs: Cool I didn't know about `inheritdoc` at all

Lojjs (approved):

PR #10915, opened 2018-01-29
Kernel API for traversals

pontusmelke:
Switches to Kernel API for
- `NodeProxy:getRelationships` and `NodeProxy:hasRelationships`
- Cypher interpreted runtimes, i.e. `QueryContext:getRelationshipsForIds` and `QueryContext:getRelationshipsForIdsPrimitive`
- Cypher compiled runtime: `Expand(All)`, `Expand(Into)`, and `type(r)`.

Also adds two new concrete helper cursors in the Kernel API, which use RelationshipTraversalCursor and RelationshipGroupCursor to iterate over the relationships of a node, and filter for type and direction. Also add common interface so node-density is possible to abstract away if wanted to.

Replaces https://github.com/neo4j/neo4j/pull/10900


community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -1007,77 +1030,94 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
...
-        nodeCursor.close()
+       close()
       }
       current
     }
   }
 
-  private def nodeValueIndexCursorToIterator(nodeCursor: NodeValueIndexCursor): Iterator[NodeValue] =  new Iterator[NodeValue] {
-    private var _next: NodeValue = fetchNext()
+  class RelationShipCursorIterator(selectionCursor: RelationshipSelectionCursor) extends RelationshipIterator {
```
thobe: A ship full of relations?

thobe (commented):
Looking at the use of the cursors from #10900 here, they are again used as the basis for constructing iterators, which makes me think that the interface we are exposing through the `RelationshipSelectionCursor` is more powerful than it needs to be.

pontusmelke:
@thobe They are used as a basis for `Iterator[RelationshipValue]` as well as the primitive `RelationshipIterator` in the interpreted runtime and now in the compiled runtime they are used more directly. I think we need to do something about the duplicated logic around getting the right `RelationshipSelection[Cursor/Iterator]` given the denseness, direction etc. As it is now we have basically the same code in three places with some small subtle differences.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/RelationshipSelections.java
```diff
@@ -0,0 +1,156 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.internal.kernel.api.helpers;
+
+import org.neo4j.graphdb.Direction;
```
thobe: We wanted to avoid depending on the graphdb API here, did we not?

thobe: I'd change the design of this class to provide 6 static methods:

* `outgoingIterator(...)`
* `incomingIterator(...)`
* `allIterator(...)`
* `outgoingCursor(...)`
* `incomingCursor(...)`
* `allCursor(...)`

That way there would be no need to depend on `graphdb.Direction`, and if some caller has another encoding of direction it wouldn't need to translate.

thobe (approved):
Looks to be fitting together pretty nicely.

PR #10917, opened 2018-01-29
Semi-automatic code cleanup, part 2

klaren:
Semi-automatic code cleanup

- Explicit <>
- Collapse identical catch blocks
- Prefer try-with-resource
- Prefer lambdas and method references
- Collections.sort to List.sort
- Unnecessary casts


MishaDemianenko (approved):

MishaDemianenko:
@klaren have pushed several small updates and approved

PR #10920, opened 2018-01-29
Fixes assumption in SeekCursor that node is LEAF

tinwelint:
in between two reads from the same node that node could be deleted and reused as an internal node,
which would make reading keys with the assumption that it was a LEAF read garbage data,
potentially causing out of bounds.


burqen (approved):
Merging

PR #10922, opened 2018-01-29
Get rid of static state in TxState

andreikoval:


community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/NodeStateImpl.java
```diff
@@ -41,8 +40,141 @@
...
+        @Override
+        public boolean isPropertyRemoved( int propertyKeyId )
+        {
+            return false;
+        }
+
+        @Override
+        public PrimitiveLongIterator getAddedRelationships( Direction direction )
+        {
+            return null;
```
MishaDemianenko: next 4 methods should return an empty iterator instead, to make it consistent across methods that return iterators at least

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/TxState.java
```diff
@@ -353,18 +348,32 @@ public boolean hasChanges()
...
     {
-        return LABEL_STATE.getOrCreate( this, labelId ).getOrCreateNodeDiffSets();
+        if ( labelStatesMap == null )
+        {
+            labelStatesMap = Primitive.intObjectMap();
+        }
+        return labelStatesMap.computeIfAbsent( labelId, unused -> new DiffSets<>() );
+    }
+
+    private ReadableDiffSets<Long> getRoLabelStateNodeDiffSets( int labelId )
```
MishaDemianenko: what does that 'Ro' part stand's for here and in other methods?

MishaDemianenko (changes requested):
Couple of small questions/comments

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/TxState.java
andreikoval: It meant to be read-only, but I'll remove it.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/state/NodeStateImpl.java
andreikoval: I didn't want to change/break existing contracts, but I agree it should be fixed.

MishaDemianenko (approved):

PR #10923, opened 2018-01-29
Remove wrapping of proxies

pontusmelke:
By instead letting `NodeProxy` and `RelationshipProxy` extend `NodeValue`
and `RelationshipValue` respectively we can remove a lot of wrapping in
the interpreted runtime. We cannot get rid of the wrapping altogether
since we must still support passing in `Node` and `Relationship` via
parameters.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
@@ -46,6 +46,13 @@ import org.neo4j.kernel.api.StatementTokenNameLookup;
 import org.neo4j.kernel.api.schema.index.IndexDescriptor;
 import org.neo4j.kernel.impl.coreapi.schema.PropertyNameUtils;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
+import org.neo4j.values.storable.TextArray;
+import org.neo4j.values.storable.TextValue;
+import org.neo4j.values.storable.Values;
+import org.neo4j.values.virtual.MapValue;
+import org.neo4j.values.virtual.NodeValue;
+import org.neo4j.values.virtual.RelationshipValue;
+import org.neo4j.values.virtual.VirtualValues;
 
 import static org.neo4j.kernel.api.schema.index.IndexDescriptor.Type.GENERAL;
 
@@ -220,19 +227,17 @@ public class SchemaProcedure
         return graphResult;
     }
 
-    private static class VirtualRelationshipHack implements Relationship
+    private static class VirtualRelationshipHack extends RelationshipValue implements Relationship
     {
-
         private static AtomicLong MIN_ID = new AtomicLong( -1 );
 
-        private final long id;
         private final Node startNode;
         private final Node endNode;
         private final RelationshipType relationshipType;
 
         VirtualRelationshipHack( final VirtualNodeHack startNode, final VirtualNodeHack endNode, final String type )
         {
-            this.id = MIN_ID.getAndDecrement();
+            super( MIN_ID.getAndDecrement() );
             this.startNode = startNode;
             this.endNode = endNode;
             relationshipType = () -> type;
@@ -293,6 +298,24 @@ public class SchemaProcedure
         }
 
         @Override
+        public long getStartNodeId()
+        {
+            return 0;
+        }
+
+        @Override
+        public long getEndNodeId()
+        {
+            return 0;
+        }
+
+        @Override
+        public long getOtherNodeId( long id )
+        {
+            return 0;
+        }
+
+        @Override
         public GraphDatabaseService getGraphDatabase()
         {
             return null;
@@ -345,20 +368,43 @@ public class SchemaProcedure
         {
             return String.format( "VirtualRelationshipHack[%s]", id );
         }
+
+        @Override
+        public NodeValue startNode()
+        {
+            return null;
+        }
+
+        @Override
+        public NodeValue endNode()
+        {
+            return null;
+        }
+
+        @Override
+        public TextValue type()
+        {
+            return null;
+        }
+
+        @Override
+        public MapValue properties()
+        {
+            return null;
+        }
     }
 
-    private static class VirtualNodeHack implements Node
+    private static class VirtualNodeHack extends NodeValue implements Node
     {
 
         private final HashMap<String,Object> propertyMap = new HashMap<>();
 
         private static AtomicLong MIN_ID = new AtomicLong( -1 );
-        private final long id;
         private final Label label;
 
         VirtualNodeHack( final String label, Map<String,Object> properties )
         {
-            this.id = MIN_ID.getAndDecrement();
+            super( MIN_ID.getAndDecrement() );
             this.label = Label.label( label );
             propertyMap.putAll( properties );
             propertyMap.put( "name", label );
@@ -554,6 +600,18 @@ public class SchemaProcedure
         public Map<String,Object> getProperties( String... keys )
         {
             return null;
+        }
+
+        @Override
+        public TextArray labels()
+        {
+            return Values.EMPTY_TEXT_ARRAY;
+        }
+
+        @Override
+        public MapValue properties()
+        {
+            return VirtualValues.EMPTY_MAP;
         }
 
         @Override
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
@@ -49,6 +49,8 @@ import org.neo4j.internal.kernel.api.exceptions.LabelNotFoundKernelException;
 import org.neo4j.internal.kernel.api.exceptions.PropertyKeyIdNotFoundKernelException;
 import org.neo4j.internal.kernel.api.exceptions.explicitindex.AutoIndexingKernelException;
 import org.neo4j.internal.kernel.api.exceptions.schema.ConstraintValidationException;
+import org.neo4j.internal.kernel.api.exceptions.schema.IllegalTokenNameException;
+import org.neo4j.internal.kernel.api.exceptions.schema.TooManyLabelsException;
 import org.neo4j.kernel.api.KernelTransaction;
 import org.neo4j.kernel.api.ReadOperations;
 import org.neo4j.kernel.api.Statement;
@@ -56,12 +58,14 @@ import org.neo4j.kernel.api.StatementTokenNameLookup;
 import org.neo4j.kernel.api.exceptions.EntityNotFoundException;
 import org.neo4j.kernel.api.exceptions.RelationshipTypeIdNotFoundKernelException;
 import org.neo4j.kernel.api.exceptions.Status;
-import org.neo4j.internal.kernel.api.exceptions.schema.IllegalTokenNameException;
-import org.neo4j.internal.kernel.api.exceptions.schema.TooManyLabelsException;
 import org.neo4j.kernel.impl.api.operations.KeyReadOperations;
+import org.neo4j.kernel.impl.util.ValueUtils;
 import org.neo4j.storageengine.api.EntityType;
+import org.neo4j.values.storable.TextArray;
 import org.neo4j.values.storable.Value;
 import org.neo4j.values.storable.Values;
+import org.neo4j.values.virtual.MapValue;
+import org.neo4j.values.virtual.NodeValue;
 
 import static java.lang.String.format;
 import static org.neo4j.collection.primitive.PrimitiveIntCollections.map;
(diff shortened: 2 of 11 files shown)


pontusmelke:
@tinwelint @thobe hates this, maybe we want it it closed please discus with him.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -292,6 +297,24 @@ public boolean isType( RelationshipType type )
             return false;
         }
 
+        @Override
+        public long getStartNodeId()
+        {
+            return 0;
```
thobe, line 303: This cannot be right.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -292,6 +297,24 @@ public boolean isType( RelationshipType type )
...
+        }
+
+        @Override
+        public long getEndNodeId()
+        {
+            return 0;
```
thobe, line 309: No way this is correct.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -292,6 +297,24 @@ public boolean isType( RelationshipType type )
...
+        }
+
+        @Override
+        public long getOtherNodeId( long id )
+        {
+            return 0;
```
thobe, line 315: Wrong.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -345,20 +368,43 @@ public String toString()
         {
             return String.format( "VirtualRelationshipHack[%s]", id );
         }
+
+        @Override
+        public NodeValue startNode()
+        {
+            return null;
```
thobe, line 375: I would think we could do better than this.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -345,20 +368,43 @@ public String toString()
...
+        }
+
+        @Override
+        public NodeValue endNode()
+        {
+            return null;
```
thobe, line 381: We should be able to do better.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -345,20 +368,43 @@ public String toString()
...
+        }
+
+        @Override
+        public TextValue type()
+        {
+            return null;
```
thobe, line 387: We can do better than this.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -345,20 +368,43 @@ public String toString()
...
+        }
+
+        @Override
+        public MapValue properties()
+        {
+            return null;
```
thobe, line 393: If we expect this to not be invoked, throwing an exception is much better than returning a value that can wreck havoc in other places.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -556,6 +602,18 @@ public Object removeProperty( String key )
             return null;
         }
 
+        @Override
+        public TextArray labels()
+        {
+            return Values.EMPTY_TEXT_ARRAY;
```
thobe, line 608: Do virtual nodes never have labels?

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/SchemaProcedure.java
```diff
@@ -220,19 +227,17 @@ private GraphResult getGraphResult( final Map<String,VirtualNodeHack> nodeMap,
...
 
-    private static class VirtualRelationshipHack implements Relationship
+    private static class VirtualRelationshipHack extends RelationshipValue implements Relationship
     {
-
         private static AtomicLong MIN_ID = new AtomicLong( -1 );
 
-        private final long id;
         private final Node startNode;
         private final Node endNode;
```
thobe, line 235: Perhaps this code would work a lot better if these were `VirtualNodeHack` objects?

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -550,13 +553,20 @@ public int compareTo( Object node )
...
     public boolean equals( Object o )
     {
-        return o instanceof Node && this.getId() == ((Node) o).getId();
+        if ( o instanceof Node )
+        {
+            return this.getId() == ((Node) o).getId();
+        }
+        else
+        {
+            return super.equals( o );
```
thobe, line 562: This gives me an extremely uneasy feeling.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -779,6 +789,25 @@ public int getDegree( RelationshipType type, Direction direction )
...
+
+    @Override
+    public TextArray labels()
+    {
+        ArrayList<String> ls = new ArrayList<>();
+        for ( Label label : getLabels() )
+        {
+            ls.add( label.name() );
+        }
+        return Values.stringArray( ls.toArray( new String[ls.size()] ) );
```
thobe, line 802: This seems like the wrong way around - if the implementation is the way it is here, with `NodeProxy` being both a `NodeValue` and a `Node`, surely the `getLabels()` method should be based on this method rather than the other way around.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -779,6 +789,25 @@ public int getDegree( RelationshipType type, Direction direction )
...
+    }
+
+    @Override
+    public MapValue properties()
+    {
+        return ValueUtils.asMapValue( getAllProperties() );
```
thobe, line 808: This seems like the wrong way around, going up the abstraction hierarchy only to go back down.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/RelationshipProxy.java
```diff
@@ -458,13 +462,20 @@ public int compareTo( Object rel )
...
     public boolean equals( Object o )
     {
-        return o instanceof Relationship && this.getId() == ((Relationship) o).getId();
+        if ( o instanceof Relationship )
+        {
+            return this.getId() == ((Relationship) o).getId();
+        }
+        else
+        {
+            return super.equals( o );
```
thobe, line 471: This feels very dangerous - since it means `Relationship` objects might now be equal to other `RelationshipValue` objects.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/RelationshipProxy.java
```diff
@@ -486,6 +497,47 @@ public String toString()
...
+    @Override
+    public TextValue type()
+    {
+        return Values.stringValue( getType().name() );
+    }
+
+    @Override
+    public MapValue properties()
+    {
+        return ValueUtils.asMapValue( getAllProperties() );
```
thobe, line 525: This seems backwards.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/CoreNodeWrappingNodeValue.java
```diff
@@ -31,13 +31,17 @@
 import org.neo4j.values.virtual.NodeValue;
 import org.neo4j.values.virtual.VirtualValues;
 
-public class NodeProxyWrappingNodeValue extends NodeValue
+/**
+ * This NodeValue wraps a Core API Node and delegates method call to the
+ * Node instance.
+ */
+public class CoreNodeWrappingNodeValue extends NodeValue
```
thobe, line 38: So even with `NodeProxy` now being a `NodeValue` we _still_ have the wrappers. Then what do we gain with this change?

community/kernel/src/main/java/org/neo4j/kernel/impl/util/CoreRelationshipWrappingValue.java
```diff
@@ -25,21 +25,25 @@
...
+import org.neo4j.values.virtual.RelationshipValue;
 import org.neo4j.values.virtual.VirtualNodeValue;
 import org.neo4j.values.virtual.VirtualValues;
 
-public class RelationshipProxyWrappingValue extends RelationshipValue
+/**
+ * This RelationshipValue wraps a Core API Relationship and delegates method calls to the
+ * Relationship instance.
+ */
+public class CoreRelationshipWrappingValue extends RelationshipValue
```
thobe, line 38: The `RelationshipProxy` objects were never the problem, and the problematic `Relationship` implementations we still wrap - so we gain nothing by this change...?

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
```diff
@@ -326,11 +326,25 @@ else if ( map.containsKey( "latitude" ) && map.containsKey( "longitude" ) )
...
     {
-        return new NodeProxyWrappingNodeValue( node );
+        if ( node instanceof NodeValue )
+        {
+            return (NodeValue) node;
+        }
+        else
+        {
+            return new CoreNodeWrappingNodeValue( node );
+        }
```
thobe, line 336: Even if `NodeProxy` were not to extend `NodeValue` we could still do something special for `NodeProxy` here, and return *sane* `NodeValue` implementations in that case, reserving the unfortunate hack of `NodeValue` with the capability of having behaviour for external implementations of `Node` - which would help us to _limit_ the spread of that disease.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
```diff
@@ -326,11 +326,25 @@ else if ( map.containsKey( "latitude" ) && map.containsKey( "longitude" ) )
...
     {
-        return new RelationshipProxyWrappingValue( relationship );
+        if ( relationship instanceof RelationshipValue )
+        {
+            return (RelationshipValue) relationship;
+        }
+        else
+        {
+            return new CoreRelationshipWrappingValue( relationship );
+        }
```
thobe, line 348: Same thing goes for here as for `fromNodeProxy()` - we can treat `RelationshipProxy` special even if it doesn't extend `RelationshipValue`.

thobe (commented):
I think this change is completely backwards. It's a shame, because the title of it holds such great promise.

We don't want to have `NodeValue` and `RelationshipValue` implementations that wrap `Node` and `Relationship` because we don't want to have `Value`s that have behaviour outside of the database. The only reason we do that today is because there are various hacks (in `db.schema()` and APOC) that rely on this ability in order to create "virtual" nodes and relationships.

This PR further cements that anti-pattern by making `NodeProxy` and `RelationshipProxy` extend `NodeValue` and `RelationshipValue`, meaning we now have `Value` implementations that have behaviour associated with them, are they really values at that point? It also makes the definition of `equals()` in `NodeProxy` and `RelationshipProxy` really strange, and potentially harmful.

Not only that, but this PR doesn't do anything to address the original case for wrapping `Node` and `Relationship` objects in `NodeValue` and `RelationshipValue` implementations. The wrappers still remain.

The only good thing this PR does is give slightly better names to what used to be called `NodeProxyWrappingNodeValue` and `RelationshipProxyWrappingValue`.

What would be better would be to change the converter methods that convert `Node` and `Relationship` objects to `NodeValue` and `RelationshipValue` instances to give `NodeProxy` and `RelationshipProxy` special treatment since they are _trusted implementations_ and return a sane and simple `NodeValue` or `RelationshipValue` implementation in those cases, leaving the wrapping-hack to the cases where we are dealing with _foreign_ `Node` and `Relationship` implementations. So while it is good to change the name of `NodeProxyWrappingNodeValue` and `RelationshipProxyWrappingValue`, perhaps even better names to change to would be `ForeignNodeValue` and `ForeignRelationshipValue`?

thobe:
Github doesn't have a review option of "reject", and "request changes" didn't seem right when what I'm asking for is for it to be completely redone from scratch. Should I just close this, or should I let others have a say first?

PR #10924, opened 2018-01-30
 Orderability for TemporalValues

sherfert:
It remained figuring out what global orderability works best for Durations. This is one attempt.


community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
```diff
@@ -226,6 +227,41 @@ else if ( seconds > 0 && nanos < 0 )
...
+                .thenComparingLong( DurationValue::getMonths )
+                .thenComparingLong( DurationValue::getDays )
+                .thenComparingLong( DurationValue::getSeconds )
+                .thenComparingInt( DurationValue::getNanos ).compare( this, other );
+    }
+
+    private long averageLengthInNanos()
+    {
+        return this.nanos + this.seconds * NANOS_PER_SECOND + this.days * NANOS_PER_SECOND * SECONDS_PER_DAY +
+                this.months * NANOS_PER_SECOND * (long) (SECONDS_PER_DAY * AVERAGE_DAYS_PER_MONTH);
```
thobe: This will overflow for durations longer than 292 years, which isn't very long in the grand scheme of things.

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
```diff
@@ -226,6 +227,41 @@ else if ( seconds > 0 && nanos < 0 )
         this.nanos = (int) nanos;
     }
 
+    public int compareTo( DurationValue other )
+    {
+        return Comparator.comparingLong(DurationValue::averageLengthInNanos)
+                .thenComparingLong( DurationValue::getMonths )
+                .thenComparingLong( DurationValue::getDays )
+                .thenComparingLong( DurationValue::getSeconds )
+                .thenComparingInt( DurationValue::getNanos ).compare( this, other );
```
thobe: Tiny implementation note: the Comparator should be statically defined, it is unnecessary to recreate a new Comparator every time we compare two DurationValue objects.

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
```diff
@@ -226,6 +227,41 @@ else if ( seconds > 0 && nanos < 0 )
...
+
+    public long getSeconds()
+    {
+        return seconds;
+    }
+
+    public int getNanos()
+    {
+        return nanos;
+    }
```
thobe: If these methods are for the comparator only, there is no need to make them public, in fact they could be inlined into the comparator as a lambda accessing the field directly (e.g. `d -> d.months`).

thobe (commented):
I have another implementation of pretty much the same logic, although maybe not as elegantly: https://github.com/thobe/neo4j/commit/b8bbab872728489eed3423b93977863bc6fa0930#diff-5cd163402f62d45ee1da73ebe101cde7R217

thobe:
It seems quite doable to define a total order for durations. What we need to be aware of is that the order we define is compatible with potential future comparison operators.

The problematic cases then are things like `P30D` and `P1M`.

The thing is that while we can't say anything about `P1M` compared to `P28D`, `P29D`, `P30D` or `P31D`, we know that `P1M` is shorter than `P32D` and longer than `P27D` - so a potential future comparison may or may not be defined in those cases.

Diving further we find:
* `P28D` <= `P1M` <= `P31D` (shortest possible: February)
* `P59D` <= `P2M` <= `P62D` (shortest possible two: Jan-Feb or Feb-Mar, longest possible two: Jul-Aug or Dec-Jan)
* `P89D` <= `P3M` <= `P92D` (shortest possible three: Feb-Apr, longest possible three: Nov-Jan or Jun-Aug or Jul-Sep)
* `P120D` <= `P4M` <= `P123D` (shortest possible four: Jan-Apr or Feb-May or Nov-Feb, longest possible four: May-Aug or Jul-Oct or Oct-Jan)
* `P150D` <= `P5M` <= `P153D` (shortest possible five: Feb-Jun, longest possible five: any range not including Feb)
* `P181D` <= `P6M` <= `P184D` (shortest possible six: Jan-Jun or Feb-Jul, longest possible six: Jul-Dec or Aug-Jan)
* `P212D` <= `P7M` <= `P215D` (shortest possible seven: any range including Feb, longest possible seven: Jul-Jan)
* `P242D` <= `P8M` <= `P245D` (shortest possible eight: Feb-Sep or Sep-Apr or Nov-Jun, longest possible eight: May-Dec or Jun-Jan)
* `P273D` <= `P9M` <= `P276D` (shortest possible nine: _many_, longest possible nine: May-Jan)
* `P303D` <= `P10M` <= `P306D` (shortest possible ten: Feb-Nov or Sep-Jun, longest possible ten: Mar-Dec or Apr-Jan)
* `P334D` <= `P11M` <= `P337D` (shortest possible eleven: _many_, longest possible eleven: Mar-Jan)
* `P365D` <= `P12M` = `P1Y` <= `P366D` (longest possible: leap year)
* `P1460D` <= `P4Y` <= `P1461D` (every 100 years, we skip a leap year)
* `P36524D` <= `P100Y` <= `P36525D` (every 400 years, we skip skipping the leap year)
* `P400Y` = `P146097D` (this it the shortest number of months where we know exactly how many days it is)

Caveat: quick calculations - some could be wrong.

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
```diff
@@ -192,10 +192,20 @@ DurationValue create(
 
     private static final DurationValue ZERO = new DurationValue( 0, 0, 0, 0 );
     private static final List<TemporalUnit> UNITS = unmodifiableList( asList( MONTHS, DAYS, SECONDS, NANOS ) );
+    // This comparator is safe until 292,271,023,045 years. After that, we have an overflow.
```
thobe, line 195: This makes it safe for all "natural" durations - i.e. the ones that can be obtained by asking for the duration between two instants that can be represented. Probably good enough, although I think we can do even better if we wanted to.

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
```diff
@@ -192,10 +192,20 @@ DurationValue create(
...
+            // At this point, the durations have the same length and we compare by the individual fields.
+            .thenComparingLong( (d) -> d.months )
+            .thenComparingLong( (d) -> d.days )
+            .thenComparingLong( (d) -> d.seconds );
+
     static final int NANOS_PER_SECOND = 1_000_000_000;
     static final long SECONDS_PER_DAY = DAYS.getDuration().getSeconds();
     /** 30.4375 days = 30 days, 10 hours, 30 minutes */
     private static final double AVERAGE_DAYS_PER_MONTH = 365.25 / 12;
+    static final long AVG_SECONDS_PER_MONTH = 2_629_800;
```
thobe: Actually `2_629_746` if we want to be really exact...

thobe: The smallest duration in months and years for which there is an exact number of days is `400 years` = `((365*4+1)*25-1)*4+1 days` = `146_097 days` which is `12_622_780_800 seconds` (discounting leap seconds, which our implementation does). Divide that by `400*12` to get the average number of seconds in a month: `2_629_746`.

At least it is an exact whole number of seconds.

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
sherfert: That would require going to BigInteger, though.

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
sherfert: Yes. Since the rest of the implementation and the specs are based on 365.25 days per year, I used this value instead.

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
thobe: Only if we want it to pass through as a single object somewhere, which isn't necessary.

PR #10928, opened 2018-01-30
Native memory allocation trackers

MishaDemianenko:
Introduce possibility to track native memory that we allocate.
allow track allocations in global and local contexts, where local contexts
can be the context of any granularity.
Update primitive off-heap collections to report memory allocation and
de-allocations.

Force all native allocations and deallocation to track 
requested and freed bytes

### DIFF
--- a/community/common/src/main/java/org/neo4j/memory/GlobalMemoryTracker.java
+++ b/community/common/src/main/java/org/neo4j/memory/GlobalMemoryTracker.java
@@ -17,32 +17,41 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-package org.neo4j.collection.primitive.hopscotch;
+package org.neo4j.memory;
 
-import org.neo4j.memory.MemoryAllocationTracker;
+import java.util.concurrent.atomic.LongAdder;
 
-public class LongKeyUnsafeTable<VALUE> extends UnsafeTable<VALUE>
+/**
+ * Global memory tracker that can be used in a global multi threaded context to record
+ * allocation and de-allocation of native memory.
+ * @see org.neo4j.memory.MemoryAllocationTracker
+ * @see MemoryTracker
+ */
+public class GlobalMemoryTracker implements MemoryTracker, MemoryAllocationTracker
 {
-    public LongKeyUnsafeTable( int capacity, VALUE valueMarker, MemoryAllocationTracker allocationTracker )
+    private LongAdder longAdder = new LongAdder();
+
+    public static final GlobalMemoryTracker INSTANCE = new GlobalMemoryTracker();
+
+    private GlobalMemoryTracker()
     {
-        super( capacity, 8, valueMarker, allocationTracker );
     }
 
     @Override
-    protected long internalKey( long keyAddress )
+    public long usedDirectMemory()
     {
-        return alignmentSafeGetLongAsTwoInts( keyAddress );
+        return longAdder.sum();
     }
 
     @Override
-    protected void internalPut( long keyAddress, long key, VALUE value )
+    public void allocate( long allocatedBytes )
     {
-        alignmentSafePutLongAsTwoInts( keyAddress, key );
+        longAdder.add( allocatedBytes );
     }
 
     @Override
-    protected Table<VALUE> newInstance( int newCapacity )
+    public void deallocate( long deAllocatedBytes )
     {
-        return new LongKeyUnsafeTable<>( newCapacity, valueMarker, allocationTracker );
+        longAdder.add( -deAllocatedBytes );
     }
 }
--- a/community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
+++ b/community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
@@ -19,6 +19,7 @@
  */
 package org.neo4j.io.mem;
 
+import org.neo4j.memory.MemoryAllocationTracker;
 import org.neo4j.unsafe.impl.internal.dragons.UnsafeUtil;
 
 import static org.neo4j.io.ByteUnit.kibiBytes;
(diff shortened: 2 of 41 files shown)


community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
```diff
@@ -162,8 +166,13 @@ private void initCause( NativeMemoryAllocationRefusedError error, OutOfMemoryErr
...
     {
         super.finalize();
-        Grab current = grabs;
+        close();
+    }
 
+    @Override
+    public void close()
+    {
+        Grab current = grabs;
```
chrisvest: The tests that need this could also call `finalize` directly. The reason we free in `finalize` is because we rely on the GC to tell us when there are no more page cursors that can access any native memory. Having a `close` method makes it look like a mistake that we are not calling it anywhere in the product code (except for this finalizer). If you insist on having it, then we should strengthen the documentation around the particular way that we are keeping track of the life cycle of the native memory in the page cache, and also the method should be `synchronized` to match its siblings in this class.

community/primitive-collections/src/main/java/org/neo4j/collection/primitive/hopscotch/UnsafeTable.java
```diff
@@ -219,4 +225,14 @@ protected static long alignmentSafeGetLongAsTwoInts( long address )
         long msb = UnsafeUtil.getInt( address + Integer.BYTES ) & 0xFFFFFFFFL;
         return lsb | (msb << Integer.SIZE);
     }
+
+    private long allocateMemory( long bytesToAllocate )
+    {
+        return UnsafeUtil.allocateMemory( bytesToAllocate, allocationTracker );
```
klaren: Previously this was done in the constructor and could only be done once. Here we allow more than one call and a potential memory leak.

community/unsafe/src/main/java/org/neo4j/unsafe/impl/internal/dragons/UnsafeUtil.java
```diff
@@ -350,14 +352,15 @@ public static String newSharedArrayString( char[] chars )
      * The memory is aligned such that it can be used for any data type.
      * The memory is uninitialised, so it may contain random garbage, or it may not.
      */
-    public static long allocateMemory( long sizeInBytes )
+    public static long allocateMemory( long sizeInBytes, MemoryAllocationTracker allocationTracker )
```
klaren, line 361: Since the `MemoryAllocationTracker` is only different in tests, and all other calls refer to the global singleton I would prefer an overload instead.
```java
    public static long allocateMemory( long sizeInBytes )
    {
        return allocateMemory( sizeInBytes, GlobalMemoryTracker.INSTANCE );
    }
```
Same thing could be done for the pimitive collections as well.

community/common/src/main/java/org/neo4j/memory/GlobalMemoryTracker.java
```diff
@@ -0,0 +1,57 @@
...
+ * @see MemoryTracker
+ */
+public class GlobalMemoryTracker implements MemoryTracker, MemoryAllocationTracker
+{
+    private LongAdder longAdder = new LongAdder();
+
+    public static final GlobalMemoryTracker INSTANCE = new GlobalMemoryTracker();
+
+    private GlobalMemoryTracker()
+    {
```
klaren, line 37: Follow the best practice and `throw new AssertionError`.

community/io/src/main/java/org/neo4j/io/mem/GrabAllocator.java
klaren: Definitely add a comment for future victims to digest :)

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/PageListTest.java
```diff
@@ -87,7 +89,7 @@
     public static void setUpStatics()
     {
         executor = Executors.newCachedThreadPool( new DaemonThreadFactory() );
-        mman = MemoryAllocator.createAllocator( "1 MiB" );
+        mman = MemoryAllocator.createAllocator( "1 MiB", new LocalMemoryTracker() );
```
klaren: Same here.

community/io/src/test/java/org/neo4j/io/pagecache/impl/muninn/LargePageListIT.java
```diff
@@ -43,9 +45,9 @@ public void veryLargePageListsMustBeFullyAccessible() throws Exception
         long pageCacheSize = ByteUnit.gibiBytes( 513 ) + pageSize;
         int pages = Math.toIntExact( pageCacheSize / pageSize );
 
-        MemoryAllocator mman = MemoryAllocator.createAllocator( "2 GiB" );
+        MemoryAllocator mman = MemoryAllocator.createAllocator( "2 GiB", new LocalMemoryTracker() );
```
klaren: This one is using a local tracker while the others in this tests are using the global instance. Looks suspicious.

klaren (changes requested):
Overall good, a few comments here and there.

I noticed we never expose `usedDirectMemory()` anywhere, but I'm guessing there is a future PR for that.

community/common/src/main/java/org/neo4j/memory/GlobalMemoryTracker.java
MishaDemianenko: thats not really possible since we still need one instance :)

klaren: Silly me :see_no_evil:

MishaDemianenko:
@chrisvest @klaren all comments should be fixed now, they are in a separate commit to simplify review.
I have restored finalization, let's fight about that in some future pr instead

chrisvest (approved):

klaren (approved):

tinwelint (approved):

PR #10929, opened 2018-01-30
Bump checkstyle plugin and tool version

MishaDemianenko:


klaren (approved):
Looks good! Now, if we could only sneak in the built-in `google_checks.xml` :smile:

PR #10930, opened 2018-01-30
Move feature toggles out of the dungeon to commons

MishaDemianenko:
Feature toggles are commonly used across the modules and there is no
reason to keep them in a package that is not related to what they are
doing.

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/transport/NettyServer.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/transport/NettyServer.java
@@ -39,7 +39,7 @@ import org.neo4j.helpers.PortBindException;
 import org.neo4j.kernel.configuration.BoltConnector;
 import org.neo4j.kernel.configuration.ConnectorPortRegister;
 import org.neo4j.kernel.lifecycle.LifecycleAdapter;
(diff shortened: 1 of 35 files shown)


community/common/src/main/java/org/neo4j/helpers/FeatureToggles.java
```diff
@@ -17,7 +17,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-package org.neo4j.unsafe.impl.internal.dragons;
+package org.neo4j.helpers;
```
chrisvest: Does this mean it's public API now? I'd prefer to keep it non-public.

chrisvest (commented):
Just one comment.

community/common/src/main/java/org/neo4j/helpers/FeatureToggles.java
MishaDemianenko: Its really good point; thats wrong them - have moved into another package `org.neo4j.util` inside commons

MishaDemianenko:
@chrisvest updated

chrisvest (approved):

PR #10931, opened 2018-01-30
Raft pre elections should be compatible with refuse_to_be_leader

andrewkerr9000:


andrewkerr9000:
I'll rearrange code a bit to make the diff clearer and get rid of the TODOs. Also fix committer name.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Follower.java
```diff
@@ -167,49 +192,127 @@ public Outcome handle( RaftMessages.NewEntry.BatchRequest batchRequest ) throws
...
+        private static ElectionTimeoutHandler instance = new PreVoteUnsupportedHandler();
     }
 
-    class PreVoteActiveHandler extends PreVoteSupportedHandler
+    private static class PreVoteUnsupportedRefusesToLead implements ElectionTimeoutHandler
     {
+        @Override
+        public Outcome handle( RaftMessages.Timeout.Election election, Outcome outcome, ReadableRaftState ctx, Log log ) throws IOException
+        {
+            log.info( "Election timeout triggered but refusing to be leader" );
```
martinfurmanski, line 247: This will be very useful in debugging!

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Follower.java
```diff
@@ -167,49 +192,127 @@ public Outcome handle( RaftMessages.NewEntry.BatchRequest batchRequest ) throws
...
+    {
+        @Override
+        public Outcome handle( RaftMessages.Timeout.Election election, Outcome outcome, ReadableRaftState ctx, Log log ) throws IOException
         {
-            super( ctx, log );
+            log.info( "Election timeout triggered but refusing to be leader" );
+            Set<MemberId> memberIds = ctx.votingMembers();
+            if ( memberIds != null && memberIds.contains( ctx.myself() ) )
+            {
+                outcome.setPreElection( true );
```
martinfurmanski, line 263: Why are we manipulating this if we are anyway refusing to be leader?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Follower.java
```diff
@@ -81,17 +81,24 @@ public Outcome handle( RaftMessages.RaftMessage message, ReadableRaftState ctx,
...
+        private final ElectionTimeoutHandler electionTimeoutHandler;
 
-        Handler( ReadableRaftState ctx, Log log )
+        Handler( PreVoteRequestHandler preVoteRequestHandlerDelegate, PreVoteResponseHandler preVoteResponseHandler,
+                ElectionTimeoutHandler electionTimeoutHandler, ReadableRaftState ctx, Log log )
         {
             this.ctx = ctx;
             this.log = log;
             this.outcome = new Outcome( FOLLOWER, ctx );
+            this.preVoteRequestHandlerDelegate = preVoteRequestHandlerDelegate;
```
martinfurmanski: Is there a reason to single this one out with Delegate in the name? Hard for me to see without IDE-support, but can't really notice anything special through cursory browsing here.

martinfurmanski:
I had just very minor comments. I think the functionality looks good enough for a merge, however the PR run seems to have failed miserably. Don't know if just flaky or perhaps some lifecycle issue in test refactoring.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Follower.java
andrewkerr9000: Probably an oversight from an earlier version.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/consensus/roles/Follower.java
andrewkerr9000: `preElection` needs to be true to participate in pre elections (`PreVoteRequestVotingHandler`). However it won't send pre vote requests and it will ignore pre vote responses.

See lines 361-376.

andrewkerr9000:
Looks like a PreElectionIT is failing, and causing a cascade of failures by not cleaning up after itself.

andrewkerr9000:
@martinfurmanski Tests should be stable now.

tl;dr: we need to prevent members that refuse to be leader bootstrapping the cluster. HazelcastDiscoveryService already did this, but SharedDiscoveryService did not. Unfortunately the logic is coupled to implementation specific data structures and cannot easily be extracted further up the call hierarchy.

andrewkerr9000:
Cypher test failed, rebased.

martinfurmanski (approved):

PR #10932, opened 2018-01-30
Remove snapshoting of index provider metafiles

MishaDemianenko:
Cleanup index provided meta files snapshoting since we do not have any.


burqen (approved):
LGTM

PR #10935, opened 2018-01-30
Basic implementation of native schema string indexing layer

tinwelint:
Much of the changes in here are about making test classes and utilities
generic to value type as well as copy type-specific indexing code from
number schema index classes.

Here's also a basic, non-optimized implementation of the string schema
index accessor, string layout and friends which focuses on getting the
functionality in place first and foremost. This string index isn't hooked in
at any place, only some tests assert its functionality.

There are still some TODOs scattered around the code, which are mostly 
points to attack next or to consider more carefully when moving into
performance and integration work, please don't take too much notice
of them as they are here mostly for reminders when doing further development.

co-author: @burqen


community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringLayout.java
```diff
@@ -0,0 +1,105 @@
...
+abstract class StringLayout extends Layout.Adapter<StringSchemaKey,NativeSchemaValue>
+{
+    @Override
+    public StringSchemaKey newKey()
+    {
+        return new StringSchemaKey();
+    }
+
+    @Override
+    public StringSchemaKey copyKey( StringSchemaKey key, StringSchemaKey into )
```
MishaDemianenko, line 40: can you please describe when it should be used (like node splitting) because its not clear why its used for example during the consistency check

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * Includes value and entity id (to be able to handle non-unique values). A value can be any {@link String},
+ * or rather any string that {@link GBPTree} can handle.
+ */
+class StringSchemaKey extends ValueWriter.Adapter<RuntimeException> implements NativeSchemaKey
+{
+    static final int ENTITY_ID_SIZE = Long.BYTES;
+
+    private long entityId;
+    private boolean entityIdIsSpecialTieBreaker;
```
MishaDemianenko: what is this field about?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringLayoutNonUnique.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.neo4j.index.internal.gbptree.Layout;
+
+/**
+ * {@link Layout} for strings that doesn't need to be unique.
+ */
+public class StringLayoutNonUnique extends StringLayout
+{
+    private static final String IDENTIFIER_NAME = "NUSI";
+    static final int MAJOR_VERSION = 0;
```
MishaDemianenko: all constants here can be private

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringLayoutUnique.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.neo4j.index.internal.gbptree.Layout;
+
+/**
+ * {@link Layout} for strings that need to be unique.
+ */
+class StringLayoutUnique extends StringLayout
+{
+    private static final String IDENTIFIER_NAME = "USI";
+    static final int MAJOR_VERSION = 0;
```
MishaDemianenko: all fields here can be private

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaIndexProvider.java
```diff
@@ -0,0 +1,120 @@
...
+import org.neo4j.values.storable.ValueGroup;
+
+/**
+ * Schema index provider for native indexes backed by e.g. {@link GBPTree}.
+ */
+public class StringSchemaIndexProvider extends NativeSchemaIndexProvider<StringSchemaKey,NativeSchemaValue>
+{
+    public static final String KEY = "string";
+    public static final Descriptor STRING_PROVIDER_DESCRIPTOR = new Descriptor( KEY, "1.0" );
+    static final IndexCapability CAPABILITY = new StringIndexCapability();
```
MishaDemianenko: CAPABILITY field should be private

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaIndexProvider.java
```diff
@@ -0,0 +1,120 @@
...
+import org.neo4j.kernel.impl.api.index.sampling.IndexSamplingConfig;
+import org.neo4j.values.storable.ValueGroup;
+
+/**
+ * Schema index provider for native indexes backed by e.g. {@link GBPTree}.
+ */
+public class StringSchemaIndexProvider extends NativeSchemaIndexProvider<StringSchemaKey,NativeSchemaValue>
+{
+    public static final String KEY = "string";
+    public static final Descriptor STRING_PROVIDER_DESCRIPTOR = new Descriptor( KEY, "1.0" );
```
MishaDemianenko: As I can see we have versions for layout and for the provider, how would potential future supported migration paths look like?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaIndexProvider.java
```diff
@@ -0,0 +1,120 @@
...
+                indexId, samplingConfig );
+    }
+
+    @Override
+    public IndexCapability getCapability( IndexDescriptor indexDescriptor )
+    {
+        return CAPABILITY;
+    }
+
+    private static class StringIndexCapability implements IndexCapability
```
MishaDemianenko: please add a javadoc that describe underlying capabilities

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaIndexReader.java
```diff
@@ -0,0 +1,117 @@
...
+    void validateQuery( IndexOrder indexOrder, IndexQuery[] predicates )
+    {
+        if ( predicates.length != 1 )
+        {
+            throw new UnsupportedOperationException();
+        }
+
+        if ( indexOrder != IndexOrder.NONE )
+        {
+            throw new UnsupportedOperationException( "unsupported order " + indexOrder );
```
MishaDemianenko: In StringCapabilities we claim that we support `ASCENDING` index order. Why do we not allow it here?, and shouldn't we use capabilities here for validation of predicates as well?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaIndexReader.java
```diff
@@ -0,0 +1,117 @@
...
+        IndexQuery predicate = predicates[0];
+        switch ( predicate.type() )
+        {
+        case exists:
+            treeKeyFrom.initAsLowest();
+            treeKeyTo.initAsHighest();
+            break;
+        case exact:
+            ExactPredicate exactPredicate = (ExactPredicate) predicate;
+            treeKeyFrom.from( Long.MIN_VALUE, exactPredicate.value() );
```
MishaDemianenko, line 64: i find following predicates too low level at this point: key.initAsLowest and key.initAsHighest are really good examples how keys should be initialized on this level and I would prefer to have helper methods instead that will perform initialisations for exact and range cases as well

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
```diff
@@ -0,0 +1,226 @@
...
+    public void from( long entityId, Value... values )
+    {
+        this.entityId = entityId;
+        entityIdIsSpecialTieBreaker = false;
+        assertValidValue( values ).writeTo( this );
+    }
+
+    private TextValue assertValidValue( Value... values )
+    {
+        if ( values.length > 1 )
```
MishaDemianenko, line 87: this more or less duplicate capability checks, maybe use that instead?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
```diff
@@ -0,0 +1,226 @@
...
+     * Compares the value of this key to that of another key.
+     * This method is expected to be called in scenarios where inconsistent reads may happen (and later retried).
+     *
+     * @param other the {@link StringSchemaKey} to compare to.
+     * @return comparison against the {@code other} {@link StringSchemaKey}.
+     */
+    int compareValueTo( StringSchemaKey other )
+    {
+        // TODO cover all cases of bytes == null and special tie breaker and document
+        if ( bytes != other.bytes )
```
MishaDemianenko, line 147: this is wrong: for example, comparing highest and lowest key now will say they are the same

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
```diff
@@ -0,0 +1,226 @@
...
+        for ( int i = 0; i < length; i++ )
+        {
+            int compare = Byte.compare( a[i], b[i] );
+            if ( compare != 0 )
+            {
+                return compare;
+            }
+        }
+
+        if ( a.length < b.length )
```
MishaDemianenko: Integer.compare( a.length, b.length )

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        try
+        {
+            // TODO change to not throw
+            return codePointByteArrayCompare( bytes, other.bytes );
+        }
+        catch ( Exception e )
+        {
+            // We can not throw here because we will visit this method inside a pageCursor.shouldRetry() block.
```
MishaDemianenko, line 170: not sure if its safe to do a fallback here since it will mean that comparison order is wrong anyway

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/NativeSchemaIndexProviderTest.java
```diff
@@ -0,0 +1,422 @@
...
+                fail( "Should have failed" );
+            }
+            catch ( IndexEntryConflictException e )
+            {
+                // good
+            }
+        }
+    }
+
+    protected abstract Value someValue();
```
MishaDemianenko: nice to see this hidden abstract method between the tests, otherwise its to easy to find all f those abstract things...

community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/NumberSchemaIndexAccessorTest.java
```diff
@@ -81,4 +82,28 @@ public void respectIndexOrder() throws Exception
             assertTrue( "found all values", i == expectedValues.length );
```
MishaDemianenko: assertEquals?

MishaDemianenko (changes requested):
Couple of comments here and there, have tried to avoid commenting todos here and there as noted in description

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringLayout.java
burqen: The `ConsistencyChecker` that is using it is actually the consistency checker for the `GBPTree` itself, not the Neo consistency checker. You use it when you want a copy of the key that you can make changes to without destroying the original state.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
burqen: Keys look like `<stringValue,nodeId>`. We want to do a range scan over `"A"-"C"` _inclusive_ in both ends. We initiate our keys that define our range as `<"A",Long.MIN_VALUE> - <"C",Long.MAX_VALUE>` which will cover the entire range of possible values. The problem is that if do not explicitly take the node id part into account we will only compare against the string value in an exclusive fashion. So to tell the comparator that he needs to take the id into consideration as well we do `entityIdIsSpecialTieBreaker=true`.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaIndexProvider.java
burqen: Keep `native-lucene-1.0` around as is.
Introduce new `native-lucene-2.0` that include native string.
Do no automatic migration and instead let user use drop+create just as when we introduced native number.

So far we have not had the need to utilise the layout version but it's their so that we _can_ if we need in the future. One example would be if we wound move native number layout from "fixed" to "dynamic". The migration path in that case is not super clear.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
burqen: Yes, the validation of "what goes where" happens all over the place right now. Created a card for cleaning this up. Ok if we leave as future work?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
burqen: It's actually ok because we want to `compareValueInThisKey` and not `compareKey`. Validation of highest and lowest is done outside of this method.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
burqen: This will be re-written in future work.

burqen:
@MishaDemianenko Comments addressed and rebase done :D

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
```diff
@@ -0,0 +1,226 @@
...
+        for ( int i = 0; i < length; i++ )
+        {
+            int compare = Byte.compare( a[i], b[i] );
+            if ( compare != 0 )
+            {
+                return compare;
+            }
+        }
+
+        if ( Integer.compare( a.length, b.length ) < 0 )
```
MishaDemianenko: _whole_ thing below this line can be replaced with single Integer.compare

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
MishaDemianenko: then why not call this field `compareIds` or something that will not force future reader to break his tie?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaIndexProvider.java
MishaDemianenko: talked offline

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaIndexReader.java
MishaDemianenko: well, initFromForRange and initFromForRange is still here

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
MishaDemianenko: oke

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
MishaDemianenko: your future reader would be glad to know that there are specific values that considered outside of this method and also be curious why its so

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
MishaDemianenko: All of the lines below the first comparison can be replaced with only one integer.compare thingy

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
burqen: Haha, yes ofc! xD

burqen:
@MishaDemianenko One more try

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
burqen: Just like the current readers then ;) Do you think javadoc would be enough?

MishaDemianenko (approved):

PR #10936, opened 2018-01-31
Remove entity manipulation methods from GraphDatabaseQueryService

fickludd:
Minor simplification of cypher-kernel interaction surface

Depends on https://github.com/neo4j/neo4j-documentation/pull/176 to fix docs build.


SaschaPeukert (approved):

henriknyman:
@fickludd If I remember correctly `GraphDatabaseQueryService` was introduced as an intentional wrapper to streamline the surface between kernel and Cypher, and as a step to make Cypher independent on the Core API. This PR seems to undo these objectives.

fickludd:
@henriknyman That might be the case, but since that work wasn't continued I think we just ended up with another interface without very much value. The Kernel API project + cypher versioning at the logical plan will enable that goal much better than this interface, I just started cleaning up a little.

PR #10938, opened 2018-01-31
Native indexes recovery cleanup.

MishaDemianenko:
Do not try to recover native label scan store on CC.
Allow native scan store to shutdown even if init failed.
Cleanup in RecoverRequiredChecker.

### DIFF
--- a/community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
+++ b/community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
@@ -37,7 +37,6 @@ import org.neo4j.consistency.statistics.VerboseStatistics;
 import org.neo4j.function.Suppliers;
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
 import org.neo4j.helpers.progress.ProgressMonitorFactory;
-import org.neo4j.index.internal.gbptree.RecoveryCleanupWorkCollector;
 import org.neo4j.io.fs.DefaultFileSystemAbstraction;
 import org.neo4j.io.fs.FileSystemAbstraction;
 import org.neo4j.io.pagecache.PageCache;
@@ -244,7 +243,7 @@ public class ConsistencyCheckService
 
             LabelScanStore labelScanStore =
                     new NativeLabelScanStore( pageCache, storeDir, FullStoreChangeStream.EMPTY, true, monitors,
-                            RecoveryCleanupWorkCollector.IMMEDIATE );
+                            RECOVERY_PREVENTING_COLLECTOR );
             life.add( labelScanStore );
 
             int numberOfThreads = defaultConsistencyCheckThreadsNumber();
--- a/community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
+++ b/community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
@@ -37,6 +37,7 @@ import org.neo4j.consistency.statistics.VerboseStatistics;
 import org.neo4j.function.Suppliers;
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
 import org.neo4j.helpers.progress.ProgressMonitorFactory;
+import org.neo4j.index.internal.gbptree.RecoveryCleanupWorkCollector;
 import org.neo4j.io.fs.DefaultFileSystemAbstraction;
 import org.neo4j.io.fs.FileSystemAbstraction;
 import org.neo4j.io.pagecache.PageCache;
@@ -63,7 +64,6 @@ import org.neo4j.logging.Log;
 import org.neo4j.logging.LogProvider;
 
 import static java.lang.String.format;
-import static org.neo4j.consistency.internal.SchemaIndexExtensionLoader.RECOVERY_PREVENTING_COLLECTOR;
 import static org.neo4j.consistency.internal.SchemaIndexExtensionLoader.instantiateKernelExtensions;
 import static org.neo4j.consistency.internal.SchemaIndexExtensionLoader.loadSchemaIndexProviders;
 import static org.neo4j.io.file.Files.createOrOpenAsOuputStream;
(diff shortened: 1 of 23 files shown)


community/consistency-check/src/main/java/org/neo4j/consistency/ConsistencyCheckService.java
```diff
@@ -243,7 +242,7 @@ public Result runFullConsistencyCheck( final File storeDir, Config config, Progr
 
             LabelScanStore labelScanStore =
                     new NativeLabelScanStore( pageCache, storeDir, FullStoreChangeStream.EMPTY, true, monitors,
-                            RecoveryCleanupWorkCollector.IMMEDIATE );
+                            RECOVERY_PREVENTING_COLLECTOR );
```
burqen: First of all, it is clear that we should never execute the cleanup job because consistency check should be a read only thing. This change I like!
However, I don't think it is the right thing to block consistency checker from doing it's job since we know that we are not doing a "real" recovery of the store. At the same time, I don't think silently ignoring is a good thing either. Perhaps we could report a dirty tree as any other inconsistency? What do you think?
If this turns out to be a big change, it can wait for some future PR.

burqen (changes requested):
Nice cleanups!
It's not super clear to be what would be the correct behaviour. See comment and let's discuss.

burqen:
@MishaDemianenko Did we get anywhere with this?

MishaDemianenko:
@burqen sorry didn't have time to come back into this, will try to do things as we agreed as soon as i will switch back into it

MishaDemianenko:
@burqen introduced CC checks for the dirty label and native indexes. Not sure I'm in the fan zone for that kind of check, but let's see wdyt.

community/consistency-check/src/test/java/org/neo4j/consistency/checking/AllNodesInStoreExistInLabelIndexTest.java
```diff
@@ -83,6 +86,31 @@ public void mustReportSuccessfulForConsistentLabelScanStore() throws Exception
         assertTrue( "Expected consistency check to succeed", result.isSuccessful() );
     }
 
+    @Test
+    public void reportNotCleanLabelIndex() throws IOException, ConsistencyCheckIncompleteException
```
burqen, line 90: This test a label scan store that is actually missing some label and thus "really" is corrupt. We also want to test the situation where label index actually contain the correct information but is dirty.

Something like this:
```
    @Test
    public void reportNotCleanLabelIndexWithCorrectData() throws IOException, ConsistencyCheckIncompleteException
    {
        File storeDir = db.getStoreDir();
        someData();
        db.resolveDependency( CheckPointer.class ).forceCheckPoint( new SimpleTriggerInfo( "forcedCheckpoint" ) );
        File labelIndexFileCopy = new File( storeDir, "label_index_copy" ); // <- COPIED FILE HAS ALL THE STUFFZ

        copyFile( new File( storeDir, NativeLabelScanStore.FILE_NAME ), labelIndexFileCopy );

        db.shutdownAndKeepStore();

        copyFile( labelIndexFileCopy, new File( storeDir, NativeLabelScanStore.FILE_NAME ) );

        ConsistencyCheckService.Result result = fullConsistencyCheck();
        assertTrue( "Expected consistency check to fail", result.isSuccessful() ); // <- SUCCESSFUL!
        assertThat( readReport( result ),
                hasItem( containsString("WARN : Label index was not properly shutdown and rebuild is required.") ) );
    }
```

community/consistency-check/src/test/java/org/neo4j/consistency/checking/IndexConsistencyIT.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private final AssertableLogProvider log = new AssertableLogProvider();
+    private static final Label[] LABELS = new Label[]{LABEL_ONE, LABEL_TWO, LABEL_THREE};
+    private static final String PROPERTY_KEY = "numericProperty";
+    private static final double DELETE_RATIO = 0.2;
+    private static final double UPDATE_RATIO = 0.2;
+    private static final int NODE_COUNT_BASELINE = 10;
+
+    @Test
+    public void reportNotCleanNativeIndex() throws IOException, ConsistencyCheckIncompleteException
```
burqen, line 77: Same thing goes for this test

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
```diff
@@ -1297,4 +1301,9 @@ private void closeCursor()
             }
         }
     }
+
+    public boolean isDirty()
```
burqen: Misleading name since state could have changed since startup, `wasDirtyOnStartup` or something to that effect. Also, this functionality already exist through the `Monitor`. If we want to rely on that we can edit `private GBPTree.Monitor treeMonitor()` in `NativeLabelScanStore` and `NativeSchemaNumberIndex`.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStore.java
```diff
@@ -467,12 +470,20 @@ public void stop() throws IOException
     @Override
     public void shutdown() throws IOException
     {
-        index.close();
+        if ( index != null )
+        {
+            index.close();
```
burqen, line 475: Maybe take the opportunity to set `index = null` here?

community/kernel/src/test/java/org/neo4j/kernel/impl/index/labelscan/NativeLabelScanStoreRebuildTest.java
```diff
@@ -108,6 +95,46 @@ public void mustBeDirtyIfFailedDuringRebuild() throws Exception
...
+        nativeLabelScanStore.start();
+
+        assertTrue( monitor.notValid );
+        assertFalse( monitor.rebuilt );
+        assertFalse( monitor.rebuilding );
+        nativeLabelScanStore.shutdown();
+    }
+
+    @Test
+    public void labelScanStoreIdDirtyWhenIndexIsNotClean() throws IOException
```
burqen: labelScanStore**Is**DirtyWhenIndexIsNotClean

burqen (changes requested):
I think this is a quite nice compromise between "kill checkpoint on dirty state" and "silently allow checkpoint even if state is dirty".

Found some improvements here and there.

community/index/src/main/java/org/neo4j/index/internal/gbptree/GBPTree.java
MishaDemianenko: no i will not gonna use monitor for that, its not a monitor business to return you some kind of object state, that just wrong

burqen: Alright, just the name change then 😄

MishaDemianenko:
@burqen here we go. Comments should be resolved and also includes `the renaming` :)

burqen (approved):

PR #10939, opened 2018-01-31
More Core API migration

fickludd:
Migrate more Core API methods to use Kernel API

### DIFF
--- a/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
+++ b/community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
@@ -99,9 +99,9 @@ final class TransactionBoundQueryContext(txContext: TransactionalContextWrapper)
   override def createRelationship(start: Node, end: Node, relType: String) =
     start.createRelationshipTo(end, withName(relType))
 
-  override def createRelationship(start: Long, end: Long, relType: Int) = {
+  override def createRelationship(start: Long, end: Long, relType: Int): Relationship = {
     val relId = txContext.statement.dataWriteOperations().relationshipCreate(relType, start, end)
-    relationshipOps.getById(relId)
+    entityAccessor.newRelationshipProxy(relId, start, relType, end)
   }
 
   override def getOrCreateRelTypeId(relTypeName: String): Int =
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
@@ -919,7 +919,7 @@ public class GraphDatabaseFacade implements GraphDatabaseAPI, EmbeddedProxySPI
         }
     }
 
-    private static abstract class PrefetchingNodeResourceIterator implements ResourceIterator<Node>
+    private abstract static class PrefetchingNodeResourceIterator implements ResourceIterator<Node>
     {
         private final Statement statement;
         private final NodeFactory nodeFactory;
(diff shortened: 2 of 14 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -888,21 +922,48 @@ public GraphPropertiesProxy newGraphPropertiesProxy()
...
+        }
+
+        @Override
+        void closeResources( Statement statement )
+        {
+            IOUtils.closeAllSilently( statement, cursor );
+        }
+    }
+
+    private abstract class PrefetchingNodeResourceIterator implements ResourceIterator<Node>
```
pontusmelke: Maybe a bit picky but it is not really prefetching, at least not in the way the normal `PrefetchingIterator` is. 

Furthermore I think we should make this static, I know wasn't done here but looking at it now I think it would be nicer to have it static. Only reason it is not is the `newNodeProxy` but it should be easy enough to fix.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
```diff
@@ -99,9 +99,9 @@ final class TransactionBoundQueryContext(txContext: TransactionalContextWrapper)
   override def createRelationship(start: Node, end: Node, relType: String) =
     start.createRelationshipTo(end, withName(relType))
 
-  override def createRelationship(start: Long, end: Long, relType: Int) = {
+  override def createRelationship(start: Long, end: Long, relType: Int): Relationship = {
     val relId = txContext.statement.dataWriteOperations().relationshipCreate(relType, start, end)
-    relationshipOps.getById(relId)
+    entityAccessor.newRelationshipProxy(relId, start, relType, end)
```
pontusmelke, line 104: This change should probably be done in `2.3` and in `runtime.interpreted` as well

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -757,43 +768,73 @@ public String toString()
         return spi.name() + " [" + getStoreDir() + "]";
     }
 
-    private static class PropertyValueFilteringNodeIdIterator extends PrimitiveLongCollections.PrimitiveLongBaseIterator
+    private class NodeLabelPropertyIterator extends PrefetchingNodeResourceIterator
```
pontusmelke: Why not static? I would prefer having it static

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
```diff
@@ -885,21 +926,48 @@ public GraphPropertiesProxy newGraphPropertiesProxy()
     {
         return new GraphPropertiesProxy( this );
     }
-    private static final long UNINITIALIZED = -2L;
-    private static final long NO_ID = -1L;
 
-    private final class NodeCursorResourceIterator<CURSOR extends NodeIndexCursor> implements ResourceIterator<Node>
+    private final class NodeCursorResourceIterator<CURSOR extends NodeIndexCursor> extends PrefetchingNodeResourceIterator
```
pontusmelke: I think it is me being lazy but I think this one should be static as well.

pontusmelke (approved):
Needs a rebase, otherwise looks really nice.

I had some minor comments, fix now, later or ignore completely.

community/kernel/src/main/java/org/neo4j/kernel/impl/factory/GraphDatabaseFacade.java
fickludd: Why isn't it prefetching? It has `fetchNext` and keeps it in `next`. You can change it if you like, but I feel like it's prefetching, so left it like that :).

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/TransactionBoundQueryContext.scala
fickludd: Yeah that's a good idea. Did that.

PR #10940, opened 2018-02-01
More usage of Kernel API in compiled runtime

pontusmelke:
Ports the following to use Kernel API instead:
- all token operations
- node index seek
- node by id

### DIFF
--- a/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/compiled/codegen/ir/LoopDataGenerator.scala
+++ b/enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/compiled/codegen/ir/LoopDataGenerator.scala
@@ -22,16 +22,16 @@ package org.neo4j.cypher.internal.compatibility.v3_4.runtime.compiled.codegen.ir
 import org.neo4j.cypher.internal.compatibility.v3_4.runtime.compiled.codegen.spi.MethodStructure
 import org.neo4j.cypher.internal.compatibility.v3_4.runtime.compiled.codegen.{CodeGenContext, Variable}
 
-// Generates the code that moves data into local variables from the iterator being consumed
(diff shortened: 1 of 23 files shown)


enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/compiled/codegen/ir/LoopDataGenerator.scala
```diff
@@ -22,16 +22,16 @@ package org.neo4j.cypher.internal.compatibility.v3_4.runtime.compiled.codegen.ir
...
+// Generates the code that moves data into local variables from the iterator or cursor being consumed.
 trait LoopDataGenerator {
 
-  def hasNext[E](generator: MethodStructure[E], iterVar: String): E
+  def checkNext[E](generator: MethodStructure[E], iterVar: String): E
 
   def init[E](generator: MethodStructure[E])(implicit context: CodeGenContext): Unit
 
-  def produceNext[E](nextVar: Variable, iterVar: String, generator: MethodStructure[E])(implicit context: CodeGenContext): Unit
+  def getNext[E](nextVar: Variable, iterVar: String, generator: MethodStructure[E])(implicit context: CodeGenContext): Unit
```
henriknyman, line 32: These method renamings seem slightly superfluous to me

henriknyman (approved):
Variables holding cursors are still called `iterVar`, but I guess it is fine with the way they are being used.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/compiled/codegen/ir/LoopDataGenerator.scala
pontusmelke: Yeah you might be right, now the LoopDataGenerator is used both for cursors and iterators in the cursor case we are not really producing, that is done in the `next` call.

PR #10943, opened 2018-02-01
Fix bug in bytecode generation for multiple ands and ors

henriknyman:
This affected e.g. the generated equals-methods used by grouping keys
for distinct and aggregation, so that only the first two keys (in a
compile-time determined accidental ordering) were compared.

### DIFF
--- a/community/codegen/src/test/java/org/neo4j/codegen/CodeGenerationTest.java
+++ b/community/codegen/src/test/java/org/neo4j/codegen/CodeGenerationTest.java
@@ -1057,6 +1057,150 @@ public class CodeGenerationTest
     }
 
     @Test
+    public void shouldGenerateMethodUsingMultipleAnds() throws Throwable
+    {
+        // given
+        ClassHandle handle;
+        try ( ClassGenerator simple = generateClass( "SimpleClass" ) )
+        {
+            try ( CodeBlock conditional = simple.generateMethod( boolean.class, "conditional",
+                    param( boolean.class, "test1" ),
+                    param( boolean.class, "test2" ),
+                    param( boolean.class, "test3" ) ) )
+            {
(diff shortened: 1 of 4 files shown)


community/codegen/src/test/java/org/neo4j/codegen/CodeGenerationTest.java
```diff
@@ -1056,6 +1056,150 @@ public void shouldGenerateMethodUsingAnd() throws Throwable
...
+
+        // then
+        assertThat( conditional.invoke( true, true, true ), equalTo( true ) );
+        assertThat( conditional.invoke( true, false, true ), equalTo( false ) );
+        assertThat( conditional.invoke( false, true, true ), equalTo( false ) );
+        assertThat( conditional.invoke( false, false, true ), equalTo( false ) );
+        assertThat( conditional.invoke( true, true, false ), equalTo( false ) );
+        assertThat( conditional.invoke( true, false, false ), equalTo( false ) );
+        assertThat( conditional.invoke( false, true, false ), equalTo( false ) );
+        assertThat( conditional.invoke( false, false, false ), equalTo( false ) );
```
fickludd, line 1092: These 4 tests are great, if somewhat similar. Did you consider extracting common structure, or you just prefer it like this? I'm fine with either, just curious.

fickludd (approved):
What a crisp and understandable bug. Nice fix. Good tests. Approved.

community/codegen/src/test/java/org/neo4j/codegen/CodeGenerationTest.java
henriknyman: I don't have a really good excuse, but we tend to allow more duplication in test code and this implementation is not expected to change very often.

PR #10944, opened 2018-02-01
Fixes as issue with applying index updates to unique indexes

tinwelint:
Given a uniqueness contraint on Foo:bar. If a transaction make changes
such that in the end, the constraint is not violated but some intermediate
individual step violates the constraint the index would still throw an
exception.

The reason this happens is that native index validates the uniqueness
constraint on every insert and can not defer this validation to later,
unlike Lucene.

We avoid this problem by first applying all removals and then adds. Also
changes are split up into a remove and an add and applied accordingly.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
@@ -58,6 +58,7 @@ import static java.lang.String.format;
 
 import static org.neo4j.collection.primitive.PrimitiveIntCollections.contains;
 import static org.neo4j.kernel.impl.api.index.IndexPopulationFailure.failure;
+import static org.neo4j.kernel.impl.api.index.IndexUpdateProcessor.applyIndexUpdatesByMode;
 
 /**
  * {@link IndexPopulator} that allow population of multiple indexes during one iteration.
@@ -384,7 +385,8 @@ public class MultipleIndexPopulator implements IndexPopulator
                 // Generifying IndexUpdateProcessor with exception type doesn't work due to multiple exception types thrown.
                 try
                 {
-                    IndexUpdateProcessor.applyIndexUpdatesByMode( queue, update -> storeScan.acceptUpdate( updater, update, currentlyIndexedNodeId ) );
+                    applyIndexUpdatesByMode( extractUpdatesFromQueue(),
+                            update -> storeScan.acceptUpdate( updater, update, currentlyIndexedNodeId ) );
                 }
                 catch ( IOException | IndexEntryConflictException e )
                 {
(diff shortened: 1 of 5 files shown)


community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobTest.java
```diff
@@ -492,7 +492,7 @@ public void process( IndexEntryUpdate<?> update ) throws IOException, IndexEntry
                             added.add( Pair.of( update.getEntityId(), update.values()[0].asObjectCopy() ) );
                             break;
                         default:
-                            throw new IllegalArgumentException( update.updateMode().name() );
+                            // ignore removals
```
MishaDemianenko: do not think that's a right way of ignoring removals, instead add an additional case for removals and keep default case that throw exception

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
```diff
@@ -375,13 +377,19 @@ private void populateFromQueueIfAvailable( long currentlyIndexedNodeId )
         {
             try ( MultipleIndexUpdater updater = newPopulatingUpdater( storeView ) )
             {
-                do
+                // TODO the generics of IndexEntryUpdate in combination with that of IndexUpdater make the IndexUpdateProcessor
```
MishaDemianenko, line 381: do not think that removing generics and adding a long comment that something will not happen is a wright thing to do ever, i think generics should be restored and new code needs to be updated instead.

MishaDemianenko (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
tinwelint: Fixing this in 3.3 will be a bigger thing and a lot of this is undergoing a change in 3.4, so changing anything regarding generics here will only cause a lot of other problems. This should be dealt with in 3.4, after the current work happening in 3.4

community/kernel/src/test/java/org/neo4j/kernel/impl/api/index/IndexPopulationJobTest.java
tinwelint: This is in a test and the test asserts what the populator sees, but I can make this change anyway.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
MishaDemianenko: ok lets do that in 3.4 then only later on

MishaDemianenko (approved):

PR #10945, opened 2018-02-01
Close procedure stream on exception

henriknyman:
Make sure we call any close handlers attached to the procedure
stream on exception. Exceptions that occur during close are
suppressed. Also improves the error message.


fickludd (approved):
Very reasonable changes so that `Stream.close` is called regardless of how and when a procedure succeeds or crashes.

fickludd:
@henriknyman this PR needs some attention =).

henriknyman:
Replaced with https://github.com/neo4j/neo4j/pull/11036

PR #10946, opened 2018-02-01
un-version slotted runtime package

fickludd:
Remove `compatibility.v3_4.` from the slotted runtime packaged. There more places to do this, but taking it in chunks.


pontusmelke (approved):
👍

PR #10948, opened 2018-02-02
Adding temporal array types.

sherfert:
Also fix some remainders around temporal values in some other classes.


community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/ParameterConverter.java
```diff
@@ -365,37 +367,37 @@ public void writeDuration( long months, long days, long seconds, int nanos ) thr
...
     {
-        writeValue( LocalTimeValue.localTime( nanoOfDay ) );
+        writeValue( LocalTime.ofNanoOfDay( nanoOfDay ) );
     }
 
     @Override
     public void writeTime( long nanosOfDayUTC, int offsetSeconds ) throws RuntimeException
     {
-        writeValue( TimeValue.time( nanosOfDayUTC, ZoneOffset.ofTotalSeconds( offsetSeconds ) ) );
+        writeValue( OffsetTime.of( LocalTime.ofNanoOfDay( nanosOfDayUTC ), ZoneOffset.ofTotalSeconds( offsetSeconds ) ) );
```
Lojjs, line 382: Why are you using `OffsetTime.of( LocalTime.ofNanoOfDay( nanosOfDayUTC ), ... )` instead of `OffsetTime.ofInstant( Instant.ofEpochSecond( 0, nanosOfDayUTC ), ... )`  which agrees more with both the former `TimeValue` implementation and `writeLocalDateTime below?` 

Also, the same in `BaseToObjectValueWriter`.

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/ParameterConverter.java
```diff
@@ -365,37 +367,37 @@ public void writeDuration( long months, long days, long seconds, int nanos ) thr
...
     {
-        writeValue( DateTimeValue.datetime( epochSecondUTC, nano, ZoneOffset.ofTotalSeconds( offsetSeconds ) ) );
+        writeValue( ZonedDateTime.ofInstant( Instant.ofEpochSecond(epochSecondUTC, nano), ZoneOffset.ofTotalSeconds( offsetSeconds ) ) );
     }
 
     @Override
     public void writeDateTime( long epochSecondUTC, int nano, String zoneId ) throws RuntimeException
     {
-        writeValue( DateTimeValue.datetime( epochSecondUTC, nano, ZoneId.of( zoneId ) ) );
+        writeValue( ZonedDateTime.of( LocalDate.ofEpochDay( epochSecondUTC ), LocalTime.ofNanoOfDay( nano ), ZoneId.of( zoneId ) ) );
```
Lojjs: Why are this method using `ZonedDateTime.of` and the one above`ZonedDateTime.ofInstant`? They are both doing the same thing, just having the time zone on different formats.

Also, the same in `BaseToObjectValueWriter`.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
```diff
@@ -135,14 +135,7 @@ else if ( object instanceof Stream<?> )
...
             {
-                if ( object instanceof Point )
-                {
-                    return asPointValue( (Point) object );
-                }
-                else
-                {
-                    return asGeometryValue( (Geometry) object );
-                }
+                return asGeometryValue( (Geometry) object );
```
Lojjs, line 138: What was the reason to remove the `asPointValue`option here?

community/values/src/test/java/org/neo4j/values/storable/LocalTimeValueTest.java
```diff
@@ -109,6 +111,18 @@ public void writeLocalTime( long nanoOfDay ) throws RuntimeException
...
     }
 
+    @Test
+    public void shouldEqualItself()
+    {
+        assertEqual( localTime( 10, 52, 5, 6 ), localTime( 10, 52, 5, 6 ) );
+    }
+
+    @Test
+    public void shouldNotEqualSameTimeButDifferentTimezone()
```
Lojjs: Copy pasta, no time zones here!  The name of this test should be `shouldNotEqualOther`

community/values/src/main/java/org/neo4j/values/storable/ArrayValue.java
```diff
@@ -66,31 +65,46 @@ public AnyValue next()
...
+        if ( other == null )
+        {
+            return false;
+        }
+
+        if ( other instanceof SequenceValue )
+        {
+            return this.equals( (SequenceValue) other );
+        }
+        return other instanceof Value && equals( (Value) other );
```
Lojjs: Can an `ArrayValue` ever equal something which is not another array and thus an instance of `SequenceValue`? I would assume that this row always returns false.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/executionplan/procs/ProcedureExecutionResult.scala
```diff
@@ -108,6 +110,12 @@ class ProcedureExecutionResult(context: QueryContext,
           case CTBoolean => transform(res(pos), booleanValue)
           case CTPoint => transform(res(pos), (p: Point) => asPointValue(p))
           case CTGeometry => transform(res(pos), (g: Geometry) => asGeometryValue(g))
+          case CTDateTime => transform(res(pos), (g: ZonedDateTime) => DateTimeValue.datetime(g))
+          case CTLocalDateTime => transform(res(pos), (g: LocalDateTime) => LocalDateTimeValue.localDateTime(g))
+          case CTDate => transform(res(pos), (g: LocalDate) => DateValue.date(g))
+          case CTTime => transform(res(pos), (g: OffsetTime) => TimeValue.time(g))
+          case CTLocalTime => transform(res(pos), (g: LocalTime) => LocalTimeValue.localTime(g))
+          case CTDuration => transform(res(pos), (g: TemporalAmount) => Values.durationValue(g))
```
Lojjs, line 118: In many places in the PR you are using the built-in Java classes for everything except `Duration`, where you use our own class `DurationValue` instead, see e.g. `CastSupport.scala`. Here, you are using the Java version `TemporalAmount`.
1) Why are you not using `TemporalAmount` in the other places? Has it to do with `TemporalAmount` being an interface and not a class as e.g. `OffsetTime`?
2) Why are you using it here? That feels a bit inconsistent to me.

community/values/src/main/java/org/neo4j/values/storable/ByteArray.java
```diff
@@ -22,31 +22,36 @@
...
-import org.neo4j.values.SequenceValue;
 import org.neo4j.values.ValueMapper;
 
 import static java.lang.String.format;
 
-public abstract class ByteArray extends IntegralArray
+public class ByteArray extends IntegralArray
 {
-    abstract byte[] value();
+    final byte[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/CharArray.java
```diff
@@ -19,23 +19,33 @@
...
-import org.neo4j.values.SequenceValue;
 import org.neo4j.values.ValueMapper;
 
 import static java.lang.String.format;
 
-public abstract class CharArray extends TextArray
+public class CharArray extends TextArray
 {
-    abstract char[] value();
+    final char[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/DateArray.java
```diff
@@ -0,0 +1,106 @@
...
+import java.time.LocalTime;
+import java.time.OffsetTime;
+import java.time.ZonedDateTime;
+import java.util.Arrays;
+
+import org.neo4j.values.ValueMapper;
+
+public class DateArray extends TemporalArray<LocalDate, DateValue>
+{
+    final LocalDate[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/DateTimeArray.java
```diff
@@ -0,0 +1,106 @@
...
+import java.time.LocalTime;
+import java.time.OffsetTime;
+import java.time.ZonedDateTime;
+import java.util.Arrays;
+
+import org.neo4j.values.ValueMapper;
+
+public class DateTimeArray extends TemporalArray<ZonedDateTime, DateTimeValue>
+{
+    final ZonedDateTime[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/DoubleArray.java
```diff
@@ -22,31 +22,36 @@
...
-import org.neo4j.values.SequenceValue;
 import org.neo4j.values.ValueMapper;
 
 import static java.lang.String.format;
 
-public abstract class DoubleArray extends FloatingPointArray
+public class DoubleArray extends FloatingPointArray
 {
-    abstract double[] value();
+    final double[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/DurationArray.java
```diff
@@ -0,0 +1,187 @@
...
+
+import org.neo4j.graphdb.spatial.Geometry;
+import org.neo4j.values.AnyValue;
+import org.neo4j.values.ValueMapper;
+
+import static java.lang.String.format;
+
+public class DurationArray extends NonPrimitiveArray<DurationValue>
+{
+    final DurationValue[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/FloatArray.java
```diff
@@ -22,31 +22,36 @@
...
-import org.neo4j.values.SequenceValue;
 import org.neo4j.values.ValueMapper;
 
 import static java.lang.String.format;
 
-public abstract class FloatArray extends FloatingPointArray
+public class FloatArray extends FloatingPointArray
 {
-    abstract float[] value();
+    final float[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/IntArray.java
```diff
@@ -22,31 +22,36 @@
...
-import org.neo4j.values.SequenceValue;
 import org.neo4j.values.ValueMapper;
 
 import static java.lang.String.format;
 
-public abstract class IntArray extends IntegralArray
+public class IntArray extends IntegralArray
 {
-    abstract int[] value();
+    final int[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/LongArray.java
```diff
@@ -22,31 +22,36 @@
...
-import org.neo4j.values.SequenceValue;
 import org.neo4j.values.ValueMapper;
 
 import static java.lang.String.format;
 
-public abstract class LongArray extends IntegralArray
+public class LongArray extends IntegralArray
 {
-    abstract long[] value();
+    final long[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/PointArray.java
```diff
@@ -19,23 +19,31 @@
...
-import org.neo4j.values.SequenceValue;
 import org.neo4j.values.ValueMapper;
 
-import static java.lang.String.format;
-
-public abstract class PointArray extends ArrayValue
+public class PointArray extends NonPrimitiveArray<PointValue>
 {
-    abstract PointValue[] value();
+    final PointValue[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/NonPrimitiveArray.java
```diff
@@ -0,0 +1,147 @@
...
+        return false;
+    }
+
+    @Override
+    public NumberType numberType()
+    {
+        return NumberType.NO_NUMBER;
+    }
+
+    public final int compareTo( NonPrimitiveArray<T> other )
```
Lojjs, line 88: Is it possible to call `compareTo` on one type of NonPrimitiveArray with another type as input parameter, e.g. doing `compareTo` between a PointArray and a DateArray? If that is possible, I would like to see some testing showing that `compareTo` work as expected in such a case.

community/values/src/main/java/org/neo4j/values/storable/ShortArray.java
```diff
@@ -22,31 +22,36 @@
...
-import org.neo4j.values.SequenceValue;
 import org.neo4j.values.ValueMapper;
 
 import static java.lang.String.format;
 
-public abstract class ShortArray extends IntegralArray
+public class ShortArray extends IntegralArray
 {
-    abstract short[] value();
+    final short[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/StringArray.java
```diff
@@ -19,47 +19,57 @@
...
-import org.neo4j.values.SequenceValue;
 import org.neo4j.values.ValueMapper;
 
 import static java.lang.String.format;
 
-public abstract class StringArray extends TextArray
+public class StringArray extends TextArray
 {
-    abstract String[] value();
+    final String[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/DurationArray.java
```diff
@@ -0,0 +1,83 @@
...
+package org.neo4j.values.storable;
+
+import java.util.Arrays;
+
+import org.neo4j.values.AnyValue;
+import org.neo4j.values.ValueMapper;
+
+public class DurationArray extends NonPrimitiveArray<DurationValue>
+{
+    final DurationValue[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/LocalDateTimeArray.java
```diff
@@ -0,0 +1,72 @@
...
+package org.neo4j.values.storable;
+
+import java.time.LocalDateTime;
+import java.util.Arrays;
+
+import org.neo4j.values.ValueMapper;
+
+public class LocalDateTimeArray extends TemporalArray<LocalDateTime, LocalDateTimeValue>
+{
+    final LocalDateTime[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/LocalTimeArray.java
```diff
@@ -0,0 +1,72 @@
...
+package org.neo4j.values.storable;
+
+import java.time.LocalTime;
+import java.util.Arrays;
+
+import org.neo4j.values.ValueMapper;
+
+public class LocalTimeArray extends TemporalArray<LocalTime, LocalTimeValue>
+{
+    final LocalTime[] value;
```
Lojjs: Should be private

community/values/src/main/java/org/neo4j/values/storable/TimeArray.java
```diff
@@ -0,0 +1,72 @@
...
+package org.neo4j.values.storable;
+
+import java.time.OffsetTime;
+import java.util.Arrays;
+
+import org.neo4j.values.ValueMapper;
+
+public class TimeArray extends TemporalArray<OffsetTime, TimeValue>
+{
+    final OffsetTime[] value;
```
Lojjs: Should be private

Lojjs (changes requested):
First of all, really good work! I especially liked your simplification of Values.equals were we now have got rid of a lot of methods only returning false. Many of my comments are simply stating `should be private` The values field in the array classes can always be private, so I have added a comment on each occurrence where it’s not private, not only in the new classes. I’m not sure if the new array constructors should have some other visibility as well or if package-private is fine.

community/values/src/main/java/org/neo4j/values/storable/NonPrimitiveArray.java
Lojjs: Actually I realise now, that we already has this kind of testing in `ValueComparisonTest` so apparently it work as expected

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/ParameterConverter.java
sherfert: An `OffsetTime` is internally comprised of a `LocalTime` and a `ZoneId`. Therefore, this constructor comes more naturally. For `LocalDateTime` we cannot have the more natural choice (`LocalDate` and `LocalTime`), since `epochSecond` is part of both Date and Time. This splitting was chosen since it makes most sense for the PropertyStore (I think).

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/ParameterConverter.java
sherfert: This is a bug. They should both use `ofInstant`.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/executionplan/procs/ProcedureExecutionResult.scala
sherfert: Most places should use `TemporalAmount`, since values coming from "the outside" can be build in other ways and might not be `DurationValues`. We can safely use `DurationValue` only if we know that no values from the outside can get there.

I changed it at least in `CastSupport`. What other places did you have in mind?

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
sherfert: `asGeometryValue` just checks if it is a point and then calls `toPoint`. We don't need both of them there.

community/values/src/main/java/org/neo4j/values/storable/ArrayValue.java
sherfert: Sounds right.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/compatibility/v3_4/runtime/executionplan/procs/ProcedureExecutionResult.scala
Lojjs: Now I can only find ParameterConverterTest, line 314 apart from the one that you fixed. Maybe I was looking at some cases that was changed in later commits. Because it is in a test you don't have to change it

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
Lojjs: Ok

Lojjs (approved):

PR #10949, opened 2018-02-02
Adding support for scalar temporal values in the PropertyStore

sherfert:
This does only support scalar values, except DateTimes with
ZoneIds. It does not add array support yet.


community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyStore.java
```diff
@@ -124,6 +124,22 @@
...
+ * LOCALTIME: [xxxx,xxxx] [xxxx,xxxx] [xxxx,xxxx] [xxxx,xxx1] [ 02 ,type][K][K][K] nanoOfDay
+ * LOCALTIME: [    ,    ] [    ,    ] [    ,    ] [    ,   0] [ 02 ,type][K][K][K] nanoOfDay in next long block
+ * LOCALDTIME:[xxxx,xxxx] [xxxx,xxxx] [xxxx,xxxx] [xxxx,xxxx] [ 03 ,type][K][K][K] nanoOfSecond
+ *            epochSecond in next long block
+ * TIME:      [xxxx,xxxx] [xxxx,xxxx] [xxxx,xxxx] [xxxx,xxxx] [ 04 ,type][K][K][K] secondOffset (=ZoneOffset)
+ *            nanoOfDay in next long block
+ * DATETIME:  [xxxx,xxxx] [xxxx,xxxx] [xxxx,xxxx] [xxxx,xxx1] [ 05 ,type][K][K][K] nanoOfSecond
+ *            epochSecond in next long block
+ *            secondOffset in next long block
+ * TODO alternative DateTime format
```
tinwelint, line 138: Should there be an alternative DateTime format?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TemporalType.java
```diff
@@ -0,0 +1,452 @@
...
+                    if ( storingZoneOffset( valueBlocks[0] ) )
+                    {
+                        int nanoOfSecond = (int) (valueBlocks[offset] >>> 33);
+                        long epochSecond = valueBlocks[1 + offset];
+                        int secondOffset = (int) valueBlocks[2 + offset];
+                        return DateTimeValue.datetime( epochSecond, nanoOfSecond, ZoneOffset.ofTotalSeconds( secondOffset ) );
+                    }
+                    else
+                    {
+                        throw new UnsupportedOperationException( "..." );
```
tinwelint, line 176: Likely to happen? Should there be a better description of this error?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TemporalType.java
```diff
@@ -0,0 +1,452 @@
...
+                public int calculateNumberOfBlocksUsedForTemporal( long firstBlock )
+                {
+                    if ( storingZoneOffset( firstBlock ) )
+                    {
+                        return 3;
+                    }
+                    else
+                    {
+                        // TODO proper number
+                        return 3;
```
tinwelint, line 190: Then what is a proper number?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TemporalType.java
```diff
@@ -0,0 +1,452 @@
...
+
+    public static int calculateNumberOfBlocksUsed( long firstBlock )
+    {
+        TemporalType geometryType = find( getTemporalType( firstBlock ) );
+        return geometryType.calculateNumberOfBlocksUsedForTemporal( firstBlock );
+    }
+
+    private static TemporalType find( int temporalType )
+    {
+        if ( temporalType < TYPES.length )
```
tinwelint: `&& temporalType >= 0` ?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/format/Capability.java
```diff
@@ -57,6 +57,11 @@
      */
     POINT_PROPERTIES( CapabilityType.STORE ),
 
+    /**
+     * Temporal types are an addition to the format, not a change
+     */
+    TEMPORAL_PROPERTIES( CapabilityType.STORE ),
```
tinwelint, line 63: Since this is used in combination with `POINT_PROPERTIES` in the property store, perhaps it would be enough with one capability for the both, named something like `POINT_AND_TEMPORAL_PROPERTIES`?

tinwelint (changes requested):
Had a few comments/questions, but generally looks reasonably straight forward to me.

Lots of TODOs, but mostly about array support, which will arrive soon presumably?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyStore.java
sherfert: Yes, DateTimes can have either ZoneIds or ZoneOffsets. This PR only supports DateTimes with ZoneOffsets. After having the TimeZoneTokenStore, we will also support (and document) the other format.

Array storage will also arrive soon in a separate PR.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TemporalType.java
sherfert: Sorry, this is when we encounter a DateTime with ZoneId (ergo should never happen right now) and the else block will then be implemented as soon as we support that format.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TemporalType.java
sherfert: We don't know yet. This will also be figured out when we implement the format for DateTimes with ZoneIds.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/format/Capability.java
sherfert: I found it nice to have them split, since then the error message will exactly tell you what Capability the store is missing and might be more helpful if it distinguishes between the two.

tinwelint: OK if there is a use case for having them split, then it's alright. I just thought there wasn't.

tinwelint (approved):
Apart from record test failures then 👍

PR #10950, opened 2018-02-02
Autoclean of sonar errors

klaren:


enterprise/auth-plugin-api/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/spi/AuthenticationPlugin.java
```diff
@@ -62,7 +62,7 @@
      * @see CustomCacheableAuthenticationInfo
      * @see AuthProviderOperations#setAuthenticationCachingEnabled(boolean)
      */
-    AuthenticationInfo authenticate( AuthToken authToken ) throws AuthenticationException;
+    AuthenticationInfo authenticate( AuthToken authToken );
```
MishaDemianenko: hm do not think we can remove this one? it looks like its some sort of public interface for auth plugins, also javadoc still claims that exception can be throw

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/Neo4jError.java
```diff
@@ -174,7 +174,7 @@ private static Neo4jError fromThrowable( Throwable any, boolean isFatal )
         // In this case, an error has "slipped out", and we don't have a good way to handle it. This indicates
         // a buggy code path, and we need to try to convince whoever ends up here to tell us about it.
 
-        return new Neo4jError( Status.General.UnknownError, any.getMessage(), any, isFatal );
+        return new Neo4jError( Status.General.UnknownError, any != null ? any.getMessage() : null, any, isFatal );
```
MishaDemianenko, line 177: not sure how this came? is it also safe to send null as error message here?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
```diff
@@ -160,7 +159,7 @@
     private Type type;
     private long transactionId;
     private long commitTime;
-    private volatile int reuseCount;
+    private volatile int reuseCount; // TODO: why volatile?
```
MishaDemianenko: todo

MishaDemianenko (commented):
I went over most of the man source files, have couple of questions, please fix those and then i think you can go and merge this massive cleanup as soon as builds are green

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/Neo4jError.java
klaren: It's a potential `NullPointerException` since `any` can be `null`, the message is checked for null every time it's used so I believe it's okay.

enterprise/auth-plugin-api/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/spi/AuthenticationPlugin.java
klaren: That looks true, maybe I should add it to revapi to protect it in the future?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/KernelTransactionImplementation.java
klaren: Removed TODO, sonar flagged it as a non-atomic operation on a volatile variable. After closer inspections, it is accessed from multiple threads, but it's only one thread that can update it at a time so it should fine.

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/Neo4jError.java
MishaDemianenko: Should it be impossible to come here without exception right?

enterprise/auth-plugin-api/src/main/java/org/neo4j/server/security/enterprise/auth/plugin/spi/AuthenticationPlugin.java
MishaDemianenko: please talk with security guys to figured out if we need to do so

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/Neo4jError.java
klaren: You could still call `Neo4jError.from()` with `null`, impossible is a very strong word :stuck_out_tongue: I don't think this hurts.

klaren:
New commits @MishaDemianenko, turns out that the auto-fix was not transient so I hade to fix point iterate it until it found 0 redundant throws, that's why there are a couple more.

MishaDemianenko (approved):

PR #10952, opened 2018-02-02
Reuse Kernel API cursors in InstanceCache

fickludd:
This is done to reduce GC pressure during heavy Core API loads, where we would allocate
new cursors for every iterator. This is extremely expensive for eg NodeProxy.singleRelationship(),
so now we are instead reusing the object if no-one else has handed back their cursor in between.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultNodeCursor.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultNodeCursor.java
@@ -20,6 +20,7 @@
 package org.neo4j.kernel.impl.newapi;
 
 import java.util.Set;
+import java.util.function.Consumer;
 
 import org.neo4j.collection.primitive.Primitive;
 import org.neo4j.collection.primitive.PrimitiveIntSet;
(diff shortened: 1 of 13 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultNodeCursor.java
```diff
@@ -242,21 +246,27 @@ public boolean shouldRetry()
...
-        hasChanges = HasChanges.MAYBE;
-        addedNodes = emptySet();
-        reset();
+
+        if ( pageCursor != null )
+        {
+            pageCursor.close();
+            pageCursor = null;
+
+            if ( pool != null )
```
tinwelint: Why are we doing this null check? Are we expecting this pool to be null in some real scenarios? If this check is here to allow tests to pass in null then I think that they should pass in a no-op consumer instead so that we can get rid of this condition. (many similar places below too).

tinwelint (changes requested):

fickludd:
This is not the solution we are looking for, will return with this functionality and more in a separate PR. This crashes because there is currently only one `DefaultCursors` in the db, meaning concurrent transactions will trash the instance caches in interesting ways. Didn't realize that before. Will change to one `DefaultCursors` per tx.

PR #10953, opened 2018-02-02
Improve CypherComparisonSupport

Lojjs:
Includes:
- Create 2 new handy short TestConfigs
- Simplify a lot of configuration code

Note:
`resultAssertionInTx = None` was required to make the tests work, but maybe the base scenario should be able to use the same resultAssertionInTx function as the other scenarios does? Or is it running in a transaction anyway?

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
@@ -145,44 +145,59 @@ trait CypherComparisonSupport extends CypherTestSupport {
                             params: Map[String, Any] = Map.empty): InternalExecutionResult = {
     // Never consider Morsel even if test requests it
     val expectSucceedEffective = expectSucceed - Configs.Morsel
-    val expectedDifferentResultsEffective = expectedDifferentResults - Configs.Morsel
 
-    val compareResults = expectSucceedEffective - expectedDifferentResultsEffective
-    val baseScenario =
-      if (expectSucceedEffective.scenarios.nonEmpty) extractBaseScenario(expectSucceedEffective, compareResults)
-      else TestScenario(Versions.Default, Planners.Default, Runtimes.Interpreted)
+    if (expectSucceedEffective.scenarios.nonEmpty) {
+      val expectedDifferentResultsEffective = expectedDifferentResults - Configs.Morsel
+      val compareResults = expectSucceedEffective - expectedDifferentResultsEffective
+      val baseScenario = extractBaseScenario(expectSucceedEffective, compareResults)
+      val explicitlyRequestedExperimentalScenarios = expectSucceedEffective.scenarios intersect Configs.Experimental.scenarios
 
-    val explicitlyRequestedExperimentalScenarios = expectSucceedEffective.scenarios intersect Configs.Experimental.scenarios
-    val positiveResults = ((Configs.AbsolutelyAll.scenarios ++ explicitlyRequestedExperimentalScenarios) - baseScenario).flatMap {
-      thisScenario =>
-        executeScenario(thisScenario, query, expectSucceedEffective.containsScenario(thisScenario), executeBefore, params, resultAssertionInTx)
+      val positiveResults = ((Configs.AbsolutelyAll.scenarios ++ explicitlyRequestedExperimentalScenarios) - baseScenario).flatMap {
+        thisScenario =>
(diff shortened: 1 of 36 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -632,23 +649,23 @@ object CypherComparisonSupport {
...
     If you are unsure what you need, this is a good start. It's not really all scenarios, but this is testing all
     interesting scenarios.
      */
-    def All: TestConfiguration =
-      TestConfiguration(Versions.V3_4, Planners.Cost, Runtimes(Runtimes.CompiledSource, Runtimes.CompiledBytecode)) +
-        TestConfiguration(Versions.Default, Planners.Default, Runtimes(Runtimes.Interpreted, Runtimes.Slotted)) +
-        TestConfiguration(Versions.V2_3 -> Versions.V3_1, Planners.all, Runtimes.Default) +
-        TestScenario(Versions.Default, Planners.Rule, Runtimes.Default) +
-        TestScenario(Versions.V3_3, Planners.Cost, Runtimes.Default)
+    def All: TestConfiguration = AbsolutelyAll - Procs
```
sherfert, line 666: When Alex and I made all configurations explicit (instead of using each other with + and -), we tried to make it easier to understand, what a Configuration actually contains. At least for all the "pre-canned" Configurations offered here.

I think that what we came up with is too verbose and, even though it makes explicit what is used, is not that helpful. I am really unsure which approach is more convenient/error-prone.

Just sharing thoughts here :)

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
```diff
@@ -145,44 +145,59 @@ trait CypherComparisonSupport extends CypherTestSupport {
...
+          executeScenario(thisScenario, query, expectSucceedEffective.containsScenario(thisScenario), executeBefore, params, resultAssertionInTx)
+      }
 
-    baseScenario.prepare()
-    executeBefore()
-    val baseResult = innerExecute(s"CYPHER ${baseScenario.preparserOptions} $query", params)
-    baseScenario.checkResultForSuccess(query, baseResult)
-    planComparisonStrategy.compare(expectSucceedEffective, baseScenario, baseResult)
+      //Must be run last and have no rollback to be able to do certain result assertions
+      val baseOption = executeScenario(baseScenario, query, expectedToSucceed = true, executeBefore, params, resultAssertionInTx = None, rollback = false)
```
sherfert, line 161: I am confused. Passing in `None` instead of `resultAssertionInTx` means, we are _not_ running inTx assertions for the baseScenario, doesn't it?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
Lojjs: I see what you mean. I'm also not sure which the best way is. My thinking here was that if you want to add a new scenario to `AbsolutelyAll`, you don't have to add it to `All` as well. And while `All` and `AbsolutelyAll` are not the best names for something which does not contain all configs, this at least shows what the difference between them are, which was not very obvious before

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/CypherComparisonSupport.scala
Lojjs: Yes, I'm also confused about this part. 😆  My main intention with this work was not actually to solve the transaction thing, but to simplify the test configs. I then found some improvements that could be made in CypherComparisonSupport and me and Olivia thought I had fixed this card as a bonus. When I looked back at the code however I think you are right and that I actually have failed to do so. So I can either retry to solve it or you can regard this PR as a separete do-what-you-want work and not mark the transaction card as done.

sherfert: I think the code improvements are good to have, regardless of solving the inTx assertions. I'll put that back into prioritized TODO.

sherfert (approved):

PR #10954, opened 2018-02-02
Adds a new token store for time zones

sherfert:
This store will be used as a lookup table for DateTimes with ZoneIDs.
Since there only a couple 100 time zones, it is more space efficient to
store them separately in the new store as strings and only refer to them
using the id.

This touches many parts of kernel, ha, cc; and I don't know if all these propagations are actually necessary. Please check that carefully. I might also miss lots of tests and am thankful for hints where to add some.

I've stumbled upon a lot of unused or strange code. I put TODOs there so the reviewers can verify what should be done.


community/kernel/src/main/java/org/neo4j/kernel/impl/store/StoreAccess.java
```diff
@@ -186,6 +201,7 @@ public CountsAccessor getCounts()
         return counts;
     }
 
+    // TODO never used???
```
tinwelint, line 204: Perhaps a good opportunity to remove it then :)

community/kernel/src/main/java/org/neo4j/kernel/impl/store/format/standard/StandardFormatSettings.java
```diff
@@ -30,6 +30,8 @@
     static final int PROPERTY_MAXIMUM_ID_BITS = 36;
     public static final int DYNAMIC_MAXIMUM_ID_BITS = 36;
     public static final int LABEL_TOKEN_MAXIMUM_ID_BITS = 32;
+    // TODO decide on correct max bits
```
tinwelint: What does the store format of temporal values in #10949 suggest/assume?

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/command/PhysicalLogCommandReaderV3_0_2.java
```diff
@@ -82,6 +83,9 @@ protected Command read( byte commandType, ReadableChannel channel ) throws IOExc
             return visitRelationshipTypeTokenCommand( channel );
         case NeoCommandType.LABEL_KEY_COMMAND:
             return visitLabelTokenCommand( channel );
+            // TODO do older physical logs need to support time zone tokens?
```
tinwelint: No and this addition should bump the tx log format, see `org.neo4j.kernel.impl.storageengine.impl.recordstorage.RecordStorageCommandReaderFactory`. Basically don't add this time zone reading capability here, instead copy this class and add it to the new version instead. See outline of how to do that in `org.neo4j.kernel.impl.transaction.log.entry.LogEntryVersion` class javadoc, which I hope is up2date.

tinwelint (changes requested):
Tokens are everywhere, across the whole db stack. I can only imagine the dread of adding such a store. An applaud-able effort and looks well made. 👍 

Had basically only one concern, which is that of the log reader and hopefully I pointed you in the right direction for how we usually go about such changes.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/format/standard/StandardFormatSettings.java
sherfert: Nothing yet. There are ~500 different time zones, which we need to be able to address (plus a buffer for future time zones). I will adapt the number accordingly.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/command/PhysicalLogCommandReaderV3_0_2.java
sherfert: Oki doki.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/StoreAccess.java
phughk: I saw a few other `TODO` in the PR as well in case it was missed.

phughk (approved):
Seems good, though I think @martinfurmanski should take a look when he is available

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/store/format/highlimit/v310/HighLimitV3_1_0.java
```diff
@@ -92,6 +94,12 @@ public HighLimitV3_1_0()
         return new LabelTokenRecordFormat();
     }
 
+    @Override
+    public RecordFormat<TimeZoneTokenRecord> timeZoneToken()
```
martinfurmanski, line 98: Should we really touch store formats < 3.4? Perhaps introduce a new one?

enterprise/ha/src/main/java/org/neo4j/kernel/ha/MasterClient214.java
```diff
@@ -205,6 +205,14 @@ protected boolean shouldCheckStoreId( RequestType<Master> type )
         return sendRequest( requestTypes.type( HaRequestTypes.Type.CREATE_LABEL ), context, serializer, deserializer );
     }
 
+    @Override
+    public Response<Integer> createTimeZone( RequestContext context, String name )
```
martinfurmanski, line 209: I have forgotten how most of this stuff works, but updating MasterClient214 which is the oldest one seems incorrect. I don't know at the moment if there is a need for a new version because of this. Should be investigated.

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/core/state/machines/CoreStateMachinesTest.java
```diff
@@ -168,7 +170,7 @@ public void shouldReturnLastAppliedOfAllStateMachines() throws Exception
     private final DummyMachine dummySM = mock( DummyMachine.class );
     private final RecoverConsensusLogIndex recoverConsensusLogIndex = mock( RecoverConsensusLogIndex.class );
 
-    private final CoreStateMachines coreStateMachines = new CoreStateMachines( txSM, labelTokenSM,
+    private final CoreStateMachines coreStateMachines = new CoreStateMachines( txSM, labelTokenSM, timeZoneTokenSM,
```
martinfurmanski, line 173: Makes me wonder about upgrade paths from < 3.4.

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/state/SchemaRuleCommandTest.java
```diff
@@ -90,7 +90,7 @@
     private final IndexBatchTransactionApplier indexApplier = new IndexBatchTransactionApplier( indexes,
             labelScanStoreSynchronizer, indexUpdatesSync, mock( NodeStore.class ),
             new PropertyPhysicalToLogicalConverter( propertyStore ) );
-    private final BaseCommandReader reader = new PhysicalLogCommandReaderV3_0_2();
+    private final BaseCommandReader reader = new PhysicalLogCommandReaderV3_0_11();
```
martinfurmanski, line 93: Far from obviously correct.

martinfurmanski (commented):
This is a very fundamental change in the product. Have to make sure to exercise this new store in stress tests (e.g. xx-robustness), long-running soaks should have workloads exercising these parts and upgrade tests should test upgrade paths. There are probably a bunch of ITs as well that should exercise this new store, especially upgrade-testing or recovery testing ITs.

enterprise/ha/src/main/java/org/neo4j/kernel/ha/MasterClient214.java
tinwelint: Right, good catch. I definitely think we need to introduce a new version actually.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/store/format/highlimit/v310/HighLimitV3_1_0.java
tinwelint: Thoughts:

Introducing a new store is a bit different than changing format of a store. I wonder what it should do... throw exception instead? If we make it throw (together with 3_2_0 of course) then we'll have to make sure not to call it when opening older stores. To do that we'll have to look at the format Capabilities and exclude timeZone store from StoreType[] list passed to `StoreFactory#openAllNeoStores` / `#openNeoStores`, which is perhaps something that StoreFactory could do itself.

Otoh creating that store doesn't damage the old store, but it's slightly weird to create it though.

community/kernel/src/test/java/org/neo4j/kernel/impl/transaction/state/SchemaRuleCommandTest.java
sherfert: Isn't this file always testing with the latest physical log version? It seemed to me that way, and that is why I replaced that line.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/store/format/highlimit/v310/HighLimitV3_1_0.java
sherfert: I agree it is weird, but not harmful. Do you remember how that was handled the last time a new store was added?

tinwelint: Last time a store was added we didn't have multiple formats to consider

sherfert:
Given the difficulties and the little time before feature freeze, we decided to take a different approach for storing TimeZones.

PR #10955, opened 2018-02-02
Reduced overhead of dynamic-size tree node format

burqen:
The dynamic tree node has an offset array, which is 2 bytes per entry. It also had 2 bytes key size and 2 bytes value size. These key/value sizes is what this PR reduces, from 4 down to a minimum of 1 byte.

co-author: @tinwelint


community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
```diff
@@ -28,28 +28,73 @@
...
+ * - Key tombstone, first bit in keyValueSize
+ *
+ * Format of key/value size is dynamic in itself, first byte being:
+ * <pre>
+ * [T,K,V,k,k,k,k,k]
+ * </pre>
+ * If {@code T} is set key is dead.
+ * If {@code K} is set the next byte contains the higher order bits of the key size.
+ * If {@code V} is set there is a value size to be read directly after key size.
+ * This first byte can fit key size <= 32 and we only need the second byte if key size is larger.
```
tinwelint: I believe it should be `key size < 32`, since there's only 5 bits to store it on in the first byte.

community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
```diff
@@ -28,28 +28,73 @@
...
+ * Together with the second byte we can fit key size <= 8192.
+ *
+ * Byte following key size bytes (second or third byte depending on how many bytes needed for key size):
+ * <pre>
+ * [V,v,v,v,v,v,v,v]
+ * </pre>
+ * If {@code V} is set the next byte contains the higher order bits of the value size.
+ * This first value size byte can fit value size <= 128 and with the second byte we can fit value size <= 32768.
```
tinwelint: likewise `value size < 128`, and `value size < 32768`

community/index/src/main/java/org/neo4j/index/internal/gbptree/TreeNodeDynamicSize.java
```diff
@@ -125,17 +126,15 @@ KEY keyAt( PageCursor cursor, KEY into, int pos, Type type )
...
             cursor.setCursorException( format( "Read unreliable key, keySize=%d, keyValueSizeCap=%d, keyHasTombstone=%b, offset=%d, pos=%d",
-                    keySize, keyValueSizeCap, hasTombstone( keySize ), offset, pos ) );
+                    keySize, keyValueSizeCap, extractTombstone( keyValueSize ), offset, pos ) );
             return into;
         }
-        if ( type == LEAF )
-        {
-            progressCursor( cursor, bytesValueSize() );
-        }
+        // TODO don't rely on readKeySize to progress beyond value size
```
tinwelint: Should we do something about this TODO in this PR?

community/index/src/test/java/org/neo4j/index/internal/gbptree/SimpleByteArrayLayout.java
```diff
@@ -82,8 +84,16 @@ public void writeValue( PageCursor cursor, RawBytes rawBytes )
...
     {
-        into.bytes = new byte[keySize];
-        cursor.getBytes( into.bytes );
+        int offsetBeforeRead = cursor.getOffset();
+        try
+        {
+            into.bytes = new byte[keySize];
+            cursor.getBytes( into.bytes );
+        }
+        catch ( BufferUnderflowException e )
```
tinwelint: Oh, I don't think this debug-catch should be here, should it?

community/index/src/test/java/org/neo4j/index/internal/gbptree/TreeNodeTestBase.java
```diff
@@ -518,6 +519,7 @@ public void shouldDefragLeafWithMultipleTombstonesInterleavedEven() throws Excep
         assertKeyEquals( key( 4 ), node.keyAt( cursor, layout.newKey(), 1, LEAF ) );
     }
 
+    @Seed( 1517581480887L )
```
tinwelint: This test doesn't look that random all of a sudden :)

tinwelint (changes requested):
Awesome! I made a note about some debug stuff we left behind and some javadoc comments too.

tinwelint (approved):

PR #10956, opened 2018-02-02
Cluster seeding

martinfurmanski:
changelog: Fix cluster seeding failing in certain scenarios


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClient.java
```diff
@@ -77,7 +71,7 @@ public boolean onFileContent( CompletableFuture<Long> signal, FileChunk fileChun
                         public void onFileStreamingComplete( CompletableFuture<Long> signal,
                                 StoreCopyFinishedResponse response )
                         {
-                            log.info( "Finished streaming %s", destination );
+                            log.info( "Finished streaming %s" );
```
RagnarW: this will not log destination

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/RemoteStore.java
```diff
@@ -123,20 +124,40 @@ private long getPullIndex( File storeDir ) throws IOException
...
+     *
+     * If there are no transaction logs then we can pull from and including
+     * the index which the metadata store points to. This would be the case
+     * for example with a backup taken during an idle period of the system.
+     *
+     * However, if there are transaction logs then we want to find out where
+     * they end and pull from there, excluding the last one so that we do not
+     * get duplicate entries.
+     */
+    public CatchupResult tryCatchingUp( AdvertisedSocketAddress from, StoreId expectedStoreId ) throws StoreCopyFailedException, IOException
```
RagnarW: I think we are still of by one. Assume that i = last index, T=tx, M=metadata.

In the scenario were tx logs are present we would before see:
    CLIENT -> Ti -> SERVER
    SERVER -> Ti+1 -> CLIENT

Now we have:
  CLIENT -> Ti-i -> SERVER
  SERVER -> Ti -> CLIENT.

In summary what we get from server now is one of:
Ti
Mi or Mi+1

What we want is:
Ti+1
Mi or Mi+1 

isn't it?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/StoreCopyClient.java
martinfurmanski: Sloppy removal, will remove %s as well. The destination was removed on purpose because that is already logged when receiving and is not interesting to log again on completion.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/RemoteStore.java
RagnarW: Disregard this comment. I was wrong.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
```diff
@@ -252,4 +253,15 @@ private void dropAvailabilityGuard()
         availabilityGuard.fulfill( currentRequirement );
         currentRequirement = null;
     }
+
+    public boolean hasTxLogs() throws RuntimeException
```
andrewkerr9000: Does it make sense to declare `throws RuntimeException`?

andrewkerr9000 (approved):
One comment, but I'm not particularly familiar with this area.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/storecopy/LocalDatabase.java
martinfurmanski: Kind of pointless indeed.

PR #10957, opened 2018-02-02
Active page cache warmup

chrisvest:
[cl] Introduces a new _active page cache warmup_ feature in the Enterprise Edition. The memory contents of the Neo4j page cache are periodically profiled, and these profiles will then be used to quickly reheat the cache on startup. This decreases the time-to-performance for restarts or crashes.

### DIFF
--- a/community/common/src/main/java/org/neo4j/scheduler/JobScheduler.java
+++ b/community/common/src/main/java/org/neo4j/scheduler/JobScheduler.java
@@ -202,6 +202,8 @@ public interface JobScheduler extends Lifecycle
          */
         public static Group vmPauseMonitor = new Group( "VmPauseMonitor" );
 
+        public static Group pageCacheIOHelper = new Group( "PageCacheIOHelper" );
+
         private Groups()
         {
         }
@@ -235,7 +237,9 @@ public interface JobScheduler extends Lifecycle
     /** Expose a group scheduler as an {@link Executor} */
     Executor executor( Group group );
 
-    /** Creates an {@link ExecutorService} that does works-stealing - read more about this in {@link ForkJoinPool}*/
+    /**
+     * Creates an {@link ExecutorService} that does works-stealing - read more about this in {@link ForkJoinPool}
+     */
     ExecutorService workStealingExecutor( Group group, int parallelism );
 
     /**
--- a/community/common/src/main/java/org/neo4j/scheduler/JobScheduler.java
+++ b/community/common/src/main/java/org/neo4j/scheduler/JobScheduler.java
@@ -160,49 +160,52 @@ public interface JobScheduler extends Lifecycle
         /**
          * UDC timed events.
          */
-        public static Group udc  = new Group( "UsageDataCollection" );
+        public static final Group udc  = new Group( "UsageDataCollection" );
 
         /**
          * Storage maintenance.
          */
-        public static Group storageMaintenance = new Group( "StorageMaintenance" );
+        public static final Group storageMaintenance = new Group( "StorageMaintenance" );
 
         /**
          * Raft timers.
          */
-        public static Group raft = new Group( "RaftTimer" );
+        public static final Group raft = new Group( "RaftTimer" );
 
         /**
          * Native security.
          */
-        public static Group nativeSecurity = new Group( "NativeSecurity" );
+        public static final Group nativeSecurity = new Group( "NativeSecurity" );
 
         /**
          * File watch service group
          */
-        public static Group fileWatch = new Group( "FileWatcher" );
+        public static final Group fileWatch = new Group( "FileWatcher" );
 
         /**
          * Recovery cleanup.
          */
-        public static Group recoveryCleanup = new Group( "RecoveryCleanup" );
+        public static final Group recoveryCleanup = new Group( "RecoveryCleanup" );
 
         /**
          * Kernel transaction timeout monitor.
          */
-        public static Group transactionTimeoutMonitor = new Group( "TransactionTimeoutMonitor" );
+        public static final Group transactionTimeoutMonitor = new Group( "TransactionTimeoutMonitor" );
 
         /**
          * Kernel transaction timeout monitor.
          */
-        public static Group cypherWorker = new Group( "CypherWorker" );
+        public static final Group cypherWorker = new Group( "CypherWorker" );
 
         /**
          * VM pause monitor
          */
-        public static Group vmPauseMonitor = new Group( "VmPauseMonitor" );
+        public static final Group vmPauseMonitor = new Group( "VmPauseMonitor" );
 
-        public static Group pageCacheIOHelper = new Group( "PageCacheIOHelper" );
+        /**
+         * IO helper threads for page cache and IO related stuff.
+         */
+        public static final Group pageCacheIOHelper = new Group( "PageCacheIOHelper" );
 
         private Groups()
         {
--- a/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
+++ b/community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
@@ -570,7 +570,7 @@ public class GraphDatabaseSettings implements LoadableConfig
     @Internal
     @Description( "The profiling frequency for the page cache. Accurate profiles allow the page cache to do active " +
                   "warmup after a restart, reducing the mean time to performance." )
-    public static Setting<Duration> pagecache_warmup_profiling_interval =
+    public static final Setting<Duration> pagecache_warmup_profiling_interval =
             setting( "unsupported.dbms.memory.pagecache.warmup.profile.interval", DURATION, "1m" );
 
     /**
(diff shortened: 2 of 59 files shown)


community/common/src/main/java/org/neo4j/scheduler/JobScheduler.java
```diff
@@ -197,6 +197,8 @@ public String threadName( Map<String, String> metadata )
          */
         public static Group cypherWorker = new Group( "CypherWorker" );
 
+        public static Group pageCacheIOHelper = new Group( "PageCacheIOHelper" );
```
klaren: `final`

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -568,6 +568,12 @@
     public static final Setting<String> pagecache_swapper =
             setting( "dbms.memory.pagecache.swapper", STRING, null );
 
+    @Internal
+    @Description( "The profiling frequency for the page cache. Accurate profiles allow the page cache to do active " +
+                  "warmup after a restart, reducing the mean time to performance." )
+    public static Setting<Duration> pagecache_warmup_profiling_interval =
```
klaren: `final`

community/kernel/src/main/java/org/neo4j/helpers/Format.java
```diff
@@ -159,6 +159,13 @@ public static String duration( long durationMillis, TimeUnit highestGranularity,
             }
         }
 
+        if ( builder.length() == 0 )
```
klaren, line 162: Nice catch!

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/PageCacheWarmer.java
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    private long profile( PagedFile file ) throws IOException
+    {
+        long pagesInMemory = 0;
+        File outputNext = profileOutputFileNext( file );
+
+        try ( OutputStream outputStream = compressedOutputStream( outputNext );
+              PageCursor cursor = file.io( 0, PF_SHARED_READ_LOCK | PF_NO_FAULT ) )
+        {
```
klaren, line 237: While correct, all this shifting makes me dizzy, might I suggest:
```
        {
            int stepper = 0;
            int b = 0;
            for ( ; ; )
            {
                if ( stopped )
                {
                    return pagesInMemory;
                }
                if ( !cursor.next() )
                {
                    break; // Exit the loop if there are no more pages.
                }
                if ( cursor.getCurrentPageId() != PageCursor.UNBOUND_PAGE_ID )
                {
                    pagesInMemory++;
                    b |= (1 << stepper);
                }
                stepper++;
                if ( stepper == 8 )
                {
                    outputStream.write( b );
                    b = 0;
                    stepper = 0;
                }
            }
            outputStream.write( b );
            outputStream.flush();
        }
```
instead?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/PageCacheWarmer.java
```diff
@@ -0,0 +1,324 @@
...
+        catch ( CompressorException e )
+        {
+            IOUtils.closeAllSilently( sink );
+            throw new IOException( "Exception when building compressor.", e );
+        }
+    }
+
+    private CompressorStreamFactory compressors()
+    {
+        return new CompressorStreamFactory( true, 1024 );
```
klaren: Can't we reuse the factory?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/PageCacheWarmer.java
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    private File profileOutputFileFinal( PagedFile file )
+    {
+        File mappedFile = file.file();
+        String profileOutputName = "." + mappedFile.getName() + SUFFIX_CACHEPROF;
```
klaren, line 303: Do we want the files to be hidden?

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/PageCacheWarmerKernelExtension.java
```diff
@@ -0,0 +1,174 @@
...
+        long frequencyMillis = config.get( GraphDatabaseSettings.pagecache_warmup_profiling_interval ).toMillis();
+        profileHandle = scheduler.schedule(
+                pageCacheIOHelper, this::tryStartProfile, frequencyMillis, TimeUnit.MILLISECONDS );
+    }
+
+    private void tryStartProfile()
+    {
+        if ( profilingStarted.compareAndSet( false, true ) )
+        {
+            // At this point, we are currently executing inside a *scheduled* executor thread. There are, at this time
```
klaren, line 129: This relies on too much undocumented implementation details of the `Neo4jJobScheduler`. If the problem is that we only have two threads, we can increase that.

If the purpose of the scheduled executor is to just spawn new jobs at a delay, refactor `Neo4jJobScheduler` to do this instead of implementing it here. If you had this problem others will surely follow, and this hidden relationship with the causal clustering is very easy to miss.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/PageLoader.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.pagecache;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+abstract class PageLoader implements Closeable
```
klaren: Can't this be an interface?

integrationtests/src/test/java/org/neo4j/kernel/PageCacheWarmupCcIT.java
```diff
@@ -0,0 +1,108 @@
...
+import org.neo4j.concurrent.BinaryLatch;
+import org.neo4j.graphdb.factory.GraphDatabaseSettings;
+import org.neo4j.kernel.impl.pagecache.PageCacheWarmerMonitor;
+import org.neo4j.kernel.monitoring.Monitors;
+import org.neo4j.test.causalclustering.ClusterRule;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+public class PageCacheWarmupCcIT extends PageCacheWarmupTestSupport
```
klaren, line 41: It really nice we get warmup for cluster members as well! :tada:

integrationtests/src/test/java/org/neo4j/kernel/PageCacheWarmupTestSupport.java
```diff
@@ -0,0 +1,94 @@
...
+    {
+        try ( Transaction tx = db.beginTx() )
+        {
+            Label label = Label.label( "Label" );
+            RelationshipType relationshipType = RelationshipType.withName( "REL" );
+            long[] largeValue = new long[1024];
+            for ( int i = 0; i < 1000; i++ )
+            {
+                Node node = db.createNode( label );
+                node.setProperty( "Niels", "Borh" );
```
klaren, line 47: :denmark:

klaren (changes requested):
This looks really good. An awesome feature to have!

I'm not sure about the default profile interval of `1m`, do we have any feeling of how long the profiling will take on a larger database?

klaren:
Also, shouldn't this be tagged as `changelog`?

chrisvest:
@klaren We don't have very much feeling about how good a default `1m` is. I observed in our soak testing that a profile takes about 50ms to 100ms to complete, so I saw no reason to increase the interval. We do want the profiles to be fairly recent, since we at this level don't have any idea about how the workload changes over time.

integrationtests/src/test/java/org/neo4j/kernel/PageCacheWarmupTestSupport.java
chrisvest: ⚛️

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/PageCacheWarmer.java
chrisvest: I don't think people want to see these by default since they're not super important. I hope hiding them means we'll get fewer support cases that are just about asking what these files are.

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/PageCacheWarmer.java
chrisvest: Looks simpler 👍

enterprise/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/PageCacheWarmerKernelExtension.java
chrisvest: Let's do that in a different PR.

chrisvest:
@klaren I believe I've addressed all of your comments.

klaren (approved):

PR #10958, opened 2018-02-03
Working with Scala 2.12.3\n\nDropped support for v2.3, v3.1 and v3.2

collymy:
Hi,

I have managed to get a version of 3.3.1 working with Scala.2.12.3.

I have dropped backward compatibility with Cypher 2.3, 3.1 and 3.2 because it would have been a lot of extra work to convert them to Scala 2.12 as well and I am only interested in doing some feasibility testing with Neo4j before making a technology decision.

All tests pass except for some in tests in org.neo4j.cypher.internal.javacompat.NotificationAcceptanceTest.java. The tests were giving sensible errors but they did not pass the tests, presumably because of the missing backward compatibility for Cypher 2.3, 3.1 and 3.2.

With this version I have been able to use Neo4j and Cypher without any problems with my Scala 2.12.3 codebase.

I am not sure if this is any use to the project but I compiled with stricter flags available for the Scala 2.12 compiler than you use for 2.11 and I found a number of problems which I have fixed.

If you want to use any of this and need me to sign the CLA let me know and I will do so.

Colly


sherfert:
Hi @collymy and thanks for the contribution. Unfortunately, we won't be able to merge this PR, given that it removes backwards compatibility. When we consider updating our supported scala version, we might come back to this, but for now this is not planned anytime soon.

Therefore, I am closing this PR.

PR #10962, opened 2018-02-06
Fix index regression

pontusmelke:
For an exact index seek there is no reason to fetch the values from the index.


tinwelint (approved):
Yup, nice one!

PR #10963, opened 2018-02-06
Traversal regressions

pontusmelke:
We saw some regressions from changing to the new Kernel API for traversals, this fix includes

- pooling cursors
- faster `assertInUnterminatedTransaction`
- Some minor optimizations on the relationship iterators.


pontusmelke:
@tinwelint you can hold off here a little bit, we still investigating if this fixes the regressions.

PR #10964, opened 2018-02-06
SeekCursor faster for ranged queries

tinwelint:
SeekCursor reads keys/values off of tree nodes while they are potentially
concurrently changing, getting deleted, getting reused and what not.
Therefore there's rigorous validation made after reading a key/value pair.
This validation is expensive and slows down ranged queries because every result
goes through this validation.

What this commit does is to read more key/value pairs while it's reading anyway,
and does validation of the read once per such batch. With a default batch size
of 20 this means that the expensive validation is made for basically every
20th result, instead of for every result.

The batch size can be experimented with, but having too big a batch size
may hurt by means of creating too many KEY/VALUE instances. Also interesting
coupled with this type of change is to consider SeekCursor pooling or to
let user explicitly allocate these SeekCursors and close when done.

### DIFF
--- a/community/index/src/test/java/org/neo4j/index/internal/gbptree/SeekCursorTestBase.java
+++ b/community/index/src/test/java/org/neo4j/index/internal/gbptree/SeekCursorTestBase.java
@@ -21,6 +21,7 @@ package org.neo4j.index.internal.gbptree;
 
 import org.apache.commons.lang3.mutable.MutableBoolean;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
 
 import java.io.IOException;
@@ -35,6 +36,7 @@ import java.util.function.Supplier;
 
 import org.neo4j.io.pagecache.PageCursor;
 import org.neo4j.io.pagecache.impl.DelegatingPageCursor;
+import org.neo4j.test.rule.RandomRule;
 
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
@@ -44,6 +46,7 @@ import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.neo4j.index.internal.gbptree.GenerationSafePointerPair.pointer;
+import static org.neo4j.index.internal.gbptree.SeekCursor.DEFAULT_MAX_READ_AHEAD;
 import static org.neo4j.index.internal.gbptree.TreeNode.Type.INTERNAL;
 import static org.neo4j.index.internal.gbptree.TreeNode.Type.LEAF;
 import static org.neo4j.index.internal.gbptree.ValueMergers.overwrite;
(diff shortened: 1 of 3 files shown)


community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
```diff
@@ -1565,18 +1565,16 @@ public void shouldThrowIllegalStateExceptionOnCallingNextAfterClose() throws Exc
...
-                {
-                    // when
-                    seek.next();
-                    fail( "Should have failed" );
-                }
-                catch ( IllegalStateException e )
-                {
-                    // then good
+                while ( seek.next() )
+                {   // exhaust it, sooner or later it will need to read from the cursor and then notice it's closed
```
burqen: We should not be able to continue using a closed cursor. Even if we get results from the read ahead buffer we need to verify that the cursor is still open.
I think this test should look like it did before to verify this behaviour.

community/index/src/test/java/org/neo4j/index/internal/gbptree/SeekCursorTestBase.java
```diff
@@ -1199,7 +1199,7 @@ public void mustRereadHeadersOnRetry() throws Exception
         // WHEN
         try ( SeekCursor<KEY,VALUE> cursor = new SeekCursor<>( this.cursor,
                 node, from, to, layout, stableGeneration, unstableGeneration, () -> 0L, failingRootCatchup,
-                unstableGeneration, exceptionDecorator ) )
+                unstableGeneration, exceptionDecorator, 1 ) )
```
burqen, line 1208: If we are only testing with `maxCacheSize=1` we are not actually testing this new behaviour. Ideally we would test with some different cache sizes. Maybe even `random(1,20)`.
Some tests would need to be updated for this, must notably `mustFindKeyInsertedOnSeekPosition`.

burqen (changes requested):
Changes look good overall! Excited to see what performance improvements this provides.

Got 2 objections on testing and closing behaviour. See comments.

community/index/src/test/java/org/neo4j/index/internal/gbptree/GBPTreeTest.java
tinwelint: That means an added check on the fast path, sure we can add that, but what is it that makes that absolutely necessary, mostly being nice and fail fast and all that?

burqen: Yes, fail fast and don't hide bugs. The check should only be a simple boolean check (not even volatile) so I think it's worth it.

tinwelint:
@burqen take a look again, I amended the revert of the changed `GBPTreeTest` together with a check in `SeekCursor#next` into the main commit and then added a commit for extended testing of `SeekCursorTestBase`. Turned out better with your ideas I think.

tinwelint:
Benchmarking shows great improvements for result sets of > 1, but a smallish regression for result sets of 1. Marking as NOT READY FOR MERGE while investigating.

EDIT: results are looking good now

burqen (approved):

PR #10965, opened 2018-02-06
Calculates correct max memory to use for caching

tinwelint:


community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/Configuration.java
```diff
@@ -286,15 +298,20 @@ static long calculateMaxMemoryFromPercent( int percent )
...
 
         double factor = percent / 100D;
-        return round( (freePhysicalMemory - Runtime.getRuntime().maxMemory()) * factor );
+        return round( (totalPhysicalMemory - Runtime.getRuntime().maxMemory()) * factor );
+    }
+
+    public static void main( String[] args )
+    {
+
     }
```
chrisvest: Stray `main`.

tinwelint:
<blush>

Den 7 feb. 2018 00:16 skrev "Chris Vest" <notifications@github.com>:

yoyo
```diff
@@ -0,0 +1,132 @@
+ community/codegen/src/main/java/org/neo4j/codegen/Expression.java                                                                                     |   2 +-
```
ragadeeshu: This probably shouldn't be here?

ragadeeshu (changes requested):

ragadeeshu (approved):

PR #10966, opened 2018-02-06
Removes special handing of 2.3 test case that does not seem flaky

sherfert:
This test was supposed to be flaky, but I can never get it to pass.
Since we are not fixing the bug itself, this simply keeps that
test on the blacklist and removes the special flakiness handling.

It also prevents NullPointerExceptions when catching an Exception
that has a null error message.


Lojjs (approved):
I think it looks fine. Hard to say whether the flakiness is gone or not, but at least it works locally. The failing causal clustering test also work locally.

PR #10968, opened 2018-02-06
Rename the Cypher frontend module to openCypher-frontend

Mats-SX:
This is a small second step in the process of moving this module to the realm of `openCypher`. (The first step was relicensing the module to Apache 2.0.) 

This renames the module from `neo4j-cypher-frontend-3.4` to `openCypher-frontend-1`. The version stays the same (tied to Neo4j, latest `3.4.0-SNAPSHOT`). The description is also extended.

This would accomplish two things: signal to all users of this module that it belongs in `openCypher` space, and remove the Neo4j version from the name. A subsequent move of this module to another repository with a detached release process could choose a different version scheme, as long as it uses version numbers above `3.4`.

Possibly, we should also change the version, to not occupy the version space for the new `artifactId`. Although this may be a moot point because we expect the `groupId` to change once we make the move.

### DIFF
--- a/community/cypher/cypher-logical-plans-3.4/pom.xml
+++ b/community/cypher/cypher-logical-plans-3.4/pom.xml
@@ -65,7 +65,7 @@
 
     <dependency>
       <groupId>org.neo4j</groupId>
-      <artifactId>neo4j-cypher-frontend-3.4</artifactId>
+      <artifactId>openCypher-frontend-1</artifactId>
       <version>${project.version}</version>
     </dependency>
 
@@ -79,7 +79,7 @@
 
     <dependency>
       <groupId>org.neo4j</groupId>
-      <artifactId>neo4j-cypher-frontend-3.4</artifactId>
+      <artifactId>openCypher-frontend-1</artifactId>
       <version>${project.version}</version>
       <type>test-jar</type>
       <scope>test</scope>
--- a/community/cypher/cypher-planner-3.4/pom.xml
+++ b/community/cypher/cypher-planner-3.4/pom.xml
@@ -88,7 +88,7 @@
 
     <dependency>
       <groupId>org.neo4j</groupId>
-      <artifactId>neo4j-cypher-frontend-3.4</artifactId>
+      <artifactId>openCypher-frontend-1</artifactId>
       <version>${project.version}</version>
     </dependency>
 
@@ -137,7 +137,7 @@
 
     <dependency>
       <groupId>org.neo4j</groupId>
-      <artifactId>neo4j-cypher-frontend-3.4</artifactId>
+      <artifactId>openCypher-frontend-1</artifactId>
       <version>${project.version}</version>
       <type>test-jar</type>
       <scope>test</scope>
--- a/community/cypher/cypher/pom.xml
+++ b/community/cypher/cypher/pom.xml
@@ -296,7 +296,7 @@
 
     <dependency>
       <groupId>org.neo4j</groupId>
-      <artifactId>neo4j-cypher-frontend-3.4</artifactId>
+      <artifactId>openCypher-frontend-1</artifactId>
       <version>${project.version}</version>
       <type>test-jar</type>
       <scope>test</scope>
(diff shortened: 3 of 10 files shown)


Mats-SX:
@boggle @systay @fickludd @craigtaverner Thoughts on this?

community/cypher/cypher-logical-plans-3.4/pom.xml
```diff
@@ -65,7 +65,7 @@
 
     <dependency>
       <groupId>org.neo4j</groupId>
-      <artifactId>neo4j-cypher-frontend-3.4</artifactId>
+      <artifactId>openCypher-frontend</artifactId>
```
craigtaverner: I'm guessing the removal of the version number from the module is because this frontend should only be used once within a single project (ie. unlike the cypher-compiler-x.y module we will not have multiple versions in a single Neo4j instance)?

However, doesn't the cypher-compiler-x.y depend on cypher-frontend-x.y, so we would still need versioned frontends for the versioned compilers (ie. when Neo4j includes two compilers, they each depend on a different frontend)?

craigtaverner (changes requested):
I'm concerned about the removal of the version number from the module. If we remove it, we should presumably also remove it from the package names (but they are still versioned). However, I question whether it can be removed at all, since the compiler (or planner now) depends on it, and Cypher will depend on multiple compilers (planners) and they will require multiple versions of the frontend, requiring both the module and the packages to have the version in the name.

Mats-SX:
@craigtaverner From my understanding, the version in the module name is redundant. For `frontend-3.4` there will only be versions `3.4.x`; for `frontend-3.3` there will only be versions `3.3.x` and so on. There will never be a `frontend-3.1` with version `3.3.0` for example.

A future version of Neo4j Cypher `3.5.x` would depend on two versions of `openCypher-frontend`: `3.4.0` and `3.5.x`. But perhaps this is where the problem arises? Is multiple versions of the same artifact an issue in general, or something we can solve with namespacing or exclusions?

For the package: I'm not sure. Perhaps it could be renamed too, but I think the value compared to the effort might not be great.

Mats-SX:
@craigtaverner I realised another thing that might make the discussion on multiple versions moot: we don't intend for this to be a long-term solution. The long-term solution is to completely move this module out of the code base into its own separate repository, with detached release process.

That means that there will probably not be a `3.5.x` version of the `org.neo4j.openCypher-frontend` module (or just one or two alphas, depending on when we actually do it), but instead a new module `org.opencypher.openCypher-frontend` with its own versions. Whether Neo4j can depend on multiple versions of that, or whether we'd need to redesign our compatibility layer I'm not sure of, but I think it is solvable.

community/cypher/cypher-logical-plans-3.4/pom.xml
Mats-SX: I'm not sure that `cypher-compiler-3.5` depending on `openCypher-frontend` version `1.0` and `cypher-compiler-4.0` depending on `openCypher-frontend` version `2.5` is a problem, is it? Will they conflict on the classpath?

Couldn't we redesign the compatibility layer so that Neo4j Cypher only depends on a single version of `openCypher-frontend` which is backwards compatible, and then have multiples of `cypher-compiler-x.y` to which it redirects based on compatibility mode selected by the user?

craigtaverner (approved):

Mats-SX:
Blocked due to uncertainty of the future of this module and its versioning scheme and usage. Main crux: can not depend on multiple versions of the same module. If we were to stick a version name into the artifact name, what version should that be?

systay:
@Mats-SX Yeah, that is what I was thinking. Could name it `openCypher-frontend-1.0` or maybe `-9.0`?

Mats-SX:
@craigtaverner @systay I now renamed it to `openCypher-frontend-1`, keeping its version to `3.4.0` (tied to Neo4j). That's the most minimal step that still achieves putting `openCypher` into the name of this artifact that I can come up with.

I don't feel like the `-1` matters a lot, but allows us to keep the existing multi-version setup even though it may not end up becoming a problem without.

Mats-SX:
@craigtaverner Test failures are docs build (probably depends on module name) and an unrelated test. This should be good to merge.

PR #10970, opened 2018-02-06
Report the actual listening Bolt port in the log

technige:
This change alters the log output to report the actual port(s) on which the Bolt connector(s) is/are listening.

While not commonly used outside testing environments, it is possible to configure the socket to [bind to port zero](https://www.dnorth.net/2012/03/17/the-port-0-trick/), which asks the network stack to select a vacant, high-numbered port. Previously, this was reported in the log as `0` rather than as the actual port number selected:
```
2018-02-06 17:28:50.788+0000 INFO  ======== Neo4j 3.4.0-SNAPSHOT ========
2018-02-06 17:28:50.813+0000 INFO  Starting...
2018-02-06 17:28:51.973+0000 INFO  Bolt enabled on [::]:0.
```

This PR moves the logging to a point when the actual port number is known and simply reports that instead:
```
2018-02-06 17:28:50.788+0000 INFO  ======== Neo4j 3.4.0-SNAPSHOT ========
2018-02-06 17:28:50.813+0000 INFO  Starting...
2018-02-06 17:28:51.973+0000 INFO  Bolt enabled on [::]:41359.
```

### DIFF
--- a/community/bolt/src/main/java/org/neo4j/bolt/transport/NettyServer.java
+++ b/community/bolt/src/main/java/org/neo4j/bolt/transport/NettyServer.java
@@ -28,8 +28,6 @@ import io.netty.channel.EventLoopGroup;
 import io.netty.channel.epoll.Epoll;
 
 import java.net.InetSocketAddress;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ThreadFactory;
 
@@ -40,7 +38,6 @@ import org.neo4j.helpers.ListenSocketAddress;
 import org.neo4j.helpers.PortBindException;
 import org.neo4j.kernel.configuration.BoltConnector;
 import org.neo4j.kernel.configuration.ConnectorPortRegister;
-import org.neo4j.kernel.impl.logging.LogService;
 import org.neo4j.kernel.lifecycle.LifecycleAdapter;
 import org.neo4j.logging.Log;
 import org.neo4j.util.FeatureToggles;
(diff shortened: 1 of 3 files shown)


community/bolt/src/main/java/org/neo4j/bolt/transport/NettyServer.java
```diff
@@ -109,6 +112,9 @@ public void start() throws Throwable
                                 .bind( protocolInitializer.address().socketAddress() ).sync();
                 InetSocketAddress localAddress = (InetSocketAddress) channelFuture.channel().localAddress();
                 connectionRegister.register( boltConnector.key(), localAddress );
+                ListenSocketAddress listenSocketAddress = new ListenSocketAddress(
+                        protocolInitializer.address().getHostname(), localAddress.getPort() );
+                log.info( "Bolt enabled on %s.", listenSocketAddress );
```
lutovich: Is it possible to log `localAddress` and not even create `listenSocketAddress`?

lutovich (approved):
@technige looks good, one small question.

community/bolt/src/main/java/org/neo4j/bolt/transport/NettyServer.java
technige: This is wrapped in a `ListenSocketAddress` because that has support for better formatting. The default `toString` that comes with `InetSocketAddress` is a bit messier IMHO, especially with IPv6. Plus doing this retains format consistency with what we have today.

technige: We could alternatively extract the formatting logic into a separate static function.

lutovich: Understood, it's just not immediately clear why is `ListenSocketAddress` required. I think it's okey to keep it as is, or call `toString()` explicitly after construction just to make it clear:

```java
String listenSocketAddress = new ListenSocketAddress(
  protocolInitializer.address().getHostname(), localAddress.getPort() ).toString();
```

PR #10971, opened 2018-02-07
Store Arrays of temporal values.

sherfert:
This is still missing the storage of DateTimes with ZoneIds (and Arrays thereof), which awaits https://github.com/neo4j/neo4j/pull/10954 to be mergeable.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/DynamicArrayStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/DynamicArrayStore.java
@@ -113,7 +113,8 @@ public class DynamicArrayStore extends AbstractDynamicStore
     {
         super( fileName, configuration, idType, idGeneratorFactory, pageCache,
                 logProvider, TYPE_DESCRIPTOR, dataSizeFromConfiguration, recordFormats.dynamic(), recordFormats.storeVersion(), openOptions );
-        allowStorePointsAndTemporal = recordFormats.hasCapability( Capability.POINT_PROPERTIES );
+        allowStorePointsAndTemporal = recordFormats.hasCapability( Capability.POINT_PROPERTIES )
+                && recordFormats.hasCapability( Capability.TEMPORAL_PROPERTIES );
     }
 
     @Override
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/store/DynamicArrayStore.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/store/DynamicArrayStore.java
@@ -23,6 +23,11 @@ import java.io.File;
 import java.lang.reflect.Array;
 import java.nio.ByteBuffer;
 import java.nio.file.OpenOption;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.OffsetTime;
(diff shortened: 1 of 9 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/store/DynamicArrayStore.java
```diff
@@ -106,7 +113,7 @@ public DynamicArrayStore(
     {
         super( fileName, configuration, idType, idGeneratorFactory, pageCache,
                 logProvider, TYPE_DESCRIPTOR, dataSizeFromConfiguration, recordFormats.dynamic(), recordFormats.storeVersion(), openOptions );
-        allowStorePoints = recordFormats.hasCapability( Capability.POINT_PROPERTIES );
+        allowStorePointsAndTemporal = recordFormats.hasCapability( Capability.POINT_PROPERTIES );
```
tinwelint: did you forget to also add `recordFormats.hasCapability( Capability.TEMPORAL_PROPERTIES )`?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/PropertyType.java
```diff
@@ -159,6 +159,10 @@ else if ( itemType <= GEOMETRY.byteValue() )
             {
                 return headOf( recordBytes, DynamicArrayStore.GEOMETRY_HEADER_SIZE );
             }
+            else if ( itemType <= TEMPORAL.byteValue() )
```
tinwelint: The `<=` comparison looks weird. I see that this is most likely a copy-paste issue stemming from the `DOUBLE` comparison above, which `GEOMETRY` also copy-pasted. The `DOUBLE` comparison is, I believe, a special case for covering all integral and floating point types in one comparison, knowing that the `DOUBLE` `byteValue()` is the largest of those. I think it makes sense to change this condition to `==` and potentially also that of `GEOMETRY` while you're at it.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TemporalType.java
```diff
@@ -77,9 +89,22 @@ public int calculateNumberOfBlocksUsedForTemporal( long firstBlock )
...
+                        LocalDate[] dates = new LocalDate[numbers.length()];
+                        for ( int i = 0; i < dates.length; i++ )
+                        {
+                            dates[i] = LocalDate.ofEpochDay( numbers.longValue( i ) );
+                        }
+                        return Values.dateArray( dates );
+                    }
+                    else
+                    {
+                        throw new RuntimeException( "Array with unexpected type. Actual:" + dataValue.getClass().getSimpleName() + ". Expected: LongArray." );
```
tinwelint: Would be better to throw some more specific and coherent exception, like `InvalidRecordException` or the likes. Likewise for the other similar places.

tinwelint (changes requested):
I think it looks good. Some minor findings, that's all

community/kernel/src/main/java/org/neo4j/kernel/impl/store/DynamicArrayStore.java
sherfert: Indeed

tinwelint (approved):

PR #10972, opened 2018-02-07
Add a PageCursor.copyTo(long,ByteBuffer) method.

chrisvest:
Using this method, we can choose to quickly copy memory out of our page cursors for interpretation,
so we can simplify the cases where inconsistent-read-robust record decoding is too difficult to do.

### DIFF
--- a/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
+++ b/community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
@@ -21,6 +21,7 @@ package org.neo4j.io.pagecache.impl.muninn;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.util.Objects;
 
 import org.neo4j.io.pagecache.CursorException;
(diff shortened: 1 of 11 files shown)


community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
```diff
@@ -4367,6 +4369,174 @@ public void copyToMustCheckBounds() throws Exception
...
+        buffer.clear();
+        copied = cursor.copyTo( 1, buffer );
+        assertFalse( cursor.checkAndClearBoundsFlag() );
+        assertThat( copied, is( filePageSize - 1 ) );
+        assertThat( buffer.position(), is( filePageSize - 1 ) );
+        assertThat( buffer.remaining(), is( 1 ) );
+        buffer.clear();
+
+        // Smaller buffer at offset zero.
+        zapBuffer( buffer );
```
MishaDemianenko, line 4461: not sure why do we zapBuffer here since next operation on top of that buffer is clear

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
```diff
@@ -789,6 +790,53 @@ public int copyTo( int sourceOffset, PageCursor targetCursor, int targetOffset,
         return 0;
     }
 
+    @Override
+    public int copyTo( int sourceOffset, ByteBuffer buf )
+    {
+        if ( buf.getClass() == UnsafeUtil.directByteBufferClass )
```
MishaDemianenko: I think buf.isReadOnly() should be checked not only for `UnsafeUtil.directByteBufferClass` but for all buffers that provided; then you will not gonna need  `assert isDirectWritableByteBuffer( buf )` since that will be duplicated check.
Also `buf.getClass() == UnsafeUtil.directByteBufferClass` implies that byteBuffer classloader and UnsafeUtilClass loader is the same, maybe we need to mention it at least as a comment.

MishaDemianenko (approved):
Have only two minor comments, will approve directly, but still would be nice to get reply on those

community/io/src/main/java/org/neo4j/io/pagecache/impl/muninn/MuninnPageCursor.java
chrisvest: I can move the checks into the if-condition, sending callers with weird byte buffers into the fall-back code.

They are JDK classes, so they should have the same class loaders, either the system class loader or the boot class loader.

community/io/src/test/java/org/neo4j/io/pagecache/PageCacheTest.java
chrisvest: The `zapBuffer` method overwrites the data from further up with zeros. The `buffer.clear` method resets the buffer position, mark and limit, but does not change any data.

MishaDemianenko: it does not change any data but we also do not check byte buffer implementation right

chrisvest: I think the check is sufficient for the test.

PR #10973, opened 2018-02-07
Let dbms.procedures YIELD mode

### LINKED ISSUE #9170: Feature Request: let dbms.procedures() YIELD mode
Currently `dbms.procedures()` YIELDS roles (and only for the enterprise edition), but it would be helpful to YIELD mode as well, which would save from having to decode mode based on roles values.

sherfert:
Fixes #9170 

changelog: dbms.procedures() which lists all built-in procedures now also list their mode


Lojjs (approved):
Nice!

PR #10975, opened 2018-02-07
Less indirection

pontusmelke:
Instead of first creating iterator and then wrap them in progressors we
create progressors directly. This PR introduce some duplication but in a future not too distant we should be able to get rid of some of the iterator variants

### DIFF
--- a/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/DocValuesCollector.java
+++ b/community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/DocValuesCollector.java
@@ -55,6 +55,8 @@ import org.neo4j.helpers.collection.ArrayIterator;
 import org.neo4j.helpers.collection.PrefetchingIterator;
 import org.neo4j.index.impl.lucene.explicit.EmptyIndexHits;
 import org.neo4j.kernel.impl.api.explicitindex.AbstractIndexHits;
+import org.neo4j.storageengine.api.schema.IndexProgressor;
+import org.neo4j.values.storable.Value;
 
 /**
  * Collector to record per-segment {@code DocIdSet}s and {@code LeafReaderContext}s for every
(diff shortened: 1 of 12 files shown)


community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/DocValuesCollector.java
```diff
@@ -318,7 +330,7 @@ private void replayTo( Collector collector ) throws IOException
      * Two, lookup a value for the current doc in a sidecar {@code NumericDocValues} field.
      * That is, this iterator has a main field, that drives the iteration and allow for lookups
      * in other, secondary fields based on the current document of the main iteration.
-     *
+     * <p>
```
SaschaPeukert, line 333: You might want to remove that one?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/reader/SimpleIndexReader.java
```diff
@@ -139,6 +141,58 @@ public PrimitiveLongResourceIterator query( IndexQuery... predicates ) throws In
...
+            assertNotComposite( predicates );
+            IndexQuery.StringContainsPredicate scp = (IndexQuery.StringContainsPredicate) predicate;
+            return containsString( client, scp.contains() );
+        case stringSuffix:
+            assertNotComposite( predicates );
+            IndexQuery.StringSuffixPredicate ssp = (IndexQuery.StringSuffixPredicate) predicate;
+            return endsWith( client, ssp.suffix() );
+        default:
+            // todo figure out a more specific exception
+            throw new RuntimeException( "Index query not supported: " + Arrays.toString( predicates ) );
```
SaschaPeukert, line 192: Now would be a good time for that, right?

SaschaPeukert (commented):
Looks good to me! 👍

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/reader/SimpleIndexReader.java
pontusmelke: is now the time @burqen @tinwelint ?

burqen: The time to figure out at more specific exception? Yes, it's always a good time for that! ;)
To be honest though, I can't think of what more to add to that exception.

tinwelint (changes requested):
Why would we need two `query` methods that are pretty similar? Is one fast and one slow? Why use the slow one and when?

PR #10976, opened 2018-02-07
Removed the cipher filter enforced by new jetty

zhenlineo:
If a user configures cipher suites and protocols via configuration ssl policy, then jetty server should pick up whatever the user configured.


zhenlineo:
Enforce TLSv1.2 by default instead https://github.com/neo4j/neo4j/pull/10984

PR #10977, opened 2018-02-07
Small fixes for bloom

ragadeeshu:


enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomProcedures.java
```diff
@@ -122,6 +124,7 @@ public void setIndexedRelationshipPropertyKeys( @Name( "propertyKeys" ) List<Str
 
     private Stream<EntityOutput> queryAsStream( List<String> terms, ReadOnlyFulltext indexReader, boolean fuzzy, boolean matchAll )
     {
+        terms = terms.stream().flatMap( s -> Arrays.stream( s.split( "\\s+" ) ) ).collect( Collectors.toList() );
```
chrisvest, line 127: I feel like this is missing a test. I also don't get why we do this for the non-fuzzy queries as well.

ragadeeshu: Because doesn't hurt and results in cleaner code. Test added.

ragadeeshu:
@chrisvest updated

chrisvest (approved):

PR #10984, opened 2018-02-08
Enforce TLSv1.2 for HTTPS and Bolt server

zhenlineo:
HTTPS server and Bolt server use legacy ssl policy. This policy does not expose any configuration option for allowed security protocols or cipher suites. As a result, the REST and Bolt server were running with protocol and cipher suites provided by jvm by default. However to provide more secured neo4j service, from this version, we will enforce TLSv1.2 as default for legacy ssl policy.
 
For users who would like to have their own customized protocols or cipher suites, they could migrate their ssl policy for bolt and https to new [`dbms.ssl.policy`](https://neo4j.com/docs/operations-manual/current/security/ssl-framework/#_policy_definition) to have the full control of security settings.

For users whose jdk does not have TLSv1.2 enabled by default (such as **ibm-jdk8**), if they would like to have neo4j HTTPS and Bolt server running normally with TLSv1.2, they should run the server with jvm option `com.ibm.jsse2.overrideDefaultTLS=true`. This could be achieved by adding a line `dbms.jvm.additional=-Dcom.ibm.jsse2.overrideDefaultTLS=true` in `neo4j.config` file.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
@@ -20,6 +20,7 @@
 package org.neo4j.kernel.configuration.ssl;
 
 import java.io.File;
+import java.util.List;
 
 import org.neo4j.configuration.Description;
 import org.neo4j.configuration.Internal;
@@ -27,8 +28,12 @@ import org.neo4j.configuration.LoadableConfig;
 import org.neo4j.graphdb.config.Setting;
 
 import static org.neo4j.kernel.configuration.Settings.PATH;
+import static org.neo4j.kernel.configuration.Settings.STRING_LIST;
 import static org.neo4j.kernel.configuration.Settings.derivedSetting;
 import static org.neo4j.kernel.configuration.Settings.pathSetting;
+import static org.neo4j.kernel.configuration.Settings.setting;
+import static org.neo4j.kernel.configuration.ssl.SslPolicyConfig.CIPHER_SUITES_DEFAULTS;
+import static org.neo4j.kernel.configuration.ssl.SslPolicyConfig.TLS_VERSION_DEFAULTS;
 
 /**
  * To be removed in favour of {@link SslPolicyConfig}. The settings below are still
@@ -54,4 +59,15 @@ public class LegacySslPolicyConfig implements LoadableConfig
     public static final Setting<File> tls_key_file =
             derivedSetting( "unsupported.dbms.security.tls_key_file", certificates_directory,
                     certificates -> new File( certificates, "neo4j.key" ), PATH );
+
+    @Internal
+    @Description( "Default encryption protocol used for legacy SSl policy." )
+    static final Setting<List<String>> default_security_protocol =
+            setting( "unsupported.dbms.security.protocol", STRING_LIST, TLS_VERSION_DEFAULTS );
+
+    @Internal
+    @Description( "Default encryption protocol used for legacy SSl policy." )
+    static final Setting<List<String>> default_security_cipher_suites =
+            setting( "unsupported.dbms.security.cipher_suites", STRING_LIST, CIPHER_SUITES_DEFAULTS );
+
 }
--- a/community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
@@ -20,7 +20,6 @@
 package org.neo4j.kernel.configuration.ssl;
 
 import java.io.File;
-import java.util.List;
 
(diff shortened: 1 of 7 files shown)


community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
```diff
@@ -54,4 +59,15 @@
     public static final Setting<File> tls_key_file =
             derivedSetting( "unsupported.dbms.security.tls_key_file", certificates_directory,
                     certificates -> new File( certificates, "neo4j.key" ), PATH );
+
+    @Internal
+    @Description( "Default encryption protocol used for legacy SSl policy." )
```
martinfurmanski: It is a bit wrong to mention default in the descriptions and names here, since you are overriding a default by using the setting. Applies to all settings and descriptions.

It is also not encryption protocol, but SSL/TLS protocol versions. Since most of the things here say TLS I think it is better perhaps to stick to that only.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
```diff
@@ -54,4 +59,15 @@
...
+    static final Setting<List<String>> default_security_protocol =
+            setting( "unsupported.dbms.security.protocol", STRING_LIST, TLS_VERSION_DEFAULTS );
+
+    @Internal
+    @Description( "Default encryption protocol used for legacy SSl policy." )
+    static final Setting<List<String>> default_security_cipher_suites =
```
martinfurmanski: Maybe tls_ciphers or tls_cipher_suites instead?

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
```diff
@@ -54,4 +59,15 @@
...
             derivedSetting( "unsupported.dbms.security.tls_key_file", certificates_directory,
                     certificates -> new File( certificates, "neo4j.key" ), PATH );
+
+    @Internal
+    @Description( "Default encryption protocol used for legacy SSl policy." )
+    static final Setting<List<String>> default_security_protocol =
+            setting( "unsupported.dbms.security.protocol", STRING_LIST, TLS_VERSION_DEFAULTS );
+
+    @Internal
+    @Description( "Default encryption protocol used for legacy SSl policy." )
```
martinfurmanski: Not encryption protocol, but ciphers or cipher suites used for SSL/TLS (upper-case L). Since most of the things here say TLS I think it is better to stick to that.

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/LegacySslPolicyConfig.java
```diff
@@ -54,4 +59,15 @@
     public static final Setting<File> tls_key_file =
             derivedSetting( "unsupported.dbms.security.tls_key_file", certificates_directory,
                     certificates -> new File( certificates, "neo4j.key" ), PATH );
+
+    @Internal
+    @Description( "Default encryption protocol used for legacy SSl policy." )
+    static final Setting<List<String>> default_security_protocol =
```
martinfurmanski: Maybe tls_protocol instead of default_security_protocol?

martinfurmanski (changes requested):
Note that you are adding more flexibility to the legacy policy. While I don't see a big reason not to, I don't see a big reason to either. You can override in any deployment with the new unified SSL policy system.

zhenlineo:
Nice catch. I have no intension to extend legacy policy. Now hard-coded them.

martinfurmanski (approved):

community/kernel/src/main/java/org/neo4j/kernel/configuration/ssl/SslPolicyConfig.java
```diff
@@ -44,7 +45,8 @@
 @Group( "dbms.ssl.policy" )
 public class SslPolicyConfig
 {
-    private static final String TLS_VERSION_DEFAULTS = join( ",", new String[]{"TLSv1.2"} );
+    public static final List<String> TLS_VERSION_DEFAULTS = asList( "TLSv1.2" );
+    public static final List<String> CIPHER_SUITES_DEFAULTS = null;
```
MishaDemianenko, line 49: please use the empty list here instead, you digging a grave for your own joinList that need to check if the value is null; plus making future code not null-safe

zhenlineo: It has to be null because of the two lines added in the `SslSocketConnectorFactory.java`.
If it is null, then it will default to use jetty's cipher filter, otherwise if it an empty list (which means a user wrote this config on purpose), then it will overwrite the jetty's default filter with `sslContextFactory.setExcludeCipherSuites();`.

PR #10986, opened 2018-02-08
3.4 modifier protocols

andrewkerr9000:
Handshake modifier protocols.

Does not include installers for protocols, or much thought into what they might be.

Modifier protocol type (name + identifier) may require some thought.

May refactor to have handshake client send modifier protocol requests without waiting for an application protocol response.

### DIFF
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ClientNettyPipelineBuilder.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ClientNettyPipelineBuilder.java
@@ -17,22 +17,24 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
-package org.neo4j.causalclustering.protocol.handshake;
+package org.neo4j.causalclustering.protocol;
 
-import static org.neo4j.causalclustering.protocol.handshake.StatusCode.FAILURE;
+import io.netty.channel.ChannelPipeline;
+import io.netty.handler.codec.LengthFieldPrepender;
 
-public class ApplicationProtocolResponse extends BaseProtocolResponse implements ClientMessage
+import org.neo4j.logging.Log;
+
+public class ClientNettyPipelineBuilder extends NettyPipelineBuilder<ProtocolInstaller.Orientation.Client, ClientNettyPipelineBuilder>
 {
-    public static final ApplicationProtocolResponse NO_PROTOCOL = new ApplicationProtocolResponse( FAILURE, "", 0 );
-
-    ApplicationProtocolResponse( StatusCode statusCode, String protocolName, int version )
+    ClientNettyPipelineBuilder( ChannelPipeline pipeline, Log log )
     {
-        super( statusCode, protocolName, version );
+        super( pipeline, log );
     }
 
     @Override
-    public void dispatch( ClientMessageHandler handler )
+    public ClientNettyPipelineBuilder addFraming()
     {
-        handler.handle( this );
+        add( "frame_encoder", new LengthFieldPrepender( 4 ) );
+        return this;
     }
 }
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ModifierProtocolInstaller.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ModifierProtocolInstaller.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.causalclustering.protocol;
+
+import io.netty.handler.codec.compression.SnappyFrameDecoder;
+import io.netty.handler.codec.compression.SnappyFrameEncoder;
+
+import java.util.List;
+
+import org.neo4j.causalclustering.protocol.ProtocolInstaller.Orientation;
+
+import static java.util.Arrays.asList;
+
+public interface ModifierProtocolInstaller<O extends Orientation>
+{
+    Protocol.ModifierProtocol protocol();
+
+    <BUILDER extends NettyPipelineBuilder<O,BUILDER>> void apply( NettyPipelineBuilder<O,BUILDER> nettyPipelineBuilder );
+
+    List<ModifierProtocolInstaller<Orientation.Server>> serverCompressionInstallers = asList( SnappyServer.instance );
+    List<ModifierProtocolInstaller<Orientation.Client>> clientCompressionInstallers = asList( SnappyClient.instance );
+
+    List<ModifierProtocolInstaller<Orientation.Server>> allServerInstallers = serverCompressionInstallers;
+    List<ModifierProtocolInstaller<Orientation.Client>> allClientInstallers = clientCompressionInstallers;
+
+    class SnappyClient implements ModifierProtocolInstaller<Orientation.Client>
+    {
+        public static final SnappyClient instance = new SnappyClient();
+
+        private SnappyClient()
+        {
+        }
+
+        @Override
+        public Protocol.ModifierProtocol protocol()
+        {
+            return Protocol.ModifierProtocols.COMPRESSION_SNAPPY;
+        }
+
+        @Override
+        public <BUILDER extends NettyPipelineBuilder<Orientation.Client,BUILDER>> void apply(
+                NettyPipelineBuilder<Orientation.Client,BUILDER> nettyPipelineBuilder )
+        {
+            nettyPipelineBuilder.add( "snappy_encoder", new SnappyFrameEncoder() );
+        }
+    }
+
+    class SnappyServer implements ModifierProtocolInstaller<Orientation.Server>
+    {
+        private SnappyServer()
+        {
+        }
+
+        public static final SnappyServer instance = new SnappyServer();
+
+        @Override
+        public Protocol.ModifierProtocol protocol()
+        {
+            return Protocol.ModifierProtocols.COMPRESSION_SNAPPY;
+        }
+
+        @Override
+        public <BUILDER extends NettyPipelineBuilder<Orientation.Server,BUILDER>> void apply(
+                NettyPipelineBuilder<Orientation.Server,BUILDER> nettyPipelineBuilder )
+        {
+            nettyPipelineBuilder.add( "snappy_decoder", new SnappyFrameDecoder() );
+        }
+    }
+}
--- a/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/Protocol.java
+++ b/enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/Protocol.java
@@ -25,26 +25,36 @@ public interface Protocol
 
     int version();
 
-    enum Identifier
+    interface Identifier<T extends Protocol>
+    {
+        String canonicalName();
+    }
+
+    interface ApplicationProtocol extends Protocol
+    {
+    }
+
+    enum ApplicationProtocolIdentifier implements Identifier<ApplicationProtocol>
     {
         RAFT,
         CATCHUP;
 
+        @Override
         public String canonicalName()
         {
             return name().toLowerCase();
         }
     }
 
-    enum Protocols implements Protocol
+    enum ApplicationProtocols implements ApplicationProtocol
     {
-        RAFT_1( Identifier.RAFT, 1 ),
-        CATCHUP_1( Identifier.CATCHUP, 1 );
+        RAFT_1( ApplicationProtocolIdentifier.RAFT, 1 ),
+        CATCHUP_1( ApplicationProtocolIdentifier.CATCHUP, 1 );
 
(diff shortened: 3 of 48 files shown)


andrewkerr9000:
Functionally complete, including installing protocols, but could do with some refactoring.

Note that installing Snappy is an alternative to installing framing, not an addition to. Therefore the framing has been implemented as another compression protocol with an earlier version number, meaning that a compression protocol must be included in the negotiation.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ProtocolInstallerRepository.java
```diff
@@ -49,15 +71,26 @@ private void addTo( Map<Protocol,ProtocolInstaller<O>> tempServerMap, ProtocolIn
...
+        ApplicationProtocol applicationProtocol = protocolStack.applicationProtocol();
+        Set<Protocol.ModifierProtocol> modifierProtocols = protocolStack.modifierProtocols();
+        List<UnaryOperator<NettyPipelineBuilder>> modifierOperators = modifierProtocols.stream()
+                .map( protocol -> Optional.ofNullable( modifiers.get( protocol ) ) )
+                .flatMap( Streams::ofOptional )
+                .flatMap( List::stream )
+                .collect( Collectors.toList() );
+        ProtocolInstaller protocolInstaller = installers.get( applicationProtocol );
+
+        // TODO need at least one modifier to frame
```
martinfurmanski: I find this perhaps slightly controversial to mix framing in together with modifier protocols, and negotiating the basic framing. I think we can definitely let it in but perhaps consider it up for discussion later on when we take a fresh look at protocols and protocol specifications in isolation.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/handshake/ClientMessageDecoder.java
```diff
@@ -88,4 +79,22 @@ protected void decode( ChannelHandlerContext ctx, ByteBuf in, List<Object> out )
             return;
         }
     }
+
+    private <T extends BaseProtocolResponse> Optional<T> decodeProtocolResponse( TriFunction<StatusCode, String, Integer, T> constructor, ByteBuf in )
```
martinfurmanski: Optional here doesn't seem good. We want to fail and throw an exception if there is some sort of protocol violation. Decoding completely and correctly is not optional.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/handshake/HandshakeClient.java
```diff
@@ -19,42 +19,68 @@
...
+import java.util.stream.Collectors;
 
 import org.neo4j.causalclustering.messaging.Channel;
 import org.neo4j.causalclustering.protocol.Protocol;
+import org.neo4j.helpers.collection.Pair;
+import org.neo4j.stream.Streams;
 
 import static org.neo4j.causalclustering.protocol.handshake.StatusCode.SUCCESS;
 
 // TODO: modifier protocols
```
martinfurmanski: TODO can be removed

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/handshake/HandshakeClient.java
```diff
@@ -19,42 +19,68 @@
...
+        knownModifierProtocolVersions.forEach( modifierProtocolSelection ->
+                channel.write( new ModifierProtocolRequest( modifierProtocolSelection.identifier(), modifierProtocolSelection .versions() ) ) );
+
+        channel.writeAndFlush( new ApplicationProtocolRequest( knownApplicationProtocolVersions.identifier(), knownApplicationProtocolVersions.versions() ) );
 
         return future;
     }
 
-    private void ensureMagic()
+    private void ensureMagic( Runnable block )
```
martinfurmanski: This seems more obfuscated compared to the previous version where the pre-conditions are checked and the method aborts (throws) early when they aren't fulfilled. It creates more entangling between two pieces of functionality that could be quite separate concerns.

martinfurmanski (commented):
Looks mostly good but a few minor comments. Would be good to get #10999 in and prove we are stable before merging in more work though.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ProtocolInstallerRepository.java
andrewkerr9000: We either have framing or snappy. It is a bit odd, and I don't have a good solution for it yet.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/handshake/ClientMessageDecoder.java
andrewkerr9000: Yes, this is to have something working before implementing handling an unknown status code. There is a TODO below.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/handshake/HandshakeClient.java
andrewkerr9000: The HandshakeServer threw on a missing magic but HandshakeClient did not - it exceptionally completed its future and then happily proceeded to send messages to the server. Maybe throwing would be better here. I don't think coupling is an issue though - ensureMagic just knows it should do what it's told if the magic value has been received.

andrewkerr9000:
Needs squashing, an unused and arguably poorly named ModifierProtocolIdentifier `GRATUITOUS_OBFUSCATION` is present for testing purposes. Should be mergeable otherwise.

Snappy compression has been disabled. Will leave a commit removing it that can be reverted later.

Outstanding issues will be addressed when implementing the configuration and catch up protocol parts of the handshake story, and include:

* failing CoreToCoreCopySnapshotIT when snappy enabled
* preference order of modifier protocols for a given identifier: "latest" is not a useful guide

andrewkerr9000:
squashed

andrewkerr9000:
Protocols are described by pair of Identifier (enum/string) and Version (int).

For application protocols (Raft, Catchup) the version is an ordinal measure: a greater number is a more recent version of the algorithm, and would be preferred.

For modifier protocols (Compression) the version is nominal: it implies no ordering, only incompatibility. An implementation of Snappy is represented as (Compression, 2). An incompatible version of Snappy (e.g. with checksums) would be (Compression, y), and an entirely different algorithm such as LZH would be (Compression, z). Friendly names are included in the protocols in code and displayed in the protocol Cypher procedure, and will be used to define an order of preference in config.

andrewkerr9000:
Removed explicit timeout on IT - not necessary, and potentially makes the test flaky.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ModifierProtocolInstaller.java
```diff
@@ -0,0 +1,61 @@
...
+
+import static java.util.Arrays.asList;
+
+public interface ModifierProtocolInstaller
+{
+    Protocol.ModifierProtocol protocol();
+
+    void apply( NettyPipelineBuilder nettyPipelineBuilder );
+
+    // TODO using Snappy on Raft breaks CoreToCoreCopySnapshotIT.
```
martinfurmanski: Did you figure out why yet?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/handshake/HandshakeClient.java
```diff
@@ -54,6 +76,7 @@ private void ensureMagic()
         if ( !magicReceived )
         {
             decline( "Magic value not received." );
+            throw new IllegalStateException( "Magic value not received." );
```
martinfurmanski, line 79: Not part of your PR, but I wonder if the handler for the magic message below also should throw. Now it will decline(..) yet set magicReceived = true.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/Protocol.java
```diff
@@ -53,13 +63,68 @@ public String canonicalName()
...
+    enum ModifierProtocols implements ModifierProtocol
+    {
+        // TODO Snappy on Raft breaks CoreToCoreCopySnapshotIT
+        COMPRESSION_NONE( ModifierProtocolIdentifier.COMPRESSION, 1, "Uncompressed" );
+
+        private final int version;
+        private final ModifierProtocolIdentifier identifier;
+        private final String friendlyName;
+
+        ModifierProtocols( ModifierProtocolIdentifier identifier, int version, String friendlyName )
```
martinfurmanski, line 104: This is very much different than what was done for application protocols. Not sure version makes sense anymore as a name for this, or even identifier. Identifier seems to be a class of modifiers while the version decided the particular implementation. Will friendly name be sent over the wire? I think that is important for the application protocols and makes it possible to still print and complain on something reasonable on an instance which lacks support, instead of just "not supporting compression variant 3".

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ProtocolInstallerRepository.java
```diff
@@ -49,16 +58,44 @@ private void addTo( Map<Protocol,ProtocolInstaller<O>> tempServerMap, ProtocolIn
...
 
-    public ProtocolInstaller installerFor( Protocol protocol )
+    public ProtocolInstaller<O> installerFor( ProtocolStack protocolStack )
     {
-        ProtocolInstaller protocolInstaller = installers.get( protocol );
-        if ( protocolInstaller == null )
+        ApplicationProtocol applicationProtocol = protocolStack.applicationProtocol();
+        ProtocolInstaller.Factory<O,?> protocolInstaller = installers.get( applicationProtocol );
+
+        // TODO need at least one modifier to frame
```
martinfurmanski: I wonder if we actually always need the basic framing for POJOs. Snappy does some chunking and framing, but I suspect it might not work as we need it to. We need it to emit entire ByteBufs representing complete messages for the subsequent decoders. Perhaps this is the reason some tests are failing? From looking at the source code of SnappyFrameEncoder I think for example that sending a message above 32K will yield several ByteBuf chunks on the decoding side. I might be totally wrong of course as well.

martinfurmanski (commented):
Will continue looking tomorrow, but have a few comments already.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ProtocolInstallerRepository.java
andrewkerr9000: Mixing Snappy with basic framing breaks everything. I haven't tested Snappy with large messages.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ModifierProtocolInstaller.java
andrewkerr9000: Nope. I'm planning to revisit when I look at catchup.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/Protocol.java
andrewkerr9000: The model is a little odd. I think it makes sense to revisit when implementing configuration - these protocols need to be identified somehow.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ProtocolInstallerRepository.java
martinfurmanski: Hmm... it really shouldn't. I wonder if that is an issue of ordering in the pipeline? I think snappy should be the outermost layer and the basic framing within it, since snappy compresses a stream and the basic framing collects until it has a complete frame.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/handshake/ProtocolStack.java
```diff
@@ -19,24 +19,35 @@
...
 
-    public Protocol applicationProtocol()
+    public ApplicationProtocol applicationProtocol()
     {
         return applicationProtocol;
     }
 
+    public List<ModifierProtocol> modifierProtocols()
+    {
+        return Collections.unmodifiableList( modifierProtocols );
```
martinfurmanski: You could make it unmodifiable already in the constructor.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/handshake/ServerMessageDecoder.java
```diff
@@ -50,28 +52,38 @@ protected void decode( ChannelHandlerContext ctx, ByteBuf in, List<Object> out )
         }
         case 1:
```
martinfurmanski, line 53: We should extract these later.

martinfurmanski (approved):

andrewkerr9000:
* Split modifiers into client and server components.
* Reintroduced Snappy, with framing at the end of the pipeline.
* NettyPipelineBuilder has client/server type parameter.
* Remove no compression modifier.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/ClientNettyPipelineBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+{
+    ClientNettyPipelineBuilder( ChannelPipeline pipeline, Log log )
+    {
+        super( pipeline, log );
+    }
+
+    @Override
+    public ClientNettyPipelineBuilder addFraming()
+    {
+        add( "frame_encoder", new LengthFieldPrepender( 4 ) );
```
martinfurmanski, line 37: This will be problematic for catchup next, because it is bi-directional. Raft client is uni-directional which is why it just needs the encoders.

martinfurmanski (commented):
I wonder how this will play along with bidirectional protocols. Our working protocol so far (Raft) unfortunately is a bit odd in that it is unidirectional.

andrewkerr9000:
We'd need to add a duplex "orientation" alongside server & client. I doubt it would be challenging.

Issue #9992, opened 2017-09-05
allshortestPaths error (start/end nodes the same) with cypher.forbid_shortestpath_common_node=false


Alan-R:
I'm getting an error message when running the query below:

```
MATCH p = allshortestPaths( (start:Class_Drone)-[:tcpclient|:tcpservice|:hosting|:baseip|:nicowner|:monitoring|:wiredto|:ipowner*]-(m) )
        WHERE m.nodetype IN ['IPaddrNode', 'SystemNode', 'MonitorAction', 'ProcessNode', 'IPtcpportNode', 'Drone', 'NICNode']
        UNWIND nodes(p) AS n
        UNWIND rels(p) AS r
        RETURN [x in COLLECT(DISTINCT n) WHERE x.nodetype in ['IPaddrNode', 'SystemNode', 'MonitorAction', 'ProcessNode', 'IPtcpportNode', 'Drone', 'NICNode']] AS nodes,
        COLLECT(DISTINCT r) AS relationships
```
I have this line in my neo4j config file:
**cypher.forbid_shortestpath_common_node=false**

And I still get this message:
The shortest path algorithm does not work when the start and end nodes are the same. This can happen if you perform a shortestPath search after a cartesian product that might have the same start and end nodes for some of the rows passed to shortestPath. If you would rather not experience this exception, and can accept the possibility of missing results for those rows, disable this in the Neo4j configuration by setting `cypher.forbid_shortestpath_common_node` to false. If you cannot accept missing results, and really want the shortestPath between two common nodes, then re-write the query using a standard Cypher variable length pattern expression followed by ordering by path length and limiting to one result.

Syslog adds this phrase:
reference 78c83d45-4e40-4774-a6be-e347907e7ec9

FWIW, ***this query used to work in earlier (3.0.x) versions of Neo4j***. The intent of the query is to return the selected portions of the graph  as a subgraph. This particular one returns basically the entire graph, but other variations are more selective. They all get the same message...

Suggestions on how to work around this would be appreciated. I have to say that after reading the message, I'm not certain I understand it. It could be clearer. It's also pretty clear that being able to select and return a subgraph is a pretty important capability. If this isn't the right way, then tell me the right way...

### Operating system:
    "operating-system": "GNU/Linux",
    "machine": "x86_64",
    "processor": "x86_64",
    "hardware-platform": "x86_64",
    "kernel-name": "Linux",
    "kernel-release": "4.4.0-92-generic",
[...]

PR #10522, opened 2017-11-28
Fix typo: 'cypher.forbid_shortestpath_common_node'

Koenkk:
The configration option is called `cypher.forbid_shortestpath_common_nodes` (with an **s** at the end) not `cypher.forbid_shortestpath_common_node`. 

See #9992.

PR #10988, opened 2018-02-08
This is important, since the text of the message directs the user to do the wrong thing

Alan-R:
You can see others have commented on this - even though you closed the issue.

There is a typo in the commit message - it should have said #9992.


Alan-R:
Given that it just adds an 's' to a message, it's hard to see how the checks should all fail. Looks like a broken build system. Since I can't see what's wrong, I can't figure out if there is some kind of real problem (but I seriously doubt it).

chrisvest:
It's not a build system problem. The first thing the build does, is check that all committers are connected to contributor license agreements, and apparently this check fails.

craigtaverner:
We should rather merge the original PR at https://github.com/neo4j/neo4j/pull/10522, and that user has signed the CLA (although is not yet on the whitelist).

Alan-R:
I thought I had signed it before, but just in case I hadn't -- I sent the CLA to cla@neotechnology, and CCed @jexp for good measure.

sherfert:
Hi @Alan-R . It took as while, but now you are finally on our whitelist. I apologize for talking so so long to do that.

Unfortunately, your PR will need to be rebased for the CI checks to become green. We have changed our licence headers since then...

sherfert:
Since #10522 was merged, I am closing this PR.

PR #10989, opened 2018-02-09
More leaf operators and aggregation in morsel runtime

pontusmelke:
- Use cursors directly
- Support label scan
- Support index seek
- Don't force operators to be handled by single thread
- Some aggregation operators: `count`, `min`, `max`, `avg`, `collect`


community/values/src/main/java/org/neo4j/values/utils/ValueMath.java
```diff
@@ -0,0 +1,104 @@
...
+     */
+    public static NumberValue add( long a, long b )
+    {
+        try
+        {
+            return Values.longValue( Math.addExact( a, b ) );
+        }
+        catch ( ArithmeticException e )
+        {
+            return Values.doubleValue( (double) a + (double) b );
```
fickludd: Even though it shouldn't be too frequent, is there anyway to do this without using exceptions as a code path?

fickludd (changes requested):
I merely graced over the morsel stuff, so I cannot really determine the quality of that. I did look properly at the Values maths methods though, and those look pretty good to me. I'm wondering though if it wouldn't be enough to have specialized operators (`plus`, `minus`, `times`, `etc`) only for the `long` and `double` primitives? As you only implemented maths for those anyway, I don't really see any benefit in having the smaller primitives.

pontusmelke:
So I was going to explain how you're wrong and that using the intrinsic `addExact` et al would be so much faster for the normal case that it is totally worth taking the extra hit in the rare cases that we get an overflow. However actual numbers seems to disagree with that hypothesis - and the performance drop was steeper than I expected (and the payoff was not what I would have hoped for).

Running benchmarks with no overflow and 1% overflow

```
ExactMath.check       0 1544.297 ±  52.957   ops/ms
ExactMath.check       1 1214.406 ± 106.689  ops/ms
ExactMath.intrinsic   0 1526.825 ±  75.823  ops/ms
ExactMath.intrinsic   1   86.094 ±   8.380  ops/ms
```
 So yes you're right.

For your other question we still need to be able to handle `byte`, `short`, `int`, and `float` since they might be coming from parameters, e.g.

```
MATCH (n) RETURN avg(n.prop)
```

The plan is that this should replace what is now done in `OverFlowAwareSum`,  `Add`, `TypeSafeMathSupport` and probably a couple of other places.

fickludd (approved):

fickludd:
Wow, I'm impressed that you benchmarked it, I was just guessing =).

Ok, let's keep the all the specializations as you already coded them. We'll see later if they're needed or not.

fickludd:
@pontusmelke I think this looks good now, but it needs a little love.

pontusmelke:
@fickludd I gave it some more love. I realized that having the widening behavior by default is wrong and not how cypher works today. Currently we use strict math by default and only use the widening to double in `AVG`, so I switched to that so that we can replace `TypeSafeMathSupport` with these method (in a separate PR).  If we want the automatic widening on for all additions I think we need a `BigIntegerValue` or something which can be backed by a `BigInteger` so that we can guarantee that `a+b` is never lossy for integers.

enterprise/cypher/morsel-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/vectorized/PipelineBuilder.scala
```diff
@@ -102,6 +103,26 @@ class PipelineBuilder(slotConfigurations: SlotConfigurations, converters: Expres
...
           new MergeSortOperator(ordering, slots)
 
+        case plans.Aggregation(_, groupingExpressions, aggregationExpression) if groupingExpressions.isEmpty =>
+          val (in, out) = aggregationExpression.foldLeft[(Seq[MapperOffsets], Seq[ReducerOffsets])]((Seq.empty, Seq.empty)) {
+            case ((maps, reds), (key, expression)) =>
+              val aggregator = converters.toCommandExpression(expression).asInstanceOf[AggregationExpressionOperator]
+              val currentSlot = slots.get(key).get
+              //we need to make room for storing aggregation value in
+              //source slot
+              source.slots.newReference(key, currentSlot.nullable, currentSlot.typ)
```
systay: Hmmm... If I understand this correctly, what we'll do is to use the input morsel first to calculate the aggregate values, and then we write those values back into the same morsel. Could we not instead break the pipeline here? From what I can see, the result will be that the input-morsel will be wider than necessary (since the aggregation values cannot be written until after all the input has been consumed), and for the output, we only use the last ref-slots of the first row.

It should work but seems inefficient.

enterprise/cypher/morsel-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/vectorized/expressions/MinOrMaxOperatorExpression.scala
```diff
@@ -0,0 +1,104 @@
...
+
+class MaxReducer extends AggregationReducer with MaxChecker
+
+
+
+
+
+
+
+
```
systay: This file should go on a diet.

enterprise/cypher/morsel-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/vectorized/operators/AggregationMapperOperatorTest.scala
```diff
@@ -0,0 +1,175 @@
...
+    // Given
+    val groupSlot1 = RefSlot(0, nullable = false, CTAny)
+    val groupSlot2 = RefSlot(1, nullable = false, CTAny)
+    val slots = new SlotConfiguration(mutable.Map("node" -> LongSlot(0, nullable = false, CTNode),
+                                                  "aggregate" -> RefSlot(2, nullable = false, CTAny),
+                                                  "group1" -> groupSlot1, "group2" -> groupSlot2), 1, 3)
+    val aggregation = new AggregationMapperOperator(slots,
+                                                    Array(AggregationOffsets(2, 2, DummyEvenNodeIdAggregation(0))),
+                                                    Array(GroupingOffsets(groupSlot1, groupSlot1,
+                                                                          new DummyExpression(stringValue("A"), stringValue("B"))),
```
systay, line 70: The line width here is itching. Could we indent things differently maybe?

enterprise/cypher/morsel-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/vectorized/PipelineBuilder.scala
```diff
@@ -88,6 +103,45 @@ class PipelineBuilder(slotConfigurations: SlotConfigurations, converters: Expres
           source = source.addOperator(preSorting)
           new MergeSortOperator(ordering, slots)
 
+        case plans.Aggregation(_, groupingExpressions, aggregationExpression) if groupingExpressions.isEmpty =>
```
systay, line 106: Hmmm... If I understand this correctly, what we'll do is to use the input morsel first to calculate the aggregate values, and then we write those values back into the same morsel. Could we not instead break the pipeline here? From what I can see, the result will be that the input-morsel will be wider than necessary (since the aggregation values cannot be written until after all the input has been consumed), and for the output, we only use the last ref-slots of the first row.

It should work but seems inefficient.

(Same thing for both these cases)

fickludd: @pontusmelke @systay If you two agree on this, please make a card to remember that optimization. I will merge this PR now before it gets stale.

pontusmelke: 👍

PR #10992, opened 2018-02-09
Store DateTimes with ZoneIds

sherfert:
This implements the storage of DateTimes with ZoneIds for both scalar
and array values. It stores ZoneIds as shorts, and uses a hardcoded
table to map between String and short.

changelog: Any temporal value can be stored in a property. This includes Date, Time, DateTime, LocalTime, LocalDateTime, and Duration.


community/kernel/src/main/java/org/neo4j/kernel/impl/store/TimeZoneMapping.java
```diff
@@ -0,0 +1,664 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.store;
+
+import io.netty.util.collection.ShortObjectHashMap;
+import io.netty.util.collection.ShortObjectMap;
```
thobe: These are the wrong collections to use here.

community/kernel/src/test/java/org/neo4j/kernel/impl/core/TestPropertyTypes.java
```diff
@@ -595,8 +595,21 @@ public void testDateTimeArrayWithZoneOffset()
...
+     * Update {@link TimeZoneMapping#TIME_ZONE_SHORT_TO_STRING} if this test fails.
+     * To do that, run {@link ZoneId#getAvailableZoneIds()} and add new time zones
+     * at the end.
+     *
+     * After that, change the test to expect the new Java version.
+     */
+    @Test
+    public void timeZoneTokenListIsUpToDate()
+    {
+        assertEquals( "When updating to a new Java version, make sure to update to time zone mappings.",
```
thobe: This is insufficient. Since it is possible for end users to update the Time Zone Database in their JRE, we really ought to update the time zone mapping every time there is a new release of the IANA Time Zone Database.

community/kernel/src/test/java/org/neo4j/kernel/impl/core/TestPropertyTypes.java
```diff
@@ -595,8 +595,21 @@ public void testDateTimeArrayWithZoneOffset()
...
+                "1.8", Runtime.class.getPackage().getSpecificationVersion() );
```
thobe: This assertion is insufficient. We should assert on the actual table of tzids, since even official JRE releases will update the Time Zone Database on patch releases.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TimeZoneMapping.java
```diff
@@ -0,0 +1,664 @@
...
+        sToI.put( putAndReturn( iToS, ++i, "Australia/Lindeman" ), i);
+        sToI.put( putAndReturn( iToS, ++i, "America/Los_Angeles" ), i);
+        sToI.put( putAndReturn( iToS, ++i, "SystemV/EST5EDT" ), i);
+        sToI.put( putAndReturn( iToS, ++i, "Pacific/Majuro" ), i);
+        sToI.put( putAndReturn( iToS, ++i, "America/Argentina/Buenos_Aires" ), i);
+        sToI.put( putAndReturn( iToS, ++i, "Europe/Nicosia" ), i);
+        sToI.put( putAndReturn( iToS, ++i, "Pacific/Guadalcanal" ), i);
+        sToI.put( putAndReturn( iToS, ++i, "Europe/Athens" ), i);
+        sToI.put( putAndReturn( iToS, ++i, "US/Pacific" ), i);
+        sToI.put( putAndReturn( iToS, ++i, "Europe/Monaco" ), i);
```
thobe: It really irks me to have this defined in code. I'd much prefer to have this defined in a resource file that is read when this class loads.

Keep in mind that we also need to define mappings for tzids that have been removed from the tz database.

thobe (changes requested):
There are a few issues with this. But most importantly I'm missing a strategy for how the mapping of tzids is supposed to be kept up to date.

I have written up a short document on the things to keep in mind whit this sort of mapping: https://docs.google.com/document/d/1FS0V7B3oj09xOb23KXNXHufkjBjZxedcIeq2Qtv4pbs/edit?usp=sharing

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TimeZoneMapping.java
```diff
@@ -0,0 +1,664 @@
...
+     */
+    private TimeZoneMapping()
+    {
+    }
+
+    /**
+     * This provides a mapping from short to string for all
+     * supported time zone names.
+     */
+    public static final Map<Short,String> TIME_ZONE_SHORT_TO_STRING;
```
tinwelint: Probably better with something like a [PrimitiveIntObjectMap](https://github.com/neo4j/neo4j/blob/3.4/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/Primitive.java#L181)

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TemporalType.java
```diff
@@ -469,7 +468,19 @@ public static Value decode( long[] valueBlocks, int offset )
...
     public static long[] encodeDateTime( int keyId, long epochSecond, long nanoOfSecond, String zoneId )
     {
-        throw new UnsupportedOperationException( "Cannot yet store DateTime with ZoneID" );
+        int idBits = StandardFormatSettings.PROPERTY_TOKEN_MAXIMUM_ID_BITS;
+        short zoneNumber = TimeZoneMapping.TIME_ZONE_STRING_TO_SHORT.get( zoneId );
+
+        long keyAndType = keyId | (((long) (PropertyType.TEMPORAL.intValue()) << idBits));
+        long temporalTypeBits = TemporalType.TEMPORAL_DATE_TIME.temporalType << (idBits + 4);
+
+        long[] data = new long[3];
```
tinwelint: Has this `3` any relation to the one in `calculateNumberOfBlocksUsedForTemporal`? If so, or perhaps even if not, then introducing constants is a good idea and really aids readers of this code.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TemporalType.java
```diff
@@ -244,22 +245,18 @@ public Value decodeForTemporal( long[] valueBlocks, int offset )
...
-                    if ( storingZoneOffset( firstBlock ) )
-                    {
-                        return 3;
-                    }
-                    else
-                    {
-                        // TODO proper number
-                        return 3;
-                    }
+                    return 3;
```
tinwelint: A little outline of what's in these 3 blocks as a comment would be dandy

tinwelint (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TimeZoneMapping.java
sherfert: Which ones would be the right ones?

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TemporalType.java
sherfert: I will put that into the comment in PropertyStore, where all store layouts are summarized. Thanks for the reminder!

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TimeZoneMapping.java
```diff
@@ -0,0 +1,111 @@
...
+                {
+                    TIME_ZONE_SHORT_TO_STRING.add( line );
+                    TIME_ZONE_STRING_TO_SHORT.put( line, (short) (TIME_ZONE_SHORT_TO_STRING.size() - 1) );
+                }
+            }
+            LATEST_SUPPORTED_IANA_VERSION = latestVersion;
+        }
+        catch ( IOException e )
+        {
+            throw new RuntimeException( "Time zone id file is missing." );
```
thobe: Not the only reason why we could get an exception here. I'd reword this to just `"Failed to read time zone id file."`.

community/kernel/src/main/resources/TZIDS
```diff
@@ -0,0 +1,612 @@
...
+Europe/Kirov
+# tzdata2016g
+Asia/Yangon
+# tzdata2016i
+Asia/Famagusta
+# tzdata2016j
+Asia/Atyrau
+Europe/Saratov
+# tzdata2017a
+America/Punta_Arenas
```
thobe: I think there should be a newline at the end. Even if that means that the reader code needs to know how to skip blank lines.

community/kernel/src/main/resources/TZIDS
```diff
@@ -0,0 +1,612 @@
...
+Australia/Lindeman
+America/Los_Angeles
+SystemV/EST5EDT
+Pacific/Majuro
+America/Argentina/Buenos_Aires
+Europe/Nicosia
+Pacific/Guadalcanal
+Europe/Athens
+US/Pacific
+Europe/Monaco
```
thobe, line 587: _This is not a suggestion, it is a question._

For the initial chunk - the identifiers before `tzdata2014b` - how about we order them in a way that makes this file easier for humans?

The argument for keeping it unordered would be that it would not tempt maintainers to try to insert new entries "in order", in stead of at the end as they should.
The argument for having an order would be that it makes debugging easier.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/TimeZoneMappingTest.java
```diff
@@ -0,0 +1,179 @@
...
+        catch ( IllegalArgumentException e )
+        {
+            fail( "Our time zone table does not support Canada/East-Saskatchewan" );
+        }
+    }
+
+    @Test
+    public void updateTimeZoneMappings() throws IOException
+    {
+        String foo = "https://www.iana.org/time-zones/repository/tzdata-latest.tar.gz";
```
thobe: There is an alternative URL for this under the `data` domain that we might want to use instead: `https://data.iana.org/time-zones/tzdata-latest.tar.gz`.

thobe: We should give this variable a better name than `foo`.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/TimeZoneMappingTest.java
```diff
@@ -0,0 +1,179 @@
...
+        List<String> versionsToUpgrade = new ArrayList<>();
+
+        try ( TarArchiveInputStream tar = new TarArchiveInputStream( new GZIPInputStream( new URL( foo ).openStream() ) ) )
```
thobe: I'd like to factor this code into a some helper method:
```
void readTZDBFile( String release, String file, Predicate<String> consumer ) throws IOException
{
    String uri = release == null
            ? "https://data.iana.org/time-zones/tzdata-latest.tar.gz"
            : "https://data.iana.org/time-zones/releases/tzdata" + release + ".tar.gz";
    try ( TarArchiveInputStream tar = new TarArchiveInputStream( new GZIPInputStream( new URL( uri ).openStream() ) ) )
    {
        for ( ArchiveEntry entry; (entry = tar.getNextEntry()) != null; )
        {
            if ( file.equalsIgnoreCase( entry.getName() ) )
            {
                try ( BufferedReader reader = new BufferedReader( new InputStreamReader( tar ) ) )
                {
                    for ( String line; (line = reader.readLine()) != null; )
                    {
                        if ( !consumer.test( line ) )
                        {
                            break;
                        }
                    }
                }
                break;
            }
        }
    }
}
```

Then our two usages of this pattern would become:
```
readTZDBFile( null, "NEWS", line ->
{
    if ( line.startsWith( "Release " ) )
    {
        String substring = line.substring( line.indexOf( ' ' ) + 1 );
        String release = substring.substring( 0, substring.indexOf( ' ' ) );
        if ( TimeZoneMapping.LATEST_SUPPORTED_IANA_VERSION.equals( release ) )
        {
            return false; // stop reading
        }
        versionsToUpgrade.add( release );
    }
    return true; // continue reading
} );
```
and:
```
readTZDBFile( release, "zone1970.tab", line ->
{
    if ( !line.startsWith( "#" ) )
    {
        ianaSupportedTzs.add( line.split( "\\t" )[2] );
    }
    return true; // read all lines
} );
```

Which is much easier on the eyes (even if it does need the comments for explaining how the return value is used for control flow).

community/kernel/src/main/java/org/neo4j/kernel/impl/store/TimeZoneMapping.java
```diff
@@ -0,0 +1,111 @@
...
+{
+    /**
+     * Prevent instance creation.
+     */
+    private TimeZoneMapping()
+    {
+    }
+
+    private static final List<String> TIME_ZONE_SHORT_TO_STRING = new ArrayList<>( 1024 );
+    private static final Map<String,Short> TIME_ZONE_STRING_TO_SHORT = new HashMap<>( 1024 );
```
thobe, line 43: I'd like to change `TIME_ZONE_SHORT_TO_STRING` to just be an array, and I think we can introduce a lower overhead data structure for `TIME_ZONE_STRING_TO_SHORT`, but those are future improvements that can be made without breaking any functionality.

thobe (approved):
I think this is good enough. There are still some questions I'd like to discuss before this is merged, but overall this is good enough.

There are also some things that can be further improved, but that doesn't need to stop this from going in.

community/kernel/src/main/resources/TZIDS
sherfert: It does already know how to do that

tinwelint (approved):

community/kernel/src/main/resources/TZIDS
sherfert: I have no strong preference for one or the other.

community/kernel/src/test/java/org/neo4j/kernel/impl/store/TimeZoneMappingTest.java
sherfert: Agreed

PR #10994, opened 2018-02-09
Fix bug where shortestPath unnecessarily required named nodes

sherfert:
This fixes it for 3.4. I did that first then realized I should be doing it in 3.3 instead. So now there"s 2 PRs.

### DIFF
--- a/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathAcceptanceTest.scala
+++ b/enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathAcceptanceTest.scala
@@ -743,6 +743,18 @@ class ShortestPathAcceptanceTest extends ExecutionEngineFunSuite with CypherComp
     result.toList should equal(List(Map("length(path)" -> 2)))
   }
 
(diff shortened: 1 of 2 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathAcceptanceTest.scala
```diff
@@ -743,6 +743,18 @@ class ShortestPathAcceptanceTest extends ExecutionEngineFunSuite with CypherComp
...
+  test("should not require named nodes in shortest path") {
+    graph.execute("CREATE (a:Person)-[:WATCH]->(b:Movie)")
+
+    val query =
+      """MATCH p=shortestPath( (:Person)-[*1..4]->(:Movie) )
+        |RETURN length(p)
+      """.stripMargin
+    val result = graph.execute(query)
+
+    println(result.next())
```
Lojjs: Same as the other PR, you have forgotten a debug line here and I think it would be nice with an assertion of the result

Lojjs (changes requested):
You have a `println` statement in your commited code, please remove

Lojjs (approved):

PR #10995, opened 2018-02-09
Fix bug where shortestPath unnecessarily required named nodes

sherfert:
Null forward merge, since we have a corresponding PR for 3.4


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/ShortestPathAcceptanceTest.scala
```diff
@@ -740,6 +740,18 @@ class ShortestPathAcceptanceTest extends ExecutionEngineFunSuite with CypherComp
...
+  test("should not require named nodes in shortest path") {
+    graph.execute("CREATE (a:Person)-[:WATCH]->(b:Movie)")
+
+    val query =
+      """MATCH p=shortestPath( (:Person)-[*1..4]->(:Movie) )
+        |RETURN length(p)
+      """.stripMargin
+    val result = graph.execute(query)
+
+    println(result.next())
```
Lojjs: It looks like you forgot to remove this debug line! Maybe you should also do some kind of assertion on the result?

Lojjs (changes requested):
You have a `println` in the commited code, please remove

Lojjs (approved):

PR #10996, opened 2018-02-09
Implements specific SimpleIndexReader#query method

tinwelint:
Instead of relying on the default wrapping found in AbstractIndexReader.
This removes unnecessary indirection in SimpleIndexReader.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/api/schema/BridgingIndexProgressor.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/api/schema/BridgingIndexProgressor.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.kernel.impl.api.schema;
+
+import java.util.Queue;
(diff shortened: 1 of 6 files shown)


tinwelint:
An alternative approach to #10975 that doesn't introduce an additional `IndexReader#query` method.

pontusmelke (commented):
👍

community/kernel/src/main/java/org/neo4j/kernel/impl/api/schema/BridgingIndexProgressor.java
```diff
@@ -0,0 +1,106 @@
...
+ * Combine multiple progressor to act like one single logical progressor seen from client's perspective.
+ */
+public class BridgingIndexProgressor implements IndexProgressor.NodeValueClient, IndexProgressor
+{
+    private final NodeValueClient client;
+    private final int[] keys;
+    // This is a thread-safe queue because it can be used in parallel scenarios.
+    // The overhead of a concurrent queue in this case is negligable since typically there will be two or a very few number
+    // of progressors and each progressor has many results each
+    private final Queue<IndexProgressor> progressors = new ConcurrentLinkedQueue<>();
```
burqen, line 40: What is the parallel scenario?

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/DocValuesCollector.java
```diff
@@ -312,6 +320,92 @@ private void replayTo( Collector collector ) throws IOException
...
+                    if ( nextDoc != DocIdSetIterator.NO_MORE_DOCS )
+                    {
+                        index++;
+                        next = currentDocValues.get( nextDoc );
+                        return true;
+                    }
+                    else
+                    {
+                        currentIdIterator = null;
+                        return fetchNextEntityId();
```
burqen, line 396: Is this recursive call really necessary? If it is, that's fine, but I would prefer something that is easier to reason about.

burqen (changes requested):
Nice getting rid of that indirection! Looks good overall but got some questions.

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/index/collector/DocValuesCollector.java
tinwelint: This isn't changed though... so something for another PR/discussion?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/schema/BridgingIndexProgressor.java
tinwelint: Querying a lucene index that has multiple partitions is done in parallel.

burqen (approved):

PR #10997, opened 2018-02-09
Various little fixes of code smells and bugs suggested by sonarQube

SaschaPeukert:


community/cypher/cypher/src/main/java/org/neo4j/cypher/export/SubGraphExporter.java
```diff
@@ -90,12 +90,13 @@ private void output( PrintWriter out, String ... commands )
...
                     throw new IllegalStateException( "Indexes should have at least one property key" );
                 }
-                String key = quote( propertyKeys.next() );
+                String id = propertyKeys.next();
                 if ( propertyKeys.hasNext() )
                 {
                     throw new RuntimeException( "Exporting compound indexes is not implemented yet" );
                 }
 
+                String key = quote( id );
```
thobe, line 99: I don't understand what this improves. Sure, it doesn't hurt, but I'm curious to learn what it is supposed to improve.

community/cypher/cypher/src/main/java/org/neo4j/cypher/export/SubGraphExporter.java
```diff
@@ -106,8 +107,21 @@ private void output( PrintWriter out, String ... commands )
...
+            count = ((Collection<ConstraintDefinition>) constraints).size();
+        }
+        else
+        {
+            for ( ConstraintDefinition ignored : constraints )
+            {
+                count++;
+            }
+        }
+        final List<String> result = new ArrayList<>( count );
```
thobe, line 123: This is quite a bit more code in order to only be able to pre-size the collection. Is this method really hot enough to warrant it?

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
```diff
@@ -765,7 +765,9 @@ public Object next()
             {
                 throw new NoSuchElementException();
             }
-            return Array.get( array, position++ );
+            Object result = Array.get( array, position );
+            position++;
+            return result;
```
thobe: In what way is this an improvement?

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/ParameterConverter.java
```diff
@@ -239,8 +238,8 @@ else if ( obj instanceof Path )
             {
                 return new Iterator<PropertyContainer>()
                 {
-                    Iterator<? extends PropertyContainer> current = nodes().iterator();
-                    Iterator<? extends PropertyContainer> next = relationships().iterator();
+                    private Iterator<? extends PropertyContainer> current = nodes().iterator();
+                    private Iterator<? extends PropertyContainer> next = relationships().iterator();
```
thobe: This is just more typing. Not that it matters much, but I fail to see the point.

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/tracing/TimingCompilationTracer.java
```diff
@@ -117,27 +117,27 @@ final Clock clock()
 
     private static class Query extends Event implements QueryEvent, QueryCompilationEvent
     {
-        private final String query;
+        private final String actualQuery;
```
thobe: Why?

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/tracing/TimingCompilationTracer.java
```diff
@@ -163,24 +163,24 @@ public String query()
 
     private static class Phase extends Event implements PhaseEvent, CompilationPhaseTracer.CompilationPhaseEvent
     {
-        private final CompilationPhaseTracer.CompilationPhase phase;
+        private final CompilationPhaseTracer.CompilationPhase compilationPhase;
```
thobe, line 166: Why?

enterprise/cypher/slotted-runtime/src/main/java/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/DefaultComparatorTopTable.java
```diff
@@ -126,7 +126,8 @@ public T next()
                 {
                     throw new NoSuchElementException();
                 }
-                return (T) array[--cursor];
+                --cursor;
+                return (T) array[cursor];
```
thobe: I don't see how this is an improvement.

thobe (commented):
Most of these changes are completely unnecessary. Some just make the code more verbose, and one or two to the point where the code becomes harder to follow.

There is a bit of good cleanup mixed in as well, but in many cases it seems to be about only imposing some personal preferences on how to format code.

community/cypher/cypher/src/main/java/org/neo4j/cypher/export/SubGraphExporter.java
SaschaPeukert: The reason for that was "Move declaration closer to actual usage". Its usefulness is debatable, you are right. Say the word and I kill it

community/cypher/cypher/src/main/java/org/neo4j/cypher/export/SubGraphExporter.java
SaschaPeukert: I was asking myself the same question. SonarQube is referring to not pre-sized collections as a performance bug so I wanted to discuss this one

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
SaschaPeukert: Let me show you what sonarQube has to say about this one:

"The use of increment and decrement operators in method calls or in combination with other arithmetic operators is not recommended, because:

- It can significantly impair the readability of the code.
- It introduces additional side effects into a statement, with the potential for undefined behavior.
- It is safer to use these operators in isolation from any other arithmetic operators."

To me personally, it feels like it improves readability a (tiny) bit

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/ParameterConverter.java
SaschaPeukert: "Failing to explicitly declare the visibility of a member variable could result it in having a visibility you don't expect, and potentially leave it open to unexpected modification by other classes."

I do not need it, but it might be nice for people less experienced with java defaults? Say the word and I kill it.

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/tracing/TimingCompilationTracer.java
SaschaPeukert: "It's confusing to have a class member with the same name (case differences aside) as its enclosing class. This is particularly so when you consider the common practice of naming a class instance for the class itself."

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/tracing/TimingCompilationTracer.java
SaschaPeukert: See comment on Query

enterprise/cypher/slotted-runtime/src/main/java/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/DefaultComparatorTopTable.java
SaschaPeukert: Same reason as the increment operator in CompiledConversionUtils

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/ParameterConverter.java
thobe: Since it is an anonymous class, the type itself is not visible, and thus the fields are not visible.

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
thobe: I feel the grouping of the three things (access from array, increment, return) is wrong in this new version, making it less clear than it was before. Perhaps it would be clearer written as:
```
int offset = position++;
return Array.get( array, offset );
```

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/tracing/TimingCompilationTracer.java
thobe: I don't think there was any confusion here. Although, I don't have a strong opinion that it must be `query`, I do however think that `actualQuery` is the wrong name. `queryString` would be better.

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/tracing/TimingCompilationTracer.java
thobe: Again, I don't think there was confusion in this case. I don't oppose the change though, just questioning why we are wasting our time with such irrelevant details.

enterprise/cypher/slotted-runtime/src/main/java/org/neo4j/cypher/internal/compatibility/v3_4/runtime/slotted/DefaultComparatorTopTable.java
thobe: I think the new version is slightly more confusing. A less confusing middle-ground would be:
```
int offset = --cursor;
return (T) array[offset];
```

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledConversionUtils.java
SaschaPeukert: But wouldn't that access the wrong position (+1), thus we would need to access first and then increment?

sherfert:
I know that many of these changes can be viewed from opposing angles. But this PR has been around for almost three weeks. It would be nice if you @thobe and @SaschaPeukert could find a state where the changes can get accepted and merged.

PR #10998, opened 2018-02-09
Migrate tests to JUnit 5

MishaDemianenko:
This is the first commit in a series of those to switch neo4j to use latest junit
for tests execution.
More precise description to come...


MishaDemianenko:
Modules will be migrated independently instead to make PRs reviewable and readable

PR #10999, opened 2018-02-10
Fix race with channel interceptor

martinfurmanski:
Remove the ChannelInterceptor concept since it potentially left a gap between connect()
and adding the interceptor where messages could slip by without being intercepted. Instead
the complete handshake pipeline is installed immediately from the ChannelInitializer and
the NettyPipelineBuilder has been extended with functionality to cater for the message gating.
    
The message gate is now a free-standing handler of its own and which does as little as
possible to fulfil its role. Closing and flusing logic has been moved outside to the
components which drive the handshake.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/ReconnectingChannel.java
```diff
@@ -129,18 +126,18 @@ public boolean isOpen()
...
     @Override
-    public CompletableFuture<Void> writeAndFlush( Object msg )
+    public Future<Void> writeAndFlush( Object msg )
     {
-        return doWrite( msg, ChannelOutboundInvoker::writeAndFlush );
+        return write( msg, true );
     }
 
-    private CompletableFuture<Void> doWrite( Object msg, BiFunction<io.netty.channel.Channel, Object, Future<Void>> writer )
+    private Future<Void> write( Object msg, boolean flush )
```
andrewkerr9000, line 145: Is this a Java Future? It is much less powerful than a CompletableFuture or a NettyFuture.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/NettyPipelineBuilder.java
```diff
@@ -27,89 +27,153 @@
...
+    {
+        newHandlers.stream().map( handler -> new HandlerInfo( name, fixed, handler ) ).forEachOrdered( handlers::add );
         return this;
     }
 
-    public NettyPipelineBuilder add( ChannelHandler... newHandlers )
+    /**
+     * @see #add(String, boolean, List)
+     */
+    public NettyPipelineBuilder add( String name, boolean fixed, ChannelHandler... newHandlers )
```
andrewkerr9000: This method could be private as overrides for both `fixed` permutations are provided

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/NettyPipelineBuilderFactory.java
```diff
@@ -39,9 +39,10 @@ public NettyPipelineBuilder create( Channel channel, Log log ) throws Exception
     {
         ChannelPipeline pipeline = channel.pipeline();
         NettyPipelineBuilder builder = NettyPipelineBuilder.with( pipeline, log );
+        int i = 0;
         for ( ChannelHandler handler : wrapper.handlersFor( channel ) )
         {
-            builder.add( handler );
+            builder.add( String.format( "%s_%d", wrapper.name(), i ), handler );
```
andrewkerr9000, line 45: i always equals 0?

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/HandshakeClientGate.java
```diff
@@ -20,59 +20,69 @@
...
+        if ( handshakePromise.isDone() || msg instanceof ServerMessage )
+        {
+            ctx.write( msg, promise );
+        }
+        else
+        {
+            handshakePromise.whenComplete( ( ignored, failure ) -> {
+                if ( failure == null )
+                {
+                    channel.writeAndFlush( msg, promise );
```
andrewkerr9000: Should this be calling a method on channel or on ctx? Should we be flushing from within a write method? Should we also override the flush method on ChannelDuplexHandler?

andrewkerr9000 (commented):
Looks good. I do have some questions on HandshakeClientGate.

You've replaced a CompletableFuture with a Future, but as all we do is block on it that would be an acceptable simplification.

There will be conflicts with the modifier protocol work..

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/ReconnectingChannel.java
martinfurmanski: Yes! Sorry for removing that, but I felt there was no need for it anymore. It removes the object creation.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/NettyPipelineBuilder.java
martinfurmanski: That is true, but I guess I view it as providing a nice and flexible API.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/protocol/NettyPipelineBuilderFactory.java
martinfurmanski: Will fix!

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/HandshakeClientGate.java
martinfurmanski: I actually want to remake this a bit (later) so that the messages are queued up on an actual queue and then the entire thing has a bound and gets drained completely with writes and a single flush.

I will take another look at channel vs. ctx and clarify with a comment which is correct and why considering the pipeline directions and the position of the gate. It looks potentially dubious to me as it is.

martinfurmanski: I figured out what the issue here is. The "fixed" handler concept in the pipeline builder has really crappy semantics, since you really have no idea about which position they get on a reinstall. Right now when installing the handshaking it is added to end the which is the correct thing to gate the outbound (last->first), but after a reinstall it will have moved relative position to the beginning and thus channel is needed instead of context to get the queued messages through the pipeline. Of course the issue is with the design of the "fixed" handler concept. I will try to come up with something at least a tad bit more robust, perhaps tail/head fixed handlers. Ideas welcome.

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/messaging/ReconnectingChannel.java
```diff
@@ -151,32 +151,67 @@ public boolean isOpen()
...
     }
 
+    /**
+     * Chains a channel future to a promise. Used when the returned promise
+     * was not allocated through the channel and cannot be used as the
+     * first-hand promise for the I/O operation.
+     */
+    private static void chain( ChannelFuture when, Promise<Void> then )
+    {
+        when.addListener( f -> {
```
andrewkerr9000, line 189: Code style won't like that {

martinfurmanski: Oddly enough, it accepts this!

andrewkerr9000 (approved):

PR #11009, opened 2018-02-14
Fixes as issue with applying index updates to unique indexes

tinwelint:
Given a uniqueness contraint on Foo:bar. If a transaction make changes
such that in the end, the constraint is not violated but some intermediate
individual step violates the constraint the index would still throw an
exception.

The reason this happens is that native index validates the uniqueness
constraint on every insert (since doing so is for free).

Although actually doing so when applying online updates has never been done
in the lucene indexes because all that is handled in the kernel layer
before committing a transaction anyway. Detecting a conflict while applying
updates to a unique index means that there's a bug inside the kernel layer
and would result in a kernel panic.

Therefore we avoid this problem and do what we do for lucene indexes,
namely by not checking conclicts when applying online updates.


tinwelint:
Marked as NOT READY since there are some problems with external stress tests, investigating.

MishaDemianenko (approved):
There is one relative test failure in a test that should be updated because of conflict resolution behavior change; approved otherwise.

PR #11010, opened 2018-02-14
Implement new VM pause monitor

andreikoval:


community/common/src/main/java/org/neo4j/util/Preconditions.java
```diff
@@ -0,0 +1,47 @@
...
+package org.neo4j.util;
+
+public final class Preconditions
+{
+    private Preconditions()
+    {
+        // util class
+    }
+
+    public static long requireNonNegative( long value )
```
MishaDemianenko: would be really nice to get unit tests for preconditions

community/common/src/main/java/org/neo4j/util/Preconditions.java
```diff
@@ -0,0 +1,47 @@
...
+    {
+        if ( value < 0 )
+        {
+            throw new IllegalArgumentException( "Expected non-negative integer value, got " + value );
```
MishaDemianenko: messages claims that integers were expected while they where checking longs actually; some rewording would be nice

community/common/src/main/java/org/neo4j/util/Preconditions.java
```diff
@@ -0,0 +1,47 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.neo4j.util;
+
+public final class Preconditions
```
MishaDemianenko: and some javadocs would be nice

community/kernel/src/main/java/org/neo4j/kernel/monitoring/VmPauseMonitor.java
```diff
@@ -0,0 +1,176 @@
...
+import static java.lang.System.nanoTime;
+import static java.lang.Thread.currentThread;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+import static org.neo4j.util.Preconditions.requirePositive;
+
+public final class VmPauseMonitor
+{
+    private final long measurementDurationnNs;
```
MishaDemianenko: Durationn

community/kernel/src/main/java/org/neo4j/kernel/monitoring/VmPauseMonitor.java
```diff
@@ -0,0 +1,176 @@
...
+    private final long stallAlertThresholdNs;
+    private final Log log;
+    private final Consumer<VmPauseInfo> listener;
+    private final Thread worker;
+
+    public VmPauseMonitor( long measureIntervalMs, long stallAlertThresholdMs, Log log, Consumer<VmPauseInfo> listener )
+    {
+        this.measurementDurationnNs = MILLISECONDS.toNanos( requirePositive( measureIntervalMs ) );
```
MishaDemianenko: in a component level above we convert setting duration to millis and now here we immediately do another conversion to nanos, maybe pass duration from outside and do single conversion here?

community/kernel/src/main/java/org/neo4j/kernel/monitoring/VmPauseMonitor.java
```diff
@@ -0,0 +1,176 @@
...
+        this.stallAlertThresholdNs = MILLISECONDS.toNanos( requirePositive( stallAlertThresholdMs ) );
+        this.log = requireNonNull( log );
+        this.listener = requireNonNull( listener );
+        this.worker = new Thread( this::run );
```
MishaDemianenko: we have Neo4jJobScheduler that is responsible for all of the threads, so let's use that instead for thread things

community/kernel/src/main/java/org/neo4j/kernel/monitoring/VmPauseMonitor.java
```diff
@@ -0,0 +1,176 @@
...
+    public void start()
+    {
+        log.debug( "Starting VM pause monitor" );
+        worker.start();
+    }
+
+    public void stop()
+    {
+        log.debug( "Stopping VM pause monitor" );
+        worker.interrupt();
```
MishaDemianenko: would really like if instead of interruption we would use  own flag stop kind of flag

community/kernel/src/main/java/org/neo4j/kernel/monitoring/VmPauseMonitor.java
```diff
@@ -0,0 +1,176 @@
...
+            this.gcTime = gcTime;
+            this.gcCount = gcCount;
+        }
+
+        public long getPauseTime()
+        {
+            return pauseTime;
+        }
+
+        public long getGcTime()
```
MishaDemianenko: some getters here are declared but never used

community/kernel/src/main/java/org/neo4j/unsafe/impl/batchimport/staging/OnDemandDetailsExecutionMonitor.java
```diff
@@ -269,4 +272,20 @@ void collect()
             doneBatches = lastDoneBatches;
         }
     }
+
+    private static class VmPauseTimeAccumulator implements Consumer<VmPauseInfo>
+    {
+        private final AtomicLong totalPauseTime = new AtomicLong();
```
MishaDemianenko, line 281: you can use LongAdder here instead

community/kernel/src/main/java/org/neo4j/kernel/monitoring/VmPauseMonitor.java
```diff
@@ -0,0 +1,176 @@
...
+
+import static java.lang.String.format;
+import static java.lang.System.nanoTime;
+import static java.lang.Thread.currentThread;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+import static org.neo4j.util.Preconditions.requirePositive;
+
+public final class VmPauseMonitor
```
MishaDemianenko: Also would be really nice to have tests and confirm that actual reporting and  measuring works

MishaDemianenko (changes requested):
Nice to get reed of old monitor but some things should be changed here and there

community/kernel/src/main/java/org/neo4j/kernel/monitoring/VmPauseMonitor.java
```diff
@@ -0,0 +1,175 @@
...
+
+import static java.lang.String.format;
+import static java.lang.System.nanoTime;
+import static java.lang.Thread.currentThread;
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+import static org.neo4j.util.Preconditions.checkState;
+import static org.neo4j.util.Preconditions.requirePositive;
+
+public final class VmPauseMonitor
```
MishaDemianenko: There are still no tests that confirm we can actually report and measure something

MishaDemianenko (changes requested):
Almost everything is fine now, except tests for monitoring that are not there yet

MishaDemianenko (approved):

PR #11017, opened 2018-02-14
Elector identification now happens more reliably

digitalstain:
When an instance joins an HA cluster, it assumes all other
 members, as they are received in the configuration response,
 are alive. This means that, if the newly joining instance, is the
 lowest numbered alive one but there are others, lower numbered, but
 which are failed, it will choose not to trigger elections. Updating
 liveness state afterwards, through suspicion gossip, will not
 alter that since marking a member as failed does not result in
 elections if that member does not hold any roles.
This patch addresses this issue by making liveness information
 part of the configuration response and consulting that list on
 the first election request. By getting all alive instances and
 removing what the cluster says is currently failed, we end up with
 a reliable list of alive members which then we can check and see
 who the elector is.

### DIFF
--- a/enterprise/cluster/src/test/java/org/neo4j/cluster/protocol/election/ElectionContextTest.java
+++ b/enterprise/cluster/src/test/java/org/neo4j/cluster/protocol/election/ElectionContextTest.java
@@ -19,8 +19,6 @@
  */
 package org.neo4j.cluster.protocol.election;
 
-import org.junit.Test;
-
 import java.net.URI;
 import java.util.Collections;
 import java.util.HashMap;
@@ -30,6 +28,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Executor;
+
+import org.junit.Test;
 
 import org.neo4j.cluster.ClusterSettings;
 import org.neo4j.cluster.InstanceId;
@@ -48,6 +48,7 @@ import org.neo4j.logging.NullLogProvider;
 import static junit.framework.Assert.assertNull;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
@@ -56,7 +57,7 @@ public class ElectionContextTest
     @Test
     public void testElectionOkNoFailed()
     {
-        Set<InstanceId> failed = new HashSet<InstanceId>();
+        Set<InstanceId> failed = new HashSet<>();
 
         baseTestForElectionOk( failed, false );
     }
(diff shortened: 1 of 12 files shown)


enterprise/cluster/src/test/java/org/neo4j/cluster/protocol/election/ElectionContextTest.java
```diff
@@ -188,14 +189,52 @@ public void testElectionNotOkQuorumFailedFiveInstances()
...
+
+        Map<InstanceId, URI> members = new HashMap<>();
+        members.put( new InstanceId( 1 ), URI.create( "server1" ) );
+        members.put( new InstanceId( 2 ), URI.create( "server2" ) );
+        members.put( new InstanceId( 3 ), URI.create( "server3" ) );
+        members.put( new InstanceId( 4 ), URI.create( "server4" ) );
+        members.put( new InstanceId( 5 ), URI.create( "server5" ) );
+
+        Config config = mock( Config.class );
+        when( config.get( ClusterSettings.max_acceptors ) ).thenReturn( 10 );
```
martinfurmanski: I think you can get this behaviour quite stably even without a mock!

enterprise/cluster/src/test/java/org/neo4j/cluster/protocol/election/ElectionContextTest.java
```diff
@@ -188,14 +189,52 @@ public void testElectionNotOkQuorumFailedFiveInstances()
...
+
+        ClusterConfiguration clusterConfiguration = mock( ClusterConfiguration.class );
+        when( clusterConfiguration.getMembers() ).thenReturn( members );
+        when( clusterConfiguration.getMemberIds() ).thenReturn( members.keySet() );
+
+        ClusterContext clusterContext = mock( ClusterContext.class );
+        when( clusterContext.getConfiguration() ).thenReturn( clusterConfiguration );
+
+        MultiPaxosContext context = new MultiPaxosContext( new InstanceId(3 ), Iterables.iterable(
```
martinfurmanski: Perhaps single out the 3 as a variable to make it clear which instance is the one being tested.

enterprise/cluster/src/test/java/org/neo4j/cluster/protocol/election/ElectionContextTest.java
```diff
@@ -188,14 +189,52 @@ public void testElectionNotOkQuorumFailedFiveInstances()
...
+                new ElectionRole( "coordinator" ) ), clusterConfiguration,
+                mock( Executor.class ), NullLogProvider.getInstance(),
+                mock( ObjectInputStreamFactory.class ), mock( ObjectOutputStreamFactory.class ),
+                mock( AcceptorInstanceStore.class ), mock( Timeouts.class ),
+                mock( ElectionCredentialsProvider.class), config );
+
+        context.getHeartbeatContext().getFailed().addAll( failed );
```
martinfurmanski: Maybe standard //when comment on this?

Maybe also another sanity-check assertion just before this just to check that we were not the elector before we were informed about the failed instances?

martinfurmanski (approved):
Looks good.

PR #11021, opened 2018-02-15
Implement planner support for index queries on distance predicates

sherfert:
changelog: Index-backed support for spatial point distance searches from Cypher.

For example, the following query will return all points within 10km of the point `here`:

    WITH point({latitude: 56.5, longitude: 12.5}) as here
    MATCH (c:CoffeeShop) WHERE distance(c.location, here) < 10000
    RETURN c.name as CoffeeShop, c.address as Address, distance(c.location, here) as distance
    ORDER BY distance


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
```diff
@@ -42,6 +42,17 @@ class MatchAcceptanceTest extends ExecutionEngineFunSuite with QueryStatisticsTe
     result.toList should equal(List(Map("count(n.name)" -> 3)))
   }
 
+  test("x") {
```
OliviaYtterbrink: This doesn't look like a test...

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/PlanDescriptionArgumentSerializer.scala
```diff
@@ -50,6 +50,8 @@ object PlanDescriptionArgumentSerializer {
       case Index(label, properties) => s":$label(${properties.mkString(",")})"
       case PrefixIndex(label, property, p) => s":$label($property STARTS WITH ${asPrettyString(p)})"
       case InequalityIndex(label, property, bounds) => bounds.map(bound => s":$label($property) $bound").mkString(" AND ")
+      case PointDistanceIndex(label, property, point, distance, inclusive) =>
+        s":$label($property) WHERE distance(_,$point) <${if(inclusive) "=" else ""} $distance"
```
Lojjs, line 54: Do we really need the `else` case here?

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/steps/AbstractIndexSeekLeafPlanner.scala
```diff
@@ -41,7 +41,7 @@ abstract class AbstractIndexSeekLeafPlanner extends LeafPlanner with LeafPlanFro
                               valueExpr: QueryExpression[Expression],
                               hint: Option[UsingIndexHint],
                               argumentIds: Set[String],
-                              context: LogicalPlanningContext): Seq[Expression] => LogicalPlan
+                              context: LogicalPlanningContext): (Seq[Expression], Seq[Expression]) => LogicalPlan
```
systay: I think we can name the inputs by using something like this:
```
  protected def constructPlan(idName: String,
                              label: LabelToken,
                              propertyKeys: Seq[PropertyKeyToken],
                              valueExpr: QueryExpression[Expression],
                              hint: Option[UsingIndexHint],
                              argumentIds: Set[String],
                              context: LogicalPlanningContext)
			     (x1: Seq[Expression], x2: Seq[Expression]): LogicalPlan
```

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/steps/uniqueIndexSeekLeafPlanner.scala
```diff
@@ -32,8 +32,8 @@ object uniqueIndexSeekLeafPlanner extends AbstractIndexSeekLeafPlanner {
                               valueExpr: QueryExpression[Expression],
                               hint: Option[UsingIndexHint],
                               argumentIds: Set[String],
-                              context: LogicalPlanningContext): (Seq[Expression]) => LogicalPlan =
-    (predicates: Seq[Expression]) =>
+                              context: LogicalPlanningContext): (Seq[Expression], Seq[Expression]) => LogicalPlan =
+    (predicates: Seq[Expression], _: Seq[Expression]) =>
```
Lojjs: In `IndexSeekLeafPlanner` and `mergeUniqueIndexSeekLeafPlanner` you renamed `predicates` to `predicatesSolved`. I think you should do the same here for consistency.

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/plans/Sargable.scala
```diff
@@ -110,6 +110,34 @@ object AsValueRangeSeekable {
   }
 }
 
+// WHERE distance(p.prop, otherPoint) < number
+// and the like
+object AsDistanceSeekable {
```
Lojjs, line 115: The corresponding objects `AsStringRangeSeekable` and `AsValueRangeSeekable` have `calculateSelectvityFor...` methods in the `ExpressionSelectivityCalculator.scala`. I'm not sure, but maybe you need to add something similar for distances?

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/steps/LogicalPlanProducer.scala
```diff
@@ -200,16 +200,24 @@ case class LogicalPlanProducer(cardinalityModel: CardinalityModel, solveds: Solv
...
-    annotate(NodeIndexSeek(idName, label, propertyKeys, valueExpr, argumentIds), solved, context)
+    // We know solvedPredicates is a subset of solvedPredicatesForCardinalityEstimation
+    val solved = RegularPlannerQuery(queryGraph = queryGraph)
+    val solvedForCardinalityEstimation = RegularPlannerQuery(queryGraph.addPredicates(solvedPredicatesForCardinalityEstimation: _*))
+
+    val plan = NodeIndexSeek(idName, label, propertyKeys, valueExpr, argumentIds)
+    val cardinality = cardinalityModel(solvedForCardinalityEstimation, context.input, context.semanticTable)
+    solveds.set(plan.id, solved)
+    cardinalities.set(plan.id, cardinality)
+    plan
```
Lojjs, line 220: An idea would be to rewrite `annotate` so it still can be used here. i.e. that it takes a separate `solvedForCardinalityEstimation` which can be the same as `solved` as default. That would make it clearer that these two `solved`s don't necessarily must be the same thing.

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/LogicalPlan2PlanDescription.scala
```diff
@@ -405,6 +405,13 @@ case class LogicalPlan2PlanDescription(readOnly: Boolean, cardinalities: Cardina
             val lessThanBoundsText = lessThanBounds.bounds.map(bound =>
               s"<${bound.inequalitySignSuffix} ${bound.endPoint.asCanonicalStringVal}").toIndexedSeq
             (name, InequalityIndex(label.name, propertyKey, greaterThanBoundsText ++ lessThanBoundsText))
+          case PointDistanceSeekRangeWrapper(PointDistanceRange(point, distance, inclusive)) =>
+            val poi = point match {
+              case FunctionInvocation(Namespace(List()), FunctionName("point"), _, Seq(MapExpression(args))) =>
+                s"point(${args.map(_._1.name).mkString(",")})"
```
Lojjs: This was not so easy to understand. Maybe you can write a multiple line solution or a comment to make it more clear what `args.map(_._1.name)` returns

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/LogicalPlan2PlanDescription.scala
```diff
@@ -405,6 +405,13 @@ case class LogicalPlan2PlanDescription(readOnly: Boolean, cardinalities: Cardina
             val lessThanBoundsText = lessThanBounds.bounds.map(bound =>
               s"<${bound.inequalitySignSuffix} ${bound.endPoint.asCanonicalStringVal}").toIndexedSeq
             (name, InequalityIndex(label.name, propertyKey, greaterThanBoundsText ++ lessThanBoundsText))
+          case PointDistanceSeekRangeWrapper(PointDistanceRange(point, distance, inclusive)) =>
+            val poi = point match {
+              case FunctionInvocation(Namespace(List()), FunctionName("point"), _, Seq(MapExpression(args))) =>
```
systay: In a couple of places, we have used the hard coded strings for function names. I think it would be better if we instead used something like `org.neo4j.cypher.internal.v3_4.functions.Point.name`

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/MatchAcceptanceTest.scala
systay: Either remove or make into a real unit test?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -477,6 +477,29 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with Cypher
...
   }
 
+  test("Index query with MERGE") {
+    // Given
+    graph.createIndex("Place", "location")
+    graph.execute("CREATE (p:Place) SET p.location = point({latitude: 56.7, longitude: 12.78, crs: 'WGS-84'})")
+    graph.execute("CREATE (p:Place) SET p.location = point({latitude: 56.700001, longitude: 12.7800001, crs: 'WGS-84'})")
+
+    // When matching in merge
+    val result = innerExecuteDeprecated("MERGE (p:Place {location: point({latitude: 56.7, longitude: 12.78, crs: 'WGS-84'}) }) RETURN p.location as point", Map.empty)
```
systay: Looks like we are using this deprecated method a lot. Sounds like a bad idea - should we not try to make this new code not build on deprecated stuff?

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -235,7 +236,32 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
       indexSeekByPrefixRange(index, prefix)
     case range: InequalitySeekRange[Any] =>
       indexSeekByPrefixRange(index, range)
+    case range: PointDistanceRange[Any] =>
+      val distance: Double = range.distance match {
```
Lojjs: Don't need a type annotation here

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -235,7 +236,32 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
...
+          CartesianCalculator.boundingBox(p, distance)
+        case p:PointValue if p.getCoordinateReferenceSystem == WGS84 =>
+          HaversinCalculator.boundingBox(p, distance)
+        case _ =>
+          // when it tries to evaluate the distance on something that is not a point
+          return Iterator.empty
+      }
+
+      val (fromBounds, toBounds) =
+        if (range.inclusive)
```
Lojjs, line 259: Does this mean that we don't have support for "half-inclusive" ranges where the lower bound is inclusive and the higher exclusive or vice versa?

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/steps/AbstractIndexSeekLeafPlanner.scala
```diff
@@ -151,28 +153,33 @@ abstract class AbstractIndexSeekLeafPlanner extends LeafPlanner with LeafPlanFro
...
     case predicate@AsValueRangeSeekable(seekable) =>
       val queryExpression = seekable.asQueryExpression
       val keyName = seekable.propertyKeyName
-      IndexPlannableExpression(seekable.name, keyName, predicate, queryExpression, hints, argumentIds)
+      IndexPlannableExpression(seekable.name, keyName, predicate, queryExpression, hints, argumentIds, solvesPredicate = true)
+
+    case predicate@AsDistanceSeekable(seekable) =>
+      val queryExpression = seekable.asQueryExpression
+      val keyName = seekable.propertyKeyName
+      IndexPlannableExpression(seekable.name, keyName, predicate, queryExpression, hints, argumentIds, solvesPredicate = false)
```
systay: I'm struggling to understand why this last case does not solve the predicate. Maybe we can add a comment to make it clearer?

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/DistanceFunction.scala
```diff
@@ -104,4 +101,12 @@ object HaversinCalculator extends DistanceCalculator {
     val greatCircleDistance = 2.0 * atan2(sqrt(a), sqrt(1-a))
     EARTH_RADIUS_METERS * greatCircleDistance
   }
+
+  override def boundingBox(p: PointValue, distance: Double): (PointValue, PointValue) = {
+    val coordinates = p.coordinate()
+    val dx = 180.0 / Math.pow(2.0 * Math.PI, 2) * distance / EARTH_RADIUS_METERS
```
Lojjs: Wouldn't it be nice performance wise to pre-calculate 
`180.0 / Math.pow(2.0 * Math.PI, 2) / EARTH_RADIUS_METERS` and store in a val in the object?

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -235,7 +236,32 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
...
+        case n:NumberValue => n.doubleValue()
+        case n:Number => n.doubleValue()
+        case _ =>
+          // We can't compare against something that is not a number, so no rows will match
+          return Iterator.empty
+      }
+
+      val (from, to) = range.point match {
+        case p:PointValue if p.getCoordinateReferenceSystem == Cartesian =>
+          CartesianCalculator.boundingBox(p, distance)
```
Lojjs, line 250: The points within a certain distance from a specific point forms a circle, while the `boundingBox` is the smallest rectangle encapsulating that circle. Where does the post filtering step that filter out the false positives in the edges of the rectangle happens?

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -569,6 +592,159 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with Cypher
...
+    val result = innerExecuteDeprecated(query, Map.empty)
+
+    // Then
+    val plan = result.executionPlanDescription()
+    plan should useOperatorWithText("Projection", "point")
+    plan should useOperatorWithText("Filter", "distance")
+    result.toList should equal(List(Map("point" -> Values.pointValue(CoordinateReferenceSystem.WGS84, 11.78, 55.7))))
+  }
+
+  test("indexed points with distance query and points within bbox") {
```
Lojjs: Good test!

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -569,6 +592,159 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with Cypher
...
+  test("points with distance query and mixed crs") {
+    // Given
+    graph.execute("CREATE (p:Place) SET p.location = point({y: 55.7, x: 11.78, crs: 'cartesian'})")
+    graph.execute("CREATE (p:Place) SET p.location = point({latitude: 56.7, longitude: 12.78, crs: 'WGS-84'})")
+    graph.execute("CREATE (p:Place) SET p.location = point({latitude: 55.7, longitude: 11.78, crs: 'WGS-84'})")
+    graph.execute("CREATE (p:Place) SET p.location = point({latitude: 50.7, longitude: 12.78, crs: 'WGS-84'})")
+    graph.execute("CREATE (p:Place) SET p.location = point({latitude: 56.7, longitude: 10.78, crs: 'WGS-84'})")
+
+    val query =
+      """CYPHER runtime=slotted MATCH (p:Place)
```
Lojjs: Why are you only running with slotted, do we not support distance queries in interpreted? (this and following tests)

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/PlanDescriptionArgumentSerializer.scala
Lojjs: Also add some testing for this!

sherfert:
WGS84 Test support is missing and WGS84 bounding box calculation is currently buggy.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/DistanceFunction.scala
craigtaverner: I would expect the compiler to do this for us, and having a one liner here is easier to read.

craigtaverner: Having said that, I believe this equation is wrong. I've re-derived a better one that takes into account that distance on latitude changes with longitude. We should correct this with:

    dx_latitude = 180 * distance / (Math.PI * EARTH_RADIUS_METERS)
    dx_longitude = 180 * distance / (Math.PI * EARTH_RADIUS_METERS * Math.cos( Math.toRadians( coordinates(1) ) ))

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/steps/AbstractIndexSeekLeafPlanner.scala
craigtaverner: Sounds good. We could add a comment like:

> The planned index seek will _almost_ satisfy the predicate, but with the possibility of some false positives. Since it reduces the cardinality to almost the level of the predicate, we can use the predicate to calculate cardinality, but not mark it as solved, since the planner will still need to solve it with a Filter.

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/steps/LogicalPlanProducer.scala
craigtaverner: We thought about this, but since it is only this specific method that does things differently, it felt better to keep these changes inline. If we have more methods that do this, we can pull it out to a separate method.

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/steps/uniqueIndexSeekLeafPlanner.scala
craigtaverner: Agreed. `predicatesSolved` sounds good and explicit.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
craigtaverner: That happens in the leaf planner where we don't report the predicates as solved, so that the later planning steps are forced to plan a filter for those predicates. See my response to Andres request for a comment above. I suggested the following:

> The planned index seek will _almost_ satisfy the predicate, but with the possibility of some false positives. Since it reduces the cardinality to almost the level of the predicate, we can use the predicate to calculate cardinality, but not mark it as solved, since the planner will still need to solve it with a Filter.

sherfert:
Also, we need tests for the doughnut shape.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
craigtaverner: The inclusiveness is on the distance<x predicate, which can be translated into an index search. If we have a<distance<b we can still only use the distance<b part for the index, so we would still only have one inclusive boolean.

However, you have a point that we should have tests that verify that predicates like a<distance<b are correctly planned with spatial index seeks.

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/LogicalPlan2PlanDescription.scala
craigtaverner: The intention here was a short and readible 'other' column. In this case we will have something that looks like `point(x,y,crs)`, with the values stripped out because they are visually cluttering (include stuff like `  AUTOVAR32`) and are duplicates of other rows. The syntax `_._1.name` takes the tuple from the map, uses the key (`_1`) and only the `name` of the key.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
craigtaverner: We've been using this method a lot in many places. Clearly a sign that we should rename it and remove the `deprecated` from the name. I think it was named that under the assumption that it should not be necessary to use this method, but in retrospect, it seems very, very useful.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
craigtaverner: You are right, we should add comparison support to these tests. This is for more than this PR, but the other, main, spatial PR did the same thing, so I suggest we do this as part of a new PR specifically about getting spatial comparison support correct.

craigtaverner (changes requested):
Need to fix the bbox calculation for WGS84, which is very wrong.

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/LogicalPlan2PlanDescription.scala
sherfert: I have only found this one place...

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
sherfert: But actually we could be using `executeWith` to see it works in different runtimes...

community/cypher/runtime-util/src/main/scala/org/neo4j/cypher/internal/runtime/planDescription/PlanDescriptionArgumentSerializer.scala
sherfert: I just tested, and it inserts "()", if you don't explicitly have the else case. I'll add tests.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
sherfert: I will attempt to do it here. These things get forgotten to easily.

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/plans/Sargable.scala
sherfert: Well spotted! We missed that completely!

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialDistanceAcceptanceTest.scala
```diff
@@ -0,0 +1,474 @@
...
+        """.stripMargin
+
+      // Then
+      val expected = Set(
+        Map("point" -> Values.pointValue(CoordinateReferenceSystem.Cartesian, 0, 9.99))
+      )
+      expectResultsAndIndexUsage(query, expected, inclusiveRange = false)
+    }
+        // <= geographic
+        {
```
craigtaverner, line 284: Indented an extra 4 spaces?

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/PointDistanceSeekRangeExpression.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.neo4j.cypher.internal.runtime.interpreted.pipes.QueryState
+import org.neo4j.cypher.internal.util.v3_4.InternalException
+import org.neo4j.cypher.internal.v3_4.logical.plans.PointDistanceRange
+import org.neo4j.values.AnyValue
+
+case class PointDistanceSeekRangeExpression(range: PointDistanceRange[Expression])
+  extends Expression {
+
+  override def apply(ctx: ExecutionContext, state: QueryState): AnyValue = throw new
+      InternalException("This should never be called")
```
craigtaverner, line 32: Looks like all expressions that should never have this method called are expressions created by the expression converter. Might be nice to have them all extend a super-class with this method implemented final.

community/cypher/interpreted-runtime/src/test/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/DistanceFunctionTest.scala
```diff
@@ -0,0 +1,186 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.neo4j.cypher.internal.runtime.interpreted.commands.expressions
+
+import org.neo4j.cypher.internal.util.v3_4.test_helpers.CypherFunSuite
+import org.neo4j.values.storable.{CoordinateReferenceSystem, PointValue, Values}
+import org.scalactic.{Equality, TolerantNumerics}
+import org.scalatest.matchers.{MatchResult, Matcher}
+
+class DistanceFunctionTest extends CypherFunSuite {
```
craigtaverner, line 27: Most of the tests inside this class seem to test the distance and bbox library that currently (but not correctly) resides inside the DistanceFunction class. I think these test could or should be elsewhere (as should the calculations themselves).

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/DistanceFunction.scala
```diff
@@ -59,49 +57,99 @@ case class DistanceFunction(p1: Expression, p2: Expression) extends Expression {
...
-  def isDefinedAt(p1: PointValue, p2: PointValue): Boolean
-
-  def calculateDistance(p1: PointValue, p2: PointValue): Double
-
-  def apply(p1: PointValue, p2: PointValue): Option[Double] =
-    if (isDefinedAt(p1, p2))
-      Some(calculateDistance(p1, p2))
-    else
-      None
+trait DistanceCalculator extends PartialFunction[(PointValue, PointValue), Double] {
```
craigtaverner, line 60: These calculators are more general than the DistanceFunction itself and so should reside somewhere else. Current thinking is to move them to the CRS.

craigtaverner (approved):
Very nice work. Some refactoring suggestions I've made, like moving the distance and bbox calculations to the CoordinateReferenceSystem can be done in another PR (eg. the 3D PR which already does that)

PR #11026, opened 2018-02-15
Make use of dynamic keyValueSizeHeader

burqen:
Header describing key and value size is dynamic to use as few bytes as possible when storing. This was not used by TreeNodeDynamicSize however. This commit addresses that by calculating overhead before allocating space for key and value.


tinwelint:
Is it surprising that `FormatCompatibilityTest` isn't failing?

burqen:
@tinwelint yes, very!

tinwelint:
Oh wait, it isn't actually...

tinwelint (approved):

PR #11028, opened 2018-02-15
Cleanup usage of deprecated and removed methods in TestSecurityManager

MishaDemianenko:
As part of Java 10 security manager deprecations were cleaned up:
https://bugs.openjdk.java.net/browse/JDK-8186535
Since we do not use removed method and to allow
kernel compilation with latest jdk releases, old method in our
implementations also removed.


burqen (approved):
Seems ligit

Issue #9151, opened 2017-04-05
Access-Control-Allow-Origin header missing on responses other than 200 or 404.


Vilx2:
When performing requests against the REST endpoints, the Access-Control-Allow-Origin header is only present on the responses that have HTTP status codes 200 and 404. Other responses (like 302, 401, 400, etc) do not have this header. As a result ajax requests from the browser fail in these cases and do not report any status information to the calling script. This makes it impossible to distinguish between cases like a wrong URL, incorrect credentials, etc.

**Neo4j Version:** 3.1.3 
**Operating System:** Windows 7
**API:** Javascript from a browser using REST endpoints

### Steps to reproduce
Execute a script that uses XMLHttpRequest to retrieve http://localhost:7475/db/data/ but passes incorrect credentials.

### Expected behavior
XMLHttpRequest reports status 401.

### Actual behavior
XMLHttpRequest reports status 0

technige:
Thanks, we'll look into this.

PR #11034, opened 2018-02-16
Always attach CORS headers in REST

lutovich:
Previously CORS headers were only added to requests that passed auth. This made JS requests from browser unable to receive the actual error because CORS is not allowed with missing headers.

This PR moves the relevant filter to the top of the filter chain so that every response has CORS headers attached.

Fixes https://github.com/neo4j/neo4j/issues/9151


community/server/src/main/java/org/neo4j/server/rest/web/CorsFilter.java
```diff
@@ -0,0 +1,87 @@
...
+ * This filter adds the header "Access-Control-Allow-Origin : *" to all
+ * responses that goes through it. This allows modern browsers to do cross-site
+ * requests to us via javascript.
+ */
+public class CorsFilter implements Filter
+{
+    private static final String ACCESS_CONTROL_ALLOW_ORIGIN = "Access-Control-Allow-Origin";
+    private static final String ACCESS_CONTROL_ALLOW_METHODS = "Access-Control-Allow-Methods";
+    private static final String ACCESS_CONTROL_ALLOW_HEADERS = "Access-Control-Allow-Headers";
+    private static final String ACCESS_CONTROL_REQUEST_METHOD = "access-control-request-method";
```
jimwebber: capitalise to Access-Control-Request-Method

community/server/src/main/java/org/neo4j/server/rest/web/CorsFilter.java
```diff
@@ -0,0 +1,87 @@
...
+    private static final String ACCESS_CONTROL_REQUEST_HEADERS = "access-control-request-headers";
```
jimwebber: Capitalise to Access-Control-Request-Headers

jimwebber (changes requested):
I think those headers are capitalised to be a good citizen in the HTTP world. Otherwise this seems sane to me.

ali-ince (approved):
LGTM!

I'm aware that this filter was already part of the product as it is here, but I think it may be important to discuss the logic of setting `Access-Control-Allow-Origin` to `*`  without a way to limit the cross-origin requests originators. The same logic applies to `Access-Control-Request-Method` and `Access-Control-Allow-Methods`. WDYT?

PR #11036, opened 2018-02-16
Close procedure streams on failures

henriknyman:
Make sure we call any close handlers attached to procedure
streams on exceptions. Exceptions that occur during close are
collected as ResourceCloseFailureException and attached as suppressed.

Introduces a ResourceManager/ResourceTracker on KernelStatement to
keep track of active resources so that any non-exhausted procedure
streams are always closed if the transaction is closed during query
execution without closing the query result first.

The ResourceManager implementation is separated into a
class CloseableResourceManager so we can easily move it if we
will not keep KernelStatement in the future.

Also improves the error message.

Doc are adapted to this change with https://github.com/neo4j/neo4j-documentation/pull/185.

### DIFF
--- a/community/cypher/acceptance-spec-suite/src/test/java/org/neo4j/internal/cypher/acceptance/TestResourceProcedure.java
+++ b/community/cypher/acceptance-spec-suite/src/test/java/org/neo4j/internal/cypher/acceptance/TestResourceProcedure.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright (c) 2002-2018 "Neo Technology,"
+ * Network Engine for Objects in Lund AB [http://neotechnology.com]
+ *
+ * This file is part of Neo4j.
+ *
+ * Neo4j is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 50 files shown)


community/cypher/acceptance-spec-suite/src/test/java/org/neo4j/internal/cypher/acceptance/TestResourceProcedure.java
```diff
@@ -0,0 +1,174 @@
...
+            closeCountTestOnCloseFailingResourceProcedure = 0;
+            openCountTestResourceProcedure = 0;
+            openCountTestFailingResourceProcedure = 0;
+            openCountTestOnCloseFailingResourceProcedure = 0;
+        }
+    }
+
+    public static ComponentRegistry.Provider<Counters> countersProvider(Counters counters1)
+    {
+        return context -> counters1;
```
fickludd: `..nters1`?

community/cypher/acceptance-spec-suite/src/test/java/org/neo4j/internal/cypher/acceptance/TestResourceProcedure.java
```diff
@@ -0,0 +1,174 @@
...
+        stream.onClose( () ->
+        {
+            counters.closeCountTestResourceProcedure++;
+        } );
+        counters.openCountTestResourceProcedure++;
+        return stream;
+    }
+
+    @Procedure( name = "org.neo4j.test.testFailingResourceProcedure", mode = Mode.READ )
+    @Description( "Returns a stream of integers from 1 to the given argument, but throws an exception when reaching the last element" )
```
fickludd, line 120: This actually starts from `2`, because it increments before returning.

fickludd (approved):
Looks like a well needed addition to assert correct resource management. This looks to me like very helpful in changing resource management over in 3.4 as well. 

Got some minor pointers that I will fix myself =).

PR #11037, opened 2018-02-16
3.4 spatial support for 3D cartesian points

craigtaverner:
changelog: Added support for 3D cartesian and 3D WGS84 (GPS) points

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
@@ -256,55 +256,66 @@ public final class ValueUtils
 
     public static PointValue pointFromMap( MapValue map )
     {
-        if ( map.containsKey( "x" ) && map.containsKey( "y" ) )
+        CoordinateReferenceSystem crs;
+        double[] coordinates;
+        if ( map.containsKey( "crs" ) )
         {
-            double x = ((NumberValue) map.get( "x" )).doubleValue();
-            double y = ((NumberValue) map.get( "y" )).doubleValue();
-            double[] coordinates = map.containsKey( "z" ) ? new double[]{x, y, ((NumberValue) map.get( "z" )).doubleValue()} : new double[]{x, y};
-            CoordinateReferenceSystem crs = CoordinateReferenceSystem.Cartesian;
-            if ( map.containsKey( "crs" ) )
+            TextValue crsName = (TextValue) map.get( "crs" );
+            crs = CoordinateReferenceSystem.byName( crsName.stringValue() );
+            if ( crs == null )
             {
-                TextValue crsName = (TextValue) map.get( "crs" );
-                if ( crsName.stringValue().equals( CoordinateReferenceSystem.Cartesian.getName() ) )
-                {
-                    crs = CoordinateReferenceSystem.Cartesian;
-                }
-                else if ( crsName.stringValue().equals( CoordinateReferenceSystem.WGS84.getName() ) )
-                {
-                    crs = CoordinateReferenceSystem.WGS84;
-                }
-                else
-                {
-                    throw new IllegalArgumentException( "Unknown coordinate reference system: " + crsName.stringValue() );
-                }
-            }
-            return Values.pointValue( crs, coordinates );
-        }
-        else if ( map.containsKey( "latitude" ) && map.containsKey( "longitude" ) )
-        {
-            double latitude = ((NumberValue) map.get( "latitude" )).doubleValue();
-            double longitude = ((NumberValue) map.get( "longitude" )).doubleValue();
-            if ( !map.containsKey( "crs" ) )
-            {
-                return Values.pointValue( CoordinateReferenceSystem.WGS84, longitude, latitude );
-            }
-
-            TextValue crs = (TextValue) map.get( "crs" );
-            if ( crs.stringValue().equals( CoordinateReferenceSystem.WGS84.getName() ) )
-            {
-                return Values.pointValue( CoordinateReferenceSystem.WGS84, longitude, latitude );
-            }
-            else
-            {
-                throw new IllegalArgumentException(
-                        "Geographic points does not support coordinate reference system: " + crs.stringValue() );
+                throw new IllegalArgumentException( "Unknown coordinate reference system: " + crsName.stringValue() );
             }
         }
         else
         {
-            throw new IllegalArgumentException(
-                    "A point must contain either 'x' and 'y' or 'latitude' and 'longitude'" );
+            crs = null;
         }
+        if ( map.containsKey( "x" ) && map.containsKey( "y" ) )
+        {
+            double x = ((NumberValue) map.get( "x" )).doubleValue();
+            double y = ((NumberValue) map.get( "y" )).doubleValue();
+            coordinates = map.containsKey( "z" ) ? new double[]{x, y, ((NumberValue) map.get( "z" )).doubleValue()} : new double[]{x, y};
+            if ( crs == null )
+            {
+                crs = coordinates.length == 3 ? CoordinateReferenceSystem.Cartesian_3D : CoordinateReferenceSystem.Cartesian;
+            }
+        }
+        else if ( map.containsKey( "latitude" ) && map.containsKey( "longitude" ) )
+        {
+            double x = ((NumberValue) map.get( "longitude" )).doubleValue();
+            double y = ((NumberValue) map.get( "latitude" )).doubleValue();
+            // TODO Consider supporting key 'height'
+            if ( map.containsKey( "z" ) )
+            {
+                coordinates = new double[]{x, y, ((NumberValue) map.get( "z" )).doubleValue()};
+            }
+            else if ( map.containsKey( "height" ) )
+            {
+                coordinates = new double[]{x, y, ((NumberValue) map.get( "height" )).doubleValue()};
+            }
+            else
+            {
+                coordinates = new double[]{x, y};
+            }
+            if ( crs == null )
+            {
+                crs = coordinates.length == 3 ? CoordinateReferenceSystem.WGS84_3D : CoordinateReferenceSystem.WGS84;
+            }
+            if ( !crs.isGeographic() )
+            {
+                throw new IllegalArgumentException( "Geographic points does not support coordinate reference system: " + crs );
+            }
+        }
+        else
+        {
+            throw new IllegalArgumentException( "A point must contain either 'x' and 'y' or 'latitude' and 'longitude'" );
+        }
+        if ( crs.getDimension() != coordinates.length )
+        {
+            throw new IllegalArgumentException( "Cannot create " + crs.getDimension() + "D point with " + coordinates.length + " coordinates" );
+        }
+        return Values.pointValue( crs, coordinates );
     }
 
     private static Map<String,AnyValue> mapValues( Map<String,Object> map )
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
@@ -254,30 +254,31 @@ public final class ValueUtils
         return map( mapValues( map ) );
     }
 
-    public static PointValue fromMap( MapValue map )
+    public static PointValue pointFromMap( MapValue map )
     {
         if ( map.containsKey( "x" ) && map.containsKey( "y" ) )
         {
             double x = ((NumberValue) map.get( "x" )).doubleValue();
             double y = ((NumberValue) map.get( "y" )).doubleValue();
-            if ( !map.containsKey( "crs" ) )
+            double[] coordinates = map.containsKey( "z" ) ? new double[]{x, y, ((NumberValue) map.get( "z" )).doubleValue()} : new double[]{x, y};
+            CoordinateReferenceSystem crs = CoordinateReferenceSystem.Cartesian;
+            if ( map.containsKey( "crs" ) )
             {
-                return Values.pointValue( CoordinateReferenceSystem.Cartesian, x, y );
+                TextValue crsName = (TextValue) map.get( "crs" );
+                if ( crsName.stringValue().equals( CoordinateReferenceSystem.Cartesian.getName() ) )
+                {
+                    crs = CoordinateReferenceSystem.Cartesian;
+                }
+                else if ( crsName.stringValue().equals( CoordinateReferenceSystem.WGS84.getName() ) )
+                {
+                    crs = CoordinateReferenceSystem.WGS84;
+                }
+                else
+                {
+                    throw new IllegalArgumentException( "Unknown coordinate reference system: " + crsName.stringValue() );
+                }
             }
-
-            TextValue crs = (TextValue) map.get( "crs" );
-            if ( crs.stringValue().equals( CoordinateReferenceSystem.Cartesian.getName() ) )
-            {
-                return Values.pointValue( CoordinateReferenceSystem.Cartesian, x, y );
-            }
-            else if ( crs.stringValue().equals( CoordinateReferenceSystem.WGS84.getName() ) )
-            {
-                return Values.pointValue( CoordinateReferenceSystem.WGS84, x, y );
-            }
-            else
-            {
-                throw new IllegalArgumentException( "Unknown coordinate reference system: " + crs.stringValue() );
-            }
+            return Values.pointValue( crs, coordinates );
         }
         else if ( map.containsKey( "latitude" ) && map.containsKey( "longitude" ) )
         {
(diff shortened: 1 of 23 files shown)


enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -175,15 +184,36 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with Cypher
...
+    )
+    val dist = result.columnAs[Any]("dist").next()
+    assert(dist == null)
+  }
+
+  test("distance function should return null if provided with points with different dimensions") {
+    val result = executeWith(expectedToSucceed,
+      """WITH point({x: 2.3, y: 4.5}) as p1, point({x: 1.2, y: 3.4, z: 5.6}) as p2
+        |RETURN distance(p1,p2) as dist""".stripMargin,
+      expectedDifferentResults = Configs.Version3_1 + Configs.AllRulePlanners // TODO should rather throw error
```
sherfert: I am unsure about this TODO. I think the test is correct and distance should return null. For the same reasons it must return null in other cases. I think we can remove the TODO.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -338,14 +368,40 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with Cypher
     point.getCRS.getHref should equal("http://spatialreference.org/ref/epsg/4326/")
   }
 
+  test("3D point should be assignable to node property") {
+    // Given
+    createLabeledNode("Place")
+
+    // When
+    val config = expectedToSucceed - Configs.Cost3_1 - Configs.AllRulePlanners - Configs.Morsel
```
sherfert: There should be no need to explicitly subtract Morsel. It's not run by default.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -338,14 +368,40 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with Cypher
...
+    createLabeledNode("Place")
+    graph.execute("MATCH (p:Place) SET p.location = point({x: 1.2, y: 3.4, z: 5.6}) RETURN p.location as point")
+
+    // When
+    val result = executeWith(Configs.All, "MATCH (p:Place) RETURN p.location as point",
+      planComparisonStrategy = ComparePlansWithAssertion(_ should useOperatorWithText("Projection", "point"),
+        expectPlansToFail = Configs.AllRulePlanners))
+
+    // Then
+    val point = result.columnAs("point").toList.head.asInstanceOf[Point]
```
sherfert: Can we also have an assertion on the size of the result list? (should be 1)

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -440,6 +440,47 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with Cypher
...
+    // When
+    val result = executeWith(expectedToSucceed - Configs.Version3_1 - Configs.AllRulePlanners,
+      "MATCH (p:Place) WHERE p.location = point({x: 1.2, y: 3.4, z: 5.6}) RETURN p.location as point",
+      planComparisonStrategy = ComparePlansWithAssertion({ plan =>
+        plan should useOperatorWithText("Projection", "point")
+        plan should useOperatorWithText("NodeIndexSeek", ":Place(location)")
+      }, expectPlansToFail = Configs.AbsolutelyAll - Configs.Version3_4 - Configs.Version3_3))
+
+    // Then
+    val point = result.columnAs("point").toList.head.asInstanceOf[Point]
```
sherfert: Here also missing an assertion on result list size.

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
```diff
@@ -256,55 +256,66 @@ public static MapValue asMapValue( Map<String,Object> map )
...
-            return Values.pointValue( crs, coordinates );
         }
         else if ( map.containsKey( "latitude" ) && map.containsKey( "longitude" ) )
         {
-            double latitude = ((NumberValue) map.get( "latitude" )).doubleValue();
-            double longitude = ((NumberValue) map.get( "longitude" )).doubleValue();
-            if ( !map.containsKey( "crs" ) )
+            double x = ((NumberValue) map.get( "longitude" )).doubleValue();
+            double y = ((NumberValue) map.get( "latitude" )).doubleValue();
+            // TODO Consider supporting key 'height'
```
sherfert: Sounds weird if do not support "width" and "depth" either.

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -131,4 +196,146 @@ public int hashCode()
...
+        double[] min = new double[dimension];
+        double[] max = new double[dimension];
+        int cartesianStartIndex = 0;
+        if ( geographic )
+        {
+            // Geographic CRS default to extent of the earth in degrees
+            min[0] = -180.0;
+            max[0] = 180.0;
+            min[1] = -90.0;
+            max[1] = 90.0;
```
sherfert, line 232: Instead of having the geographic defaults defined here, and the cartesian defaults passed into as an arguments, I would like to see _all_ defaults defined in constants. Then the last two arguments to this function can be removed.

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -131,4 +196,146 @@ public int hashCode()
...
+        }
+        for ( int i = cartesianStartIndex; i < dimension; i++ )
+        {
+            min[i] = minCartesian;
+            max[i] = maxCartesian;
+        }
+        return Pair.of( min, max );
+    }
+
+    public interface Calculator
```
sherfert: Can we move `Calculator` and its two implementations to separate files?

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -131,4 +196,146 @@ public int hashCode()
...
+        {
+            assert center.getCoordinateReferenceSystem().dimension == dimension;
+            double[] coordinates = center.coordinate();
+            double dx_latitude = 180 * distance / (Math.PI * EARTH_RADIUS_METERS);
+            double dx_longitude = 180 * distance / (Math.PI * EARTH_RADIUS_METERS * cos( toRadians( coordinates[1] ) ));
+            double[] min = new double[dimension];
+            double[] max = new double[dimension];
+            for ( int i = 0; i < dimension; i++ )
+            {
+                double dx = (i == 0) ? dx_longitude : (i == 1) ? dx_latitude : distance;
```
sherfert: After your rebase, be sure to extend the tests to cover 3D bounding boxes.

community/values/src/test/java/org/neo4j/values/storable/CoordinateReferenceSystemTest.java
```diff
@@ -0,0 +1,114 @@
...
+        assertThat( "2D distance should match", crs.getCalculator().distance( geo( -45.0, 0.0 ), geo( 45.0, 0.0 ) ), closeTo( 10000000.0, 20000.0 ) );
+        //"distance function should measure distance from Copenhagen train station to Neo4j in Malmö"
+        PointValue cph = geo( 12.564590, 55.672874 );
+        PointValue malmo = geo( 12.994341, 55.611784 );
+        double expected = 27842.0;
+        assertThat( "2D distance should match", crs.getCalculator().distance( cph, malmo ), closeTo( expected, 0.1 ) );
+    }
+
+    @Test
+    public void shouldCalculateGeographicDistance3D()
```
sherfert, line 110: It seems like we need some more tests for the geographic 3D case.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -51,12 +51,13 @@ class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with Cypher
   }
 
   test("point function should work with literal map and 3D cartesian coordinates") {
-    val result = executeWith(expectedToSucceed, "RETURN point({x: 2.3, y: 4.5, z: 6.7, crs: 'cartesian'}) as point",
+    val result = executeWith(expectedToSucceed - Configs.Version3_1 - Configs.AllRulePlanners,
```
sherfert: I would like to see a test that 4D fails. (as we used to have for 3D)

sherfert (changes requested):
My main concern is the lack of acceptance tests. You are only asserting that things can be stored and retrieved again. I would like to see the same coverage as for 2D points.

* Acceptance test level tests of distance function on 3D points (both CRS).
 * Index queries with distance with points inside and outside of bounding boxes, asserting on no false positives or negatives.
* Testing the "doughnut shape" for 3D (index query with distance)
* Extending tests like "indexScan should handle multiple different types of points and also survive restart" and "change crs"
* Test for index range and equality queries. With points far apart, and so close that they are in the same index tile.
* Testing that z-coordinate of point function can be provided through query parameters or a previously defined map in Cypher.
* Property Store storage of Arrays of both 3D type points.

sherfert:
Also `SpaceFillingCurveTest.shouldHaveReasonableCoveredArea()` and the likes should have a 3D pendant, now that 3D curves are actually in use.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
craigtaverner: This is for backwards compatibility. Are you suggesting we make new releases of 3.1.x that change behaviour in those older releases to return null? Currently those releases return different results, which seems wrong. I thought we could try cause an error here in 3.4 when trying this against older releases.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
craigtaverner: We subtract morsel in many places (other untouched tests). I guess we should remove that across the board, in a separate commit.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
craigtaverner: We did not do that for the 2D case from which this was derived. We could add it to both perhaps.

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
craigtaverner: For geographic they are constants. For cartesian there are no real extents, only suggested extents for the purpose of index optimization. So the _extents_ mean a very, very different thing for the two types, and I think the code should reflect that difference. Perhaps we can highlight the difference more clearly with comments?

And in fact the correct approach for Cartesian extents is the planned work on configurable extents, which does not exist yet. The current code is a step in that direction with cartesian as parameters.

sherfert: I disagree here. They are different in the respect that geographic has bound on max and min of the values. That shouldn't mean they are not configurable. People might want to index just a country, city, etc.

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
```diff
@@ -147,4 +214,203 @@ public int hashCode()
...
+            else
+            {
+                return distance2D;
+            }
+        }
+
+        @Override
+        // http://janmatuschek.de/LatitudeLongitudeBoundingCoordinates
+        // But calculating in degrees instead of radians to avoid rounding errors
+        public Pair<PointValue,PointValue> boundingBox( PointValue center, double distance )
```
sherfert: Did you also move the bounding box tests to reflect that this logic mas moved? Cannot find it in the changeset.

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
craigtaverner: I think this should be done as part of the work to enable configuration of the index extents.

community/values/src/main/java/org/neo4j/values/storable/CoordinateReferenceSystem.java
craigtaverner: That test was written in scala, and the best new location for it is in the Values module, which currently only has java tests, so I thought to leave it alone for the moment.

community/values/src/test/java/org/neo4j/values/storable/CoordinateReferenceSystemTest.java
craigtaverner: The SpatialDistanceAcceptanceTest covers many more cases, so I thought this should cover something more real world.

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialIndexAcceptanceTest.scala
```diff
@@ -92,31 +118,67 @@ class SpatialIndexAcceptanceTest extends CypherFunSuite with GraphDatabaseTestSu
...
+      Values.pointValue(CoordinateReferenceSystem.WGS84_3D, 12.78, 56.7, 100.0),
+      Values.pointValue(CoordinateReferenceSystem.Cartesian_3D, 1.0, 2.78, 5.0))
   }
 
   test("indexSeek should handle multiple different types of points and also survive restart") {
     graph.createIndex("Place", "location")
 
     graph.execute("CREATE (p:Place) SET p.location = point({latitude: 56.7, longitude: 12.78, crs: 'WGS-84'})")
     graph.execute("CREATE (p:Place) SET p.location = point({x: 1.0, y: 2.78, crs: 'cartesian'})")
+    //graph.execute("CREATE (p:Place) SET p.location = point({latitude: 56.7, longitude: 12.78, height: 100.0, crs: 'WGS-84-3D'})")
```
sherfert: I think this should be uncommented, shouldn't it?

sherfert (changes requested):
We have a much greater coverage now, but some of the ones from my first review are not done yet, and I think they are crucial to have before merging this. Only one of them (number 3) is actually hard to code, so I think it is OK to defer the merging until these tests are in place.

1. In SpatialIndexResultsAcceptanceTest tests like "indexed points far apart in cartesian space - range query greaterThan" are missing their 3D pendant (or extension). This is true for most of the tests in that class.

2. SpatialIndexAcceptanceTest."change crs" should include changing to a 3D CRS.

3. DistanceFunctionTest."should calculate correct bounding box" should have a 3D pendant.

Lojjs (changes requested):
Apart from the tests that Satia already have mentioned, I had a look at the Spatial Index PR and I would like to see 3D points together with the 2D cartesian and WGS84 in the following test files: 
- IndexQueryTest.java
- IndexingAcceptanceTest.java
- CompositeIndexAccessorCompatibility.java
- NodeUpdatesTest.java
- SchemaConstraintProviderApprovalTest.java
- SchemaIndexProviderApprovalTest.java
- Neo4jJsonCodecTest.java
- PropertyVaueComparisonTest.java
- ExecutionResultSeralizerTest.java

Lojjs (approved):

sherfert (approved):

PR #11041, opened 2018-02-19
Fix most of Core API LBDC regressions

fickludd:
This PR contains several small optimizations and some bigger ones. Big ones are

 - Reuse Cursors in instance cache (single instance pool). 
 - Separate cursor close() and resource release into separate methods, so that resources and only released once the cursor is dropped from the pool. 
 - No unneeded txBridge interactions and statement acquires. 

The smaller optimizations were identified hotspots from async profiles. 

Beware that some code gets touched by several commits here, so unclear whether commit-by-commit or full-monkey would be the best review strategy.

### DIFF
--- a/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/RelationshipDenseSelection.java
+++ b/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/RelationshipDenseSelection.java
@@ -236,7 +236,7 @@ abstract class RelationshipDenseSelection
             break;
 
         default:
-            throw new IllegalStateException( "Lorem ipsus, Brutus" );
+            throw new IllegalStateException( "Lorem ipsus, Brutus. (could not setup cursor for Dir='" + d + "')" );
         }
     }
 
--- a/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/RelationshipDenseSelection.java
+++ b/community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/RelationshipDenseSelection.java
@@ -57,7 +57,7 @@ abstract class RelationshipDenseSelection
      * @param groupCursor Group cursor to use. Pre-initialized on node.
      * @param relationshipCursor Relationship traversal cursor to use.
      */
-    public void outgoing(
+    public final void outgoing(
             RelationshipGroupCursor groupCursor,
             RelationshipTraversalCursor relationshipCursor )
     {
@@ -71,7 +71,7 @@ abstract class RelationshipDenseSelection
      * @param relationshipCursor Relationship traversal cursor to use.
      * @param types Relationship types to traverse
      */
-    public void outgoing(
+    public final void outgoing(
             RelationshipGroupCursor groupCursor,
             RelationshipTraversalCursor relationshipCursor,
             int[] types )
@@ -94,7 +94,7 @@ abstract class RelationshipDenseSelection
      * @param groupCursor Group cursor to use. Pre-initialized on node.
      * @param relationshipCursor Relationship traversal cursor to use.
      */
-    public void incoming(
+    public final void incoming(
             RelationshipGroupCursor groupCursor,
             RelationshipTraversalCursor relationshipCursor )
     {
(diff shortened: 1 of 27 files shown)


pontusmelke:
@fickludd rebase

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultRelationshipTraversalCursor.java
```diff
@@ -507,6 +510,12 @@ private void reset()
     @Override
     protected void collectAddedTxStateSnapshot()
     {
+        if ( filterState == FilterState.NOT_INITIALIZED )
```
tinwelint, line 514: Not `&& filterStore`?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/RelationshipDenseSelection.java
```diff
@@ -187,63 +186,59 @@ protected boolean fetchNext()
...
+            onRelationship = relationshipCursor.next();
+            break;
+
+        case LOOP:
+            groupCursor.loops( relationshipCursor );
+            onRelationship = relationshipCursor.next();
+            break;
+
+        default:
+            throw new IllegalStateException( "Lorem ipsus, Brutus" );
```
tinwelint: It's cute and all, but at least include direction `d`?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipCursor.java
```diff
@@ -28,7 +28,8 @@
 abstract class RelationshipCursor extends RelationshipRecord implements RelationshipDataAccessor, RelationshipVisitor<RuntimeException>
 {
     Read read;
-    private HasChanges hasChanges = HasChanges.MAYBE;
+    private boolean hasChanges;
```
tinwelint, line 31: An alternative, if you want to keep the original design, would have been to let `hasChanges` be an `int` and `MAYBE`, `YES, and `NO` be int constants. Either one is fine I'd guess.

Question out of interest: did the overhead of the `enum` actually make a dent in the performance, showing up in profiling?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/RelationshipDenseSelectionIterator.java
```diff
@@ -72,6 +72,10 @@ public R next()
             close();
             next = RelationshipSelections.NO_ID;
         }
+        else
```
tinwelint: I'm a fan of consistent commits. Personally I would've merged this into https://github.com/neo4j/neo4j/pull/11041/commits/d141aeb00260d4cc6fc3c237e3daca809d61a0eb to aid our future selves.

tinwelint (commented):
Had some comments, but generelly looks fine to me

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultRelationshipTraversalCursor.java
fickludd: Nope. Here we ensure that the filter state is initialized so that we get the correct tx-state relationships, and therefore `filterStore` does not matter.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/RelationshipCursor.java
fickludd: Yeah I considered the int alternative, but guessed this would be faster in the happy path. 

It was pretty surprising, but before this change `hasChanges()` actually showed measurable self time in the LDBC Core read benchmark! I always though enum comparison should be superfast, as they are essentially just highly limited integers. This fix seems to remove that self time. Maybe that profile was misleading, but I though Async was pretty accurate.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/RelationshipDenseSelectionIterator.java
fickludd: Done.

tinwelint (approved):

PR #11045, opened 2018-02-19
More rigid FormatCompatibilityTest

burqen:
Add test for
- Put
- Remove


tinwelint (commented):
Question on the committed zip files: was it necessary to commit new versions of those? My expectation was that the initial state should be unchanged from what it was before this PR.

Otherwise looks great.

burqen:
Reason for new zip-files is because previous initial values was 0-20000. Now, instead we even numbers between 0-40000. This is to make it possible to later insert keys in the middle of the tree instead of only at the very end.

tinwelint (approved):

PR #11049, opened 2018-02-19
3.4 Fixed flaky PreElectionIT

phughk:
Test was failing because race condition meant heartbeats could occur during election and cancel election approval.

### DIFF
--- a/enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/PreElectionIT.java
+++ b/enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/PreElectionIT.java
@@ -19,8 +19,10 @@
  */
 package org.neo4j.causalclustering.scenarios;
 
+import org.hamcrest.Matchers;
 import org.junit.Rule;
 import org.junit.Test;
+import org.mockito.internal.hamcrest.HamcrestArgumentMatcher;
 
 import java.time.Clock;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/PreElectionIT.java
```diff
@@ -76,19 +81,15 @@ public void shouldStartAnElectionIfAllServersHaveTimedOutOnHeartbeats() throws E
...
-        {
-            if ( Role.FOLLOWER == member.raft().currentRole() )
-            {
-                futures.add( CompletableFuture.runAsync( Race.throwing( () -> member.raft().triggerElection( Clock.systemUTC() ) ) ) );
-            }
-        }
+        // when a heartbeat is missed and an election is triggered on followers
+        // (NOTE: this is actually done in a loop in the then clause since race conditions lead to a heartbeat being received
+        // sometimes during election => cancelling election)
+        futures.addAll( triggerElectionOnFollowers( cluster ) );
```
andrewkerr9000, line 87: Do we need to do this here, given it's in the async assert function?

andrewkerr9000 (changes requested):
Good work debugging this.

This test was added in 3.2, so any fixes should start there.

As discussed offline, this test is redundant - the simpler and less flaky `shouldStartElectionIfLeaderRemoved` covers the same scenario. Therefore this test will be deleted.

phughk:
Conclusion: going to delete test for 3.2 instead

PR #11050, opened 2018-02-19
Relationship properties in Kernel API

pontusmelke:
- Implement mutating property operations on relationships in the kernel API, i.e. `relationshipSetProperty` and `relationshipRemoveProperty`.
- Update Core API to use kernel API for all methods in `RelationshipProxy`
- Update interpreted runtime to use kernel API for all property access on relationships (and some operations that had been overlooked in previous work)
- Update compiled runtime to use kernel API for all property access on relationships.


community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
```diff
@@ -358,6 +361,247 @@ public void shouldSeeNewRelationshipPropertyInTransaction() throws Exception
     }
 
 
+
+    //start here
```
fickludd: `//start here`?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/ExplicitIndexCursorWritesTestBase.java
```diff
@@ -265,7 +264,7 @@ public void shouldCreateExplicitIndexTwice() throws Exception
     }
 
     //TODO unignore when we support relationship creation.
```
fickludd: Remove the TODO as well?

community/kernel/src/main/java/org/neo4j/kernel/impl/core/RelationshipProxy.java
```diff
@@ -412,24 +456,28 @@ public void setProperty( String key, Object value )
             throw new IllegalStateException( "Auto indexing encountered a failure while setting property: "
                                              + e.getMessage(), e );
         }
+        catch ( KernelException e )
+        {
+            throw new ConstraintViolationException( e.getMessage(), e );
```
fickludd: This is strange, I don't think we can get constraint violation for relationships when setting a property? When removing I guess we could violate a existence constraint, but that's anyway not checked until commit. 

In general it feels like here are way more catch clauses than necessary...

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -591,46 +588,63 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
 
     override def delete(id: Long) {
       try {
-        transactionalContext.statement.dataWriteOperations().relationshipDelete(id)
+        writes().relationshipDelete(id)
       } catch {
         case _: exceptions.EntityNotFoundException => // node has been deleted by another transaction, oh well...
       }
```
fickludd: In the Kernel API relationshipDelete is idempotent I think, so we can remove the try-catch as well.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -642,25 +656,36 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
...
-          relationship = entityAccessor.newRelationshipProxy(relationshipId, startNodeId, typeId, endNodeId)
-        }
-      })
-      Option(fromRelationshipProxy(relationship))
-    } catch {
-      case _: exceptions.EntityNotFoundException => None
-    }
+    override def getByIdIfExists(id: Long): Option[RelationshipValue] =
+      if (reads().relationshipExists(id))
+        Some(fromRelationshipProxy(entityAccessor.newRelationshipProxy(id)))
```
fickludd: Here I preferred the previous impl which cached the start/end/type of the relationship. That stuff is going to be in L1 cache anyway (given current store layout) so we might as well return it.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -674,24 +699,17 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
...
+    override def acquireExclusiveLock(obj: Long): Unit =
       transactionalContext.statement.readOperations().acquireExclusive(ResourceTypes.RELATIONSHIP, obj)
 
-    override def releaseExclusiveLock(obj: Long) =
+    override def releaseExclusiveLock(obj: Long): Unit =
       transactionalContext.statement.readOperations().acquireExclusive(ResourceTypes.RELATIONSHIP, obj)
 
-    override def exists(id: Long): Boolean = {
-      try {
-        transactionalContext.statement.readOperations().relationshipVisit(id, NoopVisitor)
```
fickludd, line 717: Now we can also remove the NoopVisitor!

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
```diff
@@ -105,9 +105,9 @@ final class TransactionBoundQueryContext(val transactionalContext: Transactional
   //We cannot assign to value because of periodic commit
   private def writes() = transactionalContext.dataWrite
   private def reads() = transactionalContext.dataRead
-  private val nodeCursor = allocateAndTraceNodeCursor()
-  private val relationshipScanCursor = allocateAndTraceRelationshipScanCursor()
-  private val propertyCursor = allocateAndTracePropertyCursor()
+  private lazy val nodeCursor = allocateAndTraceNodeCursor()
+  private lazy val relationshipScanCursor = allocateAndTraceRelationshipScanCursor()
+  private lazy val propertyCursor = allocateAndTracePropertyCursor()
```
fickludd: I'm curious: does this have a measurable effect, or why did you do it?

community/kernel/src/main/java/org/neo4j/kernel/impl/core/RelationshipProxy.java
```diff
@@ -155,6 +153,11 @@ public void delete()
             throw new IllegalStateException( "Auto indexing encountered a failure while deleting the relationship: "
                                              + e.getMessage(), e );
         }
+        catch ( EntityNotFoundException e )
+        {
+            throw new NotFoundException( "Unable to delete relationship[" +
+                                         getId() + "] since it is already deleted." );
+        }
```
fickludd: I don't think we need this catch - this is handled by the boolean return value.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -119,6 +119,11 @@ public void delete()
             throw new IllegalStateException( "Auto indexing encountered a failure while deleting the node: "
                                              + e.getMessage(), e );
         }
+        catch ( EntityNotFoundException e )
+        {
+            throw new NotFoundException( "Unable to delete node[" +
+                                         getId() + "] since it is already deleted." );
+        }
```
fickludd: This should be handled by the `nodeDelete` return value, so we don't need this catch.

fickludd (changes requested):
In general really good, well done on dragging this across all user and everywhere.

Some minor non-critical points, and there seems to be missing a locking tests for `relationshipDeleteProperty`.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundQueryContext.scala
pontusmelke: done

fickludd (approved):

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
```diff
@@ -312,6 +314,50 @@ public void shouldTraverseDenseNodeViaGroupsWithDetachedReferences() throws Exce
...
+
+                relationship.properties( property );
+                assertTrue( property.next() );
+                //First property
+                assertEquals( prop1, property.propertyKey() );
+                assertEquals( property.propertyValue(), stringValue( "hello" ) );
+                //second property
+                assertTrue( property.next() );
+                assertEquals( prop2, property.propertyKey() );
+                assertEquals( property.propertyValue(), stringValue( "world" ) );
```
chrisvest: This relies on the order in which the properties comes back from the cursor. I think that's over-specifying.

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
```diff
@@ -358,6 +361,247 @@ public void shouldSeeNewRelationshipPropertyInTransaction() throws Exception
...
+
+                //property 2, start with tx state
+                assertTrue( property.next() );
+                assertEquals( propToken2, property.propertyKey() );
+                assertEquals( property.propertyValue(), stringValue( "world" ) );
+
+                //property 1, from disk
+                assertTrue( property.next() );
+                assertEquals( propToken1, property.propertyKey() );
+                assertEquals( property.propertyValue(), stringValue( "hello" ) );
```
chrisvest: Does this imply that the properties contained in the transaction state should always be siphoned out before the persistent state?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
```diff
@@ -358,6 +361,247 @@ public void shouldSeeNewRelationshipPropertyInTransaction() throws Exception
...
+
+        try ( org.neo4j.graphdb.Transaction ignored = graphDb.beginTx() )
+        {
+            assertThat(
+                    graphDb.getRelationshipById( relationshipId ).getProperty( propKey ), equalTo( "world" ) );
+        }
+    }
+
+    @Test
+    public void shouldSeeRemovedPropertyInTransaction() throws Exception
```
chrisvest: Or should _not_ see?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/Operations.java
```diff
@@ -459,10 +459,21 @@ public Value relationshipSetProperty( long relationship, int propertyKey, Value
...
     {
+        acquireExclusiveRelationshipLock( relationship );
         ktx.assertOpen();
-        throw new UnsupportedOperationException();
+        singleRelationship( relationship );
+        Value existingValue = readRelationshipProperty( propertyKey );
+
+        if ( existingValue != NO_VALUE )
+        {
+            //no existing value, we just add it
```
chrisvest: This comment look out of place.

chrisvest (approved):
Couple of comments, but overall it looks good.

PR #11053, opened 2018-02-19
3.1 Fixed bug RE:log rotation during backup recovery

phughk:
Logs are not removed during backup recovery.

This reduces the rotation policy on backups and adds a flag to network code to retain or prune transaction file logs that are being populated during catchup/incremental backup.

### DIFF
--- a/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
@@ -157,7 +157,7 @@ class BackupService
             storeCopier.copyStore(
                     storeCopyRequester,
                     CancellationRequest.NEVER_CANCELLED,
-                    MoveAfterCopy.moveReplaceExisting() );
+                    MoveAfterCopy.moveReplaceExisting(), false );
 
             bumpDebugDotLogFileVersion( targetDirectory, timestamp );
             boolean consistent = false;
@@ -187,7 +187,7 @@ class BackupService
             throw new RuntimeException( targetDirectory + " doesn't contain a database" );
         }
 
-        Map<String,String> temporaryDbConfig = getTemporaryDbConfig( config );
+        Map<String,String> temporaryDbConfig = getTemporaryDbConfig();
         config = config.with( temporaryDbConfig );
         try ( PageCache pageCache = createPageCache( new DefaultFileSystemAbstraction(), config ) )
         {
@@ -212,13 +212,12 @@ class BackupService
         }
     }
 
-    private Map<String,String> getTemporaryDbConfig( Config userConfig )
+    private Map<String,String> getTemporaryDbConfig()
     {
         Map<String,String> tempDbConfig = new HashMap<>();
         tempDbConfig.put( OnlineBackupSettings.online_backup_enabled.name(), Settings.FALSE );
-
-        // In case someone deleted the logical log from a full backup
         tempDbConfig.put( GraphDatabaseSettings.keep_logical_logs.name(), "1 txs" );
+        tempDbConfig.put( GraphDatabaseSettings.logical_log_rotation_threshold.name(), "1M" );
         return tempDbConfig;
     }
 
--- a/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
@@ -187,7 +187,7 @@ class BackupService
             throw new RuntimeException( targetDirectory + " doesn't contain a database" );
         }
 
-        Map<String,String> temporaryDbConfig = getTemporaryDbConfig();
+        Map<String,String> temporaryDbConfig = getTemporaryDbConfig( config );
         config = config.with( temporaryDbConfig );
         try ( PageCache pageCache = createPageCache( new DefaultFileSystemAbstraction(), config ) )
         {
@@ -212,12 +212,13 @@ class BackupService
         }
     }
 
-    private Map<String,String> getTemporaryDbConfig()
+    private Map<String,String> getTemporaryDbConfig( Config userConfig )
     {
         Map<String,String> tempDbConfig = new HashMap<>();
         tempDbConfig.put( OnlineBackupSettings.online_backup_enabled.name(), Settings.FALSE );
+
         // In case someone deleted the logical log from a full backup
-        tempDbConfig.put( GraphDatabaseSettings.keep_logical_logs.name(), Settings.TRUE );
+        tempDbConfig.put( GraphDatabaseSettings.keep_logical_logs.name(), "1 txs" );
         return tempDbConfig;
     }
 
--- a/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
+++ b/enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
@@ -217,7 +217,6 @@ class BackupService
         Map<String,String> tempDbConfig = new HashMap<>();
         tempDbConfig.put( OnlineBackupSettings.online_backup_enabled.name(), Settings.FALSE );
         tempDbConfig.put( GraphDatabaseSettings.keep_logical_logs.name(), "1 txs" );
-        tempDbConfig.put( GraphDatabaseSettings.logical_log_rotation_threshold.name(), "1M" );
         return tempDbConfig;
     }
 
--- a/enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
+++ b/enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
@@ -39,6 +39,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.function.Supplier;
 import java.util.stream.LongStream;
 
 import org.neo4j.commandline.admin.AdminTool;
@@ -47,6 +48,7 @@ import org.neo4j.graphdb.Node;
 import org.neo4j.graphdb.RelationshipType;
 import org.neo4j.graphdb.Transaction;
 import org.neo4j.graphdb.factory.GraphDatabaseSettings;
+import org.neo4j.io.ByteUnit;
 import org.neo4j.io.proc.ProcessUtil;
 import org.neo4j.kernel.configuration.Settings;
 import org.neo4j.kernel.impl.store.format.highlimit.HighLimit;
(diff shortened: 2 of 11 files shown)


enterprise/backup/src/main/java/org/neo4j/backup/BackupService.java
```diff
@@ -211,12 +211,16 @@ BackupOutcome doIncrementalBackup( String sourceHostNameOrIp, int sourcePort, Fi
...
+    private Map<String,String> getTemporaryDbConfig( Config userConfig )
     {
         Map<String,String> tempDbConfig = new HashMap<>();
         tempDbConfig.put( OnlineBackupSettings.online_backup_enabled.name(), Settings.FALSE );
+
         // In case someone deleted the logical log from a full backup
-        tempDbConfig.put( GraphDatabaseSettings.keep_logical_logs.name(), Settings.TRUE );
+        if ( Settings.FALSE.equals( userConfig.get( GraphDatabaseSettings.keep_logical_logs ) ) )
+        {
+            tempDbConfig.put( GraphDatabaseSettings.keep_logical_logs.name(), Settings.TRUE );
```
phughk: `1tx`

MishaDemianenko (changes requested):
We need to have a test that confirms that now it's working as intended

enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyClient.java
```diff
@@ -171,7 +172,7 @@ public StoreCopyClient( File storeDir, Config config, Iterable<KernelExtensionFa
     }
 
     public void copyStore( StoreCopyRequester requester, CancellationRequest cancellationRequest,
-                           MoveAfterCopy moveAfterCopy ) throws Exception
+                           MoveAfterCopy moveAfterCopy, boolean isProduction ) throws Exception
```
MishaDemianenko: i think its super confusing flag name, please named it `keepTransactionLogs` instead or something

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
```diff
@@ -142,6 +158,62 @@ public void makeSureBackupCanBePerformedWithCustomPort() throws Exception
...
+                runBackupToolFromOtherJvmToGetExitCode(
+                        "--from", address,
+                        "--cc-report-dir=" + backupDir,
+                        "--backup-dir=" + backupDir,
+                        "--name=" + backupName ) );
+
+        // and the database contains a few more transactions
+        long megabytes = 1024 * 1024;
+        long numberOfBytesInTx = 100;
+        long defaultLogRotationPolicyM = 250; // TODO fine tune
```
MishaDemianenko: this todo should be fixed

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
```diff
@@ -142,6 +158,62 @@ public void makeSureBackupCanBePerformedWithCustomPort() throws Exception
...
+        assertEquals(
+                0,
+                runBackupToolFromOtherJvmToGetExitCode(
+                        "--from", address,
+                        "--cc-report-dir=" + backupDir,
+                        "--backup-dir=" + backupDir,
+                        "--name=" + backupName ) );
+
+        // and the database contains a few more transactions
+        long megabytes = 1024 * 1024;
```
MishaDemianenko: `ByteUnit.mebiBytes( 1 )`

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
```diff
@@ -142,6 +158,62 @@ public void makeSureBackupCanBePerformedWithCustomPort() throws Exception
...
+        long numberOfBytesInTx = 100;
+        long defaultLogRotationPolicyM = 250; // TODO fine tune
+        LongStream.range( 0, defaultLogRotationPolicyM ).forEach( number -> createSomeData( db, megabytes / numberOfBytesInTx ) );
```
MishaDemianenko: because of this line test runs for ages, can we make it simpler by running tx that will generate logs that are bigger than expected and then checking that those were removed?. For example, you can insert quite big properties (10M for example) and those will generate quite big tx log files.

MishaDemianenko (changes requested):

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
phughk: For whatever reason there are issues with that and also from my testing it doesn't seem to be faster (even though it should). Basically this iteration runs 26s as is or 30s with 10x larger txs but 25 iterations. Also removed config as the server setup is irrelevant, and we haven't modified backup client with an additional client to have rotation of 1tx/1Mb.

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
phughk: Addressed

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
phughk: Addressed - not changed values as they are optimal

enterprise/com/src/main/java/org/neo4j/com/storecopy/StoreCopyClient.java
phughk: Addressed

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
```diff
@@ -142,6 +158,56 @@ public void makeSureBackupCanBePerformedWithCustomPort() throws Exception
...
+                runBackupToolFromOtherJvmToGetExitCode(
+                        "--from", address,
+                        "--cc-report-dir=" + backupDir,
+                        "--backup-dir=" + backupDir,
+                        "--name=" + backupName ) );
+
+        // and the database contains a few more transactions
+        long aMegabyte = ByteUnit.mebiBytes( 1 );
+        long numberOfBytesInTx = 100;
+        LongStream.range( 0, 205 ).forEach( number -> createSomeData( db, aMegabyte / numberOfBytesInTx ) );
```
phughk, line 184: This does take longer than expected, however there is no quick fix to make this more reasonable; In order to do so would mean modifying the configuration of the embedded database which isn't necessarily what we want to. Additionally rotation is tied to the tx streaming batch size, which is hard coded.

phughk: Increasing tx size = out of memory errors and overall slower;
Reducing required tx for rotation = too difficult for something that should be so simple
Faking trafic/data to simply meet requirements = not possible

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/log/PhysicalLogFile.java
```diff
@@ -140,6 +141,10 @@ public boolean rotationNeeded() throws IOException
          * Whereas channel.size() should be fine, we're safer calling position() due to possibility
          * of this file being memory mapped or whatever.
          */
+        long position = channel.position();
+        long kbytes = position / ByteUnit.kibiBytes( 1 );
+        long mbytes = position / ByteUnit.mebiBytes( 1 );
+        System.out.printf( "position=%d (%dk, %dM), rotateSize=%d\n", position, kbytes, mbytes, rotateAtSize );
```
MishaDemianenko: no printf's please

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
```diff
@@ -142,6 +158,56 @@ public void makeSureBackupCanBePerformedWithCustomPort() throws Exception
         assertEquals( getDbRepresentation(), getBackupDbRepresentation( "customport" ) );
     }
 
+    @Test
+    public void onlyTheLatestTransactionIsKeptAfterIncrementalBackup() throws Exception
+    {
+        // given database exists with data
+        int port =  4446;
+        startDb( "" + port );
```
MishaDemianenko, line 166: be careful with this hardcoded port in more recent versions

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
```diff
@@ -142,6 +158,56 @@ public void makeSureBackupCanBePerformedWithCustomPort() throws Exception
...
+        Collection<File> txLogFiles = transactionFiles( backupLocation );
+        File expectedTxLogFile1 = new File( backupLocation, "neostore.transaction.db.1" );
+        File expectedTxLogFile2 = new File( backupLocation, "neostore.transaction.db.2" );
+        assertEquals( new HashSet<>( Arrays.asList( expectedTxLogFile1, expectedTxLogFile2 ) ), new HashSet<>( txLogFiles ) );
+    }
+
+    private Collection<File> transactionFiles( File dbLocation ) throws IOException
+    {
+        Collection<File> txFiles = new ArrayList<>();
+        DirectoryStream<Path> dirStream = Files.newDirectoryStream( dbLocation.toPath(), "neostore.transaction.db.*" );
```
MishaDemianenko, line 206: you need to close this directory stream

enterprise/com/src/main/java/org/neo4j/com/storecopy/TransactionCommittingResponseUnpacker.java
```diff
@@ -23,6 +23,7 @@
 import org.neo4j.com.TransactionStream;
 import org.neo4j.com.TransactionStreamResponse;
 import org.neo4j.graphdb.DependencyResolver;
+import org.neo4j.graphdb.factory.GraphDatabaseSettings;
```
MishaDemianenko: strange to see only import updated

MishaDemianenko (changes requested):

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
phughk: Yes, merge resolution changes it to use PortAuthority in 3.3

enterprise/backup/src/test/java/org/neo4j/backup/OnlineBackupCommandIT.java
phughk: Addressed

MishaDemianenko (approved):

phughk:
There is a different solution that may be backported https://github.com/neo4j/neo4j/pull/11647 .

PR #11054, opened 2018-02-19
3.2 Fixed bug RE:log rotation during backup recovery

phughk:


phughk:
There is a different solution that may be backported https://github.com/neo4j/neo4j/pull/11647 .

PR #11062, opened 2018-02-20
Make Bolt Server use thread pools

ali-ince:
Bolt Server originally handles bolt connections on a dedicated thread per connection. This behaviour may yield high number of threads being created, especially on high workloads. As part of bolt server resource management work, this PR changes how bolt connections are handled.

Thread pools are created per bolt connector, which means if there're more than one bolt connectors defined they will be isolated from each other and not race against. Thread pools are configurable with four config keys valid on the bolt connector level; 

| Key | Description | Default |
|:--- |:--- |:--- |
| thread_pool_core_size      | The number of threads to keep in the thread pool bound to this connector, even if they are idle. | 10 |
| thread_pool_max_size | The maximum number of threads allowed in the thread pool bound to this connector.      | 400 |
| thread_pool_keep_alive | The maximum time an idle thread in the thread pool bound to this connector will wait for new tasks. | 5m |
| thread_pool_queue_size | The queue size of the thread pool bound to this connector (-1 for unbounded, 0 for direct handoff, > 0 for bounded) | 0 |

If there isn't any available thread to serve a client's request, a `FAILURE` message is generated with `NoThreadsAvailable` status code. 

When it is detected that there's an active transaction or statement, that connection sticks to the executing thread until this condition is determined to be `false`. By sticking, it will not release the thread to do other work and continues polling jobs from the job queue. **_This is not the ideal behaviour and will be addressed sooner._**

A couple of resource clean-up efforts are also introduced in `TransactionStateMachine`, in an effort to close `TransactionalContext` and `Statement` instances upon successful / failed completion.

Also added another set of metric counters and monitor methods to achieve a bit more detailed bolt connection metrics.

TODOs:
1. Schedule `validateTransaction` calls to the `BoltStateMachine` when connection is sitting without any messages being received,
2. Refactor throttle implementations to sit together at different levels ( i.e. channel & connection ),
3. Add metrics support for monitoring thread pool sizes.


enterprise/metrics/src/main/java/org/neo4j/metrics/source/db/PageCacheMetrics.java
```diff
@@ -33,7 +33,7 @@
 {
     private static final String PAGE_CACHE_PREFIX = "neo4j.page_cache";
 
-    @Documented( "The total number of exceptions seen during the eviction process in the page cache" )
+    @Documented( "The total number of exceptions seen during the eviction processNextBatch in the page cache" )
```
chrisvest: ಠ_ಠ

packaging/standalone/standalone-community/src/main/distribution/text/community/conf/neo4j.conf
```diff
@@ -261,7 +261,7 @@ dbms.jvm.additional=-XX:+UseG1GC
 dbms.jvm.additional=-XX:-OmitStackTraceInFastThrow
 
 # Make sure that `initmemory` is not only allocated, but committed to
-# the process, before starting the database. This reduces memory
+# the processNextBatch, before starting the database. This reduces memory
```
chrisvest: 🤔

enterprise/metrics/src/main/java/org/neo4j/metrics/source/db/PageCacheMetrics.java
ali-ince: thanks @chrisvest pointing at, on it.

community/bolt/src/main/java/org/neo4j/bolt/runtime/BoltConnection.java
```diff
@@ -0,0 +1,104 @@
...
+    void enqueue( Job job );
+
+    /**
+     * Executes a batch of queued jobs, which is executed in an another thread (which is part of a thread pool)
+     *
+     * @return false when no more processing should take place (i.e. connection is closed)
+     */
+    boolean processNextBatch();
+
+    void handleSchedulingError( Throwable t );
```
lutovich: Would be nice to javadoc this method as well

community/bolt/src/main/java/org/neo4j/bolt/runtime/BoltConnectionQueueMonitorAggregate.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.neo4j.bolt.v1.runtime.Job;
+
+public class BoltConnectionQueueMonitorAggregate implements BoltConnectionQueueMonitor
+{
+    private final List<BoltConnectionQueueMonitor> monitors;
+
+    public BoltConnectionQueueMonitorAggregate( BoltConnectionQueueMonitor... monitors )
+    {
+        this.monitors = new ArrayList<>( Arrays.asList( monitors ) );
```
lutovich: Can this be just `this.monitors = Arrays.asList( monitors );`?

community/bolt/src/main/java/org/neo4j/bolt/runtime/BoltConnectionReadLimiter.java
```diff
@@ -75,25 +73,21 @@ protected int getHighWatermark()
     }
 
     @Override
-    public void enqueued( Job job )
+    public void enqueued( BoltConnection to, Job job )
     {
-        checkLimitsOnEnqueue( queueSize.incrementAndGet() );
+        checkLimitsOnEnqueue( to, counters.compute( to.id(), ( k, v ) -> v == null ? new AtomicInteger( 0 ) : v ).incrementAndGet() );
```
lutovich: `Map#computeIfAbsent()` could be used instead to make code a bit shorter

community/bolt/src/main/java/org/neo4j/bolt/runtime/BoltConnectionReadLimiter.java
```diff
@@ -75,25 +73,21 @@ protected int getHighWatermark()
...
     }
 
     @Override
-    public void dequeued( Job job )
+    public void drained( BoltConnection from, Collection<Job> batch )
     {
-        checkLimitsOnDequeue( queueSize.decrementAndGet() );
+        checkLimitsOnDequeue( from, counters.compute( from.id(), ( k, v ) -> v == null ? new AtomicInteger( 0 ) : v ).addAndGet( -batch.size() ) );
```
lutovich: Consider using `Map#computeIfAbsent()`

community/bolt/src/main/java/org/neo4j/bolt/runtime/CachedThreadPoolExecutorFactory.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+        catch ( InterruptedException ex )
+        {
+            executor.shutdownNow();
+            Thread.currentThread().interrupt();
+        }
+    }
+
+    private static BlockingQueue createTaskQueue( int queueSize )
```
lutovich: `BlockingQueue<Runnable>` to avoid raw types

community/bolt/src/main/java/org/neo4j/bolt/runtime/CachedThreadPoolExecutorFactory.java
```diff
@@ -0,0 +1,122 @@
...
+            throw new IllegalArgumentException(
+                    String.format( "The passed executor should already be created by '%s#create()'.", CachedThreadPoolExecutorFactory.class.getName() ) );
+        }
+
+        executor.shutdown();
+        try
+        {
+            if ( !executor.awaitTermination( 10, TimeUnit.SECONDS ) )
+            {
+                executor.shutdownNow();
```
lutovich: This will interrupt worker threads that might be executing transactions, right?
Such interrupts can be harmful, for example blocking IO calls will fail (writing to WAL or Lucene).
Maybe it could just call `executor.shutdown()` and then rely on `NeoStoreDataSource` to terminate all transactions later during lifecycle shutdown?

community/bolt/src/main/java/org/neo4j/bolt/runtime/BoltConnection.java
```diff
@@ -0,0 +1,104 @@
...
+     * @return netty channel
+     */
+    Channel channel();
+
+    /**
+     * Returns the principal the client used to connect
+     *
+     * @return user name
+     */
+    String principal();
```
lutovich: Method is only used in tests. Maybe it can be removed or there are future plans for it?

community/bolt/src/main/java/org/neo4j/bolt/runtime/ExecutorBoltScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+        {
+            connection.handleSchedulingError( ex );
+        }
+    }
+
+    private boolean executeBatch( BoltConnection connection )
+    {
+        Thread currentThread = Thread.currentThread();
+        String originalName = currentThread.getName();
+        String newName = String.format( "%s [%s] ", originalName, connection.remoteAddress(), connector );
```
lutovich: Pattern does not match the arguments

community/bolt/src/main/java/org/neo4j/bolt/runtime/ExecutorBoltSchedulerProvider.java
```diff
@@ -0,0 +1,90 @@
...
+                            config.get( connector.thread_pool_queue_size ), forkJoinThreadPool );
+            boltScheduler.start();
+            boltSchedulers.put( connector.key(), boltScheduler );
+        } );
+    }
+
+    @Override
+    public void stop() throws Throwable
+    {
+        boltSchedulers.values().forEach( s -> s.stop() );
```
lutovich: Would be nice to handle `s.stop()` errors so that if one `BoltScheduler` fails to stop we'll still try to shutdown others.

community/bolt/src/main/java/org/neo4j/bolt/runtime/ExecutorBoltScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+        if ( threadPool != null )
+        {
+            executorFactory.destroy( threadPool );
+        }
+    }
+
+    @Override
+    public void created( BoltConnection connection )
+    {
+        activeConnections.put( connection.id(), connection );
```
lutovich: Maybe assert that `#put()` returns null

community/bolt/src/main/java/org/neo4j/bolt/runtime/DefaultBoltConnection.java
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    @Override
+    public void stop()
+    {
+        if ( shouldClose.compareAndSet( false, true ) )
+        {
+            machine.terminate();
+
+            if ( !hasPendingJobs() )
```
lutovich: `#close()` could probably be called instead of this if statement

community/bolt/src/main/java/org/neo4j/bolt/runtime/ExecutorBoltScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    @Override
+    public void closed( BoltConnection connection )
+    {
+        String id = connection.id();
+
+        try
+        {
+            CompletableFuture currentFuture = activeWorkItems.remove( id );
```
lutovich: `CompletableFuture<Boolean>` to avoid raw types

community/bolt/src/main/java/org/neo4j/bolt/runtime/ExecutorBoltScheduler.java
```diff
@@ -0,0 +1,204 @@
...
+        {
+            return connection.processNextBatch();
+        }
+        finally
+        {
+            currentThread.setName( originalName );
+        }
+    }
+
+    private void handleCompletion( BoltConnection connection, Object shouldContinueScheduling, Throwable error )
```
lutovich: `shouldContinueScheduling` param can have type `Boolean` to avoid casts

community/bolt/src/main/java/org/neo4j/bolt/runtime/ExecutorFactory.java
```diff
@@ -0,0 +1,33 @@
...
+import java.time.Duration;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadFactory;
+
+public interface ExecutorFactory
+{
+
+    ExecutorService create( int corePoolSize, int maxPoolSize, Duration keepAlive, int queueSize, ThreadFactory threadFactory );
+
+    void destroy( ExecutorService executor );
```
lutovich: This method can probably be removed if `ExecutorBoltScheduler` only calls `#shutdown()`

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachineSPI.java
```diff
@@ -149,21 +150,30 @@ public BoltResult start() throws KernelException
                 }
                 catch ( KernelException e )
                 {
+                    transactionalContext.close( false );
```
lutovich: Maybe just `close(false)`?

community/bolt/src/main/java/org/neo4j/bolt/v1/runtime/TransactionStateMachineSPI.java
```diff
@@ -149,21 +150,30 @@ public BoltResult start() throws KernelException
...
                     onFail.apply();
                     throw new QueryExecutionKernelException( e );
                 }
                 catch ( Throwable e )
                 {
+                    transactionalContext.close( false );
```
lutovich: `close(false)`?

community/kernel/src/main/java/org/neo4j/graphdb/factory/GraphDatabaseSettings.java
```diff
@@ -784,6 +784,31 @@
     public static final Setting<File> bolt_log_filename = derivedSetting( "unsupported.dbms.logs.bolt.path",
             GraphDatabaseSettings.logs_directory, logsDir -> new File( logsDir, "bolt.log" ), PATH );
 
+    @Description( "The number of threads that will be created upon initialisation to handle standard bolt requests." )
+    @Internal
+    public static final Setting<Integer> bolt_thread_pool_std_core_size =
```
lutovich: New settings in this class are unused and can probably be removed

lutovich (commented):
@ali-ince review round completed.

I think `BoltConnection` <--> `BoltScheduler` interaction could be improved. Right now they have kind of a cyclic interaction. Connection basically tells scheduler to invoke itself in a separate thread. It would be nice to make connection submit executable jobs to the scheduler instead. Maybe `#processNextBatch()` could become some kind of `Callable`.

Other significant thing is `CachedThreadPoolExecutorFactory#destroy()` which interrupts threads. It could probably just do `executor#shutdown()`.

Rest of the comments are rather small. All fixes can be made in a separate PR(s) after this one is merged.

community/bolt/src/main/java/org/neo4j/bolt/runtime/BoltConnection.java
ali-ince: It comes from the initial design where selecting target scheduler based on principal or port. It now doesn't make sense at all.

community/bolt/src/main/java/org/neo4j/bolt/runtime/DefaultBoltConnection.java
ali-ince: I think it would be even better not to break the linearity of operations on a bolt connection, and changed this to enqueue an empty job to the queue and have it executed by the thread pool. WDYT?

ali-ince:
Thanks @lutovich for your comments. Addressed all except improving the `BoltConnection` <--> `BoltScheduler` interaction, which I'll work on a seperate PR.

PR #11064, opened 2018-02-20
Handle active tx state from Cypher

fickludd:
To support the 2-layer tx-state, this PR injects a new logical plan operator: `ActiveRead`. 

This logical plan operator ensures that all source reads are performed on
the active tx layer. This means that any writes are immediately visible
without the need for a Transaction#markAsStable call, and is used in
the read part of merge plans to ensure that different rows in the same
merge clause see previous writes.

Further the `Eager` operator is augmented to mark the transaction as stable after buffering all source rows.

### DIFF
--- a/community/cypher/cypher-logical-plans-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/logical/plans/Eager.scala
+++ b/community/cypher/cypher-logical-plans-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/logical/plans/Eager.scala
@@ -22,7 +22,7 @@ package org.neo4j.cypher.internal.v3_4.logical.plans
 import org.neo4j.cypher.internal.util.v3_4.attribution.IdGen
 
 /**
-  * Consumes and buffers all source rows, and then produces them.
+  * Consumes and buffers all source rows, marks the transaction as stable, and then produces all rows.
   */
 case class Eager(source: LogicalPlan)(implicit idGen: IdGen) extends LogicalPlan(idGen) with EagerLogicalPlan {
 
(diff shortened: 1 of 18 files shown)


community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/PlanUpdates.scala
```diff
@@ -80,8 +80,10 @@ case object PlanUpdates
 
       //CREATE ()
       case p: CreateNodePattern => context.logicalPlanProducer.planCreateNode(source, p, context)
+
       //CREATE (a)-[:R]->(b)
       case p: CreateRelationshipPattern => context.logicalPlanProducer.planCreateRelationship(source, p, context)
+
```
pontusmelke, line 86: You are wasting white space :)

community/cypher/cypher-logical-plans-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/logical/plans/Eager.scala
```diff
@@ -22,7 +22,7 @@ package org.neo4j.cypher.internal.v3_4.logical.plans
 import org.neo4j.cypher.internal.util.v3_4.attribution.IdGen
 
 /**
-  * Consumes and buffers all source rows, and then produces them.
+  * Consumes and buffers all source rows, marks the transaction as stable, and then produces all rows.
```
pontusmelke, line 25: Out of curiosity, why do we need to do this?

pontusmelke (approved):
🥇

community/cypher/cypher-planner-3.4/src/main/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/logical/PlanUpdates.scala
fickludd: I'm buying developer time :) (less time spent understanding wall of code)

community/cypher/cypher-logical-plans-3.4/src/main/scala/org/neo4j/cypher/internal/v3_4/logical/plans/Eager.scala
fickludd: Well.

We will use the Eager to ensure that all writes from a write clause are executed before continuing the execution. This done so that we can guarantee that they are visible in later reads. However, with the two-layer tx-state they will per definition not be visible unless we also move these writes into the stable state, so that's why we need to call `markAsStable` here.

PR #11065, opened 2018-02-20
Primitivization and dual state of transaction state collections

MishaDemianenko:
Switch transaction state collections to be primitive.
Introduce primitive set and map that allows dual state: active and stable.
Active - actively changing state (actual state).
Stable - state of the collection on some particular moment.


MishaDemianenko:
As we find out cypher can't really use dual state tx state, closing this experiment.

PR #11068, opened 2018-02-21
3.4 Fixed backup to retain 1 tx in tx log files

phughk:


MishaDemianenko (changes requested):
We need to have a test that confirms that now its working.

phughk:
LogicalTransactionStore is the interface for interacting with a tx file; Need to look into prune monitor maybe

PR #11069, opened 2018-02-21
RelationshipGroupCursor#count honor tx state

pontusmelke:
- Update RelationshipGroupCursor count methods to honor the transaction state.
- Update interpreted runtime(s)
- Update `NodeProxy`
- Update compiled runtime

Builds on top of https://github.com/neo4j/neo4j/pull/11050 starts with 2cc4ebc


community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultRelationshipGroupCursor.java
```diff
@@ -183,11 +184,22 @@ public int relationshipLabel()
...
+        }
+        else
         {
-            return bufferedGroup.outgoingCount;
+            count = count( outgoingRawId() );
         }
-        return count( outgoingRawId() );
+        return read.hasTxStateWithChanges()
+               ? read.txState().getNodeState( getOwningNode() )
+                       .augmentDegree( Direction.OUTGOING, count, getType() ) : count;
```
fickludd: Is this going to be correct if there is a loop added in the transaction as well? I'm suspecting `Direction.OUTGOING` is the graphdb semantics which should include loops.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultRelationshipGroupCursor.java
```diff
@@ -205,11 +205,22 @@ else if ( isBuffered() )
...
+        }
+        else
         {
-            return bufferedGroup.incomingCount;
+            count = count( incomingRawId() );
         }
-        return count( incomingRawId() );
+        return read.hasTxStateWithChanges()
+               ? read.txState().getNodeState( getOwningNode() )
+                       .augmentDegree( Direction.INCOMING, count, getType() ) : count;
```
fickludd: Same as for outgoing, I think this augment will also include loops.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_4/codegen/Methods.scala
```diff
@@ -105,9 +102,6 @@ object Methods {
   val nodeExists: MethodReference = method[Read, Boolean]("nodeExists", typeRef[Long])
   val countsForNode: MethodReference = method[ReadOperations, Long]("countsForNode", typeRef[Int])
   val countsForRel: MethodReference = method[ReadOperations, Long]("countsForRelationship", typeRef[Int], typeRef[Int], typeRef[Int])
-  val relationshipGetProperty: MethodReference = method[ReadOperations, Value]("relationshipGetProperty", typeRef[Long], typeRef[Int])
-  val nodesGetForLabel: MethodReference = method[ReadOperations, PrimitiveLongResourceIterator]("nodesGetForLabel", typeRef[Int])
-  val nodeHasLabel: MethodReference = method[ReadOperations, Boolean]("nodeHasLabel", typeRef[Long], typeRef[Int])
```
fickludd, line 110: Win!

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/RelationshipGroupCursor.java
```diff
@@ -78,4 +78,6 @@ default int totalCount()
     long incomingReference();
 
     long loopsReference();
+
+    int type();
```
fickludd: Maybe you missed `relationshipLabel`? I agree this is better, let's change to this.

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/Nodes.java
```diff
@@ -0,0 +1,162 @@
...
+ */
+package org.neo4j.internal.kernel.api.helpers;
+
+import org.neo4j.internal.kernel.api.NodeCursor;
+import org.neo4j.internal.kernel.api.RelationshipGroupCursor;
+
+/**
+ * Helper methods for working with nodes
+ */
+public final class Nodes
```
fickludd, line 28: This class is very helpful!

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/helpers/Nodes.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    public static int countOutgoing( NodeCursor nodeCursor, RelationshipGroupCursor group, int type )
+    {
+        nodeCursor.relationships( group );
+        int count = 0;
+        while ( group.next() )
+        {
+            if ( group.type() == type )
+            {
+                count += group.outgoingCount() + group.loopCount();
```
fickludd: We will only encounter the type in one group, so we can return once we found it. Same thing below.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_4/codegen/Methods.scala
```diff
@@ -60,15 +60,13 @@ object Methods {
...
-                                                                                                             typeRef[ReadOperations],
                                                                                                              typeRef[Read],
                                                                                                              typeRef[CursorFactory],
                                                                                                              typeRef[NodeCursor],
                                                                                                              typeRef[Long],
                                                                                                              typeRef[Direction],
                                                                                                              typeRef[Long])
   val connectingRelationships: MethodReference = method[CompiledExpandUtils, RelationshipSelectionCursor]("connectingRelationships",
-                                                                                                          typeRef[ReadOperations],
```
fickludd, line 74: Wow, this is pretty big. Does that mean that the compiled runtime is fully ported?

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/helpers/NodesTest.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Test
+    public void shouldCountOutgoingWithType()
+    {
+        // Given
+        StubGroupCursor groupCursor = new StubGroupCursor(
+                group( 1 ).withOutCount( 1 ).withInCount( 1 ).withLoopCount( 5 ),
+                group( 2 ).withOutCount( 1 ).withInCount( 1 ).withLoopCount( 3 ),
+                group( 1 ).withOutCount( 2 ).withInCount( 1 ).withLoopCount( 2 ),
```
fickludd: Like I commented above, I think you can only get one group per type and node.

community/kernel/src/main/java/org/neo4j/kernel/impl/core/NodeProxy.java
```diff
@@ -679,52 +680,72 @@ public int getDegree()
...
+        {   // This type doesn't even exist. Return 0
+            return 0;
         }
-        catch ( EntityNotFoundException e )
+
+        try ( Statement ignore = transaction.acquireStatement() )
         {
-            throw new NotFoundException( "Node not found.", e );
+            NodeCursor nodes = transaction.nodeCursor();
+            singleNode( transaction, nodes );
```
fickludd, line 739: I really like how `singleNode` unifies the `NotFoundException` handling in this class!

community/cypher/cypher/src/main/java/org/neo4j/cypher/internal/codegen/CompiledExpandUtils.java
```diff
@@ -141,12 +141,12 @@ static int nodeGetDegree( Read read, long node, NodeCursor nodeCursor, Relations
...
     {
-            read.singleNode( node, nodeCursor );
-            if ( !nodeCursor.next() )
-            {
-                return 0;
+        read.singleNode( node, nodeCursor );
+        if ( !nodeCursor.next() )
+        {
+            return 0;
             }
```
fickludd: Better, but the indentation still seems awkward after the `return 0;`

community/kernel-api/src/test/java/org/neo4j/internal/kernel/api/RelationshipTransactionStateTestBase.java
```diff
@@ -682,6 +682,114 @@ public void shouldCountFromTxState() throws Exception
         } );
     }
 
+    @Test
+    public void groupCursorShouldSeeNewTypes() throws Exception
```
fickludd: Oh great, I didn't know you already fixed this!

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultRelationshipGroupCursor.java
```diff
@@ -124,30 +138,93 @@ public boolean next()
...
-                return false;
+                //We have now run out of groups from the store, however there may still
+                //be new types that was added in the transaction that we haven't visited yet.
+                return nextFromTxState();
             }
             read.group( this, getNext(), page );
         } while ( !inUse() );
 
+        markTypeAsSeen( type() );
         return true;
```
fickludd, line 170: This `next()` method turned out pretty clean, nice work.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultRelationshipGroupCursor.java
```diff
@@ -89,16 +98,21 @@ else if ( edge.getSecondNode() == nodeReference ) // incoming
                     throw new IllegalStateException( "not a part of the chain! TODO: better exception" );
                 }
             }
+            //we never need to check tx state, handled by traversal
+            this.hasCheckedTxState = true;
```
fickludd: Aha, that's right! I didn't think about that before.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultRelationshipGroupCursor.java
```diff
@@ -124,30 +138,93 @@ public boolean next()
...
+    {
+        if ( txTypeIterator == null && !txTypes.isEmpty() )
+        {
+            txTypeIterator = txTypes.iterator();
+            //here it may be tempting to do txTypes.clear()
+            //however that will also clear the iterator
+        }
+        if ( txTypeIterator != null && txTypeIterator.hasNext() )
+        {
+            setType( txTypeIterator.next() );
```
fickludd, line 192: When reading this I had a spider sense tingling. I wonder if there is some problem with state (the relationship pointers) leaking from a previous in-store group when we switch to tx-state groups. WDYT about adding some existing group to the test case, or making a separate test case where there is both existing and added groups?

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultRelationshipGroupCursor.java
```diff
@@ -49,7 +49,7 @@
     private PageCursor edgePage;
     private boolean hasCheckedTxState;
     private PrimitiveIntSet txTypes = Primitive.intSet();
```
fickludd: This could turn `final`, for the lack of anything better to say...

fickludd (approved):
This is pretty awesome. I was going through commit by commit, and found a few real problems which I was going to suggest investigating or solving. However, each of them was fixed and even more in later commits, so I have nothing left to say than: LGTM. Let's get it in once the underlying PR is in place.

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/DefaultRelationshipGroupCursor.java
fickludd: Which I see you already noticed, tested and fixed in the next commit. Nice.

enterprise/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_4/codegen/Methods.scala
pontusmelke: yes 🎆

PR #11071, opened 2018-02-21
Fix default value lookup in config to return the original string

klaren:
Previously we returned the setting with type and called `toString` to get the original value out. This does not work for settings that uses java lang types such as `Duration` where the `toString` cannot be overridden.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
@@ -596,8 +596,8 @@ public class Config implements DiagnosticsProvider, Configuration
 
         synchronized ( params )
         {
-            boolean oldDefault = false;
-            boolean newDefault = false;
+            boolean oldValueIsDefault = false;
+            boolean newValueIsDefault = false;
             String oldValue;
             String newValue;
             if ( update == null || update.isEmpty() )
@@ -607,7 +607,7 @@ public class Config implements DiagnosticsProvider, Configuration
                 boolean hasDefault = overriddenDefault != null;
                 oldValue = hasDefault ? params.put( setting, overriddenDefault ) : params.remove( setting );
                 newValue = getDefaultValueOf( setting );
-                newDefault = true;
+                newValueIsDefault = true;
             }
             else
             {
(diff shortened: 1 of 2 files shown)


community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -587,6 +596,8 @@ public void updateDynamicSetting( String setting, String update )
 
         synchronized ( params )
         {
+            boolean oldDefault = false;
```
MishaDemianenko: it's really hard to guess why oldDefault and newDefault are called like this

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
```diff
@@ -610,10 +622,15 @@ public void updateDynamicSetting( String setting, String update )
...
                 oldValue = getDefaultValueOf( setting );
-                params.put( setting, update );
+                if ( params.put( setting, update ) == null )
+                {
+                    oldDefault = true;
+                }
                 newValue = update;
             }
-            log.info( "Setting changed: '%s' changed from '%s' to '%s'", setting, oldValue, newValue );
+            log.info( "Setting changed: '%s' changed from '%s' to '%s'",
```
MishaDemianenko, line 636: there is no new test for different output, also existent test that checks output now failing

MishaDemianenko (changes requested):

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
klaren: ah, this was backported from 3.4, missed the test...

community/kernel/src/main/java/org/neo4j/kernel/configuration/Config.java
klaren: indeed, renaming

klaren:
@MishaDemianenko new commits ready for your eyes to look at :eyeglasses:

MishaDemianenko (approved):

Issue #6155, opened 2015-12-21
Disabling traversal endpoint is not as simple as removing the class form the classpath


AWinterman:
The traversal api class doesn't appear to ever appear in the classpath, so the directions to disabling it seem to be misleading. (re: http://neo4j.com/docs/stable/security-server.html#_arbitrary_code_execution)

chrisvest:
Hey, sorry it took us so long to pick this up. We're going to add a setting for disabling the traversal endpoint script execution: #11078

chrisvest:
The PR has been merged and the setting will appear in the next releases.

PR #11078, opened 2018-02-21
Traversal endpoint script execution is now disabled by default, by introducing the `unsupported.dbms.security.script_enabled` setting, and making it `false` by default.

chrisvest:
This resolves the #6155 issue.

### DIFF
--- a/community/server/src/main/java/org/neo4j/server/configuration/ServerSettings.java
+++ b/community/server/src/main/java/org/neo4j/server/configuration/ServerSettings.java
@@ -194,7 +194,7 @@ public class ServerSettings implements LoadableConfig
                   "endpoints. This setting is in the 'unsupported' namespace, because the scripting feature might be " +
                   "removed entirely in a future release." )
     public static final Setting<Boolean> script_enabled =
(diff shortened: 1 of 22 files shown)


community/server/src/main/java/org/neo4j/server/configuration/ServerSettings.java
```diff
@@ -189,6 +189,14 @@ private ThirdPartyJaxRsPackage createThirdPartyJaxRsPackage( String packageAndMo
     @Internal
     public static final Setting<URI> browser_path = setting( "unsupported.dbms.uris.browser", Settings.URI, "/browser/" );
 
+    @Deprecated
+    @Description( "Whether to allow executing scripts inside the server, from external sources, e.g. via traversal " +
+                  "endpoints. This setting is in the 'unsupported' namespace, because the scripting feature might be " +
```
MishaDemianenko: i think we should say that this setting will be removed in future releases, like for sure, and also ask people to use cypher or anything else instead

MishaDemianenko (approved):
A small comment about setting description, otherwise approved. Also, we should add it to a list of things that will be removed with next major release right?

PR #11081, opened 2018-02-21
Implement comparability semantics for PointValues

sherfert:


community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
```diff
@@ -238,12 +238,14 @@ else if ( seconds > 0 && nanos < 0 )
     }
 
     @Override
-    public int compareTo( Value otherValue )
+    public int compareTo( DurationValue other )
+    {
+        return unsafeCompareTo( other );
```
fickludd: Here we should flip so that `unsafeCompareTo` calls `compareTo( (DurationValue) otherValue )`, and `compareTo` has the logic.

community/values/src/main/java/org/neo4j/values/storable/PointValue.java
```diff
@@ -107,12 +107,14 @@ public boolean eq( Object other )
     }
 
     @Override
-    public int compareTo( Value otherValue )
+    public int compareTo( PointValue other )
+    {
+        return unsafeCompareTo(other);
```
fickludd: Same as above, we can invert who calls who between `compareTo` and `unsafeCompareTo`

fickludd (changes requested):
Now it looks really good, I just think we could invert to call-sites and then it's good to go.

fickludd (approved):

PR #11089, opened 2018-02-22
Refactor topology discovery service tests

andrewkerr9000:
Refactor some of the testing infrastructure around DiscoveryService, primarily introduce an enum of DiscoveryServiceFactory with constructors.

### DIFF
--- a/enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/ClusterOverviewIT.java
+++ b/enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/ClusterOverviewIT.java
@@ -35,13 +35,14 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import org.neo4j.causalclustering.core.CausalClusteringSettings;
 import org.neo4j.causalclustering.discovery.Cluster;
 import org.neo4j.causalclustering.discovery.ClusterMember;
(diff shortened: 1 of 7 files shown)


enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/ClusterOverviewIT.java
```diff
@@ -171,20 +158,22 @@ public void shouldDiscoverReadReplicasAfterRestartingCores() throws Exception
...
         clusterRule.withNumberOfReadReplicas( 0 );
 
         Cluster cluster = clusterRule.startCluster();
 
         // when
-        cluster.addCoreMemberWithId( 3 ).start();
-        cluster.addCoreMemberWithId( 4 ).start();
+        int extraCoreMembers = 2;
+        int finalCoreMembers = initialCoreMembers + 2;
+        IntStream.range( 0, extraCoreMembers ).forEach( idx -> cluster.addCoreMemberWithId( initialCoreMembers + idx ).start() );
```
hugofirth: If we're going to go this way: "documenting" with variable names, should we factor out / replicate to make it consistent across the other tests in this IT?

hugofirth (approved):
One query about the ClusterOverviewIT, but otherwise looks good to me 👍

enterprise/causal-clustering/src/test/java/org/neo4j/causalclustering/scenarios/ClusterOverviewIT.java
andrewkerr9000: Baby steps... I'll update the other tests just as soon as I care about them :)

andrewkerr9000: although the `2` on line 169 looks like it should be `extraCoreMembers`...

PR #11093, opened 2018-02-23
Decouple scheduled task execution from task scheduling.

chrisvest:
By making the scheduled executor service send the tasks to be executed in their respective pools, instead of the scheduling pool, and instead use the scheduling pool exclusively to trigger said tasks.

This also includes a handful of refactorings around the job scheduling infrastructure. And all threads started by the `CentralJobScheduler` now run within `ThreadGroups` derived from their scheduling `Group`s.


enterprise/kernel/src/main/java/org/neo4j/kernel/impl/pagecache/PageCacheWarmerKernelExtension.java
```diff
@@ -126,13 +126,7 @@ private void tryStartProfile()
     {
         if ( profilingStarted.compareAndSet( false, true ) )
         {
-            // At this point, we are currently executing inside a *scheduled* executor thread. There are, at this time
```
klaren, line 129: :tada:

community/kernel/src/main/java/org/neo4j/kernel/impl/scheduler/CentralJobScheduler.java
```diff
@@ -0,0 +1,187 @@
...
+import org.neo4j.time.Clocks;
+
+import static org.neo4j.kernel.impl.util.DebugUtil.trackTest;
+
+public class CentralJobScheduler extends LifecycleAdapter implements JobScheduler
+{
+    private static final AtomicInteger INSTANCE_COUNTER = new AtomicInteger();
+    private static final Group SCHEDULER_GROUP = new Group( "Scheduler" );
+
+    private TimeBasedTaskScheduler scheduler;
```
klaren: `final`, same for `schedulerThread`

community/kernel/src/main/java/org/neo4j/kernel/impl/scheduler/ThreadPoolManager.java
```diff
@@ -0,0 +1,58 @@
...
+    {
+        ThreadPool threadPool = getThreadPool( group );
+        return threadPool.submit( job );
+    }
+
+    public InterruptedException shutDownAll()
+    {
+        pools.forEach( ( group, pool ) -> pool.cancelAllJobs() );
+        pools.forEach( ( group, pool ) -> pool.shutDown() );
+        return pools.values().stream().reduce( null,
```
klaren: I would prefer to have a simpler map-reduce instead:
```java
return pools.values().stream().map( ThreadPool::getShutdownException ).reduce( null, Exceptions::chain );
```

community/kernel/src/main/java/org/neo4j/kernel/impl/scheduler/TimeBasedTaskScheduler.java
```diff
@@ -0,0 +1,182 @@
...
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.LockSupport;
+
+import org.neo4j.scheduler.JobScheduler.Group;
+import org.neo4j.scheduler.JobScheduler.JobHandle;
+import org.neo4j.time.SystemNanoClock;
+
+final class TimeBasedTaskScheduler implements Runnable
```
klaren, line 31: This class seems overly complicated for what needs to be done. A simple blocking priority queue would surely suffice.

community/kernel/src/main/java/org/neo4j/kernel/impl/scheduler/ScheduledTask.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.neo4j.concurrent.BinaryLatch;
+import org.neo4j.scheduler.JobScheduler;
+import org.neo4j.scheduler.JobScheduler.CancelListener;
+import org.neo4j.scheduler.JobScheduler.JobHandle;
+
+final class ScheduledTask implements JobHandle
```
klaren: A `javadoc` that documents the states and transitions would be useful.

klaren (changes requested):
This is a move in the right direction, nice to get some structure around the scheduler.

Since it feels like my requested changes will change a lot of the code, let's start with these so I don't review code that will go away anyway.

Was there any particular reason for not using `ScheduledThreadPoolExecutor` as a dispatcher for the other thread pools? Exactly the way you solved it for the profiler before.

chrisvest:
> Was there any particular reason for not using ScheduledThreadPoolExecutor as a dispatcher for the other thread pools?

Yes, I tried this at first, but it became very unwieldy to trace cancellations and ExecutionExceptions back and forth, not to mention also handling overdue tasks and avoiding overlapping execution. The dedicated scheduler thread is definitely simpler.

community/kernel/src/main/java/org/neo4j/kernel/impl/scheduler/TimeBasedTaskScheduler.java
chrisvest: You mean a DelayQueue? I don't think it's that complicated for balancing reaction time and proper shut down.

chrisvest: I side-stepped the concurrency questions of the _blocking_ priority queue, and kept the existing concurrency scheme but switched to an ordinary priority queue.

chrisvest:
@klaren I believe I've address all of your comments.

community/kernel/src/main/java/org/neo4j/kernel/impl/scheduler/TimeBasedTaskScheduler.java
klaren: I was referring to the inbox and queue split as well. Can't submit just add it to the queue directly?

If we're having trouble with contention on the submit method, I'm more than sure we don't need to worry about timeliness.

chrisvest: I like that they are split. They do different things as it is now. And one has to deal with multi-threading, while the other is single threaded and carefree.

klaren: I don't see how multiple threads submitting and one thread polling would change that. The timekeeper thread will only work on the first object in the queue each time.

chrisvest:
@klaren Take another look.

community/kernel/src/main/java/org/neo4j/kernel/impl/scheduler/TimeBasedTaskScheduler.java
```diff
@@ -0,0 +1,108 @@
...
+    private final ThreadPoolManager pools;
+    private final PriorityBlockingQueue<ScheduledJobHandle> delayedTasks;
+    private volatile Thread timeKeeper;
+    private volatile boolean stopped;
+
+    TimeBasedTaskScheduler( SystemNanoClock clock, ThreadPoolManager pools )
+    {
+        this.clock = clock;
+        this.pools = pools;
+        delayedTasks = new PriorityBlockingQueue<>( 42, DEADLINE_COMPARATOR );
```
klaren, line 47: While 42 might be the answer to everything, somehow I doubt that includes optimal initial capacity for a blocking priority queue. But I don't what the answer to that would be...

klaren (approved):
I'm happy with the current state! One magic constant to ponder upon before merging.

community/kernel/src/main/java/org/neo4j/kernel/impl/scheduler/TimeBasedTaskScheduler.java
chrisvest: The queue grows automatically, and you have to choose something in order to provide a comparator, so here we are.

PR #11096, opened 2018-02-25
Count store from kernel API

pontusmelke:
Move (or rather copy, but will be removed soon) count store operations to kernel api which allows us to get rid of all dependency on `ReadOperation` from the compiled runtime


community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
```diff
@@ -115,6 +121,37 @@ else if ( txState.nodeIsAddedInThisTx( reference ) )
...
+        {
+            CountsRecordState counts = new CountsRecordState();
+            try
+            {
+                TransactionState txState = ktx.txState();
+                txState.accept( new TransactionCountingStateVisitor( EMPTY, storeReadLayer,
+                        statement, txState, counts ) );
+                if ( counts.hasChanges() )
+                {
+                    count += counts.nodeCount( labelId, newDoubleLongRegister() ).readSecond();
```
SaschaPeukert, line 138: Except for that ```readSecond()``` part, everything else is really good to read. Maybe renaming that method could help to better express what it does?

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundGraphStatistics.scala
```diff
@@ -21,17 +21,19 @@ package org.neo4j.cypher.internal.runtime.interpreted
...
+import org.neo4j.internal.kernel.api.Read
 import org.neo4j.kernel.api.ReadOperations
 import org.neo4j.kernel.api.exceptions.index.IndexNotFoundKernelException
 
 object TransactionBoundGraphStatistics {
-  def apply(ops: ReadOperations) = new StatisticsCompletingGraphStatistics(new BaseTransactionBoundGraphStatistics(ops))
+  def apply(read: Read, ops: ReadOperations) = new StatisticsCompletingGraphStatistics(new BaseTransactionBoundGraphStatistics(read, ops))
 
-  private class BaseTransactionBoundGraphStatistics(operations: ReadOperations) extends GraphStatistics with IndexDescriptorCompatibility {
+  //TODD remove ReadOperations when schema ops are ported
```
SaschaPeukert: Who is Todd? :-D

SaschaPeukert (changes requested):
Looks good! I only have two minor nit-picks

community/kernel/src/main/java/org/neo4j/kernel/impl/newapi/AllStoreHolder.java
pontusmelke: This is the API of the underlying count store, don't want to change that API as part of this PR.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundGraphStatistics.scala
pontusmelke: oh yes TODD, I totally meant to write that :)

SaschaPeukert (approved):

PR #11098, opened 2018-02-26
Spatial index cleanup

OliviaYtterbrink:
Changed to do sampling at SpatialFusion level and got rid of the need to have an init method in `SpatialCRSSchemaIndex`.

### DIFF
--- a/community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/fusion/SpatialFusionIndexPopulatorTest.java
+++ b/community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/fusion/SpatialFusionIndexPopulatorTest.java
@@ -28,10 +28,11 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
-import org.neo4j.internal.kernel.api.schema.LabelSchemaDescriptor;
 import org.neo4j.kernel.api.exceptions.index.IndexEntryConflictException;
 import org.neo4j.kernel.api.index.IndexEntryUpdate;
 import org.neo4j.kernel.api.schema.index.IndexDescriptor;
+import org.neo4j.kernel.api.schema.index.IndexDescriptorFactory;
+import org.neo4j.kernel.impl.api.index.sampling.IndexSamplingConfig;
 import org.neo4j.kernel.impl.index.schema.SpatialCRSSchemaIndex;
 import org.neo4j.values.storable.CoordinateReferenceSystem;
 import org.neo4j.values.storable.PointValue;
@@ -45,7 +46,6 @@ import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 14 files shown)


community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/fusion/SpatialFusionIndexPopulatorTest.java
```diff
@@ -196,47 +198,48 @@ public void markAsFailedMustMarkAll() throws Exception
...
-            fusionIndexPopulator.includeSample( update );
-
-            // then
-            SpatialKnownIndex spatialKnownIndex = indexMap.get( point.getCoordinateReferenceSystem() );
-            verify( spatialKnownIndex ).includeSample( update );
-            reset( spatialKnownIndex );
-        }
-    }
+//    @Test
+//    public void shouldIncludeSampleOnCorrectPopulator() throws Exception
```
burqen: Oops 🙈

enterprise/cypher/acceptance-spec-suite/src/test/scala/org/neo4j/internal/cypher/acceptance/SpatialFunctionsAcceptanceTest.scala
```diff
@@ -19,18 +19,15 @@
...
 import org.neo4j.graphdb.spatial.Point
 import org.neo4j.internal.cypher.acceptance.CypherComparisonSupport._
 import org.neo4j.values.storable.{CoordinateReferenceSystem, Values}
 
 class SpatialFunctionsAcceptanceTest extends ExecutionEngineFunSuite with CypherComparisonSupport {
 
-  val pointConfig = Configs.Interpreted - Configs.Version2_3
-  val equalityConfig = Configs.Interpreted - Configs.OldAndRule
+  val pointConfig: TestConfiguration = Configs.Interpreted - Configs.Version2_3
+  val equalityConfig: TestConfiguration = Configs.Interpreted - Configs.OldAndRule
```
Lojjs: Can you remove type annotation here?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/sampling/IndexSamplingConfig.java
```diff
@@ -30,9 +30,17 @@
...
-        this.sampleSizeLimit = config.get( GraphDatabaseSettings.index_sample_size_limit );
-        this.updateRatio = ((double) config.get( GraphDatabaseSettings.index_sampling_update_percentage )) / 100.0d;
-        this.backgroundSampling = config.get( GraphDatabaseSettings.index_background_sampling_enabled );
+        this( config.get( GraphDatabaseSettings.index_sample_size_limit ),
+              config.get( GraphDatabaseSettings.index_sampling_update_percentage ) / 100.0d,
+              config.get( GraphDatabaseSettings.index_background_sampling_enabled )
+        );
+    }
+
+    public IndexSamplingConfig( int sampleSizeLimit, double updateRatio, boolean backgroundSampling )
```
Lojjs: This new constructor seems only to be used from the one above. What was the reason for splitting it up?

Lojjs (commented):
Lot of good clean-up here! I especially liked that you have got rid of the `NONE` state and make init simpler. Just one super-small comment and one question:

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/sampling/IndexSamplingConfig.java
OliviaYtterbrink: I think it was added to help make some testing easier, but it seems we reverted that part without reverting this. I'll remove it while rebasing.

Lojjs (approved):

PR #11100, opened 2018-02-26
Assertions and logs for server side transaction streaming

martinfurmanski:


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/ChunkedTransactionStream.java
```diff
@@ -96,6 +98,11 @@ else if ( txCursor.next() )
 
             CommittedTransactionRepresentation tx = txCursor.get();
             lastTxId = tx.getCommitEntry().getTxId();
+            if ( lastTxId != expectedTxId )
+            {
+                throw new IllegalStateException( "Transaction cursor out of order." );
```
hugofirth: May be worth providing the expected and actual values of the transaction cursor to help with debugging from the logs?

martinfurmanski: Yes, will fix!

martinfurmanski:
@hugofirth Updated the printout to include the expected and actual values. Also fixed the failing test. PTAL (please take another look)!

hugofirth (approved):
All looks good to me - nothing specific to add.

PR #11102, opened 2018-02-26
Parsing, creating temporal values and selecting them into other temporal values

sherfert:


community/values/src/main/java/org/neo4j/values/storable/DateTimeValue.java
```diff
@@ -131,201 +155,193 @@ public static DateTimeValue truncate(
...
-                throw new UnsupportedOperationException( "not implemented" );
-            }
+                    LocalTime time;
+                    ZoneId zoneId;
+                    if ( selectingTime )
+                    {
+                        AnyValue timeField = fields.get( Field.time );
+                        if ( !(timeField instanceof TemporalValue) )
+                        {
+                            throw new IllegalArgumentException( String.format( "Cannot construct local time from: %s", timeField ) );
```
craigtaverner, line 221: The word 'local' here seems incorrect.

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
```diff
@@ -322,8 +325,43 @@ private static DurationValue parseDuration( Matcher matcher )
...
             return null;
         }
-        long months = optLong( y ) * 12 + optLong( m );
-        long days = optLong( w ) * 7 + optLong( d );
+        int pos;
+        if ( (pos = fractionPoint( y )) >= 0 )
+        {
+            if ( m != null || w != null || d != null || t != null )
+            {
+                return null;
```
craigtaverner, line 333: The regex seems to allow a lot of 'illegal' combinations that can lead to `return null`. How does the user know they have written an illegal combination? Parsing error message would be useful, right?

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
```diff
@@ -416,6 +476,26 @@ private static DurationValue parseDuration(
...
+    private static int fractionPoint( String field )
+    {
+        if ( field == null )
+        {
+            return -1;
+        }
+        int fractionPoint = field.indexOf( '.' );
+        if ( fractionPoint < 0 )
+        {
+            fractionPoint = field.indexOf( ',' );
```
craigtaverner, line 494: Is it safe to be locale neutral here? Does the CIP go to this level of detail also?

community/values/src/main/java/org/neo4j/values/storable/LocalTimeValue.java
```diff
@@ -96,40 +105,56 @@ public static LocalTimeValue truncate(
...
             @Override
-            protected LocalTimeValue selectTime( AnyValue temporal )
+            public LocalTimeValue buildInternal()
             {
-                throw new UnsupportedOperationException( "not implemented" );
+                LocalTime result;
+                if ( fields.containsKey( Field.time ) )
+                {
+                    AnyValue time = fields.get( Field.time );
+                    if ( !(time instanceof TemporalValue) )
```
craigtaverner, line 127: This code pattern seems to be duplicated in many places. I wonder if it is possible to refactor to a common location? (the entire block from fields.containsKey to setting the DEFAULT).

community/values/src/main/java/org/neo4j/values/storable/TemporalValue.java
```diff
@@ -497,228 +580,182 @@ protected final ZoneId timezoneOf( AnyValue timezone )
...
+                    time = new ConstructTime();
+                }
+                time.assign( field, value );
             }
+            return this;
         }
     }
 
-    private abstract static class DateBuilder<Input>
+    private static class SelectDateOrTimeDTBuilder extends DateTimeBuilder
```
craigtaverner, line 701: Would be much easier to have separate builders for Date and Time, as the current DateOrTime seems convoluted and duplicates checks in the code calling it, and the code inside it.

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/DurationBetweenAcceptance.feature
```diff
@@ -0,0 +1,383 @@
...
+              //duration.between(datetime("2014-07-21T21:40:36.143+0200"), date("2015-06-24")),
+              duration.between(datetime("2014-07-21T21:40:36.143+0200"), localdatetime("2016-07-21T21:45:22.142")),
+              duration.between(datetime("2014-07-21T21:40:36.143+0200"), datetime("2015-07-21T21:40:32.142+0100"))
+              //duration.between(datetime("2014-07-21T21:40:36.143+0200"), localtime("16:30")),
+              //duration.between(datetime("2014-07-21T21:40:36.143+0200"), time("16:30+0100"))
+              ] as d
+      RETURN d
+      """
+    Then the result should be, in order:
+      | d |
```
craigtaverner, line 62: Does this mean that no duration support exists yet? We can run these functions with no errors, but also get no durations produced. Seems like silently failing.

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/TemporalArithmeticAcceptance.feature
```diff
@@ -0,0 +1,117 @@
...
+    When executing query:
+      """
+      WITH localdatetime({year:1984, month:10, day:11, hour:12, minute:31, second:14, nanosecond: 645876123}) as x
+      UNWIND [duration({years: 12, months: 5, days: 14, hours: 16, minutes: 12, seconds: 70, nanoseconds: 123456789}),
+              duration({months:1, days: -14, hours: 16, minutes: -12, seconds: 70}),
+              duration({years: 12.5, months: 5.5, days: 14.5, hours: 16.5, minutes: 12.5, seconds: 70.5, nanoseconds: 123456789})] as dur
+      RETURN x+dur, x-dur
+      """
+    Then the result should be, in order:
+      | d            |
```
craigtaverner, line 76: You check for no side effects before this, but not from this test on.

craigtaverner (approved):

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
sherfert: Why would it not be safe? The CIP does not specify this.

enterprise/cypher/acceptance-spec-suite/src/test/resources/cypher/features/DurationBetweenAcceptance.feature
sherfert: Nononononono. We were just to lazy to code the expected result. This will be handled in a different PR.

community/values/src/main/java/org/neo4j/values/storable/TemporalValue.java
sherfert: Actually a DateTimeBuilder is made up of a DateBuilder and a TimeBuilder. It just contains the logic of selecting a whole `datetime`, which neither of the two others could do.

community/values/src/main/java/org/neo4j/values/storable/DurationValue.java
sherfert: It turns out that this will actually never happen. The Regex Matcher will already catch these cases. I will add tests for a useful error message, but not change this behavior here.

community/values/src/main/java/org/neo4j/values/storable/LocalTimeValue.java
sherfert: I agree the implementations of `buildInternal()` are very similar. Since they all have slightly different special cases, I find it hard to actually come up with a good way to generalize this. I am open for suggestions, though!

community/values/src/main/java/org/neo4j/values/storable/TemporalValue.java
sherfert: I will remove the duplicate code here. I agree.

craigtaverner: I'm talking about SelectDateOrTimeDTBuilder which can be deleted.

craigtaverner: OK, I missed the point where the builder contains state and is used in latter assigns. I retract the suggestion.

PR #11103, opened 2018-02-26
Remove TypeSafeMathSupport

pontusmelke:


community/values/src/main/java/org/neo4j/values/storable/NumberValue.java
```diff
@@ -137,4 +142,20 @@ else if ( numberValue instanceof FloatingPointValue )
...
+        {
+            return dividedBy( numberValue.longValue() );
+        }
+        else if ( numberValue instanceof FloatingPointValue )
+        {
+            return dividedBy( numberValue.doubleValue() );
+        }
+        else
+        {
+            throw new IllegalArgumentException( "Cannot add " + numberValue );
```
fickludd: This error msg (and similar for `times` and `minus`) seem to have copy paste artifacts.

fickludd (approved):
Very nice. Found a small thing. Like how this plays out.

community/values/src/main/java/org/neo4j/values/storable/NumberValue.java
pontusmelke: chocking 😱

PR #11111, opened 2018-02-27
 toString functionality for temporal and duration

Lojjs:


community/values/src/main/java/org/neo4j/values/storable/TemporalValue.java
```diff
@@ -172,12 +172,6 @@ public final NumberType numberType()
         return NO_NUMBER;
     }
 
-    @Override
-    public final String toString()
-    {
-        return getClass().getSimpleName() + "<" + prettyPrint() + ">";
```
sherfert, line 182: If the implementation is always `prettyPrint()`, why not just change it here in the superclass?

enterprise/cypher/spec-suite-tools/src/test/scala/cypher/features/Neo4jExceptionToExecutionFailed.scala
```diff
@@ -32,8 +32,8 @@ object Neo4jExceptionToExecutionFailed {
 
   def convert(phase: String, t: Throwable): ExecutionFailed = {
     val neo4jException = t match {
-      case re: RuntimeException => re
```
sherfert: That should not have been comitted.

community/values/src/main/java/org/neo4j/values/storable/TemporalValue.java
Lojjs: That is a good point, have changed it now

enterprise/cypher/spec-suite-tools/src/test/scala/cypher/features/Neo4jExceptionToExecutionFailed.scala
Lojjs: Oops

craigtaverner (approved):

PR #11113, opened 2018-02-27
Fixes bug with unsupported default value for explicit.remove procedures

craigtaverner:
The neo4j documentation and the original design indicates that is should be possible to not pass the third parameter to removeNode and removeRelationship procedures, however that did not work. The implementation had a misunderstanding of how optional parameters worked, and there were no tests for that specific case. This PR fixes that behaviour so the third parameter is now truly optional.

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
@@ -574,12 +574,14 @@ public class BuiltInProcedures
         return Stream.of( new BooleanResult( true ) );
     }
 
+    private static final String DEFAULT_KEY = " <[9895b15e-8693-4a21-a58b-4b7b87e09b8e]> ";
+
     @Description( "Remove a node from an explicit index with an optional key" )
     @Procedure( name = "db.index.explicit.removeNode", mode = WRITE )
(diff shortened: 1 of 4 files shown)


community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
```diff
@@ -574,12 +574,14 @@ public void resampleOutdatedIndexes()
         return Stream.of( new BooleanResult( true ) );
     }
 
+    private static final String DEFAULT_KEY = " <[9895b15e-8693-4a21-a58b-4b7b87e09b8e]> ";
```
OliviaYtterbrink, line 577: Why create a new default key here, instead of using the default one in `Name`?

OliviaYtterbrink (commented):
This looks good, just one minor question.

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
craigtaverner: Using the one in Name was what we had before, and it did not work. The code interprets this as a missing value, not an optional value, with the result that we get the exception we get. It has to differ from that one, but be as unlikely to be used as a property key.

OliviaYtterbrink (approved):

PR #11118, opened 2018-02-27
Native index consider IOLimit on flush

burqen:
During checkpoint, an io limit is imposed when flushing store files to not be to disruptive for normal operations. This io limit was not respected by indexes living on top of page cache, which were all flushed without any limit. This is not longer the case.


MishaDemianenko (approved):
Everything except tests compilation problems looks good.
Will approve in advance, but please fix those before merging!

PR #11119, opened 2018-02-27
Support graph properties in Kernel API

pontusmelke:
- Read graph properties including tx state
- setting graph properties
- removing graph properties
- update `GraphPropertiesProxy` to use Kernel API

### DIFF
--- a/community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/StoreStatement.java
+++ b/community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/StoreStatement.java
@@ -310,7 +310,7 @@ public class StoreStatement implements StorageStatement
     }
 
     @Override
-    public long nextGraphPropertyReference()
+    public long getGraphPropertyReference()
     {
         return neoStores.getMetaDataStore().getGraphNextProp();
     }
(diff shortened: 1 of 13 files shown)


community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/StoreStatement.java
```diff
@@ -308,4 +308,10 @@ public Properties properties()
     {
         return propertyStore;
     }
+
+    @Override
+    public long nextGraphPropertyReference()
```
fickludd: This method is confusingly named imho, which I guess is because of `MetaDataStore.getGraphNextProp`. I think it always returns the reference to the first graph property, so maybe just `getGraphPropertyRefernce` would be better?

fickludd (approved):
Seems legit to me. Small naming suggestion, otherwise clean.

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/StoreStatement.java
pontusmelke: or maybe even `getGraphPropertyReference` :)

PR #11121, opened 2018-02-27
 Two layered state experiment

MishaDemianenko:


MishaDemianenko:
As we find out cypher can't really use dual state tx state, closing this experiment.

PR #11126, opened 2018-02-28
Indexing for temporal values 1

OliviaYtterbrink:
Adds partial indexing capability for temporal values. 

The temporal index will eventually handle the 6 temporal types Date, DateTime, LocalDateTime, Time, LocalTime and Duration by storing them in separate parts, backed by GBPTrees. In this PR we create the base of the pattern around how to handle the lazy creation of the different parts, and implement most of the Date part.

Still to do in another PR:
* add support for the other types of temporal values
* support updates for online index
* support range queries

First commit is `Minor clean-ups and fixes`

This PR builds on https://github.com/neo4j/neo4j/pull/11098


community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/TemporalIndexCache.java
```diff
@@ -0,0 +1,248 @@
...
+            duration = factory.newDuration();
+        }
+        return duration;
+    }
+
+    @Override
+    public Iterator<T> iterator()
+    {
+        T[] ts = Iterators.array( date, dateTime, dateTimeZoned, time, timeZoned, duration );
+        return new Iterator<T>()
```
tinwelint: `return Iterators.filter( item -> item != null, iterator( date, dateTime, dateTimeZoned, time, timeZoned, duration ) );`?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/TemporalSchemaIndexProvider.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public IndexPopulator getPopulator( long indexId, IndexDescriptor descriptor, IndexSamplingConfig samplingConfig )
+    {
+        if ( readOnly )
+        {
+            throw new UnsupportedOperationException( "Can't create populator for read only index" );
+        }
+        TemporalIndexFiles files = new TemporalIndexFiles( directoryStructure(), indexId, descriptor, fs );
```
tinwelint, line 67: Seems like there could be a single instance of `TemporalIndexFiles` in this provider, doesn't it? Applicable to all methods that instantiate these

community/kernel/src/test/java/org/neo4j/kernel/api/index/IndexProviderCompatibilityTestSuite.java
```diff
@@ -50,6 +56,14 @@
 {
     protected abstract SchemaIndexProvider createIndexProvider( PageCache pageCache, FileSystemAbstraction fs, File graphDbDir );
 
+    public abstract Iterable<Value> getSupportedValues();
+
+    protected List<Value> commonValues = Arrays.asList( Values.of( "string1" ), Values.of( 42 ) );
```
tinwelint, line 61: WIth these values and what we discussed @fickludd , I was refering to [FusionIndexTestHelp](https://github.com/neo4j/neo4j/blob/3.4/community/kernel/src/test/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexTestHelp.java) and a potential co-usage of those lists somehow so that we have one place with utilities for all supported value types.

More of a nice-to-have really.

community/kernel/src/test/java/org/neo4j/kernel/api/index/SimpleIndexPopulatorCompatibility.java
```diff
@@ -149,6 +159,112 @@ public void shouldApplyUpdatesIdempotently() throws Exception
...
+        for ( Value value : supportedValues )
+        {
+            result.add( new NodeAndValue( nodeIds++, value ) );
+        }
+        return result;
+    }
+
+    private List<NodeAndValue> allValues = allValues( testSuite.getSupportedValues() );
+
+    static class NodeAndValue
```
tinwelint, line 175: Smells a bit like `IndexEntryUpdate`

community/kernel/src/test/java/org/neo4j/kernel/api/index/SimpleIndexPopulatorCompatibility.java
```diff
@@ -149,6 +159,112 @@ public void shouldApplyUpdatesIdempotently() throws Exception
...
+                    NodeValueIterator nodes = new NodeValueIterator();
+                    reader.query( nodes, IndexOrder.NONE, IndexQuery.exact( propertyKeyId, entry.value ) );
+                    assertEquals( entry.nodeId, getOnly( nodes ) );
+                }
+            }
+            accessor.close();
+        }
+    }
+
+    private long getOnly( PrimitiveLongIterator nodes )
```
tinwelint: Use `PrimitiveLongCollections#single` instead perhaps?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/TemporalIndexCache.java
```diff
@@ -0,0 +1,248 @@
...
+ * Iterating over the cache will return all currently created parts.
+ *
+ * @param <T> Type of parts
+ * @param <E> Type of exception potentially thrown during creation
+ */
+class TemporalIndexCache<T, E extends Exception> implements Iterable<T>
+{
+    private final Factory<T, E> factory;
+
+    private volatile T date;
```
tinwelint, line 43: Implementation may look cleaner and would allow for less duplication with an internal T[] instead?

tinwelint (approved):
The changes seems to generally follow established conventions in the fusion index provider and what not. A nice solution with the `TemporalIndexCache` thing in between for the multiplexing.

I'm going to approve since my comments aren't about all that important stuff.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/TemporalSchemaIndexProvider.java
fickludd: That's not really possible as we need the `indexId` and `descriptor`.

community/kernel/src/test/java/org/neo4j/kernel/api/index/IndexProviderCompatibilityTestSuite.java
fickludd: Yupp. We made a card for cleaning around there.

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/TemporalIndexCache.java
fickludd: We tried this but it felt like roughly equally long, but slightly more complex. Decided to keep as is until we find some other reason to change it.

community/kernel/src/test/java/org/neo4j/kernel/api/index/SimpleIndexPopulatorCompatibility.java
fickludd: It does! We are amending this code quite a bit in the next PR, but maybe we change to only work with `IndexEntryUpdate` when we start homogenizing the test value generation as mentioned above.

fickludd:
@Lojjs I merged this because we want to avoid having too much in-process work. If you want to or think it's necessary, please review it anyway. We can create a separate PR with fixes to any issues you find.

PR #11127, opened 2018-02-28
Make (almost all) TCKTests and AcceptanceTests executable with maven

SaschaPeukert:
Compromise edition of the reverted PR


enterprise/cypher/acceptance-spec-suite/pom.xml
```diff
@@ -89,6 +89,27 @@
       <artifactId>scalacheck_2.11</artifactId>
     </dependency>
 
+    <dependency>
```
MishaDemianenko, line 92: when we run our tests we run them in parallel fashion and so far latest junit and latest surefire can't do it (forked VMs will die in any moment junit will corrupt their sdtout). How this is achieved here?
also current version of surefire that we use is not supported at all btw

MishaDemianenko:
maybe instead contribute to https://github.com/neo4j/neo4j/pull/10998 ?

enterprise/cypher/pom.xml
```diff
@@ -21,6 +21,8 @@
     <licensing.prepend.text>notice-agpl-prefix.txt</licensing.prepend.text>
     <license-text.header>AGPL-3-header.txt</license-text.header>
     <opencypher.version>1.0.0-M09</opencypher.version>
+    <junit-plattform-runner.version>1.1.0</junit-plattform-runner.version>
+    <junit-jupiter.version>5.1.0</junit-jupiter.version>
```
MishaDemianenko, line 25: also i would really like to avoid situation when different submodules that use different versions of libraries that should be common for whole project, just because they can

sherfert:
@MishaDemianenko I disagree for contributing to the other PR. That one seems not like its gonna be merged soon. This one re-enables running all our acceptance tests, which is something quite time constrained to have.

MishaDemianenko:
@sherfert our current surefire will not gonna be able to run your tests i would really like to know how you executing those?
That PR will be merged as soon as surefire will be able to run junit 5 tests, which as i already mentioned current is not able to do.
Will come to talk f2f

fickludd (approved):
I agree that we should use the same junit across the codebase, but it is also very important for our confidence in our changesets that all cypher acceptance and compatibility tests are executed by team city. In the current 3.4 state, *they are not executed*.

With this PR, we allow the JUnit5 tests of the acceptance and compatibility spec suites to run with the JUnit4 surefire plugin, by specifying `@RunWith(classOf[JUnitPlatform])`. This has been verified. As soon as #10998 is merged, we will adapt these spec suites to run regular JUnit5 style, but in the mean time this change reduces risk during the transition.

PR #11128, opened 2018-02-28
 Implement arithmetics for temporal values

sherfert:


community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/Subtract.scala
```diff
@@ -19,9 +19,27 @@
...
+
+  override def apply(ctx: ExecutionContext, state: QueryState): AnyValue = {
+    val aVal = a(ctx, state)
+    val bVal = b(ctx, state)
+
+    (aVal, bVal) match {
+      case (x: TemporalValue[_,_], y: DurationValue) => x.minus(y)
+      case (x: DurationValue, y: DurationValue) => x.sub(y)
+      case (x: TemporalValue[_,_], y: TemporalValue[_,_]) => throw new CypherTypeException("You cannot subtract a temporal instant from another. " +
+        "To obtain the duration, use 'duration.between(temporal1, temporal2)' instead.")
```
fickludd, line 38: Nice error message, very helpful.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/Add.scala
```diff
@@ -45,6 +45,9 @@ case class Add(a: Expression, b: Expression) extends Expression {
       case (x: IntegralValue, y: TextValue) => Values.stringValue(x.longValue() + y.stringValue())
       case (x: TextValue, y: FloatValue) => Values.stringValue(x.stringValue() + y.doubleValue())
       case (x: FloatValue, y: TextValue) => Values.stringValue(x.doubleValue() + y.stringValue())
+      case (x: TemporalValue[_,_], y: DurationValue) => x.plus(y)
+      case (x: DurationValue, y: TemporalValue[_,_]) => y.plus(x)
+      case (x: DurationValue, y: DurationValue) => x.add(y)
```
fickludd, line 50: I suggest to change `DurationValue.add` to `DurationValue.plus`, to adhere to the same pattern and the other values. Same with the other arithmetics on `DurationValue`.

fickludd (approved):
This looks good to me, some naming suggestion which can be ignore or done.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/commands/expressions/Add.scala
sherfert: Since the signatures are slightly different `Temporal.plus(Duration) : Temporal` while `Duration.add(Duration) : Duration`, i think it does not hurt having them named differently.

PR #11129, opened 2018-02-28
Don't use MethodHandle#invokeWithArguments for procedures et al

pontusmelke:
Turns out that `MethodHandle#invokeWithArguments` is a lot slower than normal reflection. 

These are numbers from calling a simple UDF directly, first case is with this change and the latter one is after applying change.

```
MethodHandle#invokeWithArguments 4739.831 ± 212.183  ops/ms
Method#invoke 24860.247 ± 847.512  ops/ms
```

I think there are more improvements to be done here but after applying this fix it looks like the usage of a HashMap in  in the `ProcedureRegistry` is a bigger bottleneck than the reflection.


fickludd (approved):
LGTM. Nice and easy improvement.

PR #11137, opened 2018-03-01
Fix regressions for nodeGetDegree

pontusmelke:
This includes four separate fixes:
- Fix bug in `nodeIsDense` which caused it to always return `false`
- abort early when computing degree for a given type, a type can belong to one-and-only-one group
- Use a traversal cursor directly for non-dense nodes instead of relying on the buffering done by the group cursor
- For `Expand(Into)`: Be smarter when to compute degrees for nodes in the compiled runtime, we should only compute the degree if both nodes are dense like we do in interpreted runtime


fickludd (approved):
I can't find anything to suggest really. Looks good. I assume that you verified that it solves the regressions using benchmarks like always.

PR #11138, opened 2018-03-01
Support for temporal types in Bolt

lutovich:
This PR makes Bolt version 2 aware of six new temporal types: DateTime, LocalDateTime, Date, Time, LocalTime and Duration. Bolt server will now be able to serialize and deserialize these types.

DateTime type contains time zone information either as an offset or as a zone name. That is why it is presented as two distinct types in Bolt. First contains offset as integer and second contains zone name as string.


ali-ince (approved):
Changes look good to me!

PR #11139, opened 2018-03-01
Explicitly reserve key bit for future use

tinwelint:
One of the bits in the key/value header is now explicitly reserved
for future use. This reduces max key size to 4905, but in reality
this limit is even lower anyway.

### DIFF
--- a/community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
+++ b/community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
@@ -21,6 +21,7 @@ package org.neo4j.index.internal.gbptree;
 
 import org.neo4j.io.pagecache.PageCursor;
 
+import static java.lang.String.format;
 import static org.neo4j.index.internal.gbptree.PageCursorUtil.getUnsignedShort;
 import static org.neo4j.index.internal.gbptree.PageCursorUtil.putUnsignedShort;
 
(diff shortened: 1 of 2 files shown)


community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
```diff
@@ -89,7 +93,9 @@
     private static final int FLAG_FIRST_BYTE_TOMBSTONE = 0x80;
     private static final long FLAG_READ_TOMBSTONE = 0x80000000_00000000L;
     static final int MASK_ONE_BYTE_KEY_SIZE = 0x1F;
+    static final int MASK_TWO_BYTE_KEY_SIZE = 0xFFF;
```
MishaDemianenko: according to pictures of two byte key this mask should be 1FFF, isn't it?

community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
```diff
@@ -144,6 +155,11 @@ static void putKeyValueSize( PageCursor cursor, int keySize, int valueSize )
                 byte firstByte = (byte) (valueSize & MASK_ONE_BYTE_VALUE_SIZE); // Least significant 7 bits
                 if ( needsAdditionalValueSize )
                 {
+                    if ( valueSize > MASK_TWO_BYTE_VALUE_SIZE )
+                    {
+                        throw new IllegalArgumentException(
+                                format( "Max supported key size is %d, but tried to store key of size %d", MASK_TWO_BYTE_VALUE_SIZE, valueSize ) );
```
MishaDemianenko: value size, instead of key size in a message

MishaDemianenko (changes requested):

community/index/src/main/java/org/neo4j/index/internal/gbptree/DynamicSizeUtil.java
tinwelint: The picture says e.g. `[0,1,1,k,k,k,k,k][0,k,k,k,k,k,k,k]...` where I count 12 `k`, right?

MishaDemianenko (approved):

PR #11142, opened 2018-03-01
Add support for spatial point values to the batch importer

henriknyman:
- Add simple map-parser for PointValue.
- Move the logic for creating a PointValue with different default
coordinate reference systems depending on the given fields, so that
it can be shared with the batch importer without having to instantiate
an actual Map object. Instead it can build an array with the given
fields directly from the parser.
- Let the csv module depend on the values module so it can produce
values directly.

### DIFF
--- a/community/csv/src/main/java/org/neo4j/csv/reader/Extractors.java
+++ b/community/csv/src/main/java/org/neo4j/csv/reader/Extractors.java
@@ -90,7 +90,7 @@ public class Extractors
     private final Extractor<long[]> longArray;
     private final Extractor<float[]> floatArray;
     private final Extractor<double[]> doubleArray;
-    private final PointExtractor point_;
+    private final PointExtractor point;
 
     public Extractors( char arrayDelimiter )
     {
@@ -141,7 +141,7 @@ public class Extractors
             add( longArray = new LongArrayExtractor( arrayDelimiter ) );
             add( floatArray = new FloatArrayExtractor( arrayDelimiter ) );
             add( doubleArray = new DoubleArrayExtractor( arrayDelimiter ) );
-            add( point_ = new PointExtractor() );
+            add( point = new PointExtractor() );
         }
         catch ( IllegalAccessException e )
         {
@@ -249,9 +249,9 @@ public class Extractors
         return doubleArray;
     }
 
-    public PointExtractor point_()
+    public PointExtractor point()
     {
-        return point_;
+        return point;
     }
 
     private abstract static class AbstractExtractor<T> implements Extractor<T>
--- a/community/csv/src/main/java/org/neo4j/csv/reader/Extractors.java
+++ b/community/csv/src/main/java/org/neo4j/csv/reader/Extractors.java
@@ -20,8 +20,13 @@
 package org.neo4j.csv.reader;
 
 import java.lang.reflect.Field;
+import java.nio.CharBuffer;
 import java.util.HashMap;
 import java.util.Map;
+
+import org.neo4j.values.AnyValue;
+import org.neo4j.values.storable.PointValue;
+import org.neo4j.values.storable.Values;
 
 import static java.lang.Character.isWhitespace;
 import static java.lang.reflect.Modifier.isStatic;
@@ -85,6 +90,7 @@ public class Extractors
     private final Extractor<long[]> longArray;
     private final Extractor<float[]> floatArray;
     private final Extractor<double[]> doubleArray;
+    private final PointExtractor point_;
 
     public Extractors( char arrayDelimiter )
     {
@@ -135,6 +141,7 @@ public class Extractors
             add( longArray = new LongArrayExtractor( arrayDelimiter ) );
             add( floatArray = new FloatArrayExtractor( arrayDelimiter ) );
             add( doubleArray = new DoubleArrayExtractor( arrayDelimiter ) );
+            add( point_ = new PointExtractor() );
         }
         catch ( IllegalAccessException e )
         {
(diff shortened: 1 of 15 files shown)


community/csv/src/main/java/org/neo4j/csv/reader/Extractors.java
```diff
@@ -85,6 +90,7 @@
     private final Extractor<long[]> longArray;
     private final Extractor<float[]> floatArray;
     private final Extractor<double[]> doubleArray;
+    private final PointExtractor point_;
```
sherfert: Is there a particular reason this field ends with an underscore?

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
```diff
@@ -338,4 +274,18 @@ public static RelationshipValue fromRelationshipProxy( Relationship relationship
...
     }
+
+    /**
+     * Creates a {@link Value} from the given object, or if it is already a Value it is returned as it is.
+     * <p>
+     * This is different from {@link Values#of} which explicitly fails if given a Value.
+     */
+    public static Value asValue( Object value )
+    {
+        if ( value instanceof Value )
```
sherfert: `Values.of` checks if the argument is a Point, before checking if the argument is a Value. Hence, passing `PointValue`s works. Is this method really needed?

One reason I can imagine is performace, since `Values.of` will create copies of passed `PointValue`s. If that is the reason for this method, it should be documented as such.

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInputBatchImportIT.java
```diff
@@ -452,6 +475,34 @@ private void buildUpExpectedData(
...
+                {
+                    assertEquals( expectedValue, actualValue );
+                };
+                propertyVerifiers.put( (String) node.propertyKey( i ), verify  );
+            }
+
+            // Special verifier for point property
+            Consumer verifyPoint = actualValue ->
+            {
+                // The y-coordinate should match the node number
```
sherfert, line 495: Is the test simply set up like this or is there another reason why Y is equal to the node number?

community/values/src/main/java/org/neo4j/values/storable/PointValue.java
```diff
@@ -255,4 +261,268 @@ public boolean withinRange( PointValue lower, boolean includeLower, PointValue u
         }
         return true;
     }
+
+    public static PointValue fromMap( MapValue map )
+    {
+        AnyValue[] fields = new Value[PointValueField.__MAX_VALUE__.ordinal()];
```
sherfert: It feels like you could as well use `PointValueField.values().length` and get rid of this `__MAX_VALUE__` altogether.

community/values/src/main/java/org/neo4j/values/storable/PointValue.java
```diff
@@ -255,4 +261,268 @@ public boolean withinRange( PointValue lower, boolean includeLower, PointValue u
...
+        }
+        return Values.pointValue( crs, coordinates );
+    }
+
+    private enum PointValueField
+    {
+        X
+                {
+                    @Override
+                    ValueGroup valueType()
```
sherfert: This becomes way shorter and more readable when you create a field in this enum and a Constructor, instead of overriding the same method everywhere.

E.g.:

```
 private enum PointValueField
    {
        X( NUMBER );

        private ValueGroup valType;

        PointValueField( ValueGroup valType )
        {
            this.valType = valType;
        }

        ValueGroup valueType()
        {
            return ValueGroup.NO_VALUE;
        }
    }
```

sherfert (changes requested):
I have some minor questions and comments. All in all it looks good to me!

Please have a look at the test failures in `SpatialFunctionsAcceptanceTest`

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
henriknyman: Oh, you are right that `Values.of` happens to work for PointValue (and temporal values). But it seems to go against the original purpose of the method according to its documentation. Also we don't want to create new objects unnecessarily during batch import. 

We may probably want to move all extractors over to producing `Value`s directly (now that we have the module dependency), in which case the conversion through this method will not be needed anymore.

community/csv/src/main/java/org/neo4j/csv/reader/Extractors.java
henriknyman: No, I guess I based it on a copy of one that had an underscore :)

community/kernel/src/test/java/org/neo4j/unsafe/impl/batchimport/input/csv/CsvInputBatchImportIT.java
henriknyman: It was just set up like that to get some controlled variation of the lines

community/csv/src/main/java/org/neo4j/csv/reader/Extractors.java
MishaDemianenko: oh :no_good: looks like we missing some checkstyle rule :nut_and_bolt:

sherfert (approved):

sherfert:
Your latest changes seem to have introduced other test failures. Apart from that it looks good to me now!

community/kernel/src/main/java/org/neo4j/kernel/impl/util/ValueUtils.java
fickludd: +1 on creating `Value`s directly instead of going via java Boxed objects.

PR #11144, opened 2018-03-02
Fix remaining JS engine initialisation test failures

MishaDemianenko:


chrisvest:
@MishaDemianenko This should actually be fixed in 3.2, right?

MishaDemianenko:
replaced by https://github.com/neo4j/neo4j/pull/11147

PR #11145, opened 2018-03-02
Backup command should print to stdout

klaren:


phughk (approved):
Well spotted, thanks!

PR #11146, opened 2018-03-02
Index provider for Fulltext Indexes

ragadeeshu:
Not part of this pr:
- Consistency check support
- New kernel api support
- Core api support (await indexes)
- Randomized integration tests
- Recovery tests
- Deferred updates
- DbStructureCollector support
- TransactionLogAnalyzer support
- DumpStore support
- DumpCountsStore support
- Batchinserter tests

- [x] Make old `CommandReader` implementations not need the `IndexProviderMap` dependency, and instead hardcode them to whatever descriptor implementation was there before. See `RecordStorageCommandReaderFactory`
- [ ] Feature toggle to disable as much of the new code as possible.
- [ ] Schema store format migration.


enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/catchup/tx/TxPullResponseDecoder.java
```diff
@@ -42,8 +43,9 @@ protected void decode( ChannelHandlerContext ctx, ByteBuf msg, List<Object> out
     {
         NetworkReadableClosableChannelNetty4 logChannel = new NetworkReadableClosableChannelNetty4( msg );
         StoreId storeId = StoreIdMarshal.INSTANCE.unmarshal( logChannel );
+        //TODO empty index provider map, investigate if we need a rel one.
```
MishaDemianenko: TODO

enterprise/causal-clustering/src/main/java/org/neo4j/causalclustering/core/state/machines/token/ReplicatedTokenRequestSerializer.java
```diff
@@ -120,8 +121,9 @@ public static ReplicatedTokenRequest unmarshal( ByteBuf buffer )
         ByteBuf txBuffer = Unpooled.wrappedBuffer( commandBytes );
         NetworkReadableClosableChannelNetty4 channel = new NetworkReadableClosableChannelNetty4( txBuffer );
 
+        //TODO empty index provider map, investigate if we need a rel one.
```
MishaDemianenko: todo

community/kernel/src/main/java/org/neo4j/kernel/impl/storageengine/impl/recordstorage/RecordStorageCommandReaderFactory.java
```diff
@@ -34,15 +35,15 @@
     // All supported readers. Key/index is LogEntryVersion byte code.
     private final CommandReader[] readers;
 
-    public RecordStorageCommandReaderFactory()
+    public RecordStorageCommandReaderFactory( IndexProviderMap indexProviderMap )
```
MishaDemianenko: we should not update old command readers, also our command readers should not depend on whatever providers or indexes are available in a classpath since those are dynamically registered. We should be able to read our commands always by kernel only

packaging/standalone/standalone-community/src/main/distribution/text/community/conf/neo4j.conf
```diff
@@ -245,7 +245,7 @@ dbms.tx_log.rotation.retention_policy=1 days
 # under the mountpoints specified. Uncomment this line to mount the
 # org.neo4j.examples.server.unmanaged.HelloWorldResource.java from
 # neo4j-server-examples under /examples/unmanaged, resulting in a final URL of
-# http://localhost:#{default.http.port}/examples/unmanaged/helloworld/{nodeId}
+# http://localhost:#{default.http.port}/examples/unmanaged/helloworld/{entityId}
```
MishaDemianenko: do not think this should be updated

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/ConsistencyCheckTasks.java
```diff
@@ -90,13 +91,13 @@
         this.numberOfThreads = numberOfThreads;
     }
 
-    public List<ConsistencyCheckerTask> createTasksForFullCheck( boolean checkLabelScanStore, boolean checkIndexes,
-            boolean checkGraph )
+    public List<ConsistencyCheckerTask> createTasksForFullCheck( boolean checkLabelScanStore, boolean checkIndexes, boolean checkGraph,
+            IndexProviderMap indexes )
```
MishaDemianenko, line 95: we should not pass IndexProviderMap here, we already have access to IndexAccessors here and should use those instead

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/IndexCheck.java
```diff
@@ -38,7 +38,8 @@ public IndexCheck( IndexRule indexRule )
     @Override
     public void check( IndexEntry record, CheckerEngine<IndexEntry, ConsistencyReport.IndexConsistencyReport> engine, RecordAccess records )
     {
-        int labelId = indexRule.schema().getLabelId();
+        //TODO support fulltext indexes in index consistency check
```
MishaDemianenko: TODO and access by [0]

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
```diff
@@ -95,8 +96,8 @@ private void matchIndexesToNode(
         PrimitiveIntObjectMap<PropertyBlock> nodePropertyMap = null;
         for ( IndexRule indexRule : indexes.onlineRules() )
         {
-            long labelId = indexRule.schema().getLabelId();
-            if ( labels.contains( labelId ) )
+            int[] labelIds = indexRule.schema().getEntityTokenIds();
+            if ( Arrays.stream( labelIds ).asLongStream().anyMatch( labels::contains ) )
```
MishaDemianenko: Instead of converting primitive int array to long objects to find matches please use primitive arrays instead (update NodeLabelReader to return primitives always as well)

community/consistency-check/src/main/java/org/neo4j/consistency/internal/SchemaIndexExtensionLoader.java
```diff
@@ -25,42 +25,48 @@
...
-    public static KernelExtensions instantiateKernelExtensions( File storeDir, FileSystemAbstraction fileSystem,
-            Config config, LogService logService, PageCache pageCache,
-            RecoveryCleanupWorkCollector recoveryCollector, DatabaseInfo databaseInfo, Monitors monitors )
+    public static KernelExtensions instantiateKernelExtensions( File storeDir, FileSystemAbstraction fileSystem, Config config, LogService logService,
+            PageCache pageCache, RecoveryCleanupWorkCollector recoveryCollector, DatabaseInfo databaseInfo, Monitors monitors,
+            PropertyKeyTokenHolder propkeyTokenHolder, LabelTokenHolder labelTokenHolder, RelationshipTypeTokenHolder relationshipTypeTokenHolder )
     {
         Dependencies deps = new Dependencies();
-        deps.satisfyDependencies( fileSystem, config, logService, pageCache, recoveryCollector, monitors );
+        //TODO figure out a way to not depend on procedures here, it is caused by the fulltext index.
```
MishaDemianenko: TODO

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v2_3/SchemaDescriptorTranslation.scala
```diff
@@ -21,15 +21,16 @@ package org.neo4j.cypher.internal.spi.v2_3
...
 trait SchemaDescriptorTranslation {
-  implicit def cypherToKernel(index: SchemaTypes.IndexDescriptor): KernelIndexDescriptor =
-    IndexDescriptorFactory.forLabel(index.labelId, index.propertyId)
+  implicit def cypherToKernel(index: SchemaTypes.IndexDescriptor): SchemaIndexDescriptor =
+    SchemaIndexDescriptorFactory.forLabel(index.labelId, index.propertyId)
 
   implicit def kernelToCypher(index: KernelIndexDescriptor): SchemaTypes.IndexDescriptor =
     if (index.schema().getPropertyIds.length == 1)
-      SchemaTypes.IndexDescriptor(index.schema().getLabelId, index.schema().getPropertyId)
+    //TODO we use a zero index here as to not bleed multi-token descriptors into cypher. At least for now.
```
MishaDemianenko: TODO

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/SchemaDescriptorTranslation.scala
```diff
@@ -22,15 +22,16 @@ package org.neo4j.cypher.internal.spi.v3_1
...
 trait SchemaDescriptorTranslation {
-  implicit def toKernel(index: SchemaTypes.IndexDescriptor): KernelIndexDescriptor =
-    IndexDescriptorFactory.forLabel(index.labelId, index.propertyId)
+  implicit def toKernel(index: SchemaTypes.IndexDescriptor): SchemaIndexDescriptor =
+    SchemaIndexDescriptorFactory.forLabel(index.labelId, index.propertyId)
 
   implicit def toCypher(index: KernelIndexDescriptor): SchemaTypes.IndexDescriptor = {
     assertSingleProperty(index.schema())
-    SchemaTypes.IndexDescriptor(index.schema().getLabelId, index.schema().getPropertyId())
+    //TODO we use a zero index here as to not bleed multi-token descriptors into cypher. At least for now.
```
MishaDemianenko: TODO

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_3/TransactionBoundPlanContext.scala
```diff
@@ -87,7 +88,8 @@ class TransactionBoundPlanContext(readOperationsSupplier: () => ReadOperations,
 
   private def getOnlineIndex(descriptor: KernelIndexDescriptor): Option[IndexDescriptor] =
     readOperationsSupplier().indexGetState(descriptor) match {
-      case InternalIndexState.ONLINE => Some(IndexDescriptor(descriptor.schema().getLabelId, descriptor.schema().getPropertyIds))
+      //TODO we use a zero index here as to not bleed multi-token descriptors into cypher. At least for now.
```
MishaDemianenko: TODO

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundPlanContext.scala
```diff
@@ -92,7 +93,8 @@ class TransactionBoundPlanContext(tc: TransactionalContextWrapper, logger: Inter
 
   private def getOnlineIndex(descriptor: KernelIndexDescriptor): Option[IndexDescriptor] =
     tc.statement.readOperations().indexGetState(descriptor) match {
-      case InternalIndexState.ONLINE => Some(IndexDescriptor(descriptor.schema().getLabelId, descriptor.schema().getPropertyIds))
+      //TODO we use a zero index here as to not bleed multi-token descriptors into cypher. At least for now.
```
MishaDemianenko: TODO

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/IndexDescriptorCompatibility.scala
```diff
@@ -20,24 +20,25 @@
...
+import org.neo4j.kernel.api.schema.index.{SchemaIndexDescriptorFactory, IndexDescriptor => KernelIndexDescriptor}
 
 trait IndexDescriptorCompatibility {
   def cypherToKernel(index: CypherIndexDescriptor): KernelIndexDescriptor =
-    IndexDescriptorFactory.forLabel(index.label.id, index.properties.map(_.id):_*)
+    SchemaIndexDescriptorFactory.forLabel(index.label.id, index.properties.map(_.id):_*)
 
   def kernelToCypher(index: KernelIndexDescriptor): CypherIndexDescriptor =
-    CypherIndexDescriptor(index.schema().getLabelId, index.schema().getPropertyIds)
+  //TODO we use a zero index here as to not bleed multi-token descriptors into cypher. At least for now.
```
MishaDemianenko: TODO

community/cypher/cypher-planner-3.4/src/test/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/DbStructureLogicalPlanningConfiguration.scala
```diff
@@ -63,8 +63,9 @@ object DbStructureLogicalPlanningConfiguration {
     }
   }
 
-  private def indexSet(indices: util.Iterator[Pair[String, Array[String]]]): Set[(String, Seq[String])] =
-    indices.asScala.map { pair => pair.first() -> pair.other().to[Seq] }.toSet
+  private def indexSet(indices: util.Iterator[Pair[Array[String], Array[String]]]): Set[(String, Seq[String])] =
+    //TODO we use a zero index here as to not bleed multi-token descriptors into cypher. At least for now.
```
MishaDemianenko: TODO and the same [0] access pattern that should not be used

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomProcedures.java
```diff
@@ -47,93 +53,205 @@
...
+    public void awaitPopulation() throws IndexPopulationFailedKernelException
     {
-        provider.awaitPopulation();
+        await( BLOOM_NODES );
+        await( BLOOM_RELATIONSHIPS );
+    }
+
+    private void await( String indexName ) throws IndexPopulationFailedKernelException
+    {
+        //TODO This is here because the core api really doesn't play nicely with multi token schema yet.
```
MishaDemianenko: TODO

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomProcedures.java
```diff
@@ -47,93 +53,205 @@
...
     {
-        return provider.getProperties( BLOOM_NODES, NODES ).stream().map( PropertyOutput::new );
+        try ( Statement stmt = tx.acquireStatement() )
+        {
+            ReadOperations readOperations = stmt.readOperations();
+            return accessor.propertyKeyStrings( readOperations.indexGetForName( BLOOM_NODES ) ).map( PropertyOutput::new );
+        }
+        catch ( SchemaRuleNotFoundException e )
+        {
+            return Stream.empty();
```
MishaDemianenko: why error is empty stream? how do we recognize the difference between error and no result? the same question for almost all procedures in this class

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextAccessor.java
```diff
@@ -17,18 +17,23 @@
...
-    /**
-     * Wait for the index operation to complete. Returns immediately if the operation has already completed.
-     * @throws ExecutionException if the asynchronous operation failed with an exception.
-     */
-    void awaitCompletion() throws ExecutionException;
+    String FIELD_ENTITY_ID = "__lucene__fulltext__addon__internal__id__";
+
+    Stream<String> propertyKeyStrings( IndexDescriptor descriptor );
+
+    IndexDescriptor indexDescriptorFor( String name, EntityType type, String[] entityTokens, String... properties );
```
MishaDemianenko: always passes new String[0] as value for entityTokens in production code

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexAccessor.java
```diff
@@ -0,0 +1,226 @@
...
+    @Override
+    public void close() throws IOException
+    {
+        luceneFulltext.close();
+    }
+
+    @Override
+    public IndexReader newReader()
+    {
+        return IndexReader.EMPTY;
```
MishaDemianenko: why do we return empty reader?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexAccessor.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    @Override
+    public BoundedIterable<Long> newAllEntriesReader()
+    {
+        //TODO support consistency check at some point
+        BoundedIterable<Document> empty = new BoundedIterable<Document>()
+        {
+            @Override
+            public long maxCount()
```
MishaDemianenko: todo

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexPopulator.java
```diff
@@ -0,0 +1,151 @@
...
+import org.neo4j.kernel.api.index.PropertyAccessor;
+import org.neo4j.kernel.impl.api.index.sampling.IndexSamplingConfig;
+import org.neo4j.storageengine.api.schema.IndexSample;
+
+public class FulltextIndexPopulator implements IndexPopulator
+{
+    private final FulltextIndexDescriptor descriptor;
+    private final WritableFulltext index;
+
+    FulltextIndexPopulator( long indexId, FulltextIndexDescriptor descriptor, IndexSamplingConfig samplingConfig, WritableFulltext index )
```
MishaDemianenko: half of the params are not used

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexPopulator.java
```diff
@@ -0,0 +1,151 @@
...
+import org.neo4j.kernel.api.impl.fulltext.lucene.LuceneFulltextDocumentStructure;
+import org.neo4j.kernel.api.impl.fulltext.lucene.WritableFulltext;
+import org.neo4j.kernel.api.index.IndexEntryUpdate;
+import org.neo4j.kernel.api.index.IndexPopulator;
+import org.neo4j.kernel.api.index.IndexUpdater;
+import org.neo4j.kernel.api.index.PropertyAccessor;
+import org.neo4j.kernel.impl.api.index.sampling.IndexSamplingConfig;
+import org.neo4j.storageengine.api.schema.IndexSample;
+
+public class FulltextIndexPopulator implements IndexPopulator
```
MishaDemianenko: this class is almost exact copy of LuceneIndexPopulator, looks like LuceneIndexPopulator should be used as superclass instead

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProvider.java
```diff
@@ -0,0 +1,230 @@
...
+    @Override
+    public FulltextIndexDescriptor indexDescriptorFor( SchemaDescriptor schema, String name, String metadata )
+    {
+        try
+        {
+            return new FulltextIndexDescriptor( schema, name, propertyKeyTokenHolder, metadata );
+        }
+        catch ( TokenNotFoundException e )
+        {
+            throw new RuntimeException( e );
```
MishaDemianenko: hm, runtime exception?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProvider.java
```diff
@@ -0,0 +1,230 @@
...
+    @Override
+    public boolean compatible( IndexDescriptor indexDescriptor )
+    {
+        return indexDescriptor instanceof FulltextIndexDescriptor;
+    }
+
+    @Override
+    public StoreMigrationParticipant storeMigrationParticipant( FileSystemAbstraction fs, PageCache pageCache )
+    {
+        return new SchemaIndexMigrator( fs, this );
```
MishaDemianenko: migrate schema indexes? noooo

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProvider.java
```diff
@@ -0,0 +1,230 @@
...
+        if ( fulltextIndexAccessor == null )
+        {
+            throw new IndexNotFoundKernelException( "The requested fulltext index could not be accessed. Perhaps population has not completed yet?" );
+        }
+        return fulltextIndexAccessor.query( queryString );
+    }
+
+    private class BadSchemaException extends IllegalArgumentException
+    {
+        BadSchemaException( String message )
```
MishaDemianenko: bad schema error? with the only usage with a message that its impossible to index property?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/FulltextFactory.java
```diff
@@ -76,21 +72,14 @@ private Analyzer getAnalyzer( String analyzerClassName )
...
 
-    LuceneFulltext createFulltextIndex( String identifier, FulltextIndexType type, List<String> properties )
+    public LuceneFulltext createFulltextIndex( long indexId, FulltextIndexDescriptor descriptor )
     {
-        File indexRootFolder = new File( indexDir, identifier );
-        LuceneIndexStorageBuilder storageBuilder = LuceneIndexStorageBuilder.create();
-        storageBuilder.withFileSystem( fileSystem ).withIndexFolder( indexRootFolder );
-        PartitionedIndexStorage storage = storageBuilder.build();
-        return new LuceneFulltext( storage, partitionFactory, properties, analyzer, identifier, type );
+        PartitionedIndexStorage storage = indexStorageFactory.indexStorageOf( indexId, config.get( GraphDatabaseSettings.archive_failed_index ) );
```
MishaDemianenko: `indexStorageFactory.indexStorageOf( indexId, config.get( GraphDatabaseSettings.archive_failed_index ) )` should be extracted into separte method

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/FulltextQueryHelper.java
```diff
@@ -17,18 +17,27 @@
...
-import static org.neo4j.kernel.api.impl.fulltext.integrations.bloom.BloomKernelExtensionFactory.BLOOM_NODES;
+import java.util.Collection;
 
-public class StubLuceneFulltext extends LuceneFulltext
+import static java.util.stream.Collectors.joining;
+
+public class FulltextQueryHelper
 {
-    StubLuceneFulltext()
+    public static String createQuery( Collection<String> query, boolean fuzzy, boolean matchAll )
```
MishaDemianenko: this whole method should be just rewritten and tests should be provided

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/LuceneFulltext.java
```diff
@@ -17,73 +17,51 @@
...
 import org.neo4j.kernel.api.impl.index.partition.PartitionSearcher;
-import org.neo4j.kernel.api.impl.index.partition.WritableIndexPartitionFactory;
 import org.neo4j.kernel.api.impl.index.storage.PartitionedIndexStorage;
 import org.neo4j.kernel.api.impl.schema.writer.PartitionedIndexWriter;
+import org.neo4j.storageengine.api.EntityType;
 
-class LuceneFulltext extends AbstractLuceneIndex
+import static java.util.Collections.singletonMap;
+
+public class LuceneFulltext extends AbstractLuceneIndex implements Closeable
```
MishaDemianenko: LuceneFulltext copies LuceneSchemaIndex behavior by copy-pasting, it should extend it instead.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/LuceneFulltextDocumentStructure.java
```diff
@@ -49,18 +50,30 @@ private static DocWithId reuseDocument( long id )
         return doc;
     }
 
-    static Document documentRepresentingProperties( long id, Map<String,Object> values )
+    public static Document documentRepresentingProperties( long id, Map<String,Value> values )
```
MishaDemianenko: this method is not used

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/ReadOnlyFulltext.java
```diff
@@ -17,27 +17,17 @@
...
-            }
+    /**
+     * Queires the fulltext index with the given lucene-syntax query
+     * @param query the lucene query
+     * @return A {@link ScoreEntityIterator} over the results
+     */
+    ScoreEntityIterator query( String query );
+
+    @Override
+    void close();
```
MishaDemianenko: this interface overrides AutoCloseable close to pretend it does not throw anything while two implementations do that in reality and try to catch and rethrow exceptions while they do not need to do that since in a places where its used already we are free to throw IOExceptions since lucene do that anyway

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/SimpleFulltextReader.java
```diff
@@ -85,28 +66,9 @@ public void close()
...
-        {
-            multiFieldQueryParser.setDefaultOperator( QueryParser.Operator.AND );
-        }
-        else
-        {
-            multiFieldQueryParser.setDefaultOperator( QueryParser.Operator.OR );
-        }
         Query query;
         try
         {
```
MishaDemianenko: why query error is the same as empty result?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/WritableFulltext.java
```diff
@@ -56,43 +54,23 @@ public void drop() throws IOException
...
-    PartitionedIndexWriter getIndexWriter()
+    public void markAsOnline() throws IOException
     {
-        return indexWriter;
-    }
-
-    Set<String> getProperties()
-    {
-        return luceneIndex.getProperties();
+        luceneIndex.markAsOnline();
```
MishaDemianenko: copy paste leak: locking needs to be used here as well

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomFulltextIndexBenchmarks.java
```diff
@@ -35,17 +35,13 @@
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
 import org.neo4j.test.rule.TestDirectory;
 
-import static org.neo4j.kernel.api.impl.fulltext.integrations.bloom.BloomFulltextConfig.bloom_enabled;
-
-@Ignore( "These are rudimentary benchmarks, but implemented via the jUnit framework to make them easy to run " +
-         "from an IDE." )
+@Ignore( "These are rudimentary benchmarks, but implemented via the jUnit framework to make them easy to run " + "from an IDE." )
```
MishaDemianenko: i'm sure i already mentioned this before: BloomFulltextIndexBenchmarks should not be Ignored and live in the product. It should be moved into benchmarks repo and run there.

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProviderTest.java
```diff
@@ -0,0 +1,390 @@
...
+
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+import static org.neo4j.graphdb.Label.label;
+
+public class FulltextIndexProviderTest
```
MishaDemianenko: this class is a fountain of copy paste!

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/lucene/LuceneFulltextTestSupport.java
```diff
@@ -0,0 +1,226 @@
...
+        {
+            Node node = db.getNodeById( nodeId );
+            node.setProperty( propertyKey, value );
+            tx.success();
+        }
+    }
+
+    protected void await( IndexDescriptor fulltextIndexDescriptor ) throws IndexNotFoundKernelException
+    {
+        //TODO change this once you can use normal index await with multi-token indexes
```
MishaDemianenko: TODO and continue inside is useless

community/kernel/src/main/java/org/neo4j/kernel/api/exceptions/schema/SchemaRuleException.java
```diff
@@ -33,9 +33,10 @@
  */
 class SchemaRuleException extends SchemaKernelException
 {
-    protected final SchemaDescriptor descriptor;
     protected final String messageTemplate;
     protected final SchemaRule.Kind kind;
+    protected SchemaDescriptor descriptor;
```
MishaDemianenko: two new fields should be final

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexProvider.java
```diff
@@ -90,32 +93,10 @@
  * {@link #getOnlineAccessor(long, IndexDescriptor, IndexSamplingConfig) online accessor} to
  * write to the index.
  */
-public abstract class SchemaIndexProvider extends LifecycleAdapter implements Comparable<SchemaIndexProvider>
+public abstract class IndexProvider<DESCRIPTOR extends IndexDescriptor> extends LifecycleAdapter implements Comparable<IndexProvider<?>>
```
MishaDemianenko: `Comparable<IndexProvider<?>>` ? maybe `Comparable<IndexProvider<DESCRIPTOR>>` instead?

community/kernel/src/main/java/org/neo4j/kernel/api/schema/index/IndexDescriptor.java
```diff
@@ -36,15 +36,20 @@
...
+    {
+        return "";
+    }
+
     public enum Type
     {
         GENERAL,
-        UNIQUE
+        UNIQUE,
+        NON_SCHEMA
```
MishaDemianenko: this should be renamed together with all related things to FULL_TEXT

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/schema/SchemaDescriptor.java
```diff
@@ -34,8 +35,15 @@
...
  */
-public interface SchemaDescriptor
+public interface SchemaDescriptor extends SchemaDescriptorSupplier
 {
+    boolean isAffected( long[] entityIds );
+
+    enum PropertySchemaType
+    {
+        SCHEMA_ALL_TOKENS,
+        NON_SCHEMA_ANY_TOKEN
```
MishaDemianenko: this should be renamed as well

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/schema/SchemaDescriptor.java
```diff
@@ -34,8 +35,15 @@
  * how this is done in eg. LabelSchemaDescriptor, and the SchemaProcessor and SchemaComputer interfaces need to be
  * extended with methods taking the new concrete type as argument.
  */
-public interface SchemaDescriptor
+public interface SchemaDescriptor extends SchemaDescriptorSupplier
 {
+    boolean isAffected( long[] entityIds );
+
+    enum PropertySchemaType
```
MishaDemianenko: would be nice to get javadoc for this class since the purpose of this enum is not really clear

community/kernel/src/main/java/org/neo4j/kernel/api/schema/MultiTokenSchemaDescriptor.java
```diff
@@ -0,0 +1,167 @@
...
+    @Override
+    public int[] getEntityTokenIds()
+    {
+        return entityTokens;
+    }
+
+    @Override
+    public int keyId()
+    {
+        //TODO MultiTokenSchema support for the new kernel api. This method is only used in the AllStoreHolder
```
MishaDemianenko, line 105: TODO

community/kernel/src/main/java/org/neo4j/kernel/api/ReadOperations.java
```diff
@@ -67,6 +67,7 @@
  */
 public interface ReadOperations
 {
+    long UNLABELLED = Integer.MAX_VALUE + 1L;
```
MishaDemianenko: what is it for? there are no tests that check this as well

community/kernel/src/main/java/org/neo4j/kernel/api/SchemaWriteOperations.java
```diff
@@ -65,4 +67,7 @@ RelExistenceConstraintDescriptor relationshipPropertyExistenceConstraintCreate(
             RepeatedPropertyInCompositeSchemaException;
 
     void constraintDrop( ConstraintDescriptor constraint ) throws DropConstraintFailureException;
+
+    IndexDescriptor nonSchemaIndexCreate( IndexDescriptor nonSchemaIndexDescriptor )
```
MishaDemianenko: nice try but no, please rename the whole stack of those methods

community/kernel/src/main/java/org/neo4j/kernel/builtinprocs/BuiltInProcedures.java
```diff
@@ -892,7 +893,8 @@ public RelationshipResult( Relationship relationship )
     private enum IndexType
     {
         NODE_LABEL_PROPERTY( "node_label_property" ),
-        NODE_UNIQUE_PROPERTY( "node_unique_property" );
+        NODE_UNIQUE_PROPERTY( "node_unique_property" ),
+        NON_SCHEMA( "non_schema" );
```
MishaDemianenko, line 897: this should be renamed to full text

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/EntityUpdates.java
```diff
@@ -29,67 +29,69 @@
...
+import static org.neo4j.kernel.impl.api.index.EntityUpdates.PropertyValueType.Changed;
+import static org.neo4j.kernel.impl.api.index.EntityUpdates.PropertyValueType.NoValue;
+import static org.neo4j.kernel.impl.api.index.EntityUpdates.PropertyValueType.UnChanged;
 
 /**
  * Subclasses of this represent events related to property changes due to node addition, deletion or update.
  * This is of use in populating indexes that might be relevant to node label and property combinations.
  */
-public class NodeUpdates implements PropertyLoader.PropertyLoadSink
+public class EntityUpdates implements PropertyLoader.PropertyLoadSink
```
MishaDemianenko, line 52: class was heavily updated but there are no new tests

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
```diff
@@ -272,25 +270,35 @@ public void start() throws Exception
...
+                {
+                    rebuildingNodeDescriptors.put( indexId, descriptor );
+                }
+                else
+                {
+                    rebuildingRelationshipDescriptors.put( indexId, descriptor );
+                }
+            }
+
+            if ( !rebuildingNodeDescriptors.isEmpty() )
```
MishaDemianenko: i'm sure that starting those 2 jobs as they are now will not gonna play well since each of those will gonna try to perform multi-threaded index population with tons of threads fighting with each other for no reason

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexMap.java
```diff
@@ -45,10 +48,14 @@
 public final class IndexMap implements Cloneable
```
MishaDemianenko: class was heavily updated but again no new tests

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexStoreView.java
```diff
@@ -41,26 +42,39 @@
...
+
+    /**
+     * Produces {@link EntityUpdates} objects from reading node {@code nodeId}, its labels and properties
      * and puts those updates into node updates container.
      *
      * @param nodeId id of node to load.
      * @return node updates container
      */
-    NodeUpdates nodeAsUpdates( long nodeId );
+    //TODO This seems to only be used in tests
```
MishaDemianenko: TODO again

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
```diff
@@ -297,6 +312,7 @@ void flipAfterPopulation()
         {
             try
             {
+                Thread.sleep( 1000 );
```
MishaDemianenko: yes right, we already too fast lets give a cpu some time to sleep :)

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
```diff
@@ -317,9 +333,14 @@ private IntStream propertyKeyIds( IndexPopulation population )
         return IntStream.of( population.schema().getPropertyIds() );
     }
 
-    private int[] labelIds()
+    private int[] entityTokenIds()
     {
-        return populations.stream().mapToInt( population -> population.schema().getLabelId() ).toArray();
+        if ( populations.stream().anyMatch( indexPopulation -> indexPopulation.schema().getEntityTokenIds().length == 0 ) )
+        {
+            //No token is any token
```
MishaDemianenko: this should not be here, empty array and [0] element of array is leaking everywhere

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -371,9 +372,11 @@ public long nodeGetFromUniqueIndexSeek(
             ExactPredicate... predicates )
             throws IndexNotFoundKernelException, IndexBrokenKernelException, IndexNotApplicableKernelException
     {
+        assertIndexHasOneEntityToken( index.schema() );
         assertIndexOnline( state, index );
         assertPredicatesMatchSchema( index.schema(), predicates );
-        int labelId = index.schema().getLabelId();
+        // This is fine because of the assertion above
+        int labelId = index.schema().getEntityTokenIds()[0];
```
MishaDemianenko: same leak here

community/kernel/src/main/java/org/neo4j/kernel/impl/api/LockingStatementOperations.java
```diff
@@ -502,6 +529,64 @@ public String indexGetFailure( Statement state, IndexDescriptor descriptor )
         return schemaReadDelegate.indexGetFailure( state, descriptor );
     }
 
+    private void lockSchemaExclusive( KernelStatement state, SchemaDescriptor schema )
```
MishaDemianenko: what this 2 new methods are doing? i can't see any tests for them as well?

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/SchemaImpl.java
```diff
@@ -132,7 +133,8 @@ private IndexDefinition descriptorToDefinition( final ReadOperations statement,
     {
         try
         {
-            Label label = label( statement.labelGetName( index.schema().getLabelId() ) );
+            //TODO we use a zero index here to avoid bleeding into core api. We need to figure out how to deal with the core api in regards to fulltext indexes.
```
MishaDemianenko: TODO again

MishaDemianenko (changes requested):
First bunch of comments:
* also FulltextIndexUpdater is almost a copy of LuceneIndexUpdater;
* GraphDatabaseFacade contains a hack
* package of whole thing is `org.neo4j.kernel.api.impl.fulltext.*`
which its not, have nothing to do with kernel
* lots of places do something like `descriptor.getIndexDescriptor().schema().entityType() == EntityType.NODE` and then operate on top of descriptor itself - so again leak of abstraction there but i'm not done with that part fully so will only mention it as general note
* there are no consistency checker updates for new indexes?
* now changes of analyzer property will gonna drop all of the full text indexes. That behaviour is different compared to other indexes and we need to warn about that in property description and write warning messages into the log file at least.
*we talked about flag to disable the whole thing 2 days ago and i do not see it. Where is it?

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
klaren: Also, the reason they differ is that `getEntityTokenIds` returns `int` and `NodeLabelReader` returns `long`. We should resolve this conflict where two different methods claim we have a different length of label ids.

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexEntryUpdate.java
```diff
@@ -35,7 +36,7 @@
  *
  * @param <INDEX_KEY> {@link LabelSchemaSupplier} specifying the schema
```
klaren: `SchemaDescriptorSupplier`

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexProvider.java
klaren: I think this change introduces too many raw types. Could we look into removing the generic type?

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/schema/MultiTokenSchemaDescriptor.java
```diff
@@ -17,19 +17,8 @@
...
-package org.neo4j.kernel.impl.api.index;
+package org.neo4j.internal.kernel.api.schema;
 
-import java.util.function.Consumer;
-import java.util.function.Function;
-
-import org.neo4j.kernel.api.index.SchemaIndexProvider;
-
-public interface SchemaIndexProviderMap extends Function<SchemaIndexProvider.Descriptor, SchemaIndexProvider>
+public interface MultiTokenSchemaDescriptor extends SchemaDescriptor
```
klaren: A javadoc describing what multi token is would be nice.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextAccessor.java
ragadeeshu: Yes, because bloom is the only production code we have right now, and that is the way they wanted it last time I asked. However this will not be the case for the fulltext surface when that arrives.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProvider.java
ragadeeshu: Nice catch, we need to investigate this more!

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProvider.java
ragadeeshu: At a first glance, it looked like the SchemaIndexMigrator would do the job just fine. Tests for this will be written at some point, but if you are certain that it won't work (why? :) ) we will look into it now.

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomFulltextIndexBenchmarks.java
ragadeeshu: Oh sorry, this is an old class that has just been moved around. These tests were used as an tool when developing the old fulltext index, will clean up.

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/lucene/LuceneFulltextTestSupport.java
ragadeeshu: TODO beacuse of core api support being in the future. The continue is in order to pass check-style.

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProviderTest.java
ragadeeshu: Sorry, this test was a development tool that never got a proper cleanup, I'll see to it.

community/kernel/src/main/java/org/neo4j/kernel/api/ReadOperations.java
```diff
@@ -68,6 +69,7 @@
...
```
klaren: This looks dangerous. The `-1` is "protected" by `org.neo4j.kernel.impl.store.id.IdGeneratorImpl#INTEGER_MINUS_ONE` and `org.neo4j.kernel.impl.store.id.validation.IdValidator#isReservedId` (we already have a card to make this relationship clear). Might be better to have it `Long.MIN_VALUE` or something not in the path of future label ids if we migrate to `long` for that.

community/kernel/src/main/java/org/neo4j/kernel/api/schema/index/IndexDescriptor.java
ragadeeshu: This isn't called full_text because it is not necessarily fulltext specific. I mean, it is right now, but there is nothing stating that is has to be, it could be any index that isn't a schema index. There is also no mention of fulltext anywhere in the kernel code as a whole, all of the fulltext aware code lives in the fulltext module, to me it feels like a nice thing that it doesn't bleed into the kernel code.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/RelationshipStoreScan.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    private static boolean containsAnyRelType( int[] relTypeFilter, int relType )
+    {
+        // No relationship types means we take all relationship types:
+        if ( relTypeFilter.length == 0 )
+        {
+            return true;
+        }
+        for ( long candidate : relTypeFilter )
```
klaren: `int`

community/kernel-api/src/main/java/org/neo4j/internal/kernel/api/schema/SchemaDescriptor.java
ragadeeshu: Not sure I agree, but it definitely should be documented, very good point.

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/storeview/RelationshipStoreScan.java
```diff
@@ -0,0 +1,222 @@
...
+import org.neo4j.kernel.impl.store.record.Record;
+import org.neo4j.kernel.impl.store.record.RelationshipRecord;
+import org.neo4j.kernel.impl.util.Validators;
+import org.neo4j.storageengine.api.schema.PopulationProgress;
+import org.neo4j.values.storable.Value;
+
+import static java.util.Collections.emptyIterator;
+import static org.neo4j.kernel.impl.store.record.RecordLoad.FORCE;
+
+public class RelationshipStoreScan<FAILURE extends Exception> implements StoreScan<FAILURE>
```
klaren: Lots of duplicate code here, e.g `acceptUpdate`, `getProgress`, `valueOf` etc. is dientical to `StoreViewNodeStoreScan`.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexStoreView.java
ragadeeshu: Right, this todo here... It only seems to be used in tests, but as far as I can tell, it has been that way for quite a while, and since we did not cause it, only noticed it, we did not prioritize fixing it. Would you rather see that we took time to remove it, removed the todo, or neither?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
ragadeeshu: Do you have any proposals for better defining any label/reltype than the empty array? I saw it being defined as empty array somewhere in existing code, and decided to follow that pattern since I couldn't figure out anything better.

community/kernel/src/main/java/org/neo4j/kernel/impl/coreapi/schema/SchemaImpl.java
ragadeeshu: This is the bleeding edge towards the not-yet-supported core api.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexProviderMap.java
```diff
@@ -0,0 +1,66 @@
...
+package org.neo4j.kernel.impl.api.index;
+
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import org.neo4j.kernel.api.index.IndexProvider;
+import org.neo4j.kernel.api.schema.index.IndexDescriptor;
+import org.neo4j.kernel.api.schema.index.SchemaIndexDescriptor;
+
+public interface IndexProviderMap extends Function<IndexProvider.Descriptor,IndexProvider>
```
klaren: Really no need to extend `Function` here.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexProviderMap.java
```diff
@@ -0,0 +1,66 @@
...
+
+        @Override
+        public void accept( Consumer<IndexProvider> visitor )
+        {
+            //Sure
+        }
+    };
+
+    @Override
+    IndexProvider apply( IndexProvider.Descriptor descriptor ) throws IndexProviderNotFoundException;
```
klaren: rename

community/kernel/src/main/java/org/neo4j/kernel/api/schema/MultiTokenSchemaDescriptor.java
ragadeeshu: This is the bleeding edge towards the not-yet-supported new kernel api.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
```diff
@@ -93,21 +96,23 @@
...
-    // Populators are added into this list. The same thread adding populators will later call #indexAllNodes.
+    // Populators are added into this list. The same thread adding populators will later call #indexAllEntities.
     // Multiple concurrent threads might fail individual populations.
     // Failed populations are removed from this list while iterating over it.
     final List<IndexPopulation> populations = new CopyOnWriteArrayList<>();
 
     private final IndexStoreView storeView;
     private final LogProvider logProvider;
     protected final Log log;
+    private EntityType type;
```
klaren: `final`

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
```diff
@@ -326,9 +338,14 @@ private IntStream propertyKeyIds( IndexPopulation population )
...
     }
 
-    private int[] labelIds()
+    private int[] entityTokenIds()
     {
-        return populations.stream().mapToInt( population -> population.schema().getLabelId() ).toArray();
+        if ( populations.stream().anyMatch( indexPopulation -> indexPopulation.schema().getEntityTokenIds().length == 0 ) )
+        {
+            //No token is any token
+            return new int[0];
```
klaren: At least make this a constant with a proper name and remove the comment.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexAccessor.java
ragadeeshu: Yeah, consistency checker isn't supported yet, and this is to keep track of where the rough edges are.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomProcedures.java
ragadeeshu: This is the bleeding edge towards the not-yet-supported core api.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/IndexDescriptorCompatibility.scala
ragadeeshu: This is the bleeding edge towards the not-yet-supported cypher. I'm not sure that cypher will ever need this, but if that is the case we should probably figure out a nicer transition.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v2_3/SchemaDescriptorTranslation.scala
ragadeeshu: See the other cypher comment.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_1/SchemaDescriptorTranslation.scala
ragadeeshu: See the other cypher comment.

community/cypher/cypher/src/main/scala/org/neo4j/cypher/internal/spi/v3_3/TransactionBoundPlanContext.scala
ragadeeshu: See the other cypher comment.

community/cypher/interpreted-runtime/src/main/scala/org/neo4j/cypher/internal/runtime/interpreted/TransactionBoundPlanContext.scala
ragadeeshu: See the other cypher comment.

community/cypher/cypher-planner-3.4/src/test/scala/org/neo4j/cypher/internal/compiler/v3_4/planner/DbStructureLogicalPlanningConfiguration.scala
ragadeeshu: See the other cypher comment.

community/consistency-check/src/main/java/org/neo4j/consistency/internal/SchemaIndexExtensionLoader.java
ragadeeshu: This will be resolved when we do the other consistency checker-work.

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/IndexCheck.java
ragadeeshu: This is the bleeding edge towards the not-yet-supported consistency checker.

packaging/standalone/standalone-community/src/main/distribution/text/community/conf/neo4j.conf
ragadeeshu: The dangers of intellij renaming.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/PropertyCommandsExtractor.java
```diff
@@ -105,6 +112,17 @@ public boolean visitPropertyCommand( PropertyCommand command )
...
+            List<PropertyCommand> group = propertyCommandsByRelationshipIds.get( relId );
+            if ( group == null )
+            {
+                propertyCommandsByRelationshipIds.put( relId, group = new ArrayList<>() );
+            }
+            group.add( command );
+            hasUpdates = true;
+        }
         return false;
     }
```
klaren, line 130: `containsAnyNodeOrPropertyUpdate` affects relationships new as well. Rename to reflect this.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/ConstraintEnforcingEntityOperations.java
```diff
@@ -371,9 +372,11 @@ public long nodeGetFromUniqueIndexSeek(
             ExactPredicate... predicates )
             throws IndexNotFoundKernelException, IndexBrokenKernelException, IndexNotApplicableKernelException
     {
+        assertIndexHasOneEntityToken( index.schema() );
         assertIndexOnline( state, index );
         assertPredicatesMatchSchema( index.schema(), predicates );
-        int labelId = index.schema().getLabelId();
+        // This is fine because of the assertion above
```
klaren: Make this explicit. E.g change `assertIndexHasOneEntityToken` to `getOneLableIdOrThrow` and return the id from there.

community/kernel/src/main/java/org/neo4j/kernel/impl/store/SchemaStore.java
```diff
@@ -42,7 +43,7 @@
 import static java.util.Collections.singleton;
 import static org.neo4j.kernel.impl.store.record.RecordLoad.CHECK;
 
-public class SchemaStore extends AbstractDynamicStore implements Iterable<SchemaRule>
+public class SchemaStore extends AbstractDynamicStore// implements Iterable<SchemaRule>
```
klaren: Implementing comments is not supported :wink:

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/DefaultIndexProviderMap.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+package org.neo4j.kernel.impl.transaction.state;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Consumer;
+
+import org.neo4j.internal.kernel.api.IndexCapability;
```
klaren: Unused

community/kernel/src/main/java/org/neo4j/kernel/impl/transaction/state/DefaultIndexProviderMap.java
```diff
@@ -0,0 +1,104 @@
...
+            {
+                throw new IllegalArgumentException(
+                        "Tried to load multiple index providers with the same provider descriptor " + providerDescriptor + ". First loaded " + existing +
+                                " then " + provider );
+            }
+        }
+    }
+
+    @Override
+    public IndexProvider getDefaultSchemaIndexProvider()
```
klaren: Changes return type from specified to raw.

community/kernel/src/main/java/org/neo4j/unsafe/batchinsert/internal/BatchInserterImpl.java
```diff
@@ -521,31 +519,59 @@ private void repopulateAllIndexes() throws IOException, IndexEntryConflictExcept
...
             }
         }
     }
 
+    private int[] entityTokenIds( List<SchemaDescriptor> nodeDescriptors )
+    {
+        if ( nodeDescriptors.stream().anyMatch( descriptor -> descriptor.schema().getEntityTokenIds().length == 0 ) )
+        {
+            //No token is any token
+            return new int[0];
```
klaren: Constant

community/kernel/src/main/java/org/neo4j/kernel/impl/api/StateHandlingStatementOperations.java
```diff
@@ -726,19 +726,31 @@ public void constraintDrop( KernelStatement state, ConstraintDescriptor constrai
...
-                    state.txState().indexDiffSetsByLabel( descriptor.getLabelId() ).apply( rules ) );
+            rules = state.txState().indexDiffSetsBySchema( descriptor ).apply( rules );
         }
         return singleOrNull( rules );
     }
 
+    @Override
+    public IndexDescriptor indexGetForName( KernelStatement state, String name )
+    {
+        // This is only used for querying, and thus is is "fine" (at least for now) that it is not tx-state aware
```
klaren: At least assert that `state.hasTxStateWithChanges()` is false and throw otherwise.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextConfig.java
```diff
@@ -17,30 +17,21 @@
...
 import org.neo4j.configuration.LoadableConfig;
 import org.neo4j.graphdb.config.Setting;
 
-import static org.neo4j.kernel.configuration.Settings.BOOLEAN;
-import static org.neo4j.kernel.configuration.Settings.FALSE;
 import static org.neo4j.kernel.configuration.Settings.STRING;
 import static org.neo4j.kernel.configuration.Settings.setting;
 
 /**
  * Configuration parameters for the bloom fulltext addon.
```
klaren: Still referencing bloom.

community/consistency-check/src/main/java/org/neo4j/consistency/checking/full/PropertyAndNodeIndexedCheck.java
ragadeeshu: Arrays.stream with a int array creates a primitive int stream, so that should be fine, no? I updated NodeLabelReader to work with primitive sets.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexAccessor.java
ragadeeshu: Because the only use this will see is form consistency checker, and we do not support that yet.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexAccessor.java
ragadeeshu: This will be seen to when we add consistency checker support.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProvider.java
```diff
@@ -0,0 +1,230 @@
...
+        this.relationshipTypeTokenHolder = relationshipTypeTokenHolder;
+        this.indexStorageFactory = new IndexStorageFactory( directoryFactory, fileSystem, directoryStructure() );
+
+        analyzerClassName = config.get( FulltextConfig.fulltext_default_analyzer );
+        factory = new FulltextFactory( indexStorageFactory, analyzerClassName, config );
+        accessors = new HashMap<>();
+        accessorsByName = new HashMap<>();
+    }
+
+    public ScoreEntityIterator query( IndexDescriptor descriptor, String query ) throws IOException
```
klaren: Should probably be `FulltextIndexProvider` for the parameter as the map expects that.

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexProvider.java
ragadeeshu: I'm not sure, I think you should be able to compare different types of index providers?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProviderFactory.java
```diff
@@ -0,0 +1,90 @@
...
+import org.neo4j.kernel.impl.spi.KernelContext;
+
+import static org.neo4j.kernel.api.impl.index.LuceneKernelExtensions.directoryFactory;
+import static org.neo4j.kernel.api.index.IndexDirectoryStructure.directoriesByProvider;
+import static org.neo4j.kernel.api.index.IndexDirectoryStructure.directoriesBySubProvider;
+
+@Service.Implementation( KernelExtensionFactory.class )
+public class FulltextIndexProviderFactory extends KernelExtensionFactory<FulltextIndexProviderFactory.Dependencies>
+{
+    public static final String KEY = "fulltext";
```
klaren: `private`

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/FulltextFactory.java
```diff
@@ -17,48 +17,44 @@
...
     private final Analyzer analyzer;
+    private final Config config;
 
     /**
      * Creates a factory for the specified location and analyzer.
      *
-     * @param fileSystem The filesystem to use.
-     * @param storeDir Store directory of the database.
      * @param analyzerClassName The Lucene analyzer to use for the {@link LuceneFulltext} created by this factory.
+     * @param config
```
klaren: Missing javadoc

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/FulltextFactory.java
```diff
@@ -17,48 +17,44 @@
...
      * @throws IOException
```
klaren: Does it throw?

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/IndexingService.java
ragadeeshu: In what sense will they fight? They scan different stores and apply updates to different indexes?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/LuceneFulltextFieldEncoding.java
```diff
@@ -17,7 +17,7 @@
  * You should have received a copy of the GNU Affero General Public License
```
klaren: This class can be inlined, the static method has 1 usage which only pass the parameters as is.

enterprise/fulltext-addon/src/main/resources/META-INF/services/org.neo4j.kernel.extension.KernelExtensionFactory
```diff
@@ -1 +1 @@
-org.neo4j.kernel.api.impl.fulltext.integrations.bloom.BloomKernelExtensionFactory
\ No newline at end of file
+org.neo4j.kernel.api.impl.fulltext.integrations.kernel.FulltextIndexProviderFactory
```
klaren: missing newline

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomBackupIT.java
```diff
@@ -55,6 +56,8 @@
 {
```
klaren, line 56: Is this bloom specific or could it be renamed to fulltext?

klaren (changes requested):
My first iteration.

General comments:
* We talked about having the full-text index be configured to be eventually consistent or not. Is this planned?
* "Non-schema" that is accessed through the schema makes me a bit confused. Maybe we can come up with a better name.
* Raw types should be avoided in new code. Some effort needs to be dedicated to this.
* Still a lot of TODOs left, I did not mention them since I'm assuming those will be acted upon before this is merged.
* Should we keep the bloom name and have it live side by side or migrate bloom to use the final surface later?

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomProcedures.java
ragadeeshu: The exception we catch here is specifically schemarulenotfound, which means that we haven't created the bloom index yet (or we have deleted it), and thus we do not have any indexed properties and an empty stream is the correct result.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/LuceneFulltext.java
ragadeeshu: While I see where you are coming from, there is also a lot of stuff in LuceneSchemaIndex that we do not want, such as verifying uniqueness and dealing with IndexReader. The things that are identical are isOnline, markAsOnline and markAsFailed. Perhaps we should extract those to the AbstractLuceneIndex instead.

enterprise/fulltext-addon/src/test/java/org/neo4j/kernel/api/impl/fulltext/integrations/bloom/BloomBackupIT.java
ragadeeshu: Bloom name makes sense, as it uses the bloom surface.

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/PropertyCommandsExtractor.java
ragadeeshu: updated

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/integrations/kernel/FulltextIndexProvider.java
ragadeeshu: I did it the other way around, hope that works for you

enterprise/fulltext-addon/src/main/resources/META-INF/services/org.neo4j.kernel.extension.KernelExtensionFactory
ragadeeshu: fixed

community/kernel/src/main/java/org/neo4j/kernel/impl/api/index/MultipleIndexPopulator.java
ragadeeshu: Replaced with a constant for now, feels a bit better at least.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/SimpleFulltextReader.java
ragadeeshu: Good catch, we need to put something better here.

community/kernel/src/main/java/org/neo4j/kernel/api/index/IndexEntryUpdate.java
ragadeeshu: fixed

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/LuceneFulltext.java
ragadeeshu: Extracted.

enterprise/fulltext-addon/src/main/java/org/neo4j/kernel/api/impl/fulltext/lucene/SimpleFulltextReader.java
ragadeeshu: Upon further consideration, I think we should just throw this guy, since in some cases the query might be user-provided. If we are to catch it , we should do so in the bloom surface... Maybe.

PR #11151, opened 2018-03-02
Fixed bug with date types stored after other properties

craigtaverner:


community/kernel/src/test/java/org/neo4j/kernel/impl/core/TestCombinedPropertyTypes.java
```diff
@@ -0,0 +1,117 @@
...
+import org.neo4j.graphdb.Node;
+import org.neo4j.kernel.impl.AbstractNeo4jTestCase;
+import org.neo4j.values.storable.DateTimeValue;
+import org.neo4j.values.storable.DateValue;
+import org.neo4j.values.storable.LocalTimeValue;
+import org.neo4j.values.storable.Value;
+
+import static org.junit.Assert.assertEquals;
+
+public class TestCombinedPropertyTypes extends AbstractNeo4jTestCase
```
sherfert, line 37: I really like these tests and it seems like a very useful thing to have for all our types (basically any two different ones in succession). Nothing that this PR needs to have, but it feels like this behavior has been largely untested.

sherfert (approved):

PR #11152, opened 2018-03-02
Native string index in fusion index

tinwelint:


tinwelint:
Depends on #11180

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexBase.java
```diff
@@ -0,0 +1,172 @@
...
+ * Acting as a simplifier for the multiplexing that is going in inside a fusion index. A fusion index consists of multiple parts,
+ * each handling one or more value groups. Each instance, be it a reader, populator or accessor should extend this class
+ * to get that multiplexing at a low cost. All parts will live in an array with specific slot constants to each specific part.
+ *
+ * @param <T> type of instances
+ */
+public abstract class FusionIndexBase<T>
+{
+    private static final int INSTANCE_COUNT = 5;
+    private static final String[] NAMES = { "string", "number", "spatial", "temporal", "lucene" };
```
MishaDemianenko: this is quite an ordering assumption that will be easy to break in future, since there is indirect dependency from order defined in FusionSchemaIndexProvider array creation in the constructor. When javadoc claims that they will live in their own part you do not really expect that those part numbers are hardcoded. Would be really good to have that contract enforced somehow: additional checks or update structure to not really on hard-coded indexes and positions

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexReader.java
```diff
@@ -30,86 +30,79 @@
...
         }
 
-        if ( predicates[0] instanceof GeometryRangePredicate )
+        if ( predicate instanceof GeometryRangePredicate )
         {
-            return spatialReader.query( predicates );
+            return instances[SPATIAL].query( predicates );
         }
 
 // TODO: support temporal range queries
```
MishaDemianenko, line 110: do you guys know when/who will gonna fix those todos or we can remove them?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/NumberHitIterator.java
```diff
@@ -55,9 +56,12 @@ protected boolean fetchNext()
     {
         try
         {
-            if ( seeker.next() )
+            while ( seeker.next() )
             {
-                return next( seeker.get().key().getEntityId() );
+                if ( acceptValue( seeker.get().key().asValue() ) )
```
MishaDemianenko: you can store key into intermediate variable here like:
```
                KEY key = seeker.get().key();
                if ( acceptValue( key.asValue() ) )
                {
                    return next( key.getEntityId() );
                }
```

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/StringSchemaKey.java
```diff
@@ -156,6 +173,7 @@ public void writeString( String value )
     @Override
     public void writeString( char value )
     {
-        throw new UnsupportedOperationException( "Not supported a.t.m. should it be?" );
+        // TODO more efficient impl
```
MishaDemianenko: do not think this todo should be here

community/lucene-index/src/main/java/org/neo4j/kernel/api/impl/schema/NativeLuceneFusionSchemaIndexProviderFactory.java
```diff
@@ -87,7 +88,9 @@ public static FusionSchemaIndexProvider newInstance( PageCache pageCache, File s
     {
         IndexDirectoryStructure.Factory childDirectoryStructure = subProviderDirectoryStructure( storeDir );
         boolean readOnly = isReadOnly( config, operationalMode );
-        NumberSchemaIndexProvider nativeProvider =
+        StringSchemaIndexProvider stringProvider =
```
MishaDemianenko: should we introduce a feature flag that will gonna allow us to switch native sting indexes and fallback to lucene?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FilteringNativeHitIndexProgressor.java
```diff
@@ -0,0 +1,53 @@
...
+import org.neo4j.cursor.RawCursor;
+import org.neo4j.index.internal.gbptree.Hit;
+import org.neo4j.internal.kernel.api.IndexQuery;
+import org.neo4j.values.storable.Value;
+
+class FilteringNativeHitIndexProgressor<KEY extends NativeSchemaKey, VALUE extends NativeSchemaValue> extends NativeHitIndexProgressor<KEY,VALUE>
+{
+    private final IndexQuery[] filter;
+
+    FilteringNativeHitIndexProgressor( RawCursor<Hit<KEY,VALUE>,IOException> seeker, NodeValueClient client,
```
MishaDemianenko, line 34: would b good to have separate test for this one

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/FilteringNumberHitIterator.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.io.IOException;
+import java.util.Collection;
+
+import org.neo4j.cursor.RawCursor;
+import org.neo4j.index.internal.gbptree.Hit;
+import org.neo4j.internal.kernel.api.IndexQuery;
+import org.neo4j.values.storable.Value;
+
+class FilteringNumberHitIterator<KEY extends NativeSchemaKey, VALUE extends NativeSchemaValue> extends NumberHitIterator<KEY,VALUE>
```
MishaDemianenko: and for this one

MishaDemianenko (changes requested):
Have couple of small comments/questions here and there

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexBase.java
tinwelint: Yes, it's a bit annoying... it's only used for allowing to for-loop over sub providers when collecting index failures, i.e. for throwing `IllegalStateException` if that method is called at the wrong state. I'll actually hard code that in there instead, or?

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexReader.java
tinwelint: This TODO is from code that was just moved around in this PR. Pretty sure there's a PR up now (Temporal IV) that will implement those

tinwelint:
@MishaDemianenko have another look now

community/kernel/src/main/java/org/neo4j/kernel/impl/index/schema/fusion/FusionIndexReader.java
MishaDemianenko: ok

MishaDemianenko (approved):

PR #11153, opened 2018-03-04
Improved couple Bolt tests

lutovich:
* Stop all threads after `TransportWriteThrottleTest`
 Test method `#shouldLockWhenNotWritable()` uses a separate thread to acquire a throttle lock when channel is not writable. This thread was not stopped and spun in the background generating garbage. It invoked mock of `ThrottleLock#lock()` in a tight loop. Every mock invocation is recorded and creates a stacktrace. This caused test to sometimes make forked VM to crash with OOM.
    This commit makes the test stop all threads and sleep a bit on every invocation of `ThrottleLock#lock()` to create less mock invocations.

* Remove test that contains no assertions


ali-ince (approved):
:+1:

MishaDemianenko:
Thx guys!

PR #11158, opened 2018-03-05
Public static fields should also be final

klaren:
Also removed unnecessary boxing of booleans.


community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
```diff
@@ -36,9 +36,9 @@
     public static final int STATUS_SUCCESS = 0;
     public static final int STATUS_ERROR = 1;
     // These fields are mutable for in-process AdminTool testing purposes.
-    public static String NEO4J_HOME = System.getenv().getOrDefault( "NEO4J_HOME", "" );
-    public static String NEO4J_CONF = System.getenv().getOrDefault( "NEO4J_CONF", "" );
-    public static String NEO4J_DEBUG = System.getenv().getOrDefault( "NEO4J_DEBUG", "" );
+    public static final String NEO4J_HOME = System.getenv().getOrDefault( "NEO4J_HOME", "" );
+    public static final String NEO4J_CONF = System.getenv().getOrDefault( "NEO4J_CONF", "" );
+    public static final String NEO4J_DEBUG = System.getenv().getOrDefault( "NEO4J_DEBUG", "" );
```
chrisvest, line 40: These are mutable for testing, and exploited in the blockdevice repo, I believe.

chrisvest: The comment just above also says so.

klaren: Well spotted. Although, why can't the affected tests instantiate `AdminTool` directly as all other tests do?

chrisvest: Maybe that wasn't possible at that time.

enterprise/com/src/main/java/org/neo4j/com/Client.java
```diff
@@ -208,7 +208,7 @@ public void start()
             protected ChannelContext create()
             {
                 msgLog.info( threadInfo() + "Trying to open a new channel from " + origin + " to " + destination,
-                        true );
+                        Boolean.TRUE );
```
MishaDemianenko: we should be able to just remove those true's from log messages in this and other classes that do the same thing

community/common/src/main/java/org/neo4j/function/Predicates.java
```diff
@@ -262,7 +262,7 @@ public static void awaitForever( BooleanSupplier condition, long checkInterval,
         };
     }
 
-    public static IntPredicate ALWAYS_TRUE_INT = v -> true;
+    public static final IntPredicate ALWAYS_TRUE_INT = v -> true;
```
MishaDemianenko: please move this constant from the middle of the class to the top

MishaDemianenko (changes requested):

MishaDemianenko:
Also, the thing that Chris mentioned if its not a problem anymore comment on top of those fields should be removed

chrisvest:
@MishaDemianenko I agree. Let's remove the comment, make those fields final, and fix up the blockdevice repo.

community/command-line/src/main/java/org/neo4j/commandline/admin/AdminTool.java
klaren: Should hopefully be possible now, I'll have a look at it.

klaren:
@MishaDemianenko addressed. I'm also looking into blockdevice to fix compilation errors.

MishaDemianenko (approved):

PR #11161, opened 2018-03-05
Fix bug that concatenation of covariant lists would specify wrong type

### LINKED ISSUE #10978: REDUCE function giving error.
The REDUCE function is not able to execute when iterating over a concatenation of two sub lists. When I tried to get the sum of the elements of two separate lists it throws error of type mismatch.

**Neo4j Version:** 3.2.9  
**Operating System:** Windows 10
**API:** The built-in web app running over bolt

### Steps to reproduce
1. Start the database service on your local system
2. Go to the neo4j console in your web browser (localhost:7474)
3. Execute the below sample query
`RETURN REDUCE(s = 0, p in [5,8,2,9] + [1,2] | s + p)`

### Expected behavior
It should return one 27 which is the sum of all the elements in both list.

### Actual behavior
It throws the below error 
`Type mismatch: accumulator is Integer but expression has type Float (line 1, column 57 (offset: 56))
"RETURN REDUCE(s = 0, p in [5,8,2,9] + [1,2] | s + p)"
                                                     ^`

sherfert:
The SemanticChecker would compute coercions of covariant types wrong.
This led to problems when specifying the type of add(List<?>, List<?>).

Fixes #10978

### DIFF
--- a/community/cypher/frontend-3.2/src/main/scala/org/neo4j/cypher/internal/frontend/v3_2/symbols/IntegerType.scala
+++ b/community/cypher/frontend-3.2/src/main/scala/org/neo4j/cypher/internal/frontend/v3_2/symbols/IntegerType.scala
@@ -22,7 +22,7 @@ package org.neo4j.cypher.internal.frontend.v3_2.symbols
 object IntegerType {
   val instance = new IntegerType() {
     val parentType = CTNumber
-    override lazy val coercibleTo: Set[CypherType] = Set(CTFloat)
+    override lazy val coercibleTo: Set[CypherType] = Set(CTFloat) ++ parentType.coercibleTo
     override val toString = "Integer"
     override val toNeoTypeString = "INTEGER?"
 
--- a/community/cypher/frontend-3.2/src/test/scala/org/neo4j/cypher/internal/frontend/v3_2/ast/AddTest.scala
+++ b/community/cypher/frontend-3.2/src/test/scala/org/neo4j/cypher/internal/frontend/v3_2/ast/AddTest.scala
@@ -57,6 +57,25 @@ class AddTest extends InfixExpressionTestBase(Add(_, _)(DummyPosition(0))) {
 
     testValidTypes(CTNode, CTList(CTNode))(CTList(CTNode))
     testValidTypes(CTFloat, CTList(CTFloat))(CTList(CTFloat))
+
+    testValidTypes(CTList(CTAny), CTList(CTAny))(CTList(CTAny))
+  }
+
+  test("should handle covariant types") {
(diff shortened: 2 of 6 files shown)


community/cypher/frontend-3.2/src/test/scala/org/neo4j/cypher/internal/frontend/v3_2/ast/AddTest.scala
```diff
@@ -57,6 +57,25 @@ class AddTest extends InfixExpressionTestBase(Add(_, _)(DummyPosition(0))) {
...
+  }
+
+  test("should handle covariant types") {
+    testValidTypes(CTString.covariant, CTString.covariant)(CTString)
+    testValidTypes(CTString.covariant, CTInteger)(CTString)
+    testValidTypes(CTString, CTInteger.covariant)(CTString)
+    testValidTypes(CTInteger.covariant, CTInteger.covariant)(CTInteger)
+    testValidTypes(CTInteger.covariant, CTFloat)(CTFloat)
+    testValidTypes(CTInteger, CTFloat.covariant)(CTFloat)
+    testValidTypes(CTFloat.covariant, CTFloat.covariant)(CTFloat.covariant)
```
Lojjs: `CTFloat = CTFloat.covariant` so use `CTFloat` as expected here for consistency with the String test

community/cypher/frontend-3.2/src/main/scala/org/neo4j/cypher/internal/frontend/v3_2/symbols/IntegerType.scala
```diff
@@ -22,7 +22,7 @@ package org.neo4j.cypher.internal.frontend.v3_2.symbols
 object IntegerType {
   val instance = new IntegerType() {
     val parentType = CTNumber
-    override lazy val coercibleTo: Set[CypherType] = Set(CTFloat)
+    override lazy val coercibleTo: Set[CypherType] = Set(CTFloat) ++ parentType.coercibleTo
```
Lojjs, line 25: Neither `NumberType` nor `AnyType` do override `coercibleTo`, so the ` ++ parentType.coercibleTo` will add an empty list from the `CypherType` implementation, i.e. have no effect. What was the reason for adding this? Is it if anyone later changes any of the superclasses' implementations?

sherfert: Excatly. The idea behind `coercibleTo` is to specify your coercions and then add those of all parent types (which are empty as of now). I actually forgot that in `ListType`, gonna add it there too.

Lojjs (approved):

PR #11164, opened 2018-03-05
Reworked index drop during ongoing population.

MishaDemianenko:
Allow index to be dropped during population without waiting ongoing
population job to fully completed.
In the cases when population job have more than 1 index to populate allow those
populations to continue, before all of those would be canceled for
no reason if one of the indexes where dropped.
Do not use futures-timeouts for ongoing index drop/close operations.


tinwelint (approved):
Looks good

